PR #960, opened 2021-07-01
[optimizer][CBO] count(*) with table statistics

### LINKED ISSUE #914: [optimizer][CBO] count(*) with table statistics
**Summary**

For query `select count(*) from system.numbers(100000)`, ReadDataSourcePlan has `Statistics`:
https://github.com/datafuselabs/datafuse/blob/28c8e3eda9fb8de881635343d3225b3e7be07b4b/common/planners/src/plan_read_datasource.rs#L17-L26

In this case, we can use the `read_rows` as the final `count`:
https://github.com/datafuselabs/datafuse/blob/28c8e3eda9fb8de881635343d3225b3e7be07b4b/common/planners/src/plan_statistics.rs#L6-L11

We need a `StatisticsConstant` optimizer rule, like:
https://github.com/apache/arrow-datafusion/pull/620/files

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

## Summary

Summary about this PR

#914 

Please note that currently we can optimize `count(*)` and `count(0)` but not `count()` now, it's ugly to write it in an optimizer, we can do it later.

```
mysql> select count(*) from numbers_mt(100000000);
+-----------+
| count(0)  |
+-----------+
| 100000000 |
+-----------+
1 row in set (0.01 sec)

mysql> explain select count(*) from numbers_mt(100000000);
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| explain                                                                                                                                                                                                                                                                                                                                           |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Projection: count(0):UInt64
  AggregatorFinal: groupBy=[[]], aggr=[[count(0)]]
    Projection: {"Struct":[{"UInt64":100000000}]} as count(0):Utf8
      Expression: {"Struct":[{"UInt64":100000000}]}:Utf8 (Exact Statistics)
        ReadDataSource: scan partitions: [1], scan schema: [dummy:UInt8], statistics: [read_rows: 1, read_bytes: 1] |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

## Changelog

- Improvement
- Performance Improvement

## Related Issues

Fixes #914

## Test Plan

- Unit Test


TennyZhuang:
@BohuTANG @sundy-li @zhang2014

fusequery/query/src/optimizers/optimizer_statistics_exact.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct StatisticsExactOptimizer {
+    ctx: FuseQueryContextRef,
+}
+
+impl<'plan> PlanRewriter<'plan> for StatisticsExactImpl<'_> {
+    fn rewrite_aggregate_partial(
+        &mut self,
+        plan: &'plan AggregatorPartialPlan,
+    ) -> Result<PlanNode> {
+        let new_plan = match (
```
zhang2014, line 30: What happens when execute following queries?
```sql
SELECT COUNT() FROM numbers(1000) WHERE number = 500;
SELECT COUNT() FROM numbers(1000) HAVING number = 500;
```

TennyZhuang: Nothing will happen, we will only try to match:

```
    AggregatorPartial: groupBy=[[]], aggr=[[COUNT(whatever)]]
      Expression: whatever
         ReadDataSource: whatever with exact statistics
```

Any filter will disable the optimization

TennyZhuang: What'more, `count(0)` is required, because `count()` wil missing a (Before GroupBy) Expression.

We can write another optimizer to omit any arguments in count, and make this more generic

sundy-li:
How about introducing a new setting to control this optimization, like `set optimize_trivial_count_query = 1`, defaults to 1.

TennyZhuang:
> How about introducing a new setting to control this optimization, like `set optimize_trivial_count_query = 1`, defaults to 1.
> 
> We may need this because we need the perf results without this optimization.

How about introducing a general switch for every optimizer rule?

BohuTANG:
> > How about introducing a new setting to control this optimization, like `set optimize_trivial_count_query = 1`, defaults to 1.
> > We may need this because we need the perf results without this optimization.
> 
> How about introducing a general switch for every optimizer rule? not only for count_query

Right, min/max ...

fusequery/query/src/optimizers/optimizer_statistics_exact.rs
sundy-li: It' ok.  BTW how about checking the nullable for the first expression?

http://sqlfiddle.com/#!9/34d4ba/5

sundy-li (changes requested):

fusequery/query/src/optimizers/optimizer_statistics_exact.rs
TennyZhuang: Good catch, so we can only omit count(args) with non-null fields. We should check the aggr.rule

Or for simple, we can only optimize for `count(literal)` and `count()`

sundy-li: I prefer just to optimize `count(literal)` and `count()` is ok.  

People who use count(other_expressions) may do not like the optimization, like `select count(sleep(2)) from numbers(10)`

TennyZhuang: Fixed

sundy-li (approved):
LGTM, thanks @TennyZhuang

PR #978, opened 2021-07-03
Improve group by modulo

sundy-li:

## Summary

Improve group by modulo by using the proper data_type_coercion 

## Changelog

 
- Improvement
 
## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
```
mysql>  select 256   %  3;
+-----------+
| (256 % 3) |
+-----------+
|      NULL |
+-----------+
1 row in set (0.01 sec)
```

We should allow force cast even with the loss of accuracy.

https://github.com/rust-num/num-traits/blob/master/src/cast.rs#L136-L164

BohuTANG (approved):
LGTM
Thank you @sundy-li

Issue #271, opened 2021-04-08
FuseStore v0.1 


drmingdrmer:
### Roadmap of FuseStore

[project store](https://github.com/datafuselabs/datafuse/projects/21) tracks the detail progress.

- [x] Define metadata types with protobuf. Build an in-memory storage engine, for testing purposes.
- [ ] Setup consensus group to store meta data.
- [x] . . Basic replication logical impl Leader/Follower/NonVoter management.
- [x] . . Apply create database/table action to raft metadata cluster.
- [x] . . #993
- [x] . . . . #775
- [ ] . . test multi candidates cluster and leadership transfer.
- [ ] . . #1125
- [x] . . . . #1117
- [ ] . . Disk backed raft.
- [x] . . . . #1056
- [x] . . . . #1065
- [x] . . . . #1230
- [ ] . . . . #1080
- [ ] . . . . #1081
- [ ] . . . . [store] test: Test: restarting a cluster
- [x] Catalog design @dantengsky
- [ ] . . Flight API @dantengsky
- [ ] Complete the cluster deployment workflow. **PLANNING** @drmingdrmer
- [x] . . #1250
- [ ] . . #1286 @ariesdevil 
- [ ] Store quota and isolation #1127
- [ ] anything else 🤔 ?

BohuTANG:
Add #1127

BohuTANG:
@drmingdrmer this issue maybe should rename or we can close it?

drmingdrmer:
Just close it XDDD

drmingdrmer:
G

PR #984, opened 2021-07-04
[store]: refactor: Cmd::UpsertKV use MatchSeq to specify what seq to match

drmingdrmer:

##### [store]: refactor: Cmd::UpsertKV use MatchSeq to specify what seq to match
The fully functional UpsertKV now supports all write operations.

Since a sequence number is always positive:
- `MatchSeq::Exact(0)` to perform an add-if-absent operation.
- `MatchSeq::Exact(n)` to perform an update on some specified version.
- `MatchSeq::GE(1)` to perform an update-any operation.
- `MatchSeq::Any` to perform an update or insert that always takes effect.

59 LOC reduced! 

## Changelog

- Improvement

## Related Issues

#271

### DIFF
--- a/common/flights/src/impls/kv_api_impl.rs
+++ b/common/flights/src/impls/kv_api_impl.rs
@@ -5,6 +5,7 @@
 
 use common_exception::ErrorCode;
 use common_exception::Result;
+use common_metatypes::MatchSeq;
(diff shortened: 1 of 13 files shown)


dantengsky (approved):
LGTM!

PR #992, opened 2021-07-05
Tracking: Catalog (issue #867)

dantengsky:
I hereby agree to the terms of the CLA available at: https://datafuse.rs/policies/cla/

## Summary

TBD

## Changelog

- New Feature

## Related Issues

Fixes #835
Fixes #867

## Test Plan

Unit Tests

Stateless Tests

PR #1023, opened 2021-07-08
Change database, table and data parts operations from meta to meta_node

ariesdevil:

## Summary

change database, table and data parts operations from meta to meta_node

## Changelog

- Improvement

## Related Issues

https://github.com/datafuselabs/datafuse/issues/993
https://github.com/datafuselabs/datafuse/issues/775

### DIFF
--- a/fusestore/store/src/engine/mem_engine.rs
+++ b/fusestore/store/src/engine/mem_engine.rs
@@ -105,7 +105,9 @@ impl MemEngine {
         let table_id = self
             .dbs
             .get(&cmd.db_name)
-            .ok_or_else(|| ErrorCode::UnknownDatabase("database not found yet"))?
+            .ok_or_else(|| {
+                ErrorCode::UnknownDatabase(format!("database not found: {}", &cmd.db_name))
+            })?
             .table_name_to_id
             .get(&cmd.table_name);
 
@@ -170,7 +172,7 @@ impl MemEngine {
         table_name: String,
     ) -> common_exception::Result<Table> {
         let db = self.dbs.get(&db_name).ok_or_else(|| {
-            ErrorCode::UnknownDatabase(format!("database not found yet: {:}", db_name))
+            ErrorCode::UnknownDatabase(format!("database not found: {:}", db_name))
         })?;
 
         let table_id = db
--- a/fusestore/store/src/engine/mem_engine.rs
+++ b/fusestore/store/src/engine/mem_engine.rs
@@ -105,7 +105,7 @@ impl MemEngine {
         let table_id = self
             .dbs
             .get(&cmd.db_name)
-            .ok_or_else(|| ErrorCode::UnknownDatabase("database not found"))?
+            .ok_or_else(|| ErrorCode::UnknownDatabase("database not found yet"))?
             .table_name_to_id
             .get(&cmd.table_name);
 
@@ -165,12 +165,12 @@ impl MemEngine {
     }
 
     pub fn get_table(
-        &mut self,
+        &self,
         db_name: String,
         table_name: String,
     ) -> common_exception::Result<Table> {
         let db = self.dbs.get(&db_name).ok_or_else(|| {
-            ErrorCode::UnknownDatabase(format!("database not found: {:}", db_name))
+            ErrorCode::UnknownDatabase(format!("database not found yet: {:}", db_name))
         })?;
 
         let table_id = db
--- a/fusestore/store/src/executor/action_handler.rs
+++ b/fusestore/store/src/executor/action_handler.rs
@@ -17,7 +17,6 @@ use common_flights::storage_api_impl::AppendResult;
 use common_flights::storage_api_impl::ReadAction;
 use common_flights::RequestFor;
 use common_flights::StoreDoAction;
-use common_infallible::Mutex;
 use common_planners::PlanNode;
 use common_runtime::tokio::sync::mpsc::Sender;
 use futures::Stream;
@@ -27,7 +26,6 @@ use tonic::Status;
 use tonic::Streaming;
 
 use crate::data_part::appender::Appender;
-use crate::engine::MemEngine;
 use crate::fs::FileSystem;
 use crate::meta_service::MetaNode;
 
@@ -38,7 +36,6 @@ pub trait ReplySerializer {
 }
 
 pub struct ActionHandler {
-    pub(crate) meta: Arc<Mutex<MemEngine>>,
     /// The raft-based meta data entry.
     /// In our design meta serves for both the distributed file system and the catalog storage such as db,tabel etc.
     /// Thus in case the `fs` is a Dfs impl, `meta_node` is just a reference to the `Dfs.meta_node`.
@@ -60,11 +57,7 @@ where T: RequestFor
 
 impl ActionHandler {
     pub fn create(fs: Arc<dyn FileSystem>, meta_node: Arc<MetaNode>) -> Self {
-        ActionHandler {
-            meta: MemEngine::create(),
-            meta_node,
-            fs,
-        }
+        ActionHandler { meta_node, fs }
     }
 
     /// Handle pull-file request, which is used internally for replicating data copies.
@@ -142,8 +135,9 @@ impl ActionHandler {
         parts: Streaming<FlightData>,
     ) -> common_exception::Result<AppendResult> {
         {
-            let mut meta = self.meta.lock();
-            let _tbl_meta = meta.get_table(db_name.clone(), table_name.clone())?;
+            // TODO(ariesdevil): use meta_node instead
+            // let mut meta = self.meta.lock();
+            // let _tbl_meta = meta.get_table(db_name.clone(), table_name.clone())?;
 
             // TODO:  Validates the schema of input stream:
             // The schema of `parts` should be a subset of
@@ -159,8 +153,12 @@ impl ActionHandler {
             .append_data(format!("{}/{}", &db_name, &table_name), Box::pin(parts))
             .await?;
 
-        let mut meta = self.meta.lock();
-        meta.append_data_parts(&db_name, &table_name, &res);
+        // let mut meta = self.meta.lock(); //todo(ariesdevil): change to meta_node
+        // meta.append_data_parts(&db_name, &table_name, &res);
+        // Ok(res)
+        self.meta_node
+            .append_data_parts(&db_name, &table_name, &res)
+            .await;
         Ok(res)
     }
 
--- a/fusestore/store/src/executor/action_handler.rs
+++ b/fusestore/store/src/executor/action_handler.rs
@@ -135,10 +135,6 @@ impl ActionHandler {
         parts: Streaming<FlightData>,
     ) -> common_exception::Result<AppendResult> {
         {
-            // TODO(ariesdevil): use meta_node instead
-            // let mut meta = self.meta.lock();
-            // let _tbl_meta = meta.get_table(db_name.clone(), table_name.clone())?;
-
             // TODO:  Validates the schema of input stream:
             // The schema of `parts` should be a subset of
             // table's current schema (or following the evolution rules of table schema)
--- a/fusestore/store/src/executor/action_handler_test.rs
+++ b/fusestore/store/src/executor/action_handler_test.rs
@@ -79,7 +79,7 @@ async fn test_action_handler_add_database() -> anyhow::Result<()> {
     }
 
     /// helper to build a T
-    fn case(dbname: &str, if_not_exists: bool, want: common_exception::Result<i64>) -> T {
+    fn case(dbname: &str, if_not_exists: bool, want: common_exception::Result<u64>) -> T {
         let plan = CreateDatabasePlan {
             db: dbname.to_string(),
             if_not_exists,
@@ -96,16 +96,15 @@ async fn test_action_handler_add_database() -> anyhow::Result<()> {
         T { plan, want }
     }
 
-    // TODO: id should be started from 1
     let cases: Vec<T> = vec![
-        case("foo", false, Ok(0)),
-        case("foo", true, Ok(0)),
+        case("foo", false, Ok(1)),
+        case("foo", true, Ok(1)),
         case(
             "foo",
             false,
             Err(ErrorCode::DatabaseAlreadyExists("foo database exists")),
         ),
-        case("bar", true, Ok(1)),
+        case("bar", true, Ok(2)),
     ];
 
     {
@@ -149,7 +148,7 @@ async fn test_action_handler_get_database() -> anyhow::Result<()> {
     }
 
     /// helper to build a T
-    fn case(db_name: &'static str, want: Result<i64, &str>) -> T {
+    fn case(db_name: &'static str, want: Result<u64, &str>) -> T {
         let want = match want {
             Ok(want_db_id) => Ok(GetDatabaseActionResult {
                 database_id: want_db_id,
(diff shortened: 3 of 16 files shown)


fusestore/store/src/engine/mem_engine.rs
```diff
@@ -105,7 +105,7 @@ impl MemEngine {
         let table_id = self
             .dbs
             .get(&cmd.db_name)
-            .ok_or_else(|| ErrorCode::UnknownDatabase("database not found"))?
+            .ok_or_else(|| ErrorCode::UnknownDatabase("database not found yet"))?
```
drmingdrmer, line 108: Add database name to the context?
```
            .ok_or_else(|| ErrorCode::UnknownDatabase(format!("database not found yet: {:}", db_name)))?
```

fusestore/store/src/executor/action_handler.rs
```diff
@@ -142,8 +135,9 @@ impl ActionHandler {
         parts: Streaming<FlightData>,
     ) -> common_exception::Result<AppendResult> {
         {
-            let mut meta = self.meta.lock();
-            let _tbl_meta = meta.get_table(db_name.clone(), table_name.clone())?;
+            // TODO(ariesdevil): use meta_node instead
+            // let mut meta = self.meta.lock();
+            // let _tbl_meta = meta.get_table(db_name.clone(), table_name.clone())?;
```
drmingdrmer, line 140: Why keep these comments? 🤔

fusestore/store/src/meta_service/state_machine.rs
```diff
@@ -75,6 +79,9 @@ pub struct StateMachine {
     /// table id to table mapping
     pub tables: BTreeMap<u64, Table>,
 
+    ///
+    pub tbl_parts: HashMap<String, HashMap<String, Vec<DataPartInfo>>>,
```
ariesdevil, line 83: @drmingdrmer Should I remove this field and replace table<->parts management to parts field in `Table`.

fusestore/store/src/executor/meta_handlers.rs
```diff
@@ -23,40 +23,61 @@ use common_flights::meta_api_impl::GetDatabaseAction;
...
+                db: Database {
+                    database_id: 0,
+                    tables: HashMap::new(),
+                },
+            },
         };
 
-        let database_id = meta.create_database(cmd, plan.if_not_exists)?;
+        let rst = self.meta_node.write(cr).await.map_err(|e| {
+            ErrorCode::MetaNodeInternalError(format!("not a Database result: {}", e.to_string()))
```
drmingdrmer, line 62: What does it mean `not a Database result`?

fusestore/store/src/executor/meta_handlers.rs
```diff
@@ -91,51 +109,78 @@ impl RequestHandler<DropDatabaseAction> for ActionHandler {
...
+            AppliedState::Table { result, .. } if result.is_some() => Ok(CreateTableActionResult {
+                table_id: result.unwrap().table_id,
+            }),
+            AppliedState::Table { prev, .. } if prev.is_some() => Ok(CreateTableActionResult {
+                table_id: prev.unwrap().table_id,
+            }),
+            AppliedState::Table { prev, result } if (prev.is_none() && result.is_none()) => Err(
+                ErrorCode::TableAlreadyExists(format!("table exists: {}", table_name)),
+            ),
+            _ => Err(ErrorCode::MetaNodeInternalError("not a Table result")),
```
drmingdrmer, line 182: The pattern `_` matches negative conditions: If it is a `AppliedState::Table { result, .. }` but result is None, it should not return an error `not a Table result`.

drmingdrmer (commented):
Nice and sharp!

fusestore/store/src/meta_service/state_machine.rs
drmingdrmer: Yeah I'd like to see a non-nested mapping :DDD

If you'd like to move things forward quickly, it's all right to open another PR fixing this issue.

fusestore/store/src/executor/meta_handlers.rs
ariesdevil: Here's a quote from a previous writeup, `not a Database result` means it's not a database operation error, maybe this error occurred by meta service or something else.

drmingdrmer: I think the `not a Database result` is a bit misleading.
Here it may be a network error or else.

drmingdrmer: I think the `not a Database result` is a bit misleading.
Here it may be a network error or else.

fusestore/store/src/meta_service/state_machine.rs
ariesdevil: Created a related issue: https://github.com/datafuselabs/datafuse/issues/1102

fusestore/store/src/executor/action_handler_test.rs
```diff
@@ -65,6 +77,64 @@ async fn test_action_handler_do_pull_file() -> anyhow::Result<()> {
     Ok(())
 }
 
+struct D {
+    plan: CreateDatabasePlan,
+    want: common_exception::Result<CreateDatabaseActionResult>,
+}
+
+/// helper to build a D
+fn case_db(db_name: &str, if_not_exists: bool, want: common_exception::Result<u64>) -> D {
```
drmingdrmer, line 86: As a nested type I gave it a short name because it is unlikely to have a name conflict with others.
As a mod-level struct, a single letter `D` is too short as a name.
I'd suggest keeping it a nested type or giving it an explicit and meaningful name.

fusestore/store/src/executor/action_handler_test.rs
```diff
@@ -65,6 +77,64 @@ async fn test_action_handler_do_pull_file() -> anyhow::Result<()> {
...
+        Err(err) => Err(err), // Result<i64,_> to Result<StoreDoActionResult, _>
+    };
+
+    D { plan, want }
+}
+
+struct T {
+    plan: CreateTablePlan,
+    want: common_exception::Result<CreateTableActionResult>,
+}
```
drmingdrmer, line 106: Also, refactoring code within a `new-feature` PR may not be a good practice.

Another single PR that "Just move code, nothing really changed" is much easier to review than a PR with old codes mixed within new codes.

In this case, as you need to reuse some of the old codes, just copy-paste them to a new standalone mod level function and leaving a TODO for refactoring the duplicated stuff would be comfortable for readers.

fusestore/store/src/meta_service/state_machine.rs
```diff
@@ -289,13 +289,13 @@ impl StateMachine {
...
             } => {
-                let db = self.databases.get(db_name);
-                let db = db.unwrap(); // unsafe?
+                let db = self.databases.get_mut(db_name).unwrap();
 
-                let table_id = db.tables.get(table_name);
-                let table_id = table_id.unwrap();
-                let prev = self.tables.get(table_id).cloned();
+                db.tables.remove(table_name);
+                let prev = self.tables.remove(table_id);
```
drmingdrmer: Why do we need both table name and table id?
I suppose `db.tables[table_name]` is the table id.

drmingdrmer:
With these tests I feel much safer :DDD Thanks @ariesdevil 

@sundy-li @dantengsky 
I hope the APIs(`db/table`) in this PR is still in line with the latest design.

dantengsky:
API looks nice to me. let's switch to meta_node first, and tweaks APIs later.

dantengsky:
@ariesdevil :+1: great job!

dantengsky:
@drmingdrmer @ariesdevil Any further concerns? 
It will be great if meta synchronization in PR #992 could use the new APIs in this PR.

fusestore/store/src/meta_service/state_machine.rs
ariesdevil: As I always use `db.tables.get(table_name)` to fetch table id, these two are used to avoid mutable(`db.tables.remove`) and immutable(`db.tables.get`) borrow at the same time :(

ariesdevil:
Just tweaked things that @drmingdrmer suggested. cc @dantengsky

fusestore/store/src/executor/action_handler_test.rs
ariesdevil: Got it

fusestore/store/src/meta_service/state_machine.rs
ariesdevil: Now I learned but still confuse why use `map[k]` syntax does not generate an immutable borrow❓

drmingdrmer: > Now I learned but still confuse why use `map[k]` syntax does not generate an immutable borrow❓

Not very sure about what exactly your question is. Do you have a minimal snippet to show it?

fusestore/store/src/meta_service/state_machine.rs
```diff
@@ -212,25 +220,85 @@ impl StateMachine {
...
+
+                if db.tables.contains_key(table_name) {
+                    let table_id = db.tables.get(table_name).unwrap();
+                    let prev = self.tables.get(table_id);
+                    if *if_not_exists {
+                        Ok((prev.cloned(), None).into())
+                    } else {
+                        Ok((None::<Table>, None::<Table>).into())
+                    }
+                } else {
```
drmingdrmer, line 274: The returned value should reflect the state change, i.e., includes the state before applying the cmd and the state after applying the cmd.

Thus if nothing is done, it should returns `(prev, prev)` to indicate that nothing changed.

`(prev, None)` could be a state change after applying a delete operation.

ariesdevil:
@drmingdrmer Move all exists checks logic from state_machine to meta_handler.

fusestore/store/src/meta_service/state_machine.rs
drmingdrmer: > The returned value should reflect the state change, i.e., includes the state before applying the cmd and the state after applying the cmd.
> 
> Thus if nothing is done, it should returns `(prev, prev)` to indicate that nothing changed.
> 
> `(prev, None)` could be a state change after applying a delete operation.

Sorry for the above statement. I just found some old codes did not follow this: `(prev, None)` was used to indicate a failed `add` operation. Thus the caller can not tell the difference between a failed `add` and a successful `delete`.

An solution would be add a `seq` to every value(for now `seq` is only used by `Cmd::UpsertKV`) and `apply()` returns the state before and after applying a `Cmd`.
Thus a failed `add` returns `((seq_1, value), (seq_1, value))` while a successful `delete` returns `((seq_1, value), None)`.

Any idea? 🤔 @ariesdevil

ariesdevil: Or we can use something like [this one](https://lib.rs/crates/enum_index) in return values to indicate the different commands.

Also, add a `seq` is fine for me.

fusestore/store/src/executor/meta_handlers.rs
```diff
@@ -23,40 +22,63 @@ use common_flights::meta_api_impl::GetDatabaseAction;
...
+                Ok(CreateDatabaseActionResult {
+                    database_id: db.database_id,
+                })
+            } else {
+                Err(ErrorCode::DatabaseAlreadyExists(format!(
+                    "{} database exists",
+                    db_name
+                )))
+            };
+        }
```
drmingdrmer, line 60: Reading from `meta_node` does not guarantee it sees the latest state, e.g., the database is already deleted(a user already received a positive ack of a delete operation and then at once sent a create-db operation, which should succeed.), but the change has not yet been replicated to this node.

Thus this checking should always be done at the apply phase in the state machine.

fusestore/store/src/executor/meta_handlers.rs
```diff
@@ -91,51 +110,101 @@ impl RequestHandler<DropDatabaseAction> for ActionHandler {
...
+        let if_not_exists = plan.if_not_exists;
 
-        info!("create table: {:}: {:?}", db_name, table_name);
+        info!("create table: {:}: {:?}", &db_name, &table_name);
 
-        let mut meta = self.meta.lock();
+        // if database not exists, return early.
+        let db = self.meta_node.get_database(db_name).await.ok_or_else(|| {
+            ErrorCode::UnknownDatabase(format!("create table: database not found {:}", db_name))
+        })?;
```
drmingdrmer, line 148: For the same reason, a just-created-db may not be seen on this node. Let the state machine do this check.

fusestore/store/src/executor/meta_handlers.rs
```diff
@@ -91,51 +110,101 @@ impl RequestHandler<DropDatabaseAction> for ActionHandler {
...
+                Ok(CreateTableActionResult {
+                    table_id: *table_id,
+                })
+            } else {
+                Err(ErrorCode::TableAlreadyExists(format!(
+                    "table exists: {}",
+                    table_name
+                )))
+            };
+        }
```
drmingdrmer, line 161: Same as above.

ariesdevil: Yes you're right, I'll revert this change and add a `seq` to eliminate confusing commands

fusestore/store/src/meta_service/state_machine.rs
ariesdevil: Do these changes in this area will affect changes to the API, should I leave a todo marker here and create a new issue for this? @drmingdrmer

fusestore/store/src/executor/meta_handlers.rs
drmingdrmer: > Yes you're right, I'll revert this change and add a `seq` to eliminate confusing commands

Maybe next PR? Except the inconsistent pre-check for `Add` command, most part of this PR is all right to me. Let's merge it ASAP. Leave other refinements later.

fusestore/store/src/executor/meta_handlers.rs
```diff
@@ -91,51 +116,88 @@ impl RequestHandler<DropDatabaseAction> for ActionHandler {
...
+        let table_name = &plan.table;
 
-        info!("create table: {:}: {:?}", db_name, table_name);
+        info!("create table: {:}: {:?}", &db_name, &table_name);
 
-        let mut meta = self.meta.lock();
+        // if database not exists, return early.
+        self.meta_node.get_database(db_name).await.ok_or_else(|| {
+            ErrorCode::UnknownDatabase(format!("create table: database not found {:}", db_name))
+        })?;
```
drmingdrmer, line 153: When creating a table, reading a database outside state machine may result in returning a false error.
A just-created-db may not be seen on this node.

fusestore/store/src/executor/meta_handlers.rs
```diff
@@ -145,37 +207,83 @@ impl RequestHandler<DropTableAction> for ActionHandler {
...
-        let mut meta = self.meta.lock();
-        let _ = meta.drop_table(&act.plan.db, &act.plan.table, act.plan.if_exists)?;
-        Ok(DropTableActionResult {})
+        let db_name = &act.plan.db;
+        let table_name = &act.plan.table;
+        let if_exists = act.plan.if_exists;
+
+        let db = self.meta_node.get_database(db_name).await.ok_or_else(|| {
+            ErrorCode::UnknownDatabase(format!("drop table: database not found {:}", db_name))
+        })?;
```
drmingdrmer, line 216: No need to get the db here either 🤔 
All the update must be done inside state machine.

fusestore/store/src/meta_service/state_machine.rs
drmingdrmer: > Do these changes in this area will affect changes to the API, should I leave a todo marker here and create a new issue for this? @drmingdrmer

Absolutely yes. Another issue seems a great idea!

drmingdrmer: > Or we can use something like [this one](https://lib.rs/crates/enum_index) in return values to indicate the different commands.
> 
> Also, add a `seq` is fine for me.

This crate is interesting and what I've been looking for!
Seems another issue is coming 🤔 ?

drmingdrmer (approved):
LGTM

PR #1031, opened 2021-07-11
Try to add function

wangzhen11aaa:

## Summary

This is a draft PR. I have some questions before coding.

I  found that all the function register into the map (FactoryFuncRef type), the second  function parameter has pub type FactoryFunc = fn(name: &str) -> Result<Box<dyn Function>>; 

A lot of toInt8/16/32/64 function,  I could create a file per function like  all the functions registered in map, in that way,  a world of duplicated codes would show, that's bad. 

Why did we design the function registered into the map  has a display_name?

## Changelog
add toint8/16/32/64 .. functions.

## Related Issues
https://github.com/datafuselabs/datafuse/issues/967
Fixes #issue
https://github.com/datafuselabs/datafuse/issues/967
## Test Plan
Local Test
Unit Tests

Stateless Tests


BohuTANG:
Hi @wangzhen11aaa 
display_name used to show the function name consistent with the user side, for example:
select SLeeP(2);
SLeeP is the display name here.

wangzhen11aaa:
> Hi @wangzhen11aaa
> display_name used to show the function name consistent with the user side, for example:
> select SLeeP(2);
> SLeeP is the display name here.

Thanks.

wangzhen11aaa:
In this pull request, added the toint8 function.
Next, I want to  add  the remaining functions  like the ComparisionFunction pattern and  put all the cast function into one file(or mod). 
And add some unit tests about these functions.

sundy-li:
Since we already have FunctionCast, why not just register other cast aliases? 
This is the simplest way to work.

```
toUInt8(col) ---> Cast(col as UInt8)
```

wangzhen11aaa:
> Since we already have FunctionCast, why not just register other cast aliases?
> This is the simplest way to work.
> 
> ```
> toUInt8(col) ---> Cast(col as UInt8)
> ```

First version. I will try that method later.

wangzhen11aaa:
> Since we already have FunctionCast, why not just register other cast aliases?
> This is the simplest way to work.
> 
> ```
> toUInt8(col) ---> Cast(col as UInt8)
> ```

I want to create a new mod named cast_function_template.rs.
codes like this.
"""
   pub struct CastFunctionTemplate;

impl CastFunctionTemplate{
    pub fn try_create_func(display_name: &str) -> Result<Box<dyn Function>> {
        match display_name{
            Some("int8") => CastFunction(DateType::Int8),

        }
    }
}

"""
The display_name will be a problem. I think.

sundy-li:
See :
```
HashesFunction::register
Expression  column_name
```

common/functions/src/scalars/expressions/cast_function_template.rs
```diff
@@ -0,0 +1,20 @@
+use common_datavalues::DataType;
+use common_exception::ErrorCode;
+use common_exception::Result;
+
+use crate::scalars::CastFunction;
+use crate::scalars::Function;
+
+pub struct CastFunctionTemplate;
+
+impl CastFunctionTemplate {
```
sundy-li, line 10: Why not put these into registers.

wangzhen11aaa: Don't want to change original codes that not familiar with. :)

common/functions/src/scalars/expressions/expression.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::scalars::FactoryFuncRef;
+
+#[derive(Clone)]
+pub struct ToCastFunction;
+
+impl ToCastFunction {
+    pub fn register(map: FactoryFuncRef) -> Result<()> {
+        let mut map = map.write();
+        map.insert("toint8", |display_name| {
+            CastFunction::create(display_name.to_string(), DataType::Int8)
```
sundy-li, line 17: All is Int8 ?

common/functions/src/scalars/expressions/cast_test.rs
```diff
@@ -38,7 +38,7 @@ fn test_cast_function() -> Result<()> {
             display: "CAST",
             nullable: false,
             columns: vec![Series::new(vec!["20210305", "20211024"]).into()],
-            func: CastFunction::create(DataType::Int32),
+            func: CastFunction::create("cast".to_string(), DataType::Int32),
             cast_type: DataType::Date32,
             expect: Series::new(vec![20210305i32, 20211024]),
```
sundy-li, line 43: expect type is i32, but cast type is Date32

common/functions/src/scalars/expressions/cast_test.rs
```diff
@@ -47,11 +47,10 @@ fn test_cast_function() -> Result<()> {
...
     for t in tests {
         let rows = t.columns[0].len();
-        let func = t.func;
+        let func = t.func.unwrap();
         if let Err(e) = func.eval(&t.columns, rows) {
             assert_eq!(t.error, e.to_string());
         }
-
         // Display check.
         let expect_display = t.display.to_string();
```
sundy-li, line 55: need to test the return_type to match the expected data_type

sundy-li: And stateless test is needed.

common/functions/src/scalars/expressions/cast_test.rs
wangzhen11aaa: Ok, The original code is  DateType::Int32..
I will change it into Date32.

wangzhen11aaa: > expect type is i32, but cast type is Date32

DateType is 2020-03-05 not 20200305. 
The original code is Int32.. And it passed.

common/functions/src/scalars/expressions/cast_test.rs
wangzhen11aaa: All stateless test is passed.

sundy-li (approved):
LGTM

BohuTANG:
Thank you @wangzhen11aaa

PR #1056, opened 2021-07-15
[store] feature: sled backed raft state

drmingdrmer:

##### [store] feature: sled backed raft state
- When starting a MetaNode(and a DFS that depends on a MetaNode),
  assign a on-disk dir for it to store persisted raft state, including:
  `node_id`, `current_term`, `voted_for`.

- Add: RaftState, which is a disk backed store of raft state.

- Change: MetaStore::new() now accepts a complete Config instead of a
  single listening addr.

- Test: add new_test_context() to create a temp env for testing MetaNode
  or DFS, which includes a temp meta store dir, unique port for meta
  service etc.

- Test: all tests that involves a MetaStore assign a temp dir for meta
  storage.

## Changelog

- New Feature

- Improvement

## Related Issues

#271


common/exception/src/exception.rs
```diff
@@ -181,6 +181,11 @@ build_exceptions! {
 
     InvalidConfig(2301),
 
+    // meta store errors
+
+    MetaStoreDamaged(2401),
+    MetaStoreAlreadyExists(2401),
```
ariesdevil, line 187: 2401 -> 2402

drmingdrmer: WT...thanks...

dantengsky (approved):
LGTM

PR #1076, opened 2021-07-17
[store] feature: ser/de with value order preserved

drmingdrmer:

##### [store] feature: ser/de with value order preserved
- Add: SledOrderedSerde to impl a ser/de that respect the original value order.

   E.g. serde_json does not preserve the order of u64:
   9 -> [57], 10 -> [49, 48]
   While BigEndian encoding preserve the order.

   A value that is used as sled db key must be serialized with order
   preserved.

   `SledSerde` impl the ser/de that does not require order preservation.

- Add: SledRangeSerde to serialize a `Range` to sled db range, in order
  to simplify range query or range modification.

- Change: ser/de NodeId with order preserved.

- Move NodeId into standalone file `node_id.rs`, impl order-preserved
  serialization for NodeId.

## Changelog

- New Feature

## Related Issues

#271

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1641,6 +1641,7 @@ dependencies = [
  "anyhow",
  "async-raft",
  "async-trait",
+ "byteorder",
  "common-arrow",
  "common-building",
  "common-datablocks",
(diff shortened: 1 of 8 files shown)


ariesdevil (approved):
LGTM

Issue #433, opened 2021-04-29
Optimizer: constant folding for comparison expression


BohuTANG:
**Summary**

Like arrow-datafusion does in [constant_folding.rs](https://github.com/apache/arrow-datafusion/blob/f43dc444eb510f0ff170adecb5a23afb39c489a8/datafusion/src/optimizer/constant_folding.rs):

`expr = true` and `expr != false` to `expr` when `expr` is of boolean type
`expr = false` and `expr != true` to `!expr` when `expr` is of boolean type
`true = true` and `false = false` to `true`
`false = true` and `true = false` to `false`
`!!expr` to `expr`
 `expr = null` and `expr != null` to `null`

jyizheng:
Could you assign this task to me? @BohuTANG

BohuTANG:
@jyizheng 

you can comments as:
/assignme

PR #1088, opened 2021-07-18
Constant folding

zhang2014:

## Summary

Folding constant expression in optimizer.

## Changelog

- Improvement

## Related Issues

related #433

### DIFF
--- a/common/functions/src/scalars/function.rs
+++ b/common/functions/src/scalars/function.rs
@@ -26,4 +26,8 @@ pub trait Function: fmt::Display + Sync + Send + DynClone {
     fn return_type(&self, args: &[DataType]) -> Result<DataType>;
     fn nullable(&self, _input_schema: &DataSchema) -> Result<bool>;
     fn eval(&self, columns: &[DataColumn], _input_rows: usize) -> Result<DataColumn>;
+
+    fn is_deterministic(&self) -> bool {
+        true
+    }
 }
--- a/common/functions/src/scalars/function.rs
+++ b/common/functions/src/scalars/function.rs
@@ -27,6 +27,7 @@ pub trait Function: fmt::Display + Sync + Send + DynClone {
     fn nullable(&self, _input_schema: &DataSchema) -> Result<bool>;
     fn eval(&self, columns: &[DataColumn], _input_rows: usize) -> Result<DataColumn>;
 
+    // If function returns the same result when same arguments, it is deterministic function.
     fn is_deterministic(&self) -> bool {
         true
     }
--- a/common/planners/src/plan_expression.rs
+++ b/common/planners/src/plan_expression.rs
@@ -48,6 +48,7 @@ pub enum Expression {
     /// Column name.
     Column(String),
     /// Constant value.
+    /// Note: When literal represents a column, its column_name will not be None
     Literal {
         value: DataValue,
         column_name: Option<String>,
--- a/common/planners/src/plan_expression.rs
+++ b/common/planners/src/plan_expression.rs
@@ -48,7 +48,10 @@ pub enum Expression {
     /// Column name.
     Column(String),
     /// Constant value.
-    Literal(DataValue),
+    Literal {
+        value: DataValue,
+        column_name: Option<String>,
+    },
     /// A unary expression such as "NOT foo"
     UnaryExpression { op: String, expr: Box<Expression> },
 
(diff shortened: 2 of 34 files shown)


fusequery/query/src/optimizers/optimizer_constant_folding_test.rs
```diff
@@ -4,64 +4,111 @@
...
+            query: &'static str,
+            expect: &'static str,
+        }
 
-        let plan = PlanNode::Projection(ProjectionPlan {
-            expr: vec![col("a")],
-            schema: DataSchemaRefExt::create(vec![DataField::new("a", DataType::Utf8, false)]),
-            input: Arc::from(filter_plan),
-        });
+        let tests: Vec<Test> = vec![
```
jyizheng, line 20: Should we add more tests for boolean expressions, as mentioned in datafusion?

https://github.com/apache/arrow-datafusion/blob/f43dc444eb510f0ff170adecb5a23afb39c489a8/datafusion/src/optimizer/constant_folding.rs#L33

common/functions/src/scalars/function.rs
```diff
@@ -26,4 +26,8 @@ pub trait Function: fmt::Display + Sync + Send + DynClone {
     fn return_type(&self, args: &[DataType]) -> Result<DataType>;
     fn nullable(&self, _input_schema: &DataSchema) -> Result<bool>;
     fn eval(&self, columns: &[DataColumn], _input_rows: usize) -> Result<DataColumn>;
+
+    fn is_deterministic(&self) -> bool {
```
jyizheng, line 30: What does "deterministic" mean?

common/planners/src/plan_expression.rs
```diff
@@ -48,7 +48,10 @@ pub enum Expression {
     /// Column name.
     Column(String),
     /// Constant value.
-    Literal(DataValue),
+    Literal {
+        value: DataValue,
+        column_name: Option<String>,
```
jyizheng, line 53: Any explanation for this change?

fusequery/query/src/optimizers/optimizer_constant_folding.rs
```diff
@@ -2,132 +2,191 @@
...
+        let expression_executor = Self::expr_executor(&input_schema, expression)?;
+        let dummy_columns = vec![DataColumn::Constant(DataValue::UInt8(Some(1)), 1)];
+        let data_block = DataBlock::create(input_schema, dummy_columns);
+        let executed_data_block = expression_executor.execute(&data_block)?;
+
+        ConstantFoldingImpl::convert_to_expression(origin_name, executed_data_block)
+    }
+
+    fn convert_to_expression(column_name: String, data_block: DataBlock) -> Result<Expression> {
+        assert_eq!(data_block.num_rows(), 1);
```
jyizheng, line 97: Is it possible to handle errors better?

fusequery/query/src/optimizers/optimizer_constant_folding_test.rs
zhang2014: This may be short circuit optimizer.

BTW: we can't optimize the short circuit in Kleene logic. because we can't determine whether null and bool. DataFuse uses Kleene logic.

fusequery/query/src/optimizers/optimizer_constant_folding.rs
zhang2014: It can only return one row result, I think we should be more strict

common/functions/src/scalars/function.rs
zhang2014: Comments added.

common/planners/src/plan_expression.rs
zhang2014: Comments added

fusequery/query/src/optimizers/optimizer_constant_folding.rs
```diff
@@ -2,132 +2,191 @@
...
+        assert_eq!(data_block.num_columns(), 1);
+
+        let column_name = Some(column_name);
+        let value = data_block.column(0).to_values()?.remove(0);
```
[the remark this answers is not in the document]
zhang2014: If function returns the same result when same arguments, it is deterministic function. when the arguments of the deterministic function is constant, we execute this function in optimizer to get a constant result.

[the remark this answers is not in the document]
zhang2014: We don't need to compute this function for the duration of the query

jyizheng (approved):

PR #1120, opened 2021-07-20
[datavalues] add basic test for AlignedVec and make it better

PsiACE:

## Summary

common/datavalues/src/vec.rs

## Changelog

- Improvement
- Build/Testing/CI

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/vec.rs
+++ b/common/datavalues/src/vec.rs
@@ -11,18 +11,19 @@ use common_arrow::arrow::alloc;
 use common_arrow::arrow::array::ArrayData;
 use common_arrow::arrow::array::PrimitiveArray;
 use common_arrow::arrow::buffer::Buffer;
+use common_arrow::arrow::buffer::MutableBuffer;
 use common_arrow::arrow::datatypes::*;
 
 /// A `Vec` wrapper with a memory alignment equal to Arrow's primitive arrays.
 /// Can be useful in creating a new DataArray or Arrow Primitive array without copying.
 #[derive(Debug)]
-pub struct AlignedVec<T> {
+pub struct AlignedVec<T: ArrowNativeType> {
     pub inner: Vec<T>,
     // if into_inner is called, this will be true and we can use the default Vec's destructor
     taken: bool,
 }
 
-impl<T> Drop for AlignedVec<T> {
+impl<T: ArrowNativeType> Drop for AlignedVec<T> {
     fn drop(&mut self) {
         if !self.taken {
             let inner = mem::take(&mut self.inner);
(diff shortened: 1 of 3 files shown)


common/datavalues/src/vec.rs
```diff
@@ -265,14 +277,49 @@ impl<T> AlignedVec<T> {
...
-impl<T> Default for AlignedVec<T> {
+impl<T: ArrowNativeType> Default for AlignedVec<T> {
     fn default() -> Self {
         // Be careful here. Don't initialize with a normal Vec as this will cause the wrong deallocator
         // to run and SIGSEGV
         Self::with_capacity_aligned(0)
     }
 }
+
+#[cfg(test)]
```
BohuTANG, line 295: It;s better to make a new file vec_test.rs here

PsiACE: Ok

common/datavalues/src/vec.rs
```diff
@@ -11,31 +11,32 @@ use common_arrow::arrow::alloc;
...
+impl<T: ArrowNativeType> Drop for AlignedVec<T> {
     fn drop(&mut self) {
         if !self.taken {
             let inner = mem::take(&mut self.inner);
             let mut me = mem::ManuallyDrop::new(inner);
             let ptr: *mut T = me.as_mut_ptr();
             let ptr = ptr as *mut u8;
             let ptr = std::ptr::NonNull::new(ptr).unwrap();
-            unsafe { alloc::free_aligned::<u8>(ptr, self.capacity()) }
+            unsafe { alloc::free_aligned::<u8>(ptr, me.capacity() * mem::size_of::<T>()) }
```
sundy-li, line 34: So the previous implementation has bug? Because `self.capacity()` is not the size.

common/datavalues/src/vec.rs
```diff
@@ -265,11 +277,14 @@ impl<T> AlignedVec<T> {
         self.inner.extend(iter);
         let added = self.len() - len_before;
 
-        assert_eq!(added, cap)
+        if added != cap {
+            eprintln!("size hint was incorrect, this is UB. aborting");
+            std::process::abort()
```
sundy-li, line 282: Is that good way to just abort?
Or we can use debug_assert! .

common/datavalues/src/vec.rs
BohuTANG: Is there any tool like github actions help us to check the  unsafe?

PsiACE: > So the previous implementation has bug? Because `self.capacity()` is not the size.

Known UB.

> Is there any tool like github actions help us to check the unsafe?

cargo miri, but some preparation is needed, we can open an issue tracking.

common/datavalues/src/vec.rs
PsiACE: > Is that good way to just abort?
> Or we can use debug_assert! .

Just follow the polars way. I will look at the debug_assert! .

sundy-li: ok

sundy-li (approved):
LGTM, thanks

PR #1138, opened 2021-07-21
remove jq and add download progress bar

ZhiHanZ:

## Summary
use grep and sed hack to remove jq dependencies, since most users not install jq by default, and add a progress bar during download
## Changelog
- Improvement

- Build/Testing/CI

## Related Issues

fixes https://github.com/datafuselabs/datafuse/issues/1137

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
It would better to move `installer` to `scripts` in this patch

docker/release/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM debian as installer
 COPY ./installer/install-datafuse.sh /install.sh
```
BohuTANG, line 2: ./scripts/installer/...

BohuTANG (approved):
LGTM

Issue #1080, opened 2021-07-17
[store] impl sled db backed raft-state-machine.


drmingdrmer:
- [x] #1311
- [x] #1312
- [x] #1313
- [ ] #1139 
- [x] #1385
- [x] #1386

Xuanwo:
@drmingdrmer Is this issue still valid?

drmingdrmer:
It's done.
Thanks for reminding:D

PR #1160, opened 2021-07-22
[store] feature: add SledTree<K, V> to provided typed access to sled tree.

drmingdrmer:

##### [store] feature: add SledTree<K, V> to provided typed access to sled tree.
- Add: SledTree: sled::Tree only support serialized key and value.
  Since we have several tree-like data: raft log, nodes in state machine
  or DFS keys. A generalized typed tree SledTree will simplify the impl
  of all these sled tree based data structures.

- Refactor: re-impl RaftLog with SledTree.

## Changelog

- New Feature

## Related Issues

#271

#1080


ariesdevil (approved):

ariesdevil:
Type safety is always a pleasure:)

PR #1169, opened 2021-07-23
db meta versioning & Store API for syncing db meta

### LINKED ISSUE #1168: Provides a global version number for database meta
**Summary**

We need a global version number for the database meta information, which will be increased for each DDL statement if executed successfully (or have externally visible effects).

dantengsky:

## Summary

- adds version info to database metadata, which will be increased by one for each successful modification of the database metadata.
- extends store_api::meta_api for synchronizing database metadata by version

## Changelog

- New Feature

## Related Issues

Part of #992 
Closes #1168

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/fusestore/store/src/api/rpc/flight_service_test.rs
+++ b/fusestore/store/src/api/rpc/flight_service_test.rs
@@ -11,8 +11,12 @@ use common_flights::StorageApi;
 use common_flights::StoreClient;
 use common_metatypes::MatchSeq;
 use common_planners::CreateDatabasePlan;
+use common_planners::CreateTablePlan;
 use common_planners::DatabaseEngineType;
+use common_planners::DropDatabasePlan;
+use common_planners::DropTablePlan;
 use common_planners::ScanPlan;
+use common_planners::TableEngineType;
 use common_runtime::tokio;
 use common_tracing::tracing;
 use pretty_assertions::assert_eq;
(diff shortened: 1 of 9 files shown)


fusestore/store/src/api/rpc/flight_service_test.rs
```diff
@@ -583,3 +587,153 @@ async fn test_flight_generic_kv() -> anyhow::Result<()> {
...
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_flight_get_database_meta_empty_db() -> anyhow::Result<()> {
+    common_tracing::init_default_tracing();
+    let (_tc, addr) = crate::tests::start_store_server().await?;
+    let mut client = StoreClient::try_create(addr.as_str(), "root", "xxx").await?;
+
+    // Empty Database
+    let res = client.get_database_meta(None).await?;
+    assert_eq!(None, res);
```
drmingdrmer, line 599: Without any databases, a result `(0, vec![])` may be more convenient to use, i.e., one less `match` on an `Option<_>`.

drmingdrmer (commented):
May be we could just use raft log index to indicate the version of the database collection.
Since raft log itself is a WAL and the index is actually the **version** number of the state machine.

dantengsky:
> May be we could just use raft log index to indicate the version of the database collection.
> Since raft log itself is a WAL and the index is actually the **version** number of the state machine.

Need some further information here,  will operations of sub application statemachines(like DFS, which insides the same raft group) will increase the index as well?

drmingdrmer:
> > May be we could just use raft log index to indicate the version of the database collection.
> > Since raft log itself is a WAL and the index is actually the **version** number of the state machine.
> 
> Need some further information here, will operations of sub application statemachines(like DFS, which insides the same raft group) will increase the index as well?

Yes, but, what I mean is to record the log index somewhere for your own use, not to read the `last_applied_log_index` from the latest state machine.

In other word, you still need another storage to save the `version`, but you do not need to maintain another seq-number.

drmingdrmer:
> > May be we could just use raft log index to indicate the version of the database collection.
> > Since raft log itself is a WAL and the index is actually the **version** number of the state machine.
> 
> Need some further information here, will operations of sub application statemachines(like DFS, which insides the same raft group) will increase the index as well?

But do not let this refactoring task block this PR yet. :DDD

fusestore/store/src/api/rpc/flight_service_test.rs
dantengsky: :D  A version of `get_database_meta` did return `(0, vec![])` when there is nothing to sync according to the params which client provides.  
But I changed it to the current signature (and return-type),  since:
-  `(0, res)` where `res` is not empty might be confusing to the caller site 
     The caller may need extra treatments, depends on the caller be defensive or else.
-   `None` is slightly efficient while being serialized

drmingdrmer: All right

drmingdrmer (approved):
Nice shot!

Issue #867, opened 2021-06-22
Catalog API


dantengsky:
**Summary**

Currently `DataSource` API provides us the essential functionalities of manipulating metadata,  but as we moving the v0.5,  some extra requirements should be fulfilled, including but not limited to:

  - <del>periodically sync metadata with(from) meta-store</del>
  - metadata of database, table should be versioned, as we are going to provide time travel features
  - exposes a session-level catalog API so that we can keep uncommitted metadata invisible
  - provide simple transaction API
    for per statement, implicit transaction. 
    Note: In v0.5 we are not going to support distributed/parallel insertion/deletion

drmingdrmer:
> * metadata of database, table should be versioned, as we are going to provide time travel features

Versions of table is a single listed list, or could it be a DAG?

```
v0 <- v1 <- v2 ...

or

 .--- v2 <----+- v3
v             |
v0 <- v1 <----'
```

BohuTANG:
I think it's a single chain like git

leiysky:
> * periodically sync metadata with(from) meta-store

How does DDL work?

drmingdrmer:
> I think it's a single chain like git

You're right. In every case it does not need a DAG history. I was overthinking.

dantengsky:
> > * metadata of database, table should be versioned, as we are going to provide time travel features
> 
> Versions of table is a single listed list, or could it be a DAG?
> 
> ```
> v0 <- v1 <- v2 ...
> 
> or
> 
>  .--- v2 <----+- v3
> |      |       |
> v      v       |
> v0 <- v1 <----'
> ```

"v0 <- v1 <- v2 " is enought for time travel.

>```
>  .--- v2 <----+- v3
> |      |       |
> v      v       |
> v0 <- v1 <----'
> ```
The above figure may describe:
  -  both `v2` and `v1` concurrently commit on the base of `v0`, and `v1` wins. `v2` successfully resolves conflicts and commits later.
  - both `v3` and `v2` concurrently commit on the base of `v1`, and `v2` wins. `v4` successfully resolves conflicts and commits later.
  
 Transaction operation log will contain an item/entry to record the `base version` (I might forget this in the last discussion though;-), but that's for post-audit/diagnostics, for time travel, linear history is enough I think.

dantengsky:
> > * periodically sync metadata with(from) meta-store
> 
> How does DDL work?

Roughly speaking, 
- DDL interpreters call catalog API 
[...]

PR #1176, opened 2021-07-23
Add meta_id and meta_version to table meta

### LINKED ISSUE #1175: Uniquely identifies Table by table_id and table_version
**Summary**

Since tables may be altered, modified(insert, update) concurrently, it is preferred that while a SQL statement is being processed (distributedly), all query nodes should work on consistent versions of tables; thus logically,  tables should be uniquely identifiable, adding  `table_id` and `table_version` to table's metadata might be a good start.

dantengsky:

## Summary

As stated in issue #1175, for a given statement we'd like to stick to a consistent view of a `Table` after the logical plan is generated. Thus we need to mark `Table`s returned from `Datasource` with table id/version.

This PR is mainly a refactoring of Datasource, no previous functionalities are changed.

- Introduce `TableMeta`, which wraps `Table` (and `TableFunction`) with  `id` and `version`; such that `plan`s, `transformer`s etc., may access table's id and version information and passes them around.
  - `TableMeta::datasource` gives us the good old `Table` trait back.
  - It may seem strange that we can not get the table's name, schema from `TableMeta` directly
    To keep this PR review-friendly, I'd like to do the relevant refactoring in another PR
- Renames `Datasource` to DatabaseCatalog (as CH does), and returns "TableMeta" instead of `Table` for `get_table` operation
  also, this is an initial effort of separating datasource from catalog. 
- Trait `Database` returns `TableMeta` for get_table operation, instead of trait `Table` 
- Table id and version are added to `ScanPlan` and `ReadSourcePlan` 
- Ids of System and Local tables are allocated from a preserved range, and those tables are not versioned.
  - `version` for those tables/views are always `None`
  -  Although it might be unnecessary to provide `id` for system tables, but to unify the TableMeta, we assign ids for system tables(as ti/crdb does)   
- Id and version of remote tables should be taken care of by meta-store

## Changelog

- New Feature

## Related Issues

Fixes #1175 
Part of #867 #992 

## Test Plan

Unit Tests

Stateless Tests


common/planners/src/plan_builder_scan.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020-2021 The Datafuse Authors.
```
BohuTANG, line 1: :+1:

BohuTANG (approved):
LGTM

PR #1181, opened 2021-07-25
[store] feature: add SledVarTypeTree to store more than one types of key-value in one sled::Tree.

drmingdrmer:

##### [store] feature: add SledVarTypeTree to store more than one types of key-value in one sled::Tree.
SledVarTypeTree is a candidate as raft state machine store.
Storing all state(such as last applied log index or internal seq number counter) in one tree makes transactional update or snapshot creation much easier.

## Changelog

- New Feature

## Related Issues

#271

#1080

### DIFF
--- a/fusestore/store/src/meta_service/raft_types.rs
+++ b/fusestore/store/src/meta_service/raft_types.rs
@@ -2,11 +2,15 @@
 //
 // SPDX-License-Identifier: Apache-2.0.
 
+use async_raft::LogId;
 pub use async_raft::NodeId;
 use byteorder::BigEndian;
 use byteorder::ByteOrder;
+use common_exception::ErrorCode;
+use sled::IVec;
 
 use crate::meta_service::sled_serde::SledOrderedSerde;
+use crate::meta_service::SledSerde;
 
 pub type LogIndex = u64;
 pub type Term = u64;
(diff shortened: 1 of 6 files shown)


fusestore/store/src/meta_service/raft_types.rs
```diff
@@ -21,3 +25,26 @@ impl SledOrderedSerde for u64 {
...
+        Ok(String::from_utf8(v.as_ref().to_vec())?)
+    }
+
+    fn order_preserved_serialize(&self, _buf: &mut [u8]) {
+        todo!()
+    }
+
+    fn order_preserved_deserialize(_buf: &[u8]) -> Self {
+        todo!()
+    }
```
dantengsky, line 46: - FYI: There is a so-called "Memory Comparable Encoding" in TiKV

https://pingcap.com/blog-cn/tikv-source-code-reading-13/ 

the relevant doc is a little bit oversimplified  (for me), taking a look at the  code may be helpful

https://github.com/tikv/tikv/blob/6708ce171792df02e1e90e1fe1e67e424d1586c8/components/codec/src/byte.rs#L64

- Is it the right time (or in the near future) to change the type of KEY (no longer stick to &str)?
  User-defined key types and order-preserving key encoding are easier to be implemented in binary format.

drmingdrmer: Sorry for the confusion. This is not really a `todo`. This function is not called. Please ignore these two todo.
Another refactoring PR is on the way!


> * Is it the right time (or in the near future) to change the type of KEY (no longer stick to &str)?
>   User-defined key types and order-preserving key encoding are easier to be implemented in binary format.

Have you ever run into a situation where you have to use bytes instead of str as the key?

dantengsky: got it

PR #1196, opened 2021-07-27
[common/tracing] fix: enable setting tracing level from env: RUST_LOG=debug

drmingdrmer:

##### [common/tracing] fix: enable setting tracing level from env: RUST_LOG=debug

Currently the log level is hard coded in common/tracing.

It need to be turned on by env variable `RUST_LOG`.

When output to stdout, use pretty format:
image

## Changelog

- Bug Fix


ariesdevil (approved):
LGTM

Issue #966, opened 2021-07-02
[functions] window funnel function 


BohuTANG:
**Summary**

ClickHouse windowfunnel:
https://clickhouse.tech/docs/en/sql-reference/aggregate-functions/parametric-functions/#windowfunnel

PR #1214, opened 2021-07-28
Generic aggregate function state

### LINKED ISSUE #1189: Reduce the memory size of AggregateFunction state
**Summary**

This test code proves that DataValue has memory size 64 bytes, because DataValue is an enum struct which uses the maximum memory for each enum.

```
 println!("memory size -> {:?}", std::mem::size_of::<DataValue>());   // 64 
 println!("memory size -> {:?}", std::mem::size_of::<Option<u64>>()); // 16
```

We should introduce a generic state implementation  to improve memory and efficiency.

### LINKED ISSUE #1215: common/io crate to have bianary serde
**Summary**

Currently, we serde the aggregate function state by *serde/json*, but it's not efficient as binary serde.

Eg:
avg_state
Current each state item has useless keys, and it's not compact format.
```
{"sum":xxx, "count":  xxx}
```

sundy-li:

## Summary

- Use Generic state representation instead of DataValue
- Merge min and max  functions into the same file.
- Introduce common/io crate, now we have BinarySer and BinaryDe for state serde.

## Changelog

- Improvement
- Performance Improvement

## Related Issues

Fixes  #1189
Fixes #1215

Prepare of #966 

## Test Plan

Unit Tests

Stateless Tests


PsiACE (commented):
Good enough for me.

jyizheng (commented):
It looks good to me!

common/io/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "common-io"
```
jyizheng, line 2: What are the users of this package?

sundy-li: see #1215

BohuTANG (approved):
LGTM
Thank you @sundy-li

PR #1220, opened 2021-07-28
bug fix: Kleene logic filter failure

### LINKED ISSUE #1112: Kleene logic filter failure
**Summary**

**How to reproduce**
```text
mysql> select * from numbers(10) where null and true;
ERROR 1105 (HY000): Code: 5, displayText = Unsupported value expression: NULL, type: Null.
```

InMySQL:
http://sqlfiddle.com/#!9/a6c585/209953

gauss1314:

## Summary

Fix kleene logic filter failure

## Changelog

- Bug Fix

## Related Issues

Fixes #1112 

## Test Plan

Unit Tests

Stateless Tests


gauss1314:
Hi, @sundy-li , I try to cast `DataType::Null` to `DataType::Boolean`  in other place like `data_df_type.rs` or  `sql_parser.rs`, but it does not make sense. I'm not sure for current commit.
And I have another 2 questions:
1.  When I add one test case in `plan_parser_test.rs`,expect text for sql `"select * from numbers(10) where null and true"` is 

`
"Projection: number:UInt64
             Filter: (NULL AND true)
                ReadDataSource: scan partitions: [8], scan schema: [number:UInt64], statistics: [read_rows: 10, read_bytes: 80]"`,
but when I execute sql `explain select * from numbers(10) where null and true;`, I got 

`
Projection: number:UInt64
  Filter: NULL
    ReadDataSource: scan partitions: [8], scan schema: [number:UInt64], statistics: [read_rows: 10, read_bytes: 80]
`.  "Filter" are different. Is there any other place should change?

2. When I execute sql `select number from numbers(10) where number is NULL or TRUE;` got `ERROR 1105 (HY000): Code: 5, displayText = Unsupported expression: number IS NULL, type: IsNull(Identifier(Ident { value: "number", quote_style: None })).` .

sundy-li:
@gauss1314  Thanks for the contribution.
> Unsupported expression: number IS NULL

This can implement in another pr. Let's just take into consideration of #1112 

For the query: 
`select number from numbers(10) where number   = null ;`

In the filter, `Number` is UInt64 column  and *null* is Constant Null Column (DataValue::Null).
It's not allowed to compare with `DataType:UInt64` and `DataType::Null`, so we can cast the Null to be DataType::UInt64 in the comparition ops, look into the codes in https://github.com/datafuselabs/datafuse/blob/master/common/datavalues/src/series/comparison.rs#L36-L55

For the query
`select number from numbers(10) where true  and null ;`

Similarly, we can cast the Null to be DataType::Bool before the comparition

fusequery/query/src/sql/plan_parser.rs
```diff
@@ -870,6 +870,9 @@ impl PlanParser {
                 sqlparser::ast::Value::Boolean(b) => {
                     Ok(Expression::create_literal(DataValue::Boolean(Some(*b))))
                 }
+                sqlparser::ast::Value::Null => {
+                    Ok(Expression::create_literal(DataValue::Boolean(None)))
```
sundy-li, line 874: Let's just take it as DataValue::Null.

zhang2014:
We also need to consider the following query:
```
SELECT * FROM `docs` WHERE NULL
```

gauss1314:
Hi @sundy-li , I try to convert `DataType::Null` to `DataType::Boolean` but get error: 
`Cast error: Casting from Null to Boolean not supported`
or 
`cannot unpack Series: "DataArray<Boolean>" of type Null into bool.`

common/datavalues/src/series/comparison.rs
```diff
@@ -40,6 +40,10 @@ fn coerce_cmp_lhs_rhs(lhs: &Series, rhs: &Series) -> Result<(Series, Series)> {
         return Ok((lhs.clone(), rhs.clone()));
     }
 
+    if lhs.data_type() == DataType::Null {
```
sundy-li, line 43: If it's null, we can use  *DataArray::<T>::full_null(lhs.len()) *  to have full null DataArray<T>

gauss1314: But the type of `lhs` is `&Series`, not `DataArray`. Please give more hints, thanks!

sundy-li: `DataArray<T>` can convert into `Series` by IntoSeries Trait

So the code may look like 

```
 if lhs.data_type() == DataType::Null { 
    let array = DataArray::<T>::full_null(size);
    let rhs = array.into_series();
    return coerce_cmp_lhs_rhs(lhs, rhs);
 }
````

sundy-li:
@gauss1314  I've add some codes to help this pr, please take a look at.

gauss1314:
> @gauss1314 I've added some codes to help this pr, please take a look at.

Thanks! 
Seems like `select * from numbers(10) where null and true;` is correct now. And I have another question, when execute sql `explain select * from numbers(10) where null;`, things are going well. But when execute sql `select * from numbers(10) where null;`, panic happened. I try to fix it, the result of `select * from numbers(10) where null;` is right, but get some error log in `nohup.out`:

thread 'thread 'thread 'thread 'thread 'thread 'tokio-runtime-workertokio-runtime-workertokio-runtime-workertokio-runtime-workertokio-runtime-workertokio-runtime-worker' panicked at '' panicked at '' panicked at '' panicked at '' panicked at '' panicked at 'assertion failed: `(left == right)`
  left: `3`,
 right: `7168`assertion failed: `(left == right)`
  left: `1`,
 right: `139755954284144`assertion failed: `(left == right)`
  left: `1`,
 right: `94005355719102`assertion failed: `(left == right)`
  left: `1`,
 right: `4173`assertion failed: `(left == right)`
  left: `1`,
 right: `139755956472384`assertion failed: `(left == right)`
  left: `1`,
 right: `139755892040464`', ', ', ', ', ', /home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs/home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs/home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs/home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs/home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs/home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs::::::383838383838::::::555555


thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `139756985496585`', /home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs:38:5
thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `139755215958256`', /home/gauss/.cargo/bin/git/checkouts/arrow2-6249446b5f8db6f7/8190d11/src/compute/filter.rs:38:5

sundy-li:
Great, but take a look at this case : 
https://www.db-fiddle.com/f/mGKShuVhq8pyrZEtwZSRDA/2

https://github.com/datafuselabs/datafuse/blob/394e86259f2fe0f0ff1da79ac65bf8c338edb874/fusequery/query/src/pipelines/transforms/transform_filter.rs#L92-L96

Better to use `cast_with_type` than `bool` ?

sundy-li:
Let's merge it first.

gauss1314:
Hi, @sundy-li . Should I raise 2 issues ?
One is about support syntax `select number from numbers(10) where number is NULL or TRUE;`
Another is pretty `nohup.out` log message when exec `select * from numbers(10) where null;`
Or I should take  high priority task?

sundy-li:
> Hi, @sundy-li . Should I raise 2 issues ?
> One is about support syntax `select number from numbers(10) where number is NULL or TRUE;`
> Another is pretty `nohup.out` log message when exec `select * from numbers(10) where null;`
> Or I should take high priority task?

Yes, that will be great

Issue #1139, opened 2021-07-21
[store] refactor: Distinguish potentially confusing operation results


ariesdevil:
**Summary**

There is a problem in the current implementation: we cannot distinguish between failed `add` and successful `delete` operations. 

By adding a `seq` field to results, we can avoid this problem like `UpsertKV`.

BTW: these changes will break the current API

cc @drmingdrmer @dantengsky @sundy-li

drmingdrmer:
> BTW: these changes will break the current API

It breaks the internal API: `MetaNode::write()` returns different values. The incompatibility should be dealt well by action handers. Components outside action handler won't be affacted.

@sundy-li will have nothing to worry about 🤔

dantengsky:
@ariesdevil 
pls break them if needed, proper and ergonomic APIs do worth the costs.

drmingdrmer:
@ariesdevil are you still working on this?

ariesdevil:
Yep, change the database and table to kv operations will cover this issue.

drmingdrmer:
> Yep, change the database and table to kv operations will cover this issue. It Will be done today.

If possible, smaller PR reduces latency and increases concurrency. (Just like the way we do with mutex. 🤔

ariesdevil:
Okay.

BohuTANG:
Ready to close?

PR #1222, opened 2021-07-28
Fix ISSUE-1102: Replace the relationship between database and table and data parts from tbl_parts to parts field in Table struct

### LINKED ISSUE #1102: [improvement] replace the relationship between database and table and data parts from tbl_parts to parts field in Table struct
**Summary**

Replace the relationship between database and table and data parts from tbl_parts to parts field in Table struct

ariesdevil:

## Summary

Replace the relationship between database and table and data parts from tbl_parts to parts field in Table struct

## Changelog

- Improvement

## Related Issues

Fixes #1102

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
@ariesdevil 

If the PR is in WIP, would better to change it to draft :)

ariesdevil:
To accomplish #1102 , I need to change the parts field of the Table structure, and I need to move Database and Table from metatypes to storage_api to circumvent the cyclic dependency. The situation gets more complicated when I finish #1139  based on this PR.

So I'm wondering if this is worth it, and so far I've come up with two options.

1. Leave the definition of the Table's parts field unchanged and still keep the `tbl_parts`, just change the `tbl_parts` in the `StateMachine` from a two-level structure to a one-level one.

2. Give up metatypes and move everything in metatypes to storage_api to avoid dependency issues.

@drmingdrmer @dantengsky I need some suggestions.

drmingdrmer:
> 2\. Give up `metatypes` crate and move everything in `metatypes` to `storage_api` crate to avoid dependency issues.
> 
> I also think that `DataPartInfo` should be part of the meta? If it is, then the current design is misleading because `DataPartInfo` is not handled in the meta_handler, and if it is not, then I should not change the `parts` field in the `Table` from `HashSet<String>` to `HashSet<DataPartInfo>`

`metatypes` is supposed to provide data types that are used by both query and store, AFAIK.
E.g. query depends on `metatypes`, store depends on `metatypes`.

`storage_api`, literally, is a collection of API for query to access store.

Thus to me, the dependency would be:

```
query -> storage_api -> store
  |          |            |
  |          |            |
  |          |            |
  `----------+------------+---> metatypes
```

drmingdrmer:
@ariesdevil 
BTW, later soon the state machine will be converted to a kv store thus all HashMap or BTreeMap will be converted to a individual key-values, e.g. the map for tables will become something like `SledVarTypeTree::as_type<Table>()`.

ariesdevil:
@drmingdrmer So the NO.1 solution is applicable? If so, I'll modify in this way.

ariesdevil:
Or leave these unchanged until we do "covert type to kv" refactoring

BohuTANG:
> Or leave these unchanged until we do "covert type to kv" refactoring

Ok, I've converted this PR status to draft

drmingdrmer:
> @drmingdrmer So the NO.1 solution is applicable? If so, I'll modify in this way.

IMHO this is preferred.

drmingdrmer:
> Or leave these unchanged until we do "covert type to kv" refactoring

Hm... actually I'd love to see small step patches.
Although it introduces more changes but the benefit is that it makes every refactoring step clear and much easier.

ariesdevil:
OK, Every step takes a thousand miles:)

drmingdrmer:
> @ariesdevil
> BTW, later soon the state machine will be converted to a kv store thus all HashMap or BTreeMap will be converted to a individual key-values, e.g. the map for tables will become something like `SledVarTypeTree::as_type<Table>()`.

My point is that a non-nested map reflects the eventual data structure more accurately...

PR #1231, opened 2021-07-29
[store] fix: do not fsync sled data in unittets. File::sync_all() takes 10 ~ 30 ms, at worst 500 ms on a Mac

drmingdrmer:

##### [store] fix: do not fsync sled data in unittets. File::sync_all() takes 10 ~ 30 ms, at worst 500 ms on a Mac
Sled tree wrappers provides an option to not to fsync after write
operation.
This should only be used in testing environment.

For more details about the slow fsync issue:
https://github.com/drmingdrmer/sledtest

## Changelog

- Bug Fix

## Related Issues

- [ ] #271

- [ ] #1080

### DIFF
--- a/fusestore/store/src/tests/service.rs
+++ b/fusestore/store/src/tests/service.rs
@@ -51,6 +51,9 @@ pub struct StoreTestContext {
 pub fn new_test_context() -> StoreTestContext {
     let mut config = configs::Config::empty();
 
+    // On mac File::sync_all() takes 10 ms ~ 30 ms, 500 ms at worst, which very likely to fail a test.
+    config.meta_no_sync = true;
+
     config.meta_api_port = next_port();
 
     let host = "127.0.0.1";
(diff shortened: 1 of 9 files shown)


fusestore/store/src/tests/service.rs
```diff
@@ -51,6 +51,9 @@ pub struct StoreTestContext {
 pub fn new_test_context() -> StoreTestContext {
     let mut config = configs::Config::empty();
 
+    // On mac File::sync_all() takes 10 ms ~ 30 ms, 500 ms at worst, which very likely to fail a test.
```
BohuTANG, line 54: Can we check the OS if it's MACOS and set it?

drmingdrmer: > Can we check the OS if it's MACOS and set it?

Good idea 🤔

BohuTANG (approved):
LGTM

fusestore/store/src/tests/service.rs
templexxx: All Non-Enterprise deice which has no Enhanced Power Loss Data Protection cannot sync enough fast. I think it's not just MacOS issue

drmingdrmer: @templexxx 
You are totally right, as always! :DDD
Any suggestions for building a stable durable storage system would be appreciated!

PR #1248, opened 2021-07-31
[store] feature: disk backed state machine

drmingdrmer:

##### [store] feature: disk backed state machine
This is more like a runnable framework of the disk backed meta-store than a
new-feature commit.
A lot things are left to be done.

- Add: a sled::Tree to store raft state machine.
  Only `nodes` are persisted so that restarting a store server would work
  as well.

  The applied key-value pairs storage and the state machine meta(such as
  `last_applied_log`) share a same sled::Tree.
  So that when creating snapshot a consistent view can be made easily
  with just one `iter()`.

- Add: `MetaNode` creates/installs a snapshot from/to sled::Tree

- Add: `SerializableSnapshot`, for converting sled::Tree iterator into a
  data chunk for transport.

- Change: define the key/value type of state-machine-meta with an `enum`.

- Change: rename `StateMachineBuilder` to `StateMachineInitializer`:
  It wont build a state machine any more, just initializes one.

- Test: add test of restarting a single node.
  Ensures data and state are restored.

TODO:

- [ ] Other data such as database, table are still pure memory.
- [ ] Test: restarting a cluster

## Changelog

- New Feature

## Related Issues

- [ ] #271
- [ ] #1080


BohuTANG (approved):
LGTM

PR #1252, opened 2021-07-31
[common/runtime] refactor: create handle outside runtime thread; remove select

drmingdrmer:

## Changelog

- Improvement


BohuTANG (approved):
Looks Great To Me :+1:

PR #1255, opened 2021-08-01
[common/stoppable] feature: borrow codes from fuse-query to make a shared graceful shutdown impl

drmingdrmer:

##### [common/stoppable] feature: borrow codes from fuse-query to make a shared graceful shutdown impl

FuseStore needs a graceful shutdown too. 

## Changelog

- New Feature


BohuTANG (approved):
LGTM

PR #1257, opened 2021-08-02
ISSUE 1254: fix race

### LINKED ISSUE #1254: Server ShutdownHandler is thread unsafe
**Summary**

We use AtomicBool(in shutdown method) to guarantee calls from wait_for_termination_request(when kill process) and `impl Drop for ShutdownHandle`(when panic). But we cannot guarantee that the code between store and load of AtomicBool is atomic(L53-L56)

https://github.com/datafuselabs/datafuse/blob/4bf2bed809c67d7ebabba7839519a3b293bac574/fusequery/query/src/servers/server.rs#L43-L97

SGZW:

## Summary

Fix #1254 

## Changelog

- Bug Fix

## Related Issues

Fixes #1254 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/fusequery/query/src/servers/server.rs
+++ b/fusequery/query/src/servers/server.rs
@@ -46,15 +46,11 @@ impl ShutdownHandle {
             shutdown_jobs.push(service.shutdown());
         }
 
-        let shutdown = self.shutdown.clone();
         let sessions = self.sessions.clone();
         let join_all = futures::future::join_all(shutdown_jobs);
         async move {
-            if !shutdown.load(Ordering::Relaxed) {
-                join_all.await;
-                sessions.shutdown(signal).await;
-                shutdown.store(true, Ordering::Relaxed);
-            }
+            join_all.await;
+            sessions.shutdown(signal).await;
         }
     }
 
@@ -65,8 +61,13 @@ impl ShutdownHandle {
 
             log::info!("Received termination signal.");
             log::info!("You can press Ctrl + C again to force shutdown.");
-            let shutdown_services = self.shutdown(Some(receiver));
-            shutdown_services.await;
+            if let Ok(false) =
+                self.shutdown
+                    .compare_exchange(false, true, Ordering::SeqCst, Ordering::Acquire)
+            {
+                let shutdown_services = self.shutdown(Some(receiver));
+                shutdown_services.await;
+            }
         }
     }
 
@@ -90,7 +91,10 @@ impl ShutdownHandle {
 
 impl Drop for ShutdownHandle {
     fn drop(&mut self) {
-        if !self.shutdown.load(Ordering::Relaxed) {
+        if let Ok(false) =
+            self.shutdown
+                .compare_exchange(false, true, Ordering::SeqCst, Ordering::Acquire)
+        {
             futures::executor::block_on(self.shutdown(None));
         }
     }


fusequery/query/src/servers/server.rs
```diff
@@ -50,10 +50,10 @@ impl ShutdownHandle {
         let sessions = self.sessions.clone();
         let join_all = futures::future::join_all(shutdown_jobs);
         async move {
-            if !shutdown.load(Ordering::Relaxed) {
+            if !shutdown.load(Ordering::Acquire) {
```
zhang2014, line 53: https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Acquire

Acquire only affects instruction reorder(in single thread). It seems that this is also thread unsafe?

SGZW: sorry about this,  change it to cas simplely

fusequery/query/src/servers/server.rs
```diff
@@ -50,10 +50,9 @@ impl ShutdownHandle {
         let sessions = self.sessions.clone();
         let join_all = futures::future::join_all(shutdown_jobs);
         async move {
-            if !shutdown.load(Ordering::Relaxed) {
+            if Ok(false) == shutdown.compare_exchange(false, true, Ordering::SeqCst, Ordering::Acquire) {
```
zhang2014, line 53: move this to `wait_for_termination_request` method and `impl Drop for ShutdownHandle`

by the way: maybe `if let Ok(false) = shutdown.compare_exchange(false, true, Ordering::SeqCst, Ordering::Acquire)` is better?

SGZW: yes, i will fix it, thanks

zhang2014 (approved):
LGTM, thanks for the contribution.

fusequery/query/src/servers/server.rs
```diff
@@ -65,8 +61,13 @@ impl ShutdownHandle {
...
             log::info!("You can press Ctrl + C again to force shutdown.");
-            let shutdown_services = self.shutdown(Some(receiver));
-            shutdown_services.await;
+            if let Ok(false) =
+                self.shutdown
+                    .compare_exchange(false, true, Ordering::SeqCst, Ordering::Acquire)
+            {
+                let shutdown_services = self.shutdown(Some(receiver));
+                shutdown_services.await;
+            }
```
drmingdrmer, line 70: @SGZW 
The condition check on `self.shutdown:Arc<AtomicBool>` should be the first step of `self.shutdown()`. 
Since `self` is a `&mut self`, no Arc or Atomic is required.

SGZW: atomic is to avoid race, but Arc is not required

drmingdrmer: Both usages of `self.shutdown` are in a scope that holds a `&mut self`. 
There can not be two threads both holding `&mut self` at any time.

The only way to let another thread have a `&mut self`(at a different time) is to send the `self` through some channel.

So, IMHO there should not be race condition. (hoping I did not miss anything 🤔 )

SGZW: in this case,  `&mut self ` move to closure,  is copy, in my sense

drmingdrmer: That's why I stated that to move the check on `self.shutdown` into `self.shutdown()`, which would simplify this issue...

Sorry I may not have expressed it very clear...

> The condition check on `self.shutdown:Arc<AtomicBool>` should be the first step of `self.shutdown()`. 
Since `self` is a `&mut self`, no Arc or Atomic is required.

SGZW: i think check outside can reduce overhead about moving variables into closure, and panic will call drop will happen in different thread?

drmingdrmer: Hm... what about this way: 
https://github.com/datafuselabs/datafuse/blob/dd03f129e77405c1540d2f231af71d1f9e409281/common/stoppable/src/stop_handle.rs#L34-L56

I didn't get it. It should drop-in-different-thread or not?

SGZW: i get this,  all stopping flag check is in one thread by "wait_to_terminate" function move

PR #1260, opened 2021-08-02
Remove anyhow for commons

zhang2014:

## Summary

Remove anyhow for commons

## Changelog


BohuTANG (approved):
LGTM

drmingdrmer:
@zhang2014 

`assert!(res.is_ok())` is almost the same as `res.unwrap()`. Both of them get rid of the detail of an error, while `res.unwrap()` is still easier to use. `res?` keeps the detail of an error and leaves the error to the testing framework to panic with error information display, which makes unit test easier to debug.

Using `ErrorCode` everywhere makes it another `anyhow::Error`.
Returning an error of a specific type(in our case, the `SeqError`) would make it easier for the caller to deal with the error:
It does not need to examine every possible error type(an `ErrorCode` could be IO error, network error, data consistency error etc) to make a decision what next to do, e.g, to retry or to shutdown.

zhang2014:
> assert!(res.is_ok()) is almost the same as res.unwrap(). Both of them get rid of the detail of an error, while res.unwrap() is still easier to use. res? keeps the detail of an error and leaves the error to the testing framework to panic with error information display, which makes unit test easier to debug.

OK, maybe we can implement a type conversion similar to "anyhow" for ErrorCode(only enabled testing).

> Using ErrorCode everywhere makes it another anyhow::Error.
Returning an error of a specific type(in our case, the SeqError) would make it easier for the caller to deal with the error:
It does not need to examine every possible error type(an ErrorCode could be IO error, network error, data consistency error etc) to make a decision what next to do, e.g, to retry or to shutdown.

👍 Yes, I just renamed it.

PR #1305, opened 2021-08-05
remove `conf` from RpcService by using session_mgr's conf 

### LINKED ISSUE #1304: maybe change conf to `sessions.get_conf()` is better?
maybe change conf to `sessions.get_conf()` is better?

_Originally posted by @zhang2014 in https://github.com/datafuselabs/datafuse/pull/1290#discussion_r683389255_

dantengsky:

## Summary

Summary about this PR

## Changelog

- Bug Fix

## Related Issues

Fixes #1304

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/fusequery/query/src/sessions/sessions.rs
+++ b/fusequery/query/src/sessions/sessions.rs
@@ -60,6 +60,10 @@ impl SessionManager {
         }))
     }
 
+    pub fn get_conf(&self) -> &Config {
+        &self.conf
+    }
+
     pub fn get_cluster(self: &Arc<Self>) -> ClusterRef {
         self.cluster.clone()
     }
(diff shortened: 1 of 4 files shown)


fusequery/query/src/sessions/sessions.rs
```diff
@@ -60,6 +60,10 @@ impl SessionManager {
         }))
     }
 
+    pub fn get_conf(&self) -> &Config {
```
zhang2014, line 63: If we want to return ref of Config, maybe get_conf -> `get_conf_ref` is better?

dantengsky: I am not sure... 

Since there are `DataSchemaRef`, `SessionManagerRef` ..., it seems that if `ref` is explicitly mentioned, it means `Arc<Something>`.

But I am Ok with it, if `get_conf_ref` is preferred, I'd like to rename it.

sundy-li: Bad name ideas from arrow, type ArrayRef = Arc<dyn Array>;

zhang2014 (approved):

PR #1325, opened 2021-08-06
[cli] fix: status test should use different temp dir or next time it reads the old data left by previous test

drmingdrmer:

## Changelog

- Bug Fix

### DIFF
--- a/fusecli/cli/src/cmds/status.rs
+++ b/fusecli/cli/src/cmds/status.rs
@@ -42,6 +42,7 @@ impl Status {
             .write(true)
             .truncate(true)
             .open(self.path.clone())?;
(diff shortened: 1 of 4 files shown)


fusecli/cli/src/cmds/status.rs
```diff
@@ -42,6 +42,7 @@ impl Status {
             .write(true)
             .truncate(true)
             .open(self.path.clone())?;
+        println!("{:?}", self);
```
BohuTANG, line 45: This is a temporary line? 🤔

drmingdrmer: Oh...sorry I pushed too quickly

Issue #1311, opened 2021-08-05
[store] feature: create a global sled db for all use in a process.


drmingdrmer:

SGZW:
what the mean of "all use"? state machine and raft log and anything else ?

drmingdrmer:
raft state, raft log, raft state machine.
Meanwhile these 3 components depends on sled::Db.

There is some testing ongoing to ensure whether this modification would fix the deadlock bug.

It seems using only one sled db still has chance of deadlock with concurrent calls to `sled::Tree::flush_async()` :(((

SGZW:
Thanks,  so the purpose of this issue is to solve the deadlock? Can you give more info about this dead lock or this problem is the internal of sled::Tree ?

drmingdrmer:
I did not yet dig deep enough yet. A deadlock occurs when using `sled::Tree::flush_async()` on multiple sled::Db in multiple `tokio::runtime`.
I have a snippet to reduce this problem:
https://github.com/drmingdrmer/sledtest/blob/main/src/main.rs

But in datafuse unittest, a single sled db has this problem too. (sled doc states that a process should open only one sled::Db which I thought is the cause of the deadlock.). The root may be in the concurrency control about `flush()` and `flush_async()`.
I've been trying to build another snippet to reproduce this.

When a tokio is dropped, it triggers a sled tree dropping.  Then another `flush()` is triggered when dropping the sled tree. 
At this point, a previous ongoing `flush_async` has chance to block forever.

PR #1326, opened 2021-08-06
[store] refactor: use master branch for sled, with the fix to deadlock within multiple tokio runtime

drmingdrmer:

##### [store] fix: open only 1 sled db in a process
To ensure the deadlock is fixed:
`cargo test -- --test-threads 10` which previously is very likely to deadlock should not hang anymore.

- Define a `lazy_static` to hold the only sled db a process opens.

- A store server opens the sled db before running.

- All unittest shares a same sled db but using different prefix to open
  a tree.

## Changelog

- Bug Fix

- Improvement

## Related Issues

- #271
- #1080


dantengsky (approved):
LGTM

drmingdrmer:
fix: #1311

PR #1331, opened 2021-08-07
[store] refactor: remove MetaStore::{open,new}, open_create() is enough

drmingdrmer:

## Changelog

- Improvement

## Related Issues

- #271

### DIFF
--- a/fusestore/store/src/meta_service/raftmeta.rs
+++ b/fusestore/store/src/meta_service/raftmeta.rs
@@ -241,7 +241,7 @@ impl RaftStorage<LogEntry, AppliedState> for MetaStore {
                     None => (0, 0).into(),
                 };
 
-                let sm_meta = sm.sm_tree.as_type::<StateMachineMeta>();
+                let sm_meta = sm.sm_tree.key_space::<StateMachineMeta>();
 
                 let last_applied_log = sm_meta
                     .get(&LastApplied)?
@@ -345,7 +345,7 @@ impl RaftStorage<LogEntry, AppliedState> for MetaStore {
             // Serialize the data of the state machine.
             let sm = self.state_machine.write().await;
 
-            let sm_meta = sm.sm_tree.as_type::<StateMachineMeta>();
+            let sm_meta = sm.sm_tree.key_space::<StateMachineMeta>();
 
             let last_applied = sm_meta
                 .get(&LastApplied)?
@@ -508,7 +508,7 @@ impl MetaStore {
             .await
             .expect("fail to get membership");
 
-        let sm_nodes = sm.sm_tree.as_type::<sledkv::Nodes>();
+        let sm_nodes = sm.sm_tree.key_space::<sledkv::Nodes>();
         let x = sm_nodes.range_keys(..).expect("fail to list nodes");
         for node_id in x {
             // it has been added into this cluster and is not a voter.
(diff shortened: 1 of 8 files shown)


fusestore/store/src/meta_service/sled_vartype_tree_test.rs
```diff
@@ -966,8 +966,8 @@ async fn test_as_multi_types() -> anyhow::Result<()> {
     let tc = new_sled_test_context();
     let db = &tc.db;
     let tree = SledVarTypeTree::open(db, tc.config.tree_name("foo"), true).await?;
-    let aslog = tree.as_type::<sledkv::Logs>();
-    let asmeta = tree.as_type::<StateMachineMeta>();
+    let log_tree = tree.key_space::<sledkv::Logs>();
+    let sm_meta = tree.key_space::<StateMachineMeta>();
```
ariesdevil, line 970: I prefer `sm_meta_tree` as of above is named `log_tree`.

fusestore/store/src/meta_service/state_machine.rs
```diff
@@ -214,7 +214,7 @@ impl StateMachine {
         };
 
         let inited = {
-            let sm_meta = sm.sm_tree.as_type::<StateMachineMeta>();
+            let sm_meta = sm.sm_tree.key_space::<StateMachineMeta>();
```
ariesdevil, line 217: Same as above.

fusestore/store/src/meta_service/raftmeta.rs
```diff
@@ -127,27 +127,13 @@ pub struct MetaStore {
...
-
     /// Open an existent `MetaStore` instance or create an new one:
     /// 1. If `open` is `Some`, try to open an existent one.
     /// 2. If `create` is `Some`, try to create one.
     /// Otherwise it panic
+    /// Returns MetaStore instance and a bool indicating if it is opened from a previous state.
+    ///
+    /// TODO(xp): make the is_open flag a field of MetaStore, maybe introduce a trait to define this behavior
     #[tracing::instrument(level = "info")]
     pub async fn open_create(
```
PsiACE, line 138: What do you think about renaming it to `open`

ariesdevil: `open_create` is fine since meta_node maybe not exists and needs to boot.

drmingdrmer: `open()` feels like just-loading-from-disk or something similar.
This function tries to open an existent one or to create a new one from nothing, depending on the two args `open` and `create`.

It's meant to remind readers that this function could be used to open, to create, or to create-if-can-not-open.
`Something::open()` being used to create something may confuse readers. 🤔

PsiACE: ok

fusestore/store/src/meta_service/sled_vartype_tree_test.rs
drmingdrmer: Right the names in unit tests aren't well thought.

I prefer `sm_meta` than `sm_meta_tree` since it is shorter...And the concept `tree` is not that important in this context.

I named it `log_tree` but not `logs` because there is a conflict...

What say you bro?

ariesdevil: All right, you convinced me.

ariesdevil (approved):

fusestore/store/src/meta_service/sled_vartype_tree_test.rs
drmingdrmer: > All right, you convinced me.

😭 Thanks for your understanding... It's not easy for me to find a perfect name...

PR #1332, opened 2021-08-07
[store] refactor: add a field `is_open` to MetaStore and RaftState to indicate if it is opened from a previously saved state

drmingdrmer:

## Changelog

- Improvement

## Related Issues

- #271
- #1080

### DIFF
--- a/fusestore/store/src/meta_service/raft_state.rs
+++ b/fusestore/store/src/meta_service/raft_state.rs
@@ -22,6 +22,9 @@ use crate::meta_service::SledSerde;
 pub struct RaftState {
     pub id: NodeId,
 
+    /// If the instance is opened(true) from an existent state(e.g. load from disk) or created(false).
+    is_open: bool,
+
     /// A unique prefix for opening multiple RaftState in a same sled::Db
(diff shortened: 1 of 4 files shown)


fusestore/store/src/meta_service/raft_state.rs
```diff
@@ -44,7 +53,7 @@ impl RaftState {
         config: &configs::Config,
         open: Option<()>,
         create: Option<()>,
```
ariesdevil, line 55: Why not just use boolean type for these two args?

drmingdrmer: In future there may be some create-specific args or open-specific args. 
E.g. 
```
open_create(
    open: Option<Path>, /* open from a path: */
    create: Option<(u64, u64)>, /* create with initial state: */ 
)
```

ariesdevil: Got it.

ariesdevil:
LGTM

drmingdrmer:
> LGTM

LGTM without an approval is not enough yet...

PR #1338, opened 2021-08-08
Improve filter performance :

ives9638:
inspect BooleanArray exit true .
un sql:
SELECT number FROM numbers_mt(100000000000) where number = 100 limit 1  ;
before:
1 rows in set. Elapsed: 66.156 sec. Processed 100.00 billion rows, 800.00 GB (1.51 billion rows/s., 12.09 GB/s.)
after:
1 rows in set. Elapsed: 37.384 sec. Processed 100.00 billion rows, 800.00 GB (2.67 billion rows/s., 21.40 GB/s.)

## Summary

Summary about this PR

## Changelog

- Performance Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


ives9638:
@BohuTANG @sundy-li  Reviewers

PsiACE:
`make lint` will make the cargo/lint ci happy

fusequery/query/src/pipelines/transforms/transform_filter.rs
```diff
@@ -93,6 +94,14 @@ impl Processor for FilterTransform {
             // Downcast to boolean array
             let filter_array = filter_array.bool()?.downcast_ref();
             // Convert to arrow record_batch
+
+            let mut  filter_exit_true =  filter_array.values().chunks::<u64>();
+
+            if  filter_exit_true.find( |p| p.count_ones() as usize > 0_usize ).is_none() &&
```
sundy-li, line 100: Good point, can use just use `!filter_exit_true.iter().any(|v| *v > 0) && ...`

ives9638: I've   pprof. It seems that chunks is better

PR #1358, opened 2021-08-09
[Build/Test/CI] Support to run tests on Macos catalina

ZhiHanZ:

## Summary
Support to run CI in catalina

## Changelog

- Build/Testing/CI

## Related Issues

fixes https://github.com/datafuselabs/datafuse/issues/1321

## Test Plan

Unit Tests

Stateless Tests


ZhiHanZ:
Mac OS latest runner in github did not install Mysql client by default, thus stateless tests cannot pass, may also need to reolve issue https://github.com/datafuselabs/datafuse/issues/1337

.github/workflows/unit-tests.yml
```diff
@@ -7,15 +7,23 @@ env:
...
+          - { os: macos-latest, toolchain: stable, target: x86_64-apple-darwin, cross: false }
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
 
       - name: Rust setup
         run: |
           bash ./scripts/setup/dev_setup.sh
-
+          bash ./scripts/setup/dev_setup.sh -o
```
drmingdrmer, line 26: Why do we need to install twice with and without `-o`?

ZhiHanZ: https://github.com/datafuselabs/datafuse/blob/master/scripts/setup/dev_setup.sh#L123, github macOS runner did not install mysql client by default(https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md), thus we need to install mysql client manually

drmingdrmer: Just a single `bash ./scripts/setup/dev_setup.sh -o` is not enough? I am supposing the `dev_setup.sh -o` won't turn off anything that are installed by `dev_setup.sh`. 🤔

ZhiHanZ: first setup would execute the following flow:
https://github.com/datafuselabs/datafuse/blob/master/scripts/setup/dev_setup.sh#L222-L228
enable operation install means that it would only install operator dependencies such as python, mysql client etc.
a better approach may be the following command
```bash
bash ./scripts/setup/dev_setup.sh -o -t
```

drmingdrmer (approved):
LGTM!

.github/workflows/unit-tests.yml
drmingdrmer: Hm... Thanks for your explanation! The control-flow in this script is not that intuitive right?

ZhiHanZ: help messages already shown in script usage pages, maybe we could add more documentation on setup part(an issue?)

drmingdrmer: 🤔 I've read this script several times before and I did notice there is a `usage` section. :(((

The help message may need to explain what it does when no options supplied?

ZhiHanZ: agreed, help message could provide some example use cases for the setup script

Issue #1360, opened 2021-08-09
[CI] Flaky store test on OSX


ZhiHanZ:
**Summary**
Unit Tests failed on OSX unit tests.
https://github.com/datafuselabs/datafuse/pull/1358/checks?check_run_id=3276265398

ZhiHanZ:
cc @drmingdrmer

drmingdrmer:
Thanks man!

PR #1364, opened 2021-08-09
[store] fix: read latest log index from active leader

drmingdrmer:

##### [store] fix: read latest log index from active leader
When testing raft related componennt, sometimes the leadership is taken
by other node, i.e., the original leader may not send a heartbeat
quickly enought.

Thus when reading the latest log index, first we have to reload the
leader id, then read log index from it.
Instead of reading log index directly from the default leader node-0.

## Changelog

- Bug Fix

## Related Issues

- #271

Fix: #1360

### DIFF
--- a/fusestore/store/src/meta_service/raftmeta_test.rs
+++ b/fusestore/store/src/meta_service/raftmeta_test.rs
@@ -284,6 +284,7 @@ async fn test_meta_node_set_file() -> anyhow::Result<()> {
         tracing::info!("start");
 
         let (mut _nlog, tcs) = setup_cluster(hashset![0, 1, 2], hashset![3, 4]).await?;
+
         let all = test_context_nodes(&tcs);
 
         // test writing on every node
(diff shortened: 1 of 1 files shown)


ZhiHanZ:
Could we rerun unit tests and stateless tests after macOS CI merged?

drmingdrmer:
> Could we rerun unit tests and stateless tests after macOS CI merged?

All right!

drmingdrmer:
> Could we rerun unit tests and stateless tests after macOS CI merged?

Surely we could but AFSIK this flakiness may not be relevant to mac os. Since I always run unit test on may laptop It is the first time I see this failure.

The root cause is that raft cluster run in an asynchronous and indeterministic  way. A message delay or loss wont affect correctness but affect the eventual state. 🤔

ZhiHanZ:
> > Could we rerun unit tests and stateless tests after macOS CI merged?
> 
> Surely we could but AFSIK this flakiness may not be relevant to mac os. Since I always run unit test on may laptop It is the first time I see this failure.
> 
> The root cause is that raft cluster run in an asynchronous and indeterministic way. A message delay or loss wont affect correctness but affect the eventual state.

Thanks for explainations! this PR LGTM

ZhiHanZ (approved):

PR #1369, opened 2021-08-09
[store] change: MetaNode::upsertKV return two equal value if no modification is applied

drmingdrmer:

##### [store] change: MetaNode::upsertKV return two equal value if no modification is applied
Previously, if an `Add` action failed, due to mismatching `seq`, it
returns `(previous value, None)`.
From now on it returns `(previous value, previous value)` to indicate
nothing change.

**Why**:

The previous solution can not distinguish the result from a successful
`delete` and the result from a failed `add`.
Thus the caller has to examine the context to determine if the return
value is a success. E.g., by considering the action(`add` or `delete`).

With the new solution, the caller knows if it succeeded by only a simple comparison on
`res.prev` and `res.result`.

## Related Issues

- #271
- #1080 
- #1139

### DIFF
--- a/fusestore/store/src/api/rpc/flight_service_test.rs
+++ b/fusestore/store/src/api/rpc/flight_service_test.rs
@@ -615,88 +615,50 @@ async fn test_scan_partition() -> anyhow::Result<()> {
 }
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-async fn test_flight_generic_kv() -> anyhow::Result<()> {
+async fn test_flight_generic_kv_mget() -> anyhow::Result<()> {
     init_store_unittest();
-
     {
-        let span = tracing::span!(tracing::Level::INFO, "test_flight_generic_kv");
+        let span = tracing::span!(tracing::Level::INFO, "test_flight_generic_kv_list");
         let _ent = span.enter();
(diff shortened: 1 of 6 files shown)


ariesdevil:
I already do this change at https://github.com/datafuselabs/datafuse/pull/1336/files#diff-5ad456229eabd43366fde76f05cf895d991637c8805ea33946ffe85db90e1b4eR526 :(

drmingdrmer:
> I already do this change at https://github.com/datafuselabs/datafuse/pull/1336/files#diff-5ad456229eabd43366fde76f05cf895d991637c8805ea33946ffe85db90e1b4eR526 :(

I need to apply this change first before I can continue on the finish the works on snapshot.

drmingdrmer:
> I already do this change at https://github.com/datafuselabs/datafuse/pull/1336/files#diff-5ad456229eabd43366fde76f05cf895d991637c8805ea33946ffe85db90e1b4eR526 :(

There has been a lot work that needs to be done in this month.

ariesdevil:
OK

ariesdevil (approved):

drmingdrmer:
> OK, I'll rebase it later.

Thanks!

PR #1376, opened 2021-08-09
Add session type and session extra info for RPCSession

zhang2014:

## Summary

Add session type and session extra info for RPCSession

## Changelog

- Improvement


sundy-li:
Need Lint the codes, btw is there a test for the codes?

zhang2014:
> Need Lint the codes, btw is there a test for the codes?

session id is random. stateless tests may not be written

sundy-li (approved):

PR #1377, opened 2021-08-09
Add sum modulo perf case

sundy-li:

## Summary

Add sum modulo perf case

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


BohuTANG (approved):
LGTM

Issue #1402, opened 2021-08-10
[FR] Support expire action in key/value API


zhang2014:
**Summary**

- Support ExpireKVAction(key, new_timeout)
- Add the timeout field in the UpsertKVAction

CC: @drmingdrmer 

---
Attached issues from your dear assignee:DDD

- [x] #1422
- [x] #1426

PR #1425, opened 2021-08-11
[store] feature: generic kv support auto expiring an entry.

drmingdrmer:

##### [store] feature: generic kv support auto expiring an entry.
Add: UpsertKV adds another `meta` field to specify value meta.  One of
the meta is `expire_at`, which is a timestamp since 1970.  Reading on an
expired entry always returns None, e.g. when reading directly from
StateMachine or when reanding it for an update(apply a raft log).

## Changelog

- New Feature

## Related Issues

- #271
- #1402 
- fix: #1422


ariesdevil (approved):
LGTM

PR #1437, opened 2021-08-11
IDX-1428: common indexing

### LINKED ISSUE #1428: [common] common/indexing mod
**Summary**

1. extract indexing algo
2. search indexing algo
3. write/read api with MetaStore or DFS

BohuTANG:

## Summary
This patch only impl the min_max and sparse index create and make a spec for pruning with `apply_index`(always return true now).
We will impl the pruning logic in another issue https://github.com/datafuselabs/datafuse/issues/1552.

Tasks:
- [x] extract indexing from the blocks which sorted for one parquet.file
     - [x] min/max indexing
     - [x] sparse indexing

## Changelog

- New Feature

## Related Issues

Fixes #1428

## Test Plan

Unit Tests

Stateless Tests


common/datablocks/src/data_block.rs
```diff
@@ -132,6 +117,20 @@ impl DataBlock {
         }
     }
 
+    /// Take the first data value of the column.
+    pub fn first(&self, col: &str) -> Result<DataValue> {
+        let series = self.try_array_by_name(col)?;
+        let values = series.to_values()?;
```
sundy-li, line 123: `to_values()` is too heavy,  try  `try_get(idx)`

common/datablocks/src/data_block.rs
```diff
@@ -132,6 +117,20 @@ impl DataBlock {
...
+        Ok(values[0].clone())
+    }
+
+    /// Take the last data value of the column.
+    pub fn last(&self, col: &str) -> Result<DataValue> {
+        let series = self.try_array_by_name(col)?;
+        let values = series.to_values()?;
+        Ok(values[values.len() - 1].clone())
```
sundy-li, line 131: .

common/datablocks/src/data_block.rs
sundy-li: There is already a method in `DataColumn` named try_get

zhang2014 (approved):
LGTM

common/datablocks/src/data_block.rs
BohuTANG: Ok, I will update it in another patch.

PR #1442, opened 2021-08-11
Object Store Data Access Layer

### LINKED ISSUE #1407: [common] add common/dal mod
**Summary**

common/datasources mod may including:
```
- dal: interface for the shared storage such as S3
```

dantengsky:

## Summary

- An Interface for accessing Object Store (supports AsyncRead/Seek)
- Implementation for S3
  Simple unit test for AsynRead/Seek passed in my local dev (using an AWS trial account)
  Unit tests are **ignored** during CI, which will be enabled till we can inject, say `AWS_SECRET_ACCESS_KEY`, `AWS_ACCESS_KEY_ID` in CI env.  
- Implementation for Local FileSystem

NOTE: impls are not well-tuned, huge room for performance improvement

Some thoughts:

- About the Object size(length of the object being fetch/seek)
  Although we can provide AsyncSeek without knowing the size of the object, that might be inefficient: since 
  - In our case (the upcoming new table format), we know the size of target Objects (in Table meta, or index data)
  - Without knowing the size, an extra http HEAD request is needed to infer the size. (As SeekFrom::End(0), is used to detect the size, if we allow blind seeking, things are going to be wrong)
  
- We might need a rust s3 sdk, which supports chunked upload
   tried [rust_s3](https://github.com/durch/rust-s3), [rusoto_s3](https://github.com/rusoto/rusoto), [aws_sdk_s3 alpha](https://github.com/awslabs/aws-sdk-rust), they all seem to be lack of chunked transform-encoding supporting while upload data to S3.. Hope I am not missing something.
    But, again,  if we know the size before uploading, this is NOT a problem (multipart uploading may be more useful)

## Changelog

- Improvement

## Related Issues

Partially fixes #1407 

## Test Plan

Unit Tests

Stateless Tests


BohuTANG (approved):
LGTM

PR #1443, opened 2021-08-12
Introducing a custom HashMap to try to improve group by performance

zhang2014:

## Summary

Experimental Pull Request.

## Changelog

- Improvement


BohuTANG:
Cool.
This mod should in common/hashtables not query/common? :thinking:

zhang2014:
> This mod should in common/hashtables not query/common? 🤔

HashTable will only be used in query. Maybe it should be in query? 
By the way, this is just an attempt to verify

BohuTANG:
> > This mod should in common/hashtables not query/common? thinking
> 
> HashTable will only be used in query. Maybe it should be in query?

Ok, make sense.

> By the way, this is just an attempt to verify

drmingdrmer:
Cool~

drmingdrmer:
Out of curiosity, what is the advantage in this implementation that makes it faster than the std hash?

jyizheng:
What is the collision resolution policy for this hashtable implementation?

zhang2014:
> Out of curiosity, what is the advantage in this implementation that makes it faster than the std hash?

Current test, it is just a little better than std. I'm not sure whether we should switch to custom HashMap for this.

Optimize(from ClickHouse HashTable):
- Before 12582912 (2 ^ 23), the capacity is expanded by 4 times each time, otherwise it is expanded by 2 times
- Merge, get and insert operations (completed in one time, like std entry)
- Use custom hash algorithm(from murmurhash)
- Used std::alloc::realloc to resize

PsiACE:
> Out of curiosity, what is the advantage in this implementation that makes it faster than the std hash?

std hash (siphash) is a not-so-fast hashing algorithm, the most common fast hash algorithm currently used by Rust is ahash for the general case. If the object is small, then fxhash and zwohash will be faster than ahash.

zhang2014:
> What is the collision resolution policy for this hashtable implementation?

It's a open addressing hash table

PsiACE:
>     * Before 12582912 (2 ^ 23), the capacity is expanded by 4 times each time, otherwise it is expanded by 2 times

It looks like this could alleviate the tail delay. But I think the impact is still visible.

BohuTANG:
> > Out of curiosity, what is the advantage in this implementation that makes it faster than the std hash?
> 
> Current test, it is just a little better than std. I'm not sure whether we should switch to custom HashMap for this.

:+1: , let's do it later.

> Optimize(from ClickHouse HashTable):
> 
> * Before 12582912 (2 ^ 23), the capacity is expanded by 4 times each time, otherwise it is expanded by 2 times
> * Merge, get and insert operations (completed in one time, like std entry)
> * Use custom hash algorithm(from murmurhash)
> * Used std::alloc::realloc to resize

zhang2014:
> 👍 , let's do it later.

Ok. Let's temporarily close this pr

zhang2014:
Reopen, performance test results from @sundy-li (without native compile flag):
```
datafuse : ) 60 rows in set. Elapsed: 1.583 sec. Processed 1.00 billion rows, 8.00 GB (631.70 million rows/s., 5.05 GB/s.)
clickhouse : ) 60 rows in set. Elapsed: 1.776 sec. Processed 1.00 billion rows, 8.00 GB (563.11 million rows/s., 4.50 GB/s.)
```

common/infallible/src/rwlock.rs
```diff
@@ -11,6 +11,8 @@ use parking_lot::RwLockWriteGuard;
 #[derive(Debug, Default)]
 pub struct RwLock<T>(ParkingRwLock<T>);
 
+unsafe impl<T> Send for RwLock<T> {}
```
sundy-li, line 14: Should add `where T:Send` ?

query/src/common/hash_table_iter.rs
```diff
@@ -0,0 +1,51 @@
...
+                    return self.zero_entity;
+                }
+            }
+
+            self.idx += 1;
+            while self.idx < self.capacity && self.entities.offset(self.idx).is_zero() {
+                self.idx += 1;
+            }
+
+            match self.idx == self.capacity {
```
leiysky, line 45: What about use `if-else` here?

leiysky:
Seems it's a linear probing hash scheme, which can utilize with vectorization easily(by supporting `batch_get` and `batch_insert`). 

I've just found a [paper](https://www.researchgate.net/publication/326669722_SIMD_Vectorized_Hashing_for_Grouped_Aggregation_22nd_European_Conference_ADBIS_2018_Budapest_Hungary_September_2-5_2018_Proceedings) about this, FYI.

query/src/common/hash_table.rs
```diff
@@ -0,0 +1,216 @@
+// Copyright 2020-2021 The Datafuse Authors.
```
BohuTANG, line 1: Aha, we have the new licenese header now :)

zhang2014:
> Seems it's a linear probing hash scheme, which can utilize with vectorization easily(by supporting batch_get and batch_insert).
> 
> I've just found a paper about this, FYI.

This is really a great idea. Let's create a issue to record it.
By the way, in actual tests, using rust auto vectorization may be better than write SIMD instructions(I guess this may be because it can always improve the utilization of registers well enough)

sundy-li (approved):
LGTM

PR #1444, opened 2021-08-12
Only store the first address of state in group by transform

### LINKED ISSUE #1438: Store one address of state in group by transform
**Summary**

If there are many aggregate functions in query, like `select sum(a), max(b) ...`.

Datafuse used `Map<Hash, vec<usize>>` to store the state memory address to each key.
It can be improved by making it as `Map<Hash, usize>`, because there are fixed offset between each address.

https://github.com/fitzgen/bumpalo/pull/117

sundy-li:

## Summary

Only store the first address of state in group by transform

## Changelog

- Improvement

## Related Issues

Fixes #1438 
## Test Plan

Unit Tests

Stateless Tests


sundy-li:
This SQL case outperforms ClickHouse in my pc even without #1443 .

```

SELECT  sum(number)  FROM numbers_mt(1000000000)  group by number % 3, number % 4, number % 5;

datafuse:

60 rows in set. Elapsed: 1.584 sec. Processed 1.00 billion rows, 8.00 GB (631.31 million rows/s., 5.05 GB/s.)

clickhouse:
60 rows in set. Elapsed: 1.733 sec. Processed 1.00 billion rows, 8.00 GB (577.08 million rows/s., 4.62 GB/s.)

```

zhang2014 (approved):
LGTM

PR #1455, opened 2021-08-12
[CI] more iterations

ZhiHanZ:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

fixes https://github.com/datafuselabs/datafuse/issues/1453

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/.github/workflows/datafuse-performance-dispatch.yaml
+++ b/.github/workflows/datafuse-performance-dispatch.yaml
@@ -16,7 +16,7 @@ jobs:
       PR_NUMBER: ${{ github.event.client_payload.PR_NUMBER }}
       LAST_COMMIT_SHA: ${{ github.event.client_payload.LAST_COMMIT_SHA }}
       UUID: ${{ github.event.client_payload.UUID }}
-      ITERATION: 10
+      ITERATION: 100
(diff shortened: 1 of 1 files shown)


.github/workflows/datafuse-performance-dispatch.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
       PR_NUMBER: ${{ github.event.client_payload.PR_NUMBER }}
       LAST_COMMIT_SHA: ${{ github.event.client_payload.LAST_COMMIT_SHA }}
       UUID: ${{ github.event.client_payload.UUID }}
-      ITERATION: 10
+      ITERATION: 100
```
PsiACE, line 19: LGTM, just wondering if we have a way to specify how many times it runs.

ZhiHanZ: During the deisgn of test infra, I thought it could be add to /run-perf as a variable payload, but that maybe too complex for developers to use it

PsiACE: Ok, keep it easy

PR #1461, opened 2021-08-13
[store]: feature: RaftState: stores the last snapshot meta. To avoid potential repliation bug.

drmingdrmer:

##### [store]: feature: RaftState: stores the last snapshot meta. To avoid potential repliation bug.
Because membership log is not applied to state machine,
Before logs are removed, the membership need to be saved.

The async-raft::MemStore saves it in a removed log slot,
which may mess up the replication:
Replication has chance sending this pointer log that was an client data.

Although, the presence of snapshot meta does not mean there is a valid snapshot data.
Since snapshot data is not saved on disk.

A snapshot meta only guarantees no smaller snapshot will be installed
when installing snapshot, and no membership config is lost when removing
logs that are included in snapshot.

## Changelog

- New Feature

## Related Issues

- #271
- #1080

### DIFF
--- a/store/Cargo.toml
+++ b/store/Cargo.toml
@@ -45,6 +45,7 @@ futures = "0.3"
 indexmap = "1.7.0"
 lazy_static = "1.4.0"
 log = "0.4"
+maplit = "1.0.2"
 metrics = "0.17.0"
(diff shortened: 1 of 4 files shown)


ariesdevil (approved):
Look Greeeat To Me!

PR #1468, opened 2021-08-15
Add test for array builder

junli1026:

## Summary

Add test for data array and builder 

## Changelog

- Testing

### DIFF
--- a/common/datavalues/src/arrays/builders/boolean_test.rs
+++ b/common/datavalues/src/arrays/builders/boolean_test.rs
@@ -1,0 +1,68 @@
+// Copyright 2020-2021 The Datafuse Authors.
+//
+// SPDX-License-Identifier: Apache-2.0.
+
+use crate::prelude::*;
+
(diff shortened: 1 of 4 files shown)


common/datavalues/src/arrays/builders/boolean_test.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright 2020-2021 The Datafuse Authors.
```
BohuTANG, line 1: We have the new license :)

junli1026: will fix.

BohuTANG (approved):
LGTM
Thank you @junli1026

BohuTANG:
CI failed, let's 'make lint'

junli1026:
> CI failed, let's 'make lint'

Thanks for reminding. I updated the PR with formatted code. Should be fine now.

PR #1473, opened 2021-08-15
issue 1349

### LINKED ISSUE #1349: executes `TruncateTableAction` by using raft consensus protocol
**Summary**

If I get it right, `TruncateTableAction` is being executed by modifying the state of `MetaStore` directly, which is correct in single-node store deployment. But for multi-nodes deployment of `Store`,  `TruncateTableAction` should be performed by using the raft consensus protocol, like "CreateTable/Databse" does

SGZW:

## Summary

solve #1349 

## Changelog

- Improvement

## Related Issues

Fixes #1349 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/store/src/executor/action_handler_test.rs
+++ b/store/src/executor/action_handler_test.rs
@@ -32,6 +32,9 @@ use common_flights::meta_api_impl::GetDatabaseAction;
 use common_flights::meta_api_impl::GetDatabaseActionResult;
 use common_flights::meta_api_impl::GetTableAction;
 use common_flights::meta_api_impl::GetTableActionResult;
+use common_flights::storage_api_impl::AppendResult;
+use common_flights::storage_api_impl::TruncateTableAction;
+use common_flights::storage_api_impl::TruncateTableResult;
 use common_planners::CreateDatabasePlan;
 use common_planners::CreateTablePlan;
 use common_planners::DatabaseEngineType;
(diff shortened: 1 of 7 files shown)


store/src/meta_service/state_machine.rs
```diff
@@ -550,6 +550,22 @@ impl StateMachine {
...
+                let tbl_id = db.tables.get(table_name);
+                if let Some(tbl_id) = tbl_id {
+                    let tbl_id = tbl_id.to_owned();
+                    self.remove_table_data_parts(db_name, table_name);
+                    let prev = self.tables.get(&tbl_id);
+                    Ok((prev.cloned(), prev.cloned()).into())
+                } else {
+                    Ok((None::<Table>, None::<Table>).into())
+                }
+            }
```
drmingdrmer, line 568: The two fields `prev`, `result` in  `AppliedState` implies the state before and after applying an `Cmd`.
For `truncate table`, no matter whether it succeeded, the state of the table never changes.
What changed is actually the table parts.

So, here the `AppliedState` should reflect the table parts changes instead of the table changes.

Since the `AppliedState` is only used internally for now, and it is inappropriate to put the entire list of parts into `AppliedState`, my suggestion is to introduce a struct to describe the table parts change. 

BTW. there should be a corresponding `seq` for a `table`, as it is for `UpsertKV`.
Then we could use the `seq` number to identify a table change.
Maybe another issue 🤔

store/src/executor/storage_handlers.rs
```diff
@@ -54,10 +57,32 @@ impl RequestHandler<TruncateTableAction> for ActionHandler {
...
+                    Ok(TruncateTableResult {})
+                } else {
+                    Err(ErrorCode::UnknownTable(format!(
+                        "table not found: {:}",
+                        tbl_name
+                    )))
+                }
+            }
+            _ => Err(ErrorCode::MetaNodeInternalError("not a Table result")),
+        }
```
drmingdrmer, line 86: For a succeeded apply, the `prev` and the `result` should be different to indicate a change.

store/src/executor/action_handler_test.rs
```diff
@@ -644,6 +646,104 @@ async fn test_action_handler_drop_table() -> anyhow::Result<()> {
...
+                }
+                Err(ref err) => {
+                    let got = rst.unwrap_err();
+                    let got: ErrorCode = got.into();
+                    assert_eq!(err.code(), got.code(), "{}", mes);
+                    assert_eq!(err.message(), got.message(), "{}", mes);
+                }
+            }
+        }
+    }
```
drmingdrmer, line 742: A test also need to refetch the table to ensure the parts are removed.

drmingdrmer (commented):
Thank you man!

store/src/meta_service/state_machine.rs
SGZW: thanks, a new struct for `AppliedState` is great idea, and i will add seq for table to identify a table change in this pr, but this seq only for debug?

store/src/executor/storage_handlers.rs
SGZW: ok

store/src/executor/action_handler_test.rs
SGZW: i will add this test

store/src/meta_service/state_machine.rs
drmingdrmer: @SGZW not for debug. It is used to identify a unique version of some data. Normally the caller just checks the `seq` to learn if the `table` is change. And maybe `seq` is used to invalidate caches.

SGZW: i add seq for table in other pr

store/src/executor/storage_handlers.rs
```diff
@@ -54,10 +57,34 @@ impl RequestHandler<TruncateTableAction> for ActionHandler {
...
+                if let Some(prev) = prev {
+                    Ok(TruncateTableResult {
+                        trancated_table_data_parts_count: prev,
+                    })
+                } else {
+                    Err(ErrorCode::UnknownTable(format!(
+                        "table not found: {:}",
+                        tbl_name
+                    )))
+                }
```
drmingdrmer, line 85: I supposed that `prev >= 0` and `result == 0` indicates a successful truncate.

`prev` being a `Some()`  does not imply some change to the table has taken place. There could be that there is such a table, but for some reason(no privilege?) truncating is canceled.

drmingdrmer (commented):
Thank you bro.

There may be a flaw about the result checking.

store/src/executor/storage_handlers.rs
SGZW: ok,  that's more clearly

store/src/meta_service/state_machine.rs
```diff
@@ -571,6 +571,23 @@ impl StateMachine {
...
             }
+
+            Cmd::TruncateTable {
+                ref db_name,
+                ref table_name,
+            } => {
+                let db = self.databases.get_mut(db_name).unwrap();
+                let tbl_id = db.tables.get(table_name);
+                if let Some(tbl_id) = tbl_id {
+                    let _tbl_id = tbl_id.to_owned();
```
drmingdrmer, line 582:
```suggestion
```

drmingdrmer (approved):
Nice shot! Thank you!

PR #1479, opened 2021-08-16
[store] feature: applies all logs to raft state machine.

drmingdrmer:

##### [store] feature: applies all logs to raft state machine.
pass all logs to apply_entry_to_state_machine(), not just Normal logs.
Pass `Entry<D>` to `apply_entry_to_state_machine()`, not just the only
`EntryPayload::Normal(normal_log)`.

Thus the state machine is able to save the membership changes if it
prefers to.

Why:

In practice, a snapshot contains info about all applied logs, including
the membership config log.
Before this change, the state machine does not receive any membership
log thus when making a snapshot, one needs to walk through all applied
logs to get the last membership that is included in state machine.

By letting the state machine remember the membership log applied,
the snapshto creation becomes more convinient and intuitive: it does not
need to scan the applied logs any more.

Related raft changes:
https://github.com/datafuse-extras/async-raft/tree/v0.6.2-alpha.8

## Changelog

- New Feature

## Related Issues

- #271
- #1080

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,7 +170,7 @@ dependencies = [
 [[package]]
 name = "async-raft"
 version = "0.6.1"
-source = "git+https://github.com/datafuse-extras/async-raft?tag=v0.6.2-alpha.7#8e0cca5241fa423430920ab5c409494569bacb01"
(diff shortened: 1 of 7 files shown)


ariesdevil (approved):
These codes look like poetry to me:)

drmingdrmer:
> These codes look like poetry to me:)

Thanks...

PR #1505, opened 2021-08-18
[functions] add if function

### LINKED ISSUE #918: [functions] Conditional Functions
**Summary**

If
Controls conditional branching. Unlike most systems, ClickHouse always evaluate both expressions then and else.
Syntax:
```
SELECT if(cond, then, else)
```

Example:

Query:
```
SELECT if(1, 2+2, 2+6);
```
Result:
```
┌─plus(2, 2)─┐
│          4 │
└────────────┘
```

Query:
```
SELECT if(0, 2+2, 2+6);
```

Result:
```
┌─plus(2, 6)─┐
│          8 │
└────────────┘
```

Reference:
https://clickhouse.tech/docs/en/sql-reference/functions/conditional-functions/#if

zhyass:

## Summary

```
SELECT if(cond, then, else)
```

## Changelog

- New Feature

## Related Issues

Fixes #918 

## Test Plan

Unit Tests

Stateless Tests


common/datavalues/src/arrays/ops/if.rs
```diff
@@ -0,0 +1,112 @@
...
+use common_arrow::arrow::array::ArrayRef;
+use common_arrow::arrow::compute::if_then_else;
+use common_exception::ErrorCode;
+use common_exception::Result;
+
+use crate::prelude::*;
+
+macro_rules! impl_if_common {
+    ($lhs:ident, $rhs:ident,$predicate:ident) => {{
+        match ($predicate.len(), $lhs.len(), $rhs.len()) {
```
sundy-li, line 26: Why not keep the order as macros define.

common/datavalues/src/columns/conditional.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+        let mut right = right.clone();
+        if right.data_type() != dtype {
+            right = right.cast_with_type(&dtype)?;
+        }
+
+        let len = self.len();
+        match left.data_type() {
+            DataType::Null => Ok(lhs.resize_constant(lhs.len())),
+            DataType::Boolean => try_get_if_result! {left, right, cond, bool, len},
```
sundy-li, line 54: Can we add a method in SeriesTrait, then there will be no need to have this macros.

zhyass:
TODO: 
Unit Tests
Stateless Tests

common/datavalues/src/columns/conditional.rs
zhyass: Sure, I will do it

common/datavalues/src/arrays/ops/if.rs
zhyass: fix

sundy-li (approved):
LGTM

Issue #1389, opened 2021-08-10
Support SHOW TABLES WHERE statement


zhang2014:
**Summary**

https://www.db-fiddle.com/f/xkhhN3tfxXnUepeDPJ86rA/1
```
CREATE TABLE test_table(id varchar(200));
SHOW TABLES WHERE `Tables_in_test` = 'ss';
SHOW TABLES WHERE `Tables_in_test` = 'test_table';
```

zhang2014:
@SGZW 
It seems @maplefu has solved this problem

SGZW:
> @SGZW
> It seems @mapleFU has implemented this : )

thanks for notice

mapleFU:
Seems that we can close this issue @zhang2014

zhang2014:
> Seems that we can close this issue @zhang2014

Yes.

> thanks for notice

 @SGZW thank you for your contribution too.

PR #1518, opened 2021-08-18
Support SHOW TABLES LIKE pattern and Where expr 

### LINKED ISSUE #1388: Support SHOW TABLES LIKE `pattern` and Where expr
**Summary**

https://www.db-fiddle.com/f/9941isstm4v3dvWhye9ZPg/0
```
CREATE TABLE `table_1`(id varchar(200));
SHOW TABLES LIKE 'TABLE_1';
SHOW TABLES LIKE '%unknown%';

SHOW TABLES LIKE 'table_1';
SHOW TABLES LIKE '%_1';
SHOW TABLES LIKE 'ta%';
SHOW TABLES LIKE '%le%';

SHOW TABLES where xxx;
```

mapleFU:

## Summary

Summary about this PR

This pr supports `like 'pattern'` and `where expr` in `show tables`. I use some interface of sqlparse, and I don't know does I really catch it.

## Changelog

- New Feature

## Related Issues

Fixes #1388 #1389

## Test Plan

Unit Tests


sundy-li:
@mapleFU LGTM, better to add a stateless test in `06_0000_show_queries.sql`

query/src/sql/sql_parser.rs
```diff
@@ -163,7 +163,20 @@ impl<'a> DfParser<'a> {
                         self.parser.next_token();
 
                         if self.consume_token("TABLES") {
-                            Ok(DfStatement::ShowTables(DfShowTables))
+                            let tok = self.parser.next_token();
+                            match &tok {
+                                Token::EOF => Ok(DfStatement::ShowTables(DfShowTables::All)),
```
zhang2014, line 168: What happens to the following SQL?
``` SQL
SHOW TABLES  ; 
SHOW TABLES -- comment
```

query/src/sql/plan_parser.rs
```diff
@@ -133,13 +134,29 @@ impl PlanParser {
...
+                            self.ctx.get_current_database()
+                        )
+                    }
+                    DfShowTables::Like(i) => {
+                        format!(
+                            "SELECT name FROM system.tables where database = '{}' AND name LIKE {} ORDER BY database, name",
+                            self.ctx.get_current_database(), i,
+                        )
+                    }
+                    DfShowTables::Where(e) => {
```
zhang2014, line 151: What happens to the following SQL?
```
SHOW TABLES WHERE 1 = 1  OR 1 = 1 
```

query/src/sql/sql_parser.rs
mapleFU: sorry, I will add some tests and check if it's ok tonight.

query/src/sql/plan_parser.rs
sundy-li: > What happens to the following SQL?
> 
> ```
> SHOW TABLES WHERE 1 = 1  OR 1 = 1 
> ```

It's ok, we don't care about it, just let it be any complex expr. The optimize / select executor will handle that.

mapleFU: According to MySQL standard,`parse_expr` can handle all of them: https://dev.mysql.com/doc/refman/5.7/en/expressions.html

Buf I didn't check the code in sqlparse-rs, so I cannot guarantee it works well.

zhang2014: May this change the behavior of the query?

After format: 
`SELECT name FROM system.tables where database = '{}' AND 1 = 1 OR 1 = 1 ORDER BY database, name`

mapleFU: > May this change the behavior of the query?
> 
> After format:
> `SELECT name FROM system.tables where database = '{}' AND 1 = 1 OR 1 = 1 ORDER BY database, name`

Got It. Sorry for misunderstanding this case! I will adding tests about it in stateless test.

mapleFU: > evaluation proceeds left to right, with the exception that assignments evaluate right to left

According to https://dev.mysql.com/doc/refman/8.0/en/operator-precedence.html , it will not return all tables. But I think maybe it's proper to build sql like:
```
SELECT name FROM system.tables where database = '{}' AND (1 = 1 OR 1 = 1) ORDER BY database, name
```

@zhang2014 do you think it's ok?

zhang2014: > @zhang2014 do you think it's ok?

It looks ok

mapleFU:
@zhang2014 @sundy-li Sorry for fixing the problem so late!

I add some tests in `stateless`, and wrap `where expr` with `()`.

There are some cases:
1. commit `d9eaee0` remove the handling about `;`, because it was already processed.
2. When I build sql like `from system.tables where ... name = t or 1 = 1` without `()`, some system table was printed. I have tried it on MySQL but mysql will not print these messages.

sundy-li (approved):
LGTM, thanks @mapleFU

Issue #1218, opened 2021-07-28
[Tracking] Refactor SQL Parser


leiysky:
Part of #1217 

## Background
To provide a more generic parser abstraction, we need to do some refactor on the current implementation.

Generally, all of the SQL dialects follow some basic rules, with which we can design our own SQL IR(i.e. datafuse specific AST).

With this IR, we can transform the SQL queries in different dialects into the same representation.

The major idea is provide series of standard AST of Datafuse that can be directly parsed from Datafuse SQL dialect. The elements can be catagorized as:

- `Expression`s, represents expression entries in a SQL query including identifier, arithmetic operators and functions etc.
- `Statement`s, represents SQL clauses, including `SELECT`, `FROM`, `WHERE`, `GROUP BY`, `HAVING`, `ORDER BY`, `LIMIT` and DDL statements.

<del>Beyond this, we can provide interface like `ASTTransformer` to transform AST produced by other parsers(e.g. MySQL, ClickHouse, Postgres) into the standard AST.</del>

### Update: 2021.8.20

We have introduced a series of standard AST API in #1478 .

The main work during this stage is:
- Provide ability to analyze an query with AST(e.g. simple rewrting, expression binding)
- Provide a framework to build logical plan from AST, with normalization.
- Design Datafuse standard SQL #1492  and implement a parser for it (under discussing in #866).

## Tasks

High priority:
- [ ] Implement SQL parser for Datafuse standard SQL #866 
    - [ ] Evaluate SQL dialect framework #1551
- [ ] Standard SQL AST
    - [x] Introduce Datafuse standard SQL AST API @leiysky #1477 
    - [x] Fold table references in FROM clause with CROSS JOIN @leiysky  #1523  
    - [x] Refine object name with qualified typed fields @leiysky #1540
    - [ ] Implement `Visitor` utility for AST @jackwener  #1511 
- [ ] Expression binding and normalization
    - [ ] Bind expression with catalog @leiysky #1537 
    - [ ] Do type check for expressions @leiysky #1538 
    - [ ] Implement rewriter for AST to help normalizing query @leiysky #1539 

Low priority:
- [ ] SQL dialect transformer for ClickHouse SQL
- [ ] SQL dialect transformer for MySQL
- [ ] SQL dialect transformer for Postgres
- [ ] Complete `TransformerSqlparser` @SGZW #1512

leiysky:
Related to #866

leiysky:
[...]

PR #1523, opened 2021-08-19
parser: Fold table references in FROM clause

leiysky:

## Summary
Part of #1218 

Minor change, refactor `FROM` clause, use `CROSS JOIN` to replace comma seperated `TableReference`.

## Changelog

- Improvement

### DIFF
--- a/query/src/sql/parser/transformer/transform_sqlparser.rs
+++ b/query/src/sql/parser/transformer/transform_sqlparser.rs
@@ -200,11 +200,24 @@ impl TransformerSqlparser {
         })
     }
 
-    fn transform_from(&self, orig_ast: &[TableWithJoins]) -> Result<Vec<TableReference>> {
-        orig_ast
+    fn transform_from(&self, orig_ast: &[TableWithJoins]) -> Result<TableReference> {
+        let mut table_refs: Vec<TableReference> = orig_ast
             .iter()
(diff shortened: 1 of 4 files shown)


query/src/sql/parser/transformer/transform_sqlparser.rs
```diff
@@ -200,11 +200,24 @@ impl TransformerSqlparser {
...
-    fn transform_from(&self, orig_ast: &[TableWithJoins]) -> Result<Vec<TableReference>> {
-        orig_ast
+    fn transform_from(&self, orig_ast: &[TableWithJoins]) -> Result<TableReference> {
+        let mut table_refs: Vec<TableReference> = orig_ast
             .iter()
             .map(|v| self.transform_table_with_joins(v))
-            .collect()
+            .collect::<Result<_>>()?;
+        if table_refs.len() == 1 {
+            Ok(table_refs.drain(..).next().unwrap())
```
PsiACE, line 209: looks good. But maybe we can reduce the use of `unwrap()`?

leiysky: I think so, do you have better idea?

PsiACE: Not yet, let's merge first

PsiACE (approved):

PR #1524, opened 2021-08-19
Add optimize histogram metric

jyz0309:
Signed-off-by: Alkaid <alkaid@AlkaiddeMacBook-Pro.local>

## Summary

Add a histogram metric for optimizer. We can use this metric to know the P99,P90....e.t.c used time by optimizer.

## Changelog

- Other 

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/optimizers/metrics.rs
+++ b/query/src/optimizers/metrics.rs
@@ -1,0 +1,15 @@
+// Copyright 2020 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


jyz0309:
@BohuTANG PTAL

jyz0309:
BTW, I think we can add this type of metric in every module. Just like `executor`,`parser`.

BohuTANG (approved):
Thank you @jyz0309

BohuTANG:
> BTW, I think we can add this type of metric in every module. Just like `executor`,`parser`.

Right, that's what we did

PR #1526, opened 2021-08-19
[store]: refactor: upgrade raft to 0.6.2-alpha.10

drmingdrmer:

##### [store]: refactor: upgrade raft to 0.6.2-alpha.10
change-membership should be log driven but not channel driven

https://github.com/datafuse-extras/async-raft/compare/v0.6.2-alpha.9...datafuse-extras:v0.6.2-alpha.10

A membership change involves two steps: the joint config phase and the
final config phase.
Each phase has a corresponding log invovled.

Previously the raft setup several channel to organize this workflow,
which makes the logic hard to understand and introduces complexity when
restarting or leadership transfered: it needs to re-establish the channels and tasks.

According to the gist of raft, all workflow should be log driven.
Thus the new approach:
- Write two log(the joint and the final) at once it recevies a
  change-membership request.
- All following job is done according to just what log is committed.

This simplifies the workflow and makes it more reliable and intuitive to
understand.

Related changes:

- When `change_membership` is called, append 2 logs at once.

- Introduce universal response channel type to send back a message when
  some internal task is done: `ResponseTx`, and a universal response
  error type: `ResponseError`.

- Internal response channel is now an `Option<ResponseTx>`, since the
  first step of membership change does not need to respond to the
  caller.

- When a new leaser established, if the **last** log is a joint config
  log, append a final config log to let the partial change-membership be
  able to complete.

  And the test is added.

- Removed membership related channels.

##### [store] refactor: add per-case logging
Introduce common_tracing::init_tracing() to create a logger that will be
released when being dropped.

Every test need to hold a `guard` to ensure the log subscriber
working.

Elaborate some tracing span arguments.

## Changelog

- Improvement

## Related Issues

- #271
- #1080

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1453,6 +1453,7 @@ dependencies = [
  "tokio-stream",
  "tonic",
  "tonic-build",
+ "tracing-appender",
  "uuid",
  "warp",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,7 +170,7 @@ dependencies = [
 [[package]]
 name = "async-raft"
 version = "0.6.1"
-source = "git+https://github.com/datafuse-extras/async-raft?tag=v0.6.2-alpha.9#8b59966dd0a6bf804eb0ba978b5375010bfbc3f3"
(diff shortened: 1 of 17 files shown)


BohuTANG (approved):
Cool
Thank you @drmingdrmer

PR #1558, opened 2021-08-21
add log of error for install_snapshot

drmingdrmer:

##### [store] test: fix the RaftError::Shutdown issue
When running unittest, with heavy cpu load, heartbeat message has chance
to be delayed too much.
When this happens, a follower tries to elect itself and revert the
current leader to a follower state.
This causes the leader to discard every in progress request it has
received. And the client receives an **inappropriate** error
`RaftError::Shuttingdown`.

This is actually not a bug.

In such case in practise, a client should re-fetch the latest leader
and retry.

For a unittest, we just extend the timeout to let tests pass happily,
even on a poor CI VM.

##### [store] refactor: refine tracing
- Every test case creates its own span as a root.
  This way the logging/tracing belonging to a single test is easy to
  grep.

- Upgrade to async-raft 0.6.2-alpha.11, with enhanced tracing:
  a span is send along with a message through channels.
  Thus a complete workflow can be tracked.

## Changelog

- Bug Fix

- Improvement
- Build/Testing/CI

## Related Issues

- #271
- #1080


drmingdrmer:
Just for test. Closed

PR #1575, opened 2021-08-23
bump arrow2

### LINKED ISSUE #1573: The dependence `arrow2` is old
**Summary**

Maybe we should dump arrow2. The dependence `arrow2` is old and should be upgraded into `v0.3.0`.

sundy-li:

## Summary

bump arrow2

## Changelog

- Improvement

## Related Issues

Fixes #1573

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/arrow/Cargo.toml
+++ b/common/arrow/Cargo.toml
@@ -16,7 +16,7 @@ simd = ["arrow/simd"]
 
 # Github dependencies
 arrow = { package = "arrow2", git="https://github.com/datafuse-extras/arrow2", rev = "824ad7e" }
-arrow-flight = { git="https://github.com/datafuse-extras/arrow2", rev = "824ad7e" }
+arrow-flight = { git="https://github.com/datafuse-extras/arrow2", rev = "43c101f" }
 parquet = {package = "parquet2", git = "https://github.com/datafuse-extras/parquet2", rev = "05854c0"}
 # Crates.io dependencies
 
--- a/common/arrow/Cargo.toml
+++ b/common/arrow/Cargo.toml
@@ -16,7 +16,7 @@ simd = ["arrow/simd"]
 
 # Github dependencies
 arrow = { package = "arrow2", git="https://github.com/datafuse-extras/arrow2", rev = "824ad7e" }
-arrow-flight = { git="https://github.com/datafuse-extras/arrow2", rev = "647fddb" }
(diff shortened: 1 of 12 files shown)


common/arrow/Cargo.toml
```diff
@@ -15,7 +15,7 @@ simd = ["arrow/simd"]
 # Workspace dependencies
 
 # Github dependencies
-arrow = { package = "arrow2", git="https://github.com/datafuse-extras/arrow2", rev = "647fddb" }
```
jackwener, line 18: It's not enough because the high-version library isn't compatible with old-version.
Need more modification.

sundy-li:
A bug has been encountered.  @jorgecarleitao

In stateless tests:

```
select if(number>1, true, false) from numbers(3) order by number;

Got:
false
false
false

Expect:
false
false
true
```

How to reproduce:

```

#[test]
fn test_gt_u64_scalar() {
    let a = UInt64Array::from_slice(vec![0,1,2]);
    let c = gt_scalar(&a, 0);
    assert_eq!(BooleanArray::from_slice(vec![false, false, true]), c);
}

running 1 test
thread 'compute::comparison::primitive::tests::test_gt_u64_scalar' panicked at 'assertion failed: `(left == right)`
  left: `BooleanArray { values: [0b_____100], validity: None, offset: 0 }`,
 right: `BooleanArray { values: [0b_____110], validity: None, offset: 0 }`', src/compute/comparison/primitive.rs:372:9
stack backtrace:
```

jorgecarleitao:
Shouldn't the test be 

```
#[test]
fn test_gt_u64_scalar() {
    let a = UInt64Array::from_slice(&[0,1,2]);
    let c = gt_scalar(&a, 0);
    assert_eq!(BooleanArray::from_slice(&[false, true, true]), c);
}
```

i.e.

```
0 > 0 = false
1 > 0 = true
2 > 0 = true
```

sundy-li:
> Shouldn't the test be

Yes, I made a mistake,  the tests code should be `gt_scalar(&a, 1)`.

But the ci in `macos-latest` is failed, test case :

`select if(number>1, true, false) from numbers(3) order by number;`

got: 

```
false
false
false
```

Can't reproduce it in my local mac, I will try find a reproduce case to fix this.

jorgecarleitao:
No worries. I also tried to reproduce it on a linux with both `simd` and no `simd` but I am also unable to do so :/

zhyass (approved):

PR #1592, opened 2021-08-24
Add common/dal README

BohuTANG:

## Summary

Add common/dal README

## Changelog

- Improvement

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/dal/README.md
+++ b/common/dal/README.md
@@ -1,0 +1,17 @@
+# DAL
+
+A data access layer (DAL) in Datafuse is a layer which provides methods for accessing to data stored in shared storage, such as AWS S3, Azure Blob, DFS or Local disk.
+
+The API like:
+```rust
+    async fn get_input_stream(
+    &self,
+    path: &str,
+    stream_len: Option<u64>,
+) -> Result<Self::InputStream>;
+
+async fn get(&self, path: &str) -> Result<Bytes>;
+
+async fn put(&self, path: &str, content: Vec<u8>) -> common_exception::Result<()>;
+```
+
--- a/common/dal/src/blob_accessor.rs
+++ b/common/dal/src/blob_accessor.rs
@@ -1,4 +1,4 @@
-//  Copyright 2020 Datafuse Labs.
+//  Copyright 2021 Datafuse Labs.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
--- a/common/dal/src/impls/aws_s3/mod.rs
+++ b/common/dal/src/impls/aws_s3/mod.rs
@@ -1,4 +1,4 @@
-//  Copyright 2020 Datafuse Labs.
+//  Copyright 2021 Datafuse Labs.
 //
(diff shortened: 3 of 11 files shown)


common/dal/README.md
```diff
@@ -0,0 +1,17 @@
...
+```rust
+    async fn get_input_stream(
+    &self,
+    path: &str,
+    stream_len: Option<u64>,
+) -> Result<Self::InputStream>;
+
+async fn get(&self, path: &str) -> Result<Bytes>;
+
+async fn put(&self, path: &str, content: Vec<u8>) -> common_exception::Result<()>;
```
sundy-li, line 15: Should we introduce`append` function?

The put function is to override the contents, `append` function can help to continuously write the stream into the local disk.

BohuTANG: @dantengsky any thought?

dantengsky: @sundy-li  

I might misunderstand the questions, pls allow me try to answer them:

> Should we introduce `append` function?

We can simulate `append` operation for AWS S3, like

concat "previous" content and the content being appended **locally**,  and then, put the whole new thing to s3 (using the same bucket and key)

but I am afraid that will not be efficient enough to be used in serious cases.

 
> The put function is to override the path with contents, append function can help to continuously write the stream into the local disk.

**local disk**? does it mean interface like this?

```
pub async fn get_object_stream<T: Write + Send, S: AsRef<str>>(
    &self,
    path: S,
    writer: &mut T
) -> Result<u16> 
```
from [rust_s3](https://durch.github.io/rust-s3/s3/bucket/struct.Bucket.html#method.get_object_stream) 

If it does mean that, we can definitely have it; I'd like to add it in another PR

BohuTANG:
@dantengsky Please take a review for this PR :)

dantengsky:
rogar

dantengsky (approved):
LGTM & thanks

common/dal/README.md
sundy-li: > If it does mean that, we can definitely have it; I'd like to add it in another PR

Yes, we want to get_object_stream from the path rather than send a batch of bytes at once.

dantengsky: rogar.

At meantime, we have a stream reading  interface

>     type InputStream: AsyncRead + AsyncSeek;
> 
>     async fn get_input_stream(
>         &self,
>         path: &str,
>         stream_len: Option<u64>,
>     ) -> Result<Self::InputStream>;

dantengsky: > from the path rather than send a batch of bytes at once.


 Oh, I think I get your concerns --- we should provide a stream uploading interface as well.

 That's right, totally agree with you.
 
 there was a method like this
 
>    type OutputStream: AsyncWrite;
>     async fn get_output_stream(&self, path: &str) -> Result<Self::OutputStream>;

But this method and its impl are deleted in PR #1407 for some technical reason.
Mainly because I failed to find a rust sdk that supports chunked transfer, which basically means we should know the length of the stream in advance, which I do not feel comfortable with.
Bue stream uploading from a local file is definitely feasible, thanks for reminding me of this issue!

PR #1604, opened 2021-08-25
Fix deploy bug about http service port

jyz0309:
Signed-off-by: jyz0309 <a1kaid@foxmail.com>

## Summary

For now, the http service default port is different from the make run default port(`datafuse-query-node-1.toml`).

## Changelog

- Bug Fix
- Documentation

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/store/src/configs/config.rs
+++ b/store/src/configs/config.rs
@@ -54,7 +54,7 @@ pub struct Config {
     )]
     pub metric_api_address: String,
 
-    #[structopt(long, env = "HTTP_API_ADDRESS", default_value = "127.0.0.1:8181")]
+    #[structopt(long, env = "HTTP_API_ADDRESS", default_value = "127.0.0.1:8081")]
     pub http_api_address: String,
 
     #[structopt(long, env = "TLS_SERVER_CERT", default_value = "")]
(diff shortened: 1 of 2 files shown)


store/src/configs/config.rs
```diff
@@ -54,7 +54,7 @@ pub struct Config {
     )]
     pub metric_api_address: String,
 
-    #[structopt(long, env = "HTTP_API_ADDRESS", default_value = "127.0.0.1:8181")]
+    #[structopt(long, env = "HTTP_API_ADDRESS", default_value = "127.0.0.1:8081")]
```
PsiACE, line 57: I don't think this can really solve this problem.

website/datafuse/docs/api/config.md
```diff
@@ -8,7 +8,7 @@ Get the config of the Datafuse query server.
 ## Examples
 
 ```
-curl http://127.0.0.1:8080/v1/configs
+curl http://127.0.0.1:8081/v1/configs
```
PsiACE, line 11: There are many relevant ports that are 8080, so we should make changes in a separate PR and make sure that all descriptions are for the same specific run

PsiACE (changes requested):

PR #1633, opened 2021-08-27
ISSUE-1632: fix range_remove index in log compaction

### LINKED ISSUE #1632: Concerns about the raft log index
Hi,

I have a question about the index of the log. Is it correct to start from`0`?
https://github.com/datafuselabs/datafuse/blob/master/store/src/meta_service/raftmeta.rs#L457

following these instructions https://github.com/datafuse-extras/async-raft/blob/master/raft-essentials.md#can-the-log-index-reset-per-term, shouldn't it be the `last snapshot index` up to the `current snapshot index`?

ariesdevil:

## Summary

When do log compaction, the range should start from `last_snapshot_index + 1`.

## Changelog

- Bug Fix

## Related Issues

Fixes #1632 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/store/src/meta_service/raftmeta.rs
+++ b/store/src/meta_service/raftmeta.rs
@@ -454,18 +454,18 @@ impl RaftStorage<LogEntry, AppliedState> for MetaStore {
             .insert(&Entry::new_snapshot_pointer(&snapshot.meta))
             .await?;
 
-        let current_snapshot_last_log_index;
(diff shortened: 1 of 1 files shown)


drmingdrmer:
https://github.com/datafuselabs/datafuse/issues/1632#issuecomment-908026668
This should not be a problem 🤔

PR #1637, opened 2021-08-28
Improve contribute document

GrapeBaBa:

## Summary

Improve contribute document

## Changelog

- Documentation

## Related Issues

no issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/website/datafuse/docs/development/contributing.md
+++ b/website/datafuse/docs/development/contributing.md
@@ -18,6 +18,8 @@ Thank you.
 
 ## Contributing
 
+Note: In order to run `make test` successfully, make sure you have installed `mysql` client.
+
 To contribute to Datafuse, ensure that you have the latest version of the codebase, run the following:
 ```
 $ git clone https://github.com/datafuselabs/datafuse


website/datafuse/docs/development/contributing.md
```diff
@@ -18,6 +18,8 @@ Thank you.
 
 ## Contributing
 
+Note: In order to run `make test` successfully, make sure you have installed `mysql` client.
```
BohuTANG, line 21: The step `make setup` will install mysql client:
https://github.com/datafuselabs/datafuse/blob/master/scripts/setup/dev_setup.sh#L326

BohuTANG: I think we should make the `OPERATIONS` to true by default in dev_setup.sh

GrapeBaBa: `if [[ "$INSTALL_BUILD_TOOLS" == "false" ]] && \
   [[ "$OPERATIONS" == "false" ]] && \
   [[ "$INSTALL_PROFILE" == "false" ]] && \
   [[ "$INSTALL_PROVER" == "false" ]] && \
   [[ "$INSTALL_CODEGEN" == "false" ]]; then
   INSTALL_BUILD_TOOLS="true"
fi`

change to true will break this?

BohuTANG: There are two ways for the dev_setup.sh to install MySQL client:
1. set `OPERATIONS` to true by default in dev_setup.sh
2. Makefile
setup:
        bash ./scripts/setup/dev_setup.sh -o

GrapeBaBa: set OPERATIONS to true by default in dev_setup.sh will cause INSTALL_BUILD_TOOLS="false", it will not install build tools,right?

BohuTANG: > set OPERATIONS to true by default in dev_setup.sh will cause INSTALL_BUILD_TOOLS="false", it will not install build tools,right?
Y.
I have check the codes dev_setup.sh, it seems `OPERATIONS` is not make sense here, MySQL client is a tool, I am not sure why we use `OPERATIONS` to enable it.
@ZhiHanZ may help here

BohuTANG: Hi @GrapeBaBa 

This issued here https://github.com/datafuselabs/datafuse/issues/1650

Shall we close the PR first?
Thank you for the contribution and suggestion.

GrapeBaBa:
ok

PR #1638, opened 2021-08-28
[common/streams] use arrow2 bitmap to improve memory usage

cppcoffee:

## Summary

Use arrow2 bitmap to improve memory usage

## Changelog

- Improvement

## Related Issues

None

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/streams/src/stream_limit_by.rs
+++ b/common/streams/src/stream_limit_by.rs
@@ -18,10 +18,11 @@ use std::task::Context;
 use std::task::Poll;
 
 use common_arrow::arrow;
+use common_arrow::arrow::array::BooleanArray;
+use common_arrow::arrow::bitmap::MutableBitmap;
 use common_datablocks::DataBlock;
 use common_datablocks::HashMethod;
 use common_datablocks::HashMethodSerializer;
-use common_datavalues::prelude::*;
 use common_exception::Result;
 use futures::Stream;
 use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


common/streams/src/stream_limit_by.rs
```diff
@@ -50,20 +52,19 @@ impl LimitByStream {
     }
 
     pub fn limit_by(&mut self, block: &DataBlock) -> Result<Option<DataBlock>> {
-        // TODO: use BitVec here.
```
jorgecarleitao, line 53: I would suggest to use `arrow2::bitmap::MutableBitmap` here and then use `BooleanArray::from_data(filter_vec.into(), None)`. This should be at least 10x faster, as we avoid running over all bits twice.

```rust
let mut filter_vec = arrow2::bitmap::MutableBitmap::from_zero_len(block.num_rows())
...
filter_vec.set(row as usize, *count <= self.limit)
...

let array = arrow2::array::BooleanArray::from_data(filter_vec.into(), None);
// convert to DFBooleanArray
```

`bitvec` and `arrow2::bitmap` (`MutableBitmap` specifically) overlap in functionality, but creating a bitmap directly from arrow avoids a full scan over all bits and an extra memory allocation. :)

cppcoffee: Ok, should I open a new PR or modify it in the current one?

BohuTANG: Modify is the simple way?

cppcoffee: Ok, I'll fix it.

BohuTANG (approved):
LGTM
Thank you @cppcoffee

PR #1652, opened 2021-08-30
CATALOG-1651: create database from engine for catalog

### LINKED ISSUE #1568: Refactoring: Lift `LocalDatabase` into DataSource APIs
**Summary** 

TL;DR:

Suggestion:

- Decouple `LocalDatabase` into
  - TableFactoryRegistry
  - TableFactory 
- Provides an embedded MetaStore
  For unit tests. in-memory or backed by embedded KV DB
- Keep all the db metadata in MetaStore (except for `system` like virtual db/tables)
-----------------------------------------

`LocalDatabase` is quite convenient for us to verifying new features, or doing performance benchmarks. 
- we can bring up a test env easily without `Store`, it is a gift for UT
- `LocalDatabase` provides the functionalities of customizing `Table`  by `plan.engine`,  e.g. reading CSV/parquet data from the local filesystem/memory, it almost looks like DataSource.

But there is always room for enhancement:

- `LocalDatabase` is treated specially in Catalog;  local dbs and tables belonging to a local database are registered/created locally. Different computation nodes may have different sets of local databases/tables.  IMHO, this is not what we want (temp/tx level tables/views are quite different things I think)
- For the `special treatment`,  the implementation of Catalog looks at least a little bit messy: -(
- Although nothing prevents us from reading data from S3 in a Table which belongs to a `LocalDatabase`, that might be somehow weird ( _but not unacceptable, after a re-thought_)

IMHO, ideally, DB metadata and the interpretation of metadata into `Table` could be separated more clearly:

- DB metadata should belong to MetaStore(no matter Remote Or Embedded)
- Catalog is a component that exposes structural DB metadata to `Query` 
  - with friendly methods
  - and Virtual Database/Tables, i.e. the `system` db and virtual tables belong to it (there may be more as we keep going)
   
Thus, I suggest (roughly)

- Decouple `LocalDatabase` into
  - TableFactoryRegistry
  - TableFactory 
- Provides an embedded MetaStore
  For unit tests. in-memory or backed by embedded KV DB
- Keep all the db metadata in MetaStore 
  - except for `system` like virtual db/tables, which are views of db meta and system state

Note: In clickhouse, one can create a database by engine, e.g. if we create a clickhouse database, which engine is `MySQL`, CK will retrieve table metadata from the specified MySQL db.  The above suggestion does NOT provide such abilities.

### LINKED ISSUE #1651: [improvement] Add database engine to catalog
**Summary**

Catalog has map for  database type register, such as:
```
catalog::register_database_type("local", LocalDatabase::create);
catalog::register_database_type("remote", RemoteDatabase::create);
catalog::register_database_type("mysql", MySQLDatabase::create);
```

When we create a database from catalog, we can check the engine type and choose the right Database type and create them, no longer concerned it's Local or Remote .

BohuTANG:

## Summary

This PR including:
* Change the catalog functions all to `sync`
* Introduce new config  `MetaConfig` in config
* Make Catalog plugable, we can register them by the engine type,  now we support 3 engines: Local/Remote/System, and easy to support a new engine like MySQL, and the session manager register them when session creating.
* Introduce two meta backends: LocalMetaBackend and RemoteMetaBackend, each backend maintains the databases and tables for the engine.

## Changelog

- Improvement

## Related Issues
Fixes #1568
Fixes #1651 

## Test Plan

Unit Tests

Stateless Tests


dantengsky (approved):
LGTM & Salute

PR #1694, opened 2021-09-01
Initial work for saving table meta in obj storage

dantengsky:

## Summary

Initial work for:

- Save table partition/statistics  data into object storage
- First step of integrating DAL/Cache layers

The remaining jobs will be split in other PRs

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/arrow/Cargo.toml
+++ b/common/arrow/Cargo.toml
@@ -11,7 +11,8 @@ edition = "2021"
 [features]
 default = ["arrow-default", "parquet-default"]
 arrow-default = ["arrow/compute", "arrow/regex", "arrow/merge_sort", "arrow/io_csv", "arrow/io_parquet", "arrow/io_json"]
-parquet-default = ["parquet/stream"]
+#parquet-default = ["parquet/stream"]
+parquet-default = ["parquet2/stream"]
 simd = ["arrow/simd"]
 
 [dependencies] # In alphabetical order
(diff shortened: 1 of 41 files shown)


common/arrow/Cargo.toml
```diff
@@ -17,7 +17,8 @@ simd = ["arrow/simd"]
 # Github dependencies
 arrow = { package = "arrow2", git="https://github.com/datafuse-extras/arrow2", rev = "7765067" }
 arrow-flight = { git="https://github.com/datafuse-extras/arrow2", rev = "7765067" }
-parquet = {package = "parquet2", git = "https://github.com/datafuse-extras/parquet2", rev = "d28330f"}
+#parquet2 = {package = "parquet2", git = "https://github.com/datafuse-extras/parquet2", rev = "d28330f"}
 # Crates.io dependencies
+parquet2 = { version = "0.3", optional = false, default_features = false, features = ["stream"] }
```
BohuTANG, line 22: Arrow2 has parquet already https://github.com/datafuse-extras/arrow2/tree/main/src/io/parquet
Do we still need the parquet2 crate here?

dantengsky: At present, we need `parquet2` to extract/manipulate parquet Statistic.

To avoid implicit parquet2 dependency in Cargo.toml, we may need to re-export those types from datafuse-extras/arrow2 ()

BohuTANG: Got, here import parquet2 crate makes sense.

jorgecarleitao: Yes, if you just want to interact with parquet without serialization and deserialization of pages, then this would also be my recommendation: add a new dependency.

I would probably bump it to 0.4 so that it is aligned with arrow2' dependency and reduce the binary size a bit (parquet2 is very small, though). [Other 2 benefits of 0.4](https://github.com/jorgecarleitao/parquet2/releases/tag/v0.4.0) are:

1. writer no longer needs `Seek`.
2. supports async writers via `AsyncWrite`.

dantengsky: @jorgecarleitao
roger & thanks

dantengsky:
@PsiACE  @BohuTANG 

I scratched an imaginary API for integrating FuseTable with Indexing , please take a look:

https://github.com/datafuselabs/datafuse/blob/cfcba2fb0a95008e68f81eae9dc0e1314a5d28a9/query/src/datasources/fuse_table/util/index_helpers.rs#L25-L55

BohuTANG:
@dantengsky 
Hello, this is a big patch and I would suggest reviewing&merging it first and then doing iterations so that everyone's interface is clearer.
If it is ready for reviewing, please make `Ready for review`.
Thank you

dantengsky:
> @dantengsky
> Hello, this is a big patch and I would suggest reviewing&merging it first and then doing iterations so that everyone's interface is clearer.
> If it is ready for reviewing, please make `Ready for review`.
> Thank you

Roger.

I  withdrew "ready for review"  cause I forget to squash the commit log (which is messy ;-).

Squashed commits have been force-pushed, I will re-request for review whenever CI is passed.

PR #1715, opened 2021-09-03
Add timeSlot() round functions

### LINKED ISSUE #853: [functions] Functions Working with Date/DateTime
From  https://clickhouse.tech/docs/en/sql-reference/functions/date-time-functions/

prio order from high to low
- [x] now (#1648)
- [x] today (#1648)
- [x] yesterday (#1673)
- [x] tomorrow (#1673)
- [x] toDate, toDateTime    (cast number/string into Date/DateTime)  
- [x] timeSlot (#1715)
- [x] toYYYYMM (#1726)
- [x] toYYYYMMDD (#1767)
- [x] toYYYYMMDDhhmmss (#1767)
- [x] toStartOfYear (#1826)
- [x] toStartOfISOYear (#1826)
- [x] toStartOfQuarter #1848
- [x] toStartOfMonth  #1848
- [x] toStartOfWeek (#1942) 
- [x] toStartOfDay (#1715)
- [x] toStartOfHour (#1715)
- [x] toStartOfMinute (#1715)
- [x] toStartOfSecond (#1715)
- [x] toStartOfFiveMinute (#1715)
- [x] toStartOfTenMinutes (#1715)
- [x] toStartOfFifteenMinutes  (#1715)
- [x] toMonth (#2042)
- [x] toDayOfYear (#2050)
- [x] toDayOfMonth (#2050)
- [x] toDayOfWeek (#2050)
- [x] toHour (#2309)
- [x] toMinute (#2309)
- [x] toSecond (#2309)
- [x] toMonday (#2317)

Low priority:
- timeZone
- toTimeZone
- timeZoneOf
- timeZoneOffset
- toStartOfInterval
- toYear
- toQuarter
- toUnixTimestamp
- toTime
- toRelativeYearNum
- toRelativeQuarterNum
- toRelativeMonthNum
- toRelativeWeekNum
- toRelativeDayNum
- toRelativeHourNum
- toRelativeMinuteNum
- toRelativeSecondNum
- toISOYear
- toISOWeek
- toWeek
- toYearWeek
- date_trunc
- date_add
- date_diff
- date_sub
- timestamp_add
- timestamp_sub
- addYears
- subtractYears
- timeSlots
- formatDateTime
- dateName
- FROM_UNIXTIME
- toModifiedJulianDay
- toModifiedJulianDayOrNull
- fromModifiedJulianDay
- fromModifiedJulianDayOrNull

dust1:

## Summary

I use the `NaiveDateTime::arse_from_str ` to convert a string to an NaiveDateTime object.
then, i check minute and Makes a new NaiveDateTime.

Add 
- toStartOfSecond
- toStartOfMinute
- toStartOfFiveMinutes
- toStartOfTenMinutes
- toStartOfFifteenMinutes
- timeSlot
- toStartOfHour
- toStartOfDay

## Changelog

- New Feature

## Related Issues

Related #853

## Test Plan

Unit Tests 

Stateless Tests [ok]


common/functions/src/scalars/dates/times.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[derive(Clone)]
+pub struct TimeSlot;
+
+impl NoArgTimeFunction for TimeSlot {
+    fn execute() -> u32 {
+        let utc = Utc::now();
+        (utc.timestamp_millis() / 1000 - 60 * 30) as u32
+    }
```
sundy-li, line 41: That's wrong, what about the current time is exactly in half-hour mod?

Try using `utc.timestamp_millis() / 1000 / 1800 * 1800`

sundy-li:
1. Better to add docs for this function
2. How about to implement `timeSlots` in this pr ?

common/functions/src/scalars/dates/times.rs
dust1: > 1. Better to add docs for this function
> 2. How about to implement `timeSlots` in this pr ?

My fault.
I misunderstood the meaning of the timeSlot function

common/functions/src/scalars/dates/timeslot.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+        Ok(false)
+    }
+
+    fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
+        let fmt = "%Y-%m-%d %H:%M:%S";
+        let time_str: DataValue = columns[0].column().clone().to_values()?[0].clone();
+        let parse_result = NaiveDateTime::parse_from_str(time_str.to_string().as_str(), fmt);
+        let date = parse_result.unwrap();
```
BohuTANG, line 54: It would be better to return Err instead of unwrap to panic the thread

tests/suites/0_stateless/02_0012_function_datetimes.sql
```diff
@@ -16,3 +16,6 @@ select today() - 1 = yesterday();
 select today() - yesterday()  = 1;
 select today() + 1 = tomorrow();
 select tomorrow() - today() = 1;
+
+select timeSlot('2012-12-12 21:12:31');
```
BohuTANG, line 20: Good :heart: 
Add another one case?
```
select timeSlot('2012-12-12 xx:xx:31');
```

common/functions/src/scalars/dates/timeslot.rs
```diff
@@ -67,19 +47,29 @@ where T: NoArgTimeFunction + Clone + Sync + Send + 'static
...
     }
 
-    fn eval(&self, _columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
-        let value = T::execute();
-        Ok(DataColumn::Constant(
-            DataValue::UInt32(Some(value)),
-            input_rows,
-        ))
+    fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
+        let fmt = "%Y-%m-%d %H:%M:%S";
```
sundy-li, line 51: Actually, we do not want this.

We must avoid Implicit conversion between string and DateTimes/Datetime.

The timeSlot function's prototype is `timeSlot( DateTime) ` or `timeSlot()` (which equals to `timeslot(now()` ).
You don't need to care about the format in timeSlot function.

We can have another function to cast string into DateTime (that's not an easy task, because the time formats are very complex to consider)

common/functions/src/scalars/dates/timeslot.rs
```diff
@@ -47,24 +46,27 @@ impl Function for TimeSlotFunction {
...
-        let minute = date.minute();
-        let mut new_minute: u32 = 0;
-        if minute >= 30 {
-            new_minute = 30;
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        let mut timestamp = Utc::now().timestamp_millis() / 1000;
+        let len = columns.len();
+        if len > 0 {
+            let args = columns[0].column().clone().to_values()?;
+            timestamp = args[0].as_i64()?;
```
sundy-li, line 54: We can't just get one row as `timestamp`.

`timeSlow(DateTime)` function not just return DataColumn::Constant, it returns `DataColumn::Array`.


You can do this via 

```
let timecol = columns[0].column();
let timecol = timecol.u32()?;


timecol.iter()....
```

dust1: Does this mean there will be a behavior like "timeSlot(DateTime1, DateTime2...)"

sundy-li: Not yet, everything is column in functions :

- `timeSlot(x)` x is a column.
- `add(a,b)` a and b are columns.

common/functions/src/scalars/dates/timeslot.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+        Ok(false)
+    }
+
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        let timestamp = Utc::now().timestamp_millis() / 1000;
+        let mut times = vec![DataValue::Int64(Some(timestamp))];
+        let len = columns.len();
+        if len > 0 {
+            times = columns[0].column().to_values()?;
```
sundy-li, line 54: `to_values()` is not efficient, we can have native value iterator, check the implementation in toYYYYMM you will find the way.

common/functions/src/scalars/dates/timeslot.rs
```diff
@@ -0,0 +1,88 @@
...
+        Ok(DataColumn::Array(Series::new(result)))
+    }
+
+    fn variadic_arguments(&self) -> Option<(usize, usize)> {
+        Some((0, 1))
+    }
+}
+
+fn slot_time(timestamp: u32) -> u32 {
+    let mut minutes = ((timestamp / 60) % (60 * 24)) % 60;
```
sundy-li, line 77: this is too complex!

Why not Just use `timestamp / 1800 * 1800` ?

dust1: Sorry, I still can't understand the meaning of `timestamp / 1800 * 1800`. 
My current thinking is to calculate from the timestamp the number of minutes at this point in the day. The value is between 0 and 60, and then make an if judgment.
: (

sundy-li: > Sorry, I still can't understand the meaning of `timestamp / 1800 * 1800`.

It's a number rounding skill using the number divide.

`number / N * N` means to get the results `R` which `R` less or equals than number and `R` is multiple of N.


`number / N` means to divide the number into x parts, each part will have `N` numbers and ignore the remainder.

If the timestamp is UNIX seconds from the epoch.

- timestmap / 60 * 60 --> to start of current minute
- timestamp / 3600 * 3600  --> toStartOfHour
-  timestamp / 1800 * 1800  -> to start of half hour
-  timestamp / ( 3600 * 24 ) * (3600*24)  --> to start of today

sundy-li: @dust1  Thanks for your contribution, I think this pr almost looks good to me, I'll update this pr for you later.

dust1: I see. Thanks for teaching me.
This is a new field for me, Sorry for not doing well in some aspects

sundy-li (approved):

sundy-li:
@dust1 

> I see. Thanks for teaching me. This is a new field for me, Sorry for not doing well in some aspects.

Take your time, hope you can always learn new things from the community.

Another tip: don't use `master` as a branch name when you want to submit feature prs.

dust1:
> Take your time, hope you can always learn new things from the community.
> 
> Another tip: don't use `master` as a branch name when you want to submit feature prs. Try `feature/xxx`

receive : )

PR #1719, opened 2021-09-04
Add config to disable local engine register

### LINKED ISSUE #1679: [feature] add config to disable local database engine
**Summary**

In some cases, we don't want local database to create, add one config like 'disable_local_database_engine', when datafuse session creating, we won't load the local database engine.

jyz0309:
Signed-off-by: jyz0309 <45495947@qq.com>

## Summary

Add config to disable local engine register

## Changelog

- New Feature

## Related Issues

Fixes #1679 

## Test Plan

Unit Tests

Stateless Tests


query/src/configs/config.rs
```diff
@@ -147,6 +148,10 @@ pub struct StoreConfig {
     )]
     #[serde(default)]
     pub rpc_tls_store_service_domain_name: String,
+
+    #[structopt(long, env = "DISABLE_LOCAL_DATABASE_ENGINE", default_value = "0")]
```
BohuTANG, line 152: Look great :heart: 
But `DISABLE_LOCAL_DATABASE_ENGINE` is a `QueryConfig` group item, it used in the query layer.
BTW, the `StoreConfig` group only includes some configs for the store service.

jyz0309: OK, I will fix the comment tonight~

PR #1726, opened 2021-09-04
Add toYYYYMM function

### LINKED ISSUE #853: [functions] Functions Working with Date/DateTime

ygf11:

## Summary

impl date16/date32/dateTime32 to YYYYMM UInt32 number.

## Changelog

- New Feature

## Related Issues

Related #853 

## Test Plan

Unit Tests

Stateless Tests


tests/suites/0_stateless/02_0012_function_datetimes.sql
```diff
@@ -16,3 +16,9 @@ select today() - 1 = yesterday();
 select today() - yesterday()  = 1;
 select today() + 1 = tomorrow();
 select tomorrow() - today() = 1;
+
+select toYYYYMM(toDateTime(1630320462));
+select toYYYYMM(today() - 30) = toYYYYMM(today()) - 1;
```
sundy-li, line 21: What if today is the 2021-10-31?

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -0,0 +1,135 @@
...
+
+use crate::scalars::Function;
+
+#[derive(Clone, Debug)]
+pub struct NumberFunction<T> {
+    display_name: String,
+    t: PhantomData<T>,
+}
+
+pub trait NumberResultFunction {
```
sundy-li, line 35: Great, we can even make it more generic using `NumberResultFunction<T> `, this can be addressed in another pr to support ` toYYYYMMDDhhmmss`, which returns UInt64.

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    fn num_arguments(&self) -> usize {
+        1
+    }
+
+    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+        Ok(false)
+    }
+
+    fn eval(&self, _columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
```
BohuTANG, line 78: _columns -> columns

common/functions/src/scalars/dates/number_function.rs
ygf11: I will address this in next pr.

tests/suites/0_stateless/02_0012_function_datetimes.sql
```diff
@@ -18,7 +18,7 @@ select today() + 1 = tomorrow();
 select tomorrow() - today() = 1;
 
 select toYYYYMM(toDateTime(1630320462));
-select toYYYYMM(today() - 30) = toYYYYMM(today()) - 1;
-select toYYYYMM(today() + 30) = toYYYYMM(today()) + 1;
-select toYYYYMM(today()) - toYYYYMM(today() - 30) = 1;
-select toYYYYMM(today() + 30) - toYYYYMM(today()) = 1;
+select toYYYYMM(today() - 31) = toYYYYMM(today()) - 1;
```
sundy-li, line 21: That's not correct too. What if today is `03-01`,  `toYYYYMM(today() - 31)` will be `toYYYYMM(today()) - 2`

ygf11: Yeah, you are right.
For the case of 03-01, all the four eq expression may return false.
As datafuse does not support inserting date16/date32/datetime32 sql like following now.
````
mysql> CREATE TABLE test(a UInt8, b date) engine = Memory;

mysql> desc test;
+-------+--------+------+
| Field | Type   | Null |
+-------+--------+------+
| a     | UInt8  | NO   |
| b     | Date16 | NO   |
+-------+--------+------+
2 rows in set (0.00 sec)

mysql> insert into test(a, b) values(1, toDate(1630320462));
ERROR 1105 (HY000): Code: 10, displayText = create_deserializer does not support type 'Date16'.
mysql> insert into test(a, b) values(1, '2021-08-10');
ERROR 1105 (HY000): Code: 10, displayText = create_deserializer does not support type 'Date16'.
```
How can I add the stateless test? Please give some idea.

sundy-li: > As datafuse does not support inserting date16/date32/datetime32 sql like following now.

Yes, we will add this in another pr.
Cureently 

```
select toUInt32(now()) returns `1630833797`
select toUInt16(today()) returns `18875`
```
How about  

```
select toYYYYMM(toDateTime(1630833797)  )  =  202109;
select toYYYYMM(toDate(18875)  )  =  202109;

```

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -0,0 +1,135 @@
...
+                "Number of arguments for function toYYYYMM doesn't match: passed 0, should be 1 or 2")
+            );
+        }
+
+        let data_type = columns[0].data_type();
+        let number_vec = match data_type {
+            DataType::Date16 | DataType::Date32 => {
+                let number_vec_result: Vec<Option<u32>> = columns[0]
+                    .column()
+                    .to_values()?
```
sundy-li, line 90: `to_values` has poor performance than native array iter.


How about using `let col = columns[0].u16()?` and `let col = columns[0].u32()?`. 
Then we can have `col.iter().map(...)`

tests/suites/0_stateless/02_0012_function_datetimes.sql
ygf11: Ok.

common/functions/src/scalars/dates/number_function.rs
ygf11: Is the u16() method in SeriesTrait?

sundy-li: Yes

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn num_arguments(&self) -> usize {
+        1
+    }
+
+    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+        Ok(false)
+    }
+
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        if columns.is_empty() {
```
sundy-li, line 79: No need to check, already checked in `num_arguments`, but it's trival.

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -0,0 +1,137 @@
...
+            return Result::Err(ErrorCode::UnknownFunction(
+                "Number of arguments for function toYYYYMM doesn't match: passed 0, should be 1 or 2")
+            );
+        }
+
+        let data_type = columns[0].data_type();
+        let number_vec: Vec<Option<u32>> = match data_type {
+            DataType::Date16 => {
+                let number_vec_result = columns[0].column().to_array()?.u16()?.iter().map(|value|{
```
sundy-li, line 88: Great, but there are also much more spaces to improve this function.

1. Constant path to check, which may help in `select toYYYYMM( 123455 ) from number(xxx)`
2. Put the whole eval function into `execute` function, this can be done in `toYYYYMMDDhhmmss`, but it's return value is u64.

sundy-li (approved):
LGTM, thanks @ygf11

common/functions/src/scalars/dates/number_function.rs
ygf11: > Constant path to check, which may help in select toYYYYMM( 123455 ) from number(xxx)

It is nice to do this, but there are somethings I am not  sure:
1. If I understand correctly , ` toYYYYMM(123455)` is the same as `select toYYYYMM(toDate(123455))` or `toYYYYMM(toDateTime(123455))`, so which one to choose? toDateTime?
2. `Constant` may be  negative value , what should return for this situation?
3. After this, `select toYYYYMM(number) from number(xxx)` will also work. Is it ok?

sundy-li: >  is the same as select toYYYYMM(toDate(123455)) or toYYYYMM(toDateTime(123455))

It's related to `DataType::Date16` or `DataType::DateTime` ..

> Constant may be negative value 

Actually, if its logical type is Date16/DateTime32, which physical type is `u16` and `u32`, the constant value will never be negative.

> toYYYYMM(number)  will never work, we just need to make toYYYYMM(Date/DateTime) work, the inner argument is physically constant.

ygf11: > It's related to DataType::Date16 or DataType::DateTime .

How do I make sure it is a DateType::Date16 or a DataType::DateTime when only giving `select toYYYYMM(123455)`?

sundy-li: Let's use 

```
let c = columns[0].column().to_array()?.u16()?;
c.apply_cast_numeric(..)
```
instead, it take consideration of the null values.

ygf11: >Let's use
....
instead, it takes consideration of the null values.

Got it.

> Constant path to check, which may help in select toYYYYMM( 123455 ) from number(xxx)

As `DataColumn.to_array()` has coverage the `Constant` case as Series, what did you mean  `Constant path to check`?

sundy-li: Constant path can have better performance, you can simply use to_array(). 
But it will build an array with the same number which has poor performance because you need to calculate each row.

ygf11: Thanks, got it.

Issue #883, opened 2021-06-23
[FR] cluster management for FuseQuery


zhang2014:
**Summary**

FuseQuery is a stateless service. It should be very easy to scale in and scale out. for example:
``` shell
# When I want to start FuseQuery with standalone mode:
: ) ./fuse_query --store fuse_store_host_name:9191
# Or
: ) ./fuse_query --store 192.168.0.1:9191;192.168.0.2:9191


# When I want to start FuseQuery with cluster mode:
: ) ./fuse_query --store fuse_store_host_name:9191 --namespace cluster_1
# Or
: ) ./fuse_query --store 192.168.0.1:9191;192.168.0.2:9191 --namespace cluster_1

# When I want to start FuseQuery and join an existing cluster
: ) ./fuse_query --store fuse_store_host_name:9191 --namespace exists_cluster_name
# Or
: ) ./fuse_query --store 192.168.0.1:9191;192.168.0.2:9191 --namespace exists_cluster_name
```

In this case, we need FuseStore as registry center. we may need the following APIs:
- Register node(namespace, node_name, ip, flight_port, CPUs, Memory, heartbeat_interval)
- Heartbeat(namespace, node_name)
- GetRegisteredNodes

CC: @dantengsky @BohuTANG

BohuTANG:
Very cool.

Now store:
1. storage
2. metadata
3. KV API for states

For cluster node register, 2 questions:
Q1: Only KV API is ok?
Q2: Need subscriber and notify?

zhang2014:
> Q1: Only KV API is ok?

It's ok, but we need expire_key api.

> Q2: Need subscriber and notify?

It's not necessary. we can fetch it once every query. but it's better to provide `subscriber` and `notify`

dantengsky:
sorry to cut in, I have a strong feeling that in the long run, we need another layer like snowflake's [cloudservice](https://docs.snowflake.com/en/user-guide/intro-key-concepts.html#snowflake-architecture), and separates `meta` `authentication/authorization` `infra mgt` from `store` layer.

BohuTANG:
`cloud service` make sense.
`cloudservice` is a builtin or builton of the store :)

dantengsky:
[...]

PR #1731, opened 2021-09-05
 ISSUE-883 continue cluster management with namespace

zhang2014:

## Summary

Support cluster management with namespace for databend query

## Changelog

- New Feature

## Related Issues
part #883


query/src/clusters/cluster.rs
```diff
@@ -12,140 +12,233 @@
...
-    pub fn create_global(cfg: Config) -> Result<ClusterRef> {
-        Ok(Arc::new(Cluster {
-            nodes: Mutex::new(HashMap::new()),
-            local_port: Address::create(&cfg.query.flight_api_address)?.port(),
+impl ClusterDiscovery {
+    // TODO(Winter): this should be disabled by compile flag
+    async fn standalone_without_metastore(cfg: &Config) -> Result<ClusterDiscoveryRef> {
+        let local_id = global_unique_id();
+
+        let local_store = LocalKVStore::new_temp().await?;
```
BohuTANG, line 47: This api will change from `async` to `sync`.
I think all the meta api all `sync` is ok.
cc @dantengsky @zhang2014

zhang2014: got it. by the way I think `async` is OK if it is not `&mut self`

common/management/src/namespace/namespace_mgr.rs
```diff
@@ -32,109 +34,208 @@ pub static NAMESPACE_API_KEY_PREFIX: &str = "__fd_namespaces";
...
+        kv_api: Arc<dyn KVApi>,
+        tenant: &str,
+        namespace: &str,
+        lift_time: Duration,
+    ) -> Result<Self> {
+        Ok(NamespaceMgr {
+            kv_api,
+            lift_time,
+            namespace_prefix: format!(
+                "{}/{}/{}/databend_query",
```
BohuTANG, line 52: For the prefix: `"{}/{}/{}/databend_query"`,  the suffix `databend_query` is used for?

PR #1738, opened 2021-09-06
[Feature]Add config table

### LINKED ISSUE #1630: [feature] system.configs table
**Summary**

Add system.configs table to show the configs:
columns:
`name | value | group | description`

```
 select * from system.configs;

┌─name──────────────────┬─value─────┬─group─────────┬─description──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ num_cpus              │ 10        │ Query         │ Maximum thread to use for the query                                                                                                  │
└───────────────────────┴───────────┴───────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

Reference:
system.settings table:
https://github.com/datafuselabs/datafuse/blob/master/query/src/datasources/system/settings_table.rs

GrapeBaBa:

## Summary

Add config table

## Changelog

- New Feature

## Related Issues

Fixes #1630 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/datasources/system/configs_table_test.rs
+++ b/query/src/datasources/system/configs_table_test.rs
@@ -19,12 +19,20 @@ use futures::TryStreamExt;
 use pretty_assertions::assert_eq;
 
 use crate::catalogs::Table;
+use crate::clusters::Cluster;
+use crate::configs::Config;
 use crate::datasources::system::configs_table::ConfigsTable;
+use crate::sessions::SessionManager;
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 async fn test_configs_table() -> Result<()> {
-    let ctx = crate::tests::try_create_context()?;
-    ctx.get_settings().set_max_threads(2)?;
+    let config = Config::default();
+    let cluster = Cluster::empty();
+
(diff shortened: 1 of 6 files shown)


query/src/datasources/system/configs_table_test.rs
```diff
@@ -0,0 +1,88 @@
...
+        ctx.get_settings().get_max_threads()? as usize,
+    )?;
+
+    let stream = table.read(ctx, &source_plan).await?;
+    let result = stream.try_collect::<Vec<_>>().await?;
+    let block = &result[0];
+    assert_eq!(block.num_columns(), 4);
+    assert_eq!(block.num_rows(), 30);
+
+    let expected = vec![
```
sundy-li, line 50: What if we add one config, the tests will be failed. So I suggest just to test some common values?

GrapeBaBa: any method can be invoked to get rows on result?

BohuTANG: Like some column names :)

GrapeBaBa:
@BohuTANG @sundy-li I will change the test later when I am available.

PR #1740, opened 2021-09-06
chore: fix ArrayAgg error message

### LINKED ISSUE #1741: The document in `ArrayAgg` is not same with its impls.
**Summary**

For Example `sum`:

```rust
    /// Aggregate the sum of the ChunkedArray.
    /// Returns `DataValue::Null` if the array is empty or only contains null values.
    fn sum(&self) -> Result<DataValue> {
        Err(ErrorCode::BadDataValueType(format!(
            "Sum operation not supported for {:?}",
            self,
        )))
    }
```

But `sum` in `DFBooleanArray` return Boolean if `is_empty` or `all_is_null`
```rust
impl ArrayAgg for DFBooleanArray {
    fn sum(&self) -> Result<DataValue> {
        if self.all_is_null() {
            return Ok(DataValue::Boolean(None));
        }
        let sum = self.into_iter().fold(0, |acc: u64, x| match x {
            Some(v) => acc + v as u64,
            None => acc,
        });

        Ok(sum.into())
    }
...
}
```

But for `sum` I think it is more reasonable return `0`. 

The document descriptions of other aggregate functions have similar problems
I think the description of most documents are more reasonable than the implementations.

yjhmelody:

## Summary

Summary about this PR

## Changelog

- Improvement: Fix aggregation error message

## Related Issues

fixes #1741

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/arrays/ops/agg.rs
+++ b/common/datavalues/src/arrays/ops/agg.rs
@@ -35,14 +35,14 @@ pub trait ArrayAgg: Debug {
     /// Returns `DataValue::Null` if the array is empty or only contains null values.
     fn sum(&self) -> Result<DataValue> {
         Err(ErrorCode::BadDataValueType(format!(
-            "Unsupported aggregate operation: sum for {:?}",
+            "Sum operation not supported for {:?}",
             self,
         )))
     }
 
     fn min(&self) -> Result<DataValue> {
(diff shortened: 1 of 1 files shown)


sundy-li (approved):
LGTM

PR #1752, opened 2021-09-07
Add Interval suppport

junli1026:

## Summary

Add toInterval support. The usage can be like "select interval '1' day". with the pattern of "Interval '[num]' [unit]". 

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Manual tested with mysql and clickhouse client. Test sql includes:
   "select interval '1' year" 
   "select now() + interval '1' year" // this will hit to arithmetic, which is not implemented yet.


common/datavalues/src/types/data_type.rs
```diff
@@ -75,6 +78,7 @@ impl DataType {
                 ArrowDataType::Struct(arrows_fields)
             }
             String => ArrowDataType::LargeBinary,
+            Interval(_) => ArrowDataType::Float64,
```
sundy-li, line 81: Why interval is Float64?

junli1026: Not necessary, it can be float32 because I hope it can support something like "interval '1.5' year".
And I don;t think this line will be executed.

sundy-li: But I think UInt64 will cover if we store intervals as NanoSeconds.

sundy-li: ClickHouse seems to store int64, because negative interval needed.

junli1026: Sure. No problem. how about Int64 store seconds number ?

zhang2014: Why not use arrow Interval type?

sundy-li: Seconds is ok,   "interval '1.5' year" we can cast `1.5 * Interval` --> `Float64` --> `Int64 interval`

common/datavalues/src/types/data_type.rs
```diff
@@ -125,3 +129,29 @@ impl fmt::Display for DataType {
...
+    Year,
+    Month,
+    Day,
+    Hour,
+    Minute,
+    Second,
+}
+
+impl IntervalUnit {
+    pub fn avg_seconds(unit: IntervalUnit) -> i32 {
```
zhang2014, line 146: For DateTime, seconds is ok, but maybe we should consider the following case:
```
SELECT
    toDate('2020-02-28') - toIntervalMonth(1) AS a,
    toDate('2020-03-28') - toIntervalMonth(1) AS b,
    toDate('2020-03-31') - toIntervalMonth(1) AS c

Query id: 4ae748f4-8914-48b7-80d7-1c54fe9cf1c8

┌──────────a─┬──────────b─┬──────────c─┐
│ 2020-01-28 │ 2020-02-28 │ 2020-02-29 │
└────────────┴────────────┴────────────┘
```

Maybe we should save the interval units and values

common/datavalues/src/types/data_type.rs
junli1026: > Why not use arrow Interval type?

The idea is from clickhouse, they don't have Interval as a built-in column type, but more like a syntax sugar. So the interval data won't be stored into parquet. -- This is my naive understanding, correct me if I am wrong.

common/datavalues/src/types/data_type.rs
junli1026: I just tried, the query "select toDate('2020-02-28')" gives me this:

mysql> select toDate('2020-02-28');
+----------------------+
| toDate('2020-02-28') |
+----------------------+
| NULL                 |
+----------------------+
1 row in set (0.01 sec)



datafuse :) select toDate('2020-02-28');

┌─toDate('2020-02-28')─┐
│           1970-01-01 │
└──────────────────────┘


We might need an extra issue to address the toDate before this one.

common/datavalues/src/types/data_type.rs
zhang2014: 😂 In ClickHouse, the interval may be complex:

There may be two ClickHouse interval types(DataTypeNumber and DataTypeInterval). For their functions processing, in https://github.com/ClickHouse/ClickHouse/blob/f559c3411392901da6d5130711e1637fb22d2cbd/src/Functions/FunctionBinaryArithmetic.h#L599, but we may be better and more elegant than ClickHouse (I'm not sure)

junli1026: > Why not use arrow Interval type?

question: will datefuse support Interval type as sql data type, Like "CREATE TABLE t (c1 interval);"   If yes, I think we should use the arrow interval type for maybe storage. Notice that arrow also support Date32 and Date64, but we we using uint32 for Date.

common/datavalues/src/types/data_type.rs
sundy-li: > We might need an extra issue to address the toDate before this one.

It's wip.

>  Notice that arrow also support Date32 and Date64, but we are using uint32 for Date.
`Date` in Datafuse/ClickHouse is u16 now, it's enough in our life.  Datafuse used our logical types other  than arrow's.
We must have `DataTypeInterval`, but it's a logical datatype,  seems arrow's interval type can meet our demand, I vote for using arrow's DataTypeInterval.

Take interval just as seconds is not enough, we must have the interval unit.

Eg:

`03-31  interval  -1 month`   ===> `02-29`  rather than `03-31` sub `30` days.

jorgecarleitao:
some extra info:

In arrow there are 3 interval formats, all of which arrow2 supports: Interval in months (`i32`), interval in (days,milliseconds) (`[i32,i32]`) and interval in (month,day,nanoseconds) `[i32,i32,i64]`. These are defined because math in time-aware timezones is often not sufficient in making everything seconds, because a day is not necessarily 24h*3600 sec/h due to leap seconds and others.

There is also the "duration" with 4 units (sec, milli, micro, nano) stored as `i64` which is not time-aware.

junli1026:
> some extra info:
> 
> In arrow there are 3 interval formats, all of which arrow2 supports: Interval in months (`i32`), interval in (days,milliseconds) (`[i32,i32]`) and interval in (month,day,nanoseconds) `[i32,i32,i64]`. These are defined because math in time-aware timezones is often not sufficient in making everything seconds, because a day is not necessarily 24h*3600 sec/h due to leap seconds and others.
> 
> There is also the "duration" with 4 units (sec, milli, micro, nano) stored as `i64` which is not time-aware.

Thanks Jorge, is the Interval(Year) in arrows based on real life calander ? For example the one @sundy-li  mentioned:
```
"03-31 interval -1 month ===> 02-29 rather than 03-31 sub 30 days." 
```

jorgecarleitao:
> Thanks Jorge, is the Interval(Year) in arrows based on real life calander ? For example the one @sundy-li mentioned:
> 
> ```
> "03-31 interval -1 month ===> 02-29 rather than 03-31 sub 30 days." 
> ```

Exactly, `Interval(YearMonth)` is an `i32` representing months. Arrow2 will support ops over intervals; we recently added support to read time-aware timestamps, the next step is the compute :)

common/datavalues/src/types/data_type.rs
junli1026: HI @sundy-li , correct me if I am wrong. For the arithmetic PLUS operation of (DateTime + Interval),  since we don't have DataValue mapping to arrow's datetime, the underneath operation would be uint32, right ? So let's say if I map our Interval to arrow's interval,  we still can't leverage arrow's arithmetic operation, unless we change the mapping. Is my understanding correct ?

sundy-li: Yes, but we can have our own `DateTime <op> Interval`. 

If we find  it's logical  `DateTime <op> Interval`, we can go to another specific path rather than the `u32 <op> i64`.

something like:
```
let datetime_col = col1.u32()?;
let interval_col = col2.i64()?;

let result = datetime_col.op_interval(&interval_col, time_unit, time_zone ...)?;
```

sundy-li: Yes, you are right, we can't benefit if introducing array::interval in computing, we must write our own computing.
That's reasonable.


And there are many computing functions we should write our own. Such as `BinaryArray.like()`, if introducing charsets, the property of this method may like `BinaryArray.like<CharSet>()`.    
As someone suggested in tikv's like_impl: https://github.com/tikv/tikv/blob/master/components/tidb_query_expr/src/impl_like.rs

junli1026: @sundy-li, something I want to raise is, arrow interval::DateTime use struct([i32; 2]) as data value, while we don't have it in datavalue.  It is pretty hard to use arrow interval here unless we can also add a new data value.
Another option is like what you did,  use arrow::extension wrapping with ArrowDataType::Int32 for our interval.

common/datavalues/src/types/data_type.rs
```diff
@@ -80,6 +100,11 @@ impl DataType {
                 ArrowDataType::Struct(arrows_fields)
             }
             String => ArrowDataType::LargeBinary,
+            Interval(unit) => ArrowDataType::Extension(
```
sundy-li, line 103: let's just use `ArrowDataType::Int64`, extension logical type currently is not allowed in parquet write.

common/functions/src/scalars/arithmetics/arithmetic.rs
```diff
@@ -74,14 +78,21 @@ impl Function for ArithmeticFunction {
     }
 
     fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        let has_date_or_date_time = columns.iter().any(|c| is_date_or_date_time(c.data_type()));
+        let has_interval = columns.iter().any(|c| is_interval(c.data_type()));
+
+        if has_date_or_date_time && has_interval {
+            //TODO(Jun): implement data/datatime +/- interval
+            unimplemented!()
```
zhang2014, line 86: maybe return Err(ErrorCode::UnImplement) is better?

common/datavalues/src/types/data_type.rs
jorgecarleitao: We can add that as an issue in arrow2: we work on that there.

sundy-li: Thanks, already addressed in https://github.com/jorgecarleitao/arrow2/issues/392

sundy-li (approved):
LGTM, we can add arithmetic ops for the interval type in another pr.

Issue #1737, opened 2021-09-05
[feature][test] local KVApi provider for test


BohuTANG:
**Summary**

When the query use common/management/namespace to do unit test, it seems hard to mock a locally KVApi provider.

BohuTANG:
😃 CC: @drmingdrmer @dantengsky

PR #1754, opened 2021-09-07
[common/management] feature: add LocalKVStore: a KVApi impl backed with a sled tree, for testing purpose mainly

drmingdrmer:

##### [common/management] feature: add LocalKVStore: a KVApi impl backed with local store, for testing purpose mainly

For use in unitest:
1. initialize sled::Db in a temp dir.
1. create a LocalKVStore with unique name for every test case:

```rust
fn test_xxx() {
    let t = tempfile::tempdir().expect("create temp dir to sled db");
    init_temp_sled_db(t);

    let kv1 = LocalKVStore::new_temp();
    let kv2 = LocalKVStore::new_temp();
}
```

For use in production:
```rust
init_sled_db("sled-db-dir");

let kv1 = LocalKVStore::new("name-of-the-underlying-sled-tree");
let kv2 = LocalKVStore::new("another-name-of-the-underlying-sled-tree");
```

Caveat: Two LocalKVStore instances with the same `name` reference to the same underlying sled::Tree.

## Changelog

- New Feature

## Related Issues

fix: #1737

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -882,10 +882,14 @@ dependencies = [
  "common-metatypes",
  "common-runtime",
  "common-store-api",
+ "common-tracing",
+ "datafuse-store",
+ "lazy_static",
  "mockall",
  "serde",
  "serde_json",
  "sha2",
+ "tempfile",
 ]
 
 [[package]]
--- a/common/management/Cargo.toml
+++ b/common/management/Cargo.toml
@@ -14,8 +14,13 @@ edition = "2021"
 common-exception= {path = "../exception"}
 common-metatypes= {path = "../metatypes"}
 common-store-api= {path = "../store-api"}
+common-tracing= {path = "../tracing"}
+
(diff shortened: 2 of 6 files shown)


PsiACE:
If it is mainly for testing, maybe we can consider not using datafuse-store? and only provide it when testing

BohuTANG:
Cool.
 I think this is not only used by namespace from the query layer, others also need, such as user api(query haven't implement this part yet).

drmingdrmer:
> If it is mainly for testing, maybe we can consider not using datafuse-store? and only provide it when testing

Right, splitting datafuse-store into several components is another issue.

BohuTANG:
@zhang2014 
You will need this API in https://github.com/datafuselabs/datafuse/pull/1731 and would like to hear your thoughts.

zhang2014:
Looks great to me.

zhang2014 (approved):

PR #1809, opened 2021-09-13
Add arithmetic support for interval type

junli1026:

## Summary

Add arithemetic support for interval type
>      Date16  [+|-] Interval
>      Date32  [+|-] Interval
>      DateTime32 [+|-] Interval

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

### DIFF
--- a/common/functions/src/scalars/arithmetics/arithmetic.rs
+++ b/common/functions/src/scalars/arithmetics/arithmetic.rs
@@ -131,42 +131,106 @@ impl ArithmeticFunction {
         &self,
         columns: &DataColumnsWithField,
     ) -> Option<Result<DataColumn>> {
-        if columns.len() != 2 {
-            return None;
-        }
-
-        match (columns[0].data_type(), columns[1].data_type()) {
-            (&DataType::Interval(_), &DataType::Date16)
-            | (&DataType::Interval(_), &DataType::Date32)
-            | (&DataType::Interval(_), &DataType::DateTime32(_))
-            | (&DataType::Date16, &DataType::Interval(_))
-            | (&DataType::Date32, &DataType::Interval(_))
-            | (&DataType::DateTime32(_), &DataType::Interval(_)) => match self.op {
-                DataValueArithmeticOperator::Plus | DataValueArithmeticOperator::Minus => {
-                    Some(self.datetime_plus_minus_interval(columns))
-                }
-                _ => None,
-            },
-            _ => None,
+        if let Some((interval, date_datetime)) = self.check_interval_date_columns(columns) {
+            let col = match interval.data_type() {
+                DataType::Interval(IntervalUnit::YearMonth) => match date_datetime.data_type() {
+                    DataType::DateTime32(_) => {
+                        self.datetime_plus_minus_year_month(date_datetime, interval)
+                    }
+                    DataType::Date16 | DataType::Date32 => {
+                        self.date_plus_minus_year_month(date_datetime, interval)
+                    }
+                    _ => unreachable!(),
+                },
+                DataType::Interval(IntervalUnit::DayTime) => match date_datetime.data_type() {
+                    DataType::DateTime32(_) => {
+                        self.datetime_plus_minus_day_time(date_datetime, interval)
+                    }
+                    DataType::Date16 | DataType::Date32 => {
+                        self.date_plus_minus_day_time(date_datetime, interval)
+                    }
+                    _ => unreachable!(),
+                },
+                _ => unreachable!(),
+            };
+            Some(col)
+        } else {
+            None
         }
     }
 
-    pub fn datetime_plus_minus_interval(
+    pub fn check_interval_date_columns<'a>(
         &self,
-        columns: &DataColumnsWithField,
-    ) -> Result<DataColumn> {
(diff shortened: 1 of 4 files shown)


sundy-li:
Hello, this implementation looks good but seems it's not efficient.

1. `try_get` is slow inside the loop.
2. There are lots of errors try return inside the loop.
3. There are lots of `match` inside the loop.

These will cause many branch predictions which make the function work slowly.

Do you have any plan to make it faster?

If the IntervalType is DayTime, seems we can use `array.apply` to apply arithmetic op in seconds with casting it into NativeTime.


```
mach interval_type {
    DataType::IntervalType(DayTime) => {
       //date  apply to seconds and +/- seconds
       
       //date32   apply to seconds and +/- seconds
       
       //datetime32  apply to seconds and +/- seconds
    
    }


  DataType::IntervalType(YearMonth) => {
       //date apply to NativeTime and +/- year month
         
       //date32  apply to NativeTime and +/- year month
       
       //datetime32  apply to NativeTime +/- plus year month
     
    }
}

```

junli1026:
> Hello, this implementation looks good but seems it's not efficient.
> 
>     1. `try_get` is slow inside the loop.
> 
>     2. There are lots of errors try return inside the loop.
> 
>     3. There are lots of `match` inside the loop.
> 
> 
> These will cause many branch predictions which make the function work slowly.
> 
> Do you have any plan to make it faster?
> 
> If the IntervalType is DayTime, seems we can use `array.apply` to apply arithmetic op in seconds with casting it into NativeTime.
> 
> ```
> mach interval_type {
>     DataType::IntervalType(DayTime) => {
>        //date  apply to seconds and +/- seconds
>        
>        //date32   apply to seconds and +/- seconds
>        
>        //datetime32  apply to seconds and +/- seconds
>     
>     }
> 
> 
>   DataType::IntervalType(YearMonth) => {
>        //date apply to NativeTime and +/- year month
>          
>        //date32  apply to NativeTime and +/- year month
>        
>        //datetime32  apply to NativeTime +/- plus year month
>      
>     }
> }
> ```

Sure, will address. A quick question, do we have SIMD support for bit operation like left shift (<<) or right shift ?  (Maybe they can be achieved by Module or Multiple, but they are slower).   Because the DayTime is a 64bit, it may help if we have SIMD for left shift or right shift.

sundy-li:
> Because the DayTime is a 64bit, it may help if we have SIMD for left shift or right shift.

We can use `array.i64().apply(|v| v << shift )`,  since the inner logic is very simple,  it'll be auto vectorized.

sundy-li:
Some comments:

1. Can we refactor the function `interval_to_day_time` to store milliseconds as i64? Thus we don't need to do any shift operation.

2. If we use  `into_nonull_iter`, we should combine the bitmap of both arrays, there is no method named `combine_validities`, or we can introduce `pub fn binary<T, D, F>` from `arrow2`.

junli1026:
>     1. Can we refactor the function `interval_to_day_time` to store milliseconds as i64? Thus we don't need to do any shift operation.

Sure. In this case we still need to to the division from milliseconds to seconds for DateTime, and from milliseconds to days for Date, right ? 

`
Current we have 
for Date
   val / 0x1_0000_0000 => days
for DateTime
   val / 0x1_0000_0000 * seconds_per_days => seconds 

After the change:
   for Date
      val / milliseconds_per_day => days
   for DateTime
      val / 1000 => seconds 
`
So that we save one multiplication operation, is that what you mean ?

sundy-li:
> So that we save one multiplication operation, is that what you mean ?

Yes, this is not the reason related to performance, but aimed to be more clear to us.

And we can add alias functions to work with arithmetic with interval, such as `addDays`, `substractMonths`.

Did you check the implementation of ClickHouse?  I think you can check it because it did a great job.

Refer:

https://github.com/ClickHouse/ClickHouse/blob/dfbb1ffec0dd38b2480c16264b409d4fef6bfa1a/src/Functions/FunctionBinaryArithmetic.h#L999-L1014

https://github.com/ClickHouse/ClickHouse/blob/master/src/Functions/FunctionBinaryArithmetic.h#L608-L657

junli1026:
> > So that we save one multiplication operation, is that what you mean ?
> 
> Yes, this is not the reason related to performance, but aimed to be more clear to us.
> 
> And we can add alias functions to work with arithmetic with interval, such as `addDays`, `substractMonths`.
> 
> Did you check the implementation of ClickHouse? I think you can check it because it did a great job.
> 
> Refer:
> 
> https://github.com/ClickHouse/ClickHouse/blob/dfbb1ffec0dd38b2480c16264b409d4fef6bfa1a/src/Functions/FunctionBinaryArithmetic.h#L999-L1014
> 
> https://github.com/ClickHouse/ClickHouse/blob/master/src/Functions/FunctionBinaryArithmetic.h#L608-L657

Thanks. maybe we can add a file named as "interval_function.rs" in the /sralars/dates folder, and reuse the function such as addDays, substractMonths both place.

junli1026:
Hi @sundy-li  I am going to add the addYear/addMonth functions in another PR. If you are fine with this one, could you approve it ? Just don't want to make the PR too big.

sundy-li (approved):
LGTM

PR #1834, opened 2021-09-15
[Store] support defaulting in store

ZhiHanZ:

## Summary

Similar to https://github.com/datafuselabs/databend/blob/master/query/src/configs/config.rs

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/store/src/configs/config.rs
+++ b/store/src/configs/config.rs
@@ -39,7 +39,9 @@ lazy_static! {
     };
 }
 
-#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, PartialEq, StructOpt, StructOptToml)]
+#[derive(
+    Clone, Debug, serde::Serialize, serde::Deserialize, PartialEq, StructOpt, StructOptToml,
+)]
 pub struct Config {
     /// Identify a config. This is only meant to make debugging easier with more than one Config involved.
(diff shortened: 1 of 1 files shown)


drmingdrmer:
Does it build something different from `Config::empty()`?

`StructOptToml` provides an impl of `trait Default`. But it should not be used in unittest. Because it reads the CLI arguments that should be passed to `cargo test`.

That's why `Config::empty()` is introduced for building a default `Config`.

BohuTANG:
Sorry, there is `empty`(haven't hear that before) there:
```
 <Self as StructOpt>::from_iter(&Vec::<&'static str>::new())
```
 @drmingdrmer Does it same as the `default`? If so query config can be use it too, it's simple and clear.

drmingdrmer:
> Sorry, there is `empty`(haven't hear that before) there:
> 
> ```
>  <Self as StructOpt>::from_iter(&Vec::<&'static str>::new())
> ```
> 
> @drmingdrmer Does it same as the `default`? If so query config can be use it too, it's simple and clear.

Yes it is meant to do the same as `::default()`.

BohuTANG:
Cool, thank you @drmingdrmer

PR #1876, opened 2021-09-18
[improvement] store client factory #1869

### LINKED ISSUE #1869: [improvement] store client factory
Part of #1855

Add a factory for create LocalKvApi(if config.meta.meta_address is empty, this is only for test) or RemoteKvAPI

dantengsky:

## Summary

fix:
- issue #1869

## Changelog

- Improvement

**Summary**

- Crate `common-fligts` has been renamed to `store-api-sdk` 
  
  - To emphasize that one can use this crate to utilize store apis.
  - A factory is provided like this:
 
  ~~~
   #[derive(Clone)]
   pub struct StoreApiProvider {
     ...
   }
  
   impl StoreApiProvider {
      pub fn new(conf: impl Into<StoreClientConf>) -> Self

      pub async fn try_get_meta_client(&self) -> Result<Arc<dyn MetaApi>> 
      pub async fn try_get_kv_client(&self) -> Result<Arc<dyn KVApi>>
      pub async fn try_get_storage_client(&self) -> Result<Arc<dyn StorageApi>>
   }
   ~~~
   
    
   since common-xxxx are not supposed to depend on crate `query::configs::Config`, we can not use `Config` directly in `store-api-sdk`.
   
   A converter form `Config` to `StoreClientConf` 
   ```
    impl From<&Config> for StoreClientConf ...
   ```
   is provide in crate `query`.
     
- The mock of KVStoreApi -- `LocalKVStore` is provided by crate `store-api-sdk`
  - `LocalKVStore` has been changed to a simple in-memory implementation
      The idea of re-use code of store-state-machine is great, but, currently, 
       1. there are circular dependencies if we place StateMachine based `LocalKVStore` into crate `store-api-sdk`
       2. also,  make `common-xxx` depend on crate `store` is not a good idea IMO
       3. let's rollback to the former implementations after the dependency issues resolved
  - The mock is provided as a part of a **non-default** feature of crate `store-api-sdk`
    crates that depend on it, like `common-management` should declare an extra dev dependency in cargo.toml, like this
    
    ```
          [dev-dependencies]
           common-store-api-sdk= {path = "../store-api-sdk", features= ["mocks"]}
    ``` 

    It  has been added to the cargo.toml of crate `common-management`

  - `management::namespace::local_kv_store_test.rs` now using `common_store_api_sdk::mocks::LocalKVStore`
     
- UserMgr Api changed

   to  runtime dispatching style:
   ~~~
    pub struct NamespaceMgr {
      kv_api: Arc<dyn KVApi>,
    }
   
    impl NamespaceMgr {
      pub fn new(kv_api: Arc<dyn KVApi>) -> Self {
        NamespaceMgr { kv_api }
      }
     ...
    }
   ~~~
   Since, the provider method signature `fn try_get_kv_client(&self) -> Result<Arc<dyn KVApi>>` indicates that the caller site does not know the concrete type of which implemented `KVApi`.  
   Previous static dispatching style of using Store Apis can also be implemented but seems to be not user-friendly (to compose the components).
   Same modification goes to `UserMgr`

- Store Apis slightly changed
  
  trait `StoreApis` are removed since we are going to split the `Store` services into multiple services.
  
  `KV/Meta/Apis` no longer take `&mut self`, since they are all wrapped in `Arc` while being passed around.
   Instead, the flight client is cloned(which is lightweight, according to [this](https://docs.rs/tonic/0.5.2/tonic/transport/struct.Channel.html#multiplexing-requests)), before calling the flights API.

## Related Issues

Fixes #1869

## Test Plan

Unit Tests

Stateless Tests


drmingdrmer:
I see the unit test failure. It is caused by RPC timeout.

Running the test `test_meta_node_add_database` takes 7 seconds on my laptop.
For CI it takes one minute and a half.

We need to find a way to deal with these time sensitive tests. 🤔

BohuTANG:
/lgtm

:+1:

common/store-api-sdk/src/mocks/local_kv_store.rs
```diff
@@ -0,0 +1,176 @@
...
+                    value: value.unwrap_or_default(),
+                };
+                let new_seq = inner.next_seq();
+                let result = (new_seq, new_v);
+                inner.kv.insert(key.to_owned(), result.clone());
+                (None, Some(result))
+            }
+        };
+        Ok(UpsertKVActionResult { prev, result })
+    }
```
drmingdrmer, line 117: A re-implementation of the meta-kv is not guaranteed to be consistent with the remote meta service.
Somebody who is not aware of there are two implementations may have changed one of them without updating the other one accordingly.
That's why a local kv should use a embedded `StateMachine`.

dantengsky: > change log:
> 
> a. there are circular dependencies if we place StateMachine based LocalKVStore into crate store-api-sdk
> b. also, make common-xxx depend on crate store is not a good idea IMO
> c. let's rollback to the former implementations after the dependency issues resolved

drmingdrmer: We both know about the reasons. A better way is to solve the dependency cycle first.

dantengsky: I do not know what exactly the deps issues are before working on this pr.

dantengsky: btw，maybe we should also try harder to NOT leak internal concepts like StateMachine to common-xxx directly

BohuTANG: > btw，maybe we should also try harder to NOT leak internal concepts like StateMachine to common-xxx directly

It seems we need a wrapper api of StateMachine?

PR #1880, opened 2021-09-18
[store] refactor: remove Config::default().

drmingdrmer:

##### [store] refactor: remove Config::default().
Since it overrides the `default()` provided by StructOptToml, there could be unexpected side effect.
Use `Config::empty()` to create a default instance without reading CLI
arguments.

## Changelog

- Improvement

## Related Issues

- #271
- #1834

### DIFF
--- a/common/functions/src/scalars/dates/date_test.rs
+++ b/common/functions/src/scalars/dates/date_test.rs
@@ -52,7 +52,7 @@ fn test_round_function() -> Result<()> {
 }
 
 #[test]
-fn test_toStartOf_function() -> Result<()> {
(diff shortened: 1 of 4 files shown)


common/functions/src/scalars/dates/date_test.rs
```diff
@@ -52,7 +52,7 @@ fn test_round_function() -> Result<()> {
 }
 
 #[test]
-fn test_toStartOf_function() -> Result<()> {
+fn test_to_start_of_function() -> Result<()> {
```
BohuTANG, line 55: :+1:

PR #1887, opened 2021-09-19
Add interval functions add/subtract months, seconds etc.

junli1026:

## Summary

Add functions of add/subtract years, months, days, hours, minutes and seconds

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Test


common/functions/src/scalars/dates/interval_function.rs
```diff
@@ -0,0 +1,727 @@
...
+        let (_y, _m, d, h, m, s) = (
+            dt.year(),
+            dt.month(),
+            dt.day(),
+            dt.hour(),
+            dt.minute(),
+            dt.second(),
+        );
+
+        // Handle month last day overflow, "2020-2-29" + "1 year" should be "2020-2-28", or "1990-1-31" + "3 month" should be "1990-4-30".
```
sundy-li, line 501:
```suggestion
        // Handle month last day overflow, "2020-2-29" + "1 year" should be "2021-2-28", or "1990-1-31" + "3 month" should be "1990-4-30".
```

common/functions/src/scalars/dates/interval_function.rs
```diff
@@ -0,0 +1,727 @@
...
+        let (interval, date) = Self::validate_input(op, a, b)?;
+        let milliseconds_per_day = 24 * 3600 * 1000;
+
+        let days: DataColumn = interval
+            .column()
+            .to_array()?
+            .i64()?
+            .apply(|ms| ms / milliseconds_per_day)
+            .into();
+        date.column().arithmetic(op.clone(), &days)
```
sundy-li, line 256: If `date` is `Date16 or Date32`, the result of `date.column().arithmetic` will be  `i64` column, and it represents days.

junli1026:
Ping :)

sundy-li:
@junli1026  
Have you read my comment message in `interval_daytime_plus_minus_date`?
 
 I think it's wrong to apply `column.arithmetic`, because `date16` minus `i64` will be `i64`.

sundy-li:
Can you add stateless-tests  SQL tests for this pr?

junli1026:
> Can you add stateless-tests SQL tests for this pr?

sure will do.

junli1026:
> @junli1026
> Have you read my comment message in `interval_daytime_plus_minus_date`?
> 
> I think it's wrong to apply `column.arithmetic`, because `date16` minus `i64` will be `i64`.

Yes, we have some logic to convert i64 to date16 after the calculation. Are you worried about the overflow ?

sundy-li:
> Yes, we have some logic to convert i64 to date16 after the calculation

OK, seems it's ok. But `result.cast_with_type(&data_type)` need extra cast operation on the array. I'd like to keep it in a single loop for better performance.

Can we make op as an argument as arrow2 did ?
```
#[inline]
pub fn binary<T, D, F>(
    lhs: &PrimitiveArray<T>,
    rhs: &PrimitiveArray<D>,
    data_type: DataType,
    op: F,
) -> Result<PrimitiveArray<T>>
where
    T: NativeType,
    D: NativeType,
    F: Fn(T, D) -> T,
```

```
Date32 op interval --> binary(lhs, rhs,  DataType::UInt32,   |a, b|   (a as u32 + b/mills   as u32)   )
Date16 op interval --> binary(lhs, rhs,  DataType::UInt16,   |a, b|   (a as u162 + b/mills    as u16)   )

```

sundy-li (approved):
Now looks great to me, thanks @junli1026

junli1026:
Will add doc in another PR soon.

PR #1891, opened 2021-09-20
ISSUE-1867: use StructOpt::from_iter to implement XXXConfig::default

### LINKED ISSUE #1867: [improvement] change the query config group default values to StructOpt default
**Summary**

In query/config, every config group has his default values, we can use structop default to do that, for example:
```
impl MetaConfig {
    pub fn default() -> Self {
        MetaConfig {
            meta_address: "".to_string(),
            meta_username: "root".to_string(),
            meta_password: "".to_string(),
            rpc_tls_meta_server_root_ca_cert: "".to_string(),
            rpc_tls_meta_service_domain_name: "localhost".to_string(),
        }
    }
}
```
->
```
impl MetaConfig {
    pub fn default() -> Self {
       <Self as StructOpt>::from_iter(&Vec::<&'static str>::new())
    }
}
```

xuyifangreeneyes:

## Summary

Summary about this PR

Implement `LogConfig`, `MetaConfig`, `StoreConfig` and `QueryConfig`'s `pub fn default() -> Self` by calling `<Self as StructOpt>::from_iter(&Vec::<&'static str>::new())` instead of constructing the struct manually. 

## Changelog

- Improvement

## Related Issues

Fixes #1867

## Test Plan

Unit Tests

### DIFF
--- a/query/src/configs/config.rs
+++ b/query/src/configs/config.rs
@@ -113,10 +113,7 @@ pub struct LogConfig {
 
 impl LogConfig {
     pub fn default() -> Self {
-        LogConfig {
-            log_level: "INFO".to_string(),
-            log_dir: "./_logs".to_string(),
-        }
+        <Self as StructOpt>::from_iter(&Vec::<&'static str>::new())
     }
 }
 
@@ -124,22 +121,22 @@ impl LogConfig {
 /// serde(default) make the toml de to default working.
 #[derive(Clone, serde::Serialize, serde::Deserialize, PartialEq, StructOpt, StructOptToml)]
 pub struct StoreConfig {
-    #[structopt(long, env = STORE_ADDRESS, default_value = "", help = "Store backend address")]
+    #[structopt(long, env = STORE_ADDRESS, default_value, help = "Store backend address")]
     #[serde(default)]
     pub store_address: String,
 
-    #[structopt(long, env = STORE_USERNAME, default_value = "", help = "Store backend user name")]
+    #[structopt(long, env = STORE_USERNAME, default_value = "root", help = "Store backend user name")]
     #[serde(default)]
     pub store_username: String,
 
-    #[structopt(long, env = STORE_PASSWORD, default_value = "", help = "Store backend user password")]
+    #[structopt(long, env = STORE_PASSWORD, default_value, help = "Store backend user password")]
     #[serde(default)]
     pub store_password: String,
 
     #[structopt(
         long,
         env = "STORE_RPC_TLS_SERVER_ROOT_CA_CERT",
-        default_value = "",
+        default_value,
         help = "Certificate for client to identify store rpc server"
     )]
     #[serde(default)]
@@ -156,13 +153,7 @@ pub struct StoreConfig {
 
 impl StoreConfig {
     pub fn default() -> Self {
-        StoreConfig {
-            store_address: "".to_string(),
-            store_username: "root".to_string(),
-            store_password: "".to_string(),
-            rpc_tls_store_server_root_ca_cert: "".to_string(),
-            rpc_tls_store_service_domain_name: "localhost".to_string(),
-        }
+        <Self as StructOpt>::from_iter(&Vec::<&'static str>::new())
     }
 }
 
@@ -180,22 +171,22 @@ impl fmt::Debug for StoreConfig {
 /// serde(default) make the toml de to default working.
 #[derive(Clone, serde::Serialize, serde::Deserialize, PartialEq, StructOpt, StructOptToml)]
 pub struct MetaConfig {
-    #[structopt(long, env = META_ADDRESS, default_value = "", help = "MetaStore backend address")]
+    #[structopt(long, env = META_ADDRESS, default_value, help = "MetaStore backend address")]
     #[serde(default)]
     pub meta_address: String,
 
-    #[structopt(long, env = META_USERNAME, default_value = "", help = "MetaStore backend user name")]
+    #[structopt(long, env = META_USERNAME, default_value = "root", help = "MetaStore backend user name")]
     #[serde(default)]
     pub meta_username: String,
 
-    #[structopt(long, env = META_PASSWORD, default_value = "", help = "MetaStore backend user password")]
+    #[structopt(long, env = META_PASSWORD, default_value, help = "MetaStore backend user password")]
     #[serde(default)]
     pub meta_password: String,
 
     #[structopt(
         long,
         env = "META_RPC_TLS_SERVER_ROOT_CA_CERT",
-        default_value = "",
+        default_value,
         help = "Certificate for client to identify meta rpc server"
     )]
     #[serde(default)]
@@ -212,13 +203,7 @@ pub struct MetaConfig {
 
 impl MetaConfig {
     pub fn default() -> Self {
-        MetaConfig {
-            meta_address: "".to_string(),
-            meta_username: "root".to_string(),
-            meta_password: "".to_string(),
-            rpc_tls_meta_server_root_ca_cert: "".to_string(),
-            rpc_tls_meta_service_domain_name: "localhost".to_string(),
-        }
+        <Self as StructOpt>::from_iter(&Vec::<&'static str>::new())
     }
 }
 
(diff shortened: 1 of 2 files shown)


query/src/configs/config.rs
```diff
@@ -113,33 +113,30 @@ pub struct LogConfig {
...
+        <Self as StructOpt>::from_iter(&Vec::<&'static str>::new())
     }
 }
 
 /// Store config group.
 /// serde(default) make the toml de to default working.
 #[derive(Clone, serde::Serialize, serde::Deserialize, PartialEq, StructOpt, StructOptToml)]
 pub struct StoreConfig {
-    #[structopt(long, env = STORE_ADDRESS, default_value = "", help = "Store backend address")]
+    #[structopt(long, env = STORE_ADDRESS, default_value, help = "Store backend address")]
```
xuyifangreeneyes, line 124: The change is due to https://docs.rs/structopt/0.3.23/structopt/index.html#default-values

query/src/configs/config.rs
```diff
@@ -113,33 +113,30 @@ pub struct LogConfig {
...
     #[serde(default)]
     pub store_address: String,
 
-    #[structopt(long, env = STORE_USERNAME, default_value = "", help = "Store backend user name")]
+    #[structopt(long, env = STORE_USERNAME, default_value = "root", help = "Store backend user name")]
     #[serde(default)]
```
xuyifangreeneyes, line 129: `structopt`'s `default_value` is `""` while we set `store_username` to `root` in `StoreConfig::default()`. Which shall we choose?

query/src/configs/config.rs
```diff
@@ -180,22 +171,22 @@ impl fmt::Debug for StoreConfig {
...
 #[derive(Clone, serde::Serialize, serde::Deserialize, PartialEq, StructOpt, StructOptToml)]
 pub struct MetaConfig {
-    #[structopt(long, env = META_ADDRESS, default_value = "", help = "MetaStore backend address")]
+    #[structopt(long, env = META_ADDRESS, default_value, help = "MetaStore backend address")]
     #[serde(default)]
     pub meta_address: String,
 
-    #[structopt(long, env = META_USERNAME, default_value = "", help = "MetaStore backend user name")]
+    #[structopt(long, env = META_USERNAME, default_value = "root", help = "MetaStore backend user name")]
     #[serde(default)]
```
xuyifangreeneyes, line 179: ditto

query/src/configs/config.rs
```diff
@@ -238,15 +223,15 @@ impl fmt::Debug for MetaConfig {
...
+    #[structopt(long, env = QUERY_TENANT, default_value, help = "Tenant id for get the information from the MetaStore")]
     #[serde(default)]
     pub tenant: String,
 
-    #[structopt(long, env = QUERY_NAMESPACE, default_value = "", help = "Namespace for construct the cluster")]
+    #[structopt(long, env = QUERY_NAMESPACE, default_value, help = "Namespace for construct the cluster")]
     #[serde(default)]
     pub namespace: String,
 
-    #[structopt(long, env = QUERY_NUM_CPUS, default_value = "0")]
```
xuyifangreeneyes, line 249: Ditto. `"0"` or `"8"`?

BohuTANG: Here we can keep it as `default_value = "0"`
But we should set it to `8` in tests/context to make the tests passed(such as explain).

query/src/configs/config.rs
BohuTANG: `default_value = "root"` is better here, show a default user for demo.

query/src/configs/config_test.rs
```diff
@@ -24,6 +24,57 @@ use crate::configs::StoreConfig;
 // Default.
 #[test]
 fn test_default_config() -> Result<()> {
+    let expect = LogConfig {
```
BohuTANG, line 27: :+1:

xuyifangreeneyes:
@BohuTANG The unit test `test_default_config` failed on ubuntu. I guess that environment variables such as `STORE_ADDRESS` can affect the return value of `StoreConfig::default()`, which leads to the failed test. For example, if `STORE_ADDRESS = "1.2.3.4:1234"`, then `StoreConfig.store_address` is `"1.2.3.4:1234"` rather than the default value `""`. Is it better if `XXXConfig::default()` is not affected by environment variables?

BohuTANG:
> @BohuTANG The unit test `test_default_config` failed on ubuntu. I guess that environment variables such as `STORE_ADDRESS` can affect the return value of `StoreConfig::default()`, which leads to the failed test. For example, if `STORE_ADDRESS = "1.2.3.4:1234"`, then `StoreConfig.store_address` is `"1.2.3.4:1234"` rather than the default value `""`. Is it better if `XXXConfig::default()` is not affected by environment variables?

I am not sure how env affected the :: default here, but it seems it affects.
We don't want env affected the default, in the config:
1. get default config
2. get the env and override the default if it set

@ZhiHanZ may help here

drmingdrmer:
> > @BohuTANG The unit test `test_default_config` failed on ubuntu. I guess that environment variables such as `STORE_ADDRESS` can affect the return value of `StoreConfig::default()`, which leads to the failed test. For example, if `STORE_ADDRESS = "1.2.3.4:1234"`, then `StoreConfig.store_address` is `"1.2.3.4:1234"` rather than the default value `""`. Is it better if `XXXConfig::default()` is not affected by environment variables?
> 
> I am not sure how env affected the :: default here, but it seems it affects in the unit test.
> We don't want env affected the default, in the config:
> 
> 1. get default config
> 2. get the env and override the default if it set
> 
> @ZhiHanZ may help here

`set_env` sets the environment variables for the current process and all unit test run  in a single process.

ZhiHanZ:
> > @BohuTANG The unit test `test_default_config` failed on ubuntu. I guess that environment variables such as `STORE_ADDRESS` can affect the return value of `StoreConfig::default()`, which leads to the failed test. For example, if `STORE_ADDRESS = "1.2.3.4:1234"`, then `StoreConfig.store_address` is `"1.2.3.4:1234"` rather than the default value `""`. Is it better if `XXXConfig::default()` is not affected by environment variables?
> 
> I am not sure how env affected the :: default here, but it seems it affects in the unit test.
> We don't want env affected the default, in the config:
> 
> 1. get default config
> 2. get the env and override the default if it set
> 
> @ZhiHanZ may help here

agreed, default config should be affected environment variables which could incur unintentional issues during deployment,  we could try the following tests in unit test:
1. cleanup env variables related to query part, and retrieve default config(left)
2. update some env variable, and retrieve default config again(right)
3. comparing left and right, in design, they should be identical

BohuTANG:
@ZhiHanZ 
Some codes suggestion?

ZhiHanZ:
> @ZhiHanZ
> Some codes suggestion?

similar to this snippet https://github.com/datafuselabs/databend/blob/master/query/src/configs/config_test.rs#L40-L98

drmingdrmer:
> > @ZhiHanZ
> > Some codes suggestion?
> 
> similar to this snippet https://github.com/datafuselabs/databend/blob/master/query/src/configs/config_test.rs#L40-L98

Since unit tests runs concurrently, `std::env::set_var` in one testing thread still has chance affecting a test running in another thread:
https://doc.rust-lang.org/nightly/std/env/fn.set_var.html
> Sets the environment variable key to the value value for the currently running process.

BohuTANG:
> > > @ZhiHanZ
> > > Some codes suggestion?
> > 
> > 
> > similar to this snippet https://github.com/datafuselabs/databend/blob/master/query/src/configs/config_test.rs#L40-L98
> 
> Since unit tests runs concurrently, `std::env::set_var` in one testing thread still has chance affecting a test running in another thread:
> https://doc.rust-lang.org/nightly/std/env/fn.set_var.html
> 
> > Sets the environment variable key to the value value for the currently running process.
@drmingdrmer 
Before this patch all test looks ok, the reason this patch caused the unit test failed is?

drmingdrmer:
> Before this patch all test looks ok, the reason this patch caused the unit test failed is?

There is a case setting up env vars, which affects `<Self as StructOpt>::from_iter(&Vec::<&'static str>::new())`.
Before this patch `StructOpt::from_iter` is not used thus there won't be a problem.

PsiACE:
I am sorry it has taken so long to respond. However, after #3499 was merged, we are now using clap v3 to replace structopt.

I will be closing this PR and the issue. Thank you for your work and let's move forward.

PR #1942, opened 2021-09-23
add toStartOfWeek

dust1:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

#853 

## Test Plan

Unit Tests ok

Stateless Tests ok


common/functions/src/scalars/dates/week_date.rs
```diff
@@ -0,0 +1,190 @@
...
+
+impl WeekResultFunction<u32> for ToStartOfWeek {
+    fn return_type() -> Result<DataType> {
+        Ok(DataType::Date16)
+    }
+    fn to_number(value: DateTime<Utc>, mode: Option<u64>) -> u32 {
+        let week_mode = mode.unwrap_or(0);
+        let mut weekday = value.weekday().number_from_monday();
+        if (week_mode & 1) == 0 {
+            weekday = value.weekday().number_from_sunday();
```
sundy-li, line 54: If week_mode is 1,should  use monday first?
```
   bool monday_first_mode = week_mode & static_cast<UInt8>(WeekModeFlag::MONDAY_FIRST);
        if (monday_first_mode)
        {
            return toFirstDayNumOfWeek(v);
        }
```

dust1: my mistake

dust1: I find clickhouse doc, if week_mode is 1,First day of week is Monday. https://clickhouse.com/docs/zh/sql-reference/functions/date-time-functions/#toweekdatemode

dust1: shouldn't I have 1 in the code? This is magic number?

sundy-li: No, just a little weird.

Default weekday should be  `number_from_sunday` because of the week_mode defaults to 0.

dust1: I understand.I will modify it right away

sundy-li (approved):

Issue #1927, opened 2021-09-22
Consider replacing `metasrv` flight API with bare grpc?


drmingdrmer:
To access meta data(purely key-value API), it does not need to follow the arrow-flight protocol, which is designed for both meta data and stream data transfer.

BohuTANG:
arrow flight is for data transfer.
For metasvr, grpc + timeout is ok(a big change).

dantengsky:
make sense

ariesdevil:
closed by #3606

Issue #1924, opened 2021-09-22
Implements database metadata management by using KV Service


dantengsky:
**Summary**

Implements database metadata management by using KV Service

depends on 
- issue #1921

BohuTANG:
Duplicated with https://github.com/datafuselabs/databend/issues/5064

PR #1952, opened 2021-09-24
ISSUE-1951: Split store API into more fine-grained crates

### LINKED ISSUE #1951: split api provider into service providers
Original disscussion:

> Some refactoring work that needs to be done is to create a standalone KVClient out of StoreClient etc. These may block future works in next couple of weeks.

does it mean that we should provide a dedicated crate, which provides `KVClient` only?

_Originally posted by @dantengsky in https://github.com/datafuselabs/databend/issues/1940#issuecomment-926277085_
____________________________

crates `store-api` `store-api-sdk` are planning to be split into more fine-grained crates (by service).

codes of crate `metasrv`, `store` may also be changed (the `use xxx` s), if there are pending large pr may conflict, please inform me, I'd like to merge later cc @drmingdrmer

dantengsky:

## Summary

Split store API / SDK into more fine-grained crates

- issue #1951 

______________

- crate `store-api` has been split into crates `common-kv-api`, `common-meta-api` and `common-dfs-api`
- `common-dfs-api` is the block transfer services in disguise, which should be migrated to real dfs client latter
- `common-meta-api`  is going to be deprecated when issue #1924 is fixed

`store-api-sdk` is left as it is, since:

- api client (not the api itself) is needed only by `query`, where all the apis are required  
    except for unit tests of individual crates, where only dev-dependencies are needed, which will not be propagated outside
- for crates like `management`,  they  only depend on `xxx-api` (not the sdk)
- kv service is going to be implemented as grpc service as issue #1927 suggested
  after that, kv client no longer depend on `StoreClient`, a grpc client will emerge naturally
- dfs service may be implemented as flight service, but it will be the only one depending on a flight client
  even if we will depend on `do_action` to impl api like `delete`,  the enums which encapsulate the actions will be constained inside the scope of dfs service (where the service side may depend on)

thus, it is not worthy of refactoring for now, as kv and dfs services are being [re-]implemented, `store-api-sdk` will be dissolved naturally.

## Changelog

- Improvement

## Related Issues

Fixes #1951

## Test Plan

Unit Tests

Stateless Tests


common/kv-apis/vo/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+use common_metatypes::SeqValue;
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct UpsertKVActionResult {
+    /// prev is the value before upsert.
+    pub prev: Option<SeqValue<KVValue>>,
+    /// result is the value after upsert.
+    pub result: Option<SeqValue<KVValue>>,
+}
+
```
drmingdrmer, line 25: What does `vo` mean indeed? 🤔

drmingdrmer:
Love this patch bro!

PR #1968, opened 2021-09-25
make raft-store deps on dfs-api rather than sdk

### LINKED ISSUE #1967: crate raft-store may depend on common-dfs-api rather than the sdk
> > common-management has to indirectly depends on store-api-sdk because of one of its dependency.
> may i ask which one? I do not have a laptop at hand now

> > common-management has to indirectly depends on store-api-sdk because of one of its dependency.

find it,  it is `kvlocal`.  
`common-management` need not depend on `kvlocal` except for `dev-dependencies`, seems you are working on it.

`raft-store` seems to be able to migrate to depend on API only, let me try it

_Originally posted by @dantengsky in https://github.com/datafuselabs/databend/issues/1940#issuecomment-927023556_

dantengsky:

## Summary

see Issue
 - #1967

## Changelog

- Improvement

## Related Issues

Fixes #1967

## Test Plan

Unit Tests

Stateless Tests


drmingdrmer:
/lgtm

PS what does `vo` mean? 🤔

dantengsky:
> /lgtm
> 
> PS what does `vo` mean?

`value object`, if i recall correctly

PR #1976, opened 2021-09-26
ISSUE-1949: Decouple ScanPlan with Table

### LINKED ISSUE #1949: Consider decouple `ScanPlan` with `Table` object
`Table` trait now depends on `ScanPlan` in `Table::read_plan`, which seems not necessary.

I found that in most of the usage of `Table::read_plan`, it will extract `table_id`, `table_version` from `ScanPlan`, while this information is more reasonable to be stored inside `Table` object.

Besides, it's unnecessary to store `ScanPlan` in `ReadDataSourcePlan` as well.

dantengsky:

## Summary

-  Trait `Table`
   - new method `Table:: database(&self) -> &str`
   - sig of method `read_plan` changed to 
  
    ```
    fn read_plan(
        &self,
        ctx: DatabendQueryContextRef,
        push_downs: Option<Extras>,
        partition_num_hint: Option<usize>,
    ) -> Result<ReadDataSourcePlan>;

    ```
- Signature of "Catalog::get_table_function" changed to

    ```
   fn get_table_function(
        &self,
        func_name: &str,
        tbl_args: Option<Expression>,
    ) -> Result<Arc<TableFunctionMeta>>
    ```
- Former tables `Number*`  changed **into** TableFunctions
    - they belong to no database, `show tables` will no longer list them
    - they are held by `DatabaseCatalog` directly
    - database name returned by them as a table, will always be an empty string

   codes that scatter around that accesses table functions by using `Catalog::get_table`, are changed to use `Catalog:get_table_function` 
   
   NOTE: there are some incoherences there:  the `get_table_function` accepts `tbl_args` as an option (which means a table function may be nullary), but while inferring whether an entity is a table or a table function, we assume there is at least one argument.  Maybe we need to polish this latter.

- Tow extra fields are added to `ReadDataSourcePlan`
    ```
     pub struct ReadDataSourcePlan {
   
      ...
       pub tbl_args: Option<Expression>,
       pub push_downs: Option<Extras>,
    }
    ```
    those are the args passed into method `Table::read_plan` and assigned to the returned `ReadDataSourcePlan`.

    The `tbl_args` field will be carried around by subsequent tasks and at some point, forward to `Catalog::get_table_function`

    The `push_downs` seems to be unnecessary for the time being, but I leave it here in case of some tables may assign different  `push_downs` to the `ReadDataSourcePlan`, in a sense of declaring that they may only support a subset of `push_downs` that is passed in via `read_plan`.

- Field `scan_plan:Arc<ScanPlan>` is NOT removed from `ReadDataSourcePlan`
  -  Since there are some plan re-writing works related to this field, I am not sure if it is proper to rm it at current stage
   - From the point of view of a `Table`,  filed `scan_plan` never used, except be assigned a value of `Default::default`

## Changelog

- Improvement

## Related Issues

Fixes #1949

## Test Plan

Unit Tests

Stateless Tests

## tasks

- [x] new Table Constructor
    which should keep `TableInfo` in the instance, so that ScanPlan can be decoupled partially
- [x] new Table::read_plan
    no longer takes scan_plan as arg, but with table_args
- [x] rm table_args from Table::read_plan
- [x] add `database` method to Table
- <del> eliminates `TableMeta`
   it is a ugly workaround, we no longer need this at current stage </del>   postponed


leiysky:
Execuse me for the unrelated comment.

Can we add an `database` method for `Table` object to expose database name of a `Table` object?

dantengsky:
> Execuse me for the unrelated comment.
> 
> Can we add an `database` method for `Table` object to expose database name of a `Table` object?

sure

BohuTANG:
Thank you @dantengsky , looks good to me.
I think @leiysky would be more interested in this patch :)

dantengsky:
>  I think @leiysky would be more interested in this patch :)
yeah. 

@leiysky pls  take your time

BohuTANG:
Why close this

dantengsky:
shoooot, closed accidentally : - (

common/planners/src/plan_read_datasource.rs
```diff
@@ -27,16 +28,21 @@ use crate::Statistics;
...
     pub table_version: Option<MetaVersion>,
     pub schema: DataSchemaRef,
+
     pub parts: Partitions,
     pub statistics: Statistics,
     pub description: String,
     pub scan_plan: Arc<ScanPlan>,
     pub remote: bool,
+
+    pub tbl_args: Option<Expression>,
```
leiysky, line 44: Maybe the type of `tbl_args` should be `Option<Vec<Expression>>`?

query/src/catalogs/table.rs
```diff
@@ -29,24 +29,27 @@ use crate::sessions::DatabendQueryContextRef;
 #[async_trait::async_trait]
 pub trait Table: Sync + Send {
     fn name(&self) -> &str;
+    fn database(&self) -> &str;
     fn engine(&self) -> &str;
     fn as_any(&self) -> &dyn Any;
     fn schema(&self) -> Result<DataSchemaRef>;
-    // Is Local or Remote.
+    fn get_id(&self) -> u64; // should we use MetaId?
```
leiysky, line 36: Indeed, I think `meta_id` is better.

query/src/catalogs/table_meta.rs
```diff
@@ -21,7 +21,9 @@ use common_metatypes::MetaVersion;
 use crate::catalogs::Table;
 use crate::catalogs::TableFunction;
 
+//#[deprecated]
```
leiysky, line 24: What's this for?

dantengsky: oh, forgot to rm this, I was planning to merge `TableMeta` into `Table`, since it provides no extra information. 
the plan has been postponed since issue #1949 do not care about it.

query/src/catalogs/table.rs
dantengsky: right, MetaId is better

common/planners/src/plan_read_datasource.rs
dantengsky: I am not sure about this, just followed the original signature. 
If list of expressions should be represented as a vec of Expression instead of a variant of `Expression`, say sth like `Expression::List`, I'd like to change the signatures

dantengsky: A subsequent "test drive" commit, which folds all the args into `TableArgs`

https://github.com/datafuselabs/databend/pull/1976/commits/4e9f8317a4f51918b0895dd0ed7654485a6f2dbc

If the last two commits are not ok, please just let me know, I 'd like to drop them from this PR, they are just experiments :D

leiysky: Seems we don't support table function with multiple arguments now.

Let's just keep it and fix it later.

leiysky (approved):
Rest LGTM

PR #1990, opened 2021-09-28
Make date32 physical type to be int32

sundy-li:

## Summary

- To be compatible with ClickHouse's [Date32](https://clickhouse.com/docs/en/sql-reference/data-types/date32/), it supports negative numbers.

- Add new aliases for SQL datatypes:
   "DATE16" => Ok(DataType::Date16),
    "DATE32" => Ok(DataType::Date32),
    "DATETIME" => Ok(DataType::DateTime32(None)),
    "DATETIME32" => Ok(DataType::DateTime32(None)),

## Changelog

related cc @ygf11  @dust1  @junli1026 

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/types/data_type.rs
+++ b/common/datavalues/src/types/data_type.rs
@@ -37,7 +37,7 @@ pub enum DataType {
     /// in days (16 bits), it's physical type is UInt16
     Date16,
     /// A 32-bit date representing the elapsed time since UNIX epoch (1970-01-01)
-    /// in days (32 bits), it's physical type is UInt32
+    /// in days (32 bits), it's physical type is Int32
     Date32,
 
     /// A 32-bit datetime representing the elapsed time since UNIX epoch (1970-01-01)
@@ -91,7 +91,7 @@ impl DataType {
             Float32 => ArrowDataType::Float32,
             Float64 => ArrowDataType::Float64,
             Date16 => ArrowDataType::UInt16,
-            Date32 => ArrowDataType::UInt32,
+            Date32 => ArrowDataType::Int32,
             // we don't use DataType::Extension because extension types are not supported in parquet
             DateTime32(_) => ArrowDataType::UInt32,
             List(dt) => ArrowDataType::LargeList(Box::new(dt.to_arrow())),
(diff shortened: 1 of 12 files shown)


zhang2014 (changes requested):
unit test failure need to fix.

ygf11:
https://github.com/datafuselabs/databend/blob/f43dcb8f8cffe437bf47f30737db6a1033f30772/common/functions/src/scalars/dates/number_function.rs#L227-L233


@sundy-li  For functions in number_function file, we need fix  constant path for date32  and add more tests :smile:?

ygf11:
https://github.com/datafuselabs/databend/blob/dddc5614e5c7bab14386e5f1ad3bd30d1f80858b/common/functions/src/scalars/dates/number_function.rs#L227-L232
@sundy-li It seem date32 still make 0 as '1970-1-1' after this pr, so for number functions we just need to fix the constant path which use as_u64 for date32?

sundy-li:
> @sundy-li It seem date32 still make 0 as '1970-1-1' after this pr, so for number functions we just need to fix the constant path which use as_u64 for date32?

This does not matter, because the numbers_function in date/datetime have specific numberic return types.

```
pub type ToYYYYMMFunction = NumberFunction<ToYYYYMM, u32>;
pub type ToYYYYMMDDFunction = NumberFunction<ToYYYYMMDD, u32>;
pub type ToYYYYMMDDhhmmssFunction = NumberFunction<ToYYYYMMDDhhmmss, u64>;

pub type ToStartOfISOYearFunction = NumberFunction<ToStartOfISOYear, u16>;
pub type ToStartOfYearFunction = NumberFunction<ToStartOfYear, u16>;
pub type ToStartOfQuarterFunction = NumberFunction<ToStartOfQuarter, u16>;
pub type ToStartOfMonthFunction = NumberFunction<ToStartOfMonth, u16>;

```

ygf11:
> This does not matter, because the numbers_function in date/datetime have specific numberic return types.

I mean that as the physical data type of date32 is `i32` now, we can using `as_i64` instead of `as_u64`  to extract value in constant path.
https://github.com/datafuselabs/databend/blob/dddc5614e5c7bab14386e5f1ad3bd30d1f80858b/common/functions/src/scalars/dates/number_function.rs#L229

PR #2033, opened 2021-09-30
[common/meta-api] Refactor: MetaApi always use `&str` as argument type instead of `String`

drmingdrmer:

## Related Issues

- fix: #2032


drmingdrmer:
I'd like to merge it now.

PR #2038, opened 2021-10-01
ISSUE-2037: add expression transform optimizer

### LINKED ISSUE #2037: [improvement] add truth transform for expression
**Summary**

Apply NOT transformation to the expression and return a new one.

Transform expression item using next rules:
```
       a AND b AND -> NOT(a) OR NOT(b) OR ...
       a OR b OR -> NOT(a) AND NOT(b) AND ...
       NOT(a) -> a
       a = b -> a != b
       a != b -> a = b
       a < b -> a >= b
       a >= b -> a < b
       a > b -> a <= b
       a <= b -> a > b
       a LIKE b -> a NOT LIKE b
       a NOT LIKE b -> a LIKE b 
       IS NULL(a) -> IS NOT NULL(a)
       IS NOT NULL(a) -> IS NULL(a)
```

zhyass:

## Summary

Add expression transform optimizer, support NOT transformer

## Changelog

- New Feature
- Improvement

## Related Issues

Fixes #2037 

## Test Plan

Unit Tests

### DIFF
--- a/query/src/optimizers/optimizer_expression_transform.rs
+++ b/query/src/optimizers/optimizer_expression_transform.rs
@@ -26,7 +26,6 @@ use common_planners::PlanNode;
 use common_planners::PlanRewriter;
 use lazy_static::lazy_static;
 
-use crate::optimizers::utils::*;
 use crate::optimizers::Optimizer;
 use crate::sessions::DatabendQueryContextRef;
 
@@ -72,7 +71,7 @@ impl ExprTransformImpl {
         let new_op = INVERSE_OPERATOR.get(op);
         match new_op {
             Some(v) => f(v, args),
-            None => create_unary_expression("NOT", vec![origin.clone()]),
+            None => Expression::create_unary_expression("NOT", vec![origin.clone()]),
         }
     }
 
@@ -103,7 +102,7 @@ impl ExprTransformImpl {
                     vec![left.as_ref().clone(), right.as_ref().clone()],
                     origin,
                     is_negated,
-                    create_binary_expression,
+                    Expression::create_binary_expression,
                 ),
             },
             Expression::UnaryExpression { op, expr } => match op.as_str() {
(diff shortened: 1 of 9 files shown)


query/src/optimizers/utils.rs
```diff
@@ -0,0 +1,63 @@
...
+                let mut v = self;
+                v.required_columns.insert(c.clone());
+                Ok(Recursion::Continue(v))
+            }
+            _ => Ok(Recursion::Continue(self)),
+        }
+    }
+}
+
+pub fn create_scalar_function(op: &str, args: Expressions) -> Expression {
```
zhang2014, line 47: Maybe move `create_scalar_function` , `create_unary_expression`,  `create_binary_expression` to Expression is better?

query/src/optimizers/optimizer_expression_transform.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        let new_op = INVERSE_OPERATOR.get(op);
+        match new_op {
+            Some(v) => f(v, args),
+            None => create_unary_expression("NOT", vec![origin.clone()]),
+        }
+    }
+
+    fn truth_transformer(origin: &Expression, is_negated: bool) -> Expression {
```
zhang2014, line 79: Maybe we can add `convert_to_negative` to Function trait.
```
trait Function {
    ...
    pub fn convert_to_negative(&self) -> Option<Self>;
    ...
}
```

query/src/optimizers/utils.rs
zhyass: Good suggestion.

query/src/optimizers/optimizer_expression_transform.rs
```diff
@@ -0,0 +1,199 @@
...
+lazy_static! {
+    static ref INVERSE_OPERATOR: HashMap<&'static str, &'static str> = {
+        let mut map = HashMap::new();
+        map.insert("=", "<>");
+        map.insert("<>", "=");
+        map.insert("<", ">=");
+        map.insert("<=", ">");
+        map.insert(">", "<=");
+        map.insert(">=", "<");
+        map.insert("LIKE", "NOT LIKE");
```
sundy-li, line 47: we should consider case insensitive

zhyass: https://github.com/datafuselabs/databend/pull/2045

PR #2047, opened 2021-10-02
Using function features for expression transform optimizer

zhang2014:

## Summary

Using function features for expression transform optimizer

## Changelog

## Related Issues

related #2038


zhyass (approved):
LGTM

PR #2056, opened 2021-10-04
Try using DAG to describe processors

zhang2014:

## Summary

Try using DAG to describe processors

## Changelog

## Related Issues

Fixes #issue


query/src/pipelines/processors/processor_dag.rs
```diff
@@ -0,0 +1,369 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{:?}",
+            Dot::with_config(&self.graph, &[Config::EdgeNoLabel])
+        )
+    }
+}
+
+// TODO(Winter): import distributed processors in DAG. We can use edge to describe them. e.g. enum EdgeAttrs { Local, Remote }
```
leiysky, line 63: Maybe this information should be encapsulated in `Processor` nodes, so we don't need to be aware of edge labels?

zhang2014: We may also need `ProcessorsDAGExecutor`. It may should control the data flow to using the properties of the graph(maybe not the properties of specific type graph node?). welcome comment : )

leiysky: > `ProcessorsDAGExecutor`

What's the difference between current executor and `ProcessorDAGExecutor`?

By the way, I'm not quite sure about the motivation of using `petgraph` to describe `Processor`. Would you mind to elaborate that for me?

BohuTANG: Is this patch intended to impl `EXPLAIN PIPELINE graph = 1` and show the graph format output?

query/src/pipelines/processors/processor_dag.rs
```diff
@@ -0,0 +1,369 @@
+use std::fmt::Debug;
```
PsiACE, line 1: add license header?

zhang2014:
move to #3380

Issue #2059, opened 2021-10-04
Decouple dependencies in catalog


drmingdrmer:
When trying to fix #2051 , some other dependency issues emerged. 
- FuseTable depends on DatabendQueryContext and DataAccesser, because it unnecessarily provides data access methods. 
  These operations should belong to something like a `TableHandler`, not the table itself.
- DataAccesser depends on DatabendQueryContext, because it needs the `context` to spawn a task to run a async job.
- ReadDataSourcePlan depends on FuseTable, because it needs to re-build a distributed ReadDataSourcePlan from a general plan.

Finally these concepts forms a dependency cycle.

What to do:
- [x] #2060
- [x] #2061
- [x] #1997
- [x] #2066
- [x] #2072
- [x] #2074
- [x] #2077 
- [x] #2079 
- [x] #2070 
- [x] #2091
- [x] #2113
- [x] #2107
- [x] #2145
- [x] #2152
- [x] #2153
- [x] #2138
- [x] #2146
- [x] #2161 
- [x] #2062

BohuTANG:
Update two issues here:
#1997 
#2070

drmingdrmer:
The goal of this issue has achieved. Closing it.

- Continue to work on #2030 .
- Some unfinished tasks are moved to #2030.

Issue #2046, opened 2021-10-02
Cyclic dependency of `TableInfo` and `ReadDataSourcePlan`


drmingdrmer:
Load a `TableInfo` by the information provided by `ReadDataSourcePlan`:

https://github.com/datafuselabs/databend/blob/3480c4467db391d1ac61b2d7a98951ce006c1640/query/src/interpreters/plan_scheduler.rs#L783-L804

Create a `ReadDataSourcePlan` from `TableInfo`:

https://github.com/datafuselabs/databend/blob/3480c4467db391d1ac61b2d7a98951ce006c1640/query/src/datasources/table/fuse/table.rs#L109-L140

BohuTANG:
In the scheduler, we have ReadDataSourcePlan already, I think we should not use db and table to get the meta.

sundy-li:
I will try to remove `read_plan` in `Table` trait and build `ReadDataSourcePlan `in query.

drmingdrmer:
Fixed with the updated API `get_table()` in #2150 
:DDD

PR #2088, opened 2021-10-06
[query] refactor: Table::read_plan does not need DatabendQueryContext anymore.

drmingdrmer:

##### [query] refactor: Table::read_plan does not need DatabendQueryContext anymore.
Why:
`Table` is a low level concept in databend code base and should be a
dependency of some other crate such as common/planner.

This commit is one of the steps to remove the `Table` dependency on `query`.

Trait `Table` references `DatabendQueryContext` as an argument type.
In this commit it is replaced with another smaller type `TableIOContext`
so that in future `Table` can be moved out of crate `query`.

`TableIOContext` provides data-access support, exposes the runtime used
by `query` itself and provides two resource value: max thread number and
node list.

- Add `TableIOContext` to provide everything a Table need to build a
  plan or else.

- `Table::read_plan()` use `TableIOContext` as argument to replace
  `DatabendQueryContext`.

- When calling `read_plan()`, a temporary `TableIOContext` is built out
  of DatabendQueryContext.

- `DatabendQueryContext` provides two additional supporting methods:
  `get_single_node_table_io_context()` and
  `get_cluster_table_io_context()`.

- fix: #2072

## Changelog

- Improvement

## Related Issues

- #2046
- #2059


PsiACE:
Looking good.

Issue #2165, opened 2021-10-10
[bug] unable to get field name after expression transform


zhyass:
**Summary**

Unable to get field name after expression transform.

**How to reproduce**

```
mysql> create table t(a int, b int) engine=Memory;
Query OK, 0 rows affected (0.86 sec)
Read 0 rows, 0 B in 0.845 sec., 0 rows/sec., 0 B/sec.

mysql> insert into t values(2,1),(null,2);
Query OK, 0 rows affected (0.02 sec)
Read 0 rows, 0 B in 0.001 sec., 0 rows/sec., 0 B/sec.

mysql> select not(a>1) from t;
ERROR 1105 (HY000): Code: 6, displayText = Unable to get field named "(NOT (a > 1))". Valid fields: ["a", "b", "(a <= 1)"].
```

**Error message and/or stacktrace**

```
[2021-10-10T00:45:44Z ERROR databend_query::servers::mysql::writers::query_result_writer] OnQuery Error: Code: 6, displayText = Unable to get field named "(NOT (a > 1))". Valid fields: ["a", "b", "(a <= 1)"].
    
       0: backtrace::backtrace::libunwind::trace
                 at /Users/nuan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.61/src/backtrace/libunwind.rs:90:5
          backtrace::backtrace::trace_unsynchronized
                 at /Users/nuan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.61/src/backtrace/mod.rs:66:5
          backtrace::backtrace::trace
                 at /Users/nuan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.61/src/backtrace/mod.rs:53:14
       1: backtrace::capture::Backtrace::create
                 at /Users/nuan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.61/src/capture.rs:176:9
          backtrace::capture::Backtrace::new
                 at /Users/nuan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.61/src/capture.rs:140:22
       2: common_exception::exception::ErrorCode::BadArguments
                 at common/exception/src/exception.rs:110:77
       3: common_datavalues::data_schema::DataSchema::index_of
                 at common/datavalues/src/data_schema.rs:83:13
       4: common_datavalues::data_schema::DataSchema::field_with_name
[...]

PR #2177, opened 2021-10-10
ISSUE-2176: support remove constant conditions

### LINKED ISSUE #2176: [Improvement] Support remove constant conditions
**Summary**

If an AND condition has `cond AND FALSE`, then the entire condition is collapsed and replaced with `ALWAYS FALSE`. 

Similarly, if an OR condition has `cond OR TRUE`, then the entire condition is replaced with `ALWAYS TRUE`. 

Else only the constant condition is removed.

zhyass:

## Summary

If an AND condition has `cond AND FALSE`, then the entire condition is collapsed and replaced with `ALWAYS FALSE`. 

Similarly, if an OR condition has `cond OR TRUE`, then the entire condition is replaced with `ALWAYS TRUE`. 

Else only the constant condition is removed.

## Changelog

- Improvement

## Related Issues

Fixes #2176 

## Test Plan

Unit Tests

### DIFF
--- a/query/src/optimizers/optimizer_constant_folding.rs
+++ b/query/src/optimizers/optimizer_constant_folding.rs
@@ -105,6 +105,113 @@ impl ConstantFoldingImpl {
             data_type,
         })
     }
+
+    fn remove_const_cond(
+        &mut self,
+        schema: &DataSchemaRef,
+        column_name: String,
+        left: &Expression,
+        right: &Expression,
+        is_and: bool,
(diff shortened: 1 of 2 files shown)


zhyass:
Will trigger bug #2165 
```
mysql> select true and number>1 from numbers(10);
ERROR 1105 (HY000): Code: 6, displayText = Unable to get field named "(true AND (number > 1))". Valid fields: ["number", "(number > 1)"].
```

Issue #2030, opened 2021-09-30
meta service refactor summary


drmingdrmer:
Goal: 
- impl meta api in `metasrv`, make it a complete meta data store service.
- `common/meta/local-store` which is a wrapper of raft-state-machine, should provide the same `MetaApi` and should be used directly as a local meta store.

### Cleanup API definition
  - [x] #2017 
  - [x] #2032
  - [x] #2039
  - [x] #2040
  - [x] #2182 
  - [x] #2184
  - [x] #2390
### Complete unit test:
  - [x] #2391
### Refactor State Machine
  - [x] #2202
### Add `MetaApi` support to StateMachine:
  - [x] #2028 
  - [x] #1919
  - [x] #2244
  - [x] #2246
  - [x] #2029
  - [x] #2508
  - [x] #2520
  - [x] #2523
### Upgrade `Catalog` to adapt the latest MetaApi impl:
  - [x] #1894 
  - [x] #2278
  - [x] #2304
  - [x] #2305
  - [x] #2310
  - [x] #2319
  - [x] #2321
  - [x] #2325
  - [x] #2326
  - [x] #2396
  - [x] #2392
  - [x] #2284
  - [x] #1895 
  - [x] #2591
### Refactor callers
  - [x] Rename `StoreClient` to `MetaClient`.
### Cleanup 
  - [x] #2016 
### Non critical tasks
  - [x] #2081 
  - [x] #2082 
  - [x] #1894 
  - [x] #2180 
  - [x] #1904 
  - [x] #2205
  - [x] #2364
  - [x] #2367
  - [x] #2031

drmingdrmer:
@BohuTANG @ariesdevil @dantengsky

drmingdrmer:
Closing.

- Other unfinished issues are moved to #2306

BohuTANG:
Great job
Give a big :rocket:

PR #2186, opened 2021-10-11
[common/base] feature: BlockingWait turns async func to sync: `async_fn().wait()` instead of `async_fn().await`

drmingdrmer:

##### [common/base] feature: BlockingWait turns async func to sync: `async_fn().wait()` instead of `async_fn().await`
Trait `BlockingWait` turns an `async` function into `sync`.

It is meant to provide convenience for building a proof-of-concept demo or else.
Always avoid using it in a real world production,
unless **you KNOW what you are doing**:

- `wait()` runs the future in current thread and **blocks** current thread until the future is finished.
- `wait_in(rt)` runs the future in the specified runtime, and **blocks** current thread until the future is finished.

```rust
use runtime::BlockingWait;
fn main() {
    async fn five() -> u8 { 5 }
    assert_eq!(5, five().wait());
}
```

## Changelog

- New Feature

## Related Issues

- #2030

### DIFF
--- a/common/base/src/lib.rs
+++ b/common/base/src/lib.rs
@@ -24,22 +24,21 @@ mod stoppable_test;
 mod profiling;
 mod progress;
 mod runtime;
+mod stop_handle;
+mod stoppable;
+mod uniq_id;
 
 pub use profiling::Profiling;
 pub use progress::Progress;
 pub use progress::ProgressCallback;
 pub use progress::ProgressValues;
+pub use runtime::BlockingWait;
 pub use runtime::Dropper;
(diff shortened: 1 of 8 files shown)


BohuTANG:
Nice shot man!
How about add a new function:
`wait_timeout(Duration)`?

dantengsky:
\lgtm

BohuTANG:
> \lgtm

It should be /lgtm :)

dantengsky:
> > \lgtm
> 
> It should be /lgtm :)

yeah :D

PR #2192, opened 2021-10-11
[CLI] generate query configs for Local profile

ZhiHanZ:

## Summary

1. define CLI api interface for query instances
2. generate query configurations according to apis
prior step for cluster runtime provision
## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
MacOS unit test failed at:
```
---- cmds::clusters::create_test::test_generate_local_query_config stdout ----
thread 'cmds::clusters::create_test::test_generate_local_query_config' panicked at 'assertion failed: `(left == right)`
  left: `"/private/tmp"`,
 right: `"/tmp"`', cli/src/cmds/clusters/create_test.rs:283:9
```

PR #2206, opened 2021-10-12
Add get_table_info(&self) for Table trait

### LINKED ISSUE #2203: Add `get_table_info(&self)` for `Table` trait
**Summary**

Currently `to_table_info` needs `db` as an argument, but `Table`  should have a determinate DB name.

sundy-li:

## Summary

- Add get_table_info(&self) for Table trait.
- Make `schema()` return none error result.
- Uniform name to `table_info`
- Make `is_local` out of `table_info`
- Make functions below as default:
  - name()
  - get_id()
  - schema()

Summary about this PR

## Changelog
 
- Improvement
 

## Related Issues

Fixes #2203

## Test Plan

Unit Tests

Stateless Tests


drmingdrmer:
/lgtm

Clean and neat! 


> @drmingdrmer How about making `is_local` out of table_info and let it declare in each table?

I did not see any reason not to do so.


> Since memory_table is constructed by table_info with local = false, but its local must be true.

BTW, why can not a MemoryTable be initialized with `is_local=true`? 🤔

PR #2300, opened 2021-10-18
chore: remove ref count

yjhmelody:

## Summary

I think the reference count here is redundant, so I tried to modify it like this.

## Changelog

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


yjhmelody:
I don’t know, I just read this piece of code casually

BohuTANG:
Reverted this PR from the main branch.

PR #2333, opened 2021-10-19
[Observability] fix query metrics exporter issue in prometheus

ZhiHanZ:

## Summary
after this fix patch, metrics would be shown in grafana metrics, we should add descriptions in metrics to let people understand the meaning of different metrics
image

## Changelog

- Bug Fix

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
/lgtm

Excellent!!

PR #2336, opened 2021-10-19
[cli] move view table command from sync to async

### LINKED ISSUE #2242: [CLI] move view table command from sync to async
**Summary**
the following command  would check health status for a databend cluster(currently support to be managed on local environment(
```
bendctl cluster view
```
for now, the table generation function run in synchronous way, which would probe on cluster health endpoint with several retries, and it may block result for a lot of time if the cluster has lots of unhealthy endpoints.

### LINKED ISSUE #2358: Rename file `cli/src/bin/benctl.rs` to `cli/src/bin/bendctl.rs`
**Summary**

Description for this improvement.

yufan022:

## Summary

using tokio to make `build_row` async

## Changelog

- Improvement

## Related Issues

Fixes #2242
Fixes #2358 

## Test Plan

Unit Tests

### DIFF
--- a/cli/src/cmds/clusters/view_test.rs
+++ b/cli/src/cmds/clusters/view_test.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::time::Duration;
-use std::time::Instant;
 
 use comfy_table::Cell;
 use comfy_table::Color;


yufan022:
@BohuTANG I noticed that the `benctl.rs` name seems wrong, is there any special reason for this?

PR #2379, opened 2021-10-21
[query] A simple rest api (http handler) for query.

### LINKED ISSUE #2241: HTTP API for query
**Summary**

Support query with sql over http API like [clickhouse HTTP Interface](https://clickhouse.com/docs/en/interfaces/http/#http-interface).

```
curl 'http://localhost:8123/?query=SELECT%201'

echo 'SELECT 1' | curl 'http://localhost:8123/?query=' --data-binary @-

echo -ne '10\n11\n12\n' | curl 'http://localhost:8123/?query=INSERT%20INTO%20t%20FORMAT%20TabSeparated' --data-binary @
```

# why we need it
        
-  easy to use with browser and curl  
-  easy  for partners and clients to integrate
-  use in environments where SQL is not available  
    - an example   [cockroachdb-serverless-and-restful-api-over-http](https://forum.cockroachlabs.com/t/cockroachdb-serverless-and-restful-api-over-http/4256/9)

more deal coming soon，after discussing face to face

#  what is NOT discussed here

- Api for management: get sys info and change settings.
- rest query api without sql:  
  - /warehouse/database/table/row can map to Resource hierarchy, especially for simple OLTP queries.
  -  some standards,  e.g. [Odata](https://www.odata.org/),  [postRest](https://github.com/PostgREST/postgrest)
    - [Why not OData?](https://docs.servicestack.net/why-not-odata)
- data import/load api (except for "insert into")
  -  independent syntax,  not necessarily in sql
  - eg. [doris Stream load](https://doris.apache.org/master/zh-CN/administrator-guide/load-data/stream-load-manual.html),  [snowflake snowpipe](https://docs.snowflake.com/en/user-guide/data-load-snowpipe-intro.html)

the first 2 cases  need many rest endpoints, it can take full advantage of rest api.  we can discuss it in another issue if needed.

query api need a few endspoints and is more complicated.

#  a brief summary of clickhouse http API

[clickhouse HTTP Interface](https://clickhouse.com/docs/en/interfaces/http/#http-interface).

- 2 kind of endpoints 
  - sql endpoint : send sql string in query parameters and body
  - Predefined endpoints :  define in config file, including endpoint and corresponding handler.
- http body is raw data (or sql with data in request), with specified format 
  - req 
    - behavior like clickhouse-client 
      - $ cat xxx.csv | clickhouse-client --query="INSERT INTO stock FORMAT CSVWithNames";
  - resp 
    - meta/summary/Progress in header
- sync response
- begin receiving data when buffer_size is reached 
  - unless wait_end_of_query is set
- session 
  - send the same session id in each request
  - End session after idle for some time.
- other features

  - auth
  - timeout
  - compression

# The problem with clickhouse http API

clickhouse HTTP  apiis special since it sends raw data in the body instead of embeding in json. This is efficient, straight forward,  and works well in most queries, but leads to some problems. 

https://github.com/ClickHouse/ClickHouse/labels/comp-http

if we want to get result early（!wait_end_of_query),  error may occur after the response code 200 is sent. currently err message is appended to data. solutions in issues:

1. [http trailer](https://github.com/ClickHouse/ClickHouse/issues/21136)  a standalone trailer part after
   1. most of http clients don't have a good support of trailers
2. SSE ([server-sent events](https://github.com/ClickHouse/ClickHouse/issues/6294))   seems most promising,  limitions:
   1. not support POST
   2. Text only
3.  [multipart/form-data](https://github.com/ClickHouse/ClickHouse/pull/28204)
4. web socket

from the discussions in issues they are [swing to grpc](https://github.com/ClickHouse/ClickHouse/issues/6294#issuecomment-734434886) instead of solving these problems.

## possible metadatas

- req 
  - session parameters
  - warehouse/database/schema/role
  - timeout
  - required resource 
- resp 
  - Error message
  - progress, summary，estimate time
  -  ”1  rows affect“
  - Warning
  - resource usage

# http api of other databases

Most db only use http api for management, only a few support query with sql via http.

- [snowflake](https://docs.snowflake.com/en/developer-guide/sql-api/index.html)
- presto （[trino](https://trino.io/docs/current/develop/client-protocol.html)， [treasure data](https://docs.treasuredata.com/display/public/PD/REST+APIs+in+Treasure+Data
  )）

- [kylin](http://kylin.apache.org/docs15/howto/howto_use_restapi.html#query)

- google [bigquery](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query) and  [spanner](https://cloud.google.com/spanner/docs/reference/rest)

  - [gRPC Transcodin](https://google.aip.dev/127)  
  - Use "session" as object/resource, and query as a verb on it.

  

very different features from clickhouse:

- body is json and data as its field, both in req and resp 
  - pros: 
    - more restful, flexible and expressiveness
    - seperate error/stats/result,  right response code with 
  - cons: 
    - **overhead to embed/extract data** 
    - client code need to know to interpret it. 
      - most apis response with the schema
- async and pagination (both return a handler for further fetch) 

  - Async if long time no result

    - Big shuffle
    - insert into ...select ..
  - pros 
    - async enable 1. poll for statuse/progress and  2. cancel 
  - cons: 
    - more round trip for large result. 
    - result set may be buffered in initial server?  need keep-alive 
-  [multi statements in a request](https://docs.snowflake.com/en/developer-guide/sql-api/guide.html#submitting-multiple-sql-statements-in-a-single-request) (snowflake only)  i.e. batching

  - [example scenarios](https://docs.snowflake.com/en/developer-guide/sql-api/guide.html#submitting-multiple-sql-statements-in-a-single-request)
  - pros 
    - Play a row of session in some scenario?
      - More clear when used to set up context
        - session variables
        - many session setting/parameters
        - Change the database, schema, warehouse, or role for statements in a request
    - save round trip overhead
      - e.g.   https://github.com/ClickHouse/ClickHouse/issues/6707#issuecomment-526357400
        - this is pending because the error handlering problem above.
  - cons: 
    - Need  to fetch result of each statement async with corresponding handler
      - or use sth like SSE
      - clickhouse support it in Predefined endpoints, which just concatenate the results
- [unique requestID](https://docs.snowflake.com/en/developer-guide/sql-api/guide.html#assigning-a-unique-request-id-for-resubmitting-requests) to ensure idempotent behavior for queries 

# features for  discussion

we can start with a MVP,  but need to make these decisions soon or later.

Some already discussed above.

1. choose how to report error after response header is sent.
   1. Except for solutions in clickhouse issues,  is pagination worth considering？
      1. It can be understood as using batching result instead of streaming chucks,  each batch is either a good data or error. 
2. do we need [Predefined http interface](https://clickhouse.com/docs/en/interfaces/http/#predefined_http_interface)?
3. Async for Long running query  for ELT
4. batching/multi-query
5. Support api for delete session to release resources quickly.
6. [unique requestID](https://docs.snowflake.com/en/developer-guide/sql-api/guide.html#assigning-a-unique-request-id-for-resubmitting-requests)
7. Put sql in a json?
   1. except for "insert into ", we don`t need to send data.
      1. Handle large inserts with dedicated endpoint
   2. seems better place then http query params and headers
   3. but need POST,  which SSE does not support,  if we choose it.

youngsofun:

## Summary

we plan to start with a simple protocol and  **trying to** be compatible with [presto-protocol](https://trino.io/docs/current/develop/client-protocol.html), for two reasons:

1.  Presto is the only mainstream DB with uses the REST API mainly. We can learn from it.
2.  **Trying to** be compatible, we hope to use preso client(lib, CLI..)  directly.  But may not be fully compatible, because the schema is diff, and it uses headers like "X-Presto-User", and the SQL/settings is also different.

The first version in this PR is not fully functional. Send this PR in advance to make sure I`m on the right track. I will complete them in the next pr. Current limitations:

1.  only support primary DataTypes(number, string, bool).
2. Return all results at once in response. But it is already prepared for async API.

## Changelog

- New Feature

## Related Issues

Fixes #2241  

## Test Plan

Unit Tests

### DIFF
--- a/query/src/servers/http/http_services.rs
+++ b/query/src/servers/http/http_services.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/query/src/servers/http/mod.rs
+++ b/query/src/servers/http/mod.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/query/src/servers/http/v1/block_to_json.rs
+++ b/query/src/servers/http/v1/block_to_json.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/query/src/servers/http/v1/block_to_json_test.rs
+++ b/query/src/servers/http/v1/block_to_json_test.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/query/src/servers/http/v1/mod.rs
+++ b/query/src/servers/http/v1/mod.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/query/src/servers/http/v1/statement.rs
+++ b/query/src/servers/http/v1/statement.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/query/src/servers/http/v1/statement_test.rs
+++ b/query/src/servers/http/v1/statement_test.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/query/src/servers/http_server.rs
+++ b/query/src/servers/http_server.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


youngsofun:
@BohuTANG  got it. sorry for misunderstanding.

BohuTANG:
> @BohuTANG got it. sorry for misunderstanding.

That's OK, we only need move them to another place :)
Welcome the first PR

Issue #2284, opened 2021-10-17
impl MetaEmbeddedSync with the MetaEmbedded.


drmingdrmer:
So that `MetaEmbeddedSync` and `MetaRemoteSync` provides the same caching functions.

- [x] #2304 
- [x] #2305

drmingdrmer:
Sorry. It is another task to make `sync` call to `async` call...
Reopen it.

Issue #2016, opened 2021-09-30
Remove EmbeddedMetaBackend, replace it with crate `common/meta/local-store`


drmingdrmer:
get this done when `kv` provides fully implemented `MetaApi`.

PR #2534, opened 2021-10-29
[query]: feature: run query with a embedded meta-store

drmingdrmer:

##### [query]: feature: run query with a embedded meta-store
- Feature: Replace `MetaEmbeddedSync` with a real impl backed with a
  raft-StateMachine.

  Get rid of the in-memoery fake meta-store.

  Since now the embedded meta and remote meta guarantee the same
  behavior.

  About unit-test: the async API set `MetaApi` are included in unittest.
  The corresponding `sync` version `MetaApiSync` has no corresponding
  unittest.
  They are gonna be removed soon.

- **NOTE** that in this version the embedded meta data is stored in a
  temp dir and will be removed when process quits.

- Config: A new config entry `meta_embedded_dir` is introduced but is not used.

- CI/test: Add building and testing script to github-workflow and makefile etc.

- fix: #2284

- fix: #2016 

## Changelog

- New Feature

- Improvement

## Related Issues

- #2030


drmingdrmer:
BTW the workflow name is tooooooo loooooong: `stateless standalone embedded meta store` blabla. 🤔

query/src/configs/config_test.rs
```diff
@@ -63,6 +63,7 @@ log_level = \"INFO\"
 log_dir = \"./_logs\"
 
 [meta]
+meta_embedded_dir = \"./_meta_embedded\"
```
BohuTANG, line 66: Cool!
meta_embedded_dir is a test purpose path, it would be better to set it as a random uuid instead config it in the config file?
If so, We can improve it in the another PR

query/src/configs/config_meta.rs
```diff
@@ -71,6 +76,7 @@ pub struct MetaConfig {
 impl MetaConfig {
     pub fn default() -> Self {
         MetaConfig {
```
BohuTANG, line 78: Cool!
meta_embedded_dir is a test purpose path, it would be better to set it as a random uuid instead config it in the config file?
If so, We can improve it in the another PR
cc @drmingdrmer

drmingdrmer:
Hm...
Forgot to let @sundy-li know about this feature...

Issue #1780, opened 2021-09-10
fuse table implementations


dantengsky:
**Summary**

- Data Ingestion
  - [x] #2193
  - [x] #2194
  - [x] #1781
  - [x] #1783
  - [x] #1785
    - [x] #2361
    - [x] #2383
- Data Scanning
  - [x] Integration with Indexing
     - [x] #1782
     - [x] #1801
     - [x] #2362
     - [x] "optimise" the min/max value if necessary (keep the size of values of boud reasonable, sacrifices some accuracy) 
  - [x] #1784
       Consider caching the following items
       - Table Snapshot
       - Table SegmentInfo
       - Parquet Meta Section
       - Column Data (of a block)
- Compact
  - [x] distributed insertion 
  - [x] #3121
  - [x] #3147
  - [x] #3144 
  - [x] #3176
  - [x] #2946
  - [x] fine-grained compaction
- misc
  - [x] tracings  
- Refactoring
  - [x] #2378 
- Testing

  - Unit tests
  - Integration Tests
    - [x] Local FileSystem
    - [x] Mock S3 Service (MinIO?)

Xuanwo:
@dantengsky Most work on this tracking issue has been implemented. How about closing it and start a new one if needed?

dantengsky:
> @dantengsky Most work on this tracking issue has been implemented. How about closing it and start a new one if needed?

let close it :)

PR #2555, opened 2021-10-31
add endpoint url param for s3 dal

dantengsky:

## Summary

- add `endpoint_url` to `S3StorageConfig`
- using custom Region if `endpoint_url` is enabled
- use `minio` in statless-standalone-test(ubuntu-latest only)
- minor code gc

@ZhiHanZ thanks for the github action code snippets , it works really well! but
- I failed to use it in config `macos-latest`, no docker there. thus minio is enabled only for `ubuntu-latest`.  if there are better solutions, please let me know.
- `workflows/stateless-tests-standalone.yml` has been tweaked to enable minio. there are three extra steps doing nothing in `macos-latest`... not as elegant as before. is there a better way to do this? or shall we improve it later?

@drmingdrmer thanks for suggesting using moto as the mock service, but @wubx kindly reminds me that there might be real cases of using minio later, and mocking is only used in integration tests in this PR, thus minio is taken.

part of 
- issue #1780 : Mock S3 Service by using MinIO
- issue #2561

## Changelog

- New Feature

## Related Issues

part of issue #1780 #2561

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
Cool~
What happens if we run `make stateless-test` locally?

BohuTANG:
> Cool~ What happens if we run `make stateless-test` locally?

Got it, the stateless set the s3 as the default storage :)

BohuTANG:
> but @wubx kindly reminds me that there might be real cases of using minio later, and mocking is only used in integration tests in this PR, thus minio is taken.

This is stateful test(we will support it later) not stateless test, cc @wubx

PsiACE:
LGTM, although stateless-test doesn't necessarily require s3 , it can help us verify that it works.

Perhaps we also need to provide such support for azure blob or other later?

drmingdrmer:
Does it mean we mac users can only test s3 backend in a CI?

dantengsky:
> Perhaps we also need to provide such support for azure blob or other later?

yes, I think we need them

dantengsky:
> Does it mean we mac users can only test s3 backend in a CI?

for "auto" tests, unfortunately, yes.  we can improve this eventually I think.

but  by exporting the env vars before `cargo test` `make stateless-test...`, something like this:

```
          export STORAGE_TYPE=s3
          export S3_STORAGE_BUCKET=testbucket
          export S3_STORAGE_REGION=us-east-1
          export S3_STORAGE_ENDPOINT_URL=http://127.0.0.1:9000
          export S3_STORAGE_ACCESS_KEY_ID=minioadmin
          export S3_STORAGE_SECRET_ACCESS_KEY=minioadmin
          bash ./scripts/ci/ci-run-stateless-tests-standalone.sh
```
we can test s3 backend locally (but manually setup)

drmingdrmer:
> > Does it mean we mac users can only test s3 backend in a CI?
> 
> for "auto" tests, unfortunately, yes. we can improve this eventually I think.
> 
> but by exporting the env vars before `cargo test` `make stateless-test...`, something like this:
> 
> ```
>           export STORAGE_TYPE=s3
>           export S3_STORAGE_BUCKET=testbucket
>           export S3_STORAGE_REGION=us-east-1
>           export S3_STORAGE_ENDPOINT_URL=http://127.0.0.1:9000
>           export S3_STORAGE_ACCESS_KEY_ID=minioadmin
>           export S3_STORAGE_SECRET_ACCESS_KEY=minioadmin
>           bash ./scripts/ci/ci-run-stateless-tests-standalone.sh
> ```
> 
> we can test s3 backend locally (but manually setup)

What about adding these config to a script like `stateless-test-s3` and let user run it by just one click?

dantengsky:
> What about adding these config to a script like stateless-test-s3 and let user run it by just one click?

good idea!  like `ci-run-stateless-tests-cluster-tls.sh`.

PR #2569, opened 2021-11-01
ISSUE 2567: pass limit+offset to sort pipeline

### LINKED ISSUE #2567: Query with 'order by' and 'limit offset' doesn't work as expect
**Summary**

Query like 'select * from numbers(100) order by number limit 5 offset 5', datafuse will return empty result.

**How to reproduce**
"select * from numbers(100) order by number limit 5 offset 5;"

**Error message and/or stacktrace**

**Additional context**

junli1026:

## Summary

Fix issue that 'select number from numbers(100) order by number limit 10 offset 10;' returns 0 rows;

## Changelog
- Bug Fix

## Related Issues

Fixes #2567

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
Great, can we have a stateless test to cover #2567 ?

sundy-li (approved):

Issue #1866, opened 2021-09-17
New testing style  in Rust crate


sundy-li:
**Summary**

Description for this feature.

Databend used Golang testing style currently,  but there is a more rustacean testing style of cargo, refer to https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html

jorgecarleitao:
I moved all tests in `arrow2` to a separate location and it made it much faster to write new tests, as we do not need to re-compile the crate. It also made it clear when a certain API was only "public" within the module (and not to users of the crate). Highly recommend!

PsiACE:
I prefer the Rust style.

BohuTANG:
Thank you @jorgecarleitao , your information is important, Databend test is getting slower and will continue on.
@PsiACE It's time to start doing it, go Rust style.

PR #2576, opened 2021-11-01
[test] apply the new test style to bendctl

PsiACE:

## Summary

apply the new test style to bendctl

## Changelog

- Build/Testing/CI

## Related Issues

Part of #1866

## Test Plan

Unit Tests

Stateless Tests


cli/Cargo.toml
```diff
@@ -7,6 +7,10 @@ license = "Apache-2.0"
 publish = false
 edition = "2021"
 
+[lib]
+doctest = false
+test = false
```
PsiACE, line 12: `doctest = false` avoid doc tests
`test = false` just compile once

PsiACE:
Unless necessary, all tests will be moved to the corresponding `tests/it` directory.

cli/Cargo.toml
PsiACE: There will be two fewer packages to compile

BohuTANG:
Cool.
After the test refined, how we measure the code coverage?

PsiACE:
> Cool. After the test refined, how we measure the code coverage?

bendctl involves almost no code coverage, I will check this issue in other package improvements. 

Continue with the previous method for now. There may be some degree of reduction, but I don't think there will be much difference overall.

Issue #2558, opened 2021-11-01
[mysql] improve cast compability


sundy-li:
**Summary**

tasks:
- [x] support partial cast options in [arrow2](https://github.com/jorgecarleitao/arrow2/pull/561)
- [ ] support `cast(x as signed)`, `cast(x as unsigned)` syntax

PR #2596, opened 2021-11-02
Add abs func

### LINKED ISSUE #2563: [functions] more functions working with number
**Summary**

Tasks list:

- [x] %, MOD	Modulo operator
- [x] +	Addition operator
- [x] -	Minus operator
- [x] /	Division operator
- [x] DIV	Integer division 

flaneur2020:

## Summary

This PR adds a `abs()` function.

```
mysql> SELECT abs(-438);
+--------------+
| abs((- 438)) |
+--------------+
|          438 |
+--------------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.002 sec., 456.13 rows/sec., 456.13 B/sec.

mysql> SELECT abs('test');
ERROR 1105 (HY000): Code: 7, displayText = expected number, but got String.
```

caveat: 

- this pr didn't take advantage of SIMD as arrow-rs has no `abs()` kernel yet, but it exists in the C++ library: https://arrow.apache.org/docs/cpp/compute.html#arithmetic-functions 

## Changelog

- Improvement

## Related Issues

Releated #2563

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
Good job!
It would be nice to have a stateless test for the numeric functions(there already a `PI` function), such as:
02_0014_function_numeric.sql

flaneur2020:
thank you for reminding, i'd add a 02_0014_function_maths.sql file with the test cases included : )

common/functions/src/scalars/maths/abs.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        match columns[0].data_type() {
+            DataType::Int8 => impl_abs_function!(columns[0], DataType::Int8, i8),
+            DataType::Int16 => impl_abs_function!(columns[0], DataType::Int16, i16),
+            DataType::Int32 => impl_abs_function!(columns[0], DataType::Int32, i32),
+            DataType::Int64 => impl_abs_function!(columns[0], DataType::Int64, i64),
+            DataType::Float32 => impl_abs_function!(columns[0], DataType::Float32, f32),
+            DataType::Float64 => impl_abs_function!(columns[0], DataType::Float64, f64),
+            DataType::UInt8 | DataType::UInt16 | DataType::UInt32 | DataType::UInt64 => {
```
sundy-li, line 88: For string array, we should rewrite to `abs(cast (c as bigint) )`.

```
> select abs('123');
+------------+
| abs('123') | 
+------------+
| 123         | 
+------------+
1 row in set (0.04 sec)
>  
```

common/functions/src/scalars/maths/abs.rs
```diff
@@ -0,0 +1,103 @@
...
+impl Function for AbsFunction {
+    fn name(&self) -> &str {
+        "abs"
+    }
+
+    fn num_arguments(&self) -> usize {
+        1
+    }
+
+    fn return_type(&self, args: &[DataType]) -> Result<DataType> {
```
sundy-li, line 72: It's wrong, should we set int8 => uint8 ?

tests/suites/0_stateless/02_0014_function_maths.sql
```diff
@@ -0,0 +1,4 @@
+SELECT pi();
+SELECT abs(-1);
+SELECT abs(-10086);
+SELECT abs('str');
```
BohuTANG, line 4: SELECT abs('str'); -- {ErrorCode 7}
This will check the error code by the stateless and can't dump the error result.

We shouldn't add the error message with the line no to the result:
```
ERROR 1105 (HY000) at line 4: Code: 7, displayText = Expected numeric types, but got String.
```

common/functions/src/scalars/maths/abs.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    pub fn desc() -> FunctionDescription {
+        FunctionDescription::creator(Box::new(Self::try_create))
+            .features(FunctionFeatures::default().deterministic())
+    }
+}
+
+macro_rules! impl_abs_function {
+    ($column:expr, $data_type:expr, $type:ident) => {{
+        let array = $column.column().to_array()?.get_array_ref();
```
sundy-li, line 47: Column::Constant is break by to_array().

we can use `to_minimal_array` and apply abs then resize_constant.

See `apply_arithmetic`

tests/suites/0_stateless/02_0014_function_maths.result
```diff
@@ -0,0 +1,4 @@
+ERROR 1105 (HY000) at line 4: Code: 7, displayText = Expected numeric types, but got String.
```
sundy-li, line 1: -- {ErrorCode 1105}

To match the error code.

common/functions/src/scalars/maths/abs.rs
flaneur2020: can I take the `cast_with_type()` method to cast `DFPrimitiveArray<Vec<u8>>` into `DFPrimitiveArray<i64>`?

sundy-li: Looks like it's ok.

tests/suites/0_stateless/02_0014_function_maths.result
flaneur2020: fixed, thank you

tests/suites/0_stateless/02_0014_function_maths.sql
flaneur2020: fixed, thank you

common/functions/src/scalars/maths/abs.rs
flaneur2020: TIL, i switched to `to_minimal_array`, thank you

common/functions/src/scalars/maths/abs.rs
flaneur2020: I added a cast parameter in the macro, coercing int8 to uint8, int16 to uint16, etc

common/functions/src/scalars/maths/abs.rs
flaneur2020: after `cast_with_type()`, it seems `abs('blah')` would coercing the value to NULL, without an error raised, would this behaviour expected?

sundy-li: This is because mysql's cast is partial cast, but arrow2's cast is using lexical_core's cast. Releated #2558

common/functions/src/scalars/maths/abs.rs
```diff
@@ -0,0 +1,129 @@
...
+                    .cast_with_type(&DataType::Float64)?
+                    .get_array_ref(),
+            ),
+            _ => (
+                $column.data_type().clone(),
+                $column.column().to_minimal_array()?.get_array_ref(),
+            ),
+        };
+
+        let primitive_array = arrow_array
```
sundy-li, line 63: We don't want users to use arrow2's array directly, because we already wrap arrow's array.

Try use the style of https://github.com/datafuselabs/databend/blob/main/common/functions/src/scalars/dates/number_function.rs#L389-L397

flaneur2020: thank you, i've take DFPrimitiveArray's `apply_cast_numeric` instead, much cleaner than the low level arrow stuff :+1:

sundy-li (approved):

PR #2609, opened 2021-11-02
add a JSON formatted schema field to system.tables

### LINKED ISSUE #2608: Allow get the fields of a table in a easy parsed format
**Summary**

currently we can get the table schema by `show create table xxx`, but it's in a sql format:

```
mysql> mysql> show create table system.metrics;
+---------+--------------------------------------------------------------------------------------------------------------------------+
| Table   | Create Table                                                                                                             |
+---------+--------------------------------------------------------------------------------------------------------------------------+
| metrics | CREATE TABLE `metrics` (
  `metric` String,
  `kind` String,
  `labels` String,
  `value` String,
) ENGINE=SystemMetrics |
+---------+--------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
Read 0 rows, 0 B in 0.000 sec., 0 rows/sec., 0 B/sec.
```

it's human friendly, but hard for machine to interpret the table schema.

currently I want build a sql IDE for databend, it has a explorer panel allowing user to view a table's fields, and searchable, but there's no way to easily discover the table fields yet.

there's a `system.columns` table in clickhouse: https://clickhouse.com/docs/en/operations/system-tables/columns/

maybe we could take a similar approach

flaneur2020:

## Summary

this PR adds a JSON formatted `schema` field into the system table `system.tables`, allowing user to easily make some tools with the knowledge of table fields.

## Changelog

- Improvement

## Related Issues

Fixes #2608

## Test Plan

Unit Tests
Stateless Tests


zhang2014:
Maybe we should add the `system`.`columns` table to describe it better?

sundy-li:
There is a bug, seems MySQL's default nullable is true if not set.

flaneur2020:
> Maybe we should add the `system`.`columns` table to describe it better?
> 
> `system`.`columns` table schema:
> 
> ```
> database String
> table String
> name String
> type String
> is_nullable tinyint
> ```

good idea 👍

flaneur2020:
> Maybe we should add the `system`.`columns` table to describe it better?
> 
> `system`.`columns` table schema:
> 
> ```
> database String
> table String
> name String
> type String
> is_nullable tinyint
> ```

in clickhouse there's a system.columns table, I'll close this PR and work on this approach instead.

https://clickhouse.com/docs/en/operations/system-tables/columns/

PR #2635, opened 2021-11-03
Try to remove PlanScan in PlanReadSource

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

Summary about this PR

- Try to remove PlanScan in PlanReadSource
- Make pushdowns work for fuse table
- Add benefit_column_prune in table trait

## Changelog

 
- Bug Fix
 

## Related Issues

Fixes #2632

## Test Plan

Unit Tests

Stateless Tests

PR #2648, opened 2021-11-04
[common/planners] refactor: remove PlanBuilder::scan(), which just builds a default push down and is never actually used.

drmingdrmer:

## Changelog

- Improvement

## Related Issues


sundy-li:
Love it, this is much cleaner than #2635

query/src/sql/plan_parser.rs
```diff
@@ -793,43 +778,27 @@ impl PlanParser {
...
-                                // TODO(xp): remove partitions, partitioning hint has been included in io_ctx.max_threads and io_ctx.query_nodes
-                                Some(partitions),
-                            )
-                            .map(PlanNode::ReadSource)
-                    }
-                    _unreachable_plan => panic!("Logical error: Cannot downcast to scan plan"),
-                })
+
+                // TODO: Move ReadSourcePlan to SelectInterpreter
+                let source_plan = table.read_plan(
```
sundy-li, line 793: table.read_plan will get the statistics by push_down, if we give it a default Extras, it'll return full statistics.

A quick fix is to read_plan 2 times.

1. read_plan in parser to build source_plan with default extras.
2. read_plan in plan_scheduler to build really source_plan.

But read_plan seems to be very expensive.

sundy-li: It's better than before, let's merge it firstly!

Issue #2343, opened 2021-10-20
Add monotonic check in FunctionFactory


zhyass:
**Summary**

```
/// The property of monotonicity for a certain range.
struct Monotonic {
    // Is the function monotonous (nondecreasing or nonincreasing).
    is_monotonic: bool, 
    // true if the function is nondecreasing, false, if notincreasing. If is_monotonic = false, then it does not matter.
    is_positive: bool, 
    // Is true if function is monotonic on the whole input range
    is_always_monotonic: bool, 
}
```

PR #2665, opened 2021-11-05
number table apply top-n (order-by and limit)

### LINKED ISSUE #2617: number table apply optimizer's limit and order by
**Summary**

For query like 'select number from numbers(10000000000) order by number limit 10', the read_table should just return first 10 number ordered asc, no need to return all numbers in the range.

junli1026:

## Summary

When order-by and limit are set, we just return top-n rows for every scan part.

## Changelog

- Improvement

## Related Issues

Fixes #2617

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/datasources/table_func/numbers_stream.rs
+++ b/query/src/datasources/table_func/numbers_stream.rs
@@ -17,6 +17,7 @@ use std::task::Poll;
 use std::usize;
 
 use common_datablocks::DataBlock;
+use common_datablocks::SortColumnDescription;
 use common_datavalues::prelude::*;
 use common_exception::Result;
 use futures::stream::Stream;
@@ -34,15 +35,24 @@ pub struct NumbersStream {
     schema: DataSchemaRef,
     block_index: usize,
     blocks: Vec<BlockRange>,
+    sort_columns_descriptions: Vec<SortColumnDescription>,
+    limit: Option<usize>,
 }
 
 impl NumbersStream {
-    pub fn try_create(ctx: DatabendQueryContextRef, schema: DataSchemaRef) -> Result<Self> {
+    pub fn try_create(
+        ctx: DatabendQueryContextRef,
+        schema: DataSchemaRef,
+        sort_columns_descriptions: Vec<SortColumnDescription>,
+        limit: Option<usize>,
+    ) -> Result<Self> {
         Ok(Self {
             ctx,
             schema,
             block_index: 0,
             blocks: vec![],
+            sort_columns_descriptions,
+            limit,
         })
     }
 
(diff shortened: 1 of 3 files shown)


query/src/datasources/table_func/numbers_stream.rs
```diff
@@ -108,6 +118,11 @@ impl NumbersStream {
 
             let series = DFUInt64Array::new_from_aligned_vec(av).into_series();
             let block = DataBlock::create_by_array(self.schema.clone(), vec![series]);
+            if !self.sort_columns_descriptions.is_empty() {
```
sundy-li, line 121: This did not help to block generate.

select number from numbers(1000) order by number desc limit 3;

`SortPartialTransform` already did the limit improvement for us.

We should push down the sort/limit into `try_get_one_block`.

junli1026: "We should push down the sort/limit into try_get_one_block."
-- This change is actually inside of function try_get_one_block. Or are you suggesting change the BlockRange ? 

Also @sundy-li , I have some questions related to this task:
1. Do we have plan to support query like "order by Func(number)". The push down to BlockRange can only apply for "order by number" right ? If we have plan, the execution of "Func(number)" can be pushed down to remote machine, right ? 
2. Another question is, is the "SortPartialTransform" executed on local machine ? I am thinking, the read_table can be executed in remote machine, thus it can decrease the amount of data to transfer right ?

junli1026: > This did not help to block generate.
> 
> select number from numbers(1000) order by number desc limit 3;
> 
> `SortPartialTransform` already did the limit improvement for us.
> 
> We should push down the sort/limit into `try_get_one_block`.

It is actually inside of function try_get_one_block. I think at least one of the benefit is, it decreases the data block size to transmit, right ? Every thread just send  limit number of rows, right ?

sundy-li: > I think at least one of the benefit is, it decreases the data block size to transmit, right
Yes, currently it benefits the data transfer to other nodes. It saves io mostly and it's better than previous.

But if you cut to limit the block, it can save io & CPU both, it's even better.

sundy-li: The number generation in try_get_block is in sequence, so it's easy to apply sort & limit pushdown optimization.

junli1026: sure, will address

sundy-li: For other functions like:  `order by number + 1`, `order by (number * 3)`, this is related to #2343.

In this feature, we just make it simple using column name match.

junli1026: Thanks! Yes, the monotonic check is something I thought about. And I thought you the logical would be implemented in the DataBlock sort, That is why I chose use DataBlock sort, instead of jut changing DataRange in the PR.
Thanks for clarifying, will address accordingly.

BTW, @sundy-li @BohuTANG , someone with investing background sent me a message, asking about the creator of this project. I forwarded the message to you in the Slack, could you take a look ?

BohuTANG: Sorry for later, sure :)
Thank you

sundy-li (approved):
/LGTM

PR #2669, opened 2021-11-05
[functions] add log func

### LINKED ISSUE #2563: [functions] more functions working with number

flaneur2020:

## Summary

Summary about this PR

this PR add a log() function which similar with the mysql function: https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_log

this PR tries take both literal & array value as the argument, but i'm not quite sure if this is the suggest way to accomplish this: 

- take `apply_with_idx` to iterate the num series with idx, then use `.try_get(idx)` to join the value of the other column

this PR only adds one `log()` func yet, if this function is ok, i'd add `ln, log2, log10` in another PR together.

## Changelog

- Improvement

## Related Issues

Fixes #2563

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/maths/log.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        let result = if columns.len() == 1 {
+            // Log(num) with default_base if one arg
+            let base_column = &DataColumn::Constant(self.default_base.into(), 1);
+            let num_series = columns[0]
+                .column()
+                .to_minimal_array()?
+                .cast_with_type(&DataType::Float64)?;
+            let base = DFTryFrom::try_from(base_column.try_get(0)?)?;
+            num_series.f64()?.apply_cast_numeric(|v| v.log(base))
```
sundy-li, line 73: why not just use ` let base =  self.default_base;`

common/functions/src/scalars/maths/log.rs
```diff
@@ -0,0 +1,97 @@
...
+                .f64()?
+                .apply_cast_numeric(|v| v.log(self.default_base))
+        } else {
+            // Log(base, num) if two args
+            let num_series = columns[1]
+                .column()
+                .to_minimal_array()?
+                .cast_with_type(&DataType::Float64)?;
+            let base_column: &DataColumn =
+                &columns[0].column().cast_with_type(&DataType::Float64)?;
```
sundy-li, line 81: see `ops/like.rs`

common/functions/src/scalars/maths/log.rs
flaneur2020: oh my, yes it is
fixed thank you

sundy-li (changes requested):

common/functions/src/scalars/maths/log.rs
flaneur2020: I guess you mean taking a `zip()` instead of `apply_with_idx()` ?

I try replaced it with a `zip()`, thank you

sundy-li: And we should combine the null bitmap.

flaneur2020: why the bitmap is required? q.q

in the string like scenery, i guess a bitmap can help represent the filtered rowset, here's a data tranform but do not really need filtering?

common/functions/tests/it/scalars/maths/log.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use std::f64::consts::E;
+
+use common_datavalues::prelude::*;
+use common_exception::Result;
+use common_functions::scalars::*;
+
+#[test]
+fn test_log_function() -> Result<()> {
+    #[allow(dead_code)]
```
pymongo, line 23: I think this line `#[allow(dead_code)]` can remove

flaneur2020: thank you for reminding, I've removed it, it's copy-pasted >.<

common/functions/src/scalars/maths/log.rs
sundy-li: > `bitmap` is to represent arrow's null bitmap

Int32Array:  `[Some(3), None, Some(5)]`

in arrow's memory layout, it contains int32 vec[3,0,5] and null bitmap b101 .

See `Validity bitmaps`  in https://arrow.apache.org/docs/format/Columnar.html.

common/functions/src/scalars/maths/log.rs
```diff
@@ -0,0 +1,107 @@
...
+            match base_column {
+                DataColumn::Constant(v, _) => {
+                    let base = DFTryFrom::try_from(v.clone())?;
+                    num_series.f64()?.apply_cast_numeric(|v| v.log(base))
+                }
+                DataColumn::Array(base_series) => {
+                    let iter = num_series
+                        .f64()?
+                        .into_no_null_iter()
+                        .zip(base_series.f64()?.into_no_null_iter())
```
sundy-li, line 91: You can add a test cast:

log(3, null) -> null
log(null, 3) -> null

common/functions/src/scalars/maths/log.rs
sundy-li: see: https://github.com/jorgecarleitao/arrow2/blob/main/src/compute/arity.rs#L138-L160

flaneur2020: TIL, thank you a lot, i need more learnings on arrow

tests/suites/0_stateless/02_0014_function_maths.sql
```diff
@@ -10,5 +10,8 @@ SELECT abs(TRUE); -- {ErrorCode 7}
 SELECT abs(NULL); -- {ErrorCode 7}
 SELECT abs(value) FROM math_sample_numbers;
 SELECT abs(value) + abs(-1) FROM math_sample_numbers;
+SELECT log(NULL);
+SELECT log(NULL, NULL); -- {ErrorCode 10}
+SELECT log('10', 100);
 
-DROP TABLE math_sample_numbers;
+DROP TABLE math_sample_numbers;
```
sundy-li, line 17: need new line

common/functions/src/scalars/maths/log.rs
flaneur2020: > see: https://github.com/jorgecarleitao/arrow2/blob/main/src/compute/arity.rs#L138-L160
do we have a wrapped `binary` function in the current code base? q.q

flaneur2020:
our arrow version do not support casting NULL to a float null array yet, so the stateless test LOG(NULL) would raise an casting error but not the expected a NULL result: https://github.com/jorgecarleitao/arrow2/blob/v0.6.2/src/compute/cast/mod.rs#L260

but in the latest arrow version it is supported: https://github.com/apache/arrow-rs/blob/v0.7.0/arrow/src/compute/kernels/cast.rs#L72

i think after upgrading it to v0.7.0, the NULL behaviour would become expected.

common/functions/src/scalars/maths/log.rs
flaneur2020: our arrow version do not support casting NULL to a float null array yet, so the stateless test LOG(NULL) would raise an casting error but not the expected a NULL result: https://github.com/jorgecarleitao/arrow2/blob/v0.6.2/src/compute/cast/mod.rs#L260

but in the latest arrow version it is supported: https://github.com/apache/arrow-rs/blob/v0.7.0/arrow/src/compute/kernels/cast.rs#L72

i think after upgrading it to v0.7.0, the NULL behaviour would become expected.

common/functions/src/scalars/maths/log.rs
sundy-li: Currently not, but we can have.

flaneur2020:
@sundy-li a bit confuse about the `into_no_null_iter()` method:

```
    pub fn into_no_null_iter(
        &'a self,
    ) -> impl TrustedLen<Item = &'a T> + Send + Sync + ExactSizeIterator {
        // .copied was significantly slower in benchmark, next call did not inline?
        self.array.values().iter()
    }
}
```

after goes into arrow's `values()` method body:

```
    /// The values [`Buffer`].
    /// Values on null slots are undetermined (they can be anything).
    #[inline]
    pub fn values(&self) -> &Buffer<T> {
        &self.values
    }
```

it shows that this `no_null_iterator` do contains the NULL values? but it's a undetermined value on NULL.

sundy-li:
> it shows that this no_null_iterator actually do contains the NULL values? but it's a undetermined value on NULL.

Not, the `values` is data vector like `vec<u64>, vec<u32>`, it will store default value if some items are null.

BohuTANG:
Thank you @pymongo for review :D

Issue #2627, opened 2021-11-03
Support `copy` command to load csv/parquet or s3 file into table


sundy-li:
**Summary**

Description for this feature.

 support copy command to load csv file into table.
 
 ```
COPY table_name FROM '/path/to/test.csv'
```

task list:

- [x] support copy stage csv/parquet file into table #2878 
- [x] support async read stream to parse formats #2786
- [x] #3082 

Reference:
https://www.postgresql.org/docs/13/sql-copy.html
https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html
https://duckdb.org/docs/sql/statements/copy

**Comment by ZhiHanZ** [comment]
I think it should be a query semantic in http handler?

**Comment by sundy-li** [comment]
>  I think it should be a query semantic in http handler?

But we need to upload localfile, how could http handler handle this?

**Comment by BohuTANG** [comment]
> > I think it should be a query semantic in http handler?
> 
> But we need to upload localfile, how could http handler handle this?

1. http handler provides a put file(like `xx_yy.csv`) command to a path as `tmp_dir` which configed in the query -- this upload a file to the internal stage
2. query supports the `COPY` command: `COPY table_name FROM 'xx_yy.csv'` -- query read the file from the internal stage

https://docs.snowflake.com/en/sql-reference/sql/put.html
https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#loading-files-from-an-internal-stage

**Comment by flaneur2020** [comment]
> > > I think it should be a query semantic in http handler?
> > 
> > 
> > But we need to upload localfile, how could http handler handle this?
> 
> 1. http handler provides a put file(like `xx_yy.csv`) command to a path as `tmp_dir` which configed in the query -- this upload a file to the internal stage
> 2. query supports the `COPY` command: `COPY INTO table_name FROM 'xx_yy.csv'` -- query read the file from the internal stage
> 
> https://docs.snowflake.com/en/sql-reference/sql/put.html https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#loading-files-from-an-internal-stage

the csv file have to located on the bend-query instance's local filesystem? 🤔

**Comment by flaneur2020** [comment]
how about copy file from a s3 or hdfs path? like:

```
COPY INTO table_name FROM 's3://xx_yy.csv?s3_token=xxxx'
```

sundy-li:
> how about copy file from a s3 or hdfs path? like:
> 
> ```
[...]

PR #2682, opened 2021-11-07
Copy command to load local file into table [patch-1]

sundy-li:

## Summary

Summary about this PR

## Changelog

- New Feature
 

## Related Issues

Related #2627

## Test Plan

Unit Tests

Stateless Tests


query/Cargo.toml
```diff
@@ -91,6 +91,8 @@ sha2 = "0.9.8"
 structopt = "0.3"
 structopt-toml = "0.5.0"
 threadpool = "1.8.1"
+
+tokio = { version = "1.13.0", features = ["fs"] }
```
BohuTANG, line 95: Better to use common/base/runtime?

sundy-li: A mistake made.

Issue #2589, opened 2021-11-02
[meta] panicked at 'called `Result::unwrap()` on an `Err` value: Code: 2407, displayText = targeting version 1, current version 2.


BohuTANG:
**Summary**

Failed at https://github.com/datafuselabs/databend/runs/4070390168?check_suite_focus=true

Supplementary info:

- A first guess is that `MetaCached` cached table info thus no updated table can be re-fetched in a CAS-loop. This is fixed in #2603 . Keep watching other CI to see if this exception disspears.
- Another issue is the `TableVersionMissMatch` error is not dealt with. The following TODO is meant to get this done:

https://github.com/datafuselabs/databend/blob/088f8bbb95d2a7c04cd0fe67df795765cd23ce3f/query/src/datasources/table/fuse/table_do_append.rs#L70-L96

```
test servers::clickhouse::clickhouse_handler_test::test_clickhouse_insert_to_fuse_table has been running for over 60 seconds
test servers::clickhouse::clickhouse_handler_test::test_clickhouse_insert_to_fuse_table ... FAILED

failures:

---- servers::clickhouse::clickhouse_handler_test::test_clickhouse_insert_to_fuse_table stdout ----
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Code: 2407, displayText = targeting version 1, current version 2.

   0: common_exception::exception::ErrorCode::TableVersionMissMatch
             at /home/runner/work/databend/databend/common/exception/src/exception.rs:110:77
   1: common_meta_embedded::meta_api_impl::<impl common_meta_api::meta_api::MetaApi for common_meta_embedded::meta_embedded::MetaEmbedded>::upsert_table_option::{{closure}}
             at /home/runner/work/databend/databend/common/meta/embedded/src/meta_api_impl.rs:266:24
   2: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/b69fe57261086e70aea9d5b58819a1794bf7c121/library/core/src/future/mod.rs:80:19
   3: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/b69fe57261086e70aea9d5b58819a1794bf7c121/library/core/src/future/future.rs:119:9
   4: <databend_query::catalogs::backends::impls::meta_cached::MetaCached as common_meta_api::meta_api::MetaApi>::upsert_table_option::{{closure}}
             at src/catalogs/backends/impls/meta_cached.rs:113:9
[...]

PR #2688, opened 2021-11-07
[query/server/http] Add /v1/query.

### LINKED ISSUE #2604: post json instead of sql in  /v1/statement/
**Summary**

the first version of http handler,  post raw sql statement to   /v1/statement/.
I think it is better to post json instead to carry more context info.

```
{
"sql": "select * from number(10)" 
}
```

## reasons  

1.  one place to organize all context infos.
2. header and params are key-value, need other effort to embed complex info 
3. params and header have encode requirements  and length limitation.

## how

since  /v1/statement/  is used in  cli and playground. we need to discuss the way to do it.

1.  let  /v1/statement/ accept  json only 
3. add a new endpoint /v1/query,   abandon  /v1/statement/
4. add a new endpoint /v1/query for main use,  keep  /v1/statement/ for simple cases.

## possible reason to reserve the raw_sql api: 

1.  make it easy to write for tools like curl.
2.  it looks like the presto protocol (but it is difficult to be really compatible)
3. click house use it to insert data (but we will have other api for data import)

sorry for not thinking  it clearly to begin with.

@flaneur2020 @ZhiHanZ 

## related issue

https://github.com/datafuselabs/databend/issues/2241

youngsofun:

## Summary

Add new endpoint /v1/query to support sql.

# handler usage

1. post query with JSON with format HttpQueryRequest, and return JSON with format QueryResponse
2. Pagination with long polling.
3. result data is embedded in QueryResponse.
4. return the following  4 URIs,   the client should rely on the endpoint  /v1/query and URIs returned in the response. the other endpoints may change without notices.
   1. next: get the next page (together with progress) with long polling.   page N is released when Page N+1 is requested. 
   2. state: return struct like QueryResponse,  but the data field is empty. short polling.
   3. kill: kill running query
   4. delete: kill running query and delete it

[test_async](https://github.com/datafuselabs/databend/blob/aeeffa6e9a582d5073745b5b7b446da0c5166880/query/src/servers/http/v1/http_query_handlers_test.rs#L75) can be a demo for their usage.  

``` rust
pub struct HttpQueryRequest {
    pub sql: String,
}
```

```rust
pub struct QueryResponse {
    pub id: Option<String>,
    pub columns: Option<DataSchemaRef>,
    pub data: JsonBlockRef,
    pub next_uri: Option<String>,
    pub state_uri: Option<String>,
    pub kill_uri: Option<String>,
    pub delete_uri: Option<String>,
    pub query_state: Option<String>,
    pub request_error: Option<String>,
    pub query_error: Option<String>,
    pub query_progress: Option<ProgressValues>,
}

```

## internal 

1. The web layer (query_handlers.rs)  is separated from a more structured internal implementation in mod v1/query. 
     2. /v1/statement is kept to post raw SQL, may reuse the internal implementation, this will be done some others changes in another PR.
     3.   hope it may event be used to support GRPC if need later?
3. make sure the internal tokio task stop fast when the query is killed by SQL command or HTTP request

## TODO

soon( new PR in 1 week or so)：

1. classification and organization of Errors returned will be polished later.
2. more tests.
3. rewrite /v1/statements.

Long term:

1. Add client session support.
2. Adjust the handler interface to a stable version with a formal doc.  only necessary fields are added currently.  including 2 JSON formats and optional URL parameters and headers.
3. better control of memory.

## Changelog

- New Feature

## Related Issues

Fixes #2604 

## Test Plan

Unit Tests

### DIFF
--- a/query/src/servers/http/v1/http_query_handlers.rs
+++ b/query/src/servers/http/v1/http_query_handlers.rs
@@ -1,0 +1,242 @@
+// Copyright 2021 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::cmp::Ordering;
+use std::sync::Arc;
+use std::time::Duration;
+use std::time::Instant;
+
+use common_base::ProgressValues;
+use common_datavalues::DataSchemaRef;
+use common_exception::Result;
+use hyper::http::header;
+use poem::get;
+use poem::http::StatusCode;
+use poem::post;
+use poem::web::Data;
+use poem::web::Json;
+use poem::web::Path;
+use poem::web::Query;
+use poem::IntoResponse;
+use poem::Response;
+use poem::Route;
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::servers::http::v1::block_to_json::JsonBlockRef;
+use crate::servers::http::v1::query::execute_state::HttpQueryRequest;
+use crate::servers::http::v1::query::http_query::HttpQueryResponseInternal;
+use crate::servers::http::v1::query::result_data_manager::Wait;
+use crate::sessions::SessionManagerRef;
+
+pub fn make_page_uri(query_id: &str, page_no: usize) -> String {
+    format!("/v1/query/{}/page/{}", query_id, page_no)
+}
+
+pub fn make_state_uri(query_id: &str) -> String {
+    format!("/v1/query/{}", query_id)
+}
+
+pub fn make_delete_uri(query_id: &str) -> String {
+    format!("/v1/query/{}/kill?delete=true", query_id)
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct QueryResponse {
+    pub id: Option<String>,
+    pub columns: Option<DataSchemaRef>,
+    pub data: JsonBlockRef,
+    pub next_uri: Option<String>,
+    pub state_uri: Option<String>,
+    pub delete_uri: Option<String>,
+    // TODO(youngsofun): consider better response for error
+    // 1. another json format for request error
+    // 2. return 400 for some situation
+    // 3. more detail from ErrorCode (does not support Serialization))
+    // 4. use error code to ease the handling in client program
+    pub request_error: Option<String>,
+    pub query_error: Option<String>,
+    pub query_state: Option<String>,
+    pub query_progress: Option<ProgressValues>,
+}
+
+impl QueryResponse {
+    fn from_internal(id: String, result: &Result<HttpQueryResponseInternal>) -> QueryResponse {
+        match result {
+            Ok(r) => {
+                let (data, next_url) = match &r.data {
+                    Some(d) => (
+                        d.page.data.clone(),
+                        d.next_page_no.map(|n| make_page_uri(&id, n)),
+                    ),
+                    None => (Arc::new(vec![]), None),
+                };
+                let columns = match &r.initial_state {
+                    Some(v) => v.schema.clone(),
+                    None => None,
+                };
+                QueryResponse {
+                    data,
+                    columns,
+                    id: Some(id.clone()),
+                    next_uri: next_url,
+                    state_uri: Some(make_state_uri(&id)),
+                    delete_uri: Some(make_delete_uri(&id)),
+                    query_error: r.state.error.clone(),
+                    query_progress: r.state.progress.clone(),
+                    query_state: Some(r.state.state.to_string()),
+                    request_error: None,
+                }
+            }
+            Err(e) => QueryResponse::simple_error(Some(id), e.message()),
+        }
+    }
+
+    fn not_found(query_id: String) -> QueryResponse {
+        QueryResponse::simple_error(None, format!("query id not found {}", query_id))
+    }
+
+    fn simple_error(id: Option<String>, message: String) -> QueryResponse {
+        QueryResponse {
+            id,
+            data: Arc::new(vec![]),
+            query_error: None,
+            columns: None,
+            query_progress: None,
+            next_uri: None,
+            state_uri: None,
+            delete_uri: None,
+            query_state: None,
+            request_error: Some(message),
+        }
+    }
+}
+
+impl IntoResponse for QueryResponse {
+    fn into_response(self) -> Response {
+        let body = serde_json::to_vec(&self).unwrap();
+        // TODO(youngsofun): when should we return other status code here?
+        let status = StatusCode::OK;
+        let content_type = "application/json";
+        Response::builder()
+            .status(status)
+            .header(header::CONTENT_TYPE, content_type)
+            .body(body)
+    }
+}
+
+#[derive(Deserialize, Debug)]
+pub(crate) struct CancelParams {
+    delete: Option<bool>,
+}
+
+#[poem::handler]
+async fn query_cancel_handler(
+    sessions_extension: Data<&SessionManagerRef>,
+    Query(params): Query<CancelParams>,
+    Path(query_id): Path<String>,
+) -> impl IntoResponse {
+    let session_manager = sessions_extension.0;
+    let http_query_manager = session_manager.get_http_query_manager();
+    match http_query_manager.get_query_by_id(&query_id).await {
+        Some(query) => {
+            query.kill().await;
+            if params.delete.unwrap_or(false) {
+                http_query_manager.remove_query_by_id(&query_id).await;
+            }
+            StatusCode::OK
+        }
+        None => StatusCode::NOT_FOUND,
+    }
+}
+
+#[poem::handler]
+async fn query_state_handler(
+    sessions_extension: Data<&SessionManagerRef>,
+    Path(query_id): Path<String>,
+) -> impl IntoResponse {
+    let session_manager = sessions_extension.0;
+    let http_query_manager = session_manager.get_http_query_manager();
+    match http_query_manager.get_query_by_id(&query_id).await {
+        Some(query) => {
+            let response = query.get_response_state_only().await;
+            QueryResponse::from_internal(query_id, &Ok(response))
+        }
+        None => QueryResponse::not_found(query_id),
+    }
+}
+
+#[derive(Deserialize, Debug)]
+pub(crate) struct PageParams {
+    // for now, only used for test
+    wait_time: Option<i32>,
+}
+
+impl PageParams {
+    fn get_wait_type(&self) -> Wait {
+        let t = self.wait_time.unwrap_or(10);
+        match t.cmp(&0) {
+            Ordering::Greater => Wait::Deadline(Instant::now() + Duration::from_secs(t as u64)),
+            Ordering::Equal => Wait::Async,
+            Ordering::Less => Wait::Sync,
+        }
+    }
+}
+
+#[poem::handler]
+async fn query_page_handler(
+    sessions_extension: Data<&SessionManagerRef>,
+    Query(params): Query<PageParams>,
+    Path((query_id, page_no)): Path<(String, usize)>,
+) -> impl IntoResponse {
+    let session_manager = sessions_extension.0;
+    let http_query_manager = session_manager.get_http_query_manager();
+    match http_query_manager.get_query_by_id(&query_id).await {
+        Some(query) => {
+            let wait_type = params.get_wait_type();
+            let result = query.get_response_page(page_no, &wait_type, false).await;
+            QueryResponse::from_internal(query_id, &result)
+        }
+        None => QueryResponse::not_found(query_id),
+    }
+}
+
+#[poem::handler]
+pub(crate) async fn query_handler(
+    sessions_extension: Data<&SessionManagerRef>,
+    Query(params): Query<PageParams>,
+    Json(req): Json<HttpQueryRequest>,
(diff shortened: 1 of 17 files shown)


sundy-li:
I noticed `query_id` is generated by the server, but what if we want to retry an insert/update query?

Time1 query sends update SQL:  `update table set money = money + 1`, the server handles it well and returns the response to the client, but the client may crash by somehow and it will retry the update SQL.

How to handle this situation?

I think the query_id can be identified by the client or make random uuid by the server if not.

youngsofun:
the query_id here is in fact a handler, corresponding to an  HTTP request. 
 in this context, ​it is better for the server to generate it，because the server is not sure client-side id is unique.

snowflake use another optional “requestID” field in JSON for the situation you describe.

we can add it later.  need to be maintained and maybe better names.

youngsofun:
both IDs are needed.

In most cases, the query is "select" without modification, so in the client(wrapper of http api), a client-id can be added automatically when the query is insert/delete

flaneur2020:
> kill: kill running query
> delete: kill running query and delete it

is it ok to only have the delete operation? q.q

query/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub next_uri: Option<String>,
+    pub state_uri: Option<String>,
+    pub kill_uri: Option<String>,
+    pub delete_uri: Option<String>,
+    // TODO(youngsofun): consider better response for error
+    // 1. another json format for request error
+    // 2. return 400 for some situation
+    // 3. more detail from ErrorCode (does not support Serialization))
+    // 4. use error code to ease the handling in client program
+    pub request_error: Option<String>,
```
flaneur2020, line 73: how about define a QueryError which implements an IntoResponse? like this:

```
query_cancel_handler() -> Result<impl IntoResponse, QueryError>
```

maybe it could get the error handling easier

query/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub query_error: Option<String>,
```
flaneur2020, line 74: how about unify these two error field into one error field with a specific Error type? A typed error object might easier to extend with some more error kinds in the future.

there's a QueryError definition in presto:

> If query failed, the error attribute will contain JSON for a QueryError object. That object contains a message, an errorCode and other information about the error. See the QueryError class for more details.
> Non-null only if the query resulted in an error. QueryResults.failureInfo of type FailureInfo has detail on the reason for the failure, including a stack trace, and FailureInfo.errorLocation, providing the query line number and column number where the failure was detected.

query/src/servers/http/v1/query/result_data_manager.rs
```diff
@@ -0,0 +1,147 @@
...
+            Async => block_rx.try_recv(),
+            Sync => match block_rx.recv().await {
+                Some(block) => Ok(block),
+                None => Err(TryRecvError::Disconnected),
+            },
+            Deadline(t) => {
+                let sleep = tokio::time::sleep_until(tokio::time::Instant::from_std(*t));
+                tokio::select! {
+                    biased;
+                    block= block_rx.recv() => {
```
flaneur2020, line 117: `block=` ---> `block =` by a `make lint`?

query/src/sessions/context_shared.rs
```diff
@@ -58,6 +59,7 @@ pub struct DatabendQueryContextShared {
     pub(in crate::sessions) ref_count: Arc<AtomicUsize>,
     pub(in crate::sessions) subquery_index: Arc<AtomicUsize>,
     pub(in crate::sessions) running_query: Arc<RwLock<Option<String>>>,
+    pub(in crate::sessions) http_query: Arc<RwLock<Option<HttpQueryHandle>>>,
```
flaneur2020, line 62: what's the major difference between the HttpQuery and the queries from mysql & clickhouse ? q.q

query/src/servers/http/v1/query/result_data_manager.rs
youngsofun: ok thanks

query/src/sessions/context_shared.rs
youngsofun: for now:  http_query may exist after the session is stopped because we now have no other place to hold the results.

but you remend of another choice:

IF:

1.  we keep server-side sessions for clients
2. Keeping the session after the query stopped does not cost too much.

then we can 

1. use (session ID，query_seq) as request handle for async query to find itself.
2. make struct http_query a optional part of a session.

youngsofun:
> how about unify these two error field into one error field with a specific Error type? A typed error object might easier to extend with some more error kinds in the future.

@flaneur2020 


request_error is for simple situations(page/query not exists ) may return 400/404

query_error is used together with the "state" field.
if the state is stopped, the client will check if  query_error is set and what is it.
and may include more details.

youngsofun:
> is it ok to only keep one delete operation without kill? (or keep the kill operation but without the delete operation) q.q

I think it ok to only keep delete for now.

flaneur2020:
> > how about unify these two error field into one error field with a specific Error type? A typed error object might easier to extend with some more error kinds in the future.
> 
> @flaneur2020
> 
> request_error is for simple situations(page/query not exists ) may return 400/404
> 
> I just fix with commit [“fail to start query" as a special error.](https://github.com/datafuselabs/databend/pull/2688/commits/87427bb817ac0d5d248777ecf0974e46e450067b)
> 
> I`m considering removing request_error out, and returning 400/404, but have not decided on the details. 

on getting 404 / 400, maybe we could define a unified response message？the response object on getting 404 might diferentates the QueryResponse, like:

```
> GET /blahblahblah 
404 Not Found
{ error: "UrlNotFound", message: "url not found" }
```

```
> POST /blahblah
400 Bad Request
{ error: "BadArgument", message: "field is missing", info: {"query": "this field is required"} }
```

> is it ok to merge this pr as it is now and change later?

sure, never mind, this is not critical. a unified error representation might make the frontend side life easier :)

query/src/sessions/context_shared.rs
flaneur2020: what will happen on we started a HTTP query and it is running, then we executed a `KILL <qid>` in the mysql console? q.q

query/src/servers/http/v1/query/result_data_manager.rs
flaneur2020: oh, maybe rustfmt got some trouble on interpreting the macros i guess 🤔

youngsofun:
> what will happen on we started a HTTP query and it is running, then we executed a KILL <qid> in the mysql console? q.q

client will find it is killed when request next_uri or state_uri

youngsofun:
> on getting 404 / 400, maybe we could define a unified response message？ the response object on getting 404 might diferentates the QueryResponse

this is just the part that I`m not sure.

do we really need a JSON? Maybe an error string in the body is enough for now.
the JSON is needed if we want to make it easier for client program to handle it. a error code may be helpful.
```JSON
{
   "code":2333
  "message": "oh no"
}
```
but 404 and 400 already contain the meaning of UrlNotFound and BadArgument. It seems enough for now.
It is clear enough since the client also knows the URI it requests.

another reason is to use plan text is we now use poem Json Extractor.  when it meat an invalid JSON, it just returns a 400 with an error message in the body.  of course, we can Extract a String and decode ourselves,  but I'm wondering is it worth.


.

youngsofun:
@ZhiHanZ 
```
Can't find any online and idle self-hosted runner in the current repository, account/organization that matches the required labels: 'stateful-runner'
Waiting for a self-hosted runner to pickup this job...
```

BohuTANG:
Please merge with the main branch, the Stateful test fixed in  https://github.com/datafuselabs/databend/pull/2717

youngsofun:
tests passed on my mac

but ci fail:

```
---- servers::clickhouse::clickhouse_handler_test::test_clickhouse_insert_to_fuse_table stdout ----
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Code: 2407, displayText = targeting version 1, current version 2.

   0: common_exception::exception::ErrorCode::TableVersionMissMatch
             at /home/runner/work/databend/databend/common/exception/src/exception.rs:110:77
   1: common_meta_embedded::meta_api_impl::<impl common_meta_api::meta_api::MetaApi for common_meta_embedded::meta_embedded::MetaEmbedded>::upsert_table_option::{{closure}}
             at /home/runner/work/databend/databend/common/meta/embedded/src/meta_api_impl.rs:260:24
```

BohuTANG:
> tests passed on my mac
> 
> but ci fail:
> 
> https://github.com/datafuselabs/databend/runs/4147948472?check_suite_focus=true
> 
> ```
> ---- servers::clickhouse::clickhouse_handler_test::test_clickhouse_insert_to_fuse_table stdout ----
> thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Code: 2407, displayText = targeting version 1, current version 2.
> 
>    0: common_exception::exception::ErrorCode::TableVersionMissMatch
>              at /home/runner/work/databend/databend/common/exception/src/exception.rs:110:77
>    1: common_meta_embedded::meta_api_impl::<impl common_meta_api::meta_api::MetaApi for common_meta_embedded::meta_embedded::MetaEmbedded>::upsert_table_option::{{closure}}
>              at /home/runner/work/databend/databend/common/meta/embedded/src/meta_api_impl.rs:260:24
> ```

cc @dantengsky @drmingdrmer

youngsofun:
insert timeout 

https://github.com/datafuselabs/databend/blob/main/query/src/servers/clickhouse/clickhouse_handler_test.rs#L100

drmingdrmer:
> > tests passed on my mac
> > but ci fail:
> > https://github.com/datafuselabs/databend/runs/4147948472?check_suite_focus=true
> > ```
> > ---- servers::clickhouse::clickhouse_handler_test::test_clickhouse_insert_to_fuse_table stdout ----
> > thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Code: 2407, displayText = targeting version 1, current version 2.
> > 
> >    0: common_exception::exception::ErrorCode::TableVersionMissMatch
> >              at /home/runner/work/databend/databend/common/exception/src/exception.rs:110:77
> >    1: common_meta_embedded::meta_api_impl::<impl common_meta_api::meta_api::MetaApi for common_meta_embedded::meta_embedded::MetaEmbedded>::upsert_table_option::{{closure}}
> >              at /home/runner/work/databend/databend/common/meta/embedded/src/meta_api_impl.rs:260:24
> > ```
> 
> cc @dantengsky @drmingdrmer

This is a known issue that `get_table()` returns stale caches.
@zhang2014 has been working on it: 
https://github.com/datafuselabs/databend/issues/2589#issuecomment-958784245

sunli829:
> another reason is to use plan text is we now use poem Json Extractor. when it meat an invalid JSON, it just returns a 400 with an error message in the body. of course, we can Extract a String and decode ourselves, but I'm wondering is it worth.
> 
> 

You can handle the errors of the JSON extractor yourself, like this:



```rust
#[poem::handler]
fn index(req: Result<Json<R>, ParseJsonError>) {
    let req = match req {
        Ok(req) => req,
        Err(err) => {
            // parse json error
        }
    };
}
```

youngsofun:
@sunli829  great!  thank you

query/src/sessions/context_shared.rs
```diff
@@ -91,6 +94,11 @@ impl DatabendQueryContextShared {
             source_abort_handle.abort();
         }
 
+        let http_query = self.http_query.read();
+        if let Some(handle) = &*http_query {
+            handle.abort();
```
youngsofun, line 99: need this because to abort the channel send/recv between the task generate datastream and handle.
and may later add query_id or something to find HTTP_query associate from the session/query  if needed

@zhang2014

query/src/sessions/sessions.rs
```diff
@@ -42,6 +44,7 @@ pub struct SessionManager {
     pub(in crate::sessions) discovery: ClusterDiscoveryRef,
     pub(in crate::sessions) catalog: Arc<DatabaseCatalog>,
     pub(in crate::sessions) user: UserManagerRef,
+    pub(in crate::sessions) http_query_manager: HttpQueryManagerRef,
```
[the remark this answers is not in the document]
zhang2014: Look good to me.

query/src/servers/http/v1/mod.rs
```diff
@@ -15,6 +15,10 @@
 pub mod block_to_json;
 #[cfg(test)]
 mod block_to_json_test;
+pub mod http_query_handlers;
```
zhang2014, line 18: Why need pub mod, maybe replace it with `mod http_query_handlers; pub use ...` is better?

youngsofun: thanks，I will check all “pub”

query/src/sessions/context_shared.rs
zhang2014: There may be no need new abort stream for http handler(All sources are also abortable).

query/src/sessions/sessions.rs
```diff
@@ -78,6 +84,9 @@ impl SessionManager {
         self.discovery.clone()
     }
 
+    pub fn get_http_query_manager(self: &Arc<Self>) -> HttpQueryManagerRef {
+        self.http_query_manager.clone()
+    }
```
zhang2014, line 89: maybe need new line

query/src/servers/http/v1/mod.rs
youngsofun: fixed, thank you

query/src/sessions/context_shared.rs
youngsofun: currently to make sure kill success, I`ll try to think of a better solution later.

youngsofun:
@zhang2014 all done, thank you

zhang2014 (approved):

youngsofun:
@BohuTANG is it ok to merge?

youngsofun:
@BohuTANG is it ok to merge?

PR #2689, opened 2021-11-07
[common/exception] feature: add AnyError as a representation of std error for transmit

drmingdrmer:

##### [common/exception] feature: add AnyError as a representation of std error for transmit
By converting std::Error to AnyError, structured error can be serialized
and transferred.
The **type** of an error is not kept. However, AnyError stores the
`description`, `source` and `backtrace` info.

E.g., to define a structured error:

```rust
pub enum MetaError {
    #[error(transparent)]
    ConnectionError(#[from] ConnectionError),
    ...
}
pub struct ConnectionError {
    msg: String,
    #[source]
    source: AnyError,
}

impl ConnectionError {
    pub fn new(source: tonic::transport::Error, msg: String) -> Self {
        Self { msg, source: AnyError::new(&source), }
    }
}

// Usage:
let err:MetaError = ConnectionError::new(tonic_err, "when connecting to 1.2.3.4").into()
```

## Changelog

- New Feature

## Related Issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1980,6 +1980,7 @@ dependencies = [
  "anyhow",
  "async-raft",
  "async-trait",
+ "backtrace",
  "byteorder",
  "common-arrow",
  "common-base",
--- a/metasrv/Cargo.toml
+++ b/metasrv/Cargo.toml
@@ -39,6 +39,7 @@ common-macros = { path = "../common/macros" }
 anyhow = "1.0.45"
 async-raft = { git = "https://github.com/datafuse-extras/async-raft", tag = "v0.6.2-alpha.14" }
 async-trait = "0.1"
+backtrace = "0.3.63"
 byteorder = "1.1.0"
 env_logger = "0.9"
 futures = "0.3"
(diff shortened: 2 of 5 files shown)


zhang2014:
Why not convert them to `ErrorCode`(may unified error is better)?

drmingdrmer:
> Why not convert them to `ErrorCode`(may unified error is better)?

Hmm... `ErrorCode` itself does not impl `Serialize`, although there is a  `SerializedError`, which requires manually conversion.

And sometimes I found a structured serializable error would be easier to use:
- Serializable: make it easier to transmit an error between nodes.
- Structured: reduce branching: I can just define an enum `Retry` to contain every error that allows to re-send an RPC. This way I do not need to check every error code one by one.
- The error type is able to contain the root error for debugging, such as a std Error, which is not serializable.


IMO, ErrorCode is suitable for end-user. For internal use, structured error is more suited to my need.

drmingdrmer:
AnyError is meant to be used internally only. Currently only inside and between metasrv nodes.

zhang2014:
Unit tests need fixes.

> IMO, ErrorCode is suitable for end-user. For internal use, structured error is more suited to my need.

Maybe move it to metaservice is better

drmingdrmer:
> Unit tests need fixes.
> 
> > IMO, ErrorCode is suitable for end-user. For internal use, structured error is more suited to my need.
> 
> Maybe move it to metaservice is better

It's still the get-table bug... 🤔 

Ok. moving to metasrv

Issue #2668, opened 2021-11-05
metasrv add cli args: `--join addr,addr,...` to bring up a metasrv node and join an existent cluster.


drmingdrmer:

drmingdrmer:
- Fixed in #2701

PR #2701, opened 2021-11-08
[metasrv] feature: add `join` API to join a node to cluster.

drmingdrmer:

##### [metasrv] feature: add `join` API to join a node to cluster.
- Feature: add cli arg `--join addr1,addr2...` to join a new node to a
  metasrv cluster.

- Feature: add RPC `forward` to forward a request to other metasrv node.
  It is used to proxy an admin request to a known leader.

- Refactor: add `MetaError` which is serializable and structured, to
  express errors a meta server emits.

  `MetaErro` can be transmitted between metasrv nodes.

- Refactor: add `MetaLeader` as a container of admin API.

- Refactor: simplify From/Into for `RaftMes`

- Test: add unit test of `join`ing node to cluster.

- part of #2668

## Changelog

- New Feature

## Related Issues

### DIFF
--- a/common/meta/raft-store/src/config.rs
+++ b/common/meta/raft-store/src/config.rs
@@ -111,6 +111,12 @@ pub struct RaftConfig {
     )]
     pub single: bool,
 
+    /// Bring up a metasrv node and join a cluster.
+    ///
+    /// The value is one or more addresses of a node in the cluster, to which this node sends a `join` request.
+    #[structopt(long, env = "METASRV_JOIN")]
+    pub join: Vec<String>,
+
     #[structopt(
     long,
     env = KVSRV_ID,
--- a/metasrv/src/api/flight_server.rs
+++ b/metasrv/src/api/flight_server.rs
@@ -28,7 +28,11 @@ use transport::ServerTlsConfig;
 
 use crate::api::rpc::MetaFlightImpl;
 use crate::configs::Config;
+use crate::meta_service::AdminRequest;
+use crate::meta_service::AdminRequestInner;
+use crate::meta_service::JoinRequest;
 use crate::meta_service::MetaNode;
+use crate::proto::meta_service_client::MetaServiceClient;
 
 pub struct FlightServer {
     conf: Config,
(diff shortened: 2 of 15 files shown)


metasrv/src/api/flight_server.rs
```diff
@@ -92,24 +96,50 @@ impl FlightServer {
...
 
-        let mn = if meta_config.boot {
-            MetaNode::boot(0, meta_config).await?
-        } else if meta_config.single {
+        let mn = if raft_config.boot {
+            MetaNode::boot(0, raft_config).await?
+        } else if raft_config.single {
             let (mn, _is_open) =
-                MetaNode::open_create_boot(meta_config, Some(()), Some(()), Some(())).await?;
+                MetaNode::open_create_boot(raft_config, Some(()), Some(()), Some(())).await?;
```
ariesdevil, line 105: If we should check `_is_open` is true to ensure metanode is available?

drmingdrmer: `_is_open` indicates whether it is opened from a previous state, not created.

If it does not return an error, meta node is definitely available.

ariesdevil: Got it.

common/meta/raft-store/src/config.rs
```diff
@@ -111,6 +111,12 @@ pub struct RaftConfig {
     )]
     pub single: bool,
 
+    /// Bring up a metasrv node and join a cluster.
+    ///
+    /// The value is one or more addresses of a node in the cluster, to which this node sends a `join` request.
+    #[structopt(long, env = "METASRV_JOIN")]
```
BohuTANG, line 117: It's would be nice to add a `help` message to tell the user how to use the `join`, I guess the need indeed.

drmingdrmer: That's the magic! `structopt` captures the doc comment and builds help message for it! :  
```
        --join <join>...
            Bring up a metasrv node and join a cluster [env: METASRV_JOIN=]
```

ariesdevil: If the user inputs multiple addresses, the help message should print an input format hint, like using comma separation:)

PR #2712, opened 2021-11-08
[metasrv] refactor: use RaftRequest as a request type, use RaftReply as reply type for raft communication

drmingdrmer:

## Changelog

- Improvement

## Related Issues


drmingdrmer:
Two tests hangs for a night. Just refactoring changes. merge.

PR #2717, opened 2021-11-09
[CI] remove unstable stateful tests

ZhiHanZ:

## Summary

will reintroduce it once runner run on public cloud

## Changelog

- Bug Fix

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


scripts/ci/ci-run-stateful-tests-cluster.sh
```diff
@@ -30,14 +30,14 @@ else
     exit 1
 fi
 
-./target/release/bendctl --databend_dir ./.databend --group local query ./tests/suites/0_stateful/ddl/create_table.sql
+./target/debug/bendctl --databend_dir ./.databend --group local query ./tests/suites/0_stateful/ddl/create_table.sql
```
BohuTANG, line 33: Why change to `debug`?

ZhiHanZ: to reduce ci compile time，similar to stateless tests

Makefile
```diff
@@ -68,6 +68,8 @@ cross-compile-release:
 cli-build:
 	bash ./scripts/build/build-cli.sh build-cli
 
+cli-build-debug:
+	bash ./scripts/build/build-cli.sh build-cli-debug
```
BohuTANG, line 72: It's better to make a break line

scripts/ci/ci-run-stateful-tests-cluster.sh
PsiACE: How long does it take to run the stateful test in debug mode? Is the total time (compile + run) actually reduced?

BohuTANG: The debug mode in loading is slower than release, bendctl release compile time is more but the load data is less, I prefer to use release to do the stateful test.

ZhiHanZ: I shall use release mode once deployed runner into cloud service

drmingdrmer:
What does the `unstable` mean?
Is there any problem with these two removed test?

ZhiHanZ:
> What does the `unstable` mean? Is there any problem with these two removed test?

we are planning to use self-hosted runner to run stateful test, and deploy self hosted runner on EKS. self hosted runner is not ready, but we could run stateful test from local

Issue #2695, opened 2021-11-08
Failed to create users with different hostnames


Junnplus:
**Summary**
Failed to create users with different hostnames.

**How to reproduce**

```
mysql> SHOW USERS;
+------+-----------+----------+-----------+
| name | hostname  | password | auth_type |
+------+-----------+----------+-----------+
| test | localhost | password |         3 |
+------+-----------+----------+-----------+
1 row in set (0.00 sec)
Read 1 rows, 70 B in 0.000 sec., 2.01 thousand rows/sec., 140.85 KB/sec.

mysql> CREATE USER 'test'@'127.0.0.1' IDENTIFIED BY 'password';
ERROR 1105 (HY000): Code: 3001, displayText = User already exists, seq [1634](while add user)..
```

**Error message and/or stacktrace**

**Additional context**

Junnplus:
Should we use like `__fd_users/test/localhost` as the key in [add_user](https://github.com/datafuselabs/databend/blob/main/common/management/src/user/user_mgr.rs#L52) to support create users with different hostnames.

BohuTANG:
Account name syntax is `'user_name'@'host_name'`, the `@'host_name'` part is optional.
So, `test@localhost` and `test@127.0.0.1` is the different account, we should use both `'user_name'@'host_name'` as the key not `user_name` only.

PR #2727, opened 2021-11-09
[user] fix create users with different hostnames

Junnplus:

## Summary

Fix create users with different hostnames.

## Changelog

- Bug Fix

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/2695

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/management/src/user/user_mgr.rs
+++ b/common/management/src/user/user_mgr.rs
@@ -84,27 +84,19 @@ impl UserMgrApi for UserMgr {
         let get_kv = async move { kv_api.get_kv(&key).await };
         let res = get_kv.await?;
         let seq_value =
-            res.ok_or_else(|| ErrorCode::UnknownUser(format!("unknown user {}", username)))?;
+            res.ok_or_else(|| ErrorCode::UnknownUser(format!("unknown user {}", user_key)))?;
 
         match MatchSeq::from(seq).match_seq(&seq_value) {
             // Ok(_) => Ok(SeqV::new(seq_value.seq, seq_value.data.try_into()?)),
             Ok(_) => Ok(seq_value.into_seqv()?),
-            Err(_) => Err(ErrorCode::UnknownUser(format!("username: {}", username))),
+            Err(_) => Err(ErrorCode::UnknownUser(format!("unknown user {}", user_key))),
         }
     }
 
-    async fn get_users(&self, username: Option<String>) -> Result<Vec<SeqV<UserInfo>>> {
-        let key = if username.is_some() {
-            format!(
-                "{}/{}",
-                self.user_prefix,
-                format!("'{}'@", username.unwrap())
-            )
-        } else {
-            self.user_prefix.clone()
-        };
+    async fn get_users(&self) -> Result<Vec<SeqV<UserInfo>>> {
+        let user_prefix = self.user_prefix.clone();
         let kv_api = self.kv_api.clone();
-        let prefix_list_kv = async move { kv_api.prefix_list_kv(&key).await };
+        let prefix_list_kv = async move { kv_api.prefix_list_kv(user_prefix.as_str()).await };
         let values = prefix_list_kv.await?;
 
         let mut r = vec![];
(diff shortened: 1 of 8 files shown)


common/management/src/user/user_mgr.rs
```diff
@@ -80,10 +87,18 @@ impl UserMgrApi for UserMgr {
         }
     }
 
-    async fn get_users(&self) -> Result<Vec<SeqV<UserInfo>>> {
-        let user_prefix = self.user_prefix.clone();
+    async fn get_users(&self, username: Option<String>) -> Result<Vec<SeqV<UserInfo>>> {
```
BohuTANG, line 90: get_users aims to get all the users prefix a tenant, why we add a username here?

BohuTANG: `get_users` is for a tenant to get all his users.
In databend, the first level is a tenant, then a tenant may has many users for queries.

Junnplus: > `get_users` is for a tenant to get all his users.

Yes, we can get all users under a tenant through `get_users(None)`.

Or, we need to separate `get_users` and `get_users_by_name` ?

BohuTANG: Thanks for explaintion.
`get_users` add the `Option` is for test only? Since we have a `get_user` api with username already.

Junnplus: Sorry, I think I might not describe it clearly.

This pr use both 'user_name'@'host_name' as the key, `get_user` aims to get one user by username and hostname, `get_users` for list users by username, get all user if username is None.

There are two scenarios for `get_users`:
1. list all user for a tenant, as as `show users`;
2. list users by filter username to check client address in login. For example, the username has two hostnames `localhost` and `%`, we should check whether the user's client address matches the hostname.

BohuTANG: >list all user for a tenant, as as show users;


tenant is configed in the query/config already, and init in:
https://github.com/datafuselabs/databend/blob/0d5fc199c0834aa11ccb2b115d561822d4df208b/common/management/src/user/user_mgr.rs#L42
`get_users()` defaults to get all the users under the tenant

> list users by filter username to check client address in login. For example, the username has two hostnames localhost and 192.168.0.0/16, we should check whether the user's client address matches the hostname.


We should check : <username, hostname> togther, no need to change the `get_users` api to get users who prefix with username first then check the hostname,  change this api looks ok:
```
get_user(&self, user: &str, hostname: &str) -> Result<UserInfo>
```
to
```
get_user(&self, user: &str, hostname: &str) -> Result<Option<UserInfo>>
```

If it is None -- No users+hostname found.

Junnplus: > On balance, we don't need to modify the get_users api.

I will revoke the changes to get_users first.

Junnplus: > We should check : <username, hostname> togther, no need to change the get_users api to get users who prefix with username first then check the hostname, get_user method works for your case

I have a confusion, if the user's client address is `192.168.10.100`, how to match the above two hostnames `localhost` and `192.168.0.0/16` through get_user.

BohuTANG: Aha, finally got your points, the host name can take many forms, and wildcards! Thank you @Junnplus 

The get_user api change looks make sense

Junnplus: > Aha, finally got your points, the host name can take many forms, and wildcards!
😅 

> But we can wild them in the get_user(First, get the users prefixed the username, and regex the hostname from the users) api and not changed the get_users api.

Or we can add get_prefix_users instead of get_user for filter username.

BohuTANG: That's 👌

Junnplus: Thanks @BohuTANG , We can seperate a PR to support authentication.

Are there any other questions for this PR?

PR #2743, opened 2021-11-11
Add monotonicity check visitor

### LINKED ISSUE #2343: Add monotonic check in FunctionFactory

junli1026:

## Summary

Add monotonicity visitor for extracting the column from monotonic function. For example 'order by number+3' can be optimized to 'order by number', because 'number+3' -> f(x) = x+3    is a monotonic function.

The new added class is  applied for top-n optimizer in the PR. Thus query like "select number*number from numbers(100000000000) order by number+number+12 limit 10" will be optimized to only fetch top 10 rows -- will be fast.
```

mysql> select number*number from numbers(100000000000) order by number+number+12 limit 10;
+-------------------+
| (number * number) |
+-------------------+
|                 0 |
|                 1 |
|                 4 |
|                 9 |
|                16 |
|                25 |
|                36 |
|                49 |
|                64 |
|                81 |
+-------------------+
10 rows in set (0.02 sec)
Read 80 rows, 640 B in 0.007 sec., 11.78 thousand rows/sec., 94.27 KB/sec.

```

It can also be used for expression rewriting logic --rewrite the expression  at the very beginning of whole pipeline. It is not covered in this PR.

## Changelog

- New Feature

## Related Issues

Fixes #2343

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/function.rs
+++ b/common/functions/src/scalars/function.rs
@@ -18,8 +18,52 @@ use common_datavalues::columns::DataColumn;
 use common_datavalues::prelude::DataColumnsWithField;
 use common_datavalues::DataSchema;
 use common_datavalues::DataType;
+use common_datavalues::DataValue;
 use common_exception::Result;
 use dyn_clone::DynClone;
+
+// Represents the node of function tree for calculating monotonicity.
+// For example, a function of Add(Neg(number), 5) will have a tree like this:
+//
+// .                   MonotonicityNode::Function --> Add
+//                         /                       \
+//                        /                         \
+//      MonotonicityNode::Function --> Neg        Monotonicity::Constant --> 5
+//                     /
+//                    /
+//     MonotonicityNode::Variable --> number
+//
+// The structure of the tree is basically the structure of the expression.
+// Simple depth first search visit the expression tree and gete monotonicity from
+// every function. Each function is responsible to implement its own monotonicity
+// function.
+// Notice!! the mechanism doesn't solve multiple variables case.
+#[derive(Clone)]
+pub enum MonotonicityNode {
+    Function(Monotonicity),
+    Constant(DataValue),
+    Variable(String),
+}
+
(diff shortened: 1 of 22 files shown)


junli1026:
Maybe I should move the optimization to plan_expression_common, just rewrite the expression at the very beginning.

common/functions/src/scalars/function.rs
```diff
@@ -34,6 +78,11 @@ pub trait Function: fmt::Display + Sync + Send + DynClone {
         None
     }
 
+    // return monotonicity node, should always return MonotonicityNode::Function
+    fn get_monotonicity(&self, _args: &[MonotonicityNode]) -> Result<MonotonicityNode> {
```
sundy-li, line 82: Did you consider key ranges?

function `sin` is monotonic in all ranges.

common/functions/src/scalars/function.rs
```diff
@@ -18,9 +18,53 @@ use common_datavalues::columns::DataColumn;
...
+    pub is_positive: bool,
+    // Is true if function is monotonic on the whole input range
+    pub is_always_monotonic: bool,
+}
+
+impl Monotonicity {
+    pub fn default() -> Self {
+        Monotonicity {
+            is_monotonic: false,
+            is_positive: false,
```
zhyass, line 62: `is_positive` default is true

zhang2014 (approved):

sundy-li (approved):

PR #2759, opened 2021-11-12
[function] Add ceil, ceiling, floor func

kyon0304:

## Summary

Add ceil, ceiling, floor func
```sql
mysql> select ceil(123.6);
+-------------+
| ceil(123.6) |
+-------------+
|         124 |
+-------------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.002 sec., 477.44 rows/sec., 477.44 B/sec.

mysql> select ceil(123);
+-----------+
| ceil(123) |
+-----------+
|       123 |
+-----------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.003 sec., 301.25 rows/sec., 301.25 B/sec.

mysql> select ceiling('-12.3aa1');
+---------------------+
| ceiling('-12.3aa1') |
+---------------------+
|                 -12 |
+---------------------+
1 row in set (0.00 sec)
Read 1 rows, 1 B in 0.000 sec., 2.26 thousand rows/sec., 2.26 KB/sec.

mysql> select ceiling('+12.3aa1');
+---------------------+
| ceiling('+12.3aa1') |
+---------------------+
|                   0 |
+---------------------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.000 sec., 2.29 thousand rows/sec., 2.29 KB/sec.

mysql> select ceiling('a123');
+-----------------+
| ceiling('a123') |
+-----------------+
|               0 |
+-----------------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.000 sec., 2.44 thousand rows/sec., 2.44 KB/sec.

mysql> select ceiling('a');
+--------------+
| ceiling('a') |
+--------------+
|            0 |
+--------------+
1 row in set (0.00 sec)
Read 1 rows, 1 B in 0.000 sec., 3.2 thousand rows/sec., 3.2 KB/sec.

mysql> select floor(true);
ERROR 1105 (HY000): Code: 7, displayText = Expected numeric types, but got Boolean.
```

## Changelog

- New Feature

## Related Issues

#2563 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/maths/ceil.rs
+++ b/common/functions/src/scalars/maths/ceil.rs
@@ -1,0 +1,97 @@
+// Copyright 2020 Datafuse Lfloor.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt;
+
+use common_datavalues::prelude::ArrayApply;
+use common_datavalues::prelude::DataColumn;
+use common_datavalues::prelude::DataColumnsWithField;
+use common_datavalues::DataSchema;
+use common_datavalues::DataType;
+use common_exception::ErrorCode;
+use common_exception::Result;
+
+use crate::scalars::function_factory::FunctionDescription;
+use crate::scalars::function_factory::FunctionFeatures;
+use crate::scalars::Function;
+
+#[derive(Clone)]
+pub struct CeilFunction {
+    display_name: String,
+}
+
+impl CeilFunction {
+    pub fn try_create(display_name: &str) -> Result<Box<dyn Function>> {
+        Ok(Box::new(CeilFunction {
+            display_name: display_name.to_string(),
+        }))
+    }
+
+    pub fn desc() -> FunctionDescription {
+        FunctionDescription::creator(Box::new(Self::try_create))
+            .features(FunctionFeatures::default().deterministic())
+    }
+}
+
+impl Function for CeilFunction {
+    fn name(&self) -> &str {
+        &*self.display_name
+    }
+
+    fn num_arguments(&self) -> usize {
+        1
+    }
+
+    fn return_type(&self, _args: &[DataType]) -> Result<DataType> {
+        Ok(DataType::Float64)
+    }
+
+    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+        Ok(false)
+    }
+
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        match columns[0].data_type() {
+            DataType::UInt8
+            | DataType::UInt16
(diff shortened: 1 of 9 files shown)


common/functions/src/scalars/maths/ceil.rs
```diff
@@ -0,0 +1,97 @@
...
+            DataType::UInt8
+            | DataType::UInt16
+            | DataType::UInt32
+            | DataType::UInt64
+            | DataType::Int8
+            | DataType::Int16
+            | DataType::Int32
+            | DataType::Int64
+            | DataType::Float32
+            | DataType::Float64 => {
```
sundy-li, line 75: Let's add String support.

https://www.db-fiddle.com/f/wasupqK75AdxQN5PUq8v2z/0

kyon0304: done: 68c2c81f23578fda233121e926634ba4c4790449

common/functions/src/scalars/maths/ceil.rs
```diff
@@ -0,0 +1,115 @@
...
+                let result = columns[0]
+                    .column()
+                    .to_minimal_array()?
+                    .cast_with_type(&DataType::Float64)?
+                    .f64()?
+                    .apply_cast_numeric(|v| v.ceil());
+                let column: DataColumn = result.into();
+                Ok(column)
+            }
+            DataType::String => {
```
sundy-li, line 87: We don't need this regex, just using cast_with_type is ok, it's already using partial_cast.

common/functions/src/scalars/maths/floor.rs
```diff
@@ -0,0 +1,115 @@
...
+                let result = columns[0]
+                    .column()
+                    .to_minimal_array()?
+                    .cast_with_type(&DataType::Float64)?
+                    .f64()?
+                    .apply_cast_numeric(|v| v.floor());
+                let column: DataColumn = result.into();
+                Ok(column)
+            }
+            DataType::String => {
```
sundy-li, line 87: .

common/functions/src/scalars/maths/ceil.rs
kyon0304: I think cast_with_type doesn't handle string casting as expected, `select ceil('a')` gives NULL instead of 0 for example.

kyon0304: Is there a way to customize the cast result?

sundy-li: Ok,seems there is a bug from Binary to float64,  let's still use `cast_with_type` and make another pr to fix this bug.

```
mysql> select cast('a' as signed),  cast('a' as Float64), cot('a'), sin('a');
+--------------------+----------------------+----------+----------+
| cast('a' as Int64) | cast('a' as Float64) | cot('a') | sin('a') |
+--------------------+----------------------+----------+----------+
|                  0 |                 NULL |     NULL |     NULL |
+--------------------+----------------------+----------+----------+
1 row in set (0.02 sec)
```

sundy-li: > Is there a way to customize the cast result?


This is because [lexical_core](https://docs.rs/lexical-core/0.8.2/lexical_core/) partial_cast 'a' as Err.

Or we can quickly fix this in `cast.rs` and add a test case for this issue.
```
impl ArrayCast for DFStringArray {
    fn cast_with_type(&self, data_type: &DataType) -> Result<Series> {
        if data_type == &DataType::Float64 || data_type == &DataType::Float32 {
             // cast array into int64
            // cast array into float64
        }
        cast_ca(&self.array, data_type)
    }
}

```

PR #2761, opened 2021-11-13
[metasrv] impl Stoppable for meta/FlightServer

drmingdrmer:

##### [metasrv] impl Stoppable for meta/FlightServer
- Feature: FlightServer::start() blocks until listening is estabelisthed.

- Refactor: FlightServer::start() takes a ref to self instead of an
  owned instance, to let user be able to call FlightServer::stop().

- Feature: impl force/graceful shutdown for FlightServer.

- Refactor: simplify FlightServer life cycle mng in tests.

## Related Issues

### DIFF
--- a/metasrv/src/api/flight_server.rs
+++ b/metasrv/src/api/flight_server.rs
@@ -19,10 +19,12 @@ use common_base::tokio;
 use common_base::tokio::sync::oneshot;
 use common_base::tokio::sync::oneshot::Receiver;
 use common_base::tokio::sync::oneshot::Sender;
+use common_base::tokio::task::JoinHandle;
+use common_base::Stoppable;
 use common_exception::ErrorCode;
-use common_exception::ToErrorCode;
 use common_tracing::tracing;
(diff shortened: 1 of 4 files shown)


drmingdrmer:
@sunisdown

BohuTANG:
/lgtm
Neat man!!

dantengsky (approved):

BohuTANG:
From the stateless test, there are many INFO level logs prints, maybe we should change the metasrv log level to ERROR at:https://github.com/datafuselabs/databend/blob/main/scripts/deploy/databend-query-cluster-3-nodes.sh#L12-L20

BohuTANG:
Oops, the stateless cluster tests always failed(I have re-run the job, but it don't help)

sunisdown:
🚀  bro, how do you write Rust code so quickly and good...

drmingdrmer:
> 🚀 bro, how do you write Rust code so quickly and good...

Because Rust just saves a lot time debugging by making everything right at compilation time? 🤔

Issue #2764, opened 2021-11-13
Issue with fsync on mac in stateless-cluster test


drmingdrmer:
There is a known issue with fsync on mac: 
- https://github.com/datafuselabs/databend/issues/1230

We already have disabled fsync to work around this issue, at two places:
- For unit test:
https://github.com/datafuselabs/databend/blob/a598d928d50644ee0bbddd8c9b36d837b1c7044a/metasrv/src/tests/service.rs#L71-L80

- and for embedded metasrv:
https://github.com/datafuselabs/databend/blob/a598d928d50644ee0bbddd8c9b36d837b1c7044a/common/meta/embedded/src/meta_embedded.rs#L50-L58

The stateless cluster test starts metasrv in a normal mode, thus the flushing log etc takes a very long time, which is very likely to fail the tests.

https://github.com/datafuselabs/databend/runs/4197636682?check_suite_focus=true

Maybe we have to disable the fsync everywhere to let these tests pass smoothly. 🤔

drmingdrmer:
By always disabling fsync on mac, there will be risk of data loss.

drmingdrmer:
@BohuTANG @ZhiHanZ

BohuTANG:
> By always disabling fsync on mac, there will be risk of data loss.

Hmm, I understand that on mac systems we are only in a development and experience environment, not a production environment, and this is acceptable.
Which config should we set to make it happy for the mac?

drmingdrmer:
> Which config should we set to make it happy for the mac?

Just simply add this snippet to `MetaNodeBuilder::build()`.

```rust
    if cfg!(target_os = "macos") {
        tracing::warn!("Disabled fsync for meta data tests. fsync on mac is quite slow");
        config.raft_config.no_sync = true;
    }
```

BohuTANG:
> > Which config should we set to make it happy for the mac?
> 
> Just simply add this snippet to `MetaNodeBuilder::build()`.
> 
> ```rust
>     if cfg!(target_os = "macos") {
>         tracing::warn!("Disabled fsync for meta data tests. fsync on mac is quite slow");
>         config.raft_config.no_sync = true;
>     }
> ```

Just do it!

drmingdrmer:
It sounds nike to me~

PR #2768, opened 2021-11-13
[metasrv] test: always disable fsync on mac, to let integration test pass on mac VM

drmingdrmer:

##### [metasrv] test: always disable fsync on mac, to let integration test pass on mac VM
- fix: #2764

## Changelog

- Improvement
- Build/Testing/CI

## Related Issues


drmingdrmer:
@ZhiHanZ 
This may be the cause to what you've mentioned before that cluster-test runs very slow. 🤔

PR #2789, opened 2021-11-15
make partial cast to float works

sundy-li:

## Summary

Summary about this PR

- make partial cast to float works.   Continue #2759
- improve math functions and tests, pre check the datatypes without evaluating the columns.

## Changelog

- Improvement
 
## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/maths/trigonometric.rs
```diff
@@ -63,8 +64,22 @@ impl Function for TrigonometricFunction {
         1
     }
 
-    fn return_type(&self, _args: &[DataType]) -> Result<DataType> {
-        Ok(DataType::Float64)
+    fn return_type(&self, args: &[DataType]) -> Result<DataType> {
+        if is_numeric(&args[0]) || args[0] == DataType::String {
+            Ok(match &args[0] {
+                DataType::Int8 => DataType::UInt8,
```
flaneur2020, line 70: maybe the return value of trigonometric functions are always float64? sin(1) -> 0.8414709848078965

sundy-li: Yes，a mistake..

PR #2800, opened 2021-11-15
Adjust response of /v1/query.

youngsofun:

## Summary

1. remove request_error  and use HTTP status codes like 404  with text messages for these errors.
2. extend error, include fields of ErrorCode
3. move progress into stats
4. rename delete_uri  to  final_uri to be in concert with next_uri, let the server decide what to do.
     5.  user SHOULD get final_uri if next_uri is null(all fetched) or does not need the remaining data. 
     6. serve as ack for the last page. (other page is acked by next_uri)

## Related Issues

/v1/query was introduced in  https://github.com/datafuselabs/databend/pull/2688

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/servers/http/v1/http_query_handlers.rs
+++ b/query/src/servers/http/v1/http_query_handlers.rs
@@ -22,6 +22,9 @@ use common_datavalues::DataSchemaRef;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use hyper::http::header;
+use poem::error::Error as PoemError;
+use poem::error::NotFound;
+use poem::error::Result as PoemResult;
 use poem::get;
 use poem::http::StatusCode;
 use poem::post;
@@ -36,7 +39,9 @@ use serde::Deserialize;
 use serde::Serialize;
 
 use crate::servers::http::v1::block_to_json::JsonBlockRef;
+use crate::servers::http::v1::query::execute_state::ExecuteStateName;
 use crate::servers::http::v1::query::execute_state::HttpQueryRequest;
+use crate::servers::http::v1::query::http_query::HttpQuery;
 use crate::servers::http::v1::query::http_query::HttpQueryResponseInternal;
 use crate::servers::http::v1::query::result_data_manager::Wait;
 use crate::sessions::SessionManagerRef;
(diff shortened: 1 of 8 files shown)


flaneur2020:
how should i use the `final_uri` field? q.q

query/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -181,15 +201,15 @@ async fn query_cancel_handler(
 async fn query_state_handler(
     sessions_extension: Data<&SessionManagerRef>,
     Path(query_id): Path<String>,
-) -> impl IntoResponse {
+) -> PoemResult<QueryResponse> {
```
flaneur2020, line 204: the `PoemResult` here is much cooler than the `impl IntoResponse`!

youngsofun:
@flaneur2020  updated, maybe I should add this as comments.

> rename delete_uri to final_uri to be in concert with next_uri, let the server decide what to do.
5. user SHOULD get final_uri if next_uri is null(all fetched) or does not need the remaining data.
6. serve as ack for the last page. (other page is acked by next_uri)

flaneur2020:
> user SHOULD get final_uri if next_uri is null(all fetched) or does not need the remaining data.

how the server determines the `next_uri` field will be null or not? q.q

youngsofun:
> how the server determines the `next_uri` field will be null or not? q.q

if all tail data is fetched in the current response, set next_uri to null

flaneur2020:
is it ok to only one `next_uri` with a cancel api? q.q

in the client side's view, there're two cases:

- the normal case: client keep iterating on the `next_uri` until all the data has been received, if the `next_uri` is null, it stops the iteration, and the server side releases the resources on the query.
- the quit-in-the middle case: the user found she has already get what she want in the first 10 rows, so she could quit the query in the middle, so the client can fire an cancel/delete request, like https://docs.snowflake.com/en/developer-guide/sql-api/reference.html#post-api-statements-statementhandle-cancel in snowflake, and the DELETE verb in presto

youngsofun:
@flaneur2020   

> is it ok to only one next_uri with a cancel api? q.q

good idea， a few small problems：

1. we still need a field to tell the user that all data is fetched
2. not explicit compared with final_url
3. reduce protocal expansibility:
   1.  user make a URL based on a url:  <next_url>/cancel,  but  next_url may be `xxx?a=1`
   2. use Delete is good, but limited to only one possible VERB,  maybe we later need delete、 cancel、and so on

flaneur2020:
> we still need a field to tell the user that all data is fetched

user could get the data fetch is completed by checking next_url is null? q.q

flaneur2020:
> user make a URL based on a url: <next_url>/cancel, but next_url may be xxx?a=1

i think we can cancel the query by a concrete api, not necessarily `<next_url>/cancel`

in snowflake it's `POST /api/statements/{statementHandle}/cancel`, as long as we know the query id, we can easily cancel it by this api

youngsofun:
the next_url pattern is a little strage， but does minimize the protocol details client need to know，and the server can change without client update, especially useful as a open-source project which may have third party clients based on the http api.

flaneur2020:
/lgtm the code in this PR

through i think there might have better ways to express the `final_uri` logic, quq

youngsofun:
@flaneur2020 It is actually from trino API v1+v2，I think they keep using the way for pagination for a good reason.

youngsofun:
@BohuTANG I think we can merge this merge now,  it is an improvement in many aspects of  #2688(/v1/query),  not a big change of feature,  and it is not just about finan_url, which can be discussed later. So I can make more changes that depend on this pr.

ZhiHanZ (approved):

PR #2821, opened 2021-11-16
ISSUE-2806: Add metrics to dal operations (label with tenant_id and cluster_id) 

### LINKED ISSUE #2806: Add metrics to dal operations (label with  tenant_id and cluster_id)

dantengsky:

## Summary

introduce a delegation of DAL  `DALWithMetric`, which intercepts input/output operations and spills simple statistics (bytes read/write) to metrics components

fixing 
- issue #2806

## Changelog

- Improvement

## Related Issues

Fixes #2806

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1203,6 +1203,7 @@ dependencies = [
  "common-datablocks",
  "common-exception",
  "common-infallible",
+ "common-metrics",
  "futures",
  "metrics",
  "pretty_assertions",
@@ -1212,6 +1213,7 @@ dependencies = [
  "rusoto_s3",
  "serde",
  "serde_json",
+ "tempfile",
 ]
 
 [[package]]
@@ -2128,8 +2130,6 @@ dependencies = [
  "quantiles",
  "rand 0.8.4",
  "reqwest",
- "rusoto_core",
- "rusoto_s3",
  "serde",
  "serde_json",
  "sha1",
--- a/common/dal/Cargo.toml
+++ b/common/dal/Cargo.toml
@@ -15,6 +15,7 @@ common-base = {path = "../base"}
 common-datablocks = {path = "../datablocks"}
 common-exception = {path = "../exception"}
 common-infallible = {path = "../infallible"}
+common-metrics= {path = "../metrics"}
 
 async-compat = "0.2.1"
 async-trait = "0.1"
@@ -32,4 +33,5 @@ reqwest = "0.11"
 [dev-dependencies]
 pretty_assertions = "1.0"
 rand = "0.8.4"
+tempfile = "3.2.0"
 
--- a/common/dal/src/data_accessor.rs
+++ b/common/dal/src/data_accessor.rs
@@ -39,8 +39,6 @@ impl<T> SeekableReader for T where T: Read + Seek {}
 pub trait DataAccessor: Send + Sync {
     fn get_input_stream(&self, path: &str, stream_len: Option<u64>) -> Result<InputStream>;
 
-    async fn get(&self, path: &str) -> Result<Bytes>;
-
     async fn put(&self, path: &str, content: Vec<u8>) -> Result<()>;
 
     async fn put_stream(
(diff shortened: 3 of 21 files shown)


query/src/datasources/common/dal_metrics.rs
```diff
@@ -0,0 +1,137 @@
...
+        Self {
+            tenant_id: tenant_id.into(),
+            cluster_id: cluster_id.into(),
+            inner: input_stream,
+        }
+    }
+}
+
+pub struct DalWrapper {
+    tenant_id: String,
```
BohuTANG, line 50: Seems we can introduce a new struct to `common/metrics/src/recorder.rs` 
```pub struct TenantLabel {
    pub tenant_id:String,
    pub cluster_id:String,
}
```

And here we can use it as:
```
pub struct DalWrapper {
tenant_label: TenantLaber
...
}
```

dantengsky: right!

query/src/datasources/common/dal_metrics.rs
```diff
@@ -0,0 +1,137 @@
...
+    cluster_id: String,
+    inner: Arc<dyn DataAccessor>,
+}
+
+impl DalWrapper {
```
BohuTANG, line 55: `DalWrapper` looks like a strange name and it is rather difficult to understand what it stands for :thinking:

dantengsky: yes, I have the same feeling here. any suggestions?

dantengsky: how about ...  MetricDal... (this is the first name that came up to me, but still "strange" to me)

query/src/datasources/common/dal_metrics.rs
```diff
@@ -0,0 +1,137 @@
...
+use common_dal::AsyncSeekableReader;
+use common_dal::DataAccessor;
+use common_dal::InputStream;
+use common_metrics::label_counter_with_val;
+use futures::Stream;
+
+pub const METRIC_DAL_READ_BYTES: &str = "dal.read.bytes";
+pub const METRIC_DAL_WRITE_BYTES: &str = "dal.write.bytes";
+
+struct InputStreamWrapper {
```
BohuTANG, line 29: InputStreamWrapper -> InputStreamWithMetric?

dantengsky: :ok_hand:

dantengsky: rename DalWrapper to DalWithMetric as well?

query/src/datasources/common/dal_metrics.rs
BohuTANG: MetricDal is ok here :)

query/src/datasources/common/dal_metrics.rs
```diff
@@ -0,0 +1,137 @@
...
+                    super::dal_metrics::METRIC_DAL_WRITE_BYTES,
+                    stream_len as u64,
+                    self.tenant_id.as_str(),
+                    self.cluster_id.as_str(),
+                )
+            })
+    }
+}
+
+impl futures::AsyncRead for InputStreamWrapper {
```
BohuTANG, line 110: Looks like the InputStreamWrapper is a common stream who with metrics.
Would be nice move it to the common/streams as stream_metrics.rs?

BohuTANG: O, the metrics is about DAL.
I think we can make a new file in dal crate called `stream_metrics.rs` out from the dal_metrics file?

dantengsky: :ok:

dantengsky: > O, the metrics(METRIC_DAL_READ_BYTES...) is about DAL. I think we can make a new file in dal crate called `stream_metrics.rs` out from the dal_metrics file?

yeah, I am trying to move both DalWithMetrics and InputStreamWithMetrics into common/dal

dantengsky:
@BohuTANG  PTAL, `DalWithMetric` and `InputstreamWithMetric` have been moved to `common-dal`, any other suggestions / concerns, pls let me know

PR #2830, opened 2021-11-17
Make Fuse Engine as default table engine

sundy-li:
Make Fuse Engine as default table engine

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
/lgtm

Maybe we could introduce a new config `default_table_engine` later

BohuTANG:
Another issue is:
ERROR 1105 (HY000): Code: 7001, displayText = unknown storage scheme [], supported schemes are S3 | Disk.

The config seems not load from default.

sundy-li:
Default storage type should be disk.

PR #2832, opened 2021-11-17
ISSUE-2831: mysql query writer support struct datatype

### LINKED ISSUE #2831: [Improvement] mysql query writer support struct
**Summary**

MySQL client support write struct. write struct as string.

zhyass:

## Summary

MySQL client write struct as string.

## Changelog

- Improvement

## Related Issues

Fixes #2831 

## Test Plan

Stateless Tests


common/datavalues/src/types/serializations/boolean.rs
```diff
@@ -17,33 +17,12 @@ use common_exception::Result;
...
-                        "false".to_owned()
-                    }
-                })
-                .unwrap_or_else(|| "NULL".to_owned())
-            })
-            .collect();
-        Ok(result)
-    }
-
+impl TypeDeserializer for BooleanDeserializer {
```
sundy-li, line 25: Seems `TypeDeserializer` and `TypeSerializer` are both ok, because we have 

`serialize` and `de` (`deserialize`) method, what about making it a new name, like :  `SerDe`

BohuTANG:
@zhyass Myabe we should not use the `force push`, it will over the diff of your patch, just keep the commit is ok, and github will give a `Outdate` hints on the review comments, it's very useful for the next review :D

zhyass:
> @zhyass Myabe we should not use the `force push`, it will over-write the diff of your patch, just keep the commit is ok, and github will give a `Outdate` hints on the review comments, it's very useful for the next review :D

Got it, I just used `force push` when I found some tiny mistake by myself. When fixing the review comments, I always commit separately.

sundy-li (approved):

PR #2833, opened 2021-11-17
Set storage_type to disk by default

BohuTANG:

## Summary

Set storage_type to `disk` by default, make databend-query works with fuse engine wihtout yaml file.

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
Get the config in databend-query:
https://github.com/datafuselabs/databend/blob/18fc42c057cab842b4f83eb90f225028def8a58f/query/src/bin/databend-query.rs#L37

The load_from_args:
https://github.com/datafuselabs/databend/blob/18fc42c057cab842b4f83eb90f225028def8a58f/query/src/configs/config.rs#L87-L93

` Config::from_args()` is init by the StructOpt not the Config::default()

sundy-li (approved):

PR #2847, opened 2021-11-17
[function] add exp math function

Junnplus:

## Summary
```
mysql> select exp(2);
+------------------+
| exp(2)           |
+------------------+
| 7.38905609893065 |
+------------------+
1 row in set (0.00 sec)

mysql> select exp(0);
+--------+
| exp(0) |
+--------+
|      1 |
+--------+
1 row in set (0.01 sec)
```

## Changelog

- New Feature

## Related Issues

https://github.com/datafuselabs/databend/issues/2563

## Test Plan

Unit Tests

Stateless Tests


common/functions/tests/it/scalars/maths/exp.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[test]
+fn test_exp_function() -> Result<()> {
+    struct Test {
+        name: &'static str,
+        display: &'static str,
+        columns: DataColumn,
+        expect: DataColumn,
+        error: &'static str,
+        func: Box<dyn Function>,
```
pymongo, line 27: exp function is a stateless function, you should remove Test struct func to make the test code more clean, then all the test cases use a same ExpFunction

pymongo (changes requested):

common/functions/tests/it/scalars/maths/exp.rs
pymongo: recommend: remove the func field in Test struct

sundy-li: Yes， we can share a single func field in all tests.

common/functions/src/scalars/maths/exp.rs
```diff
@@ -0,0 +1,77 @@
...
+impl Function for ExpFunction {
+    fn name(&self) -> &str {
+        &*self._display_name
+    }
+
+    fn num_arguments(&self) -> usize {
+        1
+    }
+
+    fn return_type(&self, _args: &[DataType]) -> Result<DataType> {
```
sundy-li, line 53: need arg type validator.

Junnplus: @sundy-li https://github.com/datafuselabs/databend/blob/main/common/functions/src/scalars/maths/angle.rs#L66 

Is it necessary to be compatible with the boolean type?

sundy-li: No boolean-type convention for now.

common/functions/src/scalars/maths/exp.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    fn num_arguments(&self) -> usize {
+        1
+    }
+
+    fn return_type(&self, args: &[DataType]) -> Result<DataType> {
+        if args[0] == DataType::Null {
+            Ok(DataType::Null)
+        } else if is_numeric(&args[0]) || args[0] == DataType::String {
+            Ok(DataType::Float64)
```
sundy-li, line 58: Let's always use Let's use `DataType::Float64` to keep the schema stable.

For DataColumn::Null, we will return DataColumn::F64 with all null values.

sundy-li (approved):

PR #2858, opened 2021-11-17
[metasrv] fix: graceful shutdown blocks for ever.

drmingdrmer:

##### [metasrv] fix: graceful shutdown blocks for ever.
- Problem:

The current impl of 2-phase shutdown(graceful then force)
always blocks on `force.recv()`:

```rust
self.shutdown_handler.shutdown(true).await;

if let Some(mut force) = force {
    log::info!("waiting for force");
    let _ = force
        .recv()
        .await
        .expect("Failed to recv the shutdown signal");
    self.shutdown_handler.shutdown(false).await;
```

Thus when a graceful shutdown is actually finished,
it still does not return.

- Solution:

The shutdown handler has to wait for both the graceful shutdown join
handle and the force.recv(),
and receive the first returned one.

- This is one of the causes of that restarting a metasrv cluster does
  not work. Because the running metasrv processes never quit.

## Changelog

- Bug Fix

## Related Issues

### DIFF
--- a/common/base/src/http_shutdown_handlers.rs
+++ b/common/base/src/http_shutdown_handlers.rs
@@ -17,6 +17,7 @@ use std::net::SocketAddr;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_tracing::tracing;
+use futures::future::Either;
 use futures::FutureExt;
 use poem::listener::Acceptor;
 use poem::listener::AcceptorExt;
@@ -27,6 +28,8 @@ use poem::listener::TcpListener;
 use poem::Endpoint;
 use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
+
+use crate::tokio::sync::broadcast;
 
 pub struct HttpShutdownHandler {
     service_name: String,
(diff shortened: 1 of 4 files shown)


ariesdevil:
@youngsofun may be interesting to review this PR.

BohuTANG:
Does this issue exist in databend-query?

drmingdrmer:
> Does this issue exist in databend-query server?

@BohuTANG 
Looks like there is a different issue with `query`, `force shutdown` won't work:
- `shutdown` takes a `&mut self`, it can be called only once before it returns.
- And graceful shutdown blocks for ever, if a graceful shutdown did not finish, there is no way to call it again with `graceful=false`, to trigger a force shutdown.

https://github.com/datafuselabs/databend/blob/76eb8624b005a29a0abc3ebbb43e5a690dc47140/query/src/servers/server.rs#L61-L69

```rust
    async fn shutdown_services(&mut self, graceful: bool) {
        let mut shutdown_jobs = vec![];
        for service in &mut self.services {
            shutdown_jobs.push(service.shutdown(graceful));
        }
        futures::future::join_all(shutdown_jobs).await;
    }
```

```rust
// service.shutdown:
    async fn shutdown(&mut self, graceful: bool) {
        self.shutdown_handler.shutdown(graceful).await;
    }
```

```rust
// shutdown_handler.shutdown:
    pub async fn shutdown(&mut self, graceful: bool) {
        if graceful {
            if let Some(abort_handle) = self.abort_handle.take() {
                let _ = abort_handle.send(());
            }
            if let Some(join_handle) = self.join_handle.take() {
                if let Err(error) = join_handle.await {
                    log::error!(
                        "Unexpected error during shutdown Http Server {}. cause {}",
                        self.service_name,
                        error
                    );
                }
            }
        } else if let Some(join_handle) = self.join_handle.take() {
            join_handle.abort();
        }
    }
```

youngsofun (approved):

BohuTANG:
@drmingdrmer Thank you for your explanation!

youngsofun:
it is a mistake when switch from axum_server to poem,  should not wait on the graceful branch.


before swith:

```
    pub async fn shutdown(&mut self, graceful: bool) {
        if graceful {
            self.abort_handle.graceful_shutdown();
        } else {
            self.abort_handle.shutdown();
            if let Some(join_handle) = self.join_handle.take() {
                if let Err(error) = join_handle.await {
                      ..
                    );
                }
            }
        }
    }

```

and the shutdown  framework in query has a small drawback:   depend on impl of shutdown(force=false) should not block.

PR #2867, opened 2021-11-18
[bugfix] check read_only when creating/dropping database

### LINKED ISSUE #2860: Execution of “create database system” succeeded
**Summary**

I tried to create a database with the same name as the system database, but it was successfully created. 

**How to reproduce**

**Error message and/or stacktrace**

image
image

[How to get the server logs](https://databend.rs/development/how-to-get-server-logs/)

**Additional context**

Veeupup:

## Summary

fix create duplicate database bug.

## Changelog

- Bug Fix

## Related Issues

Fixes #2860 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/catalogs/impls/catalog/metastore_catalog.rs
+++ b/query/src/catalogs/impls/catalog/metastore_catalog.rs
@@ -218,4 +218,17 @@ impl Catalog for MetaStoreCatalog {
 
         Ok(tbl)
     }
+
+    async fn exists_database(&self, db_name: &str) -> Result<bool> {
+        match self.get_database(db_name).await {
+            Ok(_) => Ok(true),
+            Err(err) => {
+                if err.code() == ErrorCode::UnknownDatabaseCode() {
+                    Ok(false)
+                } else {
+                    Err(err)
+                }
+            }
+        }
+    }
 }
--- a/query/src/catalogs/impls/catalog/overlaid_catalog.rs
+++ b/query/src/catalogs/impls/catalog/overlaid_catalog.rs
@@ -190,12 +190,37 @@ impl Catalog for OverlaidCatalog {
         &self,
         plan: CreateDatabasePlan,
     ) -> common_exception::Result<CreateDatabaseReply> {
+        if self.read_only.exists_database(&plan.db).await? {
+            return Err(ErrorCode::DatabaseAlreadyExists(format!(
+                "{} database exists",
+                plan.db
+            )));
+        }
         // create db in BOTTOM layer only
         self.bottom.create_database(plan).await
     }
 
     async fn drop_database(&self, plan: DropDatabasePlan) -> common_exception::Result<()> {
         // drop db in BOTTOM layer only
+        if self.read_only.exists_database(&plan.db).await? {
+            return Err(ErrorCode::UnexpectedError(format!(
+                "user can not drop {} database",
+                plan.db
+            )));
+        }
         self.bottom.drop_database(plan).await
+    }
+
+    async fn exists_database(&self, db_name: &str) -> common_exception::Result<bool> {
+        match self.get_database(db_name).await {
+            Ok(_) => Ok(true),
+            Err(err) => {
+                if err.code() == ErrorCode::UnknownDatabaseCode() {
+                    Ok(false)
+                } else {
+                    Err(err)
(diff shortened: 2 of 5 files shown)


BohuTANG:
Better to add a stateless test here, like:
https://github.com/datafuselabs/databend/blob/9c2e1d606688bbac5134eaaee33dfb3b798dfb8d/tests/suites/0_stateless/05_0001_ddl_create_database.sql#L7-L10

```
CREATE DATABASE system; -- {The ErrorCode } 
DROP DATABASE system; -- {The ErrorCode } 
```

query/src/catalogs/impls/catalog/overlaid_catalog.rs
```diff
@@ -190,6 +190,19 @@ impl Catalog for OverlaidCatalog {
...
     ) -> common_exception::Result<CreateDatabaseReply> {
+        match self.read_only.get_database(&plan.db).await {
+            Ok(_) => {
+                return Err(ErrorCode::DatabaseAlreadyExists(format!(
+                    "{} database exists",
+                    plan.db
+                )));
+            }
+            Err(err) => {
+                if err.code() != ErrorCode::UnknownDatabase("").code() {
```
BohuTANG, line 201: Why we check the `UnknownDatabase` error code here not return the error?

BohuTANG: It seems we need a `self.read_only.exists_database` returns bool or true api here

Veeupup: when creating a DB, if the DB does not exist in read_only, it should be able to create in `bottom` and other errors should be return.

Veeupup: should I add the func `exists_database` in trait `Catalog`?

BohuTANG: Make sense.

Veeupup:
@BohuTANG  I did not make trait func `exists_database` has default implementation because its signature will be
```rust
pub fn exists_database<'life0, 'life1, 'async_trait>(&'life0 self, db_name: &'life1 str) -> ::core::pin::Pin<Box<dyn ::core::future::Future<Output = Result<bool>> + ::core::marker::Send + 'async_trait>>
where
    'life0: 'async_trait,
    'life1: 'async_trait,
    Self: ::core::marker::Sync + 'async_trait
```
the signature has `Self`, and then Catalog will be not object safe.
So I have written some familiar code in different Catalog implements.

query/src/catalogs/impls/catalog/metastore_catalog.rs
```diff
@@ -218,4 +218,17 @@ impl Catalog for MetaStoreCatalog {
 
         Ok(tbl)
     }
+
+    async fn exists_database(&self, db_name: &str) -> Result<bool> {
+        match self.get_database(db_name).await {
+            Ok(_) => Ok(true),
+            Err(err) => {
+                if err.code() == ErrorCode::UnknownDatabase("").code() {
```
BohuTANG, line 226: ErrorCode::UnknownDatabase("").code() -> ErrorCode::UnknownDatabaseCode() is neat

query/src/catalogs/impls/catalog/system_catalog.rs
```diff
@@ -177,4 +177,17 @@ impl Catalog for SystemCatalog {
 
         Ok(table.clone())
     }
+
+    async fn exists_database(&self, db_name: &str) -> Result<bool> {
+        match self.get_database(db_name).await {
+            Ok(_) => Ok(true),
+            Err(err) => {
+                if err.code() == ErrorCode::UnknownDatabase("").code() {
```
BohuTANG, line 185: Same as above :)

query/src/catalogs/impls/catalog/metastore_catalog.rs
Veeupup: resolved

query/src/catalogs/impls/catalog/system_catalog.rs
Veeupup: resloved

PR #2881, opened 2021-11-18
[Functions] Digest functions #2162

### LINKED ISSUE #2162: [Functions] Digest functions
**Summary**

- [ ] MD5
- [ ] SHA1
- [ ] SHA256

mshauneu:

## Summary

Add md5, sha1 and sha2

## Changelog

- New Feature

## Related Issues

Fixes #2162

## Test Plan

Unit Tests

Stateless Tests


website/databend/docs/sqlstatement/hash-functions/md5has.md
```diff
@@ -0,0 +1,33 @@
+---
```
BohuTANG, line 1: Great work!
The document file name is md5hash not md5hash?

mshauneu: Sorry, typos. Fixed.

common/functions/src/scalars/hashes/sha2hash.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        let l_series = columns[1]
+            .column()
+            .to_minimal_array()?
+            .cast_with_type(&DataType::UInt16)?;
+
+        let opt_iter = i_series
+            .string()?
+            .into_no_null_iter()
+            .zip(l_series.u16()?.into_no_null_iter())
```
sundy-li, line 90: I'm afraid `i_series` and `l_series` may not have the same size.

Eg:
`select sha2( toString(number),  256   )  from numbers(10)`


`l_series` is more likely to be a scalar value DataValue::UInt16, we can have a specific path for this case, thus we can avoid a dynamic `match l` inside the loop to improve the performance.

PsiACE:
Nice work, bro.

common/functions/src/scalars/hashes/sha2hash.rs
mshauneu: You are right. Changed code to handle scalar case, e.g: `select sha2( toString(number), 256 ) from numbers(10)` for length of hash2.

sundy-li (approved):

common/functions/src/scalars/hashes/sha2hash.rs
sundy-li: Great, now it's better yet it's not perfect.

```
 let hash = |i: &[u8], l: &u16| match  ...;
 loop {
      hash(x, y)
 }
```

This did not fit in cpu cache.

Better to use 

```
match l {
    a =>  loop {},
    b => loop {},
     ...
}

```

for the constant case.

But it needs much more duplicated codes.

sundy-li: Wow, seems `into_no_null_iter()` did not consider null value case.

PsiACE (approved):
LGTM

PR #2895, opened 2021-11-19
[metasrv] fix: metasrv cluster restart

drmingdrmer:

##### [metasrv] fix: metasrv cluster restart
- Fix: metasrv cluster open mode args:

  - It has to kill query first.
    `query` tries to remove its liveness record from meta before shutting down.
    If meta is stopped, `query` will receive an error that hangs graceful
    shutdown.

  - Fix port assignments and dir config.

  metasrv cluster now is able to restart, but the test with a restarted
  cluster may still fails.
  Because reading on a follower is still inconsistent, there are odds
  `query` can not retrieve a table it just created, etc.

  TODO in next pr: Reading OP on a follower should be forwarded to the leader.

- Fix: upgrade async-raft to a v0.6.2-alpha.14.1:
  A just started follower should not wait too long to elect.
  This issue results in a metasrv cluster hangs for several seconds
  before the first leader is elected.

- Refactor: reduce redundant log.

## Changelog

- Bug Fix

## Related Issues


scripts/deploy/databend-query-cluster-3-nodes.sh
```diff
@@ -5,55 +5,99 @@
 SCRIPT_PATH="$(cd "$(dirname "$0")" >/dev/null 2>&1 && pwd)"
 cd "$SCRIPT_PATH/../.." || exit
 
+# Caveat: has to kill query first.
+# `query` tries to remove its liveness record from meta before shutting down.
+# If meta is stopped, `query` will receive an error that hangs graceful
+# shutdown.
```
BohuTANG, line 11: Agree :100:

PR #2901, opened 2021-11-19
[ISSUE-2894][PATCH-1] Add parser on GRANT .. TO `database`.`table`

### LINKED ISSUE #2894: Fine-grained user privilege on database & table
**Summary**

Currently we have got an UserPrivilege in the metastore, however it do not have the infomation on database & table yet.

We can allowing user grant privileges(like Select / Create) on a database or table.

Related: https://github.com/datafuselabs/databend/pull/2703 https://github.com/datafuselabs/databend/pull/2797

## thoughts on the user grants management

we may add a struct called `GrantSet` or something else, which contains a list of `GrantEntry`, it can be used both in user and role. the `SHOW GRANTS` statement directly dumps the list of the `GrantSet`:

```
mysql> SHOW GRANTS FOR 'jeffrey'@'localhost';
+------------------------------------------------------------------+
| Grants for jeffrey@localhost                                     |
+------------------------------------------------------------------+
| GRANT USAGE ON *.* TO `jeffrey`@`localhost`                      |
| GRANT SELECT, INSERT, UPDATE ON `db1`.* TO `jeffrey`@`localhost` |
+------------------------------------------------------------------+
```

a `GrantEntry` has three levels: Global, Database and Table. the grant level is determined by the `ON` part in the `GRANT` statement.

each `GrantEntry` is identified by the `ON` and `TO` parts in the statement, like ("\*.\*", "jeffery", "localhost"), ("db1.*", "jeffery", "localhost"). if two grant statements have the same identifier, then the privileges should be merged into one `Grant` item:

```
mysql> GRANT SELECT ON *.* TO u1;
mysql> GRANT INSERT ON *.* TO u1;

mysql> SHOW GRANTS FOR 'u1';
+------------------------------------------------------------------+
| Grants for u1                                                    |
+------------------------------------------------------------------+
| GRANT SELECT, INSERT ON *.* TO `u1`                              |
+------------------------------------------------------------------+
```

on checking an operation's privilege, we can get all the current user's grant set from the metasrv, and filter grant set by (user, database/table, hostname), the API might looks like this:

```
// impl Session { 
fn get_effective_grant_set(&self) -> Result<GrantSet>
```

```
// impl GrantSet

fn validate_privilege(&self, db: &str, object: &GrantObject, privilege: &PrivilegeType) -> Result<()>
```

please note that the grant set of the current session CAN NOT directly determined by the current user & hostname, because the active grant set may be changed by executing `USE ROLE` (which limits one user's effective grant set) in the current session.

this API would not change after adding RBAC: what we get is the set of effective grant set of the current user, adding an role to an user would affect this user's grant set at last, the API caller could not care about roles at all. this is why the api in session called `get_effective_grant_set`.

~on `DROP TABLE/DATABASE`, the relevant `GrantEntry` of ALL the users needs get droped altogether. (needs some more investigation on it)~

any suggestion is welcomed!

## Steps

- [x] https://github.com/datafuselabs/databend/pull/2901
- [x] https://github.com/datafuselabs/databend/issues/2779
- [x] https://github.com/datafuselabs/databend/issues/2955
- [x] https://github.com/datafuselabs/databend/issues/3173
- [x] https://github.com/datafuselabs/databend/issues/3193
- [x] https://github.com/datafuselabs/databend/issues/3262
- [x] ~drop grants on database/table is getting dropped~ (seems mysql do not drop the relevant grants after dropping a database)
- [x] https://github.com/datafuselabs/databend/issues/3169
- [x] https://github.com/datafuselabs/databend/issues/3170

flaneur2020:

## Summary

- add parser about grant privilege on `database`.`table`
- add GrantObject in the grant plan
- sanity check on interpreting the grant privilege plan
- test cases on the sanity check

maybe in another PR:

- grant user privilege on database & tables logic in the user manager

## Changelog

- Improvement

## Related Issues

Fixes #2894

## Test Plan

Unit Tests

Stateless Tests


query/src/catalogs/impls/catalog/overlaid_catalog.rs
```diff
@@ -233,4 +233,21 @@ impl Catalog for OverlaidCatalog {
             }
         }
     }
+
+    async fn exists_table(
```
BohuTANG, line 237: How these codes is show in this patch, I rember that it's in another patch

BohuTANG:
There is a suggesstion and a question:

Suggesstion:
It would be nice to do `add an exists_table method to the Catalog trait (failed on code reusing on meeting an warning about` in another patch not this one.

Question:
From my understand, the privileges is grant to the role not the user, in this patch it still grant to user, we will do it in another patch(like the `... create role...` patch) in future?

flaneur2020:
> From my understand, the privileges is grant to the role not the user, in this patch it still grant to user, 

the privileges is both granted to the user and role? 🤔  the privilege one user finally gets is the privileges she directly get granted add the user's roles' privileges

> we will do it in another patch(like the ... create role... patch) in future?

yes, we'd also add the logic about grant the privileges to roles in the future pull requests

query/src/catalogs/impls/catalog/overlaid_catalog.rs
flaneur2020: in the earlier days there's only `exists_database` but no `exists_table` yet, there's another PR working on this? 

if it's true i'd avoid adding exists_table func in this PR to avoid conflicting q.q

BohuTANG:
> > From my understand, the privileges is grant to the role not the user, in this patch it still grant to user,
> 
> the privileges is both granted to the user and role? thinking in mysql the privileges one user finally gets is the privileges she directly get granted add the user's roles' privileges
> 

Privileges are granted to roles, and roles are granted to users.

`the privileges is both granted to the user and role?` -- This is for forward compatibility, they had no role before.






> > we will do it in another patch(like the ... create role... patch) in future?
> 
> yes, we'd also add the logic about grant the privileges to roles in the future pull requests

BohuTANG:
`in mysql the privileges one user finally gets is the privileges she directly get granted add the user's roles' privileges`

Not true.
Thinking a case:
 if a role is dropped, the user role is drop and his pivileges has gone also, not revoke the privileges from the user.

flaneur2020:
> `in mysql the privileges one user finally gets is the privileges she directly get granted add the user's roles' privileges`
> 
> Not true. Thinking a case: if a role is dropped, the user role is drop and his pivileges has gone also, not revoke the privileges from the user.

it's right, when a role is dropped, the user stil has his own privileges left:

consider a scenary like a user has got a role with `CREATE ON db1` privilege, and `SELECT ON db2` which directly granted to the user.

after dropping the role, she stil remains the `SELECT` privilege on `db2`

BohuTANG:
> > `in mysql the privileges one user finally gets is the privileges she directly get granted add the user's roles' privileges`
> > Not true. Thinking a case: if a role is dropped, the user role is drop and his pivileges has gone also, not revoke the privileges from the user.
> 
> it's right, when a role is dropped, the user stil has his own privileges left:
> 
> consider a scenary like a user has got a role with `CREATE ON db1` privilege, and `SELECT ON db2` which directly granted to the user.

If the  `CREATE ON db1` privilege is role1, `SELECT ON db2` is role2, a user can only grant one of them at a time.
> 
> after dropping the role, she stil remains the `SELECT` privilege on `db2`

BohuTANG:
The follow model is more simple for us(ClickHouse did them too, [see](https://clickhouse.com/docs/en/operations/access-rights/)):
Privileges are granted to roles, and roles are granted to users.

```
CREATE USER bohu@'%';
CREATE ROLE systemadmin;
GRANT SELECT ON system.* TO systemadmin;
GRANT systemadmin TO bohu;
```

flaneur2020:
> the privileges is both granted to the user and role? -- This is for forward compatibility, they had no role before.

the user privilege + role privileges is commonly used in the popular DBMS? 🤔 

it's also available in clickhouse: https://clickhouse.com/docs/en/sql-reference/statements/grant/

```
GRANT [ON CLUSTER cluster_name] privilege[(column_name [,...])] [,...] ON {db.table|db.*|*.*|table|*} TO {user | role | CURRENT_USER} [,...] [WITH GRANT OPTION] [WITH REPLACE OPTION]
```

a GRANT statement can works to `user` or `role`. it seems postgres is also the same: https://tableplus.com/blog/2018/04/postgresql-how-to-grant-access-to-users.html

imo RBAC is powerful but it has a bit higher bar than the user privileges, it has its own scenary 🤔

BohuTANG:
For many DBMS, user and role is called object, and the privileges can grant to object, I guess they keep it, but it's more complex.
We can follow the easy way, make it work, make it right :D

flaneur2020:
> For many DBMS, user and role is called object, and the privileges can grant to object, I guess they keep it, but it's more complex. We can follow the easy way, make it work, make it right :D

in the current code base, adding `GRANT` to role needs some works in a sequence of PRs, in my previous thought, we can follow these steps:

1. we already has a code base about grant an user some privilege
2. add fine grained privileges on users
3. the checking privilege logics 
4. allow grant privilege to roles, and grant roles to user
5. allow grant roles to role

user privilege + rbac has more complexity than the RBAC only solution at last, but RBAC has more complexity than the user privilege upfront 🤔

query/src/catalogs/impls/catalog/overlaid_catalog.rs
flaneur2020: I've reverted the changes on the catalog trait, plz help review again

query/src/interpreters/interpreter_grant_privilege.rs
```diff
@@ -51,6 +73,40 @@ impl Interpreter for GrantPrivilegeInterpreter {
         _input_stream: Option<SendableDataBlockStream>,
     ) -> Result<SendableDataBlockStream> {
         let plan = self.plan.clone();
+
+        // *.`table` is not allowed
```
BohuTANG, line 77: case by case match is more complex, but not quite sure if there is a regular check for the pattern?

BohuTANG: Or we can first to check database_pattern and then table_pattern

flaneur2020: it seems we can seperate the grant object (global, database, or table) in the parsing phase

flaneur2020: how about adding a GrantObject enum on this?

it'd has three hands:

- Global
- Database(String) 
- Table(Option<String>, String) // if the database is not specified, taking the current db in session

flaneur2020: @BohuTANG I've refactored out an GrantObject enum, and checked `*.table` as invalid syntax in the parsing phase

please help review again when you get time 🤝

BohuTANG:
I saw some sub-tasks in #2894, it would be better to create a sub-task issue of the main issue #2894 for each PR, the PR will close the sub-task when it closed

I re-open the issue #2894

PR #2927, opened 2021-11-21
ISSUE-2492 prepare to remove BlockingWait

### LINKED ISSUE #2492: find the fundamental reason of issue #2460
issue #2485 is closed, but we have not found the fundamental reason:

**memo**

to reproduce the issue:

1. checkout commits 8fbf98ff7f56741c7dded7fde062d2a9a2090080 (the one before PR #2485)
2. fire up the cluster, e.g. `script/deply/databend-query-standalone.sh`, and do some inserts... see issue #2485

some clues:
 - the hanging thread is doing `std::sync::mpsc::Receiver::recv`
 - for the MySQL session (which works fine), the thread name is None
    but I am not sure if it is a "unnamed" tokio worker thread, or just a sync call stack
 - for the CK session (which hangs), the thread name is `tokio-runtime-worker..`
    this is not a good sign IMO, synchronized `recv` should not be in the async call stack, imo
    but it may not be the root cause, since while CK session executing `read_partitions`, which does synchronized `recv` as well, things doing well.

> PR #2485 fixes this problem, by using async reader in async methods of fuse table (before this PR, during `append`,  sync reader was made use of).
>  
>  BUT, I can not explain 
>  
> 1. why MySQL sessions doing well (without this patch)
> 2. `read_partitions`, which is still a sync method,  doing well for ClickHouse sessions 
> 
> I am still trying to understand the difference in the thread/runtime model between CH and MySQL.
> 

_Originally posted by @dantengsky in https://github.com/datafuselabs/databend/issues/2460#issuecomment-953557632_

zhang2014:

## Summary

Prepare to remove BlockingWait

## Changelog

- Improvement

## Related Issues

fixes #2492

### DIFF
--- a/query/src/catalogs/impls/catalog/metastore_catalog.rs
+++ b/query/src/catalogs/impls/catalog/metastore_catalog.rs
@@ -15,7 +15,6 @@
 
 use std::sync::Arc;
 
-use common_base::BlockingWait;
 use common_dal::InMemoryData;
 use common_exception::ErrorCode;
 use common_exception::Result;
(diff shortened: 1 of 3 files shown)


drmingdrmer:
/lgtm

Great! I can't wait to get it out of the way.

PR #2930, opened 2021-11-21
CATALOG-2925: ImmutableCatalog and MutableCatalog 

BohuTANG:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

Summary about this PR:
1. Strip the `Database` trait from the `Catalog` into a seprate trait, now `Catalog` trait only related to the database operations.
2. Introduce two catalogs for the `Database`: `ImmutableCatalog` and `MutableCatalog`, for each catalog we can build difference database trait object by the engine.

TODO(will do in another patch):
1. Add database register for `ImmutableCatalog` and `MutableCatalog`
2. Remove database name for some table trait methods
3. Add more unit tests

## Changelog

- Improvement

## Related Issues

Fixes #2925

## Test Plan

Unit Tests

Stateless Tests

PR #2932, opened 2021-11-21
[common/meta] refactor: rename get_tables to list_tables

drmingdrmer:

## Changelog

- Improvement

## Related Issues

### DIFF
--- a/common/meta/api/src/meta_api.rs
+++ b/common/meta/api/src/meta_api.rs
@@ -53,7 +53,7 @@ pub trait MetaApi: Send + Sync {
 
     async fn get_table(&self, db: &str, table: &str) -> Result<Arc<TableInfo>>;
 
-    async fn get_tables(&self, db: &str) -> Result<Vec<Arc<TableInfo>>>;
+    async fn list_tables(&self, db: &str) -> Result<Vec<Arc<TableInfo>>>;
 
     async fn get_table_by_id(&self, table_id: MetaId) -> Result<(TableIdent, Arc<TableMeta>)>;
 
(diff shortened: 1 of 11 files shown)


BohuTANG:
/lgtm

Ask a question here:
how can i add a `engine` filed to the `DatabaseInfo` and deal it internal the metasrv:
https://github.com/datafuselabs/databend/blob/9ed709b8e7791074a56678d3d6f8fa5615ce9e0f/common/meta/types/src/database.rs#L18-L22

It  will need at this PR:
https://github.com/datafuselabs/databend/pull/2930/files#diff-a74f09da3b101f6ef49dfbf4c675c0999b28c7d518eeaa57137a22639d90f83cR58

BohuTANG:
Oh, I edited my last approve comments, it merged by the bot :sad

drmingdrmer:
> /lgtm
> 
> Ask a question here: how can i add a `engine` filed to the `DatabaseInfo` and deal it internal the metasrv:
> 
> https://github.com/datafuselabs/databend/blob/9ed709b8e7791074a56678d3d6f8fa5615ce9e0f/common/meta/types/src/database.rs#L18-L22
> 
> It will need at this PR: https://github.com/datafuselabs/databend/pull/2930/files#diff-a74f09da3b101f6ef49dfbf4c675c0999b28c7d518eeaa57137a22639d90f83cR103-R118

As I recalled we had decided to remove db.engine and leave only one database type: the `default` one.
And variant data sources are meant to be implemented at `table` level, i.e., by introducing new `table.engine`s.

Thus `db.engine` is removed.

It can be added back. Not a big deal. 🤔

BohuTANG:
The `engine` here is just used to mark this database type for special scenarios, for the most part we can leave it empty, which I've almost done in #2930 , but this logic is completely different from the previous one we removed.

drmingdrmer:
> The `engine` here is just used to mark this database type for special scenarios, for the most part we can leave it empty, which I've almost done in #2930 , but this logic is completely different from the previous one we removed.

Got it. And we need to save db.engine into metasrv.

PR #2933, opened 2021-11-21
Apply monotonic function to range filter

junli1026:

## Summary

1. Add ranged monotonic check for arithmetic Plus/Minus
2. Add ranged monotonic check for function 'abs'
3. Integrate with range_filter 
4. unit test added to verify expression monotonic.

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/arithmetics/arithmetic.rs
+++ b/common/functions/src/scalars/arithmetics/arithmetic.rs
@@ -21,7 +21,7 @@ use common_datavalues::DataValueArithmeticOperator;
 use common_exception::Result;
 
 use crate::scalars::dates::IntervalFunctionFactory;
-use crate::scalars::function::MonotonicityNode;
+use crate::scalars::function::Monotonicity;
 use crate::scalars::function_factory::FunctionFactory;
 use crate::scalars::ArithmeticDivFunction;
 use crate::scalars::ArithmeticMinusFunction;
@@ -55,32 +55,6 @@ impl ArithmeticFunction {
 
     pub fn new(op: DataValueArithmeticOperator) -> Self {
         ArithmeticFunction { op }
-    }
-
-    pub fn eval_range_boundary(
-        &self,
-        args: &[Option<DataColumnWithField>],
-    ) -> Result<Option<DataColumnWithField>> {
-        // if any argument is None, the new boundary will be unknown thus return None.
-        if args.iter().any(|arg| arg.is_none()) {
-            return Ok(None);
-        }
-
-        let input = args
-            .iter()
-            .map(|arg_opt| arg_opt.clone().unwrap())
-            .collect::<Vec<_>>();
-        let new_data_column = self.eval(&input, 1)?;
-
-        let input_types = input
-            .iter()
-            .map(|arg| arg.data_type().clone())
-            .collect::<Vec<_>>();
-        let new_data_column_field = DataColumnWithField::new(
-            new_data_column,
-            DataField::new("", self.return_type(&input_types)?, false),
-        );
-        Ok(Some(new_data_column_field))
     }
 }
 
@@ -144,7 +118,7 @@ impl Function for ArithmeticFunction {
         Some((1, 2))
     }
 
-    fn get_monotonicity(&self, args: &[MonotonicityNode]) -> Result<MonotonicityNode> {
+    fn get_monotonicity(&self, args: &[Monotonicity]) -> Result<Monotonicity> {
         match self.op {
             Plus => ArithmeticPlusFunction::get_monotonicity(args),
             Minus => ArithmeticMinusFunction::get_monotonicity(args),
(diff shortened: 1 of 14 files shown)


common/functions/src/scalars/function.rs
```diff
@@ -17,20 +17,62 @@ use std::fmt;
...
+
 // Represents the range of data column for monotonic function calculation.
 // For example, f(x) = x+2 where x > 10 should have variable range [10, MAX), and f(x) should
 // have range [12, MAX). The range calculation from [10, MAX) to [12, MAX) should rely on the
 // function's eval method.
 #[derive(Clone)]
 pub struct Range {
-    pub begin: DataColumnWithField, // should have constant DataValue
-    pub end: DataColumnWithField,   // should have constant DataValue
+    pub begin: Option<DataColumnWithField>, // should have constant DataValue
```
zhyass, line 36: May be it's better to replace `DataColumnWithField` with `DataValue`?

common/functions/src/scalars/function.rs
```diff
@@ -55,19 +97,17 @@ pub struct Range {
 #[derive(Clone)]
 pub enum MonotonicityNode {
     // Describe a function monotonicity information for the range of data.
-    Function(Monotonicity, Option<Range>),
-    Constant(DataValue),
-    Variable(String, Option<Range>),
+    Function(Monotonicity, Range),
```
zhyass, line 100: Maybe `Function(Monotonicity, Range)` and `Variable(String, Range)` can be combined into one enum.

zhyass:
Thanks for contributing to databend. I have some ideas.
1. is_always_monotonic cannot be deleted.
2. Range filter only needs to call `check_monotonic` method to obtain the monotonicity of expression. Follow-up processing will refer to [section 5.3.7](http://vldb.org/pvldb/vol14/p3083-edara.pdf). 
e.g. 
`abs(x)>3` 's veriable expression is `(max_x<=0 and abs(min_x)>3) or (min_x>=0 and abs(max_x) >3)`
I am currently trying to do this work.

common/functions/src/scalars/arithmetics/arithmetic_minus.rs
```diff
@@ -34,8 +37,59 @@ impl ArithmeticMinusFunction {
...
+    // get monotonicity from unary negate, like -x should flip the monotonicity and range
+    fn get_unary_operator_monotonicity(node: &MonotonicityNode) -> Result<MonotonicityNode> {
+        let func = ArithmeticFunction::new(DataValueArithmeticOperator::Minus);
+
+        match node {
+            MonotonicityNode::Function(mono, Range { begin, end }) => {
+                if !mono.is_monotonic {
+                    return Ok(node.clone());
+                }
+                let new_begin = func.eval_range_boundary(&[begin.clone()])?;
```
zhyass, line 49: need check is_positive, for example: x: range[1,3]  -x:range[-3,-1]

junli1026:
> Thanks for contributing to databend. I have some ideas.
> 
>     1. `is_always_monotonic` cannot be deleted.
> 
>     2. In current design, range filter only needs to call `check_monotonic` method to obtain the monotonicity of expression. Follow-up processing will refer to [section 5.3.7](http://vldb.org/pvldb/vol14/p3083-edara.pdf).
>        e.g.
>        `abs(x)` is conditional monotonicity, in (-Inf, 0] is {true,false,false}, in [0,+Inf) is {true,true,false}
>        `abs(x)>3` 's veriable expression is `(max_x<=0 and abs(min_x)>3) or (min_x>=0 and abs(max_x) >3)`

For this example, f(x) > 3, is equivlant to 
       f(min_x) > 3 if f(x) is monotonic increasing in [min_x, max_x]
  or  f(max_x) > 3 if f(x) is monotonic decreasing in [min_x, max_x]

The veriable expression should be created differently per block, that is, with [min, max], the false expression should be different.

```rust
            let verifiable_expression = RangeFilter::try_create(&exprs.filters[0], schema)?;
            Box::new(move |v: &BlockStats| verifiable_expression.eval(v))
 ```
I think maybe the  RangeFilter::try_create should have parameter MIN and MAX;


>        I am currently trying to do this work.
>        The  rangefilter is not the final form, maybe there is a more suitable solution
> 
>     3. For range, multi-column maybe be meaningless. For example, f(x)+g(y) can directly return {false, true, false}.
>        Here is the clickhouse code, https://github.com/ClickHouse/ClickHouse/blob/master/src/Functions/FunctionBinaryArithmetic.h#L1754

**Review by zhyass — APPROVED** [review]
Great!

File: common/functions/src/scalars/arithmetics/arithmetic.rs [hunk 4]
```diff
@@ -50,7 +50,37 @@ impl ArithmeticFunction {
...
     pub fn try_create_func(op: DataValueArithmeticOperator) -> Result<Box<dyn Function>> {
-        Ok(Box::new(ArithmeticFunction { op }))
+        Ok(Box::new(ArithmeticFunction::new(op)))
+    }
+
+    pub fn new(op: DataValueArithmeticOperator) -> Self {
+        ArithmeticFunction { op }
+    }
+
+    pub fn eval_range_boundary(
```
zhyass, line 60: `eval_range_boundary`  is useless and needs to be deleted.

Issue #2702, opened 2021-11-08
metasrv: add http api for monitoring etc


drmingdrmer:
Something like v1/cluster to return a list of members

drmingdrmer:
Thank you man for taking this issue!

FYI:
There are two kinds of nodes: voters and non-voters. `MetaRaftStore::list_non_voters()` is a good reference about these concepts.
- Voters AKA `members` can be retrieved with `get_membership_config()`.
- Non-voters are stored in a sled sub tree `StateMachine::nodes()`.

sunisdown:
@drmingdrmer  thanks for the reminder.

I originally thought it was just to show the information, but when I glanced at the code I found that the http server currently can't get the sled instance, and probably needs more code changes than I had expected. the basic idea is:

1 Add a context to the http/rpc/metrics server
2 The context needs to hold the sled (metastore) instance.
3 Read the context metastore data in the http handler and return the result.

do you have any suggestion?

drmingdrmer:
The chain of these components in metasrv is:
```
FlightServer -> MetaNode -> MetaRaftStoe -> StateMachine -> SledTree
```

`membership` is only understood by `MetaRaftStore` and `MetaNode`.
Thus to retrieve a list of voters or non-voters, it's better to get it done in `MetaNode`, instead of accessing SledTree directly.

To your questions:
- The http server needs a reference to `MetaNode`. But for now the creation process for `MetaNode` is inside `FlightServer::serve()`. We need to refactor this part:  move the creation of `MetaNode` out of `FlightServer::serve`.
- Then let `HttpServer` have a clone of `MetaNode` so that it's able to get anything it needs.

sunisdown:
@drmingdrmer  


> * Voters AKA `members` can be retrieved with `get_membership_config()`.
> * Non-voters are stored in a sled sub tree `StateMachine::nodes()`.

Hi bro, May I know where can we get the IP for the voters?

drmingdrmer:
`MetaNode.sto.get_node_addr(id)` is what you want XDDD

sunisdown:
If I understand correctly, then `MetaNode.sto` contains all the nodes, which are divided into two kinds, the `voters` in the `membership_config`, and the `non-vonters` if they are not in the `membership_config`.

drmingdrmer:
[...]

PR #2943, opened 2021-11-21
[feature]: return nodes list in metasrv server

sunisdown:

## Summary

add http api, return the nodes list in the metasrc cluster

## Changelog

- New Feature

## Related Issues

related #2702

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/metasrv/src/api/http/v1/cluster_state.rs
+++ b/metasrv/src/api/http/v1/cluster_state.rs
@@ -1,0 +1,57 @@
+// Copyright 2020 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use common_exception::Result;
+use poem::http::StatusCode;
+use poem::web::Data;
+use poem::web::Html;
+use poem::web::IntoResponse;
+use poem::Response;
+
+use crate::meta_service::MetaNode;
+
+pub struct ClusterTemplate {
+    result: Result<String>,
+}
+
+impl IntoResponse for ClusterTemplate {
+    fn into_response(self) -> Response {
+        match self.result {
+            Ok(nodes) => Html(nodes).into_response(),
+            Err(cause) => Response::builder()
+                .status(StatusCode::INTERNAL_SERVER_ERROR)
+                .body(format!(
+                    "Failed to fetch cluster nodes list. cause: {}",
+                    cause
+                )),
+        }
+    }
+}
+
+// GET /v1/cluster/nodes
+// list all nodes in current databend-query cluster
+// request: None
+// cluster_state: the shared in memory state which store all nodes known to current node
+// return: return a list of cluster node information
+#[poem::handler]
+pub async fn nodes_handler(meta_node: Data<&Arc<MetaNode>>) -> ClusterTemplate {
+    let nodes = meta_node.get_nodes().await.unwrap();
+    // let nodes = meta_node.get_nodes().await.expect("Failed to fetch cluster nodes list");
+    let nodes_list = serde_json::to_string(&nodes).unwrap();
+    ClusterTemplate {
+        result: Ok(nodes_list),
+    }
+}
--- a/metasrv/src/api/http/v1/cluster_state.rs
+++ b/metasrv/src/api/http/v1/cluster_state.rs
@@ -42,14 +42,12 @@ impl IntoResponse for ClusterTemplate {
 }
 
 // GET /v1/cluster/nodes
-// list all nodes in current databend-query cluster
+// list all nodes in current databend-metasrv cluster
 // request: None
-// cluster_state: the shared in memory state which store all nodes known to current node
 // return: return a list of cluster node information
 #[poem::handler]
 pub async fn nodes_handler(meta_node: Data<&Arc<MetaNode>>) -> ClusterTemplate {
     let nodes = meta_node.get_nodes().await.unwrap();
-    // let nodes = meta_node.get_nodes().await.expect("Failed to fetch cluster nodes list");
     let nodes_list = serde_json::to_string(&nodes).unwrap();
     ClusterTemplate {
         result: Ok(nodes_list),
--- a/metasrv/src/api/http/v1/cluster_state.rs
+++ b/metasrv/src/api/http/v1/cluster_state.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Datafuse Labs.
+// Copyright 2021 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


drmingdrmer:
Nice shot! neat and clean!

- I am not familiar with `poem` and have a question about the name `ClusterTemplate`: why call it a `template`? Is it a naming convention with `xxxTemplate`?
- If it is possible, there should be a unit test to this new API.

metasrv/src/api/http/v1/cluster_state.rs
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+}
+
+// GET /v1/cluster/nodes
+// list all nodes in current databend-metasrv cluster
+// request: None
+// return: return a list of cluster node information
+#[poem::handler]
+pub async fn nodes_handler(meta_node: Data<&Arc<MetaNode>>) -> ClusterTemplate {
```
flaneur2020, line 49: maybe it can return an Result<String, PoemError> ?

the error processing might be easier with an `Result`

BohuTANG:
> Nice shot! neat and clean!
> 
> * I am not familiar with `poem` and have a question about the name `ClusterTemplate`: why call it a `template`? Is it a naming convention with `xxxTemplate`?
> * If it is possible, there should be a unit test to this new API.

For the first name question, I ask @sunli829 for you :D

sunli829:
I think ClusterResponse might be a better name.

metasrv/src/api/http/v1/cluster_state.rs
sunli829: @flaneur2020 agree!

sunisdown:
> * I am not familiar with `poem` and have a question about the name `ClusterTemplate`: why call it a `template`? Is it a naming convention with `xxxTemplate`?
I thought it's a `databend` naming style and just followed the variable name in `query`.

sunli829:
I'm sorry that I don't have a computer at hand now. Tomorrow, I will help review all the codes related to Poem and change them to the correct way of use and send a PR.

metasrv/src/api/http/v1/cluster_state.rs
flaneur2020: the rust http library implemented an `IntoResponse` for `Result<T, E>` when `T` and `E` are both implemented `IntoResponse`, like below:

```
impl IntoResponse for Result<T, E> where T: IntoResposne, E: IntoResponse
```

it's already builtin both in axum and poem, do not need us to `impl` this by hand:

https://docs.rs/poem/0.8.0/poem/web/trait.IntoResponse.html#impl-IntoResponse-for-Result%3CT%2C%20E%3E
https://docs.rs/axum/0.1.3/axum/response/trait.IntoResponse.html#impl-IntoResponse-for-Result%3CT%2C%20E%3E

BohuTANG:
Some conflicts, need merge with master :D

sunisdown:
> * If it is possible, there should be a unit test to this new API. 

@drmingdrmer 

Have added UT  for the endpoints.
How can I rerun the macos test in CI? I ran the test locally a few times and it passed.

sunisdown:
@flaneur2020  hi bro, Do you know why this CI keeps failing?
 I have ran the tests in macos and ubuntu and they both passed.

drmingdrmer:
> @flaneur2020 hi bro, Do you know why this CI keeps failing? I have ran the tests in macos and ubuntu and they both passed.

These tests starts multiple tasks and in a CI env there are not enough threads for these tasks thus it is likely to fail due to timeout. 

I've extended the timeout months ago and it works smoothly for a while.

Let have a look what changed in the code base that causes these failure or if it is just caused by an unstable CI env.

metasrv/tests/it/api/http/clsuter_state_test.rs
```diff
@@ -0,0 +1,55 @@
...
+    let (_log_guards, ut_span) = init_meta_ut!();
+    let _ent = ut_span.enter();
+    let tc = new_test_context(0);
+    let meta_node = MetaNode::open_create_boot(&tc.config.raft_config, None, Some(()), None).await?;
+
+    let cluster_router = Route::new().at("/cluster/nodes").get(get!(health_handler)).data(meta_node.clone());
+    let response = cluster_router.call(Request::new(Method::Get, Uri::from_static("/cluster/nodes"))).await?;
+    assert_eq!(response.status(), StatusCode::OK);
+    meta_node.stop().await?;
+    Ok(())
```
drmingdrmer, line 54: Is this test faking a http-service?

I supposed the test should test the service defined in the code. 🤔 
E.g., start a real metasrv(not just meta-node), and send an http request to it.
Then check the response as expected.

To bring a cluster with more than one servers, this is a good reference: 
https://github.com/datafuselabs/databend/blob/46a2e816037c6d76991cec64db70b86cd20d5b83/metasrv/tests/it/flight/metasrv_flight_api.rs#L131-L146

sunisdown: It's not a faking http-services now. 
we can test the response with multiple meta-node, no need to start a real metasrv.

drmingdrmer:
@sunisdown 
>It's not a faking http-services now.
we can test the response with multiple meta-node, no need to start a real metasrv.

The http-service routing is defined in `http_service.rs`:

```
          .at(
                "/v1/cluster/nodes",
                get(super::http::v1::cluster_state::nodes_handler),
            )
            .at(
                "/v1/cluster/state",
                get(super::http::v1::cluster_state::state_handler),
            )
```

While the test builds a new `Router` for testing:
```
 let cluster_router = Route::new()
        .at("/cluster/state", get(state_handler))
        .data(meta_node1.clone());
    let response = cluster_router
        .call(
            Request::builder()
                .uri(Uri::from_static("/cluster/state"))
                .method(Method::GET)
                .finish(),
        )
        .await;
```

These test has to ensure that the new http API works.

Is there other test covers this that I missed?

sunisdown:
> These test has to ensure that the new http API works.
got it.
Do you prefer to test the http service with a http client? 
The test case here only covers the code of the handler logic, not the http server and router. We have other test cases to test the http service. @drmingdrmer

drmingdrmer:
> > These test has to ensure that the new http API works.
> 
> got it. Do you prefer to test the http service with a http client?

I suppose the test could just assumes a `client` works correctly.
And it should only focus on if the two newly added `uri` respond as expected.

metasrv/tests/it/api/http/clsuter_state_test.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use crate::init_meta_ut;
+use crate::tests::service::new_metasrv_test_context;
+use crate::tests::tls_constants::TEST_CA_CERT;
+use crate::tests::tls_constants::TEST_CN_NAME;
+use crate::tests::tls_constants::TEST_SERVER_CERT;
+use crate::tests::tls_constants::TEST_SERVER_KEY;
+
+#[tokio::test]
+async fn test_cluster_nodes() -> common_exception::Result<()> {
```
sundy-li, line 46: there is a typo in file name.

sunisdown: updated

metasrv/tests/it/api/http/cluster_state_test.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        let resp_nodes = client.get(node_url).send().await;
+        assert!(resp_nodes.is_ok());
+        let resp_nodes = resp_nodes.unwrap();
+        assert!(resp_nodes.status().is_success());
+        assert_eq!("/v1/cluster/nodes", resp_nodes.url().path());
+        let result = resp_nodes.json::<Vec<Node>>().await.unwrap();
+        assert_eq!(result.len(), 2);
+    });
+    Ok(())
```
drmingdrmer, line 174: A `task:spawn` will just be dropped without executing it at all.

Adding a `panic!()` in it you will see it still passes.

sunisdown: I finally figured out why the test passed with `tokio::spawn`, thanks. Updated.

drmingdrmer (approved):
/lgtm

Thank you man!

sunisdown:
Hi bro, could you help review @flaneur2020

flaneur2020 (approved):

flaneur2020:
/lgtm 
great work!

PR #2944, opened 2021-11-22
Add binary ops

sundy-li:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


flaneur2020:
/lgtm

the stateless test on the NULL behaviour might need some changes together after we adopt the null semantics, like: log(NULL, 1) -> NULL, log(1, NULL) -> NULL

flaneur2020 (approved):

Issue #2480, opened 2021-10-28
Add `CLI` markdown page to document website


bohutang:
**Summary**

Add a new menu `CLI` to https://databend.rs, goals:
1. how to install bendctl
2. how to create databend cluster from bendctl
3. how to query from bendctl

PR #2408, opened 2021-10-24
[DO NOT MERGE] ISSUE-2378: split `fuse::util` into mods

dantengsky:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

rm `fuse/util`

fixes 
- issue #2378

## Changelog

- Improvement
- Not for changelog (changelog entry is not required)

## Related Issues

Fixes #2378

## Test Plan

Unit Tests

Stateless Tests

PR #2945, opened 2021-11-22
ISSUE-2814: introducing distributed insertion

### LINKED ISSUE #2914: Proposal: supports parallel insertion
**Summary**

As 
- PR #2765 

merged, tables that support the `append` operation can take data in from a datasource (`SourceTransform`).
It would be better if data blocks could be ingested into the table in a distributed manner, sth like this (simplified):

~~~
                                        Append blocks and produces segment info

                    +--------------+      +---------------+
QueryNode_1         |SourceTranform+------>AppendTransform+--+
                    +--------------+      +---------------+  |
                                                             |
                    +--------------+      +---------------+  |
QueryNode_2         |SourceTranform+------>AppendTransform+--+-------+
                    +--------------+      +---------------+  |       |
                                                             |       |
                    +--------------+      +---------------+  |       |
QueryNode_3         |SourceTranform+------>AppendTransform+--+       |
                    +--------------+      +---------------+          |
                                                                     |
                                                                     |
                                                                     |
                                                                     |
                                                                     |
                                                                     |
                                                                     |
QueryNode_N                                                          |
                                                                     |
                                                                     |
                                                                     |
                        +--------------+                             |
                        |MergeProcessor<-----------------------------+
                        +--------------+

         which gathers segments and commits the snapshot
~~~

Also, a brute-force plan for the compaction of fuse table can be rewritten as 

  `replace  t1 as select * from t1 [order by clustering_keys]`

Some basic ideas of impling this feature:

- trait `Table` should provide 
  -  a method, say `append_chunks`,  which append blocks and returns segment infos
  -  a method, say `commit`, which takes table operations in, and try commit the operations (as new snapshot)
      currently, only 'Append' operation supported
- trait `Table` may also provide a method that declares if it supports distributed insertion
- introduce `AppendDataPlan` and `ApendTransform` processor
- modify `pipeline_builder` and `plan_scheduler` accordingly

pls also see @sundy-li in https://github.com/datafuselabs/databend/pull/2765#pullrequestreview-808691793_

dantengsky:

## Summary

- issue #2914

- `Table::append` been split into `append` and `commit`   
   - `append` returns a table specific `SendableDataBlockStream`
   
      for fuse table, a `log` of append operations were encoded into the stream; for other writable tables, returns an empty stream
   
   - `commit` takes the stream returned by append, and interpreter it in a table-specific way 
      
       for fuse table, it would be a stream of segment information,  upon which, the table will try to generate a new snapshot and commit to meta-server.

       for other writable tables, the stream will be ignored (and returns ok)

  `InsertIntoInterpreter` is in charge of feeding upstream values into `Table::append` and putting what append returned into `Table::commit`  
  
-  If applicable, a distributed append schedule will be utilized 

   e.g. statements like `insert into t1 select * from numbers(10000)`, append will be done distributedly(if in cluster mode), but committed by one query node.
   
   but for stmts like `insert into n1 select sum(number) from numbers(1000) group by number %3`, the append will still be done inside one query node.
   
- the main idea of planning goes like this:

  ~~~rust
         // before schedule, and SinkPlan is where table::append happens
          let rewritten_plan = match optimized_plan {
              // if it is a StagePlan Node, we insert the a SinkPlan in between the Stage and Stage.input
              // i.e.
              //    StagePlan <~ PlanNodeA  => StagePlan <~ Sink <~ PlanNodeA
              PlanNode::Stage(r) => {
                  let prev_input = r.input.clone();
                  let sink = PlanNode::Sink(SinkPlan {
                      table_info: table_info.clone(),
                      input: prev_input,
                      cast_needed,
                  });
                  PlanNode::Stage(StagePlan {
                      kind: r.kind,
                      input: Arc::new(sink),
                      scatters_expr: r.scatters_expr,
                  })
              }
              // otherwise, we just prepend a SinkPlan
              // i.e.
              //    node <~ PlanNodeA  => Sink<~ node <~ PlanNodeA
              node => PlanNode::Sink(SinkPlan {
                  table_info: table_info.clone(),
                  input: Arc::new(node),
                  cast_needed,
              }),
          };
  
  ~~~

    but 1) since planning is rather sophisticated, if there were cases that the above arrangement can not be covered, please let me know. 2)  currently, the above logic is embedded in `InsertIntoInterpreter`, if somewhere else is preferred, pls let me know
    
    the following cases have been tested in `standalone` and `cluster-3-nodes` setups:
    
    ~~~sql
    create table n1(a uint64);
    insert into n1 select number from numbers(10000);
    select sum(a) from n1;
    
    CREATE TABLE n2(a UInt64, b UInt64);
    insert into n2 select number, number + 1 from numbers(10000);
    select count(a) from n2;
    select sum(a), sum(b) from n2;
    
    -- "self reference"
    insert into n2 select * from n2;
    select count(a) from n2;
    select sum(a), sum(b) from n2;
    
    -- aggregation
    create table n3(a uint64, b uint64);
    insert into n3 select sum(a), sum(b) from n2;
    select * from n3;
    
    -- cast
    create table s1(a String, b String);
    insert into s1 select number, number + 1 from numbers(10000);
    select sum(cast(a as uint64)), sum(cast(b as uint64)) from s1;

   ~~~
   
   
- misc.

  - about the `ExpressionTransform` that  @sundy-li kindly suggested
  
    in this PR, no explicit expression transform has been arranged, a CastStream is enabled if necessary inside the SinkTransform,  hope this is doing the same thing. 
    
  - @zhang2014 `SelectInterpreter` has been slightly refactored since `InsertIntoInterpreter` reused lots of code from it
   
      some common stuff has been extracted to `plan_scheduler_ext.rs` (any better name?)
    
   - @BohuTANG refactoring of `fuse::util` has been postponed ( to a dedicated PR)
     
      so that this PR might be easier to be reviewed
  
   - `explain pipeline insert ...` does not work yet
  

## Changelog

- Improvement

## Related Issues

Fixes #2914 

## Test Plan

Unit Tests

Stateless Tests


query/src/datasources/table/fuse/util/statistic_helper.rs
```diff
@@ -215,3 +217,29 @@ pub fn merge_stats(schema: &DataSchema, l: &Stats, r: &Stats) -> Result<Stats> {
     };
     Ok(s)
 }
+
+pub fn merge_appends(
```
BohuTANG, line 221: Emm, it seems it's time to move this helper to a folder named like statistics, utilvis a good name for others to exploring :)

dantengsky: agree ，there is a pending pr https://github.com/datafuselabs/databend/pull/2408 ，shall i merge it into this pr？

dantengsky:
btw, although ut and stateless tests passed, there are still something known issues . Trying to fix them

query/src/datasources/table/fuse/util/statistic_helper.rs
dantengsky: Oh, pls do not bother reviewing the pr #2480, forgot that the pr accumulated lots of irrelevant changes. Let me do the refactoring in this pr directly

BohuTANG: Agree.
Sorry I missed there is refine issue #2408 already, we can do the refactoring in that PR

dantengsky: got it

BohuTANG:
Merge first, some comments also welcome...

BohuTANG:
@dantengsky 
I think there are some tasks we need to do with this PR, feel free to open issues to track them as you mentioned in the summary :)

PR #2949, opened 2021-11-22
[function] Add pow math function

Junnplus:

## Summary

```
mysql> SELECT POW(2, 2);
+-----------+
| POW(2, 2) |
+-----------+
|         4 |
+-----------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.000 sec., 3.21 thousand rows/sec., 3.21 KB/sec.

mysql> SELECT POW(2, -2);
+---------------+
| POW(2, (- 2)) |
+---------------+
|          0.25 |
+---------------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.000 sec., 2.25 thousand rows/sec., 2.25 KB/sec.

mysql> SELECT POW(2, NULL);
+--------------+
| POW(2, NULL) |
+--------------+
|         NULL |
+--------------+
1 row in set (0.00 sec)
Read 1 rows, 1 B in 0.000 sec., 3.31 thousand rows/sec., 3.31 KB/sec.
```
## Changelog

- New Feature

## Related Issues

https://github.com/datafuselabs/databend/issues/2563

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/maths/pow.rs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+            _ => {
+                let x_series = x_column.to_minimal_array()?;
+                let y_series = y_column.to_minimal_array()?;
+                let opt_iter = x_series
+                    .f64()?
+                    .into_iter()
+                    .zip(y_series.f64()?.into_iter())
+                    .map(|(x, y)| match (x, y) {
```
sundy-li, line 100: Try to use `binary` function in `ops/arith.rs`, this can reduce the match inside the loop.

common/functions/src/scalars/maths/pow.rs
```diff
@@ -0,0 +1,108 @@
...
+        let x_column: &DataColumn = &columns[0].column().cast_with_type(&DataType::Float64)?;
+        let y_column: &DataColumn = &columns[1].column().cast_with_type(&DataType::Float64)?;
+
+        let result = match (x_column, y_column) {
+            (DataColumn::Array(x_series), DataColumn::Constant(y, _)) => {
+                if y.is_null() {
+                    DFFloat64Array::full_null(input_rows)
+                } else {
+                    let y: f64 = DFTryFrom::try_from(y.clone())?;
+                    let opt_iter = x_series.f64()?.into_iter().map(|x| x.map(|x| x.pow(y)));
```
sundy-li, line 80: Sorry, I forgot to tell you that we can use `ArrayApply`here to improve the performance.

sundy-li: The reason is that we take validity at last to fit the cpu cache.

Junnplus: > The reason is that we take validity at last to fit the cpu cache.

TIL and updated.

sundy-li (approved):

PR #2952, opened 2021-11-22
[Feature]Add table engine config

### LINKED ISSUE #2853: Add table engine enable config
**Summary**

For some table engines, like `CSV/PARQUET/MEMORY` will be disabled in the cloud, users can only use the fuse engine (default).
It will introduce some configs like:
```
enable_table_engine_csv -- default is false
enable_table_engine_parquet -- default is false
enable_table_engine_memory -- default is false
```
These configs are read only(only configured in the yaml file), we can't set them in the session.

GrapeBaBa:

## Summary

Add table engine config

## Changelog

- New Feature

## Related Issues

Fixes #2853 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/configs/config_query.rs
+++ b/query/src/configs/config_query.rs
@@ -41,6 +41,10 @@ const QUERY_RPC_TLS_SERVER_KEY: &str = "QUERY_RPC_TLS_SERVER_KEY";
 const QUERY_RPC_TLS_SERVER_ROOT_CA_CERT: &str = "QUERY_RPC_TLS_SERVER_ROOT_CA_CERT";
 const QUERY_RPC_TLS_SERVICE_DOMAIN_NAME: &str = "QUERY_RPC_TLS_SERVICE_DOMAIN_NAME";
 
+const TABLE_ENGINE_CSV_ENABLED: &str = "TABLE_ENGINE_CSV_ENABLED";
+const TABLE_ENGINE_PARQUET_ENABLED: &str = "TABLE_ENGINE_PARQUET_ENABLED";
+const TABLE_ENGINE_MEMORY_ENABLED: &str = "TABLE_ENGINE_MEMORY_ENABLED";
+
 /// Query config group.
 /// serde(default) make the toml de to default working.
 #[derive(
@@ -181,6 +185,18 @@ pub struct QueryConfig {
     #[serde(default)]
     pub rpc_tls_query_service_domain_name: String,
 
+    #[structopt(long, env, help = "Table engine csv enabled")]
+    #[serde(default)]
+    pub table_engine_csv_enabled: bool,
+
+    #[structopt(long, env, help = "Table engine parquet enabled")]
+    #[serde(default)]
+    pub table_engine_parquet_enabled: bool,
+
+    #[structopt(long, env, help = "Table engine memory enabled")]
+    #[serde(default)]
+    pub table_engine_memory_enabled: bool,
+
     #[structopt(
         long,
         env = QUERY_WAIT_TIMEOUT_MILLS,
@@ -213,6 +229,9 @@ impl QueryConfig {
             rpc_tls_server_key: "".to_string(),
             rpc_tls_query_server_root_ca_cert: "".to_string(),
             rpc_tls_query_service_domain_name: "localhost".to_string(),
+            table_engine_csv_enabled: false,
+            table_engine_parquet_enabled: false,
+            table_engine_memory_enabled: false,
             wait_timeout_mills: 5000,
         }
     }
@@ -326,6 +345,27 @@ impl QueryConfig {
             rpc_tls_query_service_domain_name,
             String,
             QUERY_RPC_TLS_SERVICE_DOMAIN_NAME
+        );
+        env_helper!(
+            mut_config,
+            query,
+            table_engine_csv_enabled,
+            bool,
+            TABLE_ENGINE_CSV_ENABLED
+        );
+        env_helper!(
+            mut_config,
+            query,
+            table_engine_parquet_enabled,
+            bool,
+            TABLE_ENGINE_PARQUET_ENABLED
+        );
+        env_helper!(
+            mut_config,
+            query,
+            table_engine_memory_enabled,
+            bool,
+            TABLE_ENGINE_MEMORY_ENABLED
         );
 
         env_helper!(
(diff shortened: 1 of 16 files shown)


GrapeBaBa:
@BohuTANG @sundy-li `configs_table_test.rs` with error `memory allocation of 18446744073709551480 bytes failed`

query/src/configs/config_query.rs
```diff
@@ -40,6 +40,10 @@ const QUERY_RPC_TLS_SERVER_KEY: &str = "QUERY_RPC_TLS_SERVER_KEY";
 const QUERY_RPC_TLS_SERVER_ROOT_CA_CERT: &str = "QUERY_RPC_TLS_SERVER_ROOT_CA_CERT";
 const QUERY_RPC_TLS_SERVICE_DOMAIN_NAME: &str = "QUERY_RPC_TLS_SERVICE_DOMAIN_NAME";
 
+const TABLE_ENGINE_CSV_ENABLED: &str = "TABLE_ENGINE_CSV_ENABLED";
```
BohuTANG, line 43: These configs are read-only, we can't set them from the env.

GrapeBaBa: ok

sundy-li:
> configs_table_test.rs with error memory allocation of 18446744073709551480 bytes failed

That's strange, can you use gdb to detect which code try to allocate such large bytes?

sundy-li:
@GrapeBaBa 
I use gdb to find this problem,  hope this can help you.

```
~/work/databend issue_2853 ● ? ❯ lldb target/debug/deps/databend_query-0d58f79e6f52451c                                                                 1h 39m 6s 20:58:17
(lldb) target create "target/debug/deps/databend_query-0d58f79e6f52451c"
Current executable set to '/Users/sundy/work/databend/target/debug/deps/databend_query-0d58f79e6f52451c' (arm64).
(lldb) run --test  datasources::database::system::configs_table_test::test_configs_table

Process 55785 launched: '/Users/sundy/work/databend/target/debug/deps/databend_query-0d58f79e6f52451c' (arm64)

running 1 test
memory allocation of 18446744073709551480 bytes failed
Process 55785 stopped
* thread #2, stop reason = signal SIGABRT
    frame #0: 0x00000001c390d9b8 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`__pthread_kill:
->  0x1c390d9b8 <+8>:  b.lo   0x1c390d9d8               ; <+40>
    0x1c390d9bc <+12>: pacibsp
    0x1c390d9c0 <+16>: stp    x29, x30, [sp, #-0x10]!
    0x1c390d9c4 <+20>: mov    x29, sp
Target 0: (databend_query-0d58f79e6f52451c) stopped.
(lldb) bt

frame #15: 0x0000000102f4e76c databend_query-0d58f79e6f52451c`alloc::slice::_$LT$impl$u20$alloc..borrow..ToOwned$u20$for$u20$$u5b$T$u5d$$GT$::to_owned::h9ee28234df66fad7(self=(data_ptr = "", length = 18446744073709551480)) at slice.rs:834:9
frame #16: 0x00000001023c44fc databend_query-0d58f79e6f52451c`common_datavalues::arrays::string::DFStringArray::try_get::h5c64267f3439545c(self=0x000000010cc0c290, index=29) at mod.rs:104:27
frame #17: 0x000000010220f894 databend_query-0d58f79e6f52451c`_$LT$common_datavalues..series..wrap..SeriesWrap$LT$common_datavalues..arrays..string..DFStringArray$GT$$u20$as$u20$common_datavalues..series..series_impl..SeriesTrait$GT$::try_get::hf6ad91712d056bd8(self=0x000000010cc0c290, index=29) at wrap.rs:116:26
frame #18: 0x00000001020bdbb4 databend_query-0d58f79e6f52451c`common_datablocks::data_block_debug::create_table::hb08048629a6074fe(results=&[common_datablocks::data_block::DataBlock] @ 0x0000000170002958) at data_block_debug.rs:98:41


(lldb) f 16
frame #16: 0x00000001023c44fc databend_query-0d58f79e6f52451c`common_datavalues::arrays::string::DFStringArray::try_get::h5c64267f3439545c(self=0x000000010cc0c290, index=29) at mod.rs:104:27
   101 	        }
   102 	        let v = match self.array.is_null(index) {
   103 	            true => None,
-> 104 	            false => Some(self.array.value_unchecked(index).to_owned()),
   105 	        };
   106 	        Ok(v.into())
   107 	    }
(lldb) p self.array
(arrow2::array::binary::BinaryArray<long>) $0 = {
  data_type = {}
  offsets = {
    data = {
      ptr = {
        pointer = 0x000000010b73c570
      }
      phantom = {}
    }
    offset = 0
    length = 30
  }
  values = {
    data = {
      ptr = {
        pointer = 0x000000010b73c5a0
      }
      phantom = {}
    }
    offset = 0
    length = 136
  }
  validity = {}
}
  Fix-it applied, fixed expression was:
    self->array

(lldb) p self.array
(arrow2::array::binary::BinaryArray<long>) $0 = {
data_type = {}
offsets = {
data = {
	ptr = {
	pointer = 0x000000010b73c570
	}
	phantom = {}
}
offset = 0
length = 30
}
values = {
data = {
	ptr = {
	pointer = 0x000000010b73c5a0
	}
	phantom = {}
}
offset = 0
length = 136
}
validity = {}
}
Fix-it applied, fixed expression was:
self->array
(lldb)
```

This is an unsafe method in try_get, the reason is because the  size of values `values: Vec<&str>`  is `29` but the other column's size is `32`. 

So there is some bug in `extract_config` I guess, missing `else` condition.

query/src/configs/config_query.rs
```diff
@@ -181,6 +181,18 @@ pub struct QueryConfig {
...
+    #[serde(default)]
+    pub table_engine_csv_enabled: bool,
+
+    #[structopt(long, env, help = "Table engine parquet enabled")]
+    #[serde(default)]
+    pub table_engine_parquet_enabled: bool,
+
+    #[structopt(long, env, help = "Table engine memory enabled")]
+    #[serde(default)]
+    pub table_engine_memory_enabled: bool,
```
BohuTANG, line 194: Good job, LGTM.
It seems the memory engine can be enabled by default, so we can do some memory related tests only start the databend-query and no need to enable it in the toml.

GrapeBaBa: It can't set field default true in structopt. If the field not in toml file, it thinks it is not present and make value false.

PR #2964, opened 2021-11-22
force kill databend-query/meta when fail to kill it.

youngsofun:

## Summary

the servers may fail to shutdown some times (most time it is a bug). developers who use these scripts (maybe through 'make run')  may connect to the old server and encounter strange errors,  or find their changes not working.

and it may not be easily covered by tests, because this may happen after some ops on the servers.

1.  kill -9 to proceed
2. let the user be aware of this. user may report it or it is a bug introduced by the local change.

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/scripts/deploy/databend-query-cluster-3-nodes.sh
+++ b/scripts/deploy/databend-query-cluster-3-nodes.sh
@@ -15,6 +15,14 @@ sleep 3
 killall databend-meta
 sleep 3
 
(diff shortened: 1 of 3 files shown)


PsiACE (approved):
lgtm

BohuTANG:
/lgtm

These scripts only for our tests, but the improvement is really useful

PR #2971, opened 2021-11-23
Make python support custom output

sundy-li:

## Summary

Summary about this PR

## Changelog

 
- Improvement
 

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/tests/suites/0_stateless/12_0000_insert_into_select2.py
+++ b/tests/suites/0_stateless/12_0000_insert_into_select2.py
@@ -30,3 +30,7 @@ DROP DATABASE db1;
 """
 
 client1.run(sqls)
+stdout, stderr = client1.run_with_output("select * from system.metrics")
+assert stdout is not None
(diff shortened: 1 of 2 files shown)


BohuTANG:
/lgtm

Thank you @sundy-li

tests/suites/0_stateless/12_0000_insert_into_select2.py
```diff
@@ -30,3 +30,7 @@
 """
 
 client1.run(sqls)
+stdout, stderr = client1.run_with_output("select * from system.metrics")
+assert stdout is not None
```
BohuTANG, line 34: `print(stdout)` will show in the result file?

sundy-li: yes, but we should use `print(stdout, end = '')` to ignore the Newline.

Issue #1217, opened 2021-07-28
[Tracking] Refactor SQL layer


leiysky:
To build a robust SQL processing system, it's necessary to divide the components clearly. It is helpful to understand the code.

RFC: https://github.com/datafuselabs/databend/blob/main/website/databend/docs/dev/rfcs/query/0005-new-sql-planner-framework.md

Tracking issue for RFC: #1853 

Tasks:

- [ ] Parser: the component to parse SQL string into AST(Abstract Syntax Tree) #1218 
- [ ] Binder: the component takes AST as input and produce a logical plan https://github.com/datafuselabs/datafuse/issues/1219
- [ ] Optimizer: the component can do some optimization to a logical plan, and eventually produce a nice physical plan
- [ ] Catalog: the component maintains all items in the database(e.g. table, column, index, function), which can help Binder to build a correct logical plan #867 
- [ ] Executor: the component to execute physical plan and produce query result

zhang2014:
Maybe we will follow [it](https://15721.courses.cs.cmu.edu/spring2020/slides/19-optimizer1.pdf):
![image](https://user-images.githubusercontent.com/8087042/127422881-f13afa70-4a94-48ed-ab53-16e52e9c43e7.png)

leiysky:
> Maybe we will follow [it](https://15721.courses.cs.cmu.edu/spring2020/slides/19-optimizer1.pdf):
> ![image](https://user-images.githubusercontent.com/8087042/127422881-f13afa70-4a94-48ed-ab53-16e52e9c43e7.png)

Yes, it's exactly what I mean.

Does the picture come from the lecture of CMU 15-445?

zhang2014:
https://15721.courses.cs.cmu.edu/spring2020/

leiysky:
> > Does the picture come from the lecture of CMU 15-445?
> 
> It from https://15721.courses.cs.cmu.edu/spring2020/

I've followed Andy's coureses for years. 15721 is a nice course which presents the latest academic results and industrial development every year.

xudong963:
> RFC: https://github.com/datafuselabs/databend/blob/master/website/databend/docs/rfcs/query/0005-new-sql-planner-framework.md

404

PsiACE:
> > RFC: https://github.com/datafuselabs/databend/blob/master/website/databend/docs/rfcs/query/0005-new-sql-planner-framework.md
> 
> 404

[...]

PR #2983, opened 2021-11-23
New planner framework

leiysky:

## Summary

#1217

## Changelog

- New Feature


query/src/sql/optimizer/cascades/mod.rs
```diff
@@ -0,0 +1,146 @@
+// Copyright 2020 Datafuse Labs.
```
BohuTANG, line 1: 2020 -> 2021 :D

leiysky: Sure

TennyZhuang:
You are so successful!

query/src/sql/opt/cascades/explore_rules.rs
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::sql::opt::RuleSet;
+
+pub fn get_explore_rule_set() -> RuleSet {
+    RuleSet::create_with_ids(vec![]).unwrap()
+}
+
+#[cfg(test)]
```
BohuTANG, line 21: :+1: 

The test style has changed, we place them to query/tests/it/sql

leiysky: I'll migrate them later.

zhang2014:
maybe change `query/src/sql/opt` to `query/src/sql/optimizer` is better?

query/src/sql/planner/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+        // Step 2: bind AST with catalog, and generate a pure logical SExpr
+        let binder = Binder::new(self.context.get_catalog(), self.context.clone());
+        let bind_result = binder.bind(&stmts[0]).await?;
+
+        // Step 3: optimize the SExpr with optimizers, and generate optimized physical SExpr
+        let optimize_context = OptimizeContext::create_with_bind_context(&bind_result.bind_context);
+        let optimized_expr = optimize(bind_result.s_expr().clone(), optimize_context)?;
+
+        // Step 4: build executable Pipeline with SExpr
+        let exec = Executor::create(self.context.clone(), bind_result.metadata);
```
zhang2014, line 65: Maybe we should consider how it integrates with `Interpreter`.

leiysky:
> maybe change `query/src/sql/opt` to `query/src/sql/optimizer` is better?

It's fine.

query/src/sql/planner/bind_context.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub index: IndexType,
+    pub data_type: DataType,
+    pub nullable: bool,
+    pub expr: Option<ScalarExpr>,
+}
+
+/// `BindContext` stores all the free variables in a query and tracks the context of binding procedure.
+#[derive(Clone, Default)]
+pub struct BindContext {
+    _parent: Option<Box<BindContext>>,
```
xudong963, line 39: I guess this member is reserved for correlated subqueries?

query/src/sql/planner/mod.rs
leiysky: I had tried to integrate it with `Interpreter`, but failed in the end. The problem is `Interpreter` is heavily coupled with `PlanNode`, there's no appropriate way to combine them.

Maybe we can provide a high level abstraction that takes SQL text as input and returns a `DataBlockStream` in the future to migrate the whole planner.

query/src/sql/planner/bind_context.rs
leiysky: Yes, we need to resolve outer columns with a stack-like structure.

query/src/sql/optimizer/property.rs
```diff
@@ -0,0 +1,72 @@
...
+    output_columns: ColumnSet,
+}
+
+impl RelationalProperty {
+    pub fn create(output_columns: ColumnSet) -> Self {
+        RelationalProperty { output_columns }
+    }
+
+    pub fn output_columns(&self) -> &ColumnSet {
+        &self.output_columns
```
sundy-li, line 61: We don't need ordering for `output_columns` ?

query/src/sql/planner/plan/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Relational operator
+#[derive(Clone, PartialEq, Debug)]
+pub enum Plan {
+    // Logical operators
+    LogicalGet(LogicalGet),
+    LogicalProject(LogicalProject),
+
+    // Physical operators
+    PhysicalScan(PhysicalScan),
+    PhysicalProject(PhysicalProject),
```
sundy-li, line 54: Will it make Plan too fatter if we have logical/physical plan in a single enum?

Visitor/Rewriter should consider about all match cases.

query/src/sql/optimizer/memo.rs
```diff
@@ -0,0 +1,129 @@
...
+use std::fmt::Formatter;
+
+use common_exception::Result;
+
+use crate::sql::optimizer::group::Group;
+use crate::sql::optimizer::m_expr::MExpr;
+use crate::sql::optimizer::s_expr::SExpr;
+use crate::sql::IndexType;
+
+/// `Memo` is a search space which memoize possible plans of a query.
```
Xuanwo, line 25: Why do we name this struct `Memo`, are there any agreed-upon rules?

query/src/sql/optimizer/property.rs
leiysky: We don't need to track order of attributes in a tuple during optimization, since if two different relational expressions can produce same set of attributes then they can be regarded as logically equivalent.

There is a ordered and named column list stored in `OptimizeContext`, with which we can render the final result set of query. Though I haven't utilize with it yet.

https://github.com/datafuselabs/databend/blob/0f17720ddfe8797a87523063c5a5629697b81d44/query/src/sql/optimizer/optimize_context.rs#L37-L55

query/src/sql/optimizer/memo.rs
leiysky: This name comes from the paper of cascades optimizer. I think it can help people have knowledge about query optimizer get up with the project more easily.

flaneur2020: https://www.querifylabs.com/blog/memoization-in-cost-based-optimizers i guess this is related

query/src/sql/planner/bind_context.rs
flaneur2020: a Box<BindContext> means this BindContext owns this _parent, if the _parent a reference to a parent BindContext object, it'd be a weak reference pointer?

query/src/sql/planner/plan/mod.rs
leiysky: We have requirement for down casting a base `Plan` type to concrete logical or physical plan struct.

For example:

https://github.com/datafuselabs/databend/blob/0f17720ddfe8797a87523063c5a5629697b81d44/query/src/sql/optimizer/rule/rule_implement_get.rs#L46-L53

I found that enum is the best choice in rust to implement this.

The good thing is we don't need to write too many visitors to probe the `Plan` with this framework. Every `Rule` can focus on the subtree matches its pattern expression.

query/src/sql/planner/bind_context.rs
leiysky: > a Box means this BindContext owns this _parent, if the _parent a reference to a parent BindContext object, it'd be a weak reference pointer?

It's complicated to maintain a recursive data structure in rust, so I just make parent an owned element and won't be shared.

query/src/sql/optimizer/memo.rs
leiysky: > https://www.querifylabs.com/blog/memoization-in-cost-based-optimizers i guess this is related

Yes, it's the state of memoization, just like a table in dynamic programming.

sundy-li: We can have this refer URL inside the codes for other new guys to understand it in the future.

leiysky: > We can have this refer URL inside the codes for other new guys to understand it in the future.

I'm going to explain this in RFC, and refer the RFC in comment.

query/src/sql/optimizer/cascades/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.init(expression)?;
+
+        self.explore_group(self.memo.root().unwrap().group_index())?;
+
+        self.implement_group(self.memo.root().unwrap().group_index())?;
+
+        self.find_optimal_plan()
+    }
+
+    fn explore_group(&mut self, group_index: IndexType) -> Result<()> {
```
zhang2014, line 72: Maybe should add a todo comment. we need to use while to rewrite the 
 explore task(maybe also other task)?

query/src/sql/optimizer/m_expr.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::sql::optimizer::rule::RulePtr;
+use crate::sql::optimizer::rule::TransformState;
+use crate::sql::optimizer::s_expr::PlanPtr;
+use crate::sql::optimizer::SExpr;
+use crate::sql::IndexType;
+
+/// `MExpr` is abbreviation of multiple expression, which is the representation of relational
+/// expressions inside `Memo`.
+#[derive(Clone)]
+pub struct MExpr {
```
zhang2014, line 31: Maybe rename to `GroupExpression` is better?

query/src/sql/optimizer/cascades/mod.rs
leiysky: You remind me about this. I was going to unify explore and implement transform task but forgot to do that.

query/src/sql/optimizer/m_expr.rs
leiysky: It was named in `GroupExpression` at the beginning. But I thought the word `Expression` is too common in databend, `MExpr` and `SExpr` looked more distinct for me.

BohuTANG:
If there is no other review comments, let's merge it first :rocket:

BohuTANG:
Sub tasks(or new task) will be tracked in https://github.com/datafuselabs/databend/issues/1217

PR #2986, opened 2021-11-23
[ISSUE-2779] Add user in session

### LINKED ISSUE #2779: Bind user to session
**Summary**

There are a few tasks need to do:
- [x] Add user name to InteractiveWorker when the auth is passed
- [ ] Add user name to https://github.com/datafuselabs/databend/blob/e595993705ce9b719b0df78a4e1013a02fd77510/query/src/sessions/sessions.rs#L100
- [x] Get user info by user name and bind the user info to the session

Any thoughts are welcome ~

flaneur2020:

## Summary

this PR adds a user record in the session, which recorded on the authenticate phase of mysql & clickhouse. it also adds a user field in the `system.processes` table.

## Changelog

- Improvement

## Related Issues

Fixes #2779

## Test Plan

Unit Tests
Stateless Tests

### DIFF
--- a/query/src/sessions/session_status.rs
+++ b/query/src/sessions/session_status.rs
@@ -44,10 +44,10 @@ impl MutableStatus {
     pub fn try_create() -> Result<Self> {
         Ok(MutableStatus {
             abort: Default::default(),
-            current_user: RwLock::new(None),
+            current_user: Default::default(),
+            client_host: Default::default(),
             current_database: RwLock::new("default".to_string()),
             session_settings: RwLock::new(Settings::try_create()?.as_ref().clone()),
-            client_host: Default::default(),
             io_shutdown_tx: Default::default(),
             context_shared: Default::default(),
         })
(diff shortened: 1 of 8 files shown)


query/src/sessions/session_status.rs
```diff
@@ -30,6 +30,7 @@ pub struct MutableStatus {
     abort: AtomicBool,
     current_database: RwLock<String>,
     session_settings: RwLock<Settings>,
+    current_user: RwLock<Option<String>>,
```
BohuTANG, line 33: it will the UserInfo in the feature not only a user name?

flaneur2020: I'm taking UserInfo at first but reverted it back to a simple user name, in the current code base UserInfo is a fat object:

```
  pub struct UserInfo {
      pub name: String,
      pub hostname: String,
      pub password: Vec<u8>,
      pub auth_type: AuthType,
      pub privileges: UserPrivilege,
      pub quota: UserQuota,
  }
```

the fields inside UserInfo might get mutated on the fly (like quota, privileges, even password), or mutated by another query instance. a conservative usage of UserInfo might be always fetching `UserInfo` from the metasrv, and make it transactional with the user-related operations.

in the case about session privilege, we can construct a unique identifier for the privilege grants by user_name + client_address.

query/src/sessions/session_status.rs
```diff
@@ -72,6 +74,18 @@ impl MutableStatus {
         *lock = db
     }
 
+    // Set the current user after authentication
```
BohuTANG, line 77: It would be nice to add a unit test in test crate we have some MutableStatus tests there

flaneur2020: @BohuTANG thank you for remind, i've added a test case on the getter/setter for current user

flaneur2020: to test the behaviour of session's current user, i'd split another PR to add a current_user() function

PR #2989, opened 2021-11-23
Impl RAND() func

### LINKED ISSUE #2563: [functions] more functions working with number

mshauneu:

## Summary

Impl RAND() func

## Changelog

- New Feature

## Related Issues

Fixes #2563

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/maths/random.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn return_type(&self, _args: &[DataType]) -> Result<DataType> {
+        Ok(DataType::Float64)
+    }
+
+    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+        Ok(false)
+    }
+
+    fn eval(&self, _columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
+        let arr = DFFloat64Array::full(-1.0, input_rows).apply(|_| rand::thread_rng().gen::<f64>());
```
sundy-li, line 63: Aha, we do not need to create a temp array.

Let's do it with :

```
let iter = (0..input_rows).iter().map(|_| rand::thread_rng().gen::<f64>() );
let array = f64array::new_from_iter(iter);
```

Junnplus:
https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_rand

rand should support the seed parameter, like mysql `RAND([N])`

common/functions/src/scalars/maths/random.rs
Junnplus: `select rand() from numbers(10);` return columns with the same value, not random.

@sundy-li `input_rows` seem always is 0 when call functions with constant parameters

sundy-li: It's incorrect, `input_rows` is the row count of the column.

numbers(10) may be divided into N blocks (CPU threads), if you set `max_threads = 1`,  there will be one block with 10 input_rows.

And this could be optimized in constant folding.

sundy-li: See `transform_expression_executor.rs`

mshauneu: @Junnplus don't understand 

mysql: 
```
 mysql -uroot -e 'WITH DIGITS (N) AS (SELECT 0 UNION ALL SELECT 1 UNION ALL SELECT 2 UNION ALL SELECT 3 UNION ALL   SELECT 4 UNION ALL SELECT 5 UNION ALL SELECT 6 UNION ALL SELECT 7 UNION ALL   SELECT 8 UNION ALL SELECT 9) SELECT RAND() FROM DIGITS'
+---------------------+
| RAND()              |
+---------------------+
|  0.9070275099082175 |
|  0.8253372067821558 |
| 0.40560353491977186 |
|  0.5520060849860368 |
|   0.543219850904513 |
| 0.06008103029809988 |
|  0.6707456295106053 |
| 0.17348508739237217 |
|  0.8551885791636897 |
|   0.755487664374977 |
+---------------------+
```

my random: 
```txt
mysql -h0.0.0.0 -P3307 -uroot -e 'SELECT RAND() from numbers(10)'
+-----------------------+
| RAND()                |
+-----------------------+
| 0.0035617389443762804 |
|    0.3463889841872543 |
|    0.6216070973026121 |
|   0.41610771730421137 |
|   0.06636151314764394 |
|   0.48847821091292687 |
|    0.8531923644472893 |
|    0.3569899290235252 |
|   0.06828498341928513 |
|    0.5615828780217206 |
+-----------------------+
```
imho expected result... please clarify...

Junnplus: @mshauneu My fault, local code problem.

common/functions/src/scalars/maths/random.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    fn return_type(&self, args: &[DataType]) -> Result<DataType> {
+        if args.is_empty()
+            || matches!(
+                args[0],
+                DataType::UInt8
+                    | DataType::UInt16
+                    | DataType::UInt32
+                    | DataType::UInt64
+                    | DataType::Null
```
Junnplus, line 67: Float、Int and String should also be accepted.

common/functions/src/scalars/maths/random.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
+        let r_column: DataColumn = match columns.len() {
+            1 => {
+                let d_column: &DataColumn =
+                    &columns[0].column().cast_with_type(&DataType::UInt64)?;
+
+                match d_column {
+                    DataColumn::Constant(d, _) => {
+                        if d.is_null() {
+                            let mut rng = rand::thread_rng();
```
Junnplus, line 92: null should be as seed 0.

for mysql
```
select rand(null) = rand(0); # 1
```

common/functions/src/scalars/maths/random.rs
mshauneu: @Junnplus nw, btw I just updated docs as well.

sundy-li (approved):

common/functions/src/scalars/maths/random.rs
```diff
@@ -0,0 +1,130 @@
...
+                    DataColumn::Array(d_series) => binary(
+                        &DFUInt64Array::new_from_iter(
+                            (0..input_rows).into_iter().map(|r| r as u64),
+                        ),
+                        d_series.u64()?,
+                        |_, s| {
+                            let mut rng = rand::rngs::StdRng::seed_from_u64(s);
+                            rng.gen::<f64>()
+                        },
+                    ),
```
Junnplus, line 113: Using ArrayApply instead of binary for one iterator.

```suggestion
                   DataColumn::Array(seed_series) => seed_series.u64()?.apply_cast_numeric(|seed| {
                       let mut rng = StdRng::seed_from_u64(seed);
                       rng.gen::<f64>()
               }),
```

common/functions/src/scalars/maths/random.rs
mshauneu: good catch, fixed!

common/functions/src/scalars/maths/random.rs
mshauneu: fixed

common/functions/src/scalars/maths/random.rs
mshauneu: make sense, applied in my commit

Junnplus (approved):

PR #2993, opened 2021-11-24
Add tests for http handler.

youngsofun:

## Summary

Add tests for http handler.  and limit rows per page to 10000 so the client can test on the pagination easily.

1.  pagination
2. insert

## Changelog

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/servers/http/v1/query/result_data_manager.rs
+++ b/query/src/servers/http/v1/query/result_data_manager.rs
@@ -27,6 +27,8 @@ use crate::servers::http::v1::block_to_json::block_to_json;
 use crate::servers::http::v1::block_to_json::JsonBlock;
 use crate::servers::http::v1::block_to_json::JsonBlockRef;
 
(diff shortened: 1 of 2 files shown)


query/src/servers/http/v1/query/result_data_manager.rs
```diff
@@ -125,12 +125,20 @@ impl ResultDataManager {
...
 
         let mut end = false;
         loop {
             match ResultDataManager::receive(block_rx, tp).await {
-                Ok(block) => results.push(block_to_json(&block).unwrap()),
+                Ok(block) => {
+                    rows += block.num_rows();
+                    results.push(block_to_json(&block).unwrap());
+                    // TODO(youngsofun):  set it in post if needed
+                    if rows >= 10000 {
```
flaneur2020, line 138: define a constant for it?

youngsofun: ok

PR #3077, opened 2021-11-24
Fix bendctl load.

youngsofun:

## Summary

fix:

1.    Incorrect usage of /v1/query 
2.   `.reduce_with(|a, b| format!("{}, {}", a, b));` is very slow.
3.   stateful test build with `release`,  but run with `debug`. 

```
[ok] ✅ successfully loaded 371,357 lines, rows/src: 3,946 (rows/sec). time: 94.103 sec
```

## Changelog

## Test Plan

Unit Tests

Stateless Tests


ZhiHanZ (approved):
reduce_with was used to fix some typo issues happened in CSV to make it work on ontime dataset, seems like we already fixed them well :)

PR #3092, opened 2021-11-25
Added `OCT`, `REPEAT` scalar functions.

### LINKED ISSUE #3055: REPEAT() | Repeat a string the specified number of times

### LINKED ISSUE #3068: OCT() | Return a string containing octal representation of a number

### LINKED ISSUE #3074: OCT() | Return a string containing octal representation of a number

sunli829:

## Summary

- Added `OCT`, `REPEAT` scalar functions.

- Added helper methods:

    `DataType::is_integer` `DataType::is_signed_integer` `DataType::is_unsigned_integer`

    `DataValue::is_integer` `DataValue::is_signed_integer` `DataValue::is_unsigned_integer`

## Changelog

- New Feature

## Related Issues

fixes #3074, fixes #3068, fixes #3055 

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/strings/oct.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn num_arguments(&self) -> usize {
+        1
+    }
+
+    fn variadic_arguments(&self) -> Option<(usize, usize)> {
+        None
+    }
+
+    fn return_type(&self, args: &[DataType]) -> Result<DataType> {
+        if !args[0].is_integer() {
```
sundy-li, line 113: args[0] can be DataType::null.

if args[0] is String, we can cast it into  Int64 using `cast_with_type`

MySQL support explicit cast:

select oct('56');
select oct('56a');
select oct(null);

common/functions/src/scalars/strings/repeat.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    fn return_type(&self, args: &[DataType]) -> Result<DataType> {
+        if !matches!(args[0], DataType::String) {
+            return Err(ErrorCode::IllegalDataType(format!(
+                "Expected parameter 1 is string, but got {}",
+                args[0]
+            )));
+        }
+
+        if !args[1].is_integer() {
```
sundy-li, line 117: let's just accept `args[1]` as  `unsigned integer` | `String` | `Null`.

Then we can cast column[1] to UInt64Column to reduce the matches.

common/functions/src/scalars/strings/repeat.rs
```diff
@@ -0,0 +1,234 @@
...
+    fn num_arguments(&self) -> usize {
+        2
+    }
+
+    fn variadic_arguments(&self) -> Option<(usize, usize)> {
+        None
+    }
+
+    fn return_type(&self, args: &[DataType]) -> Result<DataType> {
+        if !matches!(args[0], DataType::String) {
```
sundy-li, line 110: DataType::Null can be supported.

tests/suites/0_stateless/02_0018_function_strings_repeat.sql
```diff
@@ -0,0 +1,91 @@
...
+
+CREATE TABLE strings_repeat_sample_3(s String) engine=Memory;
+
+INSERT INTO strings_repeat_sample_3 VALUES ('abc'), ('def');
+
+select repeat(s, 3) from strings_repeat_sample_3;
+select repeat(s, 0) from strings_repeat_sample_3;
+select repeat(s, -5) from strings_repeat_sample_3;
+
+drop table strings_repeat_sample_3;
```
sundy-li, line 91: Newline needed.

common/functions/src/scalars/strings/repeat.rs
sundy-li: Cast is not expensive because it's vectorized.

sunli829: > Then we can cast column[1] to UInt64Column to reduce the matches.

Will this cause unnecessary memory allocation?

sundy-li: Yes, but I think it's ok. The old column memory will be dropped immediately.

In most case, the second column is a `DataColumn::Constant`, so we can use `let times = columns[1].to_minimal_array().cast_with_type(&DataType::UInt64)`.

sundy-li: Since we already wrote the match macro, it's ok to not change that :)

Let's add String|Null support!

common/functions/src/scalars/strings/repeat.rs
```diff
@@ -0,0 +1,234 @@
...
+use crate::scalars::function_factory::FunctionFeatures;
+use crate::scalars::Function;
+
+macro_rules! const_array {
+    ($input_string:expr, $times:expr, $input_rows:expr) => {{
+        let mut string_builder = StringArrayBuilder::with_capacity($input_rows);
+        for n in $times {
+            let value = n.and_then(|n| {
+                if *n >= 0 {
+                    Some($input_string.repeat(*n as usize))
```
sundy-li, line 31: How about adding `max_repeat_times` check.  I don't want this to be an oom killer.

sundy-li: https://github.com/ClickHouse/ClickHouse/blob/master/src/Functions/repeat.cpp#L26-L32

sunli829:
done. 🙂

sundy-li (approved):
/LGTM

PR #2998, opened 2021-11-24
QUERY-2992: Returning name with *Ref to its original form Arc<T>

BohuTANG:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

This PR only do re-name, no code logics changed.

* DatabendQueryContextRef -> Arc\<DatabendQueryContext\>
* SessionManagerRef -> Arc\<SessionManager\>
* ClusterRef -> Arc\<Cluster\>
* DatabendQueryContext -> QueryContext
* DatabendQueryContextShared -> QueryContextShared

## Changelog

- Improvement

## Related Issues

Fixes #2992 

## Test Plan

Unit Tests

Stateless Tests

PR #3097, opened 2021-11-25
Add time in http handler response.

youngsofun:

## Summary

Add time in http handler response.
the front end need it

## Changelog

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/servers/http/v1/query/execute_state.rs
+++ b/query/src/servers/http/v1/query/execute_state.rs
@@ -99,6 +99,21 @@ impl Executor {
             Stopped(f) => f.stop_time - self.start_time,
         }
     }
+    pub(crate) async fn stop(this: &ExecutorRef, reason: Result<()>, kill: bool) {
+        let mut guard = this.write().await;
+        if let Running(r) = &guard.state {
+            // release session
+            let progress = Some(r.context.get_progress_value());
+            if kill {
+                r.session.force_kill_query();
+            }
+            guard.state = Stopped(ExecuteStopped {
+                progress,
+                reason,
+                stop_time: Instant::now(),
+            });
+        };
+    }
 }
 
 pub struct HttpQueryHandle {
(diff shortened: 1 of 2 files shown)


query/src/servers/http/v1/query/http_query.rs
```diff
@@ -35,6 +35,7 @@ pub struct ResponseInitialState {
 }
 
 pub struct ResponseState {
+    pub wall_time_ms: u128,
```
youngsofun, line 38: add a field here

youngsofun:
@flaneur2020

query/src/servers/http/v1/query/execute_state.rs
```diff
@@ -154,34 +165,38 @@ impl ExecuteState {
...
                     } else {
-                        ExecuteState::stop(&state, Ok(()), false).await;
+                        ExecuteState::stop(&executor, Ok(()), false).await;
                         break;
                     }
                 }
                 log::debug!("drop block sender!");
             })?;
-        Ok((state_clone, schema))
+        Ok((executor_clone, schema))
```
flaneur2020, line 184: seems we can directly returns `Ok((executor.clone(), schema))` ? it seems the variable `executor_clone` is not used elsewhere 🤔

query/src/servers/http/v1/query/execute_state.rs
```diff
@@ -154,34 +165,38 @@ impl ExecuteState {
...
     }
 
-    pub(crate) async fn stop(this: &ExecuteStateRef, reason: Result<()>, kill: bool) {
+    pub(crate) async fn stop(this: &ExecutorRef, reason: Result<()>, kill: bool) {
```
flaneur2020, line 187: this -> self ?

query/src/servers/http/v1/query/http_query.rs
```diff
@@ -112,7 +113,9 @@ impl HttpQuery {
     pub async fn get_state(&self) -> ResponseState {
         let state = self.state.read().await;
         let (exe_state, err) = state.state.extract();
+        let wall_time_ms = state.elapsed().as_millis();
         ResponseState {
+            wall_time_ms,
```
flaneur2020, line 118: the wall time could be very useful!

query/src/servers/http/v1/query/execute_state.rs
youngsofun: I thought so,  but compile not passed ..

flaneur2020: thank god, ignore me 🙈

query/src/servers/http/v1/query/execute_state.rs
youngsofun: type ExecutorRef = Arc<RwLock<Executor>>;
so it is a not method, I saw some suggestion to use `this` in the situation.

flaneur2020: TIL

query/src/servers/http/v1/query/execute_state.rs
```diff
@@ -71,24 +73,47 @@ impl ExecuteState {
 
 use ExecuteState::*;
 
-pub(crate) type ExecuteStateRef = Arc<RwLock<ExecuteStateWrapper>>;
+pub(crate) type ExecutorRef = Arc<RwLock<Executor>>;
```
BohuTANG, line 76: It would be better to introduce a new struct wrapping Executeor with RwLock here, 

struct ExecutorState {Arc<RwLock<Executor>>} and implement ExecutorState , like https://github.com/datafuselabs/databend/pull/2998

BohuTANG: We could do it in another PR

youngsofun: thank you, I will rethink it later.

PR #3122, opened 2021-11-26
ISSUE-3121: simple chunk strategy in insertion

### LINKED ISSUE #3121: simple chunked insertion (by number of blocks)
**Summary**

enable a simple chunk strategy during insertion: 
 - try to compact a given **number** DataBlocks of upstream into one DataBlock (and form a segment for it)
 
the **number**  could be configured in table options, if not config, a default value 1000 will be used.

dantengsky:

## Summary

- issue #3121 
- issue #3147

the simple strategy goes like this:

For each query node that processes the insertion, while consuming the `SendableDatablockStream`
 - for each `DEFAULT_CHUNK_BLOCK_NUM` of blocks, let's say B, the following reshape is applied: 
    for each n successive data block in B, if the sum of their `memory_size` exceeds  `BLOCK_SIZE_THRESHOLD`, they will be merged into one larger block.
    
NOTE:
- currently, `DEFAULT_CHUNK_BLOCK_NUM` and `BLOCK_SIZE_THRESHOLD` will be fetched first from table options
   if not there, default values will be applied
- the max size of merged-block will be 2 * block_size_threshold (in memory size)
- after compression, the on-storage size of the block varies
- for block that is larger than `block_size_threshold`, they will NOT be split
    it might be better to apply fine-grained strategy in table compact/optimization

after this simple strategy is applied, the read performance might be improved for somehow:

A very **non**-scientific, local debug test:

- for main branch commit 05f3ab5472034296ed1f3be534b31de056036d8e:

~~~
SELECT
    count(1),
    avg(Year),
    sum(DayOfWeek)
FROM ontime

┌─count(1)─┬─avg(Year)─┬─sum(DayOfWeek)─┐
│  2970856 │      2020 │       11657448 │
└──────────┴───────────┴────────────────┘

1 rows in set. Elapsed: 1.313 sec.
~~~

- this PR

~~~
select count(1) ,avg(Year), sum(DayOfWeek)  from ontime;
+----------+-----------+----------------+
| count(1) | avg(Year) | sum(DayOfWeek) |
+----------+-----------+----------------+
|  2970856 |      2020 |       11657448 |
+----------+-----------+----------------+
1 row in set (0.37 sec)
Read 0 rows, 0 B in 0.287 sec., 0 rows/sec., 0 B/sec.


select count(1) ,avg(Year), sum(DayOfWeek)  from ontime;
+----------+-----------+----------------+
| count(1) | avg(Year) | sum(DayOfWeek) |
+----------+-----------+----------------+
| 11883424 |      2020 |       46629792 |
+----------+-----------+----------------+
1 row in set (1.10 sec)
Read 0 rows, 0 B in 0.894 sec., 0 rows/sec., 0 B/sec.
~~~

data of `ontime` is generated by first `v1 streaming_load`, and then `insert into ontime select * from ontime` several times.

## Changelog
- Improvement

## Related Issues

Fixes #3121

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
Why not make the merge strategy by num of rows?

dantengsky:
> Why not make the merge strategy by num of rows?

cause `DataBlock::concat_blocks` is too handy :-) 

just kidding, `by num of rows` or `by block size` is more sensible, will  be addressed in upcoming PR(s)

query/src/storages/fuse/io/block_appender.rs
```diff
@@ -33,39 +35,79 @@ use crate::storages::fuse::statistics;
...
 
 impl BlockAppender {
+    // A simple strategy of merging small blocks into larger ones:
+    // for each n successive data blocks in `blocks`, if the sum of their `memory_size` exceeds
+    //   `block_size_threshold`, they will be merged into one larger block.
+    //  NOTE:
+    //    - the max size of merge-block will be 2 * block_size_threshold
+    //    - for block that is larger than `block_size_threshold`, they will NOT be split
+    //        TODO handling split in table compact/optimize
+    pub(crate) fn reshape_blocks(
```
sundy-li, line 48: I think it's better to sort the block in previous.

Then we can use this greedy merge in order.

dantengsky: good idea!

dantengsky:
@sundy-li PTAL

sundy-li (approved):

PR #3150, opened 2021-11-29
[fuse table] insert overwrite stmt implementation

### LINKED ISSUE #3144: [fuse table] impl insert overwrite stmt
**Summary**

impl `insert overwrite ...`  stmt

goal:
* replace a fuse table's data from another table, drop previous snapshots and segments info.

this is a task in #1780

Veeupup:

## Summary

add `insert overwrite` statement

## Changelog

- New Feature

## Related Issues

Fixes #3144 

## Test Plan

Unit Tests

Stateless Tests


common/planners/src/plan_replace.rs
```diff
@@ -0,0 +1,40 @@
...
+use common_datavalues::DataSchemaRef;
+use common_meta_types::MetaId;
+
+use crate::PlanNode;
+
+#[derive(serde::Serialize, serde::Deserialize, Clone)]
+pub struct ReplacePlan {
+    pub db_name: String,
+    pub tbl_name: String,
+    pub tbl_id: MetaId,
```
drmingdrmer, line 24: For structure fields, always using full-name e.g., `table_name`, `table_id` would be better for searching etc.

Abbreviations should only be considered when being used as local variable names.

BohuTANG: Agree

dantengsky:
@Veeupup Sorry for the misleading issue #3144 that I created, I think that impls `insert overwrite` instead might be better. let's discuss this tomorrow

common/planners/src/plan_replace.rs
Veeupup: thanks for guiding : )

sundy-li:
We can have both syntaxes.  Replace | Insert override .

Veeupup:
are they only different in syntaxes? or any other point I did not get?

dantengsky:
great work! :+1: 

It would be better if there were:

-  unit test case for `Table::commit(..., overwrite=true)`
   extending `query/src/storages/fuse/table_test.rs`,  or any other way you'd like
-  a stateless test case 
   like `tests/suites/0_stateless/09_0004_remote_insert_into_select.sql` ( a simpler version will be enough)
   but if the SQL parser layer needs some further tuning, I think we can put this in another PR

Veeupup:
> great work! 👍
> 
> It would be better if there were:
> 
> * unit test case for `Table::commit(..., overwrite=true)`
>   extending `query/src/storages/fuse/table_test.rs`,  or any other way you'd like
> * a stateless test case
>   like `tests/suites/0_stateless/09_0004_remote_insert_into_select.sql` ( a simpler version will be enough)
>   but if the SQL parser layer needs some further tuning, I think we can put this in another PR
: D
done.

dantengsky (approved):
LGTM

sundy-li (approved):

Issue #3123, opened 2021-11-26
[Security] Support to use STS token credential to visit storage backend


ZhiHanZ:
**Summary**
In k8s cluster like EKS, service account could trigger some sort of controller to generate token with specified IAM permissions to manipulate S3 resources, it would be nice to support it on storage configuration
helpful reference:
https://levelup.gitconnected.com/using-iam-roles-to-allow-the-pods-in-aws-eks-to-read-the-aws-s3-bucket-be493fbdda84

https://rusoto.github.io/rusoto/rusoto_sts/struct.StsWebIdentityFederationSessionCredentialsProvider.html

PR #3165, opened 2021-11-30
[S3] add sns web identity provider to s3 client

ZhiHanZ:

## Summary

Use auto-refresh web identity token on kubernetes to access S3 bucket

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/3123

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
@mergify update

PsiACE (approved):
LGTM

common/dal/src/accessors/aws_s3/s3.rs
```diff
@@ -53,15 +53,31 @@ impl S3 {
...
-                    ErrorCode::DALTransportError(format!(
-                        "failed to create default credentials provider, {}",
-                        e
-                    ))
-                })?,
-                dispatcher,
-            ),
+            None => {
+                // check on k8s admission webhook injection
+                if std::env::var("AWS_WEB_IDENTITY_TOKEN_FILE").is_ok() {
```
BohuTANG, line 58: Why not add a new s3 storage config?

ZhiHanZ: does not seems to be make sense... using SNS web identity means to use token generated by some controllers with fixed env variable in every instance, there seems have little space on self customization,

BohuTANG: If we inline a env in the codes, it looks very tricky, and feels like a back door was left open, we should make this configuration public although we are using it from env

ZhiHanZ: https://www.kubeflow.org/docs/distributions/aws/deploy/install-kubeflow/#option-1-use-aws-iam-roles-for-service-accounts-default-and-recommended

I would like to add similar feature flag to let user use IAM policies.

BohuTANG: Nice, thank you @ZhiHanZ

PR #3177, opened 2021-11-30
Add ignore function which can be used in performance tests

sundy-li:

## Summary

Add ignore function which can be used in performance tests

Summary about this PR

## Changelog

- New Feature
- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/others/ignore.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright 2020 Datafuse Lfloor.
```
zhang2014, line 1: 2020 -> 2021

BohuTANG (approved):
lgtm

PR #3178, opened 2021-11-30
[common/meta/embedded] feature: for non-test use, init meta store in configured dir

drmingdrmer:

##### [common/meta/embedded] feature: for non-test use, init meta store in configure dir

For production use of meta-embedded, use a persistent meta-data dir,
to avoid meta-data loss when `query` is restarted.

Let `query` create a process-wise global meta-store if the remote meta-store address is empty.
Then sub routines use this global meta-store.

Unit tests do not initialize such a process-wise global meta-store thus they still build a meta-store backed by a temp-dir.

## Changelog

- New Feature

## Related Issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1392,6 +1392,7 @@ dependencies = [
  "common-tracing",
  "derive_more",
  "futures",
+ "lazy_static",
  "maplit",
  "pretty_assertions",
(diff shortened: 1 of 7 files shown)


BohuTANG (approved):
Love it

PR #3197, opened 2021-12-01
ISSUE:3175 fix try_from_literal

### LINKED ISSUE #3175: Numbers in the `u64` range are converted to `Float64`, resulting in loss of precision
**Summary**

When I test the `bin` function, and exec it:
```
select bin(13835058055282163712);
select bin(-9223372036854775808);
```
`13835058055282163712` in the range of u64, `-9223372036854775808` in the range of i64;
but in the `DataColumnsWithField`, when i check the DataType, it tell me the data type is `Float64`, and when I checked the values, I found that they did not correspond:
```
13835058055282163712 =>  1.3835058055282164e19 = 13835058055282164000
-9223372036854775808 => -9.223372036854776e18 = -9223372036854776000
```
I am using the `BIGINT` data type in Mysql

dust1:

## Summary

I found the problem is in the try_from_literal function of DataValue, some u64 data will be lost when converting `str` to i64

## Changelog

- Bug Fix

## Related Issues

Fixes #3175 

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
maybe we can add a stateless test to cover the issue #3175

dust1:
ok.

dust1:
hello @sundy-li , I have a question.
when i exec `select -1;`, I find the data type is `Int16` in `PlanNode`,it should return to me Int8.
`Expr` has extracted ‘-’, so the parameter to `DataValue:: try_from_literal` is unsigned, the result type was UInt8, and then I found code about display PlanNode to find why, i found this:
https://github.com/datafuselabs/databend/blob/335293a66f62d581fba551766dd9ccdf69df165f/common/datavalues/src/types/data_type_coercion.rs#L184-L190

I have no database-related experience, so I can’t be sure if this is a bug, this will lose accuracy when facing data between `i32::MAX ~ i64::MAX`.
Does this mean that databend will expand the range upward when it gets the type of a signed integer?

dust1:
hello @sundy-li , I have a question.
when i exec `select -1;`, I find the data type is `Int16` in `PlanNode`. I have no database work experience, so I can’t be sure if this is a bug, does this mean that databend will expand the range upward when it gets the type of a signed integer? 

`Expr` has extracted ‘-’, so the parameter to `DataValue:: try_from_literal` is unsigned, the result type was UInt8, and then I found code about display PlanNode to find why, i found this:
https://github.com/datafuselabs/databend/blob/335293a66f62d581fba551766dd9ccdf69df165f/common/datavalues/src/types/data_type_coercion.rs#L286-L293

sundy-li:
> `select -1` returns `Int16`

This is a bug, I think it's related to sqlparser,  `-` is extracted as operator, we may need to modify the sqlparser to let `-1` be represented as `literal value`.

dust1:
To complete this work, do I need to submit pr in [datafuse-extras/sqlparser-rs](https://github.com/datafuse-extras/sqlparser-rs) ?

sundy-li:
> To complete this work, do I need to submit pr in datafuse-extras/sqlparser-rs ?

That's ok, it's a small library. You can hold it.

dust1:
ok

dust1:
hello @sundy-li , I read the code related to [datafuse-extras/sqlparser-rs](https://github.com/datafuse-extras/sqlparser-rs) and looked up [articles](https://eli.thegreenplace.net/2012/08/02/parsing-expressions-by-precedence-climbing) related to sqlparser.
I found that if the value in `Number(String, bool)` is used as a signed integer to generate `Expr` like `-1`, I can’t control the consequences of the modification with my current knowledge, this processing method seems to be different from the existing sqlparser algorithm.

I rewritten some stateless test again , The sql result works fine when use unsigned integers:
```
select toTypeName(a) from (select 255 as a);                                                   => UInt8
select toTypeName(a) from (select 65535 as a);                                               => UInt16
select toTypeName(a) from (select 4294967295 as a);                                     => UInt32
select toTypeName(a) from (select 18446744073709551615 as a);                 => UInt64
select toTypeName(a) from (select 1844674407370955161522 as a);             => Float64
```
This problem may be limited to `ArithmeticFunction `, maybe we only need to fix part of the code related to `Minus`

sundy-li:
Related codes:
```
            Token::Mul => Ok(Expr::Wildcard),
            tok @ Token::Minus | tok @ Token::Plus => {
                let op = if tok == Token::Plus {
                    UnaryOperator::Plus
                } else {
                    UnaryOperator::Minus
                };
                Ok(Expr::UnaryOp {
                    op,
                    expr: Box::new(self.parse_subexpr(Self::PLUS_MINUS_PREC)?),
                })
            }
```

common/datavalues/src/data_value_ops.rs
```diff
@@ -29,29 +29,48 @@ impl DataValue {
...
+                    } else if n == i32::MAX as u64 + 1 {
+                        return Ok(DataValue::Int32(Some(n as i32)));
+                    } else if n == i64::MAX as u64 + 1 {
+                        return Ok(DataValue::Int64(Some(n as i64)));
+                    }
+
+                    let num = -(n as i64);
+                    if num >= i8::MIN as i64 {
+                        return Ok(DataValue::Int8(Some(num as i8)));
+                    } else if num >= u16::MIN as i64 {
```
sundy-li, line 57: i16, i32, i64

dust1: all stateless test is ok. do I need to submit changes to the sqlparser version in this pr?

common/datavalues/src/data_value_ops.rs
```diff
@@ -29,29 +29,48 @@ impl DataValue {
...
-                if n >= 0 {
-                    let n = literal.parse::<u64>()?;
-                    if n <= u8::MAX as u64 {
-                        return Ok(DataValue::UInt8(Some(n as u8)));
-                    } else if n <= u16::MAX as u64 {
-                        return Ok(DataValue::UInt16(Some(n as u16)));
-                    } else if n <= u32::MAX as u64 {
-                        return Ok(DataValue::UInt32(Some(n as u32)));
+                if negative_flag {
+                    // when the real number is i8::MIN or i16::MIN or i32::MIN or i64::MIN
```
sundy-li, line 41: if it's zero, use u8;

if it's negative, use i64 and check the ranges.

if it's positive, use u64 and check the ranges.

refer: https://github.com/ClickHouse/ClickHouse/blob/f2374a6916295c236abe9029d0c5b8cbeb3c8d9b/src/DataTypes/FieldToDataType.cpp

common/datavalues/src/data_value_ops.rs
sundy-li: Yes

common/datavalues/src/data_value_ops.rs
sundy-li: Better to simplify this.

dust1: ok

sundy-li:
/LGTM,  but there are conflicts to merge.

dust1:
> /LGTM, but there are conflicts to merge.

should i use `git rebase` to merge this pr? I haven't touched before

sundy-li:
> should i use git rebase to merge this pr? I haven't touched before

`git pull --rebase origin main`  then fix the conflicts.

BohuTANG:
@mergifyio update

BohuTANG:
image
I guess the rebase has some problems here

BohuTANG:
> image I guess the rebase has some problems here

It's OK now :)

Issue #1784, opened 2021-09-10
Enable caching in  table data readers 


dantengsky:

Veeupup:
"cache" mentioned here is in memory or on disk(on query node SSD)?

flaneur2020:
> "cache" mentioned here is in memory or on disk(on query node SSD)?

good question 🤔

dantengsky:
> "cache" mentioned here is in memory or on disk(on query node SSD)?

ideally, there should be tiers of cache, (1) for hot data, like frequently accessed table metadata and blocks, they should be cached in memory, (2) for cold(relatively) data, which evicted from the hot tier, put them on disk (SSD of computing node).

but, we can go with the in-memory cache first and improve the cache layer step by step.

PR #3201, opened 2021-12-02
add cache for local data accessor

Veeupup:

## Summary

cache object in local disk. s3/azure maybe like this way.
and maybe we can cache them in different types, such as meta data and user data, thay should cached in different caches.
it is a draft pr from #1784 

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/dal/src/accessors/local.rs
+++ b/common/dal/src/accessors/local.rs
@@ -21,6 +21,9 @@ use std::path::PathBuf;
 
 use async_compat::CompatExt;
 use common_base::tokio;
+use common_base::tokio::sync::RwLock;
+use common_cache::Cache;
+use common_cache::LruCache;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use futures::Stream;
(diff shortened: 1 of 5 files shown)


common/dal/src/accessors/local.rs
```diff
@@ -21,6 +21,9 @@ use std::path::PathBuf;
 
 use async_compat::CompatExt;
 use common_base::tokio;
+use common_base::tokio::sync::RwLock;
+use common_cache::Cache;
+use common_cache::LruCache;
```
PsiACE, line 26: `LruCache` is a in-memory cache, `LruDiskCache` is the disk cache.

common/cache/src/cache/lru.rs
```diff
@@ -695,3 +696,13 @@ impl<'a, K, V> ExactSizeIterator for IterMut<'a, K, V> {
         self.0.len()
     }
 }
+
+// count bytes for data accessors
+impl<K, T> Meter<K, Vec<T>> for Vec<u8> {
```
PsiACE, line 701: This bytesmeter should be placed in the `meter` mod.

common/dal/src/accessors/local.rs
Veeupup: yes, we could have cache policies to decide which kind of data should be cached in mem or disk and when to do it.

common/cache/src/cache/lru.rs
Veeupup: u are right.. will fix it.

common/dal/src/accessors/local.rs
PsiACE: > yes, we could have cache policies to decide which kind of data should be cached in mem or disk and when to do it.

This means that a module for cache management is required.

dantengsky:
Hi @Veeupup, thanks for firing up this pr.

besides the comments that @PsiACE(the author of cache crate) gives, I am also not sure if we should embed caches inside some concrete DataAccessor.

Veeupup:
> 

Do you mean we could have the separate mod to cache all the things? maybe a global cache layer in query node that all crate can share?

dantengsky:
let's keep this PR open for a while. pls let me contact you later today

PR #3212, opened 2021-12-02
try add reuse about workflow

flaneur2020:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


.github/workflows/unit-tests.yml
```diff
@@ -29,15 +29,7 @@ jobs:
...
 
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v1
-
-      - name: Rust setup
-        run: |
-          bash ./scripts/setup/dev_setup.sh
-
-      - name: Cache Rust
-        uses: Swatinem/rust-cache@v1
```
PsiACE, line 40: At least the cache for unit tests should be placed separately.

flaneur2020: thank you for reminding, i'd try another way allowing different cache strategy

PsiACE:
It looks like we need to provide a mysql client for macos to perform stateless tests.

.github/actions/setup-rust/action.yaml
```diff
@@ -0,0 +1,12 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Install Protoc
+      uses: arduino/setup-protoc@v1
+
+    - name: Rust setup
+      shell: bash
+      run: |
+        bash ./scripts/setup/dev_setup.sh
```
PsiACE, line 12: Support for installing mysql is required

.github/actions/setup-rust/action.yaml
```diff
@@ -0,0 +1,12 @@
+name: 'Setup rust'
+description: 'Setup rust environment for databend'
+runs:
+  using: "composite"
+  steps:
+    - name: Install Protoc
+      uses: arduino/setup-protoc@v1
```
PsiACE, line 7: The action queries the GitHub API to fetch releases data, to avoid rate limiting, pass the default token with the repo-token variable

```suggestion
    - name: Install Protoc
      uses: arduino/setup-protoc@v1
      with:
        repo-token: ${{ secrets.GITHUB_TOKEN }}
```

PsiACE: > API rate limit exceeded for 13.105.49.52. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

.github/actions/setup-rust/action.yaml
PsiACE: > FileNotFoundError: [Errno 2] No such file or directory: 'mysql'

PsiACE: `    - uses: ankane/setup-mysql@v1` It may be possible to solve this problem

.github/actions/setup-rust/action.yaml
flaneur2020: thank you for reminding, i've changed as you suggested

PsiACE (approved):
LGTM

ZhiHanZ (approved):

PR #3214, opened 2021-12-02
[functions] string CONCAT() finished

### LINKED ISSUE #3020: CONCAT() | Return concatenated string

Veeupup:

## Summary

string function CONCAT() finished

## Changelog

- Improvement

## Related Issues

Fixes #3020 

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/strings/concat.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        let mut strs: Vec<u8> = Vec::new();
+        for column_field in columns {
+            if column_field.data_type().is_null() {
+                return Ok(DataColumn::Constant(DataValue::Null, 1));
+            }
+            match column_field
+                .column()
+                .cast_with_type(&DataType::String)?
+                .try_get(0)?
```
sundy-li, line 73: It's just the first element?

tests/suites/0_stateless/02_0028_function_strings_concat.sql
```diff
@@ -0,0 +1,4 @@
+SELECT CONCAT('My', 'S', 'QL');
+SELECT CONCAT('My', NULL, 'QL');
+SELECT CONCAT(14.3);
+SELECT CONCAT(14.3, 'SQL');
```
sundy-li, line 4: `select CONCAT(number, '3',  number + 1) from numbers(3)`

common/functions/src/scalars/strings/concat.rs
Veeupup: each column only has one element as part of string, may have many columns to concat

sundy-li: Yes, but here is a column vectorized calculation. Everything is columnable.

BohuTANG:
This PR has some conflicts, need rebase :D

common/functions/src/scalars/strings/concat.rs
```diff
@@ -62,28 +62,33 @@ impl Function for ConcatFunction {
...
-            match column_field
+            let values = column_field
                 .column()
                 .cast_with_type(&DataType::String)?
-                .try_get(0)?
-            {
-                DataValue::String(op) => {
-                    if let Some(data) = op {
-                        strs.extend_from_slice(&data);
+                .to_values()?;
```
sundy-li, line 74: to_values is very expensive.

common/functions/src/scalars/strings/concat.rs
```diff
@@ -62,28 +62,33 @@ impl Function for ConcatFunction {
...
+            for (idx, value) in values.iter().enumerate() {
+                match value {
+                    DataValue::String(op) => {
+                        if let Some(val) = op {
+                            data[idx].extend_from_slice(val);
+                        }
+                    }
+                    value => {
+                        data[idx].extend_from_slice(value.to_string().as_bytes());
```
sundy-li, line 83: if value is null, then the result should be null.

Veeupup:
@BohuTANG  ci test in cluster mode seems strange to fail.

common/functions/src/scalars/strings/concat.rs
```diff
@@ -0,0 +1,142 @@
...
+                    l_array
+                        .into_no_null_iter()
+                        .zip(r_array.into_no_null_iter())
+                        .map(|(l, r)| [l, r].concat()),
+                    combine_validities(l_array.inner().validity(), r_array.inner().validity()),
+                )
+            }
+            (DataColumn::Array(l_array), DataColumn::Constant(rhs, len)) => {
+                let l_array = l_array.string()?;
+                let r_array = rhs.to_series_with_size(len)?;
```
sundy-li, line 69: `rhs` does not need to resize to `r_array`, just get `Option<String>` from `DataValue::String`

Veeupup: use `StringArrayBuilder` in this scenario, maybe in this way?
```
(DataColumn::Array(lhs), DataColumn::Constant(rhs, len)) => {
                let l_array = lhs.string()?;
                let r_value = match rhs {
                    DataValue::String(val) => val,
                    _ => None,
                };
                let mut builder = StringArrayBuilder::with_capacity(len);
                let mut iter = l_array.into_iter();
                loop {
                    match iter.next() {
                        Some(val) => {
                            if val.is_none() || r_value.is_none() {
                                builder.append_null();
                            } else {
                                let val = val.unwrap();
                                let val2 = r_value.as_ref().unwrap().as_slice();
                                builder.append_value([val, val2].concat());
                            }
                        }
                        None => break,
                    }
                }
                let array = builder.finish();
                DataColumn::Array(array.into_series())
            }
```

sundy-li: If `r_value` is null, just return `DFString::full_null`.

Veeupup: resolved.

sundy-li (approved):

PR #3218, opened 2021-12-03
[bench] build the SessionManager from the default config

PsiACE:

## Summary

build the SessionManager from the default config, eliminate possible warnings and decouple from `SessionManagerBuilder`.

## Changelog

## Related Issues

Related #1866

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/benches/suites/mod.rs
+++ b/query/benches/suites/mod.rs
@@ -16,9 +16,10 @@ use common_base::tokio;
 use common_exception::Result;
 use common_planners::PlanNode;
 use criterion::Criterion;
+use databend_query::configs::Config;
 use databend_query::interpreters::SelectInterpreter;
+use databend_query::sessions::SessionManager;
 use databend_query::sql::PlanParser;
-use databend_query::tests::SessionManagerBuilder;
 use futures::StreamExt;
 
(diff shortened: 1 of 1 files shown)


BohuTANG:
Hi @PsiACE ,
Why move testing purposes function to `query/src/sessions/sessions_builder.rs`?

PsiACE:
> Hi @PsiACE , Why move testing purposes function to `query/src/sessions/sessions_builder.rs`?

On the one hand, it is also used in bench, which always prompts a warning in vscode, and this move eliminates this problem.

On the other hand, this way, we can easily move the tests dir to tests/it to suit the new testing style.

Although it appears to be for testing purposes only, it makes sense to provide such a hack.

BohuTANG:
I think `SessionManagerBuilder` is only for test, and some hacks in it like:
```
  pub fn create() -> SessionManagerBuilder {
        SessionManagerBuilder::inner_create(Config::default())
            .log_dir_with_relative("../tests/data/logs") -- here
    }
```
It is not a normal function, if we want to build the normal `SessionManager` we should use:
`SessionManager::from_conf` instead of `SessionManagerBuilder`

PR #3223, opened 2021-12-03
[common/function] refactor: simplify get_layout_offsets

drmingdrmer:

## Changelog

- Improvement

## Related Issues

### DIFF
--- a/common/functions/src/aggregates/aggregate_function_state.rs
+++ b/common/functions/src/aggregates/aggregate_function_state.rs
@@ -100,25 +100,21 @@ impl From<StateAddr> for usize {
 /// # Safety
 /// layout must ensure to be aligned
 pub unsafe fn get_layout_offsets(funcs: &[AggregateFunctionRef]) -> (Layout, Vec<usize>) {
-    let mut align_aggregate_states = 0;
-    let mut total_size_aggregate_states = 0;
+    let mut max_align = 0;
+    let mut total_size = 0;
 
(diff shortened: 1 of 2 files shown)


common/functions/src/aggregates/aggregate_function_state.rs
```diff
@@ -100,25 +100,21 @@ impl From<StateAddr> for usize {
...
+    for func in funcs {
+        let layout = func.state_layout();
+        let align = layout.align();
+
+        total_size = (total_size + align - 1) / align * align;
+
+        offsets.push(total_size);
+
+        max_align = max_align.max(align);
+        total_size += layout.size();
```
sundy-li, line 116: 「size: 2, align: 8」

total_size will be `3`, max_align will be `8` ?

drmingdrmer: Hmmm? the result should be `total_size=2, max_align=8`?

That's just what the former version does in a simpler way.

drmingdrmer: Did you mean that the size should not be smaller than align or something else?

sundy-li: > Hmmm? the result should be total_size=2, max_align=8? That's just what the former version does in a simpler way.

Aha, you are right, I was wrong.

sundy-li (approved):

PR #3228, opened 2021-12-03
Complete dates function monotonicity check

zhyass:

## Summary

Add `FACTOR_OP`, check the factor eval value to get monotonicity.

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests


zhyass:
PTAL @junli1026 @sundy-li @zhang2014

sundy-li:
What does `FACTOR_OP`mean, I did not understand this change.

zhyass:
> What does `FACTOR_OP`mean? I did not understand this change, maybe we should add more comments on it.

 `FACTOR_OP`means factor operator. Used to check the monotonicity of the function.
For example, `ToDayOfYear` is NOT a monotonic function in general, unless the time range is within the same year. So we can use the factor operator `toStartOfYear` to check whether the range[left, right] are in the same year.

sundy-li (approved):

PR #3210, opened 2021-12-02
website: Migrate site to docusaurus

Xuanwo:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

This PR is a part of https://github.com/datafuselabs/databend/issues/3209

Please review this PR here: https://databend.vercel.app/

## Migrate plan

For now, I think there should be three stages:

image

### Stage 1

Stage before this PR gets merged.

At stage 1, we need to:

- [x] Fix markdown format
- [x] Fix social media and resource links (as @PsiACE said)
- [x] Figure out how to use docusaurus and add them into docs.
- [x] Catch up all changes to current docs (**heavy work here**)

### Stage 2

This PR getting merged means we already changed the databend website framework. However, until we change the `databend.rs` DNS to vercel, this change will not take effect.

- People will still see the site built by MkDocs hosted at `gh-pages`
- All changes to docs will not show up until we get into next stage.

At stage 2, we need to:

- Mirror the current design. (**I need help**)
- Figure out how to organize content (as @PsiACE said)

### Stage 3

The final stage. After DNS is changed, the site will be online and replace the current site.

At stage 3, we need to:

- Have a front-end design (as @BohuTANG said)

## Changelog

- Documentation

Issue #3209, opened 2021-12-02
Tracking issue of use docusaurus to build our website


Xuanwo:
This is a tracking issue of discussion in https://github.com/datafuselabs/databend/discussions/3099

<div type='discussions-op-text'>

<sup>Originally posted by **Xuanwo** November 25, 2021</sup>
Our current website framework is [mkdocs](https://github.com/mkdocs/mkdocs) which is lacking blog support. So I propose to use [docusaurus](https://docusaurus.io/) to build our website.

Docusaurus is maintained by Facebook Opensource, and is widely used (see showcase: https://docusaurus.io/showcase). It's easy to add a new module like `blog` and so on. Also, it has built-in multi-version support, i18n support. And we can extend it via writing a new React module.

More, maybe we can apply for vercel open-source sponsor so that we can deploy on vercel instead of github pages. After that, we can remove the `gh-pages` branch and the `github pages` workflow, we can have a PR preview of our website to avoid broken changes.</div>

## Roadmap

For now, I think there should be three stages:

![image](https://user-images.githubusercontent.com/5351546/144538030-ac9f542f-c12d-412f-9985-eb447f669963.png)

**Current Stage: 2**

### Stage 1

Stage before https://github.com/datafuselabs/databend/pull/3210 gets merged.

At stage 1, we need to:

- [x] Fix markdown format
- [x] Fix social media and resource links (as @PsiACE said)
- [x] Figure out how to use docusaurus and add them into docs.
- [x] Catch up all changes to current docs (**heavy work here**)

### Stage 2

https://github.com/datafuselabs/databend/pull/3210 getting merged means we already changed the databend website framework. However, until we change the `databend.rs` DNS to vercel, this change will not take effect.

- People will still see the site built by MkDocs hosted at `gh-pages`
- All changes to docs will not show up until we get into next stage.

At stage 2, we need to:

- [x] Mirror the current design. https://github.com/datafuselabs/databend/pull/3443
- [x] https://github.com/datafuselabs/databend/issues/3275
- [x] Figure out how to deploy the site
- [x] Only build PR preview while files under `/website` changed (implemented via vercel builtin features)
- [ ] https://github.com/datafuselabs/databend/pull/3585

### Stage 3

The final stage. After DNS is changed, the site will be online and replace the current site.

At stage 3, we need to:

- Have a front-end design (as @BohuTANG said)
- [ ] Write the first blog (maybe?)

## Preview

Take https://databend.vercel.app/ for preview.

[...]

PR #3233, opened 2021-12-04
Don't Merge: This is a demo for vercel

Xuanwo:


Xuanwo:
ping @PsiACE for a look for the vercel's native PR preview support.

PsiACE:
wow, love it

ZhiHanZ:
seems out of topic... Vercel seems to be an interesting tool for frontend prototyping, what is the difference between vercel and netlify github action?

Xuanwo:
@ZhiHanZ 

There are two things here: Website hosting and PR preview. Both vercel, netlify and gh-pages all provide support for website hosting. However, not all platform has support for per-PR preview.

- Vercel has native support (as in this PR)
- Netlify didn't have native support ([but now they have](https://www.netlify.com/products/deploy-previews/)), so the community built an action [Netlify Actions](https://github.com/marketplace/actions/netlify-actions) to fix it.
- Github Pages don't have a PR preview and it's hard to implement (because GitHub only serve content from the `gh-pages` branch).

I will start a discussion about all possible plans so that we can come a proposal to decide how we deploy our site. I expect to finish the job during stage 2 as described in #3210.

ZhiHanZ:
> @ZhiHanZ
> 
> TL;DR
> 
> They function similarly and both work.
> 
> There are two things here: Website hosting and PR preview. Both vercel, netlify and gh-pages all provide support for website hosting. However, not all platform has support for per-PR preview.
> 
> * Vercel has native support (as in this PR)
> * Netlify didn't have native support ([but now they have](https://www.netlify.com/products/deploy-previews/)), so the community built an action [Netlify Actions](https://github.com/marketplace/actions/netlify-actions) to fix it.
> * Github Pages don't have a PR preview and it's hard to implement (because GitHub only serve content from the `gh-pages` branch).
> 
> I will start a discussion about all possible plans so that we can come a proposal to decide how we deploy our site. I expect to finish the job during stage 2 as described in #3209.

Thanks for clarification! on possible improvement on vercel is to make PR preview updates only when files under /websites got changed.

Xuanwo:
> one possible improvement on vercel is to make PR preview updates only when files under /websites got changed, maybe we could incorporate it into tracking issue

Makes sense, updated.

Xuanwo:
> one possible improvement on vercel is to make PR preview updates only when files under /websites got changed, maybe we could incorporate it into tracking issue

Implemented, it's a vercel builtin feature.

image

sundy-li:
Great tool, IMO `vercel` bot can be disabled if nothing changes related to the doc.

Xuanwo:
> Great tool, IMO `vercel` bot can be disabled if nothing changes related to the doc.

Oh, I need to figure it out. There is no obvious way to do that.

Issue #2978, opened 2021-11-23
Add system.query_log


BohuTANG:
## Summary

`system.query_log` is a read-only in memory table, it stores all the query logs under the schema:

```
mysql> desc query_log;
+---------------------+------------+------+
| Field               | Type       | Null |
+---------------------+------------+------+
| type                | Int8       | NO   |
| tenant_id           | String     | NO   |
| cluster_id          | String     | NO   |
| sql_user            | String     | NO   |
| sql_user_privileges | String     | NO   |
| sql_user_quota      | String     | NO   |
| client_address      | String     | NO   |
| query_id            | String     | NO   |
| query_text          | String     | NO   |
| query_start_time    | DateTime32 | NO   |
| query_end_time      | DateTime32 | NO   |
| written_rows        | UInt64     | NO   |
| written_bytes       | UInt64     | NO   |
| read_rows           | UInt64     | NO   |
| read_bytes          | UInt64     | NO   |
| result_rows         | UInt64     | NO   |
| result_result       | UInt64     | NO   |
| memory_usage        | UInt64     | NO   |
| cpu_usage           | UInt32     | NO   |
| exception_code      | Int32      | NO   |
| exception           | String     | NO   |
| client_info         | String     | NO   |
| current_database    | String     | NO   |
| databases           | String     | NO   |
| columns             | String     | NO   |
| projections         | String     | NO   |
| server_version      | String     | NO   |
+---------------------+------------+------+
```

This issue mainly for the system.query_log implemention,  the data source of the query_log table will be addressed in another issue.

BohuTANG:
cc @ZhiHanZ @flaneur2020 
If some columns needed, feel free to add it ...

BohuTANG:
@SGZW Thank you for taking it

BohuTANG:
Hello,
Any progress on this task? If no progress I'd love to take it and released it in our 0.6 version:D

SGZW:
> Any progress on this task? If no progress I'd love to take it and released it in our 0.6 version:D

I will finish it this week，Is it too late?

BohuTANG:
> > Any progress on this task? If no progress I'd love to take it and released it in our 0.6 version:D
> 
> I will finish it this week，Is it too late?

That's ok, thank you

SGZW:
Is the querylog table bound to the session or a single query process?

BohuTANG:
> Is the querylog table bound to the session or a single query process?

It's a global table as others system table.

[...]

PR #3234, opened 2021-12-04
ISSUE #2978, Add system.query_log

SGZW:

## Summary

solve #2978 

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/configs/config_query.rs
+++ b/query/src/configs/config_query.rs
@@ -32,7 +32,7 @@ pub const QUERY_FLIGHT_API_ADDRESS: &str = "QUERY_FLIGHT_API_ADDRESS";
 pub const QUERY_HTTP_API_ADDRESS: &str = "QUERY_HTTP_API_ADDRESS";
 pub const QUERY_METRICS_API_ADDRESS: &str = "QUERY_METRIC_API_ADDRESS";
 pub const QUERY_WAIT_TIMEOUT_MILLS: &str = "QUERY_WAIT_TIMEOUT_MILLS";
-pub const MAX_QUERY_LOG_SIZE: &str = "MAX_QUERY_LOG_SIZE";
+pub const QUERY_MAX_QUERY_LOG_SIZE: &str = "QUERY_MAX_QUERY_LOG_SIZE";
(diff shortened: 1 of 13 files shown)


SGZW:
I'm sorry that this task was completed late

BohuTANG:
> I'm sorry that this task was completed late

Thanks @SGZW , this looks great.
This almost ready, unit test needs some fix :D

query/src/configs/config_query.rs
```diff
@@ -206,6 +207,14 @@ pub struct QueryConfig {
...
     #[serde(default)]
     pub wait_timeout_mills: u64,
+
+    #[structopt(
+    long,
+    env = MAX_QUERY_LOG_SIZE,
+    default_value = "10000"
+    )]
+    #[serde(default)]
+    pub max_query_log_size: usize,
```
BohuTANG, line 217: :+1:

sundy-li:
How about using fuse engine to replace memory engine? I think it's better to persist in the query logs.

SGZW:
> How about using fuse engine to replace memory engine? I think it's better to persist in the query logs.

it means multi query process can share query_log by fuse table?

SGZW:
> How about using fuse engine to replace memory engine? I think it's better to persist in the query logs.

I think we can learn from query_log's implement of clickhouse which pesist query log in merge tree Periodic ？ I can try this in another pr

BohuTANG:
> How about using fuse engine to replace memory engine? I think it's better to persist in the query logs.

We will add LRU to query_log later(The max rows is N, N default like 100000), memory engine is OK

SGZW:
> > How about using fuse engine to replace memory engine? I think it's better to persist in the query logs.
> 
> We will add LRU to query_log later(The max rows is N, N default like 100000), memory engine is OK

what is LRU mean? fifo is

BohuTANG:
> > > How about using fuse engine to replace memory engine? I think it's better to persist in the query logs.
> > 
> > 
> > We will add LRU to query_log later(The max rows is N, N default like 100000), memory engine is OK
> 
> what is LRU mean? FIFO isn't enough?

Sorry, FIFO.
It have implement in this patch

PR #3241, opened 2021-12-04
SUBSTRING_INDEX() func

### LINKED ISSUE #3033: SUBSTRING_INDEX() | Return a substring from a string before the specified number of occurrences of the delimiter

mshauneu:

## Summary

Impl of SUBSTRING_INDEX() func

## Changelog

- New Feature

## Related Issues

Fixes #3033 

## Test Plan

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/strings/substring_index.rs
+++ b/common/functions/src/scalars/strings/substring_index.rs
@@ -1,0 +1,261 @@
+// Copyright 2021 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt;
+
+use common_datavalues::prelude::*;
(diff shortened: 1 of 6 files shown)


website/databend/docs/sqlstatement/string-functions/substring_index.md
```diff
@@ -0,0 +1,43 @@
+---
```
BohuTANG, line 1: +1

common/functions/src/scalars/strings/substring_index.rs
```diff
@@ -0,0 +1,261 @@
...
+        }
+        Ok(DataType::String)
+    }
+
+    fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
+        let s_column = columns[0].column().cast_with_type(&DataType::String)?;
+        let d_column = columns[1].column().cast_with_type(&DataType::String)?;
+        let c_column = columns[2].column().cast_with_type(&DataType::Int64)?;
+
+        let r_column: DataColumn = match (s_column, d_column, c_column) {
```
sundy-li, line 84: That's great.

But I have some thoughts, how about forcing the third argument to be Constant?
I did not find any reason to use dynamic count.

Thus we can reduce lots of branches.

sundy-li (approved):

common/functions/src/scalars/strings/substring_index.rs
sundy-li: From ClickHouse doc:

https://github.com/ClickHouse/ClickHouse/blob/cf0fbbbc1704e8d232039547005424830a0c82d9/docs/en/sql-reference/functions/string-functions.md

PR #3255, opened 2021-12-05
SPACE()

### LINKED ISSUE #3040: SPACE() | Return a string of the specified number of spaces

mshauneu:

## Summary

Impl for SPACE() func

## Changelog

- New Feature

## Related Issues

Fixes #3040 

## Test Plan

Stateless Tests


common/functions/src/scalars/strings/space.rs
```diff
@@ -0,0 +1,100 @@
...
+use common_datavalues::prelude::*;
+use common_exception::ErrorCode;
+use common_exception::Result;
+
+use crate::scalars::function_factory::FunctionDescription;
+use crate::scalars::function_factory::FunctionFeatures;
+use crate::scalars::Function;
+
+#[derive(Clone)]
+pub struct SpaceFunction {
```
sundy-li, line 26: We can use `String2NumberFunction` instead.

common/functions/src/scalars/strings/space.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl fmt::Display for SpaceFunction {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.display_name)
+    }
+}
+
+#[inline]
+fn space(count: &u64) -> Vec<u8> {
```
sundy-li, line 98: better to reuse the buffer.

mshauneu: sorry didn't catch idea 
every time when new value come from const or series[i] I need a spaces with diff length...

mshauneu: can you explain more?

sundy-li: Refer to `Base64EncodeFunction` in the codes.

mshauneu: changed

sundy-li (approved):

Issue #3205, opened 2021-12-02
`table_engine_github_enabled` for disable github table engines


BohuTANG:
**Summary**

Introduce a new config for disable github table engines:

`table_engine_github_enabled`, default is true

PR #3276, opened 2021-12-06
[query] Add config for enable/disable github table engine

Junnplus:

## Summary

- Add `table_engine_github_enabled` config.
- Uniform env config for table engine.

## Changelog

- New Feature

## Related Issues

Fixes [#issue](https://github.com/datafuselabs/databend/issues/3205)

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
How about grouping these enable/disable as a group config? 

Such as  `struct Features` ?

zhang2014:
Why don't we use conditional compilation? such as: `cargo features`

BohuTANG:
> Why don't we use conditional compilation? such as: `cargo features`

Good idea.

Junnplus:
@zhang2014 @BohuTANG Do we need to support in this pr, or open an issue to discuss?

zhang2014:
Maybe we can merge this PR and then implement it in another pr? CC: @BohuTANG

PR #3330, opened 2021-12-09
[common/meta] refactor: add a trait Store to define local store API

drmingdrmer:

##### [common/meta] refactor: add a trait Store to define local store API
- Use AsTxnKeySpace::update_and_fetch() to impl seq number.

- Impl `Store` for `AsTxnKeySpace`.

- Remove update_and_fetch from TransactionSledTree. It is not acutally
  used.

## Changelog

- Improvement

## Related Issues


ariesdevil (approved):
look great to me

PR #3332, opened 2021-12-09
fix string2string

mshauneu:

## Summary

fix string2string

## Changelog

- Bug Fix

## Test Plan

Stateless Tests


BohuTANG:
Good catch, @Veeupup has found the bug and planning to fix it, thank you @mshauneu

mshauneu:
@BohuTANG - it is typo rather bug )

PR #3343, opened 2021-12-10
EXPORT_SET func()

### LINKED ISSUE #3023: EXPORT_SET() | Return a string such that for every bit set in the value bits, you get an on string and for every unset bit, you get an off string

mshauneu:

## Summary

Impl of EXPORT_SET func()

## Changelog

- New Feature

## Related Issues

Fixes #3023 

## Test Plan

Stateless Tests


tests/suites/0_stateless/02_0038_function_strings_export_set.sql
```diff
@@ -0,0 +1,106 @@
+SELECT '=== const ===';
```
BohuTANG, line 1: :+1: :+1:

common/functions/src/scalars/strings/export_set.rs
```diff
@@ -0,0 +1,1659 @@
...
+            DataColumn::Constant(DataValue::String(Some(vec![44])), r)
+        };
+        let n_column = if columns.len() > 4 {
+            columns[4].column().cast_with_type(&DataType::UInt64)?
+        } else {
+            DataColumn::Constant(DataValue::UInt64(Some(64)), r)
+        };
+
+        let r_column: DataColumn = match (b_column, e_column, d_column, s_column, n_column) {
+            // #00000
```
sundy-li, line 118: What a looooooooooooooong match.

BohuTANG:
@mergifyio update

common/functions/src/scalars/strings/export_set.rs
sundy-li: Hello, @mshauneu  really thanks for your effort to add this function.

I tried to review this function myself and I found it's hard to review and maintain/refactor the codes,  left some comments to improve this.

1. We need to force  `b_column`, `n_column` to be constant datavalue column because it covers most cases, returns errors if not.

2.  Then better to force `d_column` to be a constant datavalue column, it's unsound to have the delimiter be dynamic, returns errors if not.

Then we just cut down 2^5 = 32 branches to be  2 ^ 2 = 4 branches.

mshauneu: I was trying to follow mysql:
```
SELECT EXPORT_SET(number, number, number, number, number)  FROM numbers WHERE number < 5  ORDER BY number;
+----------------------------------------------------+
| EXPORT_SET(number, number, number, number, number) |
+----------------------------------------------------+
|                                                    |
| 1                                                  |
| 222                                                |
| 33333                                              |
| 4444444                                            |
+----------------------------------------------------+
```
Anyway It is easy to shrink than expand, so please take a look to last commit.

sundy-li: >  SELECT EXPORT_SET(number, number, number, number, number)  FROM numbers WHERE number < 5  ORDER BY number;

In this case, we can add another branch, which converts all columns into Series using `to_array`. It may be inefficient but cover all cases.

sundy-li (approved):

PR #3350, opened 2021-12-10
Allow set user for HTTP handler (without auth now).

youngsofun:

## Summary

1. HTTP handler allows set user in  JSON.
2. better Bendcli error handling when calling HTTP streaming load API.
3. better test for HTTP handler so we can see the real reason when CI fails.

postpone to maybe next pr:

1. choose the way to pass the user for the streaming load.
2. check user privilege and client address
3. user auth

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
need: make lint

PR #3356, opened 2021-12-10
Add `create table as select` syntax for the parser.

### LINKED ISSUE #3216: `CREATE TABLE table AS SELECT ...` statement
**Summary**

Create a table by selecting from an existing table (schema and data):
```
mysql> create table t1(a int);


mysql> insert into t1 values(1),(2);

mysql> create table t2 as select * from t1;

mysql> select * from t1;
+------+
| a    |
+------+
|    1 |
|    2 |
+------+

mysql> select * from t2;
+------+
| a    |
+------+
|    1 |
|    2 |
+------+

```

Reference:
https://dev.mysql.com/doc/refman/8.0/en/create-table-select.html

AngleNet:

## Summary

Support `create table as select ` syntax, such as `create table foo as select a, b from bar`

## Changelog

- New Feature

## Related Issues

Fixes #3216 

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
Conflicts needs resolved :)

AngleNet:
@databend-bot /help

AngleNet:
@BohuTANG cc please

AngleNet:
@BohuTANG cc please

query/src/sql/sql_parser.rs
```diff
@@ -1030,7 +1045,12 @@ impl<'a> DfParser<'a> {
                 break;
             }
             let name = name.unwrap();
-            self.parser.expect_token(&Token::Eq)?;
+            let eq = self.parser.expect_token(&Token::Eq);
```
sundy-li, line 1048: consume_token

AngleNet: Fixed

query/src/sql/sql_parser.rs
```diff
@@ -1045,8 +1045,7 @@ impl<'a> DfParser<'a> {
                 break;
             }
             let name = name.unwrap();
-            let eq = self.parser.expect_token(&Token::Eq);
-            if eq.is_err() {
+            if !self.parser.consume_token(&Token::Eq) {
                 // only paired values are considered as options
                 self.parser.prev_token();
```
sundy-li, line 1050: no need pre_token any more if using `consume_token`

AngleNet: I am recover the  _name_ that already been consumed since it will not be an option if the next token is not a _=_.

BohuTANG:
`make lint` will make the `Databend Base` CI happy

AngleNet:
> `make lint` will make the `Databend Base` CI happy

Fixed now. cc please

BohuTANG:
/lgtm

Thank you @AngleNet

query/src/sql/sql_parser.rs
sundy-li:
```
    fn consume_token(&mut self, expected: &str) -> bool {
        if self.parser.peek_token().to_string().to_uppercase() == *expected.to_uppercase() {
            self.parser.next_token();
            true
        } else {
            false
        }
    }
```

I do think it's buggy, if the next token is not a `=`, it will not call `next_token`.

`peek_token` will not advance the iterator.

PR #3385, opened 2021-12-11
[metasrv] test: write on follower then read on follower

drmingdrmer:

##### [metasrv] test: write on follower then read on follower
This is meant to ensure that a `query` node always works by connecting
to any of the metasrv nodes.

- Refactor: MetaNode should not try to join a cluster if the storage
  state is already initialized.

## Changelog

- Improvement
- Build/Testing/CI

## Related Issues


ariesdevil (approved):
look great to me~

PR #3393, opened 2021-12-12
[metasrv] refactor: rename AdminRequest to ForwardRequest

drmingdrmer:

##### [metasrv] refactor: rename AdminRequest to ForwardRequest
- make forward_to_leader a u64, to support deep forward.

## Changelog

- Improvement

## Related Issues

### DIFF
--- a/metasrv/src/executor/meta_handlers.rs
+++ b/metasrv/src/executor/meta_handlers.rs
@@ -52,8 +52,8 @@ use common_tracing::tracing;
 
 use crate::executor::action_handler::RequestHandler;
 use crate::executor::ActionHandler;
-use crate::meta_service::AdminRequest;
-use crate::meta_service::AdminRequestInner;
+use crate::meta_service::ForwardRequest;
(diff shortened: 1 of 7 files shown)


ariesdevil:
Personally, I like the previous boolean type of `forward_to_leader`, rest look good to me.

ariesdevil (approved):

drmingdrmer:
> Personally, I like the previous boolean type of `forward_to_leader`, rest look good to me.

There may be a case in which it can not find an effective leader by just one forward, i.e., when the forward target leader just lost the leadership.
Then a second forward is required to complete the operation.

Otherwise, the calling side(a `query` node) may be gonna have to deal with a `ForwardToLeader` error(although the odds is low).

Thus if a request should be forwarded, try to forward it two or three times would be better, IMHO. 🤔 

What say you man?

ariesdevil:
So it means we should have a retry mechanism, make sence.

PR #3391, opened 2021-12-12
[Release] bump cross patch

ZhiHanZ:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary
Update cross patch for new ci pipeline

## Changelog

- Bug Fix
- Build/Testing/CI
- Not for changelog (changelog entry is not required)

## Related Issues

resolve https://github.com/datafuselabs/databend/issues/3388

## Test Plan

Unit Tests

Stateless Tests

PR #3344, opened 2021-12-10
ci: Refactor the whole ci into stages

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

Fix #3320 

## Changelog

- Build/Testing/CI

PR #3394, opened 2021-12-12
ci: Fix build in databend-docker

Xuanwo:

## Summary

This PR is an extension of https://github.com/datafuselabs/databend/pull/3391, fix an extra place of cross.

## Changelog

- Build/Testing/CI

### DIFF
--- a/.github/workflows/databend-docker.yml
+++ b/.github/workflows/databend-docker.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Rust setup
         run: |
           bash ./scripts/setup/dev_setup.sh
-          cargo install --version 0.1.16 cross
+          cargo install cross
 
       - uses: ./.github/actions/cache-cargo-registry
         with:
--- a/.github/workflows/databend-docker.yml
+++ b/.github/workflows/databend-docker.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Rust setup
         run: |
           bash ./scripts/setup/dev_setup.sh
-          cargo install cross
+          cargo install --version 0.2.1 cross
 
       - uses: ./.github/actions/cache-cargo-registry
         with:
--- a/.github/workflows/databend-release.yml
+++ b/.github/workflows/databend-release.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Rust setup
         run: |
           bash ./scripts/setup/dev_setup.sh
-          cargo install --version 0.2.1 cross
+          cargo install cross
       - name: Build Binary
         run: |
           if [ ${{ matrix.config.os }} = "macos-11" ]; then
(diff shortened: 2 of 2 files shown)


.github/workflows/databend-release.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Rust setup
         run: |
           bash ./scripts/setup/dev_setup.sh
-          cargo install --version 0.2.1 cross
+          cargo install cross
```
ZhiHanZ, line 35: I think we should keep exact version number to get avoid of surprising breaking change

.github/workflows/databend-docker.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Rust setup
         run: |
           bash ./scripts/setup/dev_setup.sh
-          cargo install --version 0.1.16 cross
+          cargo install cross
```
Xuanwo, line 48:
```suggestion
          cargo install --version 0.2.1 cross
```

.github/workflows/databend-release.yml
Xuanwo:
```suggestion
          cargo install --version 0.2.1 cross
```

Xuanwo: Makes sense, addressed.

ZhiHanZ (approved):

BohuTANG:
Build perf tool failed, seems related to OpenSSL :

https://github.com/datafuselabs/databend/runs/4496828786?check_suite_focus=true

```
run pkg_config fail: "pkg-config has not been configured to support cross-compilation.\n\nInstall a sysroot for the target platform and configure it via\nPKG_CONFIG_SYSROOT_DIR and PKG_CONFIG_PATH, or install a\ncross-compiling wrapper for pkg-config and set it via\nPKG_CONFIG environment variable."

  --- stderr
  thread 'main' panicked at '

  Could not find directory of OpenSSL installation, and this `-sys` crate cannot
  proceed without this knowledge. If OpenSSL is installed and this crate had
  trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
  compilation process.

  Make sure you also have the development packages of openssl installed.
  For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.

  If you're in a situation where you think the directory *should* be found
  automatically, please open a bug at https://github.com/sfackler/rust-openssl
  and include information about your system as well as this message.

  $HOST = x86_64-unknown-linux-gnu
  $TARGET = aarch64-unknown-linux-gnu
  openssl-sys = 0.9.71

  ', /cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.71/build/find_normal.rs:180:5
warning: build failed, waiting for other jobs to finish...
error: build failed
Error: Process completed with exit code 101.


```

Xuanwo:
Wired,  I will figure it out tomorrow.

Xuanwo:
The problem is gone on the `main` branch, I believe there is something wrong with the `~/.cargo/bin` cache that failed workflows use the wrong `cross` binary.

Now all the build is OK and I'm working https://github.com/datafuselabs/databend/pull/3344 so that we can avoid those problems again.

PR #3404, opened 2021-12-12
Grant all privileges to the root user of localhost

### LINKED ISSUE #3348: 'root'@'localhost' superuser account for localhost
**Summary**

https://github.com/datafuselabs/databend/blob/85599e683e22f747dcb02fb12404a9cee0c33afd/query/src/users/user_mgr.rs#L32-L40

For the user name `default` and `root`, we should check the host is `localhost` or `127.0.0.1`, 'root'@'localhost' superuser account that has all privileges and can do anything.

silathdiir:

## Summary

Grant all privileges to the root user of localhost.

## Changelog

- Grant all privileges to `default` or `root` if hostname is `127.0.0.1` or `localhost`.

## Related Issues

Fixes #3348 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/users/user_mgr.rs
+++ b/query/src/users/user_mgr.rs
@@ -26,16 +26,25 @@ use crate::users::UserApiProvider;
 
 impl UserApiProvider {
     // Get one user from by tenant.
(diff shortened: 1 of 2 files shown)


BohuTANG:
/lgtm

Thank you!

PR #3422, opened 2021-12-13
[metasrv] make `logs_handler` return a bytes stream.

sunli829:
Bump poem from `1.0.38` to `1.1.1`

## Summary

This change will reduce some memory usage, but it will also reduce the performance of `logs_handler` slightly.

It also fixes incorrect results when the stream returns `Err`.

## Changelog

- Bug Fix
- Improvement

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/api/http/v1/logs.rs
+++ b/query/src/api/http/v1/logs.rs
@@ -14,11 +14,12 @@
 
 use std::sync::Arc;
 
-use common_datablocks::DataBlock;
+use common_exception::ErrorCode;
 use common_exception::Result;
 use common_streams::SendableDataBlockStream;
 use poem::http::StatusCode;
 use poem::web::Data;
+use poem::Body;
 use poem::IntoResponse;
 use tokio_stream::StreamExt;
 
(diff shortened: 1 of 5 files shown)


query/src/api/http/v1/logs.rs
```diff
@@ -38,15 +39,27 @@ pub async fn logs_handler(
...
+                let block = res?;
+                yield format!("{:?}", block);
+            },
+            None => return,
+        }
+
+        while let Some(res) = tracing_table_stream.next().await {
+            let block = res?;
+            yield format!(", {:?}", block);
+        }
```
drmingdrmer, line 59: Why does it need a `match` and a `while`? 
A single `while` is not enough? 🤔

sunli829: The final data will be separated by commas, which is the original behavior.

Maybe this comma separator doesn't make sense? 😂

```rust
let tracing_logs = tracing_table_stream
        .collect::<Result<Vec<DataBlock>>>()
        .await?;
Ok(format!("{:?}", tracing_logs))
```

drmingdrmer: I did not see the comma...

Apologize!

/lgtm

sundy-li: How it compare with 
```
if i > 0 {
  yield format!(", {:?}", block);
} else {
  i += 1;
  yield format!("{:?}", block);
}
```

BohuTANG:
Thank you @sunli829 !

PR #3433, opened 2021-12-13
Query cache

### LINKED ISSUE #1784: Enable caching in  table data readers

Veeupup:

## Summary

Add Query cache layer when reading object in fuse table.
we have 4 config in query_config:
* table_cache_enabled
* table_memory_cache_mb_size
* table_disk_cache_root : where to put disk cache files
* table_disk_cache_mb_size: disk cache max campacity

We only cache `snapshots` and `segments` on the query node now. (so the metadata can hold all in memory).

We have two tier of cache, when reading data, we will try to get data from `memory cache` -> `disk cache` -> `S3`.

When `memory cache` is full, we will put the evict item from `memory cache` to `disk cache`. And when `disk cache` is full, we will use LRU policy to evict extra items.

Now we have a trait `QueryCache` to abstract  `query cache` layer, we can add more effective cache implementations in the future.

We do have some simple metrics about cache statistics and we can get them from the metrics HTTP API.

##  dataset and config 
test dataset, `ontime` table, load data by https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateful/12_0001_streaming_load.sh

* table_memory_cache_mb_size: 512mb
* table_disk_cache_mb_size: 1024mb

## ec2 with s3 test
test with 3 SQL each 50 times,
```bash
#!/bin/bash

timestamp() {
  date +"%s" # current time
}

timestamp
for i in {1..50}
do
 echo $i
 curl -s  --request POST '127.0.0.1:8001/v1/statement/' --header 'Content-Type: text/plain' --data-raw 'SELECT DayOfWeek, count(*) AS c FROM ontime WHERE Year >= 2000 AND Year <= 2008 GROUP BY DayOfWeek ORDER BY c DESC;' > /dev/null
 curl -s  --request POST '127.0.0.1:8001/v1/statement/' --header 'Content-Type: text/plain' --data-raw 'SELECT Year, avg(DepDelay) FROM ontime GROUP BY Year;' > /dev/null
 curl -s  --request POST '127.0.0.1:8001/v1/statement/' --header 'Content-Type: text/plain' --data-raw 'SELECT DayOfWeek, count(*) AS c FROM ontime WHERE DepDelay>10 AND Year >= 2000 AND Year <= 2008 GROUP BY DayOfWeek ORDER BY c DESC;' > /dev/null
done
timestamp
```

time total
| No.        | cache with 2 layer | no cache |
| ------- | ------------------ | -------- |
| 1       | 18s                | 28s      |
| 2       | 18s                | 27s      |
| 3       | 19s                | 28s      |
| 4       | 17s                | 27s      |
| 5       | 19s                | 27s      |
| Average | 18.2s              | 27.2s    |

cache metadata is only `76KB` size, we can sure cache all meta data in memory. So it seems that S3 is very slow.

## Changelog

- Improvement

## Related Issues

Fixes #1784 

## Test Plan

Unit Tests

Stateless Tests


common/dal/src/cache.rs
```diff
@@ -0,0 +1,106 @@
...
+        buffer_bytes_usage: Arc<RwLock<u64>>,
+    ) {
+        loop {
+            while let Some(task) = input_tasks.write().await.pop_front() {
+                let mut cache = cache.write().await;
+                let _ = cache.insert_bytes(task.loc, task.data.as_slice());
+                *buffer_bytes_usage.write().await -= task.data.len() as u64;
+            }
+            // no task, sleep 5ms
+            sleep(Duration::from_millis(5)).await;
```
dantengsky, line 97: hi!  I am not sure if this PR is still in the draft or proof of concept stage, some “ahead of time“ thoughts here :) for your reference: 

- even the `sleep` here is `async`, it would be better if we could avoid sleeping. 
- if I get it right, the queue and `backend_cache_to_disk` arrangement are to prevent the user of the cache component from being blocked by the writing of the cache (in high load). 
  - can we use some async channel (bounded) here?
  - or we might leave this feature to another PR

common/dal/src/cache.rs
```diff
@@ -0,0 +1,98 @@
...
+impl DalCache {
+    pub fn create(cache: Arc<RwLock<LruDiskCache>>, tenant_id: String, cluster_id: String) -> Self {
+        Self {
+            cache,
+            tenant_id,
+            cluster_id,
+        }
+    }
+
+    pub async fn read<K: AsRef<OsStr>>(&self, loc: K, da: &dyn DataAccessor) -> Result<Vec<u8>> {
```
BohuTANG, line 49: It's better to use the full name: location: K

common/dal/src/cache.rs
```diff
@@ -0,0 +1,98 @@
...
+        let loc = loc.as_ref().to_owned();
+        let mut cache = self.cache.write().await;
```
BohuTANG, line 51: This write lock is not fine-grained, after the lock-write we do many things it may block other task here

BohuTANG:
Great job under the disk-based LRU cache!

There is a question:
Why we use a DiskLRU not a MemoryLRU?
For the cache design, the only goal is to faster the query and gain the high cache hits ratio.

The ideal two layer cache is:
1. Memory LRU is the first level, Disk LRU is the secondary level 
2. If the Memory LRU cache entry is evited, spill it to the Disk LRU

I assume that this PR is a proof of concept (POC) design not production ready.

common/dal/src/cache.rs
Veeupup: ok, will fix it. : D

dantengsky:
> The io path looks like : memory <-- disk <-- s3.

Agree, we have the same feelings during the discussion of this issue (#1784). 

For a fast prototype of (preliminary) performance testing, I suggested @Veeupup to begin with the Disk Cache (such that, the logic of  ` tiers of cache` can be ignored temporarily).

@Veeupup Is it possible to add the in-memory cache layer in this PR too?

Veeupup:
> > The io path looks like : memory <-- disk <-- s3.
> 
> Agree, we have the same feelings during the discussion of this issue (#1784).
> 
> For a fast prototype of (preliminary) performance testing, I suggested @Veeupup to begin with the Disk Cache (such that, the logic of ` tiers of cache` can be ignored temporarily).
> 
> @Veeupup Is it possible to add the in-memory cache layer in this PR too?

got it, I will make it as the `cache tier` with `memory <-- disk <-- s3`.
with `cache tier` mentioned above, `LruDiskCache` need a little improvement to support  it, I will try to fix it first.

query/src/sessions/context_shared.rs
```diff
@@ -71,8 +72,15 @@ impl QueryContextShared {
         conf: Config,
         session: Arc<Session>,
         cluster_cache: Arc<Cluster>,
-    ) -> Arc<QueryContextShared> {
-        Arc::new(QueryContextShared {
+    ) -> Result<Arc<QueryContextShared>> {
+        let dal_ctx = DalContext::create(
```
dantengsky, line 76: @zhang2014  need your help :)

Is it possible that there might be more than one instance of `QueryContextShared` co-existing in a single instance of query node?

the concern is :

`DalContext::create` here will also create a new instance of DalCache (if the cache is enabled), which uses the same disk path as cache dir. if multiple instances of `DalCache` write to the same file (and the file is large enough), we might read partially written cache.

dantengsky:
hi @Veeupup

- Seems that `DiskCache::init` is not called

   which makes the populated  caches un-usable (and may also bring some troubles for the calculations of `size` limitation)
   
    @PsiACE  
    
    is `DiskCache::init` supposed to be called after query node restarting?
    
- maybe we should re-consider the place to init the cache layer
 
   `SessionManager` seems to be a reasonable choice, where components like Catalog,  UserApiProvider inited.
   
    @BohuTANG any suggestions?

query/src/sessions/context_shared.rs
zhang2014: > Is it possible that there might be more than one instance of QueryContextShared co-existing in a single instance of query node?

No. Maybe we should move to session manager?

Veeupup:
> 

@PsiACE  disk cache seems no need to init ? as disk cache test usage here, @dantengsky https://github.com/datafuselabs/databend/blob/main/common/cache/tests/it/disk_cache.rs#L165

PsiACE:
> disk cache seems no need to init ?

`init()` is not a pub fn, it is already called on `new()`

dantengsky:
@Veeupup @PsiACE  my fault!

query/src/sessions/context_shared.rs
dantengsky: got it, thanks!

> Maybe we should move to SessionManager(single instance for single query node)?

yeah, I have the same feelings

Veeupup: @dantengsky ok, I'll put it in SessionManager.

common/exception/Cargo.toml
```diff
@@ -12,6 +12,7 @@ test = false
 
 [dependencies] # In alphabetical order
 common-arrow = { path = "../arrow" }
+common-cache = { path = "../cache" }
```
dantengsky, line 15: make `common::exception` depends on `common::cache` might not be a good idea.

common/exception/src/exception.rs
```diff
@@ -445,6 +448,12 @@ impl From<octocrab::Error> for ErrorCode {
     }
 }
 
+impl From<common_cache::DiskCacheError> for ErrorCode {
+    fn from(error: common_cache::DiskCacheError) -> Self {
+        ErrorCode::CacheInitFailed(format!("cache init failed, cause: {}", error))
```
dantengsky, line 453: not sure if it is proper to convert all the `DiskCacheError` into `ErrorCode::CacheInitFailed`, 
maybe we should do the conversion explicitly where cache initialization failed?

common/dal/src/cache.rs
```diff
@@ -0,0 +1,161 @@
...
+        let location = location.as_ref().to_owned();
+        let mut metrics = CacheDeferMetrics {
+            tenant_id: self.tenant_id.as_str(),
+            cluster_id: self.cluster_id.as_str(),
+            cache_hit: false,
+            read_bytes: 0,
+        };
+
+        // get data from memory cache
+        let mut mem_cache = self.mem_cache.write().await;
```
dantengsky, line 116: It would be better if we acquire a READ lock here. (and upgrade it to WRITE later if necessary)

common/exception/src/exception.rs
Veeupup: u r right, maybe we need do the conversion explicitly in different errors. 
And, maybe we can make conversation in `common::cache`, then `common::expection` will not depend `common::cache` ?
Do you have any suggestions ?

dantengsky: Seems the mapping from `DiskCacheError` to `ErrorCode` is not "static", the conversion may depend on context. I suggest starting with explicit conversion (e.g. map_err(....))

Veeupup: how about this way?
```rust
// common/cache/src/disk_cache.rs
impl From<Error> for ErrorCode {
    fn from(error: Error) -> Self {
        match error {
            Error::FileNotInCache => {
                ErrorCode::DiskCacheFileNotInCache("disk cache error: file not in cache")
            }
            Error::FileTooLarge => {
                ErrorCode::DiskCacheFileTooLarge("disk cache error: file too large")
            }
            Error::Io(err) => {
                ErrorCode::DiskCacheIOError(format!("disk cache io error, cause: {}", err))
            }
        }
    }
}

```

common/dal/src/cache.rs
Veeupup: because it is lru cache, it must be a WRITE lock and it will update inner state when reading.

common/exception/src/exception.rs
dantengsky: neat :)  @PsiACE would you mind making common-cache depend on common-exception?

common/dal/src/cache.rs
dantengsky: oops, sorry I did not notice that the signature of get is `fn get<Q>(&mut self, k: &Q) -> Option<&V>`

common/exception/src/exception.rs
PsiACE: > neat :) @PsiACE would you mind making common-cache depend on common-exception?

In fact, from my point of view, it looks strange. I have always regarded the common cache as a relatively independent module.

But it does look a bit neater this way.

dantengsky: I think it is an objection :D 

@Veeupup shall we go with the " explicit conversion (e.g. ad-hoc `map_err(....)`)" way?

Veeupup: using `map_err` will have some duplicate code, just like
```rust
...
        let disk_cache = Arc::new(RwLock::new(
            LruDiskCache::new(conf.disk_cache_root, conf.disk_cache_size_mb * 1024 * 1024).map_err(
                |err| match error {
                    Error::FileNotInCache => {
                        ErrorCode::DiskCacheFileNotInCache("disk cache error: file not in cache")
                    }
                    Error::FileTooLarge => {
                        ErrorCode::DiskCacheFileTooLarge("disk cache error: file too large")
                    }
                    Error::Io(err) => {
                        ErrorCode::DiskCacheIOError(format!("disk cache io error, cause: {}", err))
                    }
                },
            ),
        ));
...
        if disk_cache.contains_key(location.clone()) {
            let mut path = disk_cache.get(location).map_err(|err| match error {
                Error::FileNotInCache => {
                    ErrorCode::DiskCacheFileNotInCache("disk cache error: file not in cache")
                }
                Error::FileTooLarge => {
                    ErrorCode::DiskCacheFileTooLarge("disk cache error: file too large")
                }
                Error::Io(err) => {
                    ErrorCode::DiskCacheIOError(format!("disk cache io error, cause: {}", err))
                }
            });
            let data = read_all(&mut path)?;
...
```

can we accept the duplicate code like this?

PsiACE: > I think it is an objection :D

No need to be so serious, we can add it.

query/src/cache/local_cache.rs
```diff
@@ -0,0 +1,161 @@
...
+        let mut mem_cache = self.mem_cache.write().await;
+        if let Some(data) = mem_cache.get(&location) {
+            metrics.cache_hit = true;
+            metrics.read_bytes = data.len() as u64;
+
+            return Ok(data.clone());
+        }
+
+        // get data from disk cache
+        let mut disk_cache = self.disk_cache.write().await;
```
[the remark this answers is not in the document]
SGZW: Also，i think cache policy can be improved by another pr

dantengsky (approved):
LGTM, thanks @Veeupup, let's improve the cache layer step by step

PsiACE (approved):
Overall great. 

I think @SGZW makes a good point. We need to consider the flow from disk to mem.

PR #3452, opened 2021-12-14
Add expression nullable support

junli1026:

## Summary

Add expression nullable support.

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Stateless Tests


tests/suites/0_stateless/05_0000_ddl_create_tables.sql
```diff
@@ -19,15 +19,24 @@ DROP TABLE IF EXISTS t2;
...
+
+SELECT '====BEGIN TEST CREATE TABLE AS SELECT STATEMENT====';
 -- test 'create table as select' statement, expect db2.test3 has the data from db1.test1 with casting
 CREATE TABLE db2.test3(x Varchar, y Varchar) ENGINE=fuse AS SELECT * FROM db1.test1;
 SELECT x FROM db2.test3;
+SELECT '====END TEST CREATE TABLE AS SELECT STATEMENT====';
+
 -- clean up test databases
 DROP DATABASE db1;
 DROP DATABASE db2;
```
sundy-li, line 42: new line needed.

sundy-li:
Nullable is annoyment.

ClickHouse takes Nullable as a separated datatype.
Yet MySQL takes Nullable as a property for datatype.

So in the function, there are two kinds of ways to handle nullable:

1. if args are not nullable, the result should not be nullable except some special functions (isNull.. )
2. if args are not nullable, the result can be nullable.

I prefer the first one, thus we can control the return type.

sundy-li:
`query group by nullable(col)` may take much more memory/costs than `query group by col`.

junli1026:
> Nullable is annoyment.
> 
> ClickHouse takes Nullable as a separated datatype. Yet MySQL takes Nullable as a property for datatype.
> 
> So in the function, there are two kinds of ways to handle nullable:
> 
> 1. if args are not nullable, the result should not be nullable except some special functions (isNull.. )
Sure, will address. Do we have an example of the 'isNull' ? --  the inputs are all not nullable but output is null ? 

> 2. if args are not nullable, the result can be nullable.
> 
> I prefer the first one, thus we can control the return type.

sundy-li:
MySQL can produce nullable result by none nullable args.

 
image

junli1026:
> MySQL can produce nullable result by none nullable args.
> 
> image

Right, that is why I just check the function's nullable() return. My initial thoughts are just like, checking the input, if any of them is nullable, the output should be nullable.

sundy-li (approved):

BohuTANG:
Thank you @junli1026 !

PR #3463, opened 2021-12-14
issue-3462: historical data truncation

### LINKED ISSUE #3462: Truncate Table history
**Summary**

- Provides a table Function that can truncate table histories (historical data)
  user may truncate the historical data manually
- Enable historical data truncation for `drop table` and `truncate table`

dantengsky:

## Summary

-  Table Function that can truncate historical data of fuse table

   `select * from fuse_truncate_history(db, tbl [, 'all'])`

    count of removed snapshots, segments, and blocks will be returned as resultset. sth like:

      ```
        +------------------+-----------------+---------------+
        | snapshot_removed | segment_removed | block_removed |
        +------------------+-----------------+---------------+
        |                2 |               2 |             2 |
        +------------------+-----------------+---------------+
    ```
    it is not intended to be used by the end user directly yet
    
  - except for the data referenced by the latest snapshot, all the snapshots, segments, blocks will be removed
   
  - if the third arg is 'all', all data will be removed

- Enable historical data truncation for **drop** and **truncate** table

   - for `truncate`, **the latest snapshot will be kept**, i.e. `fuse_truncate_history(db, tbl)`
   - for `drop`, **all data will be removed**, i.e. ``fuse_truncate_history(db, tbl, 'all')`
   
   
NOTE:

 - it is an experimental feature, the behavior of physically removing data while `Truncate/Drop` table is likely to be disabled later.
 - operations are NOT transactional yet. to be addressed later.

## Changelog

- Improvement

## Related Issues

Fixes #3462

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/table_functions/fuse_history_truncate.rs
```diff
@@ -0,0 +1,275 @@
...
+        ])
+    }
+
+    async fn read(
+        &self,
+        ctx: Arc<QueryContext>,
+        _plan: &ReadDataSourcePlan,
+    ) -> Result<SendableDataBlockStream> {
+        if let Some((snapshot_removed, segment_removed, block_removed)) =
+            self.truncate_history(ctx).await?
```
BohuTANG, line 251: Why use the read trait to do the truncate operation? 

`select * from fuse_truncate_history('db_09_0007', 't');` --> `truncate table fuse_truncate_history('db_09_0007', 't') ` ?

sundy-li:
Can we have these statistics in `system.tables` ?

query/src/storages/fuse/table_functions/fuse_history_truncate.rs
dantengsky: or `fuse_truncate_history('db_09_0007', 't')` ?

dantengsky:
> Can we have these statistics in `system.tables` ?

the number of snapshots, segments, and blocks?

query/src/storages/fuse/table_functions/fuse_history_truncate.rs
dantengsky: `truncate table fuse_truncate_history('db_09_0007', 't')` is likely to be understood as table `fuse_truncate_history` is going to be truncated.

`call fuse_truncate_history('db_09_0007', 't')` might be better?

dantengsky:
> select is read-only query, why not make this write-related function be Alter QUERY ?

`alter table xxx truncate history `?

how about `call fuse_truncate_history(...)`?

query/src/storages/fuse/table_functions/fuse_history_truncate.rs
BohuTANG:
```
mysql> select * from fuse_history('default', 't1');
+----------------------------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
| snapshot_id                      | prev_snapshot_id                 | segment_count | block_count | row_count | bytes_uncompressed | bytes_compressed |
+----------------------------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
| fe6076fa4c444fd5b7089585c24f8025 | 9d95a99799d24d8984c41372e4b674f4 |             4 |           4 |         5 |                 20 |              775 |
| 9d95a99799d24d8984c41372e4b674f4 | 7cdd8c0aee364fd2a727eca6f0c506da |             3 |           3 |         4 |                 16 |              583 |
| 7cdd8c0aee364fd2a727eca6f0c506da | c158653e18f241979fc250facf1091bd |             2 |           2 |         3 |                 12 |              391 |
| c158653e18f241979fc250facf1091bd | NULL                             |             1 |           1 |         2 |                  8 |              199 |
+----------------------------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
4 rows in set (0.02 sec)
Read 4 rows, 465 B in 0.013 sec., 311.81 rows/sec., 36.25 KB/sec.

mysql> select * from fuse_truncate_history('default', 't1'); -- this likes truncate table fuse_history('default', 't1');
+------------------+-----------------+---------------+
| snapshot_removed | segment_removed | block_removed |
+------------------+-----------------+---------------+
|                3 |               0 |             0 |
+------------------+-----------------+---------------+
1 row in set (0.03 sec)
Read 1 rows, 24 B in 0.023 sec., 44.34 rows/sec., 1.06 KB/sec.

mysql> select * from fuse_history('default', 't1');
+----------------------------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
| snapshot_id                      | prev_snapshot_id                 | segment_count | block_count | row_count | bytes_uncompressed | bytes_compressed |
+----------------------------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
| fe6076fa4c444fd5b7089585c24f8025 | 9d95a99799d24d8984c41372e4b674f4 |             4 |           4 |         5 |                 20 |              775 |
+----------------------------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
1 row in set (0.03 sec)
Read 1 rows, 136 B in 0.013 sec., 74.98 rows/sec., 10.2 KB/sec.

mysql> select * from fuse_truncate_history('default', 't1');
Read 0 rows, 0 B in 0.010 sec., 0 rows/sec., 0 B/sec.

```

Not sure `fuse_truncate_history` table function is for? It seems `fuse_history` table can do all the operations what we did with the `fuse_truncate_history`.
Thanks.

dantengsky: `fuse_truncate_history` is a function with side effects, which
 
 -  truncate the historical data of the given table ( or kinda of GC)
     for an append-only table, like `t1` you given in the example, only  snapshot file(on disk/s3) will be removed, since the table is append-only, all the segments and blocks are still in use.
 - returns the number of snapshot/segment/block that has been removed.    
    
 try `insert overwrite t1 values(...)`, and then `select * from fuse_truncate_history(...)`.  here is an example:
 
 ~~~
 create table t(c int);
 insert into t select * from numbers(100);
 ~~~
 after the insertion , the storage dir may look like:
 
 ~~~
 stateless_test_data/
├── _b
│   ├── 4bdb25420b7949058b55c067028674d6.parquet
│   ├── 6c8375b2309a4a918e5a6c52bc0c4fb2.parquet
│   ├── ....
│   └── fda4f29d751f4c8d8afc93656f7dfeb8.parquet
├── _sg
│   ├── 051beb7eb7214276896ce446d52acc5d
│   ├── 27cba26cc2954315875f26d14aede9a5
│   ├── ...
│   └── f55416c3c6d74c1ea958f1a19a472bff
└── _ss
    └── 36d794938e12450f814c069c8df0ee33
 ~~~
 
 and then, we insert **OVERWRITE** and truncate
 
 ~~~
 insert overwrite t values(1);
select * from fuse_truncate_history('default', 't');
 ~~~   
 after that
 
 ~~~
stateless_test_data/
├── _b
│   └── f04eedbd25b14ab8b2dd60feb68cf299.parquet
├── _sg
│   └── ebb24d9071f34f1fab3a5981e9413602
└── _ss
    └── a1d51110980f4c9ea23faa3398d5bc49
 ~~~
 
 only the data that referenced by the last snapshot are kept
 
---------------



as stated in the summary, it is not intended to be used by the end user.

the front-end syntax has not been settled,  previously, I am planning to expose this functionality as `call` statement  (in other PR),  but any other suggestions are welcome.

the `select * from func` syntax is used by `interpreter_drop`, we could use `FuseTruncateHistory` directly in the interpreter, but exposing a fuse table specific concept into the interpreters module may not be a good idea.

dantengsky: > Why use the read trait to do the truncate operation?

forget the answer this :) :

- ideally, the GC process should be executed in a distributed manner. 
   the `read_plan`  and then distributedly  `read` mechanism is too tempting
-  promised  @wubx to provide a preview of `drop/truncate` with GC before Fri.
   unfortunately, seem there is not enough time to deliver a distributed version before Fri.
      
there must better ideas, pls let me know

dantengsky:
> Can we have these statistics in system.tables ?

Hope I get it right: 

it would be better if we could show the number of snapshots, segments, blocks by using stmt like `select .. from system.tables`. or even better, including the min/max statistics.

Yes, but 

- not all the tables have those statistics
- the performance might or might not be acceptable, depending on the usage
   those numbers are stored(or could be inferred) in storage like s3
   for the optimizer, imo, the performance might not be acceptable. for occasionally browsing,  maybe acceptable.

dantengsky:
@sundy-li  @BohuTANG 

how about this:

1. introduce two method to Trait `Table`
   - `capabilities() -> Set<Capability>`
   - `truncate_history(context, ...) -> Result<()>`
   
2. while executing the `Drop` and `Truncate` statement
     - if `truncate history` is enabled (by default , enabled)
     -  and   capabilities() contains  `Capability::TruncateHistory` 
        execute `table.truncate_history(...)`
 3. convert table function into table methods

BohuTANG:
/lgtm

We can improve it in later PR

PR #3466, opened 2021-12-14
implement function format()

### LINKED ISSUE #3028: FORMAT() | Return a number formatted to specified number of decimal places

eastfisher:

## Summary

implement function format()

## Changelog

- New Feature

## Related Issues

Fixes #3028

## Test Plan

Stateless Tests


tests/suites/0_stateless/02_0041_function_strings_format.sql
```diff
@@ -0,0 +1,14 @@
...
+SELECT FORMAT(-12332.123456, 4);
+SELECT FORMAT(12332.1, 4);
+SELECT FORMAT(12332.2, 0);
+SELECT FORMAT(12332.2, -1);
+SELECT FORMAT(12332, 2);
+SELECT FORMAT(0, 0);
+SELECT FORMAT(NULL, 1);
+SELECT FORMAT(1, NULL);
+SELECT FORMAT(NULL, NULL);
+SELECT FORMAT(12332.123456, 4, "en_US");
```
sundy-li, line 11: single quota

eastfisher: fixed

common/functions/src/scalars/strings/format.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+        Ok(false)
+    }
+
+    fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
+        match (
+            columns[0].column().cast_with_type(&DataType::String)?,
```
sundy-li, line 93: why not cast `columns[0]` into f64 column ?

sundy-li:
```

Name | Description
-- | --
N | A number which is to be formatted up to D decimal places rounded up.
D | A number indicating up to how many decimal places N will be rounded up.

 
```

format(N, D)

So the first arg N should be float64 column and the second arg D should be unsigned integer column.

eastfisher: If parse N or D error, default value is used. If N or D is NULL, the result is NULL.

```
SELECT FORMAT("hello", "world") -> '0'
SELECT FORMAT(5.25, "world") -> '5'
SELECT FORMAT("hello", 2) -> '0.00'
SELECT FORMAT(NULL, 2) -> NULL
SELECT FORMAT(5.25, NULL) -> NULL
```

Could you please give some advice?

```
    fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
        if columns[0].column().data_type().is_null() || columns[1].column().data_type().is_null() {
            return Ok(DataColumn::Constant(DataValue::Null, input_rows))
        }

        match (
            columns[0].column().cast_with_type(&DataType::Float64)?,
            columns[1].column().cast_with_type(&DataType::UInt64)?,
        ) {
            (
                // expect: if parse error, default value of f64 and u64 is used.
                DataColumn::Constant(DataValue::Float64(Some(number)), _),
                DataColumn::Constant(DataValue::UInt64(Some(precision)), _),
            ) => {
            ......
            }
```

sundy-li: > If parse N or D error, default value is used. If N or D is NULL, the result is NULL.

Yes, currently default value is used when casting reach errors from other data columns to primitive columns. 

See: 

```
select toFloat64('hello');
select toFloat64(Null);
```

sundy-li: Current implementation is cast columns into StringColumn then cast StringColumn into Float64Column/Int64Column, we can cast columns directly into specific columns.

eastfisher: Thanks! Now cast N and D to the correct type. PTAL~

tests/suites/0_stateless/02_0041_function_strings_format.sql
```diff
@@ -0,0 +1,14 @@
+SELECT FORMAT(12332.123456, 4);
+SELECT FORMAT(-12332.123456, 4);
+SELECT FORMAT(12332.1, 4);
+SELECT FORMAT(12332.2, 0);
+SELECT FORMAT(12332.2, -1);
+SELECT FORMAT(12332, 2);
```
sundy-li, line 6: What if `select format(number +10000, 2)  from numbers(10)`?

common/functions/src/scalars/strings/format.rs
sundy-li: I'm sorry, I forgot to tell you to consider none constant cases.

See https://github.com/datafuselabs/databend/blob/f5fda2e8f7c13f2c4764d941472fe7801693381d/common/functions/src/scalars/strings/repeat.rs

tests/suites/0_stateless/02_0041_function_strings_format.sql
eastfisher: It is ok when `select format(number +10000, 2) from numbers(1)` but error when `numbers(2)`.

eastfisher: It is ok when `select format(number +10000, 2) from numbers(1)` but error when `numbers(2)`.

sundy-li: As I mentioned before, we should care about none constant case.

See https://github.com/datafuselabs/databend/blob/f5fda2e8f7c13f2c4764d941472fe7801693381d/common/functions/src/scalars/strings/repeat.rs

eastfisher: Copy from repeat() and it works!

sundy-li:
Almost good to me, continue #3489

PR #3319, opened 2021-12-09
ISSUE-3257: Add created_on to TableMeta

GanZiheng:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #3257 

## Test Plan

Unit Tests

Stateless Tests

PR #3475, opened 2021-12-15
[test]: test: filter stdout to let result matching easier

drmingdrmer:
This is not meant for merge but to solve a testing issue in  @GanZiheng 's  PR
- fix: #3319

### DIFF
--- a/tests/databend-test
+++ b/tests/databend-test
@@ -77,9 +77,39 @@ def run_single_test(args, ext, client_options, case_file, stdout_file,
     os.system("LC_ALL=C sed -i -e 's/{test_db}/default/g' {file}".format(
         test_db=args.database, file=stdout_file))
 
+
+    if proc.returncode == 0:
+
+        filter_fn = result_file.rsplit(".result")[0] + ".result_filter"
+        if os.path.exists(filter_fn):
+            with open(filter_fn, 'rb') as f:
+                filters = f.readlines()
+
+            with open(stdout_file, 'rb') as f:
+                stdout_lines = f.readlines()
+
+            res = []
+            for line in stdout_lines:
+                line = line.strip()
+                for i in range(0, len(filters), 2):
+                    flt = filters[i].strip()
+                    repl = filters[i+1].strip()
+
+                    line = re.sub(flt, repl, line)
+
+                res.append(line)
+
+            stdout = b'\n'.join(res) + b"\n"
+
+            with open(stdout_file, 'wb') as f:
+                f.write(stdout)
+
+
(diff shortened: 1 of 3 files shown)


sundy-li:
I think it's too redundant, we can write shell/python scripts.

https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/00_0000_dummy_select_sh.sh

drmingdrmer:
> I think it's too redundant, we can write shell/python scripts.
> 
> https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/00_0000_dummy_select_sh.sh
> 
> shell pipeline grep should fix this case.

IMHO, we should avoid using shell script and provide a way that is easier than writing a script for developers to transform the unstable result(such as date/time):

- Handling string literal is a disaster in shell, e.g. there is no way to escape a single quote in a string quoted with single quote.
- python `re` has less surprises than grep or sed.
- shell scripts are less portable than python. E.g. On mac `gsed` is the gnu compatible `sed` and  that `sed` and `gsed` on mac have different meanings for option `-i`.

sundy-li:
So how about using python scripts, we support all executable scripts (python,shell,golang,perl ..) , all we need to do is make the file executable.

python: https://github.com/datafuselabs/databend/blob/ac5a59358873347495b8b14456279d3938d829c5/tests/suites/0_stateless/12_0000_insert_into_select2.py

BohuTANG:
I perfer using python script for the special test, it is programable.

We would choose a option, this is blocked https://github.com/datafuselabs/databend/pull/3319/files#diff-887f7231188d935f381b467a8a38898e98bb09ac489678fdb692c9fa115a8482R80

Or we merged #3319 first and then change the test to python script.
@drmingdrmer @sundy-li what do you think? :thinking:

drmingdrmer:
> I perfer using python script for the special test, it is programable.
> 
> We would choose a option, this is blocked https://github.com/datafuselabs/databend/pull/3319/files#diff-887f7231188d935f381b467a8a38898e98bb09ac489678fdb692c9fa115a8482R80
> 
> Or we merged #3319 first and then change the test to python script. @drmingdrmer @sundy-li what do you think? 🤔

@sundy-li 
@BohuTANG 

There are two problems around this issue:
- dynamic input: the input sql for a test is not static or the test has to setup supporting resources such as s3.
- dynamic output: the output of a test varies depending on the environment, even with the same input sql. I.e., `select now()`

pyhon/shell script is meant to solve the first problem.
Something like a result filter in my sub-PR is to solve the second one.

Issue #3285, opened 2021-12-07
[metasrv] make transaction sync to disk as default behavior


ariesdevil:
**Summary**

Make transaction sync as default, so that caller no need to pass `sync=true` every time.

Related code: https://github.com/datafuselabs/databend/blob/7096d74f44dcada6ed945fa370f81920d435c83d/common/meta/sled-store/src/sled_tree.rs#L93

Issue #3284, opened 2021-12-07
[metasrv] define a method to expose `AsTxnKeySpace::inner`.


ariesdevil:
**Summary**

Some methods that take `&TransactionSledTree` as parameter need to be called in subtree(aka:AsTxnKeySpace) method, since subtree already ref to `TransactionSledTree` we impl deref here to fetch inner `&TransactionSledTree`.
eg:
```rust
fn txn_incr_seq(&self, key: &str, txn_tree: &TransactionSledTree) {}
fn sub_txn_tree_do_update<'s, KS>(
    &'s self,
    sub_tree: &AsTxnKeySpace<'s, KS>,
) {
    seq_kv_value.seq = self.txn_incr_seq(KS::NAME, &*sub_tree);
    sub_tree.insert(key, &seq_kv_value);
}
```
But the current implementation is prone to misuse and misunderstanding, the `Deref` trait should only be impl for that outer and inner have similar behavior.

So we need to refactor the current impl to define a method and explicitly expose the inner field.

FYI: subtree is key spaces with different key prefix.
related codes: https://github.com/datafuselabs/databend/blob/7096d74f44dcada6ed945fa370f81920d435c83d/common/meta/sled-store/src/sled_tree.rs#L570-L592

PR #3485, opened 2021-12-15
[metasrv] refine some codes

ariesdevil:

## Summary

For #3284  with `Deref` impl, we can use `self.txn_tree` directly in `impl Store for AsTxnKeySpace` [methods](https://github.com/datafuselabs/databend/blob/7705b175e00b728d967a48a29d3fb52d02fbfad3/common/meta/sled-store/src/sled_tree.rs#L518-L566)

For #3285 I investigated some methods that can provide default arg value, but they are all not beautiful enough, so we can leave it for now.

## Changelog

- Improvement

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/meta/raft-store/src/state_machine/sm.rs
+++ b/common/meta/raft-store/src/state_machine/sm.rs
@@ -837,7 +837,7 @@ impl StateMachine {
             },
         };
 
-        seq_kv_value.seq = self.txn_incr_seq(KS::NAME, &*sub_tree)?;
+        seq_kv_value.seq = self.txn_incr_seq(KS::NAME, sub_tree)?;
 
         sub_tree.insert(key, &seq_kv_value)?;
 


common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -837,7 +837,7 @@ impl StateMachine {
             },
         };
 
-        seq_kv_value.seq = self.txn_incr_seq(KS::NAME, &*sub_tree)?;
+        seq_kv_value.seq = self.txn_incr_seq(KS::NAME, sub_tree)?;
```
drmingdrmer, line 840: Why need to deref it?

drmingdrmer (approved):

common/meta/raft-store/src/state_machine/sm.rs
ariesdevil: need inner `txn_tree` to get key_space

PR #3499, opened 2021-12-16
use clap 3.0.0-rc.5 to unify the cli arg parsing

### LINKED ISSUE #3504: Clap  `--single` is not compatible
Ubuntu platform uses `databend-query-standalone-embedded-meta.sh` which doesn't depend on `--single` option. So the root cause is the change for `--single` is not compatible. It used to support `--single=true` but now it doesn't.

To me, it's OK to use `--single` only, if so, we need to update scripts under `scripts/deploy`.

ping @BohuTANG for review.

_Originally posted by @Xuanwo in https://github.com/datafuselabs/databend/issues/3499#issuecomment-995601515_

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

## Summary

use clap 3 to unify the cli arguments parsing

## Changelog

- Improvement

## Related Issues

Close #3504 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -19,18 +19,18 @@ path = "src/bin/bendctl.rs"
 
 [dependencies]
 # Workspace dependencies
-databend-query = {path = "../query"}
-common-base = {path = "../common/base" }
+databend-query = { path = "../query" }
+common-base = { path = "../common/base" }
 common-datavalues = { path = "../common/datavalues" }
-common-tracing = {path = "../common/tracing" }
+common-tracing = { path = "../common/tracing" }
 
 itertools = "0.10.3"
-databend-meta = {path = "../metasrv" }
-common-meta-raft-store= {path = "../common/meta/raft-store"}
+databend-meta = { path = "../metasrv" }
+common-meta-raft-store = { path = "../common/meta/raft-store" }
 colored = "2.0.0"
 comfy-table = "5.0.0"
 dirs = "4.0.0"
-clap = { version = "3.0.0-rc.0", features = ["env"] }
+clap = { version = "3.0.0-rc.5", features = ["env"] }
 clap_generate = "3.0.0-rc.5"
 dyn-clone = "1.0.4"
 flate2 = "1.0.22"
@@ -51,8 +51,6 @@ thiserror = "1.0.30"
 ureq = { version = "2.3.1", features = ["json"] }
 nix = "0.23.0"
 serde_yaml = "0.8.21"
-structopt = "0.3.25"
-structopt-toml = "0.5.0"
 portpicker = "0.1.1"
 reqwest = { version = "0.11.7", features = ["json", "native-tls", "multipart", "blocking", "tokio-rustls", "stream"] }
 libc = "0.2.109"
@@ -73,4 +71,4 @@ predicates = "2.1.0"
 httpmock = "0.6.4"
 
 [build-dependencies]
-common-building = {path = "../common/building"}
+common-building = { path = "../common/building" }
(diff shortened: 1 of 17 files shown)


cli/Cargo.toml
```diff
@@ -19,18 +19,18 @@ path = "src/bin/bendctl.rs"
 
 [dependencies]
 # Workspace dependencies
-databend-query = {path = "../query"}
-common-base = {path = "../common/base" }
+databend-query = { path = "../query" }
```
BohuTANG, line 22: Cool.
These pretty format is by tool or hand?

common/meta/raft-store/src/config.rs
```diff
@@ -29,109 +29,68 @@ pub const KVSRV_BOOT: &str = "KVSRV_BOOT";
...
-    )]
+    /// The listening host for metadata communication.
+    #[clap(long, env = KVSRV_API_HOST, default_value = "127.0.0.1")]
     pub raft_api_host: String,
 
-    #[structopt(
-        long,
-        env = KVSRV_API_PORT,
-        default_value = "28004",
-        help = "The listening port for metadata communication"
```
BohuTANG, line 50: Why remove help here?

BohuTANG: The `help` will be helpful with:
databend-meta -h

cli/Cargo.toml
koushiro: by vscode plugin

common/meta/raft-store/src/config.rs
koushiro: the doc will be help info

cli/Cargo.toml
BohuTANG: Emm, what's the plugin name? Thanks :D

koushiro: Even Better TOML

BohuTANG:
This patch seems not work on macOS:
```
Waiting on databend-meta 10 seconds...
error: The value '=true' was provided to '--single' but it wasn't expecting any more values

databend-meta [OPTIONS]

For more information try --help
```

CI failure:
https://github.com/datafuselabs/databend/runs/4545281445?check_suite_focus=true

Xuanwo:
Ubuntu platform uses `databend-query-standalone-embedded-meta.sh` which doesn't depend on `--single` option. So the root cause is the change for `--single` is not compatible. It used to support `--single=true` but now it doesn't.

To me, it's OK to use `--single` only, if so, we need to update scripts under `scripts/deploy`.

ping @BohuTANG for review.

koushiro:
I could add `parse(try_from_str)` attribute for the `single` field to keep compatible

BohuTANG:
> Ubuntu platform uses `databend-query-standalone-embedded-meta.sh` which doesn't depend on `--single` option. So the root cause is the change for `--single` is not compatible. It used to support `--single=true` but now it doesn't.
> 
> To me, it's OK to use `--single` only, if so, we need to update scripts under `scripts/deploy`.
> 
> ping @BohuTANG for review.

@koushiro Could you have plan to fix it in this PR? Thank you.

BohuTANG:
`--single` is better, we can fix it in the script.

koushiro:
> `--single` is better, we can fix it in the script.

I seem to have seen the same issues before, I will fix them together later.
I'm outside for dinner now :p

BohuTANG:
Still have some issues with databend-meta:
```
The databend-query is not killed. force killing.
databend-query: no process found
=== boot new metasrv cluster of 3
Start Meta service HA cluster(3 nodes)...
error: Found argument 'true' which wasn't expected, or isn't valid in this context

USAGE:
    databend-meta [OPTIONS]
```

https://github.com/datafuselabs/databend/runs/4555094771?check_suite_focus=true

PsiACE (approved):
LGTM

PR #3506, opened 2021-12-16
ISSUE-3505: covert table function `history_truncate` to fuse table's methods

### LINKED ISSUE #3505: convert table function truncate_history to Table's Function
/lgtm

We can improve it in later PR

_Originally posted by @BohuTANG in https://github.com/datafuselabs/databend/issues/3463#issuecomment-995569309_

dantengsky:

## Summary

- table function `fuse_history_truncate` has been removed
   functionalities are merged into fuse table
- add new method `truncate_histotry` to trait `Table`
```.rust
    async fn truncate_history(
        &self,
        _ctx: Arc<QueryContext>,
        _keep_last_snapshot: bool,
    ) -> Result<()> {
        Ok(())
    }

```
  for now, values of arg `_keep_last_snapshot` passed to this method are all `false`. 
  let's keep it for a while (might be useful while implementing sth lke `alter table ... truncate history` stmts)
- interpreter_table_drop now calles `Table::truncate_history` instead

## Changelog

- Improvement

## Related Issues

Fixes #3505 

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/operations/truncate_histor.rs
```diff
@@ -0,0 +1,116 @@
+use std::collections::HashSet;
```
BohuTANG, line 1: 1. file name: truncate_histor.rs -> truncate_history.rs 
2. The licenese header is not on top ~

dantengsky: oops, got it!

tests/suites/0_stateless/09_0007_func_fuse_truncate_history.sql
```diff
@@ -8,41 +8,15 @@ insert into t values (5);
...
 select count(*) from fuse_history('db_09_0007', 't');
 
-
--------primary case--------------
-
--- do the truncation (without syntax sugar)
--- 2 snapshot removed, and not segments or blocks will be removed
-select * from fuse_truncate_history('db_09_0007', 't');
--- after truncation, count of history will 1
+-- truncate table will remove all the historical data but the latest one
```
BohuTANG, line 14: If the latest one snapshot has datas after `truncate table 't'`, does `select * from t1` has the results?

query/src/storages/storage_table.rs
```diff
@@ -118,8 +118,16 @@ pub trait Table: Sync + Send {
         _truncate_plan: TruncateTablePlan,
     ) -> Result<()> {
         Err(ErrorCode::UnImplement(format!(
-            "truncate for local table {} is not implemented",
+            "truncate for table {} is not implemented",
             self.name()
         )))
     }
+
+    async fn truncate_history(
```
BohuTANG, line 126: The function name called `optimize` maybe better?
https://dev.mysql.com/doc/refman/8.0/en/optimize-table.html

tests/suites/0_stateless/09_0007_func_fuse_truncate_history.sql
dantengsky: `select *  from t1`  returns an empty resultset in this case, the semantic of the statement `truncate table` does not change. 

but a snapshot will be kept, which references no segments or blocks.

query/src/storages/storage_table.rs
dantengsky: sure

BohuTANG:
Hello,
Does this PR only improve the truncate table?

We will do the snapshot history delete(not truncate the data) in another PR? Like using 'optimize table t1'.

dantengsky:
> Hello, Does this PR only improve the truncate table?

yeah,  the table function `fuse_truncate_history` is no longer invokable by end-users in this PR

> We will do the snapshot history delete(not truncate the data) in another PR? Like using 'optimize table t1'.

yes, `optimize table` or `alter table ... truncate history` will be addressed by another PR

dantengsky:
> We will do the snapshot history delete(not truncate the data) in another PR?

snapshot history deletion is done implicitly by statements `drop table` and `truncate table` in this PR.
no explicit statement for `history deletion ` in this PR. will be addressed in another PR

PR #3511, opened 2021-12-16
ISSUE-3510: `optimize table` statement

### LINKED ISSUE #3510: introduce `purge` argument for truncate statement
**Summary**:

introduce `purge` argument for truncate statement

- by default, `truncate` will keep the historical data (snapshot/segment/block)
- `truncate table purge` will physically remove all the historical data, except the data (snapshot/segment/block) referenced by the last snapshot

------------------

Originally  disscussion:

> We will do the snapshot history delete(not truncate the data) in another PR?

snapshot history deletion is done implicitly by statements `drop table` and `truncate table` in this PR.
no explicit statement for `history deletion ` in this PR. will be addressed in another PR

_Originally posted by @dantengsky in https://github.com/datafuselabs/databend/issues/3506#issuecomment-995788261_

dantengsky:

## Summary

- introduce `purge` argument for the `truncate` statement

  - by default ` truncate table T` truncates the data of table `T` logically, historical data of table `T` will be kept.

  - ` truncate table $T PURGE` will truncate all the historical data (segments, blocks, snapshots) as well

- rename `compact table T` to `optimize table T [purge | compact| all]`
  -  `optimize table T` 
     will purge the historical data of table T, only the last snapshot, and the data referenced by this snapshot will be kept.
  - `optimize table T purge` 
      equals ` optimize table T`
  - `optimize table T compact` 
      will compact the history data only -- try to merge small blocks 
      a new snapshot of table T will be added to the history, by this  compaction operation 
  - `optimize table T all`       
     will compact the history data, and then, purge the history, except the last one.

## Changelog

- Improvement

## Related Issues

Fixes #3510

## Test Plan

Unit Tests

Stateless Tests


query/src/interpreters/interpreter_table_optimize.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    async fn execute(
+        &self,
+        _input_stream: Option<SendableDataBlockStream>,
+    ) -> Result<SendableDataBlockStream> {
+        let do_compact;
+        let do_purge;
+        let plan = &self.plan;
+        match plan.operation {
+            Optimization::ALL => {
```
sundy-li, line 56: can we use bitflags here ? 

https://docs.rs/bitflags/latest/bitflags/

tests/suites/0_stateless/09_0008_fuse_optimize_table.result
```diff
@@ -0,0 +1,15 @@
+4
+1
+5
```
sundy-li, line 3: Magic numbers across two files, hard to review.

How about `select count(*) = 4  from fuse_history('db_09_0008', 't');`

tests/suites/0_stateless/09_0008_fuse_optimize_table.sql
```diff
@@ -0,0 +1,55 @@
...
+
+optimize table t compact;
+
+-- expects 4 history items, 3 of previous insertion, 1 for last compaction
+select count(*) from fuse_history('db_09_0008', 't');
+
+---------------------------
+
+-- purge data and history
+optimize table 't' purge;
```
BohuTANG, line 21: `optimize table 't'` equals `optimize table 't' purge`?

tests/suites/0_stateless/09_0008_fuse_optimize_table.result
dantengsky: > How about select count(*) = 4 from fuse_history('db_09_0008', 't');

in `09_0008_fuse_optimize_table.sql`, change 

```
-- expects 4 history items, 3 of previous insertion, 1 for last compaction
select count(*) from fuse_history('db_09_0008', 't');
```

to 

```
-- expects 4 history items, 3 of previous insertion, 1 for last compaction
select count(*) = 4 from fuse_history('db_09_0008', 't');
````
,  and in `09_0008_fuse_optimize_table.result`, change the corresponding line, from `4` to `1`?

sure~

to keep the style in consistent, other `select count(*)`, even expects 1 will be changed to `select count(*) = ?` as well

File: tests/suites/0_stateless/09_0008_fuse_optimize_table.sql (hunk 3 above)
Comment (dantengsky) [reply tests/suites/0_stateless/09_0008_fuse_optimize_table.sql:21]: yes, 

line 20:
```
-- purge data and history
optimize table 't' purge;
-- expect 1 snapshot left
select count(*) from fuse_history('db_09_0008', 't');
```

- wondering why there is a snapshot left? yes, it is not consistent with the behavior of table version 0 has no snapshot file. IMO, later we should save a snapshot for table version 0 as well.

the snapshot left references no segments or blocks. (checked in unit test)

- wondering why the default behavior is purge, but not `compact and purge`?
   currently, the impl of `compact` is brute-force or ... dumb, how about changing the default behavior of `optimize` to `compact & purge` till the impl of `compact` is tuned

query/src/interpreters/interpreter_table_optimize.rs
dantengsky: sure!

dantengsky:
@sundy-li PTAL, stateless test case tweaked as you suggested,  Enum is replaced with bitflags

sundy-li (approved):

Issue #3011, opened 2021-11-24
LCASE() | Synonym for LOWER()


zhang2014:

Xuanwo:
/assignme

> Take https://github.com/datafuselabs/databend/issues/3004#issuecomment-977774256 for reference.

Xuanwo:
I'm working on https://github.com/jorgecarleitao/arrow2/issues/635 to add required support in arrow.

Here is the checklist for work on arrow2 side:

- https://github.com/jorgecarleitao/arrow2/pull/641
- Add upper support

Issue #3010, opened 2021-11-24
LOWER() | Return the argument in lowercase


zhang2014:

Xuanwo:
/assignme

> Take https://github.com/datafuselabs/databend/issues/3004#issuecomment-977774256 for reference.

PR #3521, opened 2021-12-17
functions: Add lower support

Xuanwo:

## Summary

Add lower support

## Checklist

- [x] Implement functios
- [x] Add stateless tests
- [x] Add website docs

## Changelog

- New Feature

## Related Issues

- Fixes https://github.com/datafuselabs/databend/issues/3010
- Fixes https://github.com/datafuselabs/databend/issues/3011

## Test Plan

Unit Tests

Stateless Tests


sundy-li (approved):

sundy-li:
BTW Upper can be addressed too.

Xuanwo:
> BTW Upper can be addressed too.

Yep, I will added in the next PR :smile:

common/functions/Cargo.toml
```diff
@@ -38,6 +38,7 @@ hex = "0.4.3"
 base64 = "0.13.0"
 itertools = "0.10.3"
 num-format = "0.4"
+bstr = "0.2.17"
```
sundy-li, line 41: what's this crate for?

Xuanwo: `bstr` is a byte string library. I use the `ByteSlice` trait so that we don't need to implement `s.char_indices()` on `[u8]` agin.

PsiACE (approved):
LGTM

common/functions/src/scalars/strings/lower.rs
```diff
@@ -0,0 +1,32 @@
...
+use super::string2string::String2StringFunction;
+use super::string2string::StringOperator;
+
+#[derive(Clone, Default)]
+pub struct Lower;
+
+impl StringOperator for Lower {
+    #[inline]
+    fn apply_with_no_null<'a>(&'a mut self, s: &'a [u8], buffer: &mut [u8]) -> usize {
+        buffer.copy_from_slice(&s.to_lowercase());
```
sundy-li, line 26: Oh, there is an extra copy.

Maybe we can use  https://docs.rs/bstr/latest/bstr/trait.ByteSlice.html#method.to_lowercase_into

Xuanwo: After some research, I found:

- `to_lowercase_into` requires a `&mut Vec<u8>` and we can't fit the requirement.

One possible solution is to implement the `to_lowercase_into` by hand. What's your idea?

sundy-li: Just use the codes inside  `to_lowercase_into`:

```
 #[inline]
    fn to_lowercase_into(&self, buf: &mut Vec<u8>) {
        // TODO: This is the best we can do given what std exposes I think.
        // If we roll our own case handling, then we might be able to do this
        // a bit faster. We shouldn't roll our own case handling unless we
        // need to, e.g., for doing caseless matching or case folding.

        // TODO(BUG): This doesn't handle any special casing rules.

        buf.reserve(self.as_bytes().len());
        for (s, e, ch) in self.char_indices() {
            if ch == '\u{FFFD}' {
                buf.push_str(&self.as_bytes()[s..e]);
            } else if ch.is_ascii() {
                buf.push_char(ch.to_ascii_lowercase());
            } else {
                for upper in ch.to_lowercase() {
                    buf.push_char(upper);
                }
            }
        }
    }
```

Xuanwo:
Oh, the already merged...

PR #3522, opened 2021-12-17
cleanup some deps

koushiro:

## Summary

- remove useless deps checked by `cargo udeps`
- update some deps to prune duplicated deps

## Changelog

- Improvement
- Build/Testing/CI

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -546,7 +546,6 @@ dependencies = [
  "num-format",
  "portpicker",
  "predicates",
- "pretty_assertions",
  "reqwest",
  "run_script",
  "rustyline",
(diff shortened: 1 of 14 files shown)


BohuTANG:
/lgtm

Neat man, thank you @koushiro !

sundy-li (approved):

Xuanwo:
Oh, it's so nice, I think we can add this in our `check` CI steps.

@koushiro Do you have the interest to implement it?

koushiro:
@Xuanwo 
As far as I know, `cargo-udeps` does not provide suitable command options to report errors, such as the `--check` option in `cargo fmt - --check`,  `-D` option in `cargo clippy - -D warnings`.
So adding it to CI check may not be easy, at least I don’t have a good idea now.

PR #3532, opened 2021-12-17
[metasrv] move protobuf files out of metasrv

### LINKED ISSUE #3528: The first thing to do is to move `metasrv/meta.proto` out and put it into a new crate so that other crate is able to access metasrv without depending on it.
- [ ] The first thing to do is to move `metasrv/meta.proto` out and put it into a new crate so that other crate is able to access metasrv without depending on it.

_Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/issues/3515#issuecomment-996512151_

ariesdevil:

## Summary

Move protobuf files out of metasrv so that other crates can access metasrv without depending on it.

## Changelog

- Improvement

## Related Issues

Fixes #3528 

## Test Plan

Unit Tests

Stateless Tests


.licenserc.yaml
```diff
@@ -27,9 +27,7 @@ header:
     - "**/*.sql"
     - "rust-toolchain"
     - "rustfmt.toml"
-    - "**/*.proto"
```
drmingdrmer, line 30: Hmm... why remove these ignores? 🤔

common/meta/raft-store/build.rs
```diff
@@ -17,16 +17,15 @@
...
 }
 
 fn build_proto() {
     let manifest_dir =
         env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR env variable unset");
 
-    // Changing OUT_DIR fails docker build because src/proto is read-only.
-    // env::set_var("OUT_DIR", "src/proto");
+    // Changing OUT_DIR fails docker build because src/protobuf is read-only.
+    // env::set_var("OUT_DIR", "src/protobuf");
```
drmingdrmer, line 28: These 2 lines can be just removed

common/meta/raft-store/build.rs
```diff
@@ -17,16 +17,15 @@
 use std::env;
 use std::path::Path;
 fn main() {
-    common_building::setup();
```
drmingdrmer, line 20: Isn't the proto gonna moved into a standalone crate? 

The proto is about RPC protocol. IMHO, it has nothing to do with raft-store. Unless I missed anything. 🤔

ariesdevil: Cause I encounter the rust orphan rule :(

flaneur2020: > The proto is about RPC protocol. IMHO, it has nothing to do with raft-store. Unless I missed anything. 🤔

+1

how about calling it meta-proto or meta-api? q.q

Xuanwo:
Oh, the workaround does work.

ariesdevil:
> Oh, the workaround does work.

Thx a lot~

PR #3533, opened 2021-12-17
Backtrace lead  performance drawback

sundy-li:

## Summary

Summary about this PR

 

## Changelog

 
 
- Performance Improvement
 
## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/exception/src/exception.rs
+++ b/common/exception/src/exception.rs
@@ -110,7 +110,7 @@ macro_rules! build_exceptions {
                         code: $code,
                         display_text: display_text.into(),
                         cause: None,
-                        backtrace: Some(ErrorCodeBacktrace::Origin(Arc::new(Backtrace::new()))),
+                        backtrace: None,
                     }
                 }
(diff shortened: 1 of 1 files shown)


sundy-li:
```
SELECT sum(  toUInt32(number) + toUInt32(number ) )  FROM numbers(1000000000);
SELECT sum(  number + number  )  FROM numbers_mt(1000000000) ;
```

The first query works really very slow, maybe 99% slower than the second one.

I profiled and found it's related to `backtrace` crate in `common/exceptions`, but I do know the reason.

cc @zhang2014  @PsiACE  @zhyass  @Xuanwo 


HAKjQmLOcl

sundy-li:
Oh,  I found that there is an error instance in code:

```
 let error = ErrorCode::BadDataValueType(format!(
            "Unsupported cast_with_type from array: {:?} into data_type: {:?}",
            series, self.cast_type,
        ));
```

PR #3536, opened 2021-12-17
Refactor string hash functions using String2String template function

### LINKED ISSUE #3530: Refactor string hash functions using String2String template function
Can use String2String template function.

_Originally posted by @sundy-li in https://github.com/datafuselabs/databend/pull/3527#discussion_r771268042_

- [ ] md5
- [ ] sha1
- [ ] blake3

mshauneu:

## Summary

Refactor string hash functions using String2String template function

## Changelog

- Improvement
- Performance Improvement

## Related Issues

Fixes #3530 

## Test Plan

Stateless Tests


common/functions/src/scalars/hashes/sha1hash.rs
```diff
@@ -12,80 +12,25 @@
...
-impl Sha1HashFunction {
-    pub fn try_create(display_name: &str) -> Result<Box<dyn Function>> {
-        Ok(Box::new(Sha1HashFunction {
-            display_name: display_name.to_string(),
-        }))
+impl StringOperator for Sha1 {
+    #[inline]
+    fn apply_with_no_null<'a>(&'a mut self, s: &'a [u8], buffer: &mut [u8]) -> usize {
+        // TODO sha1 lib doesn't allow encode into buffer...
+        buffer.copy_from_slice(sha1::Sha1::from(s).digest().to_string().as_ref());
```
sundy-li, line 27: https://docs.rs/sha1/0.6.0/src/sha1/lib.rs.html#187 
```
impl Digest {
    /// Returns the 160 bit (20 byte) digest as a byte array.
    pub fn bytes(&self) -> [u8; DIGEST_LENGTH] {
```

BohuTANG:
CI failed:

```
failures:

---- scalars::hashes::test_blake3hash_function stdout ----
thread 'scalars::hashes::test_blake3hash_function' panicked at 'source slice length (64) does not match destination slice length (128)', common/functions/src/scalars/hashes/blake3hash.rs:27:16
stack backtrace:
   0:        0x10c41d1b1 - std::backtrace_rs::backtrace::libunwind::trace::hf3fc7adb04351f82
```

common/functions/src/scalars/hashes/sha1hash.rs
sundy-li: Or can we use `write!`, ` std::io::Write::write_fmt` to reduce the String allocation?

mshauneu: function should return hex instead of raw bytes - hex is bytes x 2
thee problem was in multi value case 
I forgot to make slice of whole slice (`copy_from_slice` reqs slices of same length)

i think `hex::encode_to_slice` works better (hex crate already in our Cargo)

sundy-li:
Great!

PR #3543, opened 2021-12-19
Move validity masking out from function to transform

junli1026:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/columns/data_column.rs
+++ b/common/datavalues/src/columns/data_column.rs
@@ -12,7 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::iter::Iterator;
+use std::sync::Arc;
+
 use common_arrow::arrow::array::ArrayRef;
+use common_arrow::arrow::bitmap::Bitmap;
+use common_exception::ErrorCode;
 use common_exception::Result;
 
 use crate::prelude::*;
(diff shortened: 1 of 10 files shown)


common/datavalues/src/columns/data_column.rs
```diff
@@ -210,6 +261,135 @@ impl DataColumn {
...
+        let bitmap_to_apply = match validity_to_apply {
+            DataColumnValidity::Array(v, _) => v,
+
+            // The validity is neither all-true nor all-false, it must NOT be a constant one.
+            DataColumnValidity::Constant(_, _) => unreachable!(),
+        };
+
+        match self {
+            DataColumn::Array(_) => {
+                let array_ref = self.get_array_ref()?;
```
sundy-li, line 319: just use `array.validity()`

common/datavalues/src/columns/data_column.rs
```diff
@@ -210,6 +261,135 @@ impl DataColumn {
...
+        for (i, v) in validities.iter().enumerate() {
+            if v.len() != array_len {
+                return Err(ErrorCode::BadArguments(format!(
+                    "Validity's length {} doesn't match column array len {}",
+                    v.len(),
+                    array_len
+                )));
+            }
+            if v.all_null() {
+                all_null_index = i as i32;
```
sundy-li, line 361: so we can `return Ok(DataColumnValidity::Constant(false, array_len));` here.

all_null_index is useless.

common/datavalues/src/columns/data_column.rs
```diff
@@ -210,6 +261,135 @@ impl DataColumn {
...
+            }
+            if !v.is_constant() {
+                bitmap_vec.push(v);
```
sundy-li, line 364: So bitmap_vec can be Vec<Option<&Bitmap>> instead.  We don't need `unreachable` later.

common/functions/src/scalars/strings/concat.rs
```diff
@@ -140,6 +140,10 @@ impl Function for ConcatFunction {
         let result = Self::concat_column(columns[0].column().clone(), &columns[1..])?;
         Ok(result)
     }
+
+    fn passthrough_null(&self) -> bool {
```
sundy-li, line 144: why this is false?

BohuTANG:
Hello, any progress of this PR? @sundy-li @junli1026

junli1026:
> Hello, any progress of this PR? @sundy-li @junli1026

Just had a call with Sundy, will update the PR soon.

sundy-li (approved):

BohuTANG:
:+1: , Thank you @junli1026 !

PR #3552, opened 2021-12-19
implement char() function

### LINKED ISSUE #3017: CHAR() | Return the character for each integer passed

BlackSmith96:

## Summary

Summary about this PR:
Implement char()

## Changelog

- New Feature

## Related Issues

Fixes #3017 

## Test Plan
Unit Tests
Stateless Tests


common/functions/src/scalars/strings/char_.rs
```diff
@@ -0,0 +1,126 @@
...
+                    l_array = DFStringArray::new_from_iter_validity(
+                        l_array
+                            .into_no_null_iter()
+                            .zip(r_array.into_no_null_iter())
+                            .map(|(l, r)| {
+                                [
+                                    l,
+                                    &r.to_be_bytes()
+                                        .iter()
+                                        .filter(|&x| *x != 0u8)
```
sundy-li, line 81: I don't think we should filter this.

```
TiDB> SELECT length(CHAR(67, 97, 0,  116));
+-------------------------------+
| length(CHAR(67, 97, 0,  116)) | 
+-------------------------------+
| 4                             | 
+-------------------------------+
1 row in set (0.00 sec)
```

common/functions/src/scalars/strings/char_.rs
```diff
@@ -0,0 +1,126 @@
...
+                continue;
+            }
+            let column = column.cast_with_type(&DataType::UInt32)?;
+            match column {
+                DataColumn::Array(r_array) => {
+                    let r_array = r_array.u32()?;
+                    l_array = DFStringArray::new_from_iter_validity(
+                        l_array
+                            .into_no_null_iter()
+                            .zip(r_array.into_no_null_iter())
```
sundy-li, line 75: Too much mem `allocation` and `memcpy` here.

sundy-li: We can use `DFStringArray::from_data_unchecked` to build result columns.

For values buffer, we can reserve the size to be `rows * args_size`.

I contributed the `char` function of ClickHouse before, hope this can help you.

https://github.com/ClickHouse/ClickHouse/blob/master/src/Functions/FunctionChar.cpp#L55-L91

BlackSmith96: Thanks for you review.
I have read your implementation of `char()` function. It seems there are some definition differences between your char() function and mysql8's. In mysql8's definition, one integer can be converted into 4 chars at most. So which definition is the one to implement? 
If mysql8's definition is the one to implement, my solution is to reserve a values buffer, the size of which is `4 * rows * args_size`.

sundy-li: >  one integer can be converted into 4 chars at most

I did not get it. Integer will be cast into u8, so it just took one char.

```
MySQL> select char(64);
+----------+
| char(64) | 
+----------+
| @        | 
+----------+
1 row in set (0.00 sec)
MySQL> select char(64 + 256) ;
+-----------------+
| char(64 + 256)  | 
+-----------------+
| @              | 
+-----------------+
1 row in set (0.00 sec)
```

BlackSmith96: How about the parameter type? Do we only support numeric type?

sundy-li: Yes, we can use `cast_with_type(&DataType::UInt8)` to cast all parameter columns.

BlackSmith96: Is there any method to generate a bitmap full of 1? One parameter's type of function `DFStringArray::from_data_unchecked` is `Option<Bitmap>`

sundy-li: Just use None!

common/functions/src/scalars/strings/char_.rs
```diff
@@ -0,0 +1,130 @@
...
+            let column = column.cast_with_type(&DataType::UInt8)?;
+            match column {
+                DataColumn::Array(uint8_arr) => {
+                    let uint8_arr = uint8_arr.u8()?;
+                    for (j, ch) in uint8_arr.into_iter().enumerate() {
+                        match ch {
+                            Some(&ch) => unsafe {
+                                *values.as_mut_ptr().add(column_count * j + i) = ch;
+                            },
+                            None => {
```
sundy-li, line 89: `None` case is not covered, but we will fill bitmap validity after eval function.

try use `uint8_arr.into_no_null_iter()`.

common/functions/src/scalars/strings/char_.rs
```diff
@@ -0,0 +1,130 @@
...
+                                return Err(ErrorCode::IllegalDataType(
+                                    "Expected args a const, an expression, an expression, a const and a const"
+                                ));
+                            }
+                        }
+                    }
+                }
+                DataColumn::Constant(uint8_arr, _) => match uint8_arr {
```
sundy-li, line 97: try use `uint8_arr.unwrap_or(0)`

tests/suites/0_stateless/02_0043_function_char.sql
```diff
@@ -0,0 +1,4 @@
+SELECT CHAR(97);
+SELECT CHAR(24929); -- 0x6161
```
sundy-li, line 2: Add a case to cover the null case .

`select char(99,100,,null)`

BlackSmith96: Do we treat NULL as valid input for char() function?

sundy-li: Yes, if one of the arguments is null, the result is null.
But in function's eval, you don't have to care about null, it's already done by merging the argument's null bitmap after function's eval. So you can write anything you want(better to use default value) if it's null.

common/functions/src/scalars/strings/char_.rs
BlackSmith96: Actually, after column `cast_with_type(uint8)`, `NULL` value would be converted to 0. There is no difference between  `into_no_null_iter()` and `into_iter()`. In this case, we can just think `char(NULL) == 0`, because the validity will tell us the real result is NULL.

sundy-li:
Almost good to me, continue in #3644

PR #3604, opened 2021-12-22
refactor fuse readers and fuse cache

Veeupup:

## Summary

* refactor fuse readers and improve fuse local cache.
* cache `snapshots` and `metadata` in memory

TODO:
* cache `block` data (raw parquet file) in ssd.(need `block_reader` support)

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
If the storage type is local disk, how the block cache work?

Veeupup:
> If the storage type is local disk, how the block cache work?

thanks for reminding me, we should disable block cache when using local disk as storage ?

common/cache/src/query/local_cache.rs
```diff
@@ -35,6 +36,10 @@ const CACHE_READ_BYTES_FROM_LOCAL: &str = "cache_read_bytes_from_local";
 const CACHE_ACCESS_COUNT: &str = "cache_access_count";
 const CACHE_ACCESS_HIT_COUNT: &str = "cache_access_hit_count";
 
+const FUSE_TBL_BLOCK_PREFIX: &str = "_b";
```
BohuTANG, line 39: Where these prefix original defines? It would be better to use the original definitions instead defining them.

Veeupup: it seem not that elegant because original definitions is in `query/storages/fuse/constants`...
so I repeat it here  : (

BohuTANG:
> > If the storage type is local disk, how the block cache work?
> 
> thanks for reminding me, we should disable block cache when using local disk as storage ?

Right,  should check the accessor is local or not.

common/cache/src/query/local_cache.rs
BohuTANG: I think the location is string is simple, not sure may it can be a structure contain s a type here.
Wondering how the location get from?

BohuTANG: The prefix type is a underlying logic, the up-front can't defines it, so we need a place tell us what's the location type exactly is.

query/src/storages/fuse/io/readers.rs
```diff
@@ -31,7 +31,6 @@ pub async fn read_obj<T: DeserializeOwned>(
     loc: impl AsRef<str>,
     cache: Arc<Option<Box<dyn QueryCache>>>,
 ) -> Result<T> {
-    // only metadata(segments and snapshots will use cache now)
```
BohuTANG, line 34: But the logical still check the meta cache only :)

common/cache/src/query/local_cache.rs
Veeupup: > 

we will get block path from `segments`, defines here: 
https://github.com/datafuselabs/databend/blob/170084277c/query/src/storages/fuse/meta/segment.rs#L21-L26

BohuTANG: Does the QueryCache mainly used by the fuse engine?
 If so we can move it to the storages/fuse folder.

dantengsky (approved):
LGTM, thanks @Veeupup

Issue #3593, opened 2021-12-22
Improve issue template


sundy-li:
**Summary**

Apache GitHub's issue template is a very good example to learn.

Eg:

https://github.com/apache/incubator-doris/blob/master/.github/ISSUE_TEMPLATE/feature-request.yml

Xuanwo:
BTW, our issues templates are too many...

PR #3610, opened 2021-12-23
github: Refactor issue template

Xuanwo:

## Summary

Simplify the issue template.

We used to have many templates but nearly no one use them. This PR tried to change the state and only include the template that we used daily.

In this PR, we removed all existing issue templates and only allow `bug_report` and `blank` instead.

- Questions and Feature Requests should go [Discussions](https://github.com/datafuselabs/databend/discussions)
- document, performance, build, other others should all be treated as `bug_report` and triaged via maintainer. A bug is a bug, it's not the users' duty to classify which type the bug is.

**This PR is highly inspired by [Doris](https://github.com/apache/incubator-doris) and [Rust](https://github.com/rust-lang/rust), thanks for their work.**

## Preview

Please preview this PR's change at https://github.com/Xuanwo/databend/issues/new/choose

## Next Step

We will add a `tracking-issue` template once our RFC process is mature.

## Changelog

- Other 

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/3593


sundy-li:
So if feature request is discussion, then we should change it into an issue if it's really a feature request?

BohuTANG:
> So if feature request is discussion, then we should change it into an issue if it's really a feature request?

If need, we can add a feature_issue.md template.
Now user can open an blank issue to do that

BohuTANG:
image

Xuanwo:
> So if feature request is discussion, then we should change it into an issue if it's really a feature request?

It depends.

- If the feature is a simple change, we can convert this discussion into an issue and close it after PR merged.
- If the feature is complex, we need to create a tracking issue with blank issue (*and we will define a tracking issue template in the future.*)

BohuTANG:
image

Do'nt like the idea github treats the checklist as taks :cry:

PR #3631, opened 2021-12-24
[deps] replace lazy_static with once_cell

PsiACE:

## Summary

- replace lazy_static with once_cell

### Why

- macro-free
- active & popular
- https://github.com/async-rs/async-std/issues/406#issuecomment-547286625

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/flight-rpc/src/dns_resolver.rs
```diff
@@ -42,17 +42,14 @@ pub struct DNSResolver {
...
-        match TokioAsyncResolver::tokio_from_system_conf() {
-            Err(error) => Result::Err(ErrorCode::DnsParseError(format!(
-                "DNS resolver create error: {}",
-                error
-            ))),
-            Ok(resolver) => Ok(Arc::new(DNSResolver { inner: resolver })),
-        }
-    };
-}
+static INSTANCE: Lazy<Result<Arc<DNSResolver>>> =
```
zhang2014, line 45: Can we extract an `create_dns_resolver` method?

common/functions/src/aggregates/aggregate_function_factory.rs
```diff
@@ -38,14 +38,12 @@ pub type AggregateFunctionCombinatorCreator = Box<
...
-    static ref FACTORY: Arc<AggregateFunctionFactory> = {
-        let mut factory = AggregateFunctionFactory::create();
-        Aggregators::register(&mut factory);
-        Aggregators::register_combinator(&mut factory);
-        Arc::new(factory)
-    };
-}
+static FACTORY: Lazy<Arc<AggregateFunctionFactory>> = Lazy::new(|| {
+    let mut factory = AggregateFunctionFactory::create();
+    Aggregators::register(&mut factory);
```
zhang2014, line 43: Move this into `AggregateFunctionFactory::create()` is better?

common/flight-rpc/src/dns_resolver.rs
PsiACE: emmm, We already have an `instance()` function. Perhaps it could be merged in without `Lazy`.

common/functions/src/aggregates/aggregate_function_factory.rs
BohuTANG: I think a bit hard:
Lazy<Arc<AggregateFunctionFactory>> is initialed global.

If we move it to `AggregateFunctionFactory::create()`, maybe we should call it in the first entry point.

BohuTANG: Maybe `once` privimite can be used here

PR #3632, opened 2021-12-24
Change nullable function's signature to accept data field

junli1026:

## Summary

1. Change nullable function's signature to accept data field, so that the constant expression's data-field can be passed in.
2. Modify LOG function not to combine validity in Eval.

## Changelog
- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/function_alias.rs
+++ b/common/functions/src/scalars/function_alias.rs
@@ -16,7 +16,6 @@ use std::fmt;
 
 use common_datavalues::columns::DataColumn;
 use common_datavalues::prelude::DataColumnsWithField;
-use common_datavalues::DataSchema;
 use common_datavalues::DataType;
 use common_exception::Result;
 
@@ -40,10 +39,6 @@ impl Function for AliasFunction {
 
     fn return_type(&self, args: &[DataType]) -> Result<DataType> {
         Ok(args[0].clone())
-    }
-
-    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
-        Ok(true)
     }
 
     fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
--- a/common/functions/src/scalars/logics/logic.rs
+++ b/common/functions/src/scalars/logics/logic.rs
@@ -16,7 +16,6 @@ use std::fmt;
 
 use common_datavalues::columns::DataColumn;
 use common_datavalues::prelude::DataColumnsWithField;
-use common_datavalues::DataSchema;
 use common_datavalues::DataType;
 use common_datavalues::DataValueLogicOperator;
 use common_exception::Result;
@@ -51,10 +50,6 @@ impl Function for LogicFunction {
 
     fn return_type(&self, _args: &[DataType]) -> Result<DataType> {
         Ok(DataType::Boolean)
-    }
-
-    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
-        Ok(false)
     }
 
     fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
--- a/common/functions/src/scalars/uuids/uuid_creator.rs
+++ b/common/functions/src/scalars/uuids/uuid_creator.rs
@@ -17,7 +17,6 @@ use std::marker::PhantomData;
 
 use common_datavalues::columns::DataColumn;
 use common_datavalues::prelude::DataColumnsWithField;
-use common_datavalues::DataSchema;
 use common_datavalues::DataType;
 use common_datavalues::DataValue;
 use common_exception::Result;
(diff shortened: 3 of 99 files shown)


common/functions/src/scalars/function_alias.rs
```diff
@@ -42,7 +42,7 @@ impl Function for AliasFunction {
         Ok(args[0].clone())
     }
 
-    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+    fn nullable(&self, _arg_fields: &[DataField]) -> Result<bool> {
```
sundy-li, line 45: Most function's nullable should use default as trait.

So it can be removed.

junli1026: > Most function's nullable should use default as trait.
> 
> So it can be removed.

Right, will update the PR

BohuTANG:
Merry Christmas!
Oops, some conflicts need to resolve :)

junli1026:
> Merry Christmas! Oops, some conflicts need to resolve :)

Merry Xmas!

common/planners/src/plan_expression_chain.rs
```diff
@@ -127,20 +129,25 @@ impl ExpressionChain {
                 self.add_expr(left)?;
                 self.add_expr(right)?;
 
-                let args_type = vec![
-                    left.to_data_type(&self.schema)?,
-                    right.to_data_type(&self.schema)?,
+                let arg_fields = vec![
```
zhyass, line 132: CC @zhang2014 , using to_data_field may affect the issue https://github.com/datafuselabs/databend/issues/3572?

common/functions/src/scalars/logics/logic.rs
```diff
@@ -53,7 +53,8 @@ impl Function for LogicFunction {
         Ok(DataType::Boolean)
     }
 
-    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+    // Null is treated as false, so it should not output any Null.
+    fn nullable(&self, _arg_fields: &[DataField]) -> Result<bool> {
```
sundy-li, line 57: select  null and 3;
select not null;
select null or 3;

all returns null

common/functions/src/scalars/uuids/uuid_creator.rs
```diff
@@ -91,7 +91,7 @@ where T: UUIDCreator + Clone + Sync + Send + 'static
         Ok(DataType::String)
     }
 
-    fn nullable(&self, _input_schema: &DataSchema) -> Result<bool> {
+    fn nullable(&self, _arg_fields: &[DataField]) -> Result<bool> {
```
sundy-li, line 94: If the arg size is zero, it's zero not null, so this method can be removed?

common/functions/src/scalars/logics/logic.rs
junli1026: OK, will fix. But if that is how MySQL handles logic function, do we really want to be  the same ? 
Especially ``` select not null; -> null``` and ``` select null or 3; -> null ```  it is confusing ...

common/functions/src/scalars/uuids/uuid_creator.rs
junli1026: Yes, it should be removed.

common/functions/src/scalars/logics/logic.rs
junli1026: Find some useful doc from postgre's doc "SQL uses a three-valued logic system with true, false, and null, which represents "unknown". "

common/planners/src/plan_expression_chain.rs
zhang2014: Looks good to me, it's ok. I will to refactor this after merged pr.

common/functions/src/scalars/logics/logic.rs
zhyass: > select null and 3; select not null; select null or 3;
> 
> all return null

`select null or 3 ` will return `1`(means true).

> OK, will fix. But if that is how MySQL handles logic function, do we really want to be the same ?
> Especially ` select not null; -> null` and `select null or 3; -> null` it is confusing ...

The return result of `select not null` in clickhouse is also `null`.

junli1026: > > select null and 3; select not null; select null or 3;
> > all return null
> 
> Correct: `select null or 3 ` will return `1`(means true).
> 
> > OK, will fix. But if that is how MySQL handles logic function, do we really want to be the same ?
> > Especially ` select not null; -> null` and `select null or 3; -> null` it is confusing ...
> 
> The return result of `select not null` in clickhouse is also `null`. So we can just use the default.

Right, I found this doc explains well https://www.postgresql.org/docs/9.1/functions-logical.html

zhyass (approved):

sundy-li (approved):

PR #3647, opened 2021-12-25
COMMON-3646: Merge all tracing layer to one

### LINKED ISSUE #3646: bug: $ ./target/debug/databend-query can't print logs
### Search before asking

### Version

main

### What's Wrong?

$ ./target/debug/databend-query canot print logs

### What You Expected?

Print the INFO level logs

### How to Reproduce?

make build-debug
$ ./target/debug/databend-query

### Anything Else?

_No response_

BohuTANG:

## Summary

1. Remove some unused codes
2. Merge all tracing layer to one
3. Change `tracing::instrument` from info to debug
4. Add Jaeger tracing to doc
5. Enable jaeger by default, but it only works in trace/debug mode

## Changelog

- Improvement

## Related Issues

Fixes #3646 

## Test Plan

Unit Tests

Stateless Tests


common/tracing/src/logging.rs
```diff
@@ -32,31 +28,16 @@ use tracing_subscriber::prelude::*;
...
-
 /// Init tracing for unittest.
 /// Write logs to file `unittest`.
 pub fn init_default_ut_tracing() {
     static START: Once = Once::new();
 
     START.call_once(|| {
-        let mut g = GLOBAL_UT_LOG_GUARD.as_ref().lock().unwrap();
-        *g = Some(init_global_tracing("unittest", "_logs", "DEBUG"));
+        let _ = init_global_tracing("unittest", "_logs", "DEBUG");
```
drmingdrmer, line 37: IMHO, the returned file guards should be kept otherwise the log file will be removed at once?

BohuTANG: Fixed, thank you!

common/tracing/src/logging.rs
```diff
@@ -70,125 +58,65 @@ static GLOBAL_UT_LOG_GUARD: Lazy<Arc<Mutex<Option<WorkerGuard>>>> =
...
+    let mut file_span_layer = None;
+    let mut jaeger_layer = None;
+    let fuse_jaeger_env = env::var("DATABEND_JAEGER").unwrap_or_else(|_| "".to_string());
+    if !fuse_jaeger_env.is_empty() {
+        {
+            let span_dir = format!("{}_span", dir);
+            let span_rolling_appender =
+                RollingFileAppender::new(Rotation::HOURLY, span_dir, app_name);
+            let (span_rolling_writer, span_rolling_writer_guard) =
+                tracing_appender::non_blocking(span_rolling_appender);
```
dantengsky, line 85: do we need another rolling appender here?

common/tracing/src/logging.rs
```diff
@@ -70,125 +58,65 @@ static GLOBAL_UT_LOG_GUARD: Lazy<Arc<Mutex<Option<WorkerGuard>>>> =
...
+            file_span_layer = Some(
+                Layer::new()
+                    .with_writer(span_rolling_writer)
+                    .with_thread_ids(true)
+                    .with_thread_names(false)
+                    .with_ansi(false)
+                    .with_span_events(fmt::format::FmtSpan::FULL),
+            );
+            guards.push(span_rolling_writer_guard);
+        }
```
drmingdrmer, line 95: By default metasrv logs events  in the span-format.
In a CI it is by default on and it's saved my `a$$` several times on debugging a distributed service.

Shall we turn the jaeger on for our CI? I am not sure if there are any side effects with jaeger enabled and the bunyun service off, in a CI env.

Otherwise, what about always enabling span logging by default, or enable span logging by setting a different env-var?

Except this, everything else looks great to me :DDD

dantengsky: > By default metasrv logs events in the **span-format.**
> In a CI it is by default on and it's saved my a$$ several times on debugging a distributed service.

got it.   let's keep it.

BohuTANG: I am not sure if we enable log span by default(Like in databend-query), does it affects the performance?

drmingdrmer: At least in non-debug mode, i.e., the `log_level` does not enable a logging, the cost is about the same as an atomic read(`LevelFilter::current()`). 

In a debug mode with log_level = debug, the performance degradation can be perceived by a human. 

The expanded `tracing::debug!()`:
```
{
        if ::tracing::Level::DEBUG <= ::tracing::level_filters::STATIC_MAX_LEVEL
            && ::tracing::Level::DEBUG <= ::tracing::level_filters::LevelFilter::current() {
            use ::tracing::__macro_support::*;
            static CALLSITE: ::tracing::__macro_support::MacroCallsite = {
                use ::tracing::__macro_support::MacroCallsite;
                static META: ::tracing::Metadata<'static> = {
                    ::tracing_core::metadata::Metadata::new(
                        concat!(
                        "event ",
                        file!( ),
                        ":",
                        line!( )
                        ),
                        module_path!( ),
                        ::tracing::Level::DEBUG,
                        Some(file!( )),
                        Some(line!( )),
                        Some(module_path!( )),
                        ::tracing_core::field::FieldSet::new(&[stringify!( message )], ::tracing_core::callsite::Identifier((&CALLSITE))),
                        ::tracing::metadata::Kind::EVENT,
                    )
                };
                MacroCallsite::new(&META)
            };
            let interest = CALLSITE.interest();
            if !interest.is_never() && CALLSITE.is_enabled(interest) {
                let meta = CALLSITE.metadata();

                ::tracing::Event::dispatch(
                    meta,
                    &{
                        #[allow(unused_imports)]
                        use ::tracing::field::{debug, display, Value};
                        let mut iter = (meta.fields()).iter();
                        (meta.fields()).value_set(&[(&(iter.next().expect("FieldSet corrupted (this is a bug)")), Some(&format_args!("id:{} replicate to sm index:{}", self.id, entry.log_id.index
                        ) as &Value))])
                    },
                );
            }
        }
    }
```

BohuTANG: Ok, I am assuming all the span log in debug level. 
I will enable the span by default later.
Thank you @drmingdrmer .

BohuTANG:
Demo show:

website/databend/docs/dev/development/tracing.md
```diff
@@ -150,3 +150,24 @@ Jun 10 16:40:36.167 DEBUG ThreadId(309) databend_query::pipelines::transforms::t
 Jun 10 16:40:36.168 DEBUG ThreadId(309) databend_query::pipelines::transforms::transform_expression_executor: (projection executor) execute, actions: [Input(ActionInput { name: "(sum((number + 1)) + 1)", return_type: UInt64 })]
 Jun 10 16:40:36.168 DEBUG ThreadId(309) databend_query::pipelines::transforms::transform_projection: Projection cost: 241.864µs
 ```
+## Distributed tracing with Jaeger
+
+### Start Databend
+```
+QUERY_LOG_LEVEL="DEBUG" ./databend-query
```
dantengsky, line 157: is it a typo? seems it sould be `LOG_LEVEL="DEBUG" ./databend-query`

BohuTANG: Fixed, thank you.

BohuTANG:
There is a question:
why this sql not show spans in Jaeger:
```
mysql> select sum(a+1)+1 from t1 where a>0 group by a%3;
```

But the insert works:
```
insert into t1 values (1);
```
image

dantengsky (approved):
LGTM

dantengsky:
> There is a question: why this insert sql not show spans in Jaeger:

`InsertInterpreter::execute` has not be instrumented yet

dantengsky:
> There is a question: why this sql not show spans in Jaeger:
> 
> ```
> mysql> select sum(a+1)+1 from t1 where a>0 group by a%3;
> ```

2021-12-26-204654_1595x904_scrot

works on main branch

dantengsky:
@BohuTANG just tested this PR locally, 

> select sum(a+1)+1 from t1 where a>0 group by a%3;

tracing of above sql works in local dev box

dantengsky:
@BohuTANG 

we may need to set env var  `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` to 1, in case the batch span message is too large in that case, there will be error msgs shown in stderr, or in the `nohup` file.

if LOG_LEVEL / RUST_LOG is set to `debug` "globally"(uniformly to all the modules), there will be many spans traced in Jager, be sure to choose the right span to investigate.

BohuTANG:
> @BohuTANG
> 
> we may need to set env var `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` to 1, in case the batch span message is too large(and dropped), in that case, there will be error msgs shown in stderr, or in the `nohup` file.
> 
> if LOG_LEVEL / RUST_LOG is set to `debug` "globally"(uniformly to all the modules), there will be many spans traced in Jager, be sure to choose the right span to investigate.

I think the jaeger distributed tracing only used under some special cases, not a normal tracing, so assmue that the user know what he did, but it would be better to introduce `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` in the document.

dantengsky:
> I think the jaeger distributed tracing only used under some special cases, not a normal tracing, so **assmue that the user know what he did**, but it would be better to **introduce OTEL_BSP_MAX_EXPORT_BATCH_SIZE in the document**.

can not agree more

PR #3465, opened 2021-12-14
[Issue-3367] Using serde to ser/de Aggregate function states in binary format

ygf11:
I hereby agree to the terms of the CLA available at: https://databend.rs/policies/cla/

## Summary

1. Using serde to ser/de Aggregate function states in binary format.
2. Clean up non used BinarySer/BinaryDe code.

## Changelog

- Improvement
- Performance Improvement

## Related Issues

Fixes #3367

## Test Plan

Unit Tests

Stateless Tests

PR #3649, opened 2021-12-25
Update bincode version

sundy-li:

## Summary
Update bincode version

cc @ygf11 

Summary about this PR

## Changelog

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,8 +612,7 @@ dependencies = [
 [[package]]
 name = "bincode"
 version = "1.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ygf11:
Looks good to me.

BohuTANG:
Why we maintain bincode crate in datafuselabs-extras?

sundy-li:
> Why we maintain bincode crate in datafuselabs-extras?

This is because we always know that the vector state has small size ranges, so it benefits a lot to use variant encoding. But for other integers, the range may be random, it's better to use native encoding.

From PR #3465

PR #3654, opened 2021-12-26
[ISSUE-3539] Optimize aggregate states se/deserialize encoding

### LINKED ISSUE #3539: Optimize to ser/deserialize  bytes arrays/vectors length  in varint encoding
Yes, but I don't want it to be a global setting.

Could it be possible to just write the `offsets`/ `length`   of (vectors,  byte arrays ) in variant encoding?
For other integers, we just use native encoding.

This is because we always know that the vector state has small size ranges, so it benefits a lot to use variant encoding. But for other integers, the range may be random, it's better to use native encoding.

_Originally posted by @sundy-li in https://github.com/datafuselabs/databend/issues/3465#issuecomment-993429044_

ygf11:

## Summary

Optimize aggregate states se/deserialize encoding.

## Changelog

- Improvement
- Performance Improvement

## Related Issues

Fixes #3539

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
LGTM. Do we need a new inline function to build the options?

ygf11:
> LGTM. Do we need a new inline function to build the options?

agree, but the involved option is with private trait when as a return type.

How about wapper serialize together?
```rust
#[inline]
fn serialize_into<W,T>(writer:W, value: &T) -> Result<()>  {
   bincode::DefaultOptions::new()
            .with_fixint_encoding()
            .with_varint_length_offset_encoding()
            .serialize_into(writer, self)?;

   Ok(())
}

```

sundy-li:
Ok, Just do not want to change all aggregate function files  when changes are related to bincode.

ygf11:
I mean to add wrapper method here in `common-io/src/utils.rs`, and it will not change any to forked bincode.

sundy-li:
> common-io/src/utils.rs

That's ok.

PR #3655, opened 2021-12-26
Support user defined function

### LINKED ISSUE #3440: Support user defined functions
**Summary**

Currently we need to write different implementation for invertible functions, for example, `isnull` & `isnotnull`. Can we create kind of function alias to avoid that? For `isnotnull`, we can have `isnotnull = not(isnull($1))`, when we parse the expression `isnotnull`, it can be rewrited as `not(isnull($1))`.

lianghanzhen:

## Summary

Support user defined function

- All UDFs will be stored in meta KV store like user stages.
- UDFs will be verified before stored which they must be: a) Not builtin scalar or aggregate functions; b) Params must match format `@\d+` and starts with `@0`; c) Params much be continuous, such as `@0, @1, @2`; d) The order of params doesn't matter; e) Params are repeatable;
- UDFs will be transformed into the combination of builtin scalar and aggregate functions in `analyze_expr` phase;
- `CREATE/DROP/SHOW` function is supported;

Examples are available in stateless tests.

## Changelog

- New Feature

## Related Issues

Fixes #3440

## Test Plan

Unit Tests

Stateless Tests


lianghanzhen:
@BohuTANG The stateless test of `05_0012_ddl_show_udf` failed, but I couldn't figure out why.

Output diff:
```
@@ -1 +1 @@
-isnotempty     not(isnull(@0)) This is a description
+isnotempty     not(isnull(@0)) This is a description
```

BohuTANG:
> @BohuTANG The stateless test of `05_0012_ddl_show_udf` failed, but I couldn't figure out why.
> 
> Output diff:
> 
> ```
> @@ -1 +1 @@	
> -isnotempty     not(isnull(@0)) This is a description
> +isnotempty     not(isnull(@0)) This is a description
> ```

`isnotempty	not(isnull(@0))	This is a description` -- The result should be a tab not 4 blanks :)

sundy-li:
Awesome, really surprised me that this pr can be landed in such a swift speed.

query/src/sql/statements/analyzer_expr.rs
```diff
@@ -374,208 +373,132 @@ impl ExprRPNItem {
...
-            Expr::Subquery(subquery) => self.visit_subquery(subquery),
-            Expr::Function(function) => self.visit_function(function),
-            Expr::Cast { expr, data_type } => self.visit_cast(expr, data_type),
-            Expr::TypedString { data_type, value } => self.visit_typed_string(data_type, value),
-            Expr::Position {
-                substr_expr,
-                str_expr,
-            } => self.visit_position(substr_expr, str_expr),
+            Expr::Value(value) => {
+                self.rpn.push(ExprRPNItem::Value(value.clone()));
```
sundy-li, line 392: That's much better now, no extra function call.

sundy-li (approved):

Cargo.lock
```diff
@@ -1563,6 +1567,14 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "common-sql"
+version = "0.1.0"
+dependencies = [
+ "common-exception",
+ "sqlparser",
+]
+
```
leiysky, line 1577: It seems this crate only involves `expr_visitor`, what about rename it to `common-ast`? Since the name `common-sql` looks too common.

BohuTANG:
Great job, let's merge!
Thank you @lianghanzhen !

Cargo.lock
lianghanzhen: I don't have a strong opinion about the naming, I will rename it in other MR.

PR #3684, opened 2021-12-28
Use expression as UDF definition

lianghanzhen:

## Summary

Use expression as UDF definition which is the first step of [lambda style UDF](https://github.com/datafuselabs/databend/discussions/3672)

- support expression as definition
- replace `=` with `AS` keyword to define UDF

## Related Issues

Fixes #issue

## Changelog

- Improvement

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
@mergify update

query/src/sql/sql_parser.rs
```diff
@@ -854,23 +854,108 @@ impl<'a> DfParser<'a> {
...
+            if next_token == Token::EOF || next_token == Token::SemiColon {
+                break;
+            }
+            if next_token == Token::RParen {
+                found_right_paren = true;
+                break;
+            }
+
+            match next_token {
+                Token::Word(Word {
```
sundy-li, line 874: `self.parser.parse_comma_separated` can be addressed.

query/src/sql/sql_parser.rs
```diff
@@ -854,23 +854,108 @@ impl<'a> DfParser<'a> {
...
+        }
+
+        Ok(params)
+    }
+
+    fn parse_udf_definition_expr(&mut self, until_token: Vec<&str>) -> Result<String, ParserError> {
+        // Match ->
+        self.parser.expect_token(&Token::Minus)?;
+        let next_token = self.parser.next_token_no_skip();
+        if next_token != Some(&Token::Gt) {
```
sundy-li, line 924: self.parser.expect_token(&Token::Gt)?;

query/src/sql/sql_parser.rs
```diff
@@ -854,23 +854,108 @@ impl<'a> DfParser<'a> {
...
-            self.parser.expect_token(&Token::Eq)?;
+        let as_token = Token::make_keyword("AS");
+        self.parser.expect_token(&as_token)?;
+
+        let desc_token = "DESC";
+        let parameters = self.parse_udf_parameters()?;
+        let definition = self.parse_udf_definition_expr(vec![desc_token])?;
+
+        let description = if self.consume_token(desc_token) {
+            self.parser.expect_token(&as_token)?;
```
sundy-li, line 950: how about keep using `desc = 'desc_msg' `, it's strange to keep two `as` in a single sql.

query/src/sql/sql_parser.rs
lianghanzhen: Don't think so. `expect_token` will skip the whitespace, what we need here is `->`.

query/src/sql/sql_parser.rs
lianghanzhen: Checked this method, it parses a comma-separated list of 1+ items, we need to handle zero param as well.

query/src/sql/sql_parser.rs
lianghanzhen: Updated.

BohuTANG:
Thank you @lianghanzhen!

PR #3700, opened 2021-12-30
Improve origin array builder & make nullable/no-nullable builder support

### LINKED ISSUE #3692: Making `ArrayBuilder` write value to inner value directly, rather than call `arrow2`'s `try_push`
Now `ArrayBuilder` will call have `MutableArray` as the inner data structure. and then the inner `push` will add val with option, just like below:

https://github.com/datafuselabs/databend/blob/1b6bf394d7/common/datavalues/src/arrays/primitive/builder.rs#L37-L44

then the `push` will call arrow2's `try_push` to do real push

https://github.com/datafuse-extras/arrow2/blob/565cfaeb02/src/array/binary/mutable.rs#L407-L433

the code above will have extra branch predicting cost, so we will want to have write values dirctly to the values. and this will work  more effectively with no `Nullable` datatype, as mentioned in https://github.com/datafuselabs/databend/issues/3664

cc @sundy-li

Veeupup:

## Summary

* improve `MutableBooleanArrayBuilder`, `MutablePrimitiveArrayBuilder` and `MutableStringArrayBuilder`, to push value directly rather then call `try_push` to avoid branch prediction cost
* add nullable generic builder support. For example, if we determine that a certain column is with no nullable value, we can have a more efficient arraybuilder, just like below:

```rust
// with no nullable value
let builder = MutableStringArrayBuilder::<false>::default(); // false means no nullable, true means the value may be null
builder.push(val);
```

## Changelog

- Improvement

## Related Issues

Fixes #3692 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/arrays/boolean/mutable.rs
+++ b/common/datavalues/src/arrays/boolean/mutable.rs
@@ -30,8 +30,8 @@ pub struct MutableBooleanArrayBuilder {
 }
 
 impl MutableArrayBuilder for MutableBooleanArrayBuilder {
-    fn data_type(&self) -> DataType {
-        DataType::Boolean
+    fn data_type(&self) -> &DataType {
+        &self.data_type
     }
 
     fn as_any(&self) -> &dyn std::any::Any {
@@ -43,9 +43,8 @@ impl MutableArrayBuilder for MutableBooleanArrayBuilder {
     }
 
     fn as_series(&mut self) -> Series {
-        let arrow_data_type = self.data_type.to_arrow();
         let array: Arc<dyn Array> = Arc::new(BooleanArray::from_data(
-            arrow_data_type,
+            self.data_type.to_arrow(),
             std::mem::take(&mut self.values).into(),
             std::mem::take(&mut self.validity).map(|x| x.into()),
         ));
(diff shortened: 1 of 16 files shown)


common/datavalues/src/arrays/boolean/mutable.rs
```diff
@@ -71,15 +78,27 @@ impl MutableBooleanArrayBuilder {
...
             values: MutableBitmap::with_capacity(capacity),
-            validity: None,
+            validity: Some(MutableBitmap::with_capacity(capacity)),
         }
     }
 
-    pub fn from_data(values: MutableBitmap, validity: Option<MutableBitmap>) -> Self {
+    pub fn push(&mut self, value: bool) {
+        self.values.push(value);
+        self.validity.as_mut().unwrap().push(true);
```
sundy-li, line 87: Maybe no need to care about `validity` for `not nullable` case, the validity is always none.

common/datavalues/src/arrays/string/mutable.rs
```diff
@@ -59,13 +59,20 @@ impl MutableArrayBuilder for MutableStringArrayBuilder {
...
 
-impl Default for MutableStringArrayBuilder {
+impl Default for MutableStringArrayBuilder<true> {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl MutableStringArrayBuilder {
+impl Default for MutableStringArrayBuilder<false> {
```
sundy-li, line 68: Can we combine them into one?

Veeupup: the compiler does not allow me to do so : (

common/datavalues/src/arrays/boolean/mutable.rs
Veeupup: yes, it it redundant.

zhang2014 (approved):
Look good to me.

sundy-li (approved):

Issue #3689, opened 2021-12-29
Conflict between python2 and python3 when running locally


yihong0618:
As pic shows.

When run locally it may `pip as python3` and run as python2 and raise the import error 

![image](https://user-images.githubusercontent.com/15976103/147624460-b0f83873-0a15-4b3a-b2a5-4222421ac47f.png)

something like.
![image](https://user-images.githubusercontent.com/15976103/147624591-43447a07-2bb5-43ab-bca4-df28787056a2.png)

PsiACE:
closed by #3690

also, I don't think we would consider supporting Python2.

yihong0618:
@PsiACE I think the code [wait_tcp](https://github.com/datafuselabs/databend/blob/main/scripts/ci/wait_tcp.py) is running in Python2 do you want to make it to Python3?

PsiACE:
> @PsiACE I think the code [wait_tcp](https://github.com/datafuselabs/databend/blob/main/scripts/ci/wait_tcp.py) is running in Python2 do you want to make it to Python3?

Certainly. It is well known that python2 is at the end of its life cycle.

yihong0618:
> > @PsiACE I think the code [wait_tcp](https://github.com/datafuselabs/databend/blob/main/scripts/ci/wait_tcp.py) is running in Python2 do you want to make it to Python3?
> 
> Certainly. It is well known that python2 is at the end of its life cycle.

Cool, can I take a try?

PsiACE:
> Cool, can I take a try?

Thank you for your willingness to help, please feel free to submit a PR

PR #3701, opened 2021-12-30
feat: wait_tcp.py to python3

yihong0618:

## Summary

Summary about this PR

## Changelog

- Improvement
- Build/Testing/CI

## Related Issues

As discussed in #3689 port the `wait_tcp.py` to Python3. 
And use `context manager` to make sure the sock close.

### DIFF
--- a/scripts/deploy/databend-query-cluster-3-nodes.sh
+++ b/scripts/deploy/databend-query-cluster-3-nodes.sh
@@ -43,7 +43,7 @@ if [ "$mode" == "boot" ]; then
 		--log-dir ./_logs1 \
 		--raft-api-port 28103 \
 		&
-	python scripts/ci/wait_tcp.py --timeout 5 --port 9191
+	python3 scripts/ci/wait_tcp.py --timeout 5 --port 9191
 
 	nohup ./target/debug/databend-meta \
 		--id 2 \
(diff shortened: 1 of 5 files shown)


scripts/deploy/databend-query-cluster-3-nodes.sh
```diff
@@ -43,7 +43,7 @@ if [ "$mode" == "boot" ]; then
 		--log-dir ./_logs1 \
 		--raft-api-port 28103 \
 		&
-	python scripts/ci/wait_tcp.py --timeout 5 --port 9191
+	python3 scripts/ci/wait_tcp.py --timeout 5 --port 9191
```
flaneur2020, line 46: is the `python3` command available in the major os environment like Darwin, ubuntu, fedora? q.q

yihong0618: ubuntu and Darwin I think the answer `yes` 

And I saw `python3` command in `dev_setup` https://github.com/datafuselabs/databend/blob/main/scripts/setup/dev_setup.sh#L298

or maybe we can make the command from `setup_dev` too?

flaneur2020: agree, we can take the python commands the same among the script files

yihong0618: users `env` may have both `python and python3` we can get the **python env command** in setup  ln the python3 command.

yihong0618:
@flaneur2020 

and I noticed the `bot` comment same message many times seems based on the label count.
image

BohuTANG:
@yihong0618 Thank you!

PR #3710, opened 2021-12-31
Add priv check

### LINKED ISSUE #3170: verify privilege on write statements
**Summary**

parent issue: https://github.com/datafuselabs/databend/issues/2894

including:

- [x] insert/truncate table
- [x] create/drop database
- [x] create/drop tables
- [ ] create/drop users

flaneur2020:

## Summary

Summary about this PR

- [x] add checks on write operations
- [x] find user identity with client ip on authenticating user.
- [x] take default@127.0.0.1 to run the stateless tests

## Changelog

- New Feature

## Related Issues

Fixes #3170

## Test Plan

Unit Tests

Stateless Tests


common/meta/types/src/user_grant.rs
```diff
@@ -205,40 +205,27 @@ impl UserGrantSet {
         &self.grants
     }
 
-    pub fn verify_global_privilege(
+    pub fn verify_privilege(
```
BohuTANG, line 208: common/meta/types mainly for type definiton.
Maybe better to put the `verify_privilege` in other place?

flaneur2020: after adding roles, we might need a seperate PrivilegeManager to cache and verify the user privileges (it needs cache all the roles' privileges, and generate an effective grant set with the privileges merged from the roles), maybe the PrivilegeManager could be the better place to verify the privilege, and leaving the types user_grant mod as plain data objects.  🤔

this refactor requires some work, maybe we can make this refactor in a stand-alone PR? q.q

BohuTANG: That's OK

query/src/sessions/session.rs
```diff
@@ -168,6 +170,28 @@ impl Session {
         self.mutable_state.set_current_user(user)
     }
 
+    pub fn validate_privilege(
```
BohuTANG, line 173: check_privilege is better?

flaneur2020: there're some `validate_` and `check_` functions in our codebase, as the existing convention, the `check_` functions often return a bool and used in an if condition, and the `validate_` functions often returns an Result<()>, and mostly used as a short-circuit call with a `?`, which breaks the control flow if any error occurred, like:

```
validate_expression(&expr)?;
```

```
   57 fn validate_function_arg(
   58     name: &str,
   59     args_len: usize,
   60     variadic_arguments: Option<(usize, usize)>,
   61     num_arguments: usize,
   62 ) -> Result<()> {```

in this case, the `validate_privilege` function would return an Result<()> and used as a short-circuit call with a `?`, i guess naming it as `validate_` follows the convension in the current codebase? 🤔

BohuTANG: Good 👍

PR #3718, opened 2021-12-31
ISSUE-3709: table read performance tweaking

### LINKED ISSUE #3709: fuse table read performance tweaking
- use BufferReader instead of raw Read/InputStream
  and add a setting for buffer size
- Pass block length from `read_plan` to `read`
  S3InputStream could use this hint to avoid HTTP requests during seeking

dantengsky:

## Summary

- use BufferReader instead of raw Read/InputStream
     - add a setting option `read_buffer_size` to `settings`, default value set to 1MB
     - souce_parquet uses `BufReader` to read from the underlying inputstream
          
- Pass block length from read_plan to read
     - length of block is encoded as part of the `Par::name`, and decoded while executing the `Table::read`
     - length of block is passed to `InputStream`'s constructor, such that
         S3InputStream could use this hint to avoid HTTP requests during seeking
     NOTE: this modification also brings an **incompatible**  change to the block meta
## Changelog

- Performance Improvement

## Related Issues

Fixes #3709

## Test Plan

Unit Tests

Stateless Tests


query/src/sessions/settings.rs
```diff
@@ -32,7 +32,8 @@ impl Settings {
         ("flight_client_timeout", u64, 60, "Max duration the flight client request is allowed to take in seconds. By default, it is 60 seconds"),
         ("min_distributed_rows", u64, 100000000, "Minimum distributed read rows. In cluster mode, when read rows exceeds this value, the local table converted to distributed query."),
         ("min_distributed_bytes", u64, 500 * 1024 * 1024, "Minimum distributed read bytes. In cluster mode, when read bytes exceeds this value, the local table converted to distributed query."),
-        ("parallel_read_threads", u64, 1, "The maximum number of parallelism for reading data. By default, it is 1.")
+        ("parallel_read_threads", u64, 1, "The maximum number of parallelism for reading data. By default, it is 1."),
+        ("read_buffer_size", u64, 1024 * 1024, "The size of buffer in bytes for buffered reader of dal, default value is 1MB")
```
BohuTANG, line 36: How about add a prefix to name like `storage_read_buffer_size`?

dantengsky:
@BohuTANG  PTAL
- `read_buffer_size` renamed to `storage_read_buffer_size`
-  `test_interpreter_interceptor` has been adjusted
   but to be honest, I do not know why the `memory_usage` increased, seems modifications of this PR should not have this effect.

BohuTANG:
Looks good 👍 
By the way, how much performance will gain with this PR?

Veeupup (approved):

dantengsky:
> how much performance will gain with this PR?

no dedicated benchmark for this PR yet. In the last day's benchmark (aws s3), **If recalled correctly**, after applying similar modifications (but not identical), for some cases(scan insensitive), the average execution time decreased about ~50%, but for other cases(scan + computation both heavy), no significant improvement found.  ultimately, the block size (or the number of blocks) matters the most (for the scenarios under benchmark).

PR #3722, opened 2022-01-01
[toolchain] nightly-2022-01-01

PsiACE:

## Summary

Happy New Year! 

- toolchain: nightly-2022-01-01

We need to have a routine maintenance event that includes at least the following.

- update the toolchain
- update dependencies that have been missed
- check that our utilities are working properly

I think once in each natural month is a relatively appropriate frequency. This allows us to check between release channels and get something fresh.

## Changelog

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


.github/actions/build_debug/action.yml
```diff
@@ -18,6 +18,8 @@ runs:
     - uses: actions-rs/toolchain@v1
       with:
         target: ${{ matrix.config.target }}
+        toolchain: nightly
```
BohuTANG, line 21: We have defined the nightly in the rust-toolchain, why here still need? :thinking:

common/clickhouse-srv/src/types/block/mod.rs
```diff
@@ -240,12 +240,14 @@ impl<K: ColumnType> Block<K> {
     }
 
     /// Add new column into this block
+    #[must_use]
```
BohuTANG, line 243: What's the `[must_use]` attribute for

.github/actions/build_debug/action.yml
PsiACE: emm, rust-toolchain.toml hints at the format of the file, which I think is better. Unfortunately, action-rs/toolchain does not seem to support direct processing and will prompt that the toolchain file cannot be found.

common/clickhouse-srv/src/types/block/mod.rs
PsiACE: https://rust-lang.github.io/rust-clippy/master/index.html#return_self_not_must_use

> Methods returning Self often create new values, having the #[must_use] attribute prevents users from “forgetting” to use the newly created value.

PsiACE:
emm, `servers::clickhouse::test_clickhouse_insert_to_fuse_table` is a flaky test, and I will be re-running the ci

BohuTANG:
The CI failed is not related to this change, let's merge.

Issue #3664, opened 2021-12-27
Move Nullable to be a DataType


sundy-li:
### Discussed in https://github.com/datafuselabs/databend/discussions/3460

Let's introduce `DataType::Nullable(DataType)` support.

1. Add DataType `DataType::Nullable(DataType)`
2. Add `DataType::Nothing`
3. Migrate `DataType::Null` into `DataType::Nullable(Nothing)`

For table column query:

1.  `column UInt32 not null` will create column with datatype `DataType::UInt32` .
2.  `column UInt32` will create column with datatype `Nullable(DataType::UInt32)` .

zhang2014:
We also should consider merging `function::nullable` and `function::return_ type`

junli1026:
Some tasks I can think of:
- [ ] Remove function's method ```nullable```. The method ```return_type``` should return ```DataTypeAndNullable```, which carries both ```data_type``` and ```nullable``` -- thus we know the nullable information of the return value.
- [ ] Arithmetic and comparison functions should support ```DataTypeAndNullable```
- [ ] Add new data_type ```Nullable(Box<DataType>)``` or ```NotNull(Box<DataType>)```  -- this new type is to replace ```DataTypeAndNullable```
- [ ] Match Mysql/Clickhouse type with new ```Nullable```

junli1026:
@sundy-li maybe we can close this one.

PR #3726, opened 2022-01-02
Add nullable data type  --part 1

junli1026:

## Summary

The PR introduced nullable data type.

## Changelog

- Improvement

## Related Issues

Related #3664

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
We have updated rust-toolchain, there are some links error:
error: missing `#[must_use]` attribute on a method returning `Self`
1565
   --> common/datavalues/src/types/data_type.rs:132:5
1566
    |
1567
132 | /     pub fn enforce_nullable(&self) -> DataType {
1568
133 | |         match self {
1569
134 | |             DataType::Boolean(false) => DataType::Boolean(true),
1570
135 | |             DataType::UInt8(false) => DataType::UInt8(true),
1571
...   |
1572
152 | |         }
1573
153 | |     }
1574
    | |_____^

BohuTANG:
@mergify update

junli1026:
@sundy-li  @zhang2014 @zhyass This is PR is hard to say a mature one, but more like a proof of concept. Feel free to comment or propose any idea.  Actually I think maybe ```Nullable<Box::new<DataType>>``` makes more sense.

sundy-li:
> Actually I think maybe Nullable<Box::new<DataType>> makes more sense.

Hello @junli1026 , after some deep thinking of this, I do agree with that. Though you have already refactored the codes a lot using `DataType::XXX(Nullable)`.

Almost all computing functions do not need to care about nullable, we can do `removeNullable` for the column before computing and apply the bitmap to the result array. With `Nullable(Box<DataType>)`, it's easier to do `removeNullable` for the datatypes and it's more simple to use `DataType::UInt64` rather than `DataType::UInt64(_)` in other places.

zhang2014:
> > Actually I think maybe Nullable<Box::new> makes more sense.
> 
> Hello @junli1026 , after some deep thinking of this, I do agree with that. Though you have already refactored the codes a lot using `DataType::XXX(Nullable)`.
> 
> Almost all computing functions do not need to care about nullable, we can do `removeNullable` for the columns/datatypes before computing and apply the bitmap to the result array. With `Nullable(Box<DataType>)`, it's easier to do `removeNullable` for the datatypes and it's more simple to use `DataType::UInt64` rather than `DataType::UInt64(_)` in other places.

Agree + 1.

junli1026:
@sundy-li  @zhang2014  @zhyass  
Cool! Feels like we all agree that ```Nullable(Box<DataType>) ``` is more clean and makes more sense. Thus let's do it in  the right way!  I will mark the PR as Draft and keep working on it with ```Nullable(Box<DataType>)``` .  Hopefully I can finish it in this week. 😄

junli1026:
Close this PR since I have another one to address the same topic.

PR #3733, opened 2022-01-03
Fix benchmark hang

### LINKED ISSUE #3723: bug: databend-benchmark can't work
### Search before asking

### Version

- Databend Git Version - 11fa94b
- Fedora 35 (5.15.11-200 x86_64 GNU/Linux)

### What's Wrong?

Executing any query with a proper connection to databend's clickhouse server will cause databend-benchmark to hang.

### What You Expected?

Execute the query and get the correct benchmarking results.

### How to Reproduce?

```shell
echo "select avg(number) from numbers(1000000)" |  ./target/debug/databend-benchmark -c 1  -i 10 -p 9001
```

### Anything Else?

It looks like the problem was caused when the [clickhouse-driver](https://github.com/datafuse-extras/clickhouse_driver/) was introduced. ( #3682 )

sundy-li:

## Summary

Summary about this PR

## Changelog

- Bug Fix

## Related Issues

Fixes #3723

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/tools/benchmark/src/main.rs
+++ b/tools/benchmark/src/main.rs
@@ -236,7 +236,7 @@ async fn execute(bench: BenchmarkRef) -> Result<()> {
                         .await
                         .map_err_to_code(ErrorCode::LogicalError, || "")?;
 
-                    while result.next().await.is_ok() {}
+                    while let Some(_) = result.next().await.unwrap() {}
 
                     let progress = &result.progress;
                     let mut stats = bench.stats.write();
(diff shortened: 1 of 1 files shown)


tools/benchmark/src/main.rs
```diff
@@ -236,7 +236,7 @@ async fn execute(bench: BenchmarkRef) -> Result<()> {
                         .await
                         .map_err_to_code(ErrorCode::LogicalError, || "")?;
 
-                    while result.next().await.is_ok() {}
+                    while let Some(_) = result.next().await.unwrap() {}
```
PsiACE, line 239: LGTM. It works fine. But need to make clippy happy.

PsiACE (approved):

PR #3757, opened 2022-01-04
ISSUE-3225: Make parquet reader works with only one stream reader

### LINKED ISSUE #3225: Make parquet reader works with only one stream reader.
**Summary**

Description for this feature.

Currently, we are unable to stream load parquet files into databend since the parquet source need `Dal` to generate streams multiple times(parallel column reading from parquet column need it).

https://github.com/datafuselabs/databend/blob/f3ecaf16b5290fbc26870660efb2b8d059299442/common/streams/src/sources/source_parquet.rs#L101-L118

Sometimes the input stream may come from outside, like HTTP body streaming in 

https://github.com/datafuselabs/databend/blob/f3ecaf16b5290fbc26870660efb2b8d059299442/query/src/servers/http/v1/load.rs#L84-L97

@dantengsky  @jorgecarleitao  any comments to help it are welcome.

dantengsky:

## Summary

-  `common_streams::ParquetSource`
   -  serially read the columns, so that
   - only "one stream reader" is needed during the processing of the parquet source
   - An `AysncRead + AsyncSeek` reader  is  enough, no longer depend on the whole DataAccessor trait
  
-  Add a dedicated `BlockReader` to fuse storage,  which will
    -  keeps reading the columns parallelly
    -  with different signature
    
        e.g. returns `Result<Datablock>` instead of `Result<Option<DataBlock>>`

        Different from `ParquetSource`, in which the caller site is supposed to keep calling the `Source::read` method, till `Err` or `Result<None>` is returned. `BlockReader` is supposed to be used while reading columns of parquet as a block in one go.

- @Veeupup 
  - Field `cache` has been removed from `BlockReader`
    since for the time being, it is only used during the construction phase of `BlockReader`. may we add it back later if necessary?
    
    
## Changelog

- Improvement

## Related Issues

Fixes #3225

## Test Plan

Unit Tests

Stateless Tests


common/streams/src/sources/source_parquet.rs
```diff
@@ -31,141 +30,94 @@ use common_exception::Result;
...
-        let buffer_size = 10;
-        let n = std::cmp::min(buffer_size, col_num);
-        let data_cols = stream.buffered(n).try_collect().await?;
-
-        self.row_group += 1;
+            .map(|idx| (row_grp.column(idx).clone(), idx));
+        let mut data_cols = Vec::with_capacity(cols.len());
+        for (col_meta, idx) in cols {
+            let col_pages =
+                get_page_stream(&col_meta, &mut self.reader, vec![], Arc::new(|_, _| true))
```
sundy-li, line 109: So we do not read columns in parallel, will it slow down the performance?

dantengsky: for fuse engine, the newly added `BlockReader` will be used, which keeps reading the columns parallelly.

Veeupup:
@dantengsky  
agree!  `storage cache` may read a column from the cache in `BlockReader`, we can add it when it is ready.

Veeupup (approved):
/LGTM

PR #3761, opened 2022-01-04
Fuse metadata cache

Veeupup:

## Summary

* refactor fuse cache
* add block reader
* cache parquet metadata in memory

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/io/block_reader.rs
```diff
@@ -0,0 +1,184 @@
...
+pub struct BlockReader {
+    data_accessor: Arc<dyn DataAccessor>,
+    path: String,
+
+    block_schema: DataSchemaRef,
+    arrow_table_schema: ArrowSchema,
+    projection: Vec<usize>,
+    row_group: usize,
+    row_groups: usize,
+    metadata: Option<FileMetaData>,
```
dantengsky, line 52: hi, how about make this field `BlockReader::metadata` typed as  `FileMeataData`, instead of (`Option<_>`)?

Veeupup: yes, it seems we have cache metadata inside. thanks for reminding me, I'll fix it.

dantengsky: yeah, maybe we cloud pass the metadata to `BlockReader` from `FuseTable::do_read`

dantengsky (approved):
LGTM

PR #3769, opened 2022-01-05
Add Nullable(Box<DataType>) data type

junli1026:

## Summary

This PR is just adding nullable type, no optimization at all.

## Changelog

- Improvement

## Related Issues

Related #3664

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/arithmetics/arithmetic_plus.rs
+++ b/common/functions/src/scalars/arithmetics/arithmetic_plus.rs
@@ -110,6 +110,8 @@ impl ArithmeticPlusFunction {
             )));
         };
 
+        let interval = interval.remove_nullable();
+        let result_type = result_type.remove_nullable();
         match interval {
             DataType::Interval(IntervalUnit::YearMonth) => match result_type.clone() {
                 DataType::Date16 => {
--- a/common/functions/src/scalars/expressions/cast.rs
+++ b/common/functions/src/scalars/expressions/cast.rs
@@ -67,7 +67,12 @@ impl Function for CastFunction {
     }
 
     fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
-        if columns[0].data_type() == &self.cast_type {
+        // TODO: ideally we don't need to do remove nullable -- the input should be always non-nullable.
+        // The null value will be masked after function.eval.
+        let source_type = columns[0].data_type().remove_nullable();
+        let cast_to = self.cast_type.remove_nullable();
+
+        if source_type == cast_to {
             return Ok(columns[0].column().clone());
         }
         let series = columns[0].column().clone().to_minimal_array()?;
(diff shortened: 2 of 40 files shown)


sundy-li:
Failed in :
```
[ FAIL ] - result differs with:
--- /home/runner/work/databend/databend/tests/suites/0_stateless/09_0001_remote_insert.result	2022-01-05 02:35:31.538074279 +0000
+++ /home/runner/work/databend/databend/tests/suites/0_stateless/09_0001_remote_insert.stdout	2022-01-05 02:42:54.026295840 +0000
@@ -2,5 +2,5 @@
 2	2	"2"-"2"
 3
 3
-1	2021-09-07 21:38:35	2021-09-07
-0	2021-09-07 21:38:35	2021-09-07
+1	1631050715	18877
+0	1631050715	18877

```

common/functions/src/scalars/arithmetics/arithmetic_plus.rs
```diff
@@ -75,6 +75,7 @@ impl ArithmeticPlusFunction {
         with_match_primitive_type!(left_type, |$T| {
             with_match_primitive_type!(right_type, |$D| {
                 let result_type = <($T, $D) as ResultTypeOfBinary>::AddMul::data_type();
+                println!("{}", result_type);
```
sundy-li, line 78: should be removed.

common/functions/src/scalars/expressions/cast.rs
```diff
@@ -67,7 +67,12 @@ impl Function for CastFunction {
     }
 
     fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
-        if columns[0].data_type() == &self.cast_type {
+        // TODO: ideally we don't need to do remove nullable -- the input should be always non-nullable.
+        // The null value will be masked after function.eval.
+        let source_type = columns[0].data_type().remove_nullable();
```
sundy-li, line 72: So TODO will be addressed in another pr?

common/functions/src/scalars/others/running_difference_function.rs
```diff
@@ -51,7 +51,7 @@ impl Function for RunningDifferenceFunction {
 
     fn return_type(&self, args: &[DataTypeAndNullable]) -> Result<DataTypeAndNullable> {
         let nullable = args.iter().any(|arg| arg.is_nullable());
-        let data_type = match args[0].data_type() {
+        let data_type = match args[0].data_type().remove_nullable() {
```
sundy-li, line 54: Maybe we should not use `remove_nullable`, if there is null, it will throw error result.

common/functions/src/scalars/expressions/cast.rs
junli1026: Right, the PR just introduce the new type, doing the best to behavior same as it is now. It doesn't refactor any return_type or eval. 

The feature we talked, like
1.  moving nullable check out of return_type 
2. pass_through_null returns multiple state,
3. remove DataTypeWithNull
Will be addressed in another PR.

common/planners/src/plan_expression.rs
```diff
@@ -226,8 +226,14 @@ impl Expression {
     pub fn to_data_field(&self, input_schema: &DataSchemaRef) -> Result<DataField> {
         let name = self.column_name();
         self.to_data_type(input_schema).and_then(|return_type| {
-            self.nullable(input_schema)
-                .map(|nullable| DataField::new(&name, return_type, nullable))
+            self.nullable(input_schema).map(|nullable| {
+                println!("return_type {}", return_type);
```
sundy-li, line 230: should be removed

common/functions/src/scalars/arithmetics/arithmetic_plus.rs
junli1026: Fixed.

common/functions/src/scalars/others/running_difference_function.rs
junli1026: Right, the next PR I am going to address will change return_type to ```return_type(&self, args: &[DataType])``` .  For now, nullable data type is directly passed into this function,  thus the remove_nullable happens inside of the function. I will change the logic to move remove_nullable outside of function.rs in the next PR.

sundy-li:
I checked the codes, it LGTM.

Yet I found if we have `Nullable<DataType>`, we need to `removeNullable` and `enforceNullable` everywhere.
It may easily confuse the developers and lead to bugs.

So I just hold this pr, need future discussion of that.

junli1026:
> I checked the codes, it LGTM.
> 
> Yet I found if we have `Nullable<DataType>`, we need to `removeNullable` and `enforceNullable` everywhere. It may easily confuse the developers and lead to bugs.

Right. This is actually my pain point when I work on this feature. Most of the changes are to find the right place, put a ```remove_nullable``` there, so that we can pass all tests. 
This is a little annoying ....

1. If we use ``` DataType::UInt8(nullable)``` -- we need to check what nullable value to use every time we use data type.
2. If we use ```Nullable<DataType>```  --- we need to check when to ```remove_nullable``` or ```enforce_nullable```everywhere. 

Solution 1 is very strict, require every developer to think what type to use before they code.
Solution 2 is not that strict, but easy to make mistakes.



> So I just hold this pr, need future discussion of that.

BohuTANG:
@junli1026 , thank you for the pre-work.

PR #3770, opened 2022-01-05
[WIP]Add enum8 and enum16

GrapeBaBa:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/types/data_type.rs
+++ b/common/datavalues/src/types/data_type.rs
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 use core::fmt;
+use std::collections::BTreeMap;
 
 use common_arrow::arrow::datatypes::DataType as ArrowDataType;
+use common_arrow::arrow::datatypes::IntegerType as ArrowIntegerType;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_macros::MallocSizeOf;
@@ -62,6 +64,9 @@ pub enum DataType {
     List(Box<DataField>),
     Struct(Vec<DataField>),
     String,
+
+    Enum8(BTreeMap<String, u8>),
+    Enum16(BTreeMap<String, u16>),
 }
 
 #[derive(
@@ -196,6 +201,11 @@ impl DataType {
         }
     }
 
+    #[inline]
+    pub fn is_enum(&self) -> bool {
+        matches!(self, DataType::Enum8(_) | DataType::Enum16(_))
+    }
+
     pub fn to_arrow(&self) -> ArrowDataType {
         use DataType::*;
         match self {
@@ -223,6 +233,12 @@ impl DataType {
             }
             String => ArrowDataType::LargeBinary,
             Interval(_) => ArrowDataType::Int64,
+            Enum8(_) => {
+                ArrowDataType::Dictionary(ArrowIntegerType::UInt8, Box::new(ArrowDataType::UInt8))
+            }
+            Enum16(_) => {
+                ArrowDataType::Dictionary(ArrowIntegerType::UInt16, Box::new(ArrowDataType::UInt16))
+            }
         }
     }
 }
@@ -272,6 +288,12 @@ impl From<&ArrowDataType> for DataType {
                 let fields: Vec<DataField> = fields.iter().map(|f| f.into()).collect();
                 DataType::Struct(fields)
             }
+
+            ArrowDataType::Dictionary(_, val) => match val.as_ref() {
+                ArrowDataType::UInt8 => DataType::UInt8,
+                ArrowDataType::UInt16 => DataType::UInt16,
+                _ => unimplemented!("data_type: {}", dt),
+            },
 
             // this is safe, because we define the datatype firstly
             _ => {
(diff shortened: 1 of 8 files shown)


sundy-li:
Good try, I think we should notice that arrow has dictionary datatypes, we can use them.

GrapeBaBa:
> Good try, I think we should notice that arrow has dictionary datatypes, we can use them.

ok, let me check what is dictionary datatypes?

common/datavalues/src/types/data_type.rs
```diff
@@ -223,6 +233,12 @@ impl DataType {
             }
             String => ArrowDataType::LargeBinary,
             Interval(_) => ArrowDataType::Int64,
+            Enum8(_) => {
+                ArrowDataType::Dictionary(ArrowIntegerType::UInt8, Box::new(ArrowDataType::Null))
```
GrapeBaBa, line 237: @sundy-li is it what's your mention?

sundy-li: Yes.  

```
/// A dictionary encoded array (`key_type`, `value_type`), where
    /// each array element is an index of `key_type` into an
    /// associated dictionary of `value_type`.
    ///
    /// Dictionary arrays are used to store columns of `value_type`
    /// that contain many repeated values using less memory, but with
    /// a higher CPU overhead for some operations.
    ///
    /// This type mostly used to represent low cardinality string
    /// arrays or a limited set of primitive types as integers.
    ///
    /// The `bool` value indicates the `Dictionary` is sorted if set to `true`.
    Dictionary(IntegerType, Box<DataType>, bool),
```

common/datavalues/src/types/data_type.rs
```diff
@@ -62,6 +64,9 @@ pub enum DataType {
     List(Box<DataField>),
     Struct(Vec<DataField>),
     String,
+
+    Enum8(BTreeMap<String, u8>),
```
GrapeBaBa, line 68: @sundy-li Is this definition OK?

common/datavalues/src/types/data_type.rs
```diff
@@ -223,6 +233,12 @@ impl DataType {
             }
             String => ArrowDataType::LargeBinary,
             Interval(_) => ArrowDataType::Int64,
+            Enum8(_) => {
+                ArrowDataType::Dictionary(ArrowIntegerType::UInt8, Box::new(ArrowDataType::UInt8))
```
GrapeBaBa, line 237: @sundy-li Is this definition right？

common/datavalues/src/types/data_type.rs
sundy-li: Yes, it's ok. I'm sorry that can you hold this pr after I finish #3794 ?

We will remove enum based `DataType` so there will be conflicts.

GrapeBaBa: Yes, I already notice that PR. But I want to understand how it works e2e first. I still cannot find the where to handle the insert statement transfer enum string to uint.

query/src/sql/sql_common.rs
```diff
@@ -41,32 +43,67 @@ impl SQLCommon {
             //custom types for databend
             // Custom(ObjectName([Ident { value: "uint8", quote_style: None }])
             SQLDataType::Custom(obj) if !obj.0.is_empty() => {
-                match obj.0[0].value.to_uppercase().as_str() {
-                    "UINT8" => Ok(DataType::UInt8),
-                    "UINT16" => Ok(DataType::UInt16),
-                    "UINT32" => Ok(DataType::UInt32),
-                    "UINT64" => Ok(DataType::UInt64),
+                let value = obj.0[0].value.to_uppercase();
```
GrapeBaBa, line 46: Is it the right place to parse enum definition?

query/src/servers/mysql/writers/query_result_writer.rs
```diff
@@ -188,6 +191,14 @@ impl<'a, W: std::io::Write> DFQueryResultWriter<'a, W> {
                                     let serializer = data_type.create_serializer();
                                     row_writer.write_col(serializer.serialize_value(&val)?)?
                                 }
+                                (DataType::Enum8(values), DataValue::UInt8(Some(v))) => {
```
GrapeBaBa, line 194: Is it the right place to handle query enum transformation?

query/src/servers/http/v1/block_to_json.rs
```diff
@@ -92,6 +93,24 @@ where
         .collect()
 }
 
+fn enum_array_to_string_array_not_null<T, U>(
```
GrapeBaBa, line 96: Is it the right place to handle enum transformation for http query?

common/datavalues/src/types/data_type.rs
```diff
@@ -273,6 +289,12 @@ impl From<&ArrowDataType> for DataType {
                 DataType::Struct(fields)
             }
 
+            ArrowDataType::Dictionary(_, val) => match val.as_ref() {
+                ArrowDataType::UInt8 => DataType::UInt8,
```
GrapeBaBa, line 293: @sundy-li I am not quite clear if this is correct. I thought it cannot convert to 
```rust
Enum8(BtreeMap<String,u8>)
```

common/datavalues/src/types/data_type.rs
sundy-li: Aha, it's related to sqlparser [https://github.com/datafuse-extras/sqlparser-rs] crate.

query/src/sql/sql_common.rs
sundy-li: No, I'm afraid it's not the right place.

We should modify the codes in sqlparser, because currently it only accept a single token word as DataType which will not parse `Enum('a' = 0, 'b' = 1, 'c' = 2, 'd' = 4)`.

GrapeBaBa:
should be

PR #3786, opened 2022-01-06
RFC new datatype

sundy-li:

## Summary

Summary about this PR

Add RFC for new datatype design

## Changelog

- Documentation
 

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/website/databend/docs/dev/rfcs/query/0006-new-datatype-system.md
+++ b/website/databend/docs/dev/rfcs/query/0006-new-datatype-system.md
@@ -1,0 +1,114 @@
+# New datatype system design
+
+## Sumamry
+
+We need to redesign the `datatype` system because current implementation had some shortcomes, including but not limited to:
+
+- `DataType` is an enum type, we must use specific type after matching. For example, if we want to create deserializer/serializer by `DataType`, we should always do matching. It does not mean that match is not necessary. If we want to add more and more functions to `DataType`, matching may be very annoyment.
+
+- `DataType` represented as enum type, we can't use it as generic argument.
+
+- `DataType` may involve some nested datatypes, such as `DataType::Struct`, but we put `DataField` inside `DataType`, it's logically unreasonable。
+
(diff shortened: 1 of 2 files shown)


website/databend/docs/dev/rfcs/query/0006-new-datatype-system.md
```diff
@@ -0,0 +1,132 @@
...
+    Nothing,
+    Int8,
+    Int16,
+    Int32,
+    ...,
+    Date,
+    DateTime,
+    List,
+    Struct,
+    Nullable,
```
junli1026, line 33: Only DataTypeNullable.is_nullable() returns true right ? Then probably  we don't need the ```Nullable``` here 
DataTypeNullable.type_id() should return inner.type_id()? instead of Nullable ?

website/databend/docs/dev/rfcs/query/0006-new-datatype-system.md
```diff
@@ -0,0 +1,132 @@
...
+
+## New Datatype system design
+
+### Enum `TypeID`
+
+`TypeID` just represents the kind of `DataType`. It does not have extra attributes.
+
+```rust
+enum TypeID {
+    Nothing,
```
junli1026, line 24: Like ```select func(Null)```, Null is a valid data value, is the Nothing mapping to Null ?

website/databend/docs/dev/rfcs/query/0006-new-datatype-system.md
sundy-li: > Only DataTypeNullable.is_nullable() returns true right

Yes.

> DataTypeNullable.type_id() should return inner.type_id()

It will return `Nullable`.


```
impl IDataType for DataTypeNullable {
    // TODO
    fn type_id(&self) -> TypeID {
        TypeID::Null
    }

    fn is_nullable(&self) -> bool {
        true
    }

    fn arrow_type(&self) -> ArrowType {
        self.inner.arrow_type()
    }
}

```

PR #3790, opened 2022-01-06
[metasrv] meta support multi-tenant

### LINKED ISSUE #3759: MetaApi supports multi tenant
## Summary

It seems MetaApi is a global view, can not get a tenant database/table:

https://github.com/datafuselabs/databend/blob/dd7a004b6b9158769d79bd0502e4802bb53490a7/query/src/catalogs/backends/impls/meta_remote.rs#L79-L137

The KVApi works well with a prefix.

ariesdevil:

## Summary

meta support multi-tenant

## Changelog

- New Feature

## Related Issues

Fixes #3759 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/io/src/binary_read.rs
+++ b/common/io/src/binary_read.rs
@@ -58,13 +58,7 @@ where T: io::Read
     where V: Unmarshal<V> + StatBuffer {
         let mut buffer = V::buffer();
         self.read_exact(buffer.as_mut())?;
-        match V::try_unmarshal(buffer.as_ref()) {
-            Some(r) => Ok(r),
-            None => Err(ErrorCode::UnmarshalError(format!(
-                "unmarshal error: {:?}",
-                buffer
-            ))),
-        }
+        V::try_unmarshal(buffer.as_ref())
     }
 
     fn read_opt_scalar<V>(&mut self) -> Result<Option<V>>
--- a/common/io/src/binary_read.rs
+++ b/common/io/src/binary_read.rs
@@ -17,6 +17,8 @@ use std::mem::MaybeUninit;
 
 use common_exception::ErrorCode;
 use common_exception::Result;
+use uuid::Bytes;
+use uuid::Uuid;
 
 use crate::stat_buffer::StatBuffer;
 use crate::unmarshal::Unmarshal;
@@ -33,6 +35,7 @@ pub trait BinaryRead {
     fn read_string(&mut self) -> Result<String>;
     fn skip_string(&mut self) -> Result<()>;
     fn read_uvarint(&mut self) -> Result<u64>;
+    fn read_uuid(&mut self) -> Result<Uuid>;
 
     // in place read
     fn read_to_scalar<V>(&mut self, v: &mut V) -> Result<()>
@@ -117,5 +120,12 @@ where T: io::Read
 
             i += 1;
         }
+    }
+
+    fn read_uuid(&mut self) -> Result<Uuid> {
+        let mut buffer: Bytes = [0_u8; 16];
+        self.read_exact(buffer.as_mut())?;
+
+        Ok(Uuid::from_bytes(buffer))
     }
 }
--- a/common/io/src/binary_read.rs
+++ b/common/io/src/binary_read.rs
@@ -17,8 +17,6 @@ use std::mem::MaybeUninit;
 
 use common_exception::ErrorCode;
 use common_exception::Result;
-use uuid::Bytes;
-use uuid::Uuid;
 
 use crate::stat_buffer::StatBuffer;
 use crate::unmarshal::Unmarshal;
@@ -35,7 +33,7 @@ pub trait BinaryRead {
     fn read_string(&mut self) -> Result<String>;
     fn skip_string(&mut self) -> Result<()>;
     fn read_uvarint(&mut self) -> Result<u64>;
-    fn read_uuid(&mut self) -> Result<Uuid>;
+    fn read_tenant_id(&mut self, delimiter: u8) -> Result<String>;
 
     // in place read
     fn read_to_scalar<V>(&mut self, v: &mut V) -> Result<()>
@@ -55,7 +53,7 @@ pub trait BinaryRead {
 }
 
 impl<T> BinaryRead for T
-where T: io::Read
+where T: io::Read + io::BufRead
 {
     fn read_scalar<V>(&mut self) -> Result<V>
     where V: Unmarshal<V> + StatBuffer {
@@ -122,10 +120,10 @@ where T: io::Read
         }
     }
 
-    fn read_uuid(&mut self) -> Result<Uuid> {
-        let mut buffer: Bytes = [0_u8; 16];
-        self.read_exact(buffer.as_mut())?;
+    fn read_tenant_id(&mut self, delimiter: u8) -> Result<String> {
+        let mut buffer = vec![];
+        self.read_until(delimiter, &mut buffer)?;
 
-        Ok(Uuid::from_bytes(buffer))
+        Ok(String::from_utf8_lossy(&buffer[..buffer.len() - 1]).to_string())
     }
 }
--- a/common/io/src/binary_read.rs
+++ b/common/io/src/binary_read.rs
@@ -33,7 +33,7 @@ pub trait BinaryRead {
     fn read_string(&mut self) -> Result<String>;
     fn skip_string(&mut self) -> Result<()>;
     fn read_uvarint(&mut self) -> Result<u64>;
-    fn read_tenant(&mut self, delimiter: u8) -> Result<String>;
+    fn read_tenant(&mut self) -> Result<String>;
 
     // in place read
     fn read_to_scalar<V>(&mut self, v: &mut V) -> Result<()>
@@ -120,10 +120,10 @@ where T: io::Read + io::BufRead
         }
     }
 
-    fn read_tenant(&mut self, delimiter: u8) -> Result<String> {
+    fn read_tenant(&mut self) -> Result<String> {
         self.read_uvarint()?;
         let mut buffer = vec![];
-        let tenant_len = self.read_until(delimiter, &mut buffer)?;
+        let tenant_len = self.read_until(b'/', &mut buffer)?;
 
         let res = String::from_utf8_lossy(&buffer[0..tenant_len - 1]).to_string();
         Ok(res)
--- a/common/io/src/binary_read.rs
+++ b/common/io/src/binary_read.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::io;
+use std::io::Read;
 use std::mem::MaybeUninit;
 
 use common_exception::ErrorCode;
@@ -33,7 +34,7 @@ pub trait BinaryRead {
     fn read_string(&mut self) -> Result<String>;
     fn skip_string(&mut self) -> Result<()>;
     fn read_uvarint(&mut self) -> Result<u64>;
-    fn read_tenant_id(&mut self, delimiter: u8) -> Result<String>;
+    fn read_tenant(&mut self, delimiter: u8) -> Result<String>;
 
     // in place read
     fn read_to_scalar<V>(&mut self, v: &mut V) -> Result<()>
@@ -120,7 +121,7 @@ where T: io::Read + io::BufRead
         }
     }
 
-    fn read_tenant_id(&mut self, delimiter: u8) -> Result<String> {
+    fn read_tenant(&mut self, delimiter: u8) -> Result<String> {
         let mut buffer = vec![];
         self.read_until(delimiter, &mut buffer)?;
 
--- a/common/io/src/binary_read.rs
+++ b/common/io/src/binary_read.rs
@@ -121,9 +121,11 @@ where T: io::Read + io::BufRead
     }
 
     fn read_tenant(&mut self, delimiter: u8) -> Result<String> {
+        self.read_uvarint()?;
         let mut buffer = vec![];
-        self.read_until(delimiter, &mut buffer)?;
+        let tenant_len = self.read_until(delimiter, &mut buffer)?;
 
-        Ok(String::from_utf8_lossy(&buffer[..buffer.len() - 1]).to_string())
+        let res = String::from_utf8_lossy(&buffer[0..tenant_len - 1]).to_string();
+        Ok(res)
     }
 }
--- a/common/io/src/binary_read.rs
+++ b/common/io/src/binary_read.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::io;
-use std::io::Read;
 use std::mem::MaybeUninit;
 
 use common_exception::ErrorCode;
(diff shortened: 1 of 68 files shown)


drmingdrmer:
Strongly recommended to put the deleted file in a separate PR 🤔

ariesdevil:
> Strongly recommended to put the deleted file in a separate PR 🤔

Yes, commander.

common/meta/raft-store/src/state_machine/database_lookup.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct DatabaseLookupKey {
+    tenant_id: String,
+    delimiter: u8,
+    database_name: String,
+}
+
+impl DatabaseLookupKey {
+    pub fn new(tenant_id: String, database_name: String) -> Self {
```
BohuTANG, line 23: It would be more appropriate to change the all `tenant_id` to `tenant`, keeping them consistent with query :)

common/io/src/binary_read.rs
```diff
@@ -33,6 +33,7 @@ pub trait BinaryRead {
     fn read_string(&mut self) -> Result<String>;
     fn skip_string(&mut self) -> Result<()>;
     fn read_uvarint(&mut self) -> Result<u64>;
+    fn read_tenant(&mut self, delimiter: u8) -> Result<String>;
```
BohuTANG, line 36: read_until_delimiter is better? The common/io should know the tenant

common/meta/api/Cargo.toml
```diff
@@ -23,3 +23,4 @@ anyhow = "1.0.52"
 async-trait = "0.1.52"
 futures = "0.3.19"
 maplit = "1.0.2"
+uuid = { version = "1.0.0-alpha.1", features = ["serde", "v4"] }
```
BohuTANG, line 26: Unused ?

common/planners/Cargo.toml
```diff
@@ -26,6 +26,7 @@ bitflags = "1.3.2"
 serde = { version = "1.0.133", features = ["derive"] }
 futures = "0.3.19"
 once_cell = "1.9.0"
+uuid = { version = "1.0.0-alpha.1", features = ["serde", "v4"] }
```
BohuTANG, line 29: unused?

BohuTANG:
Leave some comments, others look good to me :+1: 
Wait @drmingdrmer to do the final review.

common/io/src/marshal.rs
```diff
@@ -117,3 +117,13 @@ impl Marshal for bool {
...
 }
+
+impl Marshal for char {
+    fn marshal(&self, scratch: &mut [u8]) {
+        let bits = *self as u32;
+        scratch[0] = bits as u8;
+        scratch[1] = (bits >> 8) as u8;
+        scratch[2] = (bits >> 16) as u8;
+        scratch[3] = (bits >> 24) as u8;
+    }
```
drmingdrmer, line 128: Does this marshaling preserve the order before and after marshaling?

common/io/src/unmarshal.rs
```diff
@@ -113,3 +113,13 @@ impl Unmarshal<bool> for bool {
...
     }
 }
+
+impl Unmarshal<char> for char {
+    fn unmarshal(scratch: &[u8]) -> char {
+        let bits = u32::from(scratch[0])
+            | u32::from(scratch[1]) << 8
+            | u32::from(scratch[2]) << 16
+            | u32::from(scratch[3]) << 24;
+        Self::from_u32(bits).unwrap_or(' ')
```
drmingdrmer, line 123: What does `unwrap_or(' ')` mean?
Why is it a space by default?

common/io/tests/it/binary_read.rs
```diff
@@ -26,10 +26,12 @@ fn test_write_and_read() -> Result<()> {
     buff.write_scalar(&8u8)?;
     buff.write_scalar(&16u16)?;
     buff.write_scalar(&32u32)?;
+    buff.write_scalar(&'🐳')?;
```
drmingdrmer, line 29: 👍

common/meta/raft-store/src/state_machine/database_lookup.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn ser(&self) -> Result<IVec, ErrorCode> {
+        let mut buf = BytesMut::new();
+
+        if buf.write_string(&self.tenant).is_ok()
+            && buf.write_scalar(&self.delimiter).is_ok()
+            && buf.write_string(&self.database_name).is_ok()
+        {
+            return Ok(IVec::from(buf.to_vec()));
+        }
+        Err(ErrorCode::MetaStoreDamaged("invalid key IVec"))
```
drmingdrmer, line 60: It's not meta-store's fault...

BTW, IMHO, an error should include the original cause in it.
In our case, the reason for which it failed to serialize tenant or database name.

common/meta/raft-store/src/state_machine/database_lookup.rs
```diff
@@ -0,0 +1,99 @@
...
+            let database_name_result = buf_read.read_string();
+            if let Ok(database_name) = database_name_result {
+                return Ok(DatabaseLookupKey {
+                    tenant,
+                    delimiter: DB_LOOKUP_KEY_DELIMITER,
+                    database_name,
+                });
+            }
+        }
+        Err(ErrorCode::MetaStoreDamaged("invalid key IVec"))
```
drmingdrmer, line 78: Similar to the above: if it fails to read tenant, what about leave some clue in the return value? e.g., `MetaStoreDamaged("invalid tenant in IVec {:?}", v)`

common/meta/raft-store/src/state_machine/database_lookup.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+impl fmt::Display for DatabaseLookupKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "DatabaseLookupKey_{}-{}",
+            self.tenant, self.database_name
```
drmingdrmer, line 87: Why not use `self.delimiter`? 🤔

common/io/src/binary_read.rs
```diff
@@ -118,4 +119,13 @@ where T: io::Read
             i += 1;
         }
     }
+
+    fn read_tenant(&mut self) -> Result<String> {
+        self.read_uvarint()?;
```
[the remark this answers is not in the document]
ariesdevil: Yup, just advance cursor

common/meta/raft-store/src/state_machine/sm_meta_api_impl.rs
```diff
@@ -118,12 +120,12 @@ impl MetaApi for StateMachine {
...
-            let (db_name, seq_id) = r?;
+            let (db_lookup_key, seq_id) = r?;
             let seq_meta = self.get_database_meta_by_id(&seq_id.data)?;
 
             let db_info = DatabaseInfo {
                 database_id: seq_id.data,
-                db: db_name,
+                db: db_lookup_key.get_database_name(),
                 meta: seq_meta.data,
             };
```
drmingdrmer, line 130: Do we need `tenant` in `DatabaseInfo`? 🤔 
@BohuTANG

common/meta/types/src/cmd.rs
```diff
@@ -89,24 +98,30 @@ impl fmt::Display for Cmd {
             Cmd::AddNode { node_id, node } => {
                 write!(f, "add_node:{}={}", node_id, node)
             }
-            Cmd::CreateDatabase { name, meta } => {
-                write!(f, "create_db:{}={}", name, meta)
+            Cmd::CreateDatabase { tenant, name, meta } => {
+                write!(f, "create_db:{}-{}={}", tenant, name, meta)
```
drmingdrmer, line 102:
```suggestion
                write!(f, "create_db:{}/{}={}", tenant, name, meta)
```

common/meta/types/src/cmd.rs
```diff
@@ -89,24 +98,30 @@ impl fmt::Display for Cmd {
...
             }
-            Cmd::DropDatabase { name } => {
-                write!(f, "drop_db:{}", name)
+            Cmd::DropDatabase { tenant, name } => {
+                write!(f, "drop_db:{}-{}", tenant, name)
```
drmingdrmer, line 105:
```suggestion
                write!(f, "drop_db:{}/{}", tenant, name)
```

common/meta/types/src/cmd.rs
```diff
@@ -89,24 +98,30 @@ impl fmt::Display for Cmd {
...
             Cmd::CreateTable {
+                tenant,
                 db_name,
                 table_name,
                 table_meta,
             } => {
-                write!(f, "create_table:{}-{}={}", db_name, table_name, table_meta)
+                write!(
+                    f,
+                    "create_table:{}-{}-{}={}",
```
drmingdrmer, line 115:
```suggestion
                    "create_table:{}/{}-{}={}",
```

common/meta/types/src/cmd.rs
```diff
@@ -89,24 +98,30 @@ impl fmt::Display for Cmd {
...
+                    tenant, db_name, table_name, table_meta
+                )
             }
             Cmd::DropTable {
+                tenant,
                 db_name,
                 table_name,
             } => {
-                write!(f, "delete_table:{}-{}", db_name, table_name)
+                write!(f, "delete_table:{}-{}-{}", tenant, db_name, table_name)
```
drmingdrmer, line 124:
```suggestion
                write!(f, "delete_table:{}/{}-{}", tenant, db_name, table_name)
```

common/io/src/marshal.rs
ariesdevil: Yup, `char` is just `u32` in Rust

common/io/src/unmarshal.rs
ariesdevil: As above said, `char` is just `u32` but not all `u32` is a valid `char`, return space when encounter invalid char seems fine here.

common/io/src/binary_read.rs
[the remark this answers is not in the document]
drmingdrmer: Then why do we need to store it?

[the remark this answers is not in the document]
ariesdevil: Cause I use `write_string` in serializing `DatabaseLookupKey` that it first write string len

common/io/src/marshal.rs
drmingdrmer: I mean...does this hold: `char_a > char_b ↔ marshal(char_a) > marshal(char_b)`?
Because we need to list tenants in alphabetic order.

```rust
fn marshal(c: char) -> Vec<u8> {
    let mut scratch = vec![0, 0, 0, 0];

    let bits = c as u32;
    scratch[0] = bits as u8;
    scratch[1] = (bits >> 8) as u8;
    scratch[2] = (bits >> 16) as u8;
    scratch[3] = (bits >> 24) as u8;

    scratch
}

fn main() {
    let a: char = '我';
    let b: char = '发';

    let aa = marshal(a);
    let bb = marshal(b);

    println!("{} {} {:?} {:?} ", a, b, aa, bb);

    assert_eq!(a > b, aa > bb);
}

```

```
我 发 [17, 98, 0, 0] [209, 83, 0, 0] 
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`', src/main.rs:43:5
```

common/io/src/binary_read.rs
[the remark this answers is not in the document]
drmingdrmer: Then why do not you use `read_string()` to deserialize it...

common/io/src/unmarshal.rs
drmingdrmer: If it is possible to fail, then it should be better to return `Result` or `panic`.
Otherwise, it just hides a bug.

common/io/src/binary_read.rs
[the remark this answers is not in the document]
ariesdevil: We can use `read_string()` to read the whole string then we split it in deserialize function, but I prefer a `read_tenant` that I think more clearly.

ariesdevil:
PTAL again @drmingdrmer

common/io/src/unmarshal.rs
```diff
@@ -114,12 +114,12 @@ impl Unmarshal<bool> for bool {
...
-            | u32::from(scratch[2]) << 16
-            | u32::from(scratch[3]) << 24;
-        Self::from_u32(bits).unwrap_or(' ')
+impl Unmarshal<Option<char>> for Option<char> {
+    fn unmarshal(scratch: &[u8]) -> Option<char> {
+        let bits = u32::from(scratch[3])
+            | u32::from(scratch[2]) << 8
+            | u32::from(scratch[1]) << 16
+            | u32::from(scratch[0]) << 24;
+        char::from_u32(bits)
```
drmingdrmer, line 123: The trait `Unmarshal` implies `unmarshal()` has to succeed or panic.

Since normally `unmarshal()` will fail with invalid input, IMHO, `trait Unmarshal` should provide a method `try_unmarshal() -> Result<T,E>`.
And provides a default implementation for `try_unmarshal()` by calling `unmarshal()`. So that it does not affect existent implementations.

ariesdevil: lesson learned

common/io/src/binary_read.rs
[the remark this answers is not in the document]
drmingdrmer: I kinda have the same feeling that a `common/io` crate should not be aware of the concept `tenant`:
https://github.com/datafuselabs/databend/pull/3790#discussion_r780855681

If it is a method in a business logic crate, such a way makes sense.
If it is a method in `common/io`, `read_string()` would be better.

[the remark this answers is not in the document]
ariesdevil: make sense

common/meta/raft-store/src/state_machine/sm_meta_api_impl.rs
BohuTANG: No need now

common/io/src/unmarshal.rs
```diff
@@ -14,6 +14,9 @@
 
 pub trait Unmarshal<T> {
     fn unmarshal(scratch: &[u8]) -> T;
+    fn try_unmarshal(scratch: &[u8]) -> Option<T> {
+        Some(Self::unmarshal(scratch))
+    }
```
drmingdrmer, line 19: Is there any reason for using Option instead of Result?

Here using `Result<T, E>` is more convenient to express what's going wrong.

The error type could be just the ErrorCode or a type parameter with default value ErrorCode.

ariesdevil: > However, the reverse is not true: not all valid u32s are valid chars. from_u32() will return None if the input is not a valid value for a char
As I see rust use Option type for covert u32 to char, so I use Option type here.

drmingdrmer (approved):
LGTM~

PR #3798, opened 2022-01-07
ISSUE-3751: Refine cache

### LINKED ISSUE #3751: Improvement: Consider caching object(instead of bytes) in in-memory cache
For the cache in storage, it would be better if we could eliminate the de-serialization operation while reading cached item from cache

https://github.com/datafuselabs/databend/blob/a0190843ccfd1190b6e11b22e7ea08a32944d2ae/query/src/storages/fuse/io/meta_reader.rs#L27-L40

dantengsky:

## Summary

Some refinements of the table cache

- Splits in-memory cache into caches of the snapshot, segment, block metadata

   currently, all the caches are in-memory LRU Object cache.
- Introduce `CachedReader`

  Given a reader of  T, and a cache of T, while loading object, the `CachedReader` uses/populates the cache if possible/necessary.

  Table operations work with `XxxReader`s, no longer need to know the cache, excepts for `optimize` operation, where caches might be removed explicitly

NOTE: 

  1. Caches are metered by the **count** of the cached items (instead of size)
  
      For Snapshot, SegmentInfor, we could meter the cached items by their size (before de-ser), but
      unfortunately, the size of the parquet meta can not be obtained yet.  let's switch to `BytesMeter` later.
      
  2. Consequently, the metrics no longer record `read_bytes`
  
  @Veeupup Please take a look , and thanks :)

## Changelog

- Improvement

## Related Issues

Fixes #3751

## Test Plan

Unit Tests

Stateless Tests


query/tests/it/interpreters/interpreter_interceptor.rs
```diff
@@ -66,8 +66,8 @@ async fn test_interpreter_interceptor() -> Result<()> {
             "+----------+--------------+-----------+--------------+-----------+------------+-----------------+--------------+---------------+-------------+--------------+------------+------------------------------------------------------+----------+---------------------------------------------------------------------------+",
             "| log_type | handler_type | cpu_usage | memory_usage | scan_rows | scan_bytes | scan_partitions | written_rows | written_bytes | result_rows | result_bytes | query_kind | query_text                                           | sql_user | sql_user_privileges                                                       |",
```
BohuTANG, line 67: The main branch has remove the column `memory_usage` :D

dantengsky: got it, 3Q!

dantengsky: Merged with the main branch, `memory_usage` is still there ...

Veeupup:
wow! I love the idea to have a CacheManager to separate items cache and CachedReader!

query/src/sessions/sessions.rs
```diff
@@ -115,8 +96,11 @@ impl SessionManager {
         self.catalog.clone()
     }
 
-    pub fn get_storage_cache(self: &Arc<Self>) -> Arc<Option<Box<dyn StorageCache>>> {
-        self.storage_cache.clone()
+    /// get storage cache manager
+    //  we need to disable the clippy warning here, lifetime eliding leads to compile errors
+    #[allow(clippy::needless_lifetimes)]
```
Veeupup, line 101: I'm still a little confused... it seems strange to have a lifetime identifier here...

BohuTANG:
@mergify update

query/tests/it/interpreters/interpreter_interceptor.rs
BohuTANG: Sorry, I have fixed it in this  PR: https://github.com/datafuselabs/databend/pull/3809

query/src/storages/fuse/io/meta_readers.rs
```diff
@@ -0,0 +1,192 @@
...
+            }
+        })?;
+        let r = serde_json::from_slice::<V>(&buffer)?;
+        Ok(r)
+    }
+}
+
+#[async_trait::async_trait]
+impl<T> Loader<BlockMeta> for T
+where T: InputStreamProvider + Sync
```
Veeupup, line 149: Love this  abstraction！

Veeupup (approved):

dantengsky:
@BohuTANG thanks for resolving the merge conflicts!

query/src/sessions/sessions.rs
dantengsky: Thanks for pointing this out!

After staring at these lines for quite a while (at least half a min, swear to the lamp :), finally recognized that its signature DOES NOT meet the lifetime elision rule.

> pub fn get_storage_cache_manager<'a>(self: &'a **Arc<Self>**) -> &'a CacheManager {
>       .....
> }

The type of `self` should have been a simple `&self` instead. really dumb...

Issue #3740, opened 2022-01-03
Add `management-mode` for databend


BohuTANG:
## Summary

For some cases, user uses a databend-query server as a proxy and do multi-tenant operations in a same session, for example:
```
sudo use tenant  'id1';
create user xx...;
create database 'xx';
sudo use tenant  'id2';
create database 'yy';
```
Goals:
Make databend-query as a proxy, users can do some meta operations(database/table/users/stage creating) from the databend-query for different tenants.

Tasks:
- [x] Change all places where get config.query.tenant_id directly to ctx.get_tenant_id() #3760 
- [x] #3776 
- [x] `SUDO USE TENANT <tenant_id>` statement #3793 
- [x] #3759

PR #3800, opened 2022-01-07
Add the proxy-mode stateless test

BohuTANG:

## Summary

Part of #3740

1. change `USE TENANT` to `SUDO USE TENANT` since mysql driver will handle the USE statement to USE 'TENANT xx' as use database
2. Add proxy-mode stateless test
3. Refine the test suites directory structure

## Changelog

- Improvement

## Related Issues

Part of #3740

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
`sudo use tenent` is a little hacky.

How about improving it by `set tenent 'xxx' ` .

BohuTANG:
> `sudo use tenent` is a little hacky.
> 
> How about improving it by `set tenent 'xxx' ` .

sudo use tenant is a administration command and only used in proxy mode, it mainly for a client to send commands to metasrv, not a user command.
Though, we should sperate it from the user command.

BohuTANG:
This only for proxy-mode and re-struct some stateless, let's merge

Issue #3774, opened 2022-01-05
dal2: Implement s3 support


Xuanwo:
Part of https://github.com/datafuselabs/databend/issues/3677

## TODO

- [x] https://github.com/datafuselabs/databend/pull/3775
- [x] https://github.com/datafuselabs/databend/pull/3802
- [x] https://github.com/datafuselabs/databend/pull/3821
- [ ] Add integration tests upon minio

## Waiting

- SDK needs to support path style
- SDK needs to support signature v2

Xuanwo:
not sure if the signature of method `enable_path_style` is composable with the `finish` method, e.g. `builder.enable_path_style().finish()?`

_Originally posted by @dantengsky in https://github.com/datafuselabs/databend/pull/3775#discussion_r778628674_

Xuanwo:
@dantengsky, I think this style works. I will add a test case for `Builder` to make sure they works as expected.

PR #3802, opened 2022-01-07
dal2/s3: Implement unit tests

Xuanwo:

## Summary

Part of https://github.com/datafuselabs/databend/issues/3774

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/dal2/src/services/s3.rs
+++ b/common/dal2/src/services/s3.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::pin::Pin;
-use std::str::FromStr;
 use std::task::Context;
 use std::task::Poll;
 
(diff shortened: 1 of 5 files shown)


common/dal2/src/services/s3.rs
```diff
@@ -99,20 +125,60 @@ impl Builder {
...
+        // This value is scoped under `finish` function only, changed once and used
+        // once while initializing the client, so it must be safe.
+        static mut REGION: String = String::new();
+        unsafe { REGION = self.region.clone() };
+        // TODO: Maybe we can
+        //
+        // - use "default" as the default region.
+        // - use "us-east-1" as the default region.
+        // - detect the region at runtime via `ListBuckets`.
+        cfg = cfg.region(AwsS3::Region::new(unsafe { REGION.as_str() }));
```
BohuTANG, line 140: Here is a bit strange, looks like we should use some AwsS3 function to do a convert from self.region:
https://docs.rs/aws-region/0.23.3/src/awsregion/region.rs.html#137

Xuanwo: Oh, makes sense. I will upate it.

Xuanwo: I give `aws-region` a try, it's not a part of aws rust sdk, and we can't convert by this crate.

I will figure a better to handle the endpoint and region.

dantengsky: `impl<'a> From<String> for Cow<'a, str>`

https://doc.rust-lang.org/std/borrow/enum.Cow.html#impl-From%3CString%3E

Xuanwo: It works! Thanks a lot!

Xuanwo: Well, I still have question on this.

`Cow` works well for `Cow<'static, str>`, but aws `Metadata` requires a `&static str`.

```rust
#[derive(Debug)]
pub struct Metadata {
    /// URI for the endpoint.
    ///
    /// May contain `{region}` which will replaced with the region during endpoint construction
    pub uri_template: &'static str,

    /// Protocol to use for this endpoint
    pub protocol: Protocol,

    /// Credential scope to set for requests to this endpoint
    pub credential_scope: CredentialScope,

    /// Signature versions supported by this endpoint.
    ///
    /// Currently unused since the SDK only supports SigV4
    pub signature_versions: SignatureVersion,
}
```

Maybe it's better to implement a `ResolveAwsEndpoint` byself?

Xuanwo: I will refactor region and endpoint parse logic.

Xuanwo:
I refactor the logic with new `endpoint` meanings, PTAL. @dantengsky

PR #3804, opened 2022-01-07
refactor: simplify the grant verify logic

flaneur2020:

## Summary

this PR introduces a `GrantObject::contains` method to compare grant objects, with this helper function, we can replace some adhoc logics like `verify_database_privilege`, `verify_global_privilege`, `verify_table_privilege` into a unified logic: the verified object should be smaller than the object inside grant entry.

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/meta/types/src/user_grant.rs
+++ b/common/meta/types/src/user_grant.rs
@@ -104,64 +104,19 @@ impl GrantEntry {
         }
     }
 
-    pub fn verify_global_privilege(
+    pub fn verify_privilege(
         &self,
         user: &str,
         host: &str,
+        object: &GrantObject,
         privilege: UserPrivilegeType,
     ) -> bool {
         if !self.matches_user_host(user, host) {
             return false;
         }
 
-        if self.object != GrantObject::Global {
-            return false;
-        }
-
-        self.privileges.contains(privilege)
-    }
-
-    pub fn verify_database_privilege(
-        &self,
-        user: &str,
-        host: &str,
-        db: &str,
-        privilege: UserPrivilegeType,
-    ) -> bool {
-        if !self.matches_user_host(user, host) {
-            return false;
-        }
-
-        if !match &self.object {
-            GrantObject::Global => true,
-            GrantObject::Database(ref expected_db) => expected_db == db,
-            _ => false,
-        } {
-            return false;
-        }
-
-        self.privileges.contains(privilege)
-    }
-
-    pub fn verify_table_privilege(
-        &self,
-        user: &str,
-        host: &str,
-        db: &str,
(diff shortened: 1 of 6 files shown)


common/meta/types/src/user_grant.rs
```diff
@@ -28,10 +28,20 @@ pub enum GrantObject {
...
-    /// statement is meaningless for a table.
-    pub fn allow_privilege(&self, privilege: UserPrivilegeType) -> bool {
-        self.available_privileges().has_privilege(privilege)
+    /// Comparing the grant objects, the Database object contains all the Table objects inside it.
+    /// Global object contains all the Database objects.
+    pub fn contains(&self, object: &GrantObject) -> bool {
+        use GrantObject::*;
+        match (self, object) {
+            (Global, _) => true,
+            (Database(_), Global) => false,
```
BohuTANG, line 37: `GrantObject::Database` instead of `use GrantObject::*;` looks better

common/meta/types/src/user_grant.rs
```diff
@@ -43,11 +53,14 @@ impl GrantObject {
         }
     }
 
-    /// Check if there's any privilege which can not be granted to this GrantObject
-    pub fn validate_privileges(&self, privileges: UserPrivilegeSet) -> Result<()> {
+    /// Check if there's any privilege which can not be granted to this GrantObject.
+    /// Some global privileges can not be granted to a database or table, for example,
+    /// a KILL statement is meaningless for a table.
+    pub fn validate_available_privileges(&self, privileges: UserPrivilegeSet) -> Result<()> {
```
BohuTANG, line 59: validate_available_privileges -> validate_grant_privileges?

common/meta/types/src/user_grant.rs
flaneur2020: thank you, fixed

common/meta/types/src/user_grant.rs
flaneur2020: good idea, btw i think this function can be moved to interpreter_grant.rs, because it's the only call point, and GrantEntry is a low level object, do not really knows about the GRANT statement

BohuTANG: Ok, I guess many access checking can be add to interpreter, for access checking methods,  we already have a 'access' fold there

flaneur2020: oh, this function is not an access check but an argument validation.

each GrantObject has a set of available privileges, like a privilege about KILL is meaning less to a Table but available to Global, this function does this check.

PR #3805, opened 2022-01-07
support `[NOT] IN (a, b, c, ...)` InList SQL

Veeupup:

## Summary

support `[NOT] IN (a, b, c, ...)` SQL

## Changelog

- New Feature

## Related Issues

Fixes #

## Test Plan

Unit Tests

Stateless Tests


Veeupup:
@sundy-li  @zhang2014  Although `DataType::Struct` is not finished yet. But for now, maybe u can give me some advice. : D

common/functions/src/scalars/udfs/in_basic.rs
```diff
@@ -0,0 +1,258 @@
...
+impl<const NEGATED: bool> InFunction<NEGATED> {
+    pub fn try_create(_display_name: &str) -> Result<Box<dyn Function>> {
+        Ok(Box::new(InFunction::<NEGATED> {}))
+    }
+
+    pub fn desc() -> FunctionDescription {
+        FunctionDescription::creator(Box::new(Self::try_create)).features(
+            FunctionFeatures::default()
+                .bool_function()
+                .variadic_arguments(2, usize::MAX),
```
zhyass, line 44: Maybe we can set `variadic_arguments` to 2? The list as an element.

common/functions/src/scalars/udfs/in_basic.rs
```diff
@@ -0,0 +1,258 @@
...
+    fn eval(&self, columns: &DataColumnsWithField, _input_rows: usize) -> Result<DataColumn> {
+        let input_column = columns[0].column();
+
+        let input_array = match input_column {
+            DataColumn::Array(array) => array.to_owned(),
+            DataColumn::Constant(scalar, _) => scalar.to_array()?,
+        };
+
+        let input_dt = input_array.data_type();
+        if input_dt == &DataType::Null {
```
zhyass, line 143: I think we cannot cast with the `input_dt` directly. 
Consider the following example: `select 1 in (1.2, 2 )`, If `input_dt` is `DataType::UInt8`,  `1.2` will be cast to `1`, then the result will be `true`.

common/functions/src/scalars/udfs/in_basic.rs
Veeupup: sorry, I did not get your point, what do you mean by using struct?

common/functions/src/scalars/udfs/in_basic.rs
Veeupup: thanks for reminding me, it should not do cast directly here.

common/functions/src/scalars/udfs/in_basic.rs
zhyass: For example, `a in (1,2,3)`, `(1,2,3)` can be treated as `DataType::Struct` or `DataType::List`.

Veeupup: Get it. 
Hereby using more columns here because we can simply reuse `Expression::ScalarFunction` to simplify parser work here... : ), so the parser will work just as simple

zhyass: The `vec<Expr>` can be parsed to `Tuple` function.

common/functions/src/scalars/udfs/in_basic.rs
```diff
@@ -0,0 +1,300 @@
...
+            let val = array.try_get(0)?;
+            match val {
+                DataValue::$CAST_TYPE(Some(val)) => {
+                    vals_set.insert(val);
+                }
+                _ => {}
+            }
+        }
+        for idx in 0..$INPUT_ARRAY.len() {
+            let val = $INPUT_ARRAY.try_get(idx)?;
```
sundy-li, line 64: `try_get()` is very slow.

Can we avoid using datavalues ?

zhang2014:
> @sundy-li @zhang2014 Although DataType::Struct is not finished yet. But for now, maybe u can give me some advice. : D

Look good to me.

common/functions/src/scalars/udfs/in_basic.rs
```diff
@@ -0,0 +1,362 @@
...
+            }
+            return Ok(DataColumn::Array(array.as_series()));
+        }
+        let mut builder = MutableBooleanArrayBuilder::<false>::with_capacity(input_column.len());
+
+        let check_arrays = &columns[1..];
+
+        let mut least_super_dt = input_dt.clone();
+        for array in check_arrays {
+            least_super_dt = merge_types(&least_super_dt, array.data_type())?;
```
zhyass, line 207: It's better to use `aggregate_types`.

common/functions/src/scalars/udfs/in_basic.rs
```diff
@@ -0,0 +1,362 @@
...
+                false => NEGATED,
+            };
+            $BUILDER.push(v);
+        }
+    };
+    ($INPUT_DT: expr, $INPUT_ARRAY: expr, $CHECK_ARRAY: expr, $NEGATED: expr, $BUILDER: expr, $CAST_TYPE: ident, $PRIMITIVE_TYPE: ident) => {
+        let mut vals_set = HashSet::new();
+        for array in $CHECK_ARRAY {
+            let array = array.column().cast_with_type($INPUT_DT)?;
+            match array {
```
sundy-li, line 115: Must be constant!

common/functions/src/scalars/udfs/in_basic.rs
```diff
@@ -0,0 +1,362 @@
...
+                DataColumn::Constant(DataValue::$CAST_TYPE(None), _) => {
+                    continue;
+                }
+                _ => {
+                    return Err(ErrorCode::LogicalError("it's a bug"));
+                }
+            }
+        }
+        let arr = $INPUT_ARRAY.get_array_ref();
+        let arr = DFBooleanArray::from_arrow_array(arr.as_ref());
```
sundy-li, line 102: .bool()?

common/functions/src/scalars/udfs/in_basic.rs
```diff
@@ -0,0 +1,362 @@
...
+                }
+                _ => {
+                    return Err(ErrorCode::LogicalError("it's a bug"));
+                }
+            }
+        }
+        let arr = $INPUT_ARRAY.get_array_ref();
+        let arr = DFPrimitiveArray::<$PRIMITIVE_TYPE>::from_arrow_array(arr.as_ref());
+        for val in arr.into_no_null_iter() {
+            let v = match vals_set.contains(val) {
```
sundy-li, line 161: `a & b`

sundy-li: `v` && `!NEGATED`

sundy-li: ( v && !NEGATED ) || (!v && NEGATIVED)

common/functions/src/scalars/udfs/in_basic.rs
Veeupup: resolved

zhyass (approved):

PR #3813, opened 2022-01-09
[docs] add a tip for console bind

### LINKED ISSUE #3758: configure different console ports for query and meta
**update 2021-01-07**

I think just modifying the documentation will be enough. There is no need to add more complexity to the configuration for now.

---

After #3739 , we can explore and diagnose with tokio-console.

> Note that diagnostics are currently only supported for a single program, so please ensure that **only one** program has a log level of `TRACE`. Otherwise, only the first program to occupy the port will be monitored.

Consider configuring different ports for console-subscriber. Maybe it looks like this?

```rust
init_global_tracing(app_name: &str, dir: &str, level: &str, <port in config>);

let console_layer = ConsoleLayer::builder()
    // set the address the server is bound to
    .server_addr(([127, 0, 0, 1], <port in config>))

let subscriber = subscriber.with(console_layer);
```

Of course, new fields will also need to be added to the config.

PsiACE:

## Summary

using `TOKIO_CONSOLE_BIND` to assign different binds

## Changelog

- Documentation

## Related Issues

Fixes #3758 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/website/databend/docs/dev/development/tracing.md
+++ b/website/databend/docs/dev/development/tracing.md
@@ -190,7 +190,7 @@ Open http://127.0.0.1:16686/
 1. Compile with specific `RUSTFLAGS` and features. We can use `--bin` to specify binary.
 
    ```shell
-   RUSTFLAGS="--cfg tokio_unstable" mold --run cargo build --features tokio-console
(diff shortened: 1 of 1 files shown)


BohuTANG:
/lgtm

:+1:

PR #3830, opened 2022-01-11
ISSUE-3828 MySQL Driver Compalibility

### LINKED ISSUE #3828: A driver compatibility question
## Summary

If someone use [mysql python driver](https://github.com/mysql/mysql-connector-python) connect databend will have some errors:

1. mysql.connector.errors.InterfaceError: MySQL Version '0.1.0-d757471-simd(1.59.0-nightly-2022-01-10T04:17:51.060993521+00:00)' is not supported.
2. mysql.connector.errors.DatabaseError: 1105 (HY000): Code: 5, displayText = sql parser error: Expected equals sign or TO, found: 'utf8'.
3. mysql.connector.errors.DatabaseError: 1105 (HY000): Code: 5, displayText = sql parser error: Expected equals sign or TO, found: ..

For the first:

The driver is created in 2014, so the driver does not support MySQL if the Version is less than 4.1. This is a case-by-case design.

```python
elif version < (4 1):
    print("Err the version is ", version)
    raise errors.InterfaceError(
        "MySQL Version '{0}' is not supported.".format(server_version))
```

For the second:

The driver will execute query `SET NAMES utf8 COLLATE utf8_general_ci` and databend not support.

For the third:

The driver will execute query `SET @@session.autocommit = 1` and databend not support.

## How to fix

For the first:

Modify databend version format as year.x.x. Maybe like this `22.0.1`

For the second:

Modify databend parser, if the query has `SET NAME` and `SET @@` return ok.

TCeason:

## Summary

[MySQL Python connector](https://github.com/mysql/mysql-connector-python) will send some queries that databend not supported now,

so the driver will not connect the databend.

* some test not fmt, modify them.
    tests/databend-test
    tests/helpers/client.py
    tests/suites/0_stateless/00_0000_dummy_select_py.py

## Changelog

- Bug Fix

## Related Issues

Fixes #3828 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/Cargo.toml
+++ b/query/Cargo.toml
@@ -95,6 +95,7 @@ tonic = "0.6.2"
 uuid = { version = "1.0.0-alpha.1", features = ["serde", "v4"] }
 walkdir = "2.3.2"
 parquet-format-async-temp = "0.2.0"
+regex = "1.5.4"
 
 [dev-dependencies]
 clickhouse-driver = { git = "https://github.com/datafuse-extras/clickhouse_driver", rev = "6e0ecb2" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2573,6 +2573,7 @@ dependencies = [
  "pretty_assertions",
  "prost",
  "rand 0.8.4",
+ "regex",
  "reqwest",
  "serde",
  "serde_json",
--- a/query/src/servers/mysql/mysql_interactive_worker.rs
+++ b/query/src/servers/mysql/mysql_interactive_worker.rs
@@ -31,6 +31,7 @@ use msql_srv::ParamParser;
 use msql_srv::QueryResultWriter;
 use msql_srv::StatementMetaWriter;
 use rand::RngCore;
+use regex::RegexSet;
 use tokio_stream::StreamExt;
 
 use crate::interpreters::InterpreterFactory;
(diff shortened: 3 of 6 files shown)


query/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "databend-query"
-version = "0.1.0"
+version = "22.1.0"
```
PsiACE, line 3: It is better to have a version specification, and then all crates in the repo should follow

TCeason: This is just a try. In my mind:

> For the first:
> 
> Modify databend version format as year.x.x. Maybe like this 22.0.1
> 
> Or
> 
> year.month.other

BohuTANG: It looks good to add a version prefix(like '22.') to `get_fuse_version`

sundy-li (commented):
LGTM, better to  add a test in `databend-test` using this MySQL client to cover this issue.

TCeason:
> LGTM, better to add a test in `databend-test` using this MySQL client to cover this issue.

Yep I will add the test.

BohuTANG:
Let's merge first.
Please add the test in later PR if have the plan, @TCeason

Issue #3476, opened 2021-12-15
Support auth plugin


BohuTANG:
**Summary**

- [x] #3480 
- [x] #3482

flaneur2020:
jwt might be a promising authentication option, like: https://trino.io/docs/current/security/jwt.html

youngsofun:
better way for rust is to use enum？

enum {
   Password（type PasswordType，password vec）
   Ldap（config string）
}

BohuTANG:
Ok, enum looks make sense.

youngsofun:
ok, it will do it

flaneur2020:
@youngsofun can take this issue assigned to yourself by /assignme  :)

youngsofun:
plan with 3 pr:

- [ ] refactor auth with enum (user table, create/alter users parser and tests) 
- [ ] continue to refactor auth framework and support simple JWT and HTTP basic(use existing password)
- [ ] store password as hashed value as it should be

Xuanwo:
All work has been done. Let's close.

PR #3849, opened 2022-01-13
Refactor authentication.

youngsofun:

## Summary

1.  introduce Enum Authtype to support multi auth types better,
     1. especially the parser parts
     2. the validation interface is changed a little, needs further refactoring in the next pr. 
2. fix MySQL auth algorithm.

## Changelog

## Related Issues

https://github.com/datafuselabs/databend/issues/3476#issue-1080763624

Fixes #issue

## Test Plan

Unit Tests


common/meta/types/src/user_auth.rs
```diff
@@ -11,17 +11,237 @@
...
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
+pub enum AuthType {
+    NoPassword,
+    PlaintextPassword,
+    Sha256Password,
+    DoubleShaPassword,
+}
+
+use AuthType::*;
```
BohuTANG, line 29: Why `use AuthType::*;`?

common/meta/types/src/user_auth.rs
```diff
@@ -11,17 +11,237 @@
...
+                password: p,
+                password_type: _t,
+                // TODO(youngsofun): we should store password hash later
+            } => Ok(p == password_input_plaintext),
+        }
+    }
+}
+
+#[derive(Clone, serde::Serialize, serde::Deserialize, Debug, PartialEq)]
+pub struct AuthInfoRaw {
```
BohuTANG, line 191: `AuthInfoRaw` confused here with `AuthInfo`, maybe we need a better name :thinking:

youngsofun: AuthInfoRaw means it is directly mapped to SQL.
usage of it depends on the AuthType.
I choose to not make further abstraction to make it clear to the implementer of AuthType.
what about AuthInfoArgs?

common/meta/types/src/user_auth.rs
youngsofun: fixed

common/management/src/user/user_mgr.rs
```diff
@@ -150,22 +150,20 @@ impl UserMgrApi for UserMgr {
         &self,
         username: String,
         hostname: String,
-        new_password: Option<Vec<u8>>,
-        new_password_type: Option<PasswordType>,
+        auth_info_raw: AuthInfoArgs,
```
BohuTANG, line 153: auth_info_args?

common/planners/src/plan_user_alter.rs
```diff
@@ -16,15 +16,15 @@ use std::sync::Arc;
...
 
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq)]
 pub struct AlterUserPlan {
     pub if_current_user: bool,
     pub name: String,
     pub hostname: String,
-    pub new_password: Vec<u8>,
-    pub new_password_type: PasswordType,
+    // None means no change to make
+    pub auth_info_raw: Option<AuthInfoArgs>,
```
BohuTANG, line 27: auth_info_raw -> auth_info_args?

query/src/sql/sql_parser.rs
```diff
@@ -692,14 +694,18 @@ impl<'a> DfParser<'a> {
             String::from("%")
         };
 
-        let (password_type, password) = self.get_auth_option()?;
+        let auth_info_raw = self.get_auth_option_raw()?;
```
BohuTANG, line 697: Change from raw to args?

common/meta/types/src/user_auth.rs
BohuTANG: `*_raw` -> `*_args` also need changed

common/planners/src/plan_user_alter.rs
youngsofun: fixed, thank you

query/src/sql/sql_parser.rs
```diff
@@ -694,7 +694,7 @@ impl<'a> DfParser<'a> {
             String::from("%")
         };
 
-        let auth_info_args = self.get_auth_option_raw()?;
+        let auth_info_args = self.get_auth_info_args()?;
```
youngsofun, line 697: forgot this one, no "raw" any more

query/src/servers/clickhouse/interactive_worker.rs
```diff
@@ -113,7 +110,7 @@ impl ClickHouseSession for InteractiveWorker {
                 .await
             {
                 Ok(user_info) => (
-                    user_manager.auth_user(user_info.clone(), info).await,
+                    user_info.auth_info.auth_password_plaintext(password),
```
BohuTANG, line 113: if the password is sha256, `auth_password_plaintext` does not work?

BohuTANG: Looks we need a `auth_info.auth(type:ClickHouse/MySQL, Option<salt>, password)` to do all the handlers type auth?
clickhouse also has the SHA256 password type.

Xuanwo:
Hi, please attach the issue within the PR's description, so others can take participate too~

common/meta/types/src/user_auth.rs
```diff
@@ -11,17 +11,237 @@
...
+        match self {
+            AuthInfo::Password {
+                password: _,
+                password_type: t,
+            } => Some(*t),
+            _ => None,
+        }
+    }
+
+    fn decode_double_sha1(
```
Xuanwo, line 128: `decode_double_sha1` is not a good name to me, it's confusing. How could we `decode` via hash? But I can't come up better names for it too... :cry: 

Do have any ideas?

query/src/servers/clickhouse/interactive_worker.rs
youngsofun: both clickhouse and MySQL support it，and store sha2hashed value，but the protocol parts are different.

youngsofun: it seems clickhouse client transport password in plaintext directly without salt，so this interface can be used by ch and http basic

youngsofun: only mysql client implement the protocol， others just share the password stored

common/meta/types/src/user_auth.rs
```diff
@@ -11,17 +11,237 @@
...
 // limitations under the License.
+use common_exception::ErrorCode;
+
+const NO_PASSWORD_STR: &str = "no_password";
+const PLAINTEXT_PASSWORD_STR: &str = "plaintext_password";
+const SHA256_PASSWORD_STR: &str = "sha256_password";
+const DOUBLE_SHA1_PASSWORD_STR: &str = "double_sha1_password";
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
+pub enum AuthType {
```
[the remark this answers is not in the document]
youngsofun: the former corresponding to the with part of the sql.

the later means how the password is stored.

common/meta/types/src/user_auth.rs
```diff
@@ -11,17 +11,237 @@
...
+            _ => None,
+        }
+    }
+}
 
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum AuthInfo {
+    None,
+    Password {
+        password: Vec<u8>,
```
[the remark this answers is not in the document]
BohuTANG: That's right, like the idea.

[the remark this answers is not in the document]
youngsofun: agree，I planned to change the name next pr when we do store the hash（digging some dtails ）

common/meta/types/src/user_auth.rs
youngsofun: restore_sha1_mysql？

common/meta/types/src/user_auth.rs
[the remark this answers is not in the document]
flaneur2020: +1, i guess it'd better to store the auth string directly in the AuthInfo enum.

it allows AuthInfo sharing the common infomation across the different layers.

common/meta/types/src/user_auth.rs
```diff
@@ -11,17 +11,237 @@
...
+            } => Ok(p == password_input_plaintext),
+        }
+    }
+}
+
+#[derive(Clone, serde::Serialize, serde::Deserialize, Debug, PartialEq)]
+pub struct AuthInfoArgs {
+    pub arg_with: Option<AuthType>,
+    pub arg_by: Option<String>,
+}
```
flaneur2020, line 192: can we merge AuthInfoArgs and AuthInfo into a unified struct? q.q

i think we can dispatch the auth methods by name in the sql parser.

query/src/servers/clickhouse/interactive_worker.rs
flaneur2020: > Looks we need a auth_info.auth(type:ClickHouse/MySQL, Option<salt>, password) to do all the handlers type auth?

+1, a unified `auth()` would be easier to reason about? q.q

common/meta/types/src/user_auth.rs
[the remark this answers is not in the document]
Xuanwo: Yep, I like the idea from https://multiformats.io/multihash/ which represent a hash in both human-readable and machine-native way:

image

Here is a rust implementation: https://github.com/multiformats/rust-multihash

common/meta/types/src/user_auth.rs
youngsofun: I tried it at first but did not find a good way.


 there are tiny differences between "create user" and "alter user":

create user:  
- users do not need to provide an arg as long as **it has a reasonable default value**

alter user:
- user users do not need to provide an arg as long as i**t makes sense to use the current value (just change the supplied parts)** 
- it has more situations
    - diff auth type may use diff args like as, by, server in click house 
   - a SQL may be valid by itself (from parser view), but invalid combined with the current auth type. 

 so for an AuthInfo::XXX, we may need a type AuthInfoArgs::XXX or r AuthInfo::XXX::Args, I think it is not necessary for now.


## this pr:
 
AuthInfo is stored in meta, as the final records. Most Fields in AuthInfo are not optional. 

create user:  
- parser:  SQL -> AuthInfoArgs -> AuthInfo
- analyzer use AuthInfo

alter user:
- parser:  SQL -> AuthInfoArgs(check grama of the SQL itself)
- analyzer make use of AuthInfoArgs according to current AuthInfo (may check futher)

query/src/servers/clickhouse/interactive_worker.rs
youngsofun: @flaneur2020  thanks for review

double_sha by its name only means how the password is stored. if the password is safely delivered, we valid it by simply hashing the input again, that is what the clickhouse-client did, and we will deal with it the same way for http-basic. 

but for MySQL client/diver it also means an auth protocol, the validation method is totally different too.

for stored password auth only, I`d rather sep them apart.

for the general auth interface, This pr is not the final interface,  it needs further refactor when external auth is introduced. we need at least an auth manager object because each type of external auth may have multi instances with diff configs to load, and the unified auth entrance should be here. I haven't quite nailed that down yet, but I think the usage of the enum AuthInfo is a good start,  which is the main point for the pr.

common/meta/types/src/user_auth.rs
BohuTANG: So, it seems we don't need a AuthInfoArg in meta here, it only has two fields, it would better to use function argument or others do that.
I am still confused on AuthInfo and AuthInfoArgs 🤔

youngsofun: @BohuTANG. 

this is for extensions later.
 
currently,  it only has two fields now because we only implement password auth now. 
later,  diff auth type will use many diff arg names,  It may looks like 


```
pub struct AuthInfoArgs {
    pub arg_with: Option<AuthType>,
    pub arg_by: Option<String>,
    pub arg_as: Option<String>,   // for mysql style ladp
    pub arg_server: Option<String>, // for clickhouse  style ladp
    pub arg_realm: Option<String>,   // for cickhouase style kerberos
  ...
}
```


we can all use "BY":
some may need more than one args.
some may need other types like int.  
even for passwords, we may later add something like password_expired (which most auth types do not need)

with diff names and types, is easier for user to understand and for parser to check

----

passing them by function argument will make the argument list of all related functions too long and with many options fields. 

---

and in "alter user",  need to check it in 2 places:

```
create user yang IDENTIFIED WITH ldap SERVER 'server_name'
alter user yang IDENTIFIED SERVER 'server_name' //  ok
alter user yang IDENTIFIED by 'new_password' //  ok to the parser,  fail for interpreter 
```

----

with AuthInfoArgs,  the info in SQL info is collected to a struct,   ease the reading,checking and passing around.

 it can be seen as a direct result struct (AST) of a general parser (like antlr),  which is natural  and is not an extra work.

BohuTANG: Ok, thank you for explaining.

youngsofun:
rebased to resolve conflicts

youngsofun:
```
error: written amount is not handled
  --> query/src/servers/mysql/reject_connection.rs:41:9
   |
41 |         stream.write(&buffer).await?;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[deny(clippy::unused_io_amount)]` on by default
   = help: use `AsyncWriteExt::write_all` instead, or handle partial writes
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_io_amount

```

youngsofun:
fixed according to the suggestions in error messages.

common/meta/types/src/user_auth.rs
flaneur2020: > later, diff auth type will use many diff arg names, It may looks like
>
> ```
> pub struct AuthInfoArgs {
>    pub arg_with: Option<AuthType>,
>    pub arg_by: Option<String>,
>    pub arg_as: Option<String>,   // for mysql style ladp
>    pub arg_server: Option<String>, // for clickhouse  style ladp
>    pub arg_realm: Option<String>,   // for cickhouase style kerberos
>  ...
> }
> ```

i'm afraid this AuthInfoArgs may contains too many details on different protocols in one struct, the fields might overlap on different scienaries. maybe a enum is more suitable on repsenting different things?

or as @Xuanwo 's suggest https://github.com/multiformats/rust-multihash , is it possible to represent multiple auth data in one auth string? q.q

flaneur2020: > AuthInfoArgs is only an intermediate structure, not stored.

i think AuthInfoArgs is an struct mostly used in the ast layer, and it'd converted to AuthInfo in the interpreter or planner layer, and stores inside the metastore.

the `meta` mod mostly stores some seriaklizable objects in the metastore, how about move the AuthInfoArgs to the `ast` mod?

youngsofun: > it possible to represent multiple auth data in one auth string? q.q

1. rust-multihash is good but only useful for password

> we can not all use "BY":
some may need more than one args.
some may need other types like int.
even for passwords, we may later add something like password_expired (which most auth types do not need)
with diff names and types, is easier for user to understand and for parser to check

youngsofun: AuthInfoArgs  

> maybe a enum is more suitable on repsenting different things

I tried at first,  one problem is  "with" can be omitted,  so need to guess enum type from only "by xxx"
currently not sure "by" is only one auth type, depends on how we design the SQL.

BohuTANG:
Temporarily converted this pr to DRAFT, @youngsofun feel free to set it READY back when it's ready for review.

youngsofun:
@youngsofun ok

youngsofun:
@BohuTANG ok

youngsofun:
@BohuTANG @flaneur2020 

sql syntax use mysql style, with general (auth_type, auth_string).  The main checking logic move to analyze().  Including: 

1. valid auth_type names: exists and enabled.  ( except special "no_password")
2. whether "by" is needed (according to auth_type)
3. For "alter user",  check with the current authinfo


Others: 

1.  Treat "not  indentified"  is the same with "indentified with no_password"
2. `indentified by ` is not valid,  but `indentified by ""`  is allowed (tried with mysql) 
3. About the DfAuthOpition struct
   1. the name "auth_options" comes from mysql
   2. reason for using it :
      1. a standalone parse unit
      2. may change (may add fields like auth_config)  later
      3. shared between DfCreateUser and DfAlterUser
      4. Define some common function on it
   3. Used only  before analyze and inside query crate

youngsofun:
still not know whether we should use the word auth_type or auth_plugin?

flaneur2020:
> still not know whether we should use the word auth_type or auth_plugin?

i guess auth_type is feasible in this phase, we do not really have plugin mechanism like mysql did, we can switch to auth_plugin after we really have some pluggable auth methods?

flaneur2020:
nice abstractions, LGTM

youngsofun:
> i guess auth_type is feasible in this phase, we do not really have plugin mechanism like mysql did, we can switch to auth_plugin after we really have some pluggable auth methods?

agree,   "valid auth_type names" should be changed to check earlier or it is ok to keep the way it is now

common/management/tests/it/user.rs
```diff
@@ -470,29 +471,44 @@ mod drop {
...
+
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-    async fn test_update_user_normal_partial_update() -> common_exception::Result<()> {
+    async fn test_update_user_normal_update_full() -> common_exception::Result<()> {
+        test_update_user_normal(true).await
+    }
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn test_update_user_normal_update_partial() -> common_exception::Result<()> {
+        test_update_user_normal(false).await
+    }
```
BohuTANG, line 497: Code style hints: functions should add new line at the end block  :)

youngsofun:
```
error: could not compile `common-functions 
Caused by:
  process didn't exit successfully: `rustc --crate-name c
....
(signal: 9, SIGKILL: kill)
....

```

ariesdevil:
> error: could not compile `common-functions 
> Caused by:
>   process didn't exit successfully: `rustc --crate-name c
> ....
> (signal: 9, SIGKILL: kill)
> ....

ping @Xuanwo

youngsofun:
@BohuTANG  ci passed this time,  Can this PR merge?

Issue #774, opened 2021-06-08
[Store]  Complete ActionHandler test for: create_db() and get_db()


drmingdrmer:
**Summary**

- [x] test create_db
- [ ] test drop_db
- [ ] test create_table
- [ ] test drop_table
- [ ] test get_table
- [x] add get_db and test.

There should be a unit test for `ActionHandler::create_db()`:

https://github.com/datafuselabs/datafuse/blob/fbe6bd8d3f761736ca03a1c932753320cd54e1b8/fusestore/store/src/executor/action_handler.rs#L100-L122

A test should start up an `ActionHandler` with a `Dfs` as backend, and assert expected behavior.
Reference:

https://github.com/datafuselabs/datafuse/blob/fbe6bd8d3f761736ca03a1c932753320cd54e1b8/fusestore/store/src/executor/action_handler_test.rs#L24-L52

drmingdrmer:
@ariesdevil I believe your last PR covers some of these task. May you get this issue updated or just close it?

PR #3868, opened 2022-01-16
[deps] bump arrow2 to d14ae86

PsiACE:

## Summary

- [x] d14ae86
- [x] fix stateless tests

## TODO

- [ ] 9b54146
- [ ] ?

## Changelog

- Improvement
- Build/Testing/CI

## Related Issues

Related #3746 

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
```
--- /Users/runner/work/databend/databend/tests/suites/0_stateless/02_0045_function_in.result	2022-01-16 15:57:58.000000000 +0000
+++ /Users/runner/work/databend/databend/tests/suites/0_stateless/02_0045_function_in.stdout	2022-01-16 16:01:17.000000000 +0000
@@ -1,17 +1,3 @@
-1
-2
-1
-2
-19
-4999
-0
-3
-4
-5
-6
-7
-8
-9
 NULL
 1
 0

```

Flaky test of  function `in`, cc @Veeupup

PsiACE:
> Flaky test of function `in`

Other tests also fail, perhaps because I remove the `DataSchemaRef`

sundy-li:
Some tests are failed. 

@jorgecarleitao Do you know what's the big changes if we bump arrow2 into d14ae86 (https://github.com/jorgecarleitao/arrow2/pull/713/files) ?

Errors:
```
{"v":0,"name":"databend-query-test_cluster@0.0.0.0:3307","msg":"panicked at 'called `Result::unwrap()` on an `Err` value: OutOfSpec(\"Unable to get head as schema\")', query/src/api/rpc/flight_client_stream.rs:56:79","level":50,"hostname":"arch","pid":31036,"time":"2022-01-17T09:33:22.311028756Z","target":"common_tracing::panic_hook","line":24,"file":"common/tracing/src/panic_hook.rs","panic.column":79,"panic.line":56,"panic.file":"query/src/api/rpc/flight_client_stream.rs"}
```

query/src/api/rpc/flight_service_stream.rs
```diff
@@ -49,7 +49,7 @@ impl Stream for FlightDataStream {
             Some(Ok(block)) => match block.try_into() {
                 Err(error) => Some(Err(Status::from(error))),
                 Ok(record_batch) => {
-                    let (dicts, values) = serialize_batch(&record_batch, &self.options);
+                    let (dicts, values) = serialize_batch(&record_batch, &[], &self.options);
```
jorgecarleitao, line 52:
```suggestion
                    let ipc_fields = arrow2::io::ipc::write::default_ipc_fields(record_batch.schema());
                    let (dicts, values) = serialize_batch(&record_batch, &ipc_fields, &self.options);
```

I have PRed #774 to allow passing `None` to this field to avoid users having to take this default themselves.

jorgecarleitao:
> @jorgecarleitao Do you know what's the big change in flight if we bump arrow2 into d14ae86 ([jorgecarleitao/arrow2#713](https://github.com/jorgecarleitao/arrow2/pull/713)) ?

Thus far, `Field` had a `dict_id`. However, `dict_id` is IPC-specific. To reuse dictionaries over the IPC, users to be careful in writing the `dict_id`. The old API did not allow this very well, as `dict_id` was just set to some default value.
For this reason, I have separated the `dict_id` from `Field` into its own little struct (`IPCField`), so that we can think about arrow's logical `Field` without having to think about how it is represented in IPC, and also allow changing it without having to modify `Field`, as it usually indicates a change in the logical construct.

query/src/api/rpc/flight_service_stream.rs
PsiACE: emmm, thanks @jorgecarleitao , I passed `&record_batch.schema().fields` as you suggested and it doesn't seem to work properly and still reports errors in the cluster test.

```rust
                    let record_batch: RecordBatch = record_batch;
                    let ipc_fields = common_arrow::arrow::io::ipc::write::default_ipc_fields(
                        &record_batch.schema().fields,
                    );
                    let (dicts, values) =
                        serialize_batch(&record_batch, &ipc_fields, &self.options);
```

query/src/api/rpc/flight_client_stream.rs
```diff
@@ -52,11 +52,17 @@ impl FlightDataStream {
                         )
                     }
 
-                    let arrow_schema = Arc::new(schema.to_arrow());
-                    Ok(
-                        deserialize_batch(&flight_data, arrow_schema, true, &Default::default())
-                            .map(create_data_block)?,
+                    let (arrow_schema, ipc_schema) =
+                        deserialize_schemas(flight_data.data_body.as_slice()).unwrap();
+                    let arrow_schema = Arc::new(arrow_schema);
```
jorgecarleitao, line 57: Ahh, here.

An arrow stream is expected to contain the following:

`[schema][batch][batch]...[batch][batch]`

where batches are expected to be deserializable by `schema`.

I.e. we need to read the schema from the stream once (first message), and then read the multiple batches from it. I think that this is currently reading the stream as `[schema][batch][schema][batch]...[schema][batch][schema][batch]`

So, 

* read first message and store `schema` and `ipc_fields` (part of the [StreamMetadata](https://github.com/jorgecarleitao/arrow2/blob/main/src/io/ipc/read/stream.rs#L19))
* read next messages (batch messages) using `schema` and `ipc_fields` to deserialize them

PsiACE: I fixed it, thank you very much for your help

PsiACE:
PTAL @sundy-li , the rest will be done in other PRs

PR #3870, opened 2022-01-16
Move `planners` to query module

lianghanzhen:

## Summary

Move `planners` to query module

## Changelog

- Other 

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


Xuanwo:
Hi, thank you very much for your contribution, could you please add some details related to this PR? For example, an RFC or an associated issue, which would help us to provide targeted review comments.

lianghanzhen:
Sorry, I don't create an issue before starting.

I think `planners` should be only available in the compute layer, what do you think? @Xuanwo @BohuTANG

sundy-li:
What if we need the planner in the storage layer if we push down some planners?

BohuTANG:
Thanks for the work.
planners may used by other crates, like storage(maybe Dal in the common) in future as  sundy said above.
In databend, we preferred sperate them into crates rather than take them in the query folder.

lianghanzhen:
That makes sense to me. Sorry I should have created an issue.

PR #3878, opened 2022-01-17
Manage role storage in metasrv

### LINKED ISSUE #3877: [RBAC] add client to manage role data
tracking issue: https://github.com/datafuselabs/databend/issues/2793

flaneur2020:

## Summary

this PR adds the storage about RoleInfo

## Changelog

- Improvement

## Related Issues

Fixes #3877

## Test Plan

Unit Tests

Stateless Tests


common/management/src/role/role_mgr.rs
```diff
@@ -0,0 +1,199 @@
...
+                role_name
+            ))),
+        }
+    }
+
+    async fn get_roles(&self) -> Result<Vec<SeqV<RoleInfo>>> {
+        let role_prefix = self.role_prefix.clone();
+        let kv_api = self.kv_api.clone();
+        let prefix_list_kv = async move { kv_api.prefix_list_kv(role_prefix.as_str()).await };
+        let values = prefix_list_kv.await?;
```
drmingdrmer, line 129: Why put it into an `async move` block?

`let values = kv_api.prefix_list_kv(role_prefix.as_str()).await?`
Is this just enough?

flaneur2020: i also have this question, it seems all the methods in this mod are in this style...  https://github.com/datafuselabs/databend/blob/9ad3602fb5905a170feec018bc1ba013eee7b935/common/management/src/user/user_mgr.rs#L135

is there any consideration about this? 😳

drmingdrmer: If it compiles without the `async move` block, there won't be a problem IMO... 🤔

common/management/src/lib.rs
```diff
@@ -13,12 +13,15 @@
...
 
 mod cluster;
+mod role;
 mod stage;
 mod udf;
 mod user;
 
 pub use cluster::ClusterApi;
 pub use cluster::ClusterMgr;
+pub use role::role_api::RoleMgrApi;
```
BohuTANG, line 23: pub use role::RoleMgrApi; -- short use is always better

common/management/src/role/role_mgr.rs
flaneur2020: thank you, lemme fix it

common/management/src/lib.rs
flaneur2020: thank you, fixed

BohuTANG (approved):
image

common/management/src/role/role_mgr.rs
```diff
@@ -0,0 +1,191 @@
...
+use common_meta_types::UserPrivilegeSet;
+
+use crate::role::role_api::RoleMgrApi;
+
+static ROLE_API_KEY_PREFIX: &str = "__fd_roles";
+
+pub struct RoleMgr {
+    kv_api: Arc<dyn KVApi>,
+    role_prefix: String,
+}
```
drmingdrmer, line 39: There is an issue with such a impl that @ariesdevil has been working on.

`kv_api` need to renew the handshake info if the backend metasrv restarted.
Because handshake info is not persistent on metasrv.

`kv_api`(and `meta_api`) has to discover a handshake error then resend the handshake request.

@ariesdevil will soon provide a solution to this problem.

common/management/src/role/role_mgr.rs
```diff
@@ -0,0 +1,191 @@
...
+    async fn upsert_role_info(
+        &self,
+        role_info: &RoleInfo,
+        seq: Option<u64>,
+    ) -> common_exception::Result<u64> {
+        let key = format!("{}/{}", self.role_prefix, &role_info.name);
+        let value = serde_json::to_vec(&role_info)?;
+
+        let match_seq = match seq {
+            None => MatchSeq::GE(1),
```
drmingdrmer, line 58: If you allow setting value on an absent record, `MatchSeq::Any` may be better: `GE(1)` means a replace operation, which requires the record has to be present.

Since the method name is `upsert` I suppose it allows to `add` a record.
Otherwise, `update_xxx` would be better.

common/management/src/role/role_mgr.rs
```diff
@@ -0,0 +1,191 @@
...
+        let role_val_seq = self.get_role(&role_name, seq);
+        let mut role_info = role_val_seq.await?.data;
+        role_info
+            .grants
+            .revoke_privileges(&role_name, &host_name, &object, privileges);
+        let seq = self.upsert_role_info(&role_info, seq).await?;
+        Ok(Some(seq))
+    }
+
+    async fn drop_role(&self, role_name: String, seq: Option<u64>) -> Result<()> {
```
drmingdrmer, line 171: It's recommended to use type `MatchSeq` for `seq`.

`Option<_>` is a legacy way to specify which seq to match. It is unclear what to do with `None`: sometimes it means to `add`, sometimes it means to just ignore seq check.

drmingdrmer (approved):

PR #3894, opened 2022-01-18
[metasrv] fix connection failed when metasrv restart that caused by token expired

ariesdevil:

## Summary

Bugfix: fix connection failed when metasrv restart that caused by token expired.
This fix reuse raft nodes connect pool that @drmingdrmer made.

## Changelog

- Bug Fix

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/meta/grpc/src/grpc_client.rs
+++ b/common/meta/grpc/src/grpc_client.rs
@@ -133,9 +133,9 @@ impl MetaGrpcClient {
             Some(t) => t,
             None => MetaGrpcClient::handshake(&mut client, &self.username, &self.password).await?,
         };
-        {
-            *t = Some(token.clone());
-        }
+
+        *t = Some(token.clone());
+
         let client = { MetaClient::with_interceptor(channel, AuthInterceptor { token }) };
 
         Ok(client)
--- a/common/meta/grpc/src/grpc_client.rs
+++ b/common/meta/grpc/src/grpc_client.rs
@@ -16,13 +16,14 @@ use std::fmt::Debug;
 use std::time::Duration;
 
 use common_arrow::arrow_format::flight::data::BasicAuth;
+use common_containers::ItemManager;
+use common_containers::Pool;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_exception::SerializedError;
 use common_grpc::ConnectionFactory;
 use common_grpc::RpcClientTlsConfig;
 use common_meta_types::protobuf::meta_client::MetaClient;
-use common_meta_types::protobuf::GetReply;
 use common_meta_types::protobuf::GetRequest;
 use common_meta_types::protobuf::HandshakeRequest;
 use common_meta_types::protobuf::RaftRequest;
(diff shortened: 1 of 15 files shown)


BohuTANG:
Cool, from the test, it seems there is no shutdown-and-restart test for metasrv to check the expired works or not, is it possbile to add it?

ariesdevil:
> Cool, from the test, it seems there is no shutdown-and-restart test for metasrv to check the expired works or not, is it possbile to add it?

Yup

common/meta/grpc/src/grpc_client.rs
```diff
@@ -41,56 +44,86 @@ use crate::grpc_action::MetaGrpcWriteReq;
...
+
 pub struct MetaGrpcClient {
-    #[allow(dead_code)]
-    token: Vec<u8>,
-    pub(crate) client: MetaClient<InterceptedService<Channel, AuthInterceptor>>,
+    conn_pool: Pool<MetaChannelManager>,
+    addr: String,
+    username: String,
+    password: String,
+    // pub(crate) client: MetaClient<InterceptedService<Channel, AuthInterceptor>>,
```
drmingdrmer, line 78: This comment can be removed?

common/meta/grpc/src/grpc_client.rs
```diff
@@ -132,7 +165,8 @@ impl MetaGrpcClient {
         let req: Request<RaftRequest> = (&act).try_into()?;
         let req = common_tracing::inject_span_to_tonic_request(req);
 
-        let result = self.client.clone().write_msg(req).await?.into_inner();
+        let mut client = self.make_client_with_tls().await?;
+        let result = client.write_msg(req).await?.into_inner();
```
drmingdrmer, line 169: It looks like every time calling a method a new client will be created(with a cached `channel`) and a handshake message will be sent?

There is still chance metasrv restarts just after handshake is well served and before `client.write_msg()` is sent?

ariesdevil: > It looks like every time calling a method a new client will be created(with a cached channel) and a handshake message will be sent?
This can be changed to using the old token to do action, if encounter an unauthenticated error, fire a handshake request to fetch a new token and replace the old one.  What do you think about this way?

> There is still chance metasrv restarts just after handshake is well served and before client.write_msg() is sent?
Yep, cause `handshake` and `write_msg` are not an atomic operation, so I think this is inevitable. In this case, we need business layers that holds the client to do retries according to their logic.

ariesdevil: Sorry for the late cause I encounter a deadlock problem:(

common/meta/grpc/src/grpc_client.rs
```diff
@@ -25,72 +29,116 @@ use common_meta_types::protobuf::meta_client::MetaClient;
...
-            MetaClient::with_interceptor(channel, AuthInterceptor { token })
+        let mut t = self.token.write().await;
+        let token = match t.clone() {
+            Some(t) => t,
+            None => MetaGrpcClient::handshake(&mut client, &self.username, &self.password).await?,
         };
 
-        let rx = Self { token, client };
-        Ok(rx)
+        *t = Some(token.clone());
```
drmingdrmer, line 137: `*t = Some(token.clone());` 
This line can be placed inside the `None` branch of the above `match`?

drmingdrmer (approved):
LGTM~

Issue #2647, opened 2021-11-04
[Test/Integration/Tracking] implement generic test framework for integration test


ZhiHanZ:
**Summary**
Similar to integration test in kubernetes and kubebuilder https://book.kubebuilder.io/cronjob-tutorial/writing-tests.html
For local test:
1, we use cargobin to ensure binaries on desired local and bootstrap cluster through code similar to bendctl basic test.
2. For stateless and stateful test we could choose desired number of query instance to run and deploy test them uniformly on test_env.
The test framwork would be
```bash
├── integration
│   ├── Cargo.toml
│   └── src
│       ├── framework
│       │   └── local
│       ├── lib.rs
│       └── sql
```
For cluster test:
we need to ensure some third party kubernetes dependencies and do the same thing

- [ ] #3844

PsiACE:
> 
> ```shell
> ├── integration
> │   ├── Cargo.toml
> │   └── src
> │       ├── framework
> │       │   └── local
> │       ├── lib.rs
> │       └── sql
> ```
>

```shell
├── integration
│   ├── Cargo.toml
｜   ├── sql # If it is large, we can keep only a subset
│   └── src
│       ├── framework
│       │   └── local
│       └── lib.rs
> ```

Xuanwo:
Come on, Let's go! I expect that we can remove all python scripts after this issue implemented.

Xuanwo:
I'm not interested in this issue any more. Feel free to take it :rocket:

leiysky:
Is it possible to support error check in this framework? 

The current integration tests relies on hint to specify the error code, which is not a good method.

ZhiHanZ:
> Is it possible to support error check in this framework? 
> 
> The current integration tests relies on hint to specify the error code, which is not a good method.

PTAL at the new framework,  https://github.com/datafuselabs/databend/pull/5048 
It should support error regex matching . Retry and result sorting in the new framework

PsiACE:
Now we have a new sql logic test framework

PR #3901, opened 2022-01-19
tests: PoC of new test framework

Xuanwo:

## Summary

This PR is just a PoC of #2647, feel free to leave your comments here.

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6937,6 +6937,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "13a4ec180a2de59b57434704ccfad967f789b12737738798fa08798cd5824c16"
 
 [[package]]
+name = "tests-framework"
+version = "0.1.0"
+dependencies = [
+ "inventory",
+]
+
+[[package]]
+name = "tests-integration"
+version = "0.1.0"
+dependencies = [
+ "inventory",
+ "tests-framework",
+]
+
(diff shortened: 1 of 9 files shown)


Xuanwo:
Here is a dicision that we need to make:

- Rewrite all python/shell scripts in rust, so that we can be more platform unrelated.
- Keep the python and shell scripts to make the work easier, in this way, we need to find a way to manage all our python and shell dependences. For example, `poetry` or `Pipenv`, instead of `python -m pip`.

Xuanwo:
risinglightdb's logic: https://github.com/risinglightdb/risinglight/blob/main/tests/sqllogictest.rs

sundy-li:
For Python scripts, how about making these scripts run inside the docker?

ariesdevil:
> For Python scripts, how about making these scripts run inside the docker?

For CI test that's a good way, but for developer test on the local machine, using docker will increase the complexity.

ZhiHanZ:
> > For Python scripts, how about making these scripts run inside the docker?
> 
> For CI test that's a good way, but for developer test on the local machine, using docker will increase the complexity.

I think by using dev container, developers could use docker to run python related scripts with nearly zero code changes

Xuanwo:
Closed for seeking better ideas.

PR #3896, opened 2022-01-18
Use mysql-connector-python instead of deprecated mysql-connector

ariesdevil:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Use mysql-connector-python instead of deprecated mysql-connector
ref: https://pypi.org/project/mysql-connector/

## Changelog

- Build/Testing/CI

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

PR #3902, opened 2022-01-19
Use peotry to manage all python packages

Xuanwo:

## Summary

This PR adds peotry to manage all python packages, so that we can avoid `python -m pip` while setting up.

Let's use `peotry install`!

Related https://github.com/datafuselabs/databend/pull/3901

Should be updated after https://github.com/datafuselabs/databend/pull/3896

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests


PsiACE:
I think there are a few points that need attention.

1. we probably don't need to put it in the root directory? The python-related files will only be used for testing purposes.
2. also, is it really necessary to keep the lock file? If we specify a specific version.
3. please consider integrating it into our existing ci/docker. Then let's remove the python dependencies from the script.

Xuanwo:
> 1. we probably don't need to put it in the root directory? The python-related files will only be used for testing purposes.

I think `poetry install` is much convenient than `cd xxxx && poetry install`. And I prefer to manage all python deps at once.

> 2. also, is it really necessary to keep the lock file? If we specify a specific version.

Yes, lock file is important and we need to track it.

Lock file is the key to reproducible build and keep us safe from things like `faker.js`.

> 3. please consider integrating it into our existing ci/docker. Then let's remove the python dependencies from the script.

Yes, but I prefer to split the work into several PRs.

In this one, we just discuss introducing the `poetry` itself and the version that we expect to pin.

And in the coming PRs, we will discover how to use `peotry` and how to refactor our dev setup scripts and so on.

ariesdevil:
LGTM
If we can't avoid using python, I prefer using poetry:)

BohuTANG:
> I think there are a few points that need attention.
> 
> 1. we probably don't need to put it in the root directory? The python-related files will only be used for testing purposes.

I still treat `poetry` as developent requeirement(only for our tests or a bit others), so agree with @PsiACE not put them in the root if it possible, to make the root directory as clean as possible.

`cd xxxx && poetry install` will in the setup.sh

> 2. also, is it really necessary to keep the lock file? If we specify a specific version.
> 3. please consider integrating it into our existing ci/docker. Then let's remove the python dependencies from the script.

Xuanwo:
OK, let's move `pyproject.toml` into tests.

PsiACE (approved):
LGTM, let's do some subtraction next.

PR #3912, opened 2022-01-19
impl `getData` trait for boolean/string column & improve column

Veeupup:

## Summary

- [x] impl `getData` for BooleanColumn
- [ ] impl `getData` for StringColumn
- [ ] impl `Debug` for Column

## Changelog

- Improvement
- 
## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues2/src/columns/boolean/iterator.rs
+++ b/common/datavalues2/src/columns/boolean/iterator.rs
@@ -13,10 +13,10 @@
 // limitations under the License.
 
 use std::sync::Arc;
-use std::sync::RwLock;
 
 use common_arrow::arrow::bitmap::utils::BitmapIter;
 use common_arrow::arrow::bitmap::MutableBitmap;
+use common_infallible::RwLock;
 
 use crate::prelude::*;
 
--- a/common/datavalues2/src/columns/boolean/iterator.rs
+++ b/common/datavalues2/src/columns/boolean/iterator.rs
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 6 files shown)


common/datavalues2/src/columns/boolean/iterator.rs
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::Arc;
+use std::sync::RwLock;
```
PsiACE, line 16: `common_infallible::RwLock` or `parking_lot::RwLock`

common/datavalues2/src/columns/boolean/mod.rs
```diff
@@ -99,6 +124,8 @@ impl Column for BooleanColumn {
         unsafe {
             Arc::new(Self {
                 values: self.values.clone().slice_unchecked(offset, length),
```
sundy-li, line 126: It's false cached, so we do not need to clone the original slices?

common/datavalues2/src/columns/boolean/iterator.rs
Veeupup: got it.

common/datavalues2/src/columns/boolean/mod.rs
Veeupup: if not cached data, we do not need to copy the data, but if data cached, it seems that we need to copy the underlying data and make `data_cached` true.

Veeupup:
scalar type need to refactor.

PR #3934, opened 2022-01-21
dal2: Implement SeekableReader

### LINKED ISSUE #3874: Integrate dal2 into `streams` (replace dal, implement AsyncSeek)

Xuanwo:

## Summary

Fix #3874 

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests


common/dal2/src/ops/io.rs
```diff
@@ -117,3 +125,110 @@ where F: FnMut(usize)
...
+where
+    S: super::Read<S> + super::Stat<S>,
+{
+    let o = da.stat(key).await?;
+
+    // We will create a BufReader with a large buf so it only read big chunk from underlying storage.
+    Ok(BufReader::with_capacity(
+        4 * 1024 * 1024, // 4 MiB
+        SeekableReader::new(da, key, o.size),
+    ))
```
PsiACE, line 147: I think it might be beneficial to allow the buffer size to be adjusted.

Xuanwo: Makes sense. As it's a simple change which doesn't affect the whole design, how about adding it while we really need this feature?

common/dal2/src/ops/io.rs
```diff
@@ -117,3 +125,110 @@ where F: FnMut(usize)
...
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        _cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let key = self.key.clone();
+        let da = self.da.clone();
+        let pos = self.pos;
+
+        let n = block_on(async {
```
dantengsky, line 195: are there any other workarounds other than `block_on`?

Xuanwo: I tried a lot in the past three days, and I can't find a good alternative for `block_on`.

Basicly, it seems that calling `reader.poll(cx)` inside `poll_read()` will lead to dead Loop, and I can't figure where and how to resolve it.

Here is a record of my blood and tears: https://note.xuanwo.io/#/page/jan%2020th%2C%202022

common/dal2/src/ops/io.rs
PsiACE: ok

common/dal2/src/ops/io.rs
dantengsky: >  it seems that calling reader.poll(cx) inside poll_read() will lead to dead Loop
 
 may I take a look at the code of this version?

Xuanwo: Please take a look at https://github.com/Xuanwo/databend/commits/dal2-apply-2, thanks in advance!

BohuTANG:
@mergify update

common/dal2/src/ops/io.rs
dantengsky: After quick browsing of https://github.com/Xuanwo/databend/commits/dal2-apply-2 :

Instead of `dead loop`,  more likely,  the `Future` returned by `poll_read` might be kept/trapped in the pending state

https://github.com/Xuanwo/databend/blob/3ff525a704b0752ea8c8ebe4d7daad7bf2f6b5d6/common/dal2/src/ops/io.rs#L196-L210
~~~.rust
        let n = async {
            let mut builder = da.read(key.as_str());
            let r = builder
                .offset(pos)
                .size(buf.len() as u64)
                .run()
                .await
                .map_err(io::Error::other)?;


            Pin::new(r).read(buf).await
        };


        pin_mut!(n);


        let n = ready!(n.poll(cx))?;
~~~

IMO, `AsyncReader` should hold a reference to the n (the `Future`, not the `usize`), to keep it from being dropped (too early)

Xuanwo: @dantengsky Thanks for the great review! I managed to learn how rust future works, and come out a better solution for implementing the seek_reader. PTAL.

dantengsky (approved):
LGTM

PR #3940, opened 2022-01-21
Fix pruning get error when expression is not monotonic

### LINKED ISSUE #3941: bug: pruning get error when expression is not monotonic in variable range
### Search before asking

### Version

Main branch

### What's Wrong?

```
mysql> create table test(when DateTime, a UInt32, b UInt8);                                                             Query OK, 0 rows affected (0.01 sec)
Read 0 rows, 0 B in 0.001 sec., 0 rows/sec., 0 B/sec.

mysql> INSERT INTO test SELECT now() + number * 20 as when, number, number%128   FROM numbers(100000000) LIMIT 5000;
Read 40000 rows, 320 KB in 0.006 sec., 6.26 million rows/sec., 50.07 MB/sec.

mysql> select * from test where a-b>0;
ERROR 1105 (HY000): Code: 1000, displayText = Function '-' is not monotonic in the variables range.
```

### What You Expected?

`select` can return the right result

### How to Reproduce?

_No response_

### Anything Else?

_No response_

- [x] Yes I am willing to submit a PR!

zhyass:

## Summary

RangeFilter eval return true if the expression is not monotonic in the variables range.

## Changelog

- Bug Fix

## Related Issues

Fixes #3941

## Test Plan

Unit Tests

### DIFF
--- a/common/planners/src/plan_expression_monotonicity.rs
+++ b/common/planners/src/plan_expression_monotonicity.rs
@@ -173,9 +173,11 @@ impl ExpressionMonotonicityVisitor {
         expr: &Expression,
         variables: HashMap<String, (Option<DataColumnWithField>, Option<DataColumnWithField>)>,
         single_point: bool,
-    ) -> Result<Monotonicity> {
+    ) -> Monotonicity {
         let visitor = Self::create(schema, variables, single_point);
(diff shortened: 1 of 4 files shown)


dantengsky (approved):
LGTM

PR #3943, opened 2022-01-21
[CI] migrate to dev container and self hosted CI

ZhiHanZ:

## Summary

Use dev container for CI and migrate to new self hosted container 

## Changelog

- Build/Testing/CI

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
How about to use buildkite agent later?

ZhiHanZ:
> How about to use buildkite agent later?

the true challenge is to allocate and monitor suffient  resource for CI which is indispensable for both buildkite and github self hosted action runner, here I am fine tuning on that..

Xuanwo:
I'm thinking about running all tests and building inside containers on Github-hosted runners.

So we can avoid the Github env's effect and unify the same build env for both `x86-64` and `aarch`. There is no need for `cross` anymore which will also make our CI much simpler.

Another benefit is a developer can reproduce the same env inside CI.

Xuanwo:
Also, ping @PsiACE for advice. Is there anything we must be done by Github Actions?

PsiACE:
> Also, ping @PsiACE for advice. Is there anything we must be done by Github Actions?

I think it's great, it looks like we can clean up some of the complexity. The only question is, if we use containers, roughly how much time will it take to finish the build/will there be more OOM.

Xuanwo:
Closed for https://github.com/datafuselabs/databend/pull/4630 get merged. Feel free to reopen if there are other things to address.

PR #3955, opened 2022-01-23
codegen work with datavalue2

zhyass:

## Summary

Summary about this PR

## Changelog

- Other 

## Related Issues

Fixes #issue


common/codegen/src/writes/arithmetics_type.rs
```diff
@@ -45,24 +47,33 @@ pub fn codegen_arithmetic_type() {
...
-use crate::DFPrimitiveType;
+use crate::Scalar;
 
 pub trait ResultTypeOfBinary {{
-    type AddMul: DFPrimitiveType;
-    type Minus: DFPrimitiveType;
-    type IntDiv: DFPrimitiveType;
-    type Modulo: DFPrimitiveType;
-    type LeastSuper: DFPrimitiveType;
+    type AddMul: Scalar;
```
sundy-li, line 53: Scalar includes type `PrimitiveTypes` & `String` & `Boolean`, maybe we just need to make it as `PrimitiveType` ?

zhyass: > Scalar includes type `PrimitiveTypes` & `String` & `Boolean`, maybe we just need to make it as `PrimitiveType` ?

Yes, changed.

Issue #3823, opened 2022-01-11
bug: Unable to get field named "number". Valid fields: ["(number % 2)"] (while in select before projection).


BohuTANG:
## Summary

``` sql
mysql> select number%2 as c1, number%3 as c2 from numbers(10) group by c1;
ERROR 1105 (HY000): Code: 6, displayText = Unable to get field named "number". Valid fields: ["(number % 2)"] (while in select before projection).
```

` Unable to get field named "number"` -- wrong error hints.

ClickHouse:
```
Received exception from server (version 21.12.2):
Code: 215. DB::Exception: Received from localhost:9000. DB::Exception: Column `number` is not under aggregate function and not in GROUP BY: While processing number % 2 AS c1, number % 3 AS c2. (NOT_AN_AGGREGATE)

```

wangzhen11aaa:
Hi, BohuTANG.

I am tracing this problem. 
Some help wanted.
When we create new_data_fileds from the expr, like ("number%2", "number%3", "c1", "c2"), why not combine these fileds with filed ("number")?

wangzhen11aaa:
/assignme 

Let me try.

wangzhen11aaa:
select number%2 as c1, number%3 as c2 from numbers(10) group by c;

The SQL is 

select number%2 as c1, number%3 as c2 from numbers(10) group by (number%2);

and 

This sql is wrong.  the c2 is not in the group by key.

mysql> select number%2 as c1, sum(number) from numbers(10) group by (c1);
+------+-------------+
| c1   | sum(number) |
+------+-------------+
|    0 |          20 |
|    1 |          25 |
+------+-------------+
2 rows in set (2.30 sec)
Read 10 rows, 80 B in 0.609 sec., 16.43 rows/sec., 131.41 B/sec.

In my view,  this is not a good issue. @BohuTANG

wangzhen11aaa:
> 

So, we want to fix this wrong error hint problem.

PR #3956, opened 2022-01-23
Make error message for group by exception good

wangzhen11aaa:

## Summary
Make error message for group by exception good
## Changelog

- New Feature
- Bug Fix
- Improvement
https://github.com/datafuselabs/databend/issues/3823
- Performance Improvement
- Build/Testing/CI
- Documentation
- Other 

## Related Issues
https://github.com/datafuselabs/databend/issues/3823

Fixes #issue

## Test Plan
Done
Unit Tests
Done
Stateless Tests
Done

### DIFF
--- a/common/exception/src/exception.rs
+++ b/common/exception/src/exception.rs
@@ -75,10 +75,14 @@ impl ErrorCode {
     pub fn add_message_back(self, msg: impl AsRef<str>) -> Self {
         Self {
             code: self.code(),
-            display_text: format!("{}{}", self.display_text, msg.as_ref()),
+            display_text: format!("{}{}", self.rewirte_err_msg_for_groupby(&self.display_text), msg.as_ref()),
             cause: self.cause,
             backtrace: self.backtrace,
         }
+    }
+
+    pub fn rewirte_err_msg_for_groupby(&self, display_text: &String) -> String{
+        display_text.replace("\"", "`").replace("Unable to get field named", "Column").replace(". Valid fields:","is not under aggregate function and not in GROUP BY: While processing ")
     }
 
     pub fn backtrace(&self) -> Option<ErrorCodeBacktrace> {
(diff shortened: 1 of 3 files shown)


common/exception/src/exception.rs
```diff
@@ -75,12 +75,16 @@ impl ErrorCode {
...
             code: self.code(),
-            display_text: format!("{}{}", self.display_text, msg.as_ref()),
+            display_text: format!("{}{}", self.rewirte_err_msg_for_groupby(&self.display_text), msg.as_ref()),
             cause: self.cause,
             backtrace: self.backtrace,
         }
     }
 
+    pub fn rewirte_err_msg_for_groupby(&self, display_text: &String) -> String{
+        display_text.replace("\"", "`").replace("Unable to get field named", "Column").replace(". Valid fields:","is not under aggregate function and not in GROUP BY: While processing ")
```
BohuTANG, line 85: Hi, it would be better that we check and return the error from the underlying, instead of replaced.

BohuTANG: For issue #3823 , we don't need to follow the ClickHouse way, but the error is not correct for that case, it looks we can't only replace the message text.

wangzhen11aaa: > For issue #3823 , we don't need to follow the ClickHouse way, but the error is not correct for that case, it looks we can't only replace the message text.

Give more info.

BohuTANG: Thanks for reminding,  I think the issue #3823 should be closed(I have closed), it's not a critical error, from the initial plan of this issue, we want to get a better way to make the message looks more better, it looks there is no good way.
Feel free to comment.

wangzhen11aaa: > Hi, it would be better that we check and return the error from the underlying, instead of replaced.
Yes, fn add_message_back() is so fundamental.
I think the original error message is good enough to express the error, we may use this kind of field search in many places (1 to n model) . Maybe the position to make the message more specific for each case is the like 

`               Err(cause) => {
                    return Err(cause.add_message_back(" (while in select before group by)"));
}`
give a cause some pub method to change its information in every case. 
Maybe that's ok.

wangzhen11aaa:
Close now.

PR #3961, opened 2022-01-23
Add hash functions of datavalues2

sundy-li:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


tests/suites/0_stateless/02_0009_function_hashes.result
```diff
@@ -1,7 +1,7 @@
-5091324831805182738
+18110648197875983073
```
PsiACE, line 1: also need to fix the description in the documentation

common/functions/src/scalars/hashes/hash.rs
```diff
@@ -12,28 +12,59 @@
...
+
         factory.register("blake3", Blake3HashFunction::desc());
         factory.register("xxhash32", XxHash32Function::desc());
         factory.register("xxhash64", XxHash64Function::desc());
+        factory.register("siphash64", SipHash64Function::desc());
+        factory.register("siphash", SipHash64Function::desc());
+    }
+}
+
+#[cfg(test)]
```
PsiACE, line 49: It needs to be placed in the right place

PsiACE: I noticed that there are some other files that contain similar tests mod. I think they may not be involved in the compilation (with `test = false` in Cargo.toml).

PsiACE (approved):
LGTM

sundy-li:
Ok， I will improve this in another pr to reduce the valuable resources of tests.

common/functions/src/scalars/hashes/hash_base.rs
```diff
@@ -0,0 +1,156 @@
...
+        features = features.deterministic();
+        Function2Description::creator(Box::new(Self::try_create)).features(features)
+    }
+
+    fn exec<S>(column: &ColumnRef) -> Result<ColumnRef>
+    where
+        S: Scalar,
+        for<'a> <S as Scalar>::RefType<'a>: DFHash,
+    {
+        let unary = ScalarUnaryExpression::<S, R, _>::new(|a| {
```
sundy-li, line 65: @skyzh  

Function2 has `Send + Sync + Clone` bounds, so it might not be possible to have the lambda function inside functions?

Now we just create `ScalarUnaryExpression` in each batch eval.

common/datavalues/src/arrays/ops/vec_hash.rs
```diff
@@ -81,6 +81,7 @@ impl VecHash for DFStringArray {
     fn vec_hash(&self, hasher: DFHasher) -> Result<DFUInt64Array> {
         Ok(self.apply_cast_numeric(|v| {
             let mut h = hasher.clone_initial();
+            // Hash::hash_slice(v,&mut h);
```
BohuTANG, line 84: No used any more?

common/functions/src/scalars/expressions/unary.rs
```diff
@@ -0,0 +1,43 @@
...
+use common_exception::Result;
+
+/// An common struct to caculate Unary expression scalar op.
+pub struct ScalarUnaryExpression<L: Scalar, O: Scalar, F> {
+    func: F,
+    _phantom: PhantomData<(L, O)>,
+}
+
+impl<'a, L: Scalar, O: Scalar, F> ScalarUnaryExpression<L, O, F>
+where F: Fn(L::RefType<'a>) -> O
```
skyzh, line 27: This bound looks not correct. F should be valid for all lifetime (e.g., `for<'a> F: Fn(L::RefType<'a>)`). If we implement expressions like this, we will need to create an expression every time we want to evaluate an array.

You may refer to day 6 "Erase Expression Lifetime" in https://github.com/skyzh/type-exercise-in-rust for solutions.

common/functions/src/scalars/hashes/hash_base.rs
skyzh: I don't think this is related to `Send + Sync + Clone` bound. See my comments below.

common/datavalues/src/arrays/ops/vec_hash.rs
sundy-li: datavalues will be removed, so I just add a comment that the previous usage is wrong.

common/functions/src/scalars/expressions/unary.rs
BohuTANG: Thank you @skyzh .
If so we should consider it in the later PR, cc @sundy-li

sundy-li: Thanks, I get it, these will be adressed after we introduce datatypes into the creator function,  then we can have single expression to evaluate arrays multiple times.

PR #3991, opened 2022-01-25
chore(datavalue2): fix too-strong bound on ScalarBinaryExpression

skyzh:
Signed-off-by: Alex Chi <iskyzh@gmail.com>

## Summary

I made the same mistake when building my own Array type system. <del>To brief, if we constraint the `ColumnRef` to have `'a` lifetime, we have to create an expression every time we want to do evaluation:</del>

```rust
for (a, b) in columns_left.iter().zip(columns_right.iter()) {
  let func = ScalarBinaryFunction::new(some_sql_func);
  func.eval(a, b);
}
```

... instead of

```rust
let func = ScalarBinaryFunction::new(some_sql_func);
// if we keep the 'a bound, the following a, b must have the same lifetime of
// `'a` at this point, but this is impossible -- they will be dropped in each
// loop. This snippet will CE with the `'a` bound applied to `ColumnRef`.
for (a, b) in columns_left.iter().zip(columns_right.iter()) {
  func.eval(a, b);
}
```

The above example may compile even when we retain the `'a` bound... Anyway, removing unnecessary lifetime could prevent future issues, especially when implementing generics types.

## Changelog

- Improvement

## Test Plan

Compile pass

### DIFF
--- a/common/functions/src/scalars/expressions/binary.rs
+++ b/common/functions/src/scalars/expressions/binary.rs
@@ -27,7 +27,7 @@ pub struct ScalarBinaryExpression<L: Scalar, R: Scalar, O: Scalar, F> {
     _phantom: PhantomData<(L, R, O)>,
 }
 
-impl<'a, L: Scalar, R: Scalar, O: Scalar, F> ScalarBinaryExpression<L, R, O, F>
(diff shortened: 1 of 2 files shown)


zhyass (approved):
LGTM

PsiACE (approved):

PR #3993, opened 2022-01-26
Jwt Auth for http handler

youngsofun:

## Summary

1. introduce AuthMgr to init and hold extenal Authenticators
    1. not in servers/http, because the clickhouse may support ladp/kebrose later.
2. introduce JwtAuthenticator
    1. using crate `jwt_simple`, which we already used in meta. pure rust and better interface.
    2. implement a simple jwk client.
    3. temporarily limitation (but the interface for extentions is ready)
        1. only support RS256, which is most widely used
        2. only support pub key with jwk， which is more userful, file based pub key need to specify the alg or guess it from pem.
3. http handler auth
    1. support both Basic and JWT
    2. auth done in a middleware and extract UserInfo.
    3. temporarily auth is not forced. if no Credential provided, fall back to root(should check root is `no_password` when root is real user in meta).
    4. temporarily https should be but is not forced too. 
4. test for jwt is coverd by http handler test with jwt.

related but not done in this PR:

1. a real root user.
2. auto refresh of jwks keys
3. more config option for jwt/jwk. (on demand), e.g. currently assume username in standard 'subject' field int the claims.
4. jwt with pub key in local file. (optinal)

## Changelog

- New Feature

## Related Issues

 #3482

## Test Plan

Unit Tests


youngsofun:
\review @flaneur2020

common/meta/types/src/user_auth.rs
```diff
@@ -163,7 +171,7 @@ impl AuthInfo {
                 hash_value: p,
                 hash_method: t,
             } => t.to_string(p),
-            AuthInfo::None => "".to_string(),
+            AuthInfo::None | AuthInfo::JWT => "".to_string(),
```
flaneur2020, line 174: typo? it seems jwt has already pattern matched in an earlier match arm.

query/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -218,14 +219,15 @@ async fn query_page_handler(
 #[poem::handler]
 pub(crate) async fn query_handler(
     sessions_extension: Data<&Arc<SessionManager>>,
+    user_info: Data<&UserInfo>,
```
flaneur2020, line 222: 6666

query/src/users/auth/jwt.rs
```diff
@@ -0,0 +1,176 @@
...
+                    )),
+                    Some(subject) => Ok(subject),
+                },
+                Err(err) => Err(ErrorCode::AuthenticateFailure(err.to_string())),
+            },
+        }
+    }
+}
+
+mod jwk {
```
flaneur2020, line 62: move jwk a seperate file? it seems in-file mod is not very common in the current code base

flaneur2020:
jwt intergration with a cleaner auth interface, lgtm!

common/meta/types/src/user_auth.rs
youngsofun: @flaneur2020  in two diff functions get_type and get_auth_string

query/src/users/auth/jwt.rs
youngsofun: ok

youngsofun: thank you，I did noticed that, too.
I coded this way for 1. not too many code here,  2.  I want to stress that jwk is sub mod of jwt.

which is better, cc @BohuTANG :

 1.  jwk is not sub mod of jwt ,  might also make sense

- jwt/
   - mod.rs  # empty
   - jwt.rs
   - jwk.rs

 2.   current code base also not put much code in mod.rs?

- jwt/
   - mod.rs  # jwt code
   - jwk.rs

flaneur2020:
```
jwt/
mod.rs # empty
jwt.rs
jwk.rs
```

personally i prefer this style, makes mod.rs simple, only `use` statements inside 🤔

common/meta/types/src/user_auth.rs
flaneur2020: oh my bad, sorry :joy:

query/src/users/auth/jwt.rs
youngsofun: 👌

PR #3998, opened 2022-01-26
datavalues2: Logic function

### LINKED ISSUE #3971: datavalues2 && logic functions

Veeupup:

## Summary

datavalues2 logic function
- [x] and
- [x] or
- [x] xor
- [x] not

## Changelog

- New Feature

## Related Issues

Fixes #3971 

## Test Plan

Unit Tests

Stateless Tests


Veeupup:
constant folding with `And` will produce unexcepted result, I comment it for now to make it work right.
and we can remove unnecessary constant folding in another pr.

https://github.com/datafuselabs/databend/blob/12f5b58048/query/src/optimizers/optimizer_constant_folding.rs#L258-L265

common/functions/src/scalars/logics/logic.rs
```diff
@@ -12,57 +12,187 @@
...
+
+            let func = match self.op {
+                LogicOperator::And => LogicAndFunction::eval_with_null,
+                LogicOperator::Or => LogicOrFunction::eval_with_null,
+                LogicOperator::Xor => LogicXorFunction::eval_with_null,
+                LogicOperator::Not => return Err(ErrorCode::LogicalError("never happen")),
+            };
+
+            for idx in 0..input_rows {
+                let (val, valid) = func(
```
Veeupup, line 126: can we make this func `inline` to have better performence? cc @sundy-li

sundy-li: Yes, better to test it vs before.

BTW, I do think `func` is dynamically dispatched without optimization.


I think it may be better to apply:

```
macro_rules! xxx;

match op {
LogicOperator::And => xxx!{},
...
}
```

sundy-li (approved):

junli1026 (approved):

PR #4008, opened 2022-01-27
fix if with null & fix convert to nullable column bug

Veeupup:

## Summary

* modify `if` behavoior, in detail:
before:
`select if(null, 1, 2)` -> `null`
now:
`select if(null, 1, 2)` -> `2`
when considering with if, null will be used as `false`.

* fix `old-datavalues` column convert into nullable `new-datavalues` column bug

* A Special case:when casting `NullColumn` into `BooleanColumn`, it will return a `BooleanColumn` with full `false`. It will not affect other behaviors.

## Changelog

- Bug Fix
- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues2/src/convert/column_convert.rs
+++ b/common/datavalues2/src/convert/column_convert.rs
@@ -27,7 +27,7 @@ use crate::NullableColumn;
 
 pub fn convert2_new_column(column: &OldDataColumnWithField) -> ColumnWithField {
     let result = convert2_new_column_nonull(column);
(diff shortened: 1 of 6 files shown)


Veeupup:
our @databend-bot  usually makes duplicate comments... 😢

sundy-li (approved):

PsiACE (approved):
LGTM

Issue #4004, opened 2022-01-27
Constant folding Optimizer with `And` will produce unexcepted result


Veeupup:
constant folding with `And` will produce unexcepted result, I comment it for now to make it work right.
and we can remove unnecessary constant folding in another pr.

In detail: null value will be optimized as `false`, so when `select null and 1`, it will return `0` rather than `NULL`, which is not excepted.

https://github.com/datafuselabs/databend/blob/12f5b58048/query/src/optimizers/optimizer_constant_folding.rs#L258-L265

_Originally posted by @Veeupup in https://github.com/datafuselabs/databend/issues/3998#issuecomment-1022303490_

Veeupup:
@zhyass 
it seems that`or` expression will give the right result in this situation because we will consider `null` to `false` and it will go deeper to call the `LogicFunction`.
And agree with you, constant folding with special handling (just ignore null type) for `null` will be ok to solve it.

zhyass:
@Veeupup 
I test the `or` expression in branch `datavalues-dev`, and compare with `MySQL` and `ClickHouse`

In `databend`:
```
mysql> select Null or 1<0;
+-------------------+
| (NULL OR (1 < 0)) |
+-------------------+
|                 0 |
+-------------------+
1 row in set (0.01 sec)
Read 1 rows, 1 B in 0.000 sec., 2.07 thousand rows/sec., 2.07 KB/sec.

mysql> select Null or Null;
+----------------+
| (NULL OR NULL) |
+----------------+
|              0 |
+----------------+
1 row in set (0.00 sec)
Read 1 rows, 1 B in 0.000 sec., 2.56 thousand rows/sec., 2.56 KB/sec.

mysql> select Null or 1>0;
+-------------------+
| (NULL OR (1 > 0)) |
+-------------------+
|                 1 |
+-------------------+
1 row in set (0.00 sec)
Read 1 rows, 1 B in 0.000 sec., 3.64 thousand rows/sec., 3.64 KB/sec.
```

In `MySQL`:
```
mysql> select Null or 1<0;
+-------------+
| Null or 1<0 |
+-------------+
|        NULL |
+-------------+
1 row in set (0.04 sec)

mysql> select Null or Null;
+--------------+
| Null or Null |
+--------------+
|         NULL |
+--------------+
1 row in set (0.04 sec)

mysql> select Null or 1>0;
+-------------+
| Null or 1>0 |
+-------------+
|           1 |
+-------------+
[...]

PR #4014, opened 2022-01-28
Fix Unable to get field name bug

### LINKED ISSUE #4012: Unable to get field named
bug1:
```
mysql> select 1 and number>1 from numbers(10);
ERROR 1105 (HY000): Code: 1006, displayText = Unable to get field named "(1 AND (number > 1))". Valid fields: ["number", "(number > 1)"].
```

bug2:
```
mysql> select number<3 or number>5 from numbers(10);
ERROR 1105 (HY000): Code: 1006, displayText = Unable to get field named "((number < 3) OR (number > 5))". Valid fields: ["number", "((number < 3) or (number > 5))"].
```

They're caused by constant folding optimization(constant condition remove). Bug2 is  caused by the case sensitivity of the operator https://github.com/datafuselabs/databend/blob/main/query/src/optimizers/optimizer_constant_folding.rs#L164.

zhyass:

## Summary

CC @Veeupup , it's another bug, I just move `constant_cond_remove` from `constant_folding` to `expression_transform`. The problem `null value will be optimized as false` described in the issue #4004  is still exists.
I think doing this in the filter maybe has no effect.

This is the same problem as issue #2165

## Changelog

- Bug Fix

## Related Issues

Fixes #4012 

## Test Plan

Unit Tests

### DIFF
--- a/query/src/optimizers/optimizer_expression_transform.rs
+++ b/query/src/optimizers/optimizer_expression_transform.rs
@@ -149,6 +149,108 @@ impl ExprTransformImpl {
             _ => Ok(origin.not_eq(lit(0))),
         }
     }
+
+    fn constant_transformer(origin: &Expression) -> Result<Expression> {
+        let (column_name, left, right, is_and) = match origin {
+            Expression::BinaryExpression { op, left, right } => match op.to_lowercase().as_str() {
+                "and" => (origin.column_name(), left, right, true),
+                "or" => (origin.column_name(), left, right, false),
+                _ => return Ok(origin.clone()),
+            },
+            _ => return Ok(origin.clone()),
+        };
+
+        let left = Self::constant_transformer(left)?;
+        let right = Self::constant_transformer(right)?;
+
+        let mut is_remove = false;
+
+        let mut left_const = false;
+        let new_left = Self::eval_const_cond(
+            column_name.clone(),
+            &left,
+            is_and,
+            &mut left_const,
+            &mut is_remove,
+        )?;
+        if is_remove {
+            return Ok(new_left);
+        }
+
+        let mut right_const = false;
+        let new_right = Self::eval_const_cond(
+            column_name.clone(),
+            &right,
+            is_and,
+            &mut right_const,
+            &mut is_remove,
+        )?;
+        if is_remove {
+            return Ok(new_right);
+        }
+
+        match (left_const, right_const) {
+            (true, true) => {
+                if is_and {
+                    Ok(Expression::Literal {
+                        value: DataValue::Boolean(Some(true)),
+                        column_name: Some(column_name),
+                        data_type: DataType::Boolean,
+                    })
+                } else {
+                    Ok(Expression::Literal {
+                        value: DataValue::Boolean(Some(false)),
+                        column_name: Some(column_name),
+                        data_type: DataType::Boolean,
+                    })
+                }
+            }
+            (true, false) => Ok(new_right),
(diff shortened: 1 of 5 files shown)


query/tests/it/optimizers/optimizer_expression_transform.rs
```diff
@@ -115,6 +115,70 @@ async fn test_expression_transform_optimizer() -> Result<()> {
                 \n    Filter: true\
                 \n      ReadDataSource: scan schema: [number:UInt64], statistics: [read_rows: 0, read_bytes: 0, partitions_scanned: 0, partitions_total: 0], push_downs: [projections: [0], filters: [true]]",
             },
+            Test {
```
Veeupup, line 118: maybe add a new test to check it?

query/src/optimizers/optimizer_expression_transform.rs
```diff
@@ -201,7 +303,8 @@ impl PlanRewriter for ExprTransformImpl {
         }
 
         let new_input = self.rewrite_plan_node(plan.input.as_ref())?;
-        let new_predicate = Self::boolean_transformer(&plan.predicate)?;
+        let new_predicate = Self::constant_transformer(&plan.predicate)?;
```
Veeupup, line 306: `having`?

zhyass: > `constant_transformer` for `having`?

Thanks for reminding, added

zhyass:
`explain` in `MySQL`, It seems that mysql is only optimized for filter/join/having.
```
mysql> explain select true and a>0 from t;
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------+
|  1 | SIMPLE      | t     | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    2 |   100.00 | NULL  |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------+
1 row in set, 1 warning (0.08 sec)

mysql> show warnings\G
*************************** 1. row ***************************
  Level: Note
   Code: 1003
Message: /* select#1 */ select (true and (`zzq`.`t`.`a` > 0)) AS `true and a>0` from `zzq`.`t`
1 row in set (0.04 sec)

mysql> explain select * from t where true and a>0;
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra       |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
|  1 | SIMPLE      | t     | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    2 |    50.00 | Using where |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
1 row in set, 1 warning (0.04 sec)

mysql> show warnings\G
*************************** 1. row ***************************
  Level: Note
   Code: 1003
Message: /* select#1 */ select `zzq`.`t`.`a` AS `a`,`zzq`.`t`.`b` AS `b`,`zzq`.`t`.`c` AS `c` from `zzq`.`t` where (`zzq`.`t`.`a` > 0)
1 row in set (0.04 sec)
```

I

Veeupup (approved):
LGTM

PR #4018, opened 2022-01-28
Use buf reader for meta reader

### LINKED ISSUE #3938: Use buffered-reader to read table metadata
Currently, fuse table's metadata, namely the table snapshot, segment information, and the FileMetaData of parquet are all read from underlying storage(e.g S3) by using the non-buffered "bare-mental" readers. Since there are lots of scattered reads for the reading of those metadata, especially for the FileMetaData, this leads to a bad performance. A recent experiment shows that a table of common settings, with several billion rows, may take about 20s to load these metadata (they are likely to be cached for a long period of time afterward though).

Thus, buffered reading is needed for them.

This is where the InputStream is provided for the MetaReaders.

https://github.com/datafuselabs/databend/blob/a89b24cbbefa2e07c1143311ea4fb29ada0f6889/query/src/storages/fuse/io/meta_readers.rs#L161-L166

The `InputStream` is just a boxed AsyncRead + AsyncSeek + Send + Unpin in disguise.

And the above `InputStream` is needed to be wrapped by a `BufReader`, as the BlockReader did.

https://github.com/datafuselabs/databend/blob/a89b24cbbefa2e07c1143311ea4fb29ada0f6889/query/src/storages/fuse/io/block_reader.rs#L105-L108

The settings of the size of read buffer could be obtained from context, like this:

https://github.com/datafuselabs/databend/blob/a89b24cbbefa2e07c1143311ea4fb29ada0f6889/query/src/storages/fuse/operations/read.rs#L68

GrapeBaBa:

## Summary

Use buf reader for meta reader

## Changelog
- Performance Improvement

## Related Issues

Fixes #3938 
## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/storages/fuse/io/meta_readers.rs
+++ b/query/src/storages/fuse/io/meta_readers.rs
@@ -22,6 +22,7 @@ use common_exception::ErrorCode;
 use common_exception::Result;
 use common_tracing::tracing::debug_span;
 use common_tracing::tracing::Instrument;
+use futures::io::BufReader;
 use serde::de::DeserializeOwned;
 
 use crate::sessions::QueryContext;
@@ -34,12 +35,12 @@ use crate::storages::fuse::io::snapshot_location;
 use crate::storages::fuse::meta::SegmentInfo;
 use crate::storages::fuse::meta::TableSnapshot;
 
-/// Provider of [InputStream]
+/// Provider of [BufReader]
 ///
 /// Mainly used as a auxiliary facility in the implementation of [Loader], such that the acquirement
-/// of an [InputStream] can be deferred or avoided (e.g. if hits cache).
-pub trait InputStreamProvider {
-    fn input_stream(&self, path: &str, len: Option<u64>) -> Result<InputStream>;
+/// of an [BufReader] can be deferred or avoided (e.g. if hits cache).
+pub trait BufReaderProvider {
+    fn buf_reader(&self, path: &str, len: Option<u64>) -> Result<BufReader<InputStream>>;
 }
 
 /// A Newtype for [FileMetaData].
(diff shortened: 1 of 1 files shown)


query/src/storages/fuse/io/meta_readers.rs
```diff
@@ -22,6 +22,7 @@ use common_exception::ErrorCode;
 use common_exception::Result;
 use common_tracing::tracing::debug_span;
 use common_tracing::tracing::Instrument;
+use futures::io::BufReader;
```
ariesdevil, line 25: Why use `futures::io::BufReader` instead of `tokio::io::BufReader` here?

GrapeBaBa: Follow the #3938 description, `futures::io::BufReader` was used in block_reader.rs. Actually I also have confusion since `std::io:BufReader` was also used in databend.

GrapeBaBa:
@Xuanwo i wonder if some meta reader can be wrapped as `seekablereader` here?

query/src/storages/fuse/io/meta_readers.rs
dantengsky: `InputStream` is `futures::AsyncRead + futures::Seek + ... ` , which does not fit with other two `BufReader`s (at least out of box)

FYR:

https://github.com/datafuselabs/databend/blob/b9070da6d34682363acfb03ce78329052b1f40ef/common/dal/src/data_accessor.rs#L30-L32

dantengsky (approved):
looks good to me, @Xuanwo any other concerns?

BohuTANG:
Thank you @GrapeBaBa ，春节快乐!

Issue #866, opened 2021-06-22
[sqlparser] better sql parser?


BohuTANG:
**Summary**

[sqlparser-rs](https://github.com/ballista-compute/sqlparser-rs)  is not easy to extend and we are looking for a good alternative(if have), comments are welcome

Reference:
https://github.com/datafuselabs/datafuse/blob/36dce2064742a1d64bfcf9a7cd308d0f5a3e4eb5/fusequery/query/src/sql/sql_parser.rs

BohuTANG:
[nom-sql[(https://github.com/ms705/nom-sql) but no update for a long time

PsiACE:
It seems that sqlparser lacks good alternatives. If necessary, we may need to maintain one by ourselves.

[quaint](https://github.com/prisma/quaint) is not a parser, but
- An AST for building dynamic SQL queries.
- Visitors for different databases to generate SQL strings.

[toydb-parser](https://github.com/erikgrinaker/toydb/tree/master/src/sql/parser) is the parser for toydb

sundy-li:
How about parsers from yacc rules?

dantengsky:
FYI : 

lalrpop: an lr(1) parser generator   (lexer could be customized)

repo: https://github.com/lalrpop/lalrpop
doc: http://lalrpop.github.io/lalrpop/

-  never exercise it before
-  seems it lacks precedence/association rules declaration (again I am not sure.)

After googling around, I found a SQL parser using it:

https://github.com/hmwill/vervolg/blob/main/src/sql.lalrpop

leiysky:
> [nom-sql](https://github.com/ms705/nom-sql)
> but no update for a long time

`nom` is a nice parser combinator, with which we can extend SQL syntax easily.

I think we can fork [nom-sql](https://github.com/ms705/nom-sql) or make one from scratch(not a big deal).

Here's a `nom` [Cypher parser](https://github.com/leiysky/tigraph/blob/master/src/parser/parser.rs) I wrote at a Hackathon, which only cost me about few hours :rofl: . FYI.

BohuTANG:
> How about parsers from yacc rules?

https://github.com/lalrpop/lalrpop is rust yacc

leiysky:
Based on my experience, working with bison(yacc) is suffering.

Here's the MySQL's yacc file, which has about 17,000 LOC.

https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy

[...]

PR #4024, opened 2022-01-29
Add tokenizer for the new sql parser

andylokandy:

## Summary

As attempting to implement #866, and after some deep discussion with @leiysky, we've decided to reimplement the sql parser with nom. But, while the idiom of nom parses directly on `&str`, we, like most of the other sql parsers, need a tokenizer to achieve better error messaging and better maintainability of the parser.  So this is the PR to add the tokenizer as the first step of implementing the new sql parser.

I've investigated the tokenizer of `sqlparser-rs` and have tried to implement one using the crate `regex`. As the result, the crate [`logos`](https://crates.io/crates/logos) achieved higher performance while paying the least effort.

The lexer rules in this PR follow this [official reference of PostgreSQL](https://www.postgresql.org/docs/9.1/sql-syntax-lexical.html), and the keyword definitions are completely copied from `sqlparser-rs` which I think may contain the keywords not defined in PostgreSQL. 

## Changelog

- New Feature

## Related Issues

Ref https://github.com/datafuselabs/databend/issues/866

## Test Plan

Unit Tests

### DIFF
--- a/common/ast/Cargo.toml
+++ b/common/ast/Cargo.toml
@@ -8,19 +8,21 @@ edition = "2021"
 
 [lib]
 doctest = false
+test = false
 
 [dependencies] # In alphabetical order
 # Workspace dependencies
 common-exception = { path = "../exception" }
 common-functions = { path = "../functions" }
-logos = "0.12"
 
 # Github dependencies
 sqlparser = { git = "https://github.com/datafuse-extras/sqlparser-rs", rev = "c33837e" }
 
 # Crates.io dependencies
 async-trait = "0.1.52"
+logos = "0.12"
 once_cell = "1.9.0"
+thiserror = "1.0.30"
 
 [dev-dependencies]
 pretty_assertions = "1.0.0"
--- a/common/ast/Cargo.toml
+++ b/common/ast/Cargo.toml
@@ -8,12 +8,12 @@ edition = "2021"
 
 [lib]
 doctest = false
-test = false
 
 [dependencies] # In alphabetical order
 # Workspace dependencies
 common-exception = { path = "../exception" }
 common-functions = { path = "../functions" }
+logos = "0.12"
 
 # Github dependencies
 sqlparser = { git = "https://github.com/datafuse-extras/sqlparser-rs", rev = "c33837e" }
--- a/common/ast/src/parser/token.rs
+++ b/common/ast/src/parser/token.rs
@@ -1,0 +1,1125 @@
+// Copyright 2021 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use logos::Logos;
+use logos::Span;
+
+pub use self::TokenKind::*;
+
+#[allow(non_camel_case_types)]
+#[derive(Logos, Clone, Copy, Debug, PartialEq)]
+pub enum TokenKind {
+    #[error]
+    #[regex(r"[ \t\n\f]+", logos::skip)]
+    Whitespace,
+
+    #[regex(r"--[^\t\n\f]*", logos::skip)]
+    Comment,
+
+    #[regex(r"/\*([^\*]|(\*[^/]))*\*/", logos::skip)]
+    CommentBlock,
+
+    #[regex(r#"[_a-zA-Z][_$a-zA-Z0-9]*"#)]
+    Ident,
+
+    #[regex(r#""[_a-zA-Z][_$a-zA-Z0-9]*""#)]
+    QuotedIdent,
+
+    #[regex(r#"'([^'\\]|\\t|\\u|\\n|\\'|'')*'"#)]
+    LiteralString,
+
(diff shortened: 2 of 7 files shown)


common/ast/Cargo.toml
```diff
@@ -8,12 +8,12 @@ edition = "2021"
 
 [lib]
 doctest = false
-test = false
```
andylokandy, line 11: Did we intended to disable the unit tests?

Xuanwo: Hi, databend's unit tests are built too slow, so we move all of them to integration tests instead. Please take a test at https://github.com/datafuselabs/databend/discussions/3473

Xuanwo:
@andylokandy It's great to see you in the databend community!

common/ast/Cargo.toml
PsiACE: Currently, we do not have any so-called unit tests :)

andylokandy: @PsiACE @Xuanwo Ah, thanks for your fast response. So I'm curious about how do you test the private interface.

leiysky: > @PsiACE @Xuanwo Ah, thanks for your fast response. So I'm curious about how do you test the private interfaces. In addition, how do you test with some mock technics with `#[cfg(test)]`.

By making it pub 😆

PsiACE: Two methods: 
1. add the test to src/tests , remove test = false.
2. Make it public for now and put it under tests/it.

The vast majority of tests in Databend follow the 2nd method. 

We are exploring new ways to balance the testing needs of private modules.

andylokandy: Oh, then the fabulous mocking tool [mockall](https://crates.io/crates/mockall) will not work...

common/ast/src/parser/token.rs
```diff
@@ -0,0 +1,1117 @@
...
+    Whitespace,
+
+    #[regex(r"--[^\t\n\f]*", logos::skip)]
+    Comment,
+
+    #[regex(r"/\*([^\*]|(\*[^/]))*\*/", logos::skip)]
+    CommentBlock,
+
+    #[regex(r#"[_a-zA-Z][_$a-zA-Z0-9]*"#)]
+    Ident,
```
leiysky, line 34: `Ident` should be distinguished with reserved words. For example:

```
SELECT select FROM t; -- not allowed
SELECT select_result FROM t; -- allowed
```

I think we need a complicated regular expression to handle this, or make an extra pass to check validity of `Ident`s(not a good choice).

How do you think about this?

common/ast/Cargo.toml
PsiACE: @andylokandy We do use mockall, but I'm not sure how you plan to use it. Here is one of our use cases, and the only one. :sob: https://github.com/datafuselabs/databend/blob/c54f06fc3788307aad67511e494dece562cb6d67/common/management/tests/it/user.rs

common/ast/src/parser/token.rs
```diff
@@ -0,0 +1,1117 @@
...
+
+use logos::Logos;
+use logos::Span;
+
+pub use self::TokenKind::*;
+
+#[allow(non_camel_case_types)]
+#[derive(Logos, Clone, Copy, Debug, PartialEq)]
+pub enum TokenKind {
+    #[error]
```
leiysky, line 23: What about reserve a `TokenKind::Error` for this?

andylokandy: This is required by `logos` according to its documentation.

common/ast/src/parser/token.rs
andylokandy: I think this should be the work for the parser, E.g, your first example will be lexed into `SELECT SELECT FROM ident(t)` on which the parser should be able to detect a grammar error.

common/ast/Cargo.toml
andylokandy: > Two methods:
> 1. add the test to src/tests , remove test = false.
> 2. Make it public for now and put it under tests/it.

Thanks, I'm pretty clear now.

> but I'm not sure how you plan to use it.

Think of two structs `A` and `B`, where `B` is one of the fields of `A`. In the unit test, where `#[cfg(test)]` is on ), I'll mock a `B` and inject it into A to test how `A` interacts with `B`. In the integration test, where `#[cfg(test)]` is off, I'll test `A` with the real `B`.

To give a concrete example:

```rust
#[cfg(test)]
use other::MockB as B;
#[not(cfg(test))]
use other::B;

pub struct A {
    inner: B
}

mod other {
    #[cfg_attr(test, mockall::automock)]
    pub struct B {
        // ...
    }
} 
```

PsiACE: got it, thanks. It looks like our testing style is not up to the task for now.

common/ast/src/parser/token.rs
leiysky:
```rust
...
    #[error]
    Error,

    #[regex(r"[ \t\n\f]+", logos::skip)]
    Whitespace,
...
```

This can work fine. And IMO it's more reasonible to give invalid token a `Error` instead.

common/ast/src/parser/token.rs
leiysky: I was concerning about the ambiguation between reserved words and identifier, but I found `logos` do have a clear rule to handle this.

It looks good to me now.

common/ast/src/parser/token.rs
andylokandy: Addressed

leiysky:
@andylokandy Is this PR ready for merging? If so I will approve and merge it.

andylokandy:
@leiysky Ok, let's roll!

leiysky (approved):

PR #4027, opened 2022-01-29
Add utils for new sql parser

andylokandy:

## Summary

This PR implements the utils for defining SQL grammar and also the utils for testing.

To add a parser of `DROP TABLE` with thess utils, you only need to write:

```rust
pub fn drop_table<'a, Error>(i: Input<'a>) -> IResult<Input<'a>, Statement, Error>
where Error: ParseError<Input<'a>> {
    let (i, (_, _, if_exists, database, table, _)) = rule!(
        DROP ~ TABLE ~ ( IF ~ EXISTS )? ~ ( #ident ~ "." )? ~ #ident ~ ";"
    )(i)?;

    Ok((i, Statement::DropTable {
        if_exists: if_exists.is_some(),
        database: database.map(|(name, _)| name),
        table,
    }))
}
```

The special procedural macro `rule!` is designed for the ease of writing grammar spec as well as to improve maintainability, it follows these simple rules:

1. `TABLE`: match the token by token kind, in particular, `TokenKind::TABLE` in this example.
2. `";"`: match the token by token text.
3. `#fn_name`: an external nom parser function. In the example above, `ident` is a predefined parser for identifiers.
4. `a ~ b ~ c`: a sequence of parsers to take one by one. It'll get expanded into `nom::sequence::tuple`.
5. `(...)+`: one or more repeated patterns. It'll get expanded into `nom::multi::many1`.
6. `(...)*`: zero or more repeated patterns. It'll get expanded into `nom::multi::many0`.
7. `(...)?`: Optional parser. It'll get expanded into `nom::combinator::opt`.
8. `a | b | c`: Choices between a, b, and c. It'll get expanded into `nom::branch::alt`.

To test the parser, you write:

```rust
#[test]
fn test_drop_table() {
    assert_parse!(drop_table, "drop table a;", "DROP TABLE a");
    assert_parse!(
        drop_table,
        r#"drop table if exists a."b";"#,
        r#"DROP TABLE IF EXISTS a."b""#,
    );
}
```

In `assert_parse`, the tokens get parsed into an AST then pretty-printed back to a string and finally compared with the expected result.

Personally, I'd prefer to use the style of [golden test](https://ro-che.info/articles/2017-12-04-golden-tests), which in brief is to compare the test output with the expected result in strings, so one could easily copy the test output to fill the expected (accept the output as golden). This could largely minimize the work on constructing AST by hand and can even not change the test while refactoring the AST and the parser.

## Changelog

- Not for changelog

## Related Issues

Blocked by https://github.com/datafuselabs/databend/pull/4024
Ref https://github.com/datafuselabs/databend/issues/866

## Test Plan

Unit Tests


BohuTANG:
It's helpful to run `make lint` locally for lint :)

BohuTANG:
The golden testing is used for stateless and stateful test, if the unit test using it would be nice, for many unit-test, we are checking the expected and actual block result, like:
https://github.com/datafuselabs/databend/blob/main/query/tests/it/storages/system/databases_table.rs#L43

common/ast/src/parser/rule/util.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::parser::token::TokenKind;
+
+pub type Input<'a> = &'a [Token<'a>];
+
+pub fn satisfy<'a, F, Error>(cond: F) -> impl Fn(Input<'a>) -> IResult<Input<'a>, &'a Token, Error>
+where
+    F: Fn(&Token) -> bool,
+    Error: ParseError<Input<'a>>,
+{
+    move |i| match i.get(0).map(|t| {
```
leiysky, line 32: Maybe we need to check length of `i` here?

andylokandy: if `i.len() == 0` then the `get(0)` will return `None`

leiysky (approved):

PR #4039, opened 2022-02-01
Implement the parser for expression

andylokandy:

## Summary

Implement the parser for expressions. 

The core technique used here to help arrange the operator precedence and associativity is called Pratt parser, which is also known as the Top-Down Operator-Precedence (TDOP) parser. I recommend this [tdop-tutorial](https://eli.thegreenplace.net/2010/01/02/top-down-operator-precedence-parsing) if you're interested in how it works.

## Changelog

## Related Issues

Ref #866

## Test Plan

Unit Tests

### DIFF
--- a/common/ast/src/parser/ast/expression.rs
+++ b/common/ast/src/parser/ast/expression.rs
@@ -22,73 +22,68 @@ use crate::parser::ast::display_identifier_vec;
 #[derive(Debug, Clone, PartialEq)]
 #[allow(dead_code)]
 pub enum Expr {
-    // Wildcard star
-    Wildcard,
-    // Column reference, with indirection like `table.column`
+    /// Column reference, with indirection like `table.column`
     ColumnRef {
         database: Option<Identifier>,
         table: Option<Identifier>,
         column: Identifier,
     },
-    // `IS NULL` expression
-    IsNull(Box<Expr>),
-    // `IS NOT NULL` expression
-    IsNotNull(Box<Expr>),
-    // `[ NOT ] IN (expr, ...)`
+    /// `IS [ NOT ] NULL` expression
+    IsNull { expr: Box<Expr>, not: bool },
+    /// `[ NOT ] IN (expr, ...)`
     InList {
         expr: Box<Expr>,
         list: Vec<Expr>,
         not: bool,
     },
-    // `[ NOT ] IN (SELECT ...)`
+    /// `[ NOT ] IN (SELECT ...)`
     InSubquery {
         expr: Box<Expr>,
         subquery: Box<Query>,
         not: bool,
     },
-    // `BETWEEN ... AND ...`
+    /// `BETWEEN ... AND ...`
     Between {
         expr: Box<Expr>,
-        negated: bool,
         low: Box<Expr>,
         high: Box<Expr>,
+        not: bool,
     },
-    // Binary operation
+    /// Binary operation
     BinaryOp {
         op: BinaryOperator,
         left: Box<Expr>,
         right: Box<Expr>,
     },
-    // Unary operation
-    UnaryOp {
-        op: UnaryOperator,
-        expr: Box<Expr>,
-    },
-    // `CAST` expression, like `CAST(expr AS target_type)`
+    /// Unary operation
+    UnaryOp { op: UnaryOperator, expr: Box<Expr> },
+    /// `CAST` expression, like `CAST(expr AS target_type)`
     Cast {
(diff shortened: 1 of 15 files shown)


common/ast/src/parser/ast/expression.rs
```diff
@@ -22,73 +22,68 @@ use crate::parser::ast::display_identifier_vec;
...
+    CountAll,
+    /// Scalar function call
     FunctionCall {
-        // Set to true if the function is aggregate function with `DISTINCT`, like `COUNT(DISTINCT a)`
+        /// Set to true if the function is aggregate function with `DISTINCT`, like `COUNT(DISTINCT a)`
         distinct: bool,
         name: String,
         args: Vec<Expr>,
         params: Vec<Literal>,
     },
```
leiysky, line 76: I'm thinking about lifting the `distinct` field out and add a `AggregationFunc` variant for aggregation functions(SUM, AVG etc.).

common/ast/src/parser/rule/expr.rs
```diff
@@ -0,0 +1,612 @@
...
+            )));
+        }
+
+        Ok((i, expr))
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+#[allow(dead_code)]
+pub enum ExprElement {
```
leiysky, line 105: Would you add some comment to explain the function of this struct?

common/ast/src/parser/ast/expression.rs
sundy-li: In the level of `AST`, can we distinguish `functions` and `aggregate functions` ?

andylokandy: @sundy-li I think it's viable because the name of aggregate functions are keywords.

common/ast/src/parser/rule/expr.rs
andylokandy: addressed

common/ast/src/parser/ast/expression.rs
sundy-li: Yes. But if we have custom aggregate functions like: `window_funnel`, `retention`, should we add them into 
 as keywords?

leiysky: > Yes. But if we have custom aggregate functions like: `window_funnel`, `retention`, should we add them into as keywords?

In general, it's case by case, but my suggestion is treating aggregate functions names as keywords.

It's more convenient to handle special grammar of specific functions by reserving their names as keyword(e.g. `EXTRACT(... FROM ...)`, `COUNT(*)`, `SUM(DISTINCT ...)`).

Since all aggregate functions should support the `DISTINCT` grammar, and the number of aggregate functions is much lower than scalar functions, I think it's fine to reserve their names as keywords.

common/ast/src/parser/rule/expr.rs
```diff
@@ -0,0 +1,672 @@
...
+            | COUNT
+            | SUM
+            | AVG
+            | MIN
+            | MAX
+            | STDDEV_POP
+            | SQRT
+        },
+        |name| name.text.to_string(),
+    )(i)
```
andylokandy, line 671: @leiysky FYI, this list is probably incomplete. Which solution mentioned in the todo do you prefer?

common/ast/src/parser/ast/expression.rs
leiysky: > But there are two main problems I came up with:
> 
> 1. UDF Aggregate/Scalar Functions （function name is defined by the user  and stored in meta service)
> 2. Combinators:  `*If`  like sumIf.    If we add some aggregate function, we should update the keywords & combinator keywords.

The `UDF`s are context sensitive that cannot be handled by parser.

It seems we have to keep the `distinct` field here for planner to check the semantic of a unrecognizable function...

What about only reserving the special cases like `COUNT`(for `COUNT(*)`) as keyword?

common/ast/src/parser/rule/expr.rs
leiysky: I suppose we just keep `COUNT` as a token and use `Ident` for the rest, i.e. the second way?

sundy-li: > I suppose we just keep `COUNT` as a token and use `Ident` for the rest, i.e. the second way?

Agree.

leiysky (approved):

BohuTANG:
/lgtm

Thank you @andylokandy !

Issue #3206, opened 2021-12-02
`SHOW ENGINES` Statement


BohuTANG:
**Summary**

Same as MySQL:
https://dev.mysql.com/doc/refman/8.0/en/show-engines.html

SHOW ENGINES displays status information about the server's storage engines. This is particularly useful for checking whether a storage engine is supported, or to see what the default engine is.

For example:
```
mysql> SHOW ENGINES\G
*************************** 1. row ***************************
      Engine: FUSE
     Support: YES
     Comment: Fuse storage engine
```

PR #4050, opened 2022-02-04
ISSUE #3206: Support SHOW ENGINES Statement

SGZW:

## Summary

solve #3206 

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/system/engines_table.rs
```diff
@@ -0,0 +1,90 @@
...
+    async fn read(
+        &self,
+        _ctx: Arc<QueryContext>,
+        _plan: &ReadDataSourcePlan,
+    ) -> Result<SendableDataBlockStream> {
+        let mut engine_name = Vec::with_capacity(1);
+        let mut engine_support = Vec::with_capacity(1);
+        let mut engine_comment = Vec::with_capacity(1);
+        // fuse engine
+        engine_name.push("FUSE".as_bytes());
```
BohuTANG, line 76: Let's get the engines information from the StorageFactory :\

BohuTANG: https://github.com/datafuselabs/databend/blob/76e831b753eb85b274314e3ec8c4a30c0be48171/query/src/storages/storage_factory.rs#L51-L73

SGZW: ok, let me fix it

BohuTANG: Looks like we should add `desc()` to `StorageCreator` as the `comment` column:
```
pub trait StorageCreator: Send + Sync {
    fn try_create(&self, ctx: StorageContext, table_info: TableInfo) -> Result<Box<dyn Table>>;
    fn desc(&self) -> &str;
}
```

query/src/storages/storage_factory.rs
```diff
@@ -42,33 +42,58 @@ where
...
+#[derive(Default, Clone)]
+pub struct StorageDescriptor {
+    pub engine_name: String,
+    pub comment: String,
+}
+
 #[derive(Default)]
 pub struct StorageFactory {
     creators: RwLock<HashMap<String, Arc<dyn StorageCreator>>>,
+    descriptors: Vec<StorageDescriptor>,
```
BohuTANG, line 54: Thanks for the descriptor idea.
I think `StorageDescriptor` should be a part of table engine, it explicitly tell what this engine is, as defined by the engine's module not the `Factory`.

query/src/storages/system/engines_table.rs
```diff
@@ -0,0 +1,95 @@
...
+
+pub struct EnginesTable {
+    table_info: TableInfo,
+}
+
+impl EnginesTable {
+    pub fn create(table_id: u64) -> Self {
+        let schema = DataSchemaRefExt::create(vec![
+            DataField::new("Engine", DataType::String, false),
+            DataField::new("Support", DataType::String, false),
```
BohuTANG, line 40: We could remove the `Support` column here, when the engine who is not registered to the factory it means we don't want it to show up, this is a bit different from MySQL engines.

query/src/storages/storage_factory.rs
SGZW: what if every table engine implementes StorageCreateor trait and StorageCreator add interface desc() to return StorageDescriptor?

query/src/storages/system/engines_table.rs
SGZW: ok

query/src/storages/storage_factory.rs
BohuTANG: > what if every table engine implementes StorageCreateor trait and StorageCreator add interface desc() to return StorageDescriptor?

Agreed.
But inclined to change `StorageCreateor` to `StorageDescriptor`, it's not only `create` matter also `comment`.

query/src/storages/system/engines_table.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn get_table_info(&self) -> &TableInfo {
+        &self.table_info
+    }
+
+    async fn read(
+        &self,
+        _ctx: Arc<QueryContext>,
+        _plan: &ReadDataSourcePlan,
+    ) -> Result<SendableDataBlockStream> {
+        let descriptors = _ctx.get_catalog().get_storage_descriptors();
```
BohuTANG, line 72: Why we add `StorageDescriptor` to catalog and impl it for every table engine, how about get the table engine name and descriptor from self.ctx.storage_factory?
It seems we only add a function like `desc()->&str` for the table engine.

SGZW: i think i don't get the real meaning of this. can you explain it more clearly? i think every engine should has it's own creator

SGZW: the reason of why we add StorageDescriptor to catalog is that storage_factory is only MutableCatalog's member variable，so database catalog should expose a api to get descriptors, or we can get descriptors from storage_factory and construct the EnginesTable

SGZW: Descrptor is Attributes of engine，so I impl it for every table engine...

BohuTANG: How about only add one function like description() to table engine instead of the Descriptior trait? And we register the description to storage factory, this looks more clearly.
Then we can add one trait function named 'get_table_engines' to Catalog, which getting the engine details from the mutable catalog.

BohuTANG:
/lgtm
Thank you @SGZW !

Issue #4044, opened 2022-02-02
Add comments for codes to increase readability.


wangzhen11aaa:
Hi, all. 
  Good comments are as good/important as codes, they are good abstraction and summray for a bunch of codes, though codes can express themselves. For the `Maintainability` and `Readability` especially for new one.
   RFC.

wangzhen11aaa:
How about prefixing "_"  in to variable name whoes life will end out local scope?

wangzhen11aaa:
Need some help.

`pub struct Series(pub Arc<dyn SeriesTrait>);`

impl<'a> AsRef<(dyn SeriesTrait + 'a)> for Series {
    fn as_ref(&self) -> &(dyn SeriesTrait + 'a) {
        &*self.0
    }
}

I think the `pub struct Series(pub Arc<dyn SeriesTrait>);` here needs some comments, This kind of code not found in the rust programming. PLS give some references. 

@BohuTANG

wangzhen11aaa:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

BohuTANG:
> Hi, all. For the code style part, the references below may help.
> 
> Code row length: https://www.python.org/dev/peps/pep-0008/#maximum-line-length Comment part: https://www.python.org/dev/peps/pep-0008/#comments
> 
> We may can draw on theses good styles. RFC. Please give more advance.

Thank you for your suggestions, looks like we could use rustfmt comment_width to normalize them.

wangzhen11aaa:
> > Hi, all. For the code style part, the references below may help.
> > Code row length: https://www.python.org/dev/peps/pep-0008/#maximum-line-length Comment part: https://www.python.org/dev/peps/pep-0008/#comments
> > We may can draw on theses good styles. RFC. Please give more advance.
> 
> Thank you for your suggestions, looks like we could use rustfmt comment_width to normalize them.

Do not find the option or argument named `comment_width `, and I tried the rustfmt with one line containing 400+ characters, running it does not give any change. That's awkward. :(

BohuTANG:
> 

It works with to `rustfmt.toml`:
```
comment_width = 100
wrap_comments = true
```

There is a document about it here:
https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#comment_width

wangzhen11aaa:
> > 
> 
> It works with `rustfmt.toml`:
> 
> ```
> comment_width = 100
> wrap_comments = true
> ```
[...]

PR #4059, opened 2022-02-05
Make comment with limited length per row

wangzhen11aaa:

## Summary

Comment format problem.

## Changelog

- Improvement

## Related Issues
https://github.com/datafuselabs/databend/issues/4044

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/arrays/ops/apply.rs
+++ b/common/datavalues/src/arrays/ops/apply.rs
@@ -43,7 +43,8 @@ macro_rules! apply_enumerate {
 }
 
 pub trait ArrayApply<'a, A, B> {
-    /// Apply a closure elementwise and cast to a Numeric DFPrimitiveArray. This is fastest when the null check branching is more expensive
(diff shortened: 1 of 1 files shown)


PsiACE:
Glad to see your efforts. But I suggest it is better to introduce an .editorconfig and apply it to the whole project. Rather than some piecemeal commits.

wangzhen11aaa:
> #4044



> Glad to see your efforts. But I suggest it is better to introduce an .editorconfig and apply it to the whole project. Rather than some piecemeal commits.
> 
> We should point out what the problem is, what should be done, and what not to do.
> 
> You can specifically describe some suggestions and ideas about formatting, readability, editor configuration in #4044 .

Ok, I mainly tried to learn the structure of databend, fixing these comments or variable name in passing, not systematically. 
We can maintain these rules in #4044 like pep8 for python.

Issue #3677, opened 2021-12-28
Tracking issue for Vision of Databend DAL


Xuanwo:
- https://github.com/datafuselabs/databend/discussions/3662
- Related Issues
  - https://github.com/datafuselabs/databend/issues/3586
  - https://github.com/datafuselabs/databend/issues/2976
  - https://github.com/datafuselabs/databend/issues/3225

## Roadmap

- [x] https://github.com/datafuselabs/databend/issues/3436
  - https://github.com/datafuselabs/databend/pull/3575
- [x] Implement services for PoC
  - [x] https://github.com/datafuselabs/databend/pull/3695
  - [x] https://github.com/datafuselabs/databend/issues/3774
    - We need to wait for datafuselabs/opendal#8 
    - [x] Need enable_pod_iam_policy support
- [x] datafuselabs/databend#3874
  - Sometimes we only need to read till end, don't need total size
    - https://github.com/datafuselabs/databend/pull/4001#discussion_r793214217
- [x] https://github.com/datafuselabs/databend/issues/3937
- [x] https://github.com/datafuselabs/databend/pull/4067
- [x] https://github.com/datafuselabs/databend/pull/4069
- [x] https://github.com/datafuselabs/databend/pull/4081
  - [x] Remove all `block_on` to avoid performance issues
  - [x] eliminate all `unwrap` in non-testing code
  - [x] `create table test as select * from numbers(1234567890);select * from test;` doesn't works: Fixed in https://github.com/datafuselabs/databend/pull/4107
  - [x] Make sure streaming load works: https://databend.rs/user/data-loading/http-streaming-load#examples
- [x] datafuselabs/opendal#8

## Longterm Ideas

- [x] ~~Rewrite `parquet2` into cloud storage native style~~
  - Remove `AsyncSeek` trait requirement (use style like `pread` instead of `open-seek-read`)
  - parquet is designed by appending metadata in the file tail, this task is impossible.
  - Maybe we can find a more s3 native data format
- [x] Split `dal2` into a separate repo: moved to https://github.com/datafuselabs/opendal
  - more independent
  - easy to do integration tests

Xuanwo:
All required work has been done. [opendal](https://github.com/datafuselabs/opendal) will handle all new feature requests. Let's start a new journey!

Thank @dantengsky, for much valuable help!

PR #4069, opened 2022-02-08
dal2: Implement Context support

Xuanwo:

## Summary

Part of https://github.com/datafuselabs/databend/issues/3677

This PR implements a DAL Context which will implement `Layer` and `Accessor`.

## Changelog

- New Feature

## Test Plan

Unit Tests

Stateless Tests


common/dal-context/src/lib.rs
```diff
@@ -0,0 +1,153 @@
+// Copyright 2022 Datafuse Labs.
```
BohuTANG, line 1: It would be better we split lib.rs into some sprerate files, to make the lib.rs clean.
If we put them in lib.rs file, guessing that more and more codes will be put in the future.

Xuanwo: Makes sense, I will split them.

Xuanwo: Addressed, PTAL.

PsiACE (approved):

Issue #1359, opened 2021-08-09
[handler] http server handler


BohuTANG:
**Summary**

Tasks:
- [x] #2241

PR #4076, opened 2022-02-08
clear HTTP query result with timeout

youngsofun:

## Summary

1. clean up HTTP query not deleted by the client with a timeout. 
    1. for async HTTP query， set the expiration time to `None` on receiving a req, and to `now + timeout` before responding, expecting the client to request again within the timeout limit, otherwise kill it and delete it from query manager
    2. add a config for it. In most cases clients do not care about it, so not in the request JSON,  `wait_time_secs` used for long polling is in reverse（in the config, not request）.
2. mv PaginationConf to JSON body from HTTP query parameter. So the client can only set it in the first result, which is enough at most times.
3. make sleep time in unit tests shorter.
4. some small refactor.

## Changelog

## Related Issues

part of https://github.com/datafuselabs/databend/issues/1359

## Test Plan

Unit Tests

Stateless Tests


youngsofun:
\review @flaneur2020

flaneur2020 (approved):

flaneur2020:
LGTM ✪ω✪

PR #4084, opened 2022-02-09
[metasrv] refactor: remove CLI arg "--boot". Use "--single" instead.

drmingdrmer:

##### [metasrv] refactor: remove CLI arg "--boot". Use "--single" instead.
`metasrv` automatically ignores initializing arguments such as
`--single` and `--join`, if metasrv is already initialized.

Thus when restarting, metasrv can just use the same CLI args as
initializing it.

## Changelog

- Improvement

## Related Issues

### DIFF
--- a/common/meta/raft-store/src/config.rs
+++ b/common/meta/raft-store/src/config.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use clap::Parser;
-use common_meta_types::MetaError;
 use common_meta_types::MetaResult;
 use common_meta_types::NodeId;
 use once_cell::sync::Lazy;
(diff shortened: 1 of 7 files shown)


ZhiHanZ (approved):
LGTM

common/meta/raft-store/src/config.rs
```diff
@@ -159,12 +153,6 @@ impl RaftConfig {
...
 
     pub fn check(&self) -> common_exception::Result<()> {
-        if self.boot && self.single {
-            return Err(ErrorCode::InvalidConfig(
-                "--boot and --single can not be both set",
-            ));
-        }
-
         Ok(())
     }
```
PsiACE, line 157: lgtm, but in this case, I think we can remove the `check`

drmingdrmer: Hmm... Yes it's a little bit dirty in this commit.
So let me fill in some useful stuff later. Such as `--single` and `--join` can not be specified together, or `--join` must not be set to itself.

PsiACE: ok

Issue #3975, opened 2022-01-24
datavalues2 && string functions  (some of them are already done, you can take the example in `fn register2`)


sundy-li:
- [x] bin
- [x] hex
- [x] oct
- [x] repeat
- [x] mid
- [x] substr
- [x] substring
- [x] substring_index
- [x] unhex
- [x] locate
- [x] position
- [x] instr
- [x] insert
- [x] field
- [x] concat
- [x] replace
- [x] strcmp

Please feel free to comments here to take part of them.

Xuanwo:
Let's rock!

Xuanwo:
- [x] ascii
- [x] bit_length
- [x] octet_length
- [x] char_length
- [x] character_length
- [x] ord
- [x] length

PR https://github.com/datafuselabs/databend/pull/4070 resolved all upper functions.

- [ ] bin
- [ ] hex
- [ ] oct
- [ ] repeat
- [ ] mid
- [ ] substr
- [ ] substring
- [ ] substring_index
- [ ] unhex
- [ ] locate
- [ ] position
- [ ] instr
- [ ] insert
- [ ] field
- [ ] concat
- [ ] replace
- [ ] strcmp
- [ ] left
- [ ] right
- [ ] concat_ws
- [ ] elt
- [ ] space
- [ ] lpad
- [ ] rpad
- [ ] export_set
- [ ] find_in_set
- [ ] format
- [ ] char

Please feel free to comments here to take part of them.

sundy-li:
I'll take functions from `left` to `char`.

Veeupup:
@likg227 maybe u can take the following function to have some fun!
- [ ] locate
- [ ] position

sundy-li:
I will take:

- [ ] instr
- [ ] insert
- [ ] field
- [ ] concat
- [ ] replace
- [ ] strcmp

PR #4088, opened 2022-02-10
string function: bin

Veeupup:

## Summary

function string `bin`

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/strings/bin.rs
+++ b/common/functions/src/scalars/strings/bin.rs
@@ -1,4 +1,4 @@
-// Copyright 2021 Datafuse Labs.
+// Copyright 2022 Datafuse Labs.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


common/functions/src/scalars/strings/bin.rs
```diff
@@ -14,103 +14,158 @@
...
-impl Function for BinFunction {
+impl Function2 for BinFunction {
     fn name(&self) -> &str {
         "bin"
     }
 
-    fn return_type(&self, args: &[DataTypeAndNullable]) -> Result<DataTypeAndNullable> {
-        if !args[0].is_numeric() && !args[0].is_null() {
+    fn return_type(&self, args: &[&DataTypePtr]) -> Result<DataTypePtr> {
+        let dt = if args[0].is_nullable() {
```
sundy-li, line 50: I think we can use pass_through_nullable here.

common/functions/src/scalars/strings/bin.rs
```diff
@@ -14,103 +14,158 @@
...
+            remove_nullable(args[0]).data_type_id()
+        } else {
+            args[0].data_type_id()
+        };
+
+        if !dt.is_numeric() && !dt.is_string() && !dt.is_null() {
```
sundy-li, line 56: We can strictly accept numeric data types. 

 String is not allowed, null is covered in `pass_through_nullable `

sundy-li: Old wrong tests can be removed.

Veeupup: get it.

sundy-li (approved):

Xuanwo:
Please also update https://github.com/datafuselabs/databend/issues/3975. For example: leave a comment or link the issue in PR description.

BohuTANG:
I see it's still in Draft but with Approved, does PR ready to merge? cc @Veeupup

Issue #3972, opened 2022-01-24
datavalues2 && math functions


sundy-li:

sundy-li:
I will finish the sign function as an example.

PsiACE:
i will take

- [ ] abs
- [ ] angle

sundy-li:
I will take the others, let's finish the migration ASAP.

PR #4094, opened 2022-02-10
refactor(functions/maths): migrate pi/crc32/exp to function2

PsiACE:

## Summary

- [x] pi
- [x] crc32
- [x] exp

## Changelog

- Improvement

## Related Issues

Part of #3972 

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/maths/pi.rs
```diff
@@ -15,47 +15,44 @@
...
 
-    fn eval(&self, _columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
-        Ok(DataColumn::Constant(
-            DataValue::Float64(Some(PI)),
-            input_rows,
-        ))
+    fn eval(&self, _columns: &ColumnsWithField, input_rows: usize) -> Result<ColumnRef> {
+        let mut builder: ColumnBuilder<f64> = ColumnBuilder::with_capacity(1);
+        builder.append(PI);
+        Ok(builder.build(1))
```
PsiACE, line 55: It looks like this will be fine, what do you think? cc @sundy-li

sundy-li: I think
```
ConstColumn::new(Series::from_data(vec![PT]), input_rows)
```
will be fine.

PsiACE: but if change it like you said, we will get

```
Diff < left / right > :
<PrimitiveColumn 	 typeid: Float64	 len: 1	 data: [3.141592653589793]
>PrimitiveColumn 	 typeid: Float64	 len: 0	 data: []
```

sundy-li: Oh, there is a bug in:

```
        with_match_primitive_type_id!(dt, |$T| {
            fmt_dyn!(&self, PrimitiveColumn<$T>, f)
        },
```
change &self to `col` will fix this, it's ok to fix this in this pr.

PsiACE: got it

common/functions/src/scalars/maths/math.rs
```diff
@@ -40,17 +43,17 @@ use crate::scalars::TrigonometricSinFunction;
 use crate::scalars::TrigonometricTanFunction;
 use crate::scalars::TruncNumberFunction;
 
+pub type CRC32Function = BaseHashFunction<CRC32, u32>;
```
PsiACE, line 46: That's enough, and I think it might be better to move to the hash mod.

tests/suites/0_stateless/02_function/02_0014_function_maths.sql
```diff
@@ -43,7 +43,7 @@ SELECT '===exp===';
 SELECT exp(NULL);
 SELECT exp(2);
 SELECT exp('2');
-SELECT exp('a');
+SELECT exp('a'); -- {ErrorCode 1010}
```
PsiACE, line 46:
```
ERROR 1105 (HY000) at line 46: Code: 1010, displayText = Expected server error code: 1105 but got: 1010.
Cast error happens in casting from String to Float64.
```

I think that's appropriate.

Veeupup (approved):

tests/suites/0_stateless/02_function/02_0014_function_maths.sql
sundy-li: I think we should use the strict mode for casting by default, it's better than returning null or zero.

common/functions/src/scalars/maths/exp.rs
```diff
@@ -14,62 +14,64 @@
...
     }
 
-    fn return_type(&self, args: &[DataTypeAndNullable]) -> Result<DataTypeAndNullable> {
-        let nullable = args.iter().any(|arg| arg.is_nullable());
-
-        let data_type = if args[0].is_numeric() || args[0].is_string() || args[0].is_null() {
-            Ok(DataType::Float64)
+    fn return_type(&self, args: &[&DataTypePtr]) -> Result<DataTypePtr> {
+        let data_type = if args[0].data_type_id().is_numeric()
+            || args[0].data_type_id().is_string()
```
sundy-li, line 51: We should strict args[0] to be `args[0].data_type_id().is_numeric()`.

1. Reject other types like: String 
2. `passthrough_null` is enabled by default, so null_type is covered in `Function2Adaptor`:   https://github.com/datafuselabs/databend/blob/f076522bd3e12c4e3a8ef59c3a6e949446e4adb9/common/functions/src/scalars/function2_adapter.rs

common/functions/src/scalars/maths/exp.rs
```diff
@@ -14,62 +14,64 @@
...
-            .column()
-            .to_minimal_array()?
-            .cast_with_type(&DataType::Float64)?
-            .f64()?
-            .apply_cast_numeric(|v| v.exp());
-        let column: DataColumn = result.into();
-        Ok(column.resize_constant(columns[0].column().len()))
+    fn eval(&self, columns: &ColumnsWithField, _input_rows: usize) -> Result<ColumnRef> {
+        let column = cast_column_field(&columns[0], &Float64Type::arc())?;
+        let viewer = ColumnViewer::<f64>::create(&column)?;
```
sundy-li: Do not need to use `ColumnViewer`.

1. Number of args is 1, so constant case is covered in `Function2Adaptor` if we enable `passthrough_constant` by default.

2. `passthrough_null` is enabled by default, so we do not need to care about nullable case.


We can use `UnaryExpression` here.

common/functions/src/scalars/maths/exp.rs
PsiACE: Okay, so we will no longer support `exp('2')` and I will remove it from stateless.

PsiACE: or change  to `SELECT exp('2'); -- {ErrorCode 1007}`

sundy-li: I prefer the latter.

sundy-li (approved):

PR #4096, opened 2022-02-10
Refactor: migrate interval function for datavalue2

zhyass:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/expressions/binary.rs
```diff
@@ -15,8 +15,122 @@
 use std::marker::PhantomData;
 
 use common_datavalues2::prelude::*;
+use common_exception::ErrorCode;
 use common_exception::Result;
 
+#[derive(Debug, Clone)]
+pub struct EvalContext {
```
sundy-li, line 22: Let's move it into a new file, cause unary may also need it.

common/functions/src/scalars/dates/macros.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+#[macro_export]
+macro_rules! define_date_add_year_months {
+    ($l: ident, $r: ident, $ctx: ident, $date_type:ident, $op: expr) => {{
+        let factor = $ctx.factor;
+        let epoch = NaiveDate::from_ymd(1970, 1, 1);
+        let naive = epoch.checked_add_signed(Duration::days($l as i64));
+        if naive.is_none() {
+            $ctx.error = Some(ErrorCode::Overflow(format!(
```
sundy-li, line 67: Better to add a `set_error` function:

```
fn set_error(&mut self, e: ErrorCode) {
    if self.error.is_none() {
       self.error = e;
    }
}
```

sundy-li: This keeps the first error.

common/functions/src/scalars/expressions/binary.rs
```diff
@@ -17,6 +17,101 @@ use std::marker::PhantomData;
 use common_datavalues2::prelude::*;
 use common_exception::Result;
 
+use super::EvalContext;
+
+pub trait ScalarBinaryFunction2<L: Scalar, R: Scalar, O: Scalar> {
```
sundy-li, line 22: I think we just introduce `ctx` into `ScalarBinaryFunction` and refactor others using this function is better.

`ScalarBinaryExpression2` & `ScalarBinaryFunction2` can be merged into `ScalarBinaryExpression` & `ScalarBinaryRefFunction`

zhyass: > I think we just introduce `ctx` into `ScalarBinaryFunction` and refactor others using this function is better.
> 
> `ScalarBinaryExpression2` & `ScalarBinaryFunction2` can be merged into `ScalarBinaryExpression` & `ScalarBinaryRefFunction`

Yes, I will merged after add unit and stateless tests

junli1026 (approved):

sundy-li (approved):

PR #4102, opened 2022-02-11
String Func: repeat/substring/substr/mid

Veeupup:

## Summary

string func:
- [x] repeat
- [x] substring
- [x] mid
- [x] substr  
- [x] substring_index

## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/strings/substring.rs
```diff
@@ -14,235 +14,97 @@
...
-        let dt = DataType::String;
-        Ok(DataTypeAndNullable::create(&dt, nullable))
+
+        Ok(StringType::arc())
     }
 
-    fn eval(&self, columns: &DataColumnsWithField, input_rows: usize) -> Result<DataColumn> {
-        let s_column = columns[0].column().cast_with_type(&DataType::String)?;
-        let p_column = columns[1].column().cast_with_type(&DataType::Int64)?;
+    fn eval(&self, columns: &ColumnsWithField, input_rows: usize) -> Result<ColumnRef> {
```
Veeupup, line 80: thanks for column viewer, this complicated match could be avoid cc @sundy-li

PsiACE (approved):

common/functions/src/scalars/strings/substring.rs
sundy-li: Great!  But I suggest merging main into this pr to use the latest version of `Viewer `.

sundy-li: let viewer = i16::try_create_viewer(&column)?;

    assert_eq!(viewer.size(), 1024);
    assert!(!viewer.null_at(0));
    assert!(!viewer.null_at(99));

    for (idx, value) in viewer.iter().enumerate() {
        if viewer.null_at(idx) {
            assert_eq!(value, 0);
        } else {
            assert_eq!(value, 123i16);
        }
    }

sundy-li:
@mergify update

common/functions/src/scalars/strings/substring.rs
Veeupup: resolved

sundy-li:
Good job, almost -300  LOC

Issue #4083, opened 2022-02-09
Clean up MetaError


drmingdrmer:
`MetaError` includes several kinds of errors:
- Application Error: such as `IllegalUserInfoFormat`, which should return to the caller and won't affect metasrv itself.
- Internal flow control error: such as `ForwardToLeader`, which affects metasrv execution but is not necessary to return to the caller.
- Fatal error: `MetaStorageError`, which should shutdown metasrv at once.
- Others ...

Sub errors can be grouped into enum variants, removed, or replaced.

- Try to replace `UnknownError` `UnknownException` with a more specific error or remove them, if possible.
- Maybe it's a good idea to group application errors into one sub error because most of the time, we just pop such an error up and finally return it to the caller.
- `MetaStorageError` may have already included other **fatal** error.

```
pub enum MetaError {
    #[error(transparent)]
    ForwardToLeader(#[from] ForwardToLeader),

    #[error(transparent)]
    ChangeMembershipError(#[from] ChangeMembershipError),

    #[error(transparent)]
    ConnectionError(#[from] ConnectionError),

    #[error(transparent)]
    ErrorCode(#[from] SerializedError),

    #[error(transparent)]
    MetaStorageError(MetaStorageError),

    #[error("{0}")]
    UnknownError(String),

    #[error("{0}")]
    InvalidConfig(String),

    #[error("raft state present id={0}, can not create")]
    MetaStoreAlreadyExists(u64),

    #[error("raft state absent, can not open")]
    MetaStoreNotFound,

    #[error("serde_json error: {0}")]
    SerdeJsonError(String),

    #[error("{0}")]
    MetaStoreDamaged(String),

    #[error("{0}")]
    BadBytes(String),

    #[error("{0}")]
    LoadConfigError(String),

    #[error("{0}")]
    TLSConfigurationFailure(String),

    #[error("{0}")]
    StartMetaServiceError(String),

    #[error("{0}")]
    BadAddressFormat(String),

    #[error("{0}")]
    ConcurrentSnapshotInstall(String),

    #[error("{0}")]
    UnknownNode(String),

    #[error("{0}")]
    MetaServiceError(String),

    #[error("{0}")]
    CannotConnectNode(String),

    #[error("{0}")]
    IllegalRoleInfoFormat(String),

    #[error("{0}")]
    IllegalUserInfoFormat(String),

    #[error("{0}")]
    UnknownException(String),
}
```

PR #4106, opened 2022-02-11
Refactor: move network related errors to MetaNetworkError, and raft related errors to MetaRaftError

lichuang:

## Summary

Move network related errors to MetaNetworkError, and raft related errors to MetaRaftError,which is sub types of MetaError.

## Changelog

- Improvement

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4083

## Test Plan

Unit Tests

Stateless Tests


common/meta/types/src/meta_errors_into.rs
```diff
@@ -39,6 +41,18 @@ impl From<MetaError> for ErrorCode {
...
+        MetaError::MetaNetworkError(e)
+    }
+}
+
+impl From<MetaRaftError> for MetaError {
+    fn from(e: MetaRaftError) -> Self {
+        MetaError::MetaRaftError(e)
+    }
+}
+
```
drmingdrmer, line 55:
```
MetaError {
   MetaRaftError(#[from] MetaRaftError)
}
```
Will automatically build a impl of `From<>`

common/meta/types/src/meta_errors_into.rs
```diff
@@ -39,24 +38,12 @@ impl From<MetaError> for ErrorCode {
...
-    fn from(e: MetaStorageError) -> Self {
-        MetaError::MetaStorageError(e)
-    }
-}
-
 impl From<serde_json::Error> for MetaError {
     fn from(error: serde_json::Error) -> Self {
         MetaError::MetaStorageError(error.into())
     }
 }
```
drmingdrmer, line 46: If MetaStorageError is a superset of serde json Error, maybe MetaError itself does not need to include serde json Error.

If it is true, maybe we could fix it in next PR 🤔

common/meta/types/src/meta_network_errors.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Error, Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub enum MetaNetworkError {
+    #[error(transparent)]
+    ConnectionError(#[from] ConnectionError),
+
+    #[error("{0}")]
+    GetNodeAddrError(String),
+
+    #[error("{0}")]
+    UnknownNode(String),
```
drmingdrmer, line 30: Do these two errors caused by the same reason that a node is not found in the local store?

If it is true, we could merge these two into one.

common/meta/types/src/meta_raft_errors.rs
```diff
@@ -0,0 +1,54 @@
...
+    ForwardToLeader(#[from] ForwardToLeader),
+
+    #[error(transparent)]
+    ChangeMembershipError(#[from] ChangeMembershipError),
+
+    #[error("{0}")]
+    ConsistentReadError(String),
+
+    #[error("{0}")]
+    ClientWriteError(String),
```
drmingdrmer, line 36: It seems there is only one error will be wrapped in `ClientWriteError`: `openraft::Fatal`.
What about naming it with `RaftFatal` or just `Fatal`?

BTW to wrap an external error, using `AnyError` with provide more info for debugging than using a `String`.

metasrv/src/meta_service/raftmeta.rs
```diff
@@ -201,7 +203,14 @@ impl MetaNode {
...
 
         let addr_str = addr.to_string();
-        let addr = addr.parse::<std::net::SocketAddr>()?;
+        let ret = addr.parse::<std::net::SocketAddr>();
+        let addr = match ret {
+            Ok(addr) => addr,
+            Err(e) => {
+                let err: MetaNetworkError = e.into();
+                return Err(err.into());
+            }
```
drmingdrmer, line 212: If there is a double `into()`, maybe the method return value type is too general.

IMHO, a `MetaNetworkError` may be enough for this `start_grpc()` method.

drmingdrmer (commented):
It's became more clear! :DDD Love it.

drmingdrmer (approved):
LGTM:DDD

ariesdevil (approved):

PR #4109, opened 2022-02-11
[meta] refactor: cleanup MetaStorageError

drmingdrmer:

##### [meta] refactor: cleanup MetaStorageError
- MetaStorageError::SledError() use AnyError to express the causing error.
  This way the error message includes more context information and it
  get rid of some manually error convertion job, by utilizing `From`.

- Feature: add trait `WithContext` to create a wrapper error that
  includes additional context info. See: `WithContext::context()`

- Add MetaStorageError::SnapshotError, as a wrapper all snapshot errors.
  Snapshot error is a **storage** level error and it does not have to be
  a sub error of the top-level error `MetaError`.

- Remove MetaStorageError::TransactionAbort, because all abortion is
  caused by MetaStorageError::AppError.

## Changelog

- Improvement

## Related Issues

### DIFF
--- a/common/meta/sled-store/src/sled_tree.rs
+++ b/common/meta/sled-store/src/sled_tree.rs
@@ -18,9 +18,9 @@ use std::ops::Bound;
 use std::ops::Deref;
 use std::ops::RangeBounds;
 
+use common_meta_types::error_context::WithContext;
 use common_meta_types::MetaStorageError;
 use common_meta_types::MetaStorageResult;
-use common_meta_types::ToMetaStorageError;
 use common_tracing::tracing;
 use sled::transaction::ConflictableTransactionError;
 use sled::transaction::TransactionResult;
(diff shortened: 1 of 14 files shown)


common/meta/sled-store/src/sled_tree.rs
```diff
@@ -136,15 +134,15 @@ impl SledTree {
...
-                        MetaStorageError::TransactionAbort(_e) => {
-                            Err(ConflictableTransactionError::Abort(meta_sto_err))
-                        }
                         MetaStorageError::TransactionConflict => {
                             Err(ConflictableTransactionError::Conflict)
                         }
                         MetaStorageError::AppError(_app_err) => {
                             Err(ConflictableTransactionError::Abort(meta_sto_err))
                         }
+                        MetaStorageError::SnapshotError(_e) => {
```
ariesdevil, line 143: Why not just use `_` instead of  `_e` ?

drmingdrmer: Hmm... It's a good question that deserves a great answer.

common/meta/sled-store/src/sled_tree.rs
```diff
@@ -412,16 +390,13 @@ impl SledTree {
     {
         let mut res = vec![];
 
-        let range_mes = self.range_message::<KV, _>(&range);
+        let mes = || format!("range_get: {}", self.range_message::<KV, _>(&range));
```
BohuTANG, line 393: `||` used for? Likes a new postision :)

BohuTANG: I see `item.context(mes)?;`

BohuTANG:
LGTM

PR #4130, opened 2022-02-13
PLANNER-4091: Add others show plan

### LINKED ISSUE #4091: Move the sql query rewritten from sql to interpreter
**Summary**

There are some sql rewritten in the sql crate, such as:
```
show databases;
```
It will be rewritten to the follows in sql mode:
```
SELECT name AS Database FROM system.databases ORDER BY name
```

Codes:
https://github.com/datafuselabs/databend/blob/18114cf60191b13eed0d7e3232682f20845e6759/query/src/sql/statements/statement_show_databases.rs#L45

**Problem**

1. The sql mod must know the underlying logic, such as `show databases` should know which table needs to fetch, they are tied a bit tightly.
2. In the interpreter crate, we want to know what the planner exactly is, like `show databases` is a `SHOW` planner not a `SELECT` planner after rewritting, so we can check the access for management mode, now we allows all the `SELECT` query in management mode:
https://github.com/datafuselabs/databend/blob/18114cf60191b13eed0d7e3232682f20845e6759/query/src/interpreters/access/management_mode_access.rs#L58

**Proposal**

1. `show databases` parse to the normally planner:
```
 ShowDatabasePlan {
   pub expr: String
 }
 ```

2. In the interpreter_show_databases.rs, we do the sql rewritte from `ShowDatabasePlan`

**Tasks**

- [x] show databases (#4121)
- [ ] show tables (#4126)
- [ ] show functions (#4130 )
- [ ] show users  (#4130 )
- [ ] show engines  (#4130 )
- [ ] show processlist  (#4130 )
- [ ] show settings  (#4130 )
- [ ] show metrics  (#4130 )

### PR DESCRIPTION
**Pull request by BohuTANG**

## Summary

Summary about this PR:

Change all follow `sql rewritten` to `planner`:
* show functions
* show users
* show engines
* show processlist
* show settings
* show metrics

**There is NO `SHOW` command rewritten into `SELECT` in sql parser NOW**

## Changelog

- Improvement

## Related Issues

Fixes #4091

## Test Plan

Unit Tests

Stateless Tests

### CODE REVIEW

File: common/planners/src/plan_node.rs [hunk 1]
```diff
@@ -92,14 +99,25 @@ pub enum PlanNode {
 
     // Insert.
     Insert(InsertPlan),
+
+    // Copy.
     Copy(CopyPlan),
-    Sink(SinkPlan),
+
+    // Show.
+    ShowDatabases(ShowDatabasesPlan),
```
sundy-li, line 107: Can we merge them into one show plan ?

This makes plan enum struct much more fatter.

BohuTANG: > This makes plan enum struct much more fatter.

For sql statement, it looks we should transform the statement to the deterministic plan, the plan will transformed to the deterministic interpreter, each interpreter will has his own privileges check.
`show databases` -- database level check, some no privileges database will not show here
`show tables` -- table level check

sundy-li: Can we apply the ` privileges check` inside the show plans?  `Show(ShowPlan)`

```
struct ShowPlan {
  Kind: Kind
}
```

BohuTANG: > Can we apply the ` privileges check` inside the show plans? `Show(ShowPlan)`
> 
> ```
> struct ShowPlan {
>   Kind: Kind
> }
> ```

Yes, sound a good idea.
The current implemation is the flat version of the kind.

Now, there are only 8 `show []`:
* show databases
* show tables
* show functions
* show users
* show stages -- TODO
* show engines -- management mode disabled
* show metrics -- management mode disabled
* show processlist -- management mode disabled
If we bring in the `kind`, it looks we back to the `match kind` and do every query rewritten in one place.

common/planners/src/plan_node_rewriter.rs
```diff
@@ -432,6 +446,30 @@ pub trait PlanRewriter: Sized {
         Ok(PlanNode::ShowDatabases(plan.clone()))
     }
 
+    fn rewrite_show_engines(&mut self, plan: &ShowEnginesPlan) -> Result<PlanNode> {
+        Ok(PlanNode::ShowEngines(plan.clone()))
+    }
+
+    fn rewrite_show_functions(&mut self, plan: &ShowFunctionsPlan) -> Result<PlanNode> {
```
sundy-li, line 453: Fatter version may cause the code bloat like these.

BohuTANG: We don't have many `show` command, and add onme show command is easy, I think it's ok.

For the plan_node_rewriter, [common/planners/src/plan_node_rewriter.rs](https://github.com/datafuselabs/databend/pull/4130/files/2fd62a917f4e726ccf25c1fc200cce59712d2cfd#diff-5c0a6979f46d0a3071e2d231fec9dad84867b12c7e690f972954408d7adf594e) there are many empty rewrite, maybe we should do something in it.

sundy-li (approved):

PR #4131, opened 2022-02-13
Add hash function city64WithSeed

junli1026:

## Summary

Add hash function cityhash64_with_seed, mostly for bloom_filter.

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/hashes/city64_with_seed.rs
```diff
@@ -0,0 +1,170 @@
...
+        &self,
+        columns: &common_datavalues2::ColumnsWithField,
+        _input_rows: usize,
+    ) -> Result<common_datavalues2::ColumnRef> {
+        let column = columns[0].column();
+        let physical_data_type = columns[0].data_type().data_type_id().to_physical_type();
+
+        let const_col: Result<&ConstColumn> = Series::check_get(columns[1].column());
+
+        if let Ok(col) = const_col {
```
sundy-li, line 135: I think we can use `ScalarBinaryExpression` which covers the constant cases. And we do not need `cast_column_field` if we add bound trait: `AsPrimitive<u64>`

junli1026: For the constant cast, we can keep re-using the same hasher object because they have the same seed; non-constant case will need to create hasher per row.
ScalarBinaryExpression is good for this case, but I didn't figure out a way to avoid creating duplicate hasher.

sundy-li:
LGTM

PR #4139, opened 2022-02-14
refactor(functions/maths): migrate ceil/floor/sqrt to function2

PsiACE:

## Summary

- [x] ceil/ceiling
- [x] floor
- [x] sqrt

## Changelog

- Improvement

## Related Issues

Part of #3972 

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/maths/sqrt.rs
```diff
@@ -14,64 +14,58 @@
...
-            Ok(DataType::Float64)
-        } else {
-            Err(ErrorCode::IllegalDataType(format!(
-                "Expected numeric types, but got {}",
-                args[0]
-            )))
-        }?;
-
-        // sqrt on negative returns null, so nullable should be true.
-        Ok(DataTypeAndNullable::create(&data_type, true))
```
PsiACE, line 60: It looks like `ScalarUnaryExpression` cannot handle `Option<f64>` directly, for now I let sqrt on negative returns `f64::NAN`.

cc @sundy-li , should we continue to return `NULL`?

sundy-li: > now I let sqrt on negative returns f64::NAN

It's better now.

sundy-li (approved):

BohuTANG:
/lgtm

:rocket:

PR #4146, opened 2022-02-14
Datavalues2: number and round function

### LINKED ISSUE #3970: datavalues2 && date functions

LiuYuHui:

## Summary

Migrate some date functions into datavalues2

## Changelog

- Improvement

## Related Issues

Fixes #3970 

## Test Plan

Unit Tests

Stateless Tests


Xuanwo:
@LiuYuHui Sorry for the wrong robot behavior which sent you a lot of spam messages, we will fix it as soon as possible :cry:

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -155,332 +120,259 @@ impl NumberResultFunction<u16> for ToStartOfISOYear {
...
+            TypeID::Date16 => {
+                if column.is_const() {
+                    let column: &ConstColumn = Series::check_get(column)?;
+                    let date_time = Utc.timestamp(column.get(0).as_i64()? * 24 * 3600, 0_u32);
+                    let column = PrimitiveColumn::new_from_vec(vec![T::to_number(date_time)]).arc();
+                    Ok(Arc::new(ConstColumn::new(column, input_rows)).arc())
+                }else {
+                    let mut array = Vec::with_capacity(input_rows);
+                    for i in 0..column.len() {
+                        let v = column.get(i).as_i64()?;
```
sundy-li, line 337: `column.get(index)` is virtual call inside the loop.

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -155,332 +120,259 @@ impl NumberResultFunction<u16> for ToStartOfISOYear {
...
+    fn eval(
+        &self,
+        columns: &common_datavalues2::ColumnsWithField,
+        input_rows: usize,
+    ) -> Result<common_datavalues2::ColumnRef> {
+        let type_id = columns[0].field().data_type().data_type_id();
+        let column = columns[0].column();
+        let number_array= match type_id {
+            TypeID::Date16 => {
+                if column.is_const() {
```
sundy-li, line 329: Do need to care about const case, it's covered in `Function2Adaptor`.

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -155,332 +120,259 @@ impl NumberResultFunction<u16> for ToStartOfISOYear {
...
-                        .to_array()?
-                        .u32()?
-                        .apply_cast_numeric(|v| {
-                            let date_time = Utc.timestamp(v as i64, 0_u32);
-                            T::to_number(date_time)
-                        }
-                        );
-                    Ok(result.into())
+            },
+            TypeID::DateTime32 => {
```
sundy-li, line 360: Better to use unary function, eg:
```
match type_id{
  Date16 => {
             let unary = ScalarUnaryExpression::<u16, R, _>::new(T::to_number);
             let col = unary.eval(columns[0].column())?;
             Ok(col.arc())
  }

}
```

sundy-li:
BTW, type `DateTime64` is mapping to `i64` instead of `u64` now after merged #4096.

LiuYuHui:
Hi @sundy-li, currently, I am confused by this two errors, I have no idea how to fix it.

1. The first error is from unit test, it's related to monotonicity, error message is 

> ---- plan_expression_monotonicity::test_dates_function stdout ----
thread 'plan_expression_monotonicity::test_dates_function' panicked at 'assertion failed: `(left == right)`
  left: `false`,
right: `true`: f(z) = toSecond(z) is_monotonic', common/planners/tests/it/plan_expression_monotonicity.rs:82:5
2. The second error is from stateless test, it's mainly related to toStartOf-* function, the expected output should be '2021-01-01', but the real output is '18628', it seems like about datatype?

Could you provide some pointers to me? thanks a lot :D

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -489,27 +358,25 @@ where
...
         if args[0].left.is_none() || args[0].right.is_none() {
-            return Ok(Monotonicity::default());
+            return Ok(Monotonicity2::default());
         }
 
-        let left_val = func.eval(&[args[0].left.clone().unwrap()], 1)?.try_get(0)?;
-        let right_val = func
-            .eval(&[args[0].right.clone().unwrap()], 1)?
-            .try_get(0)?;
+        let left_val = func.eval(&[args[0].left.clone().unwrap()], 1)?.get(0);
```
sundy-li, line 372: need to wrap func .

```
let func = Function2Adapter::create(func);
```

sundy-li:
> The second error is from stateless test, it's mainly related to toStartOf-* function, the expected output should be '2021-01-01', but the real output is '18628', it seems like about datatype?

We need to modify `NumberFunction` or introduce `DateTimeFunction` to adjust `ToStartOf*`,
because :

```
  fn return_type(
        &self,
        _args: &[&common_datavalues2::DataTypePtr],
    ) -> Result<common_datavalues2::DataTypePtr> {
        Ok(R::to_data_type())
    }
```

It returns number type instead of `DateTime32::arc(None)`.

LiuYuHui:
Thanks a lot @sundy-li, I have changed the code, PTAL :D

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -103,6 +108,10 @@ impl NumberOperator<u16> for ToStartOfYear {
         let end: DateTime<Utc> = Utc.ymd(value.year(), 1, 1).and_hms(0, 0, 0);
         get_day(end) as u16
     }
+
+    fn return_type() -> Option<common_datavalues2::DataTypePtr> {
+        Some(Date32Type::arc())
```
sundy-li, line 113: Date16Type ?

common/functions/src/scalars/dates/number_function.rs
```diff
@@ -133,6 +146,10 @@ impl NumberOperator<u16> for ToStartOfQuarter {
         let date = Utc.ymd(value.year(), new_month, 1).and_hms(0, 0, 0);
         get_day(date) as u16
     }
+
+    fn return_type() -> Option<common_datavalues2::DataTypePtr> {
+        Some(Date32Type::arc())
```
sundy-li, line 151: .

common/functions/src/scalars/dates/number_function.rs
LiuYuHui: Fixed

common/functions/src/scalars/dates/number_function.rs
LiuYuHui: Fixed

sundy-li (approved):
Looks great to me.

PR #4147, opened 2022-02-14
[metasrv] refactor: MetaError: add AppError and Fatal error, remove ErrorCode from MetaError

drmingdrmer:

## Summary

[metasrv] refactor: MetaError: add AppError and Fatal error, remove ErrorCode from MetaError

ErrorCode is convenient but hard to manage.
Using ErrorCode everywhere means every mod has to carefully deal with all kind of
errors.

A mod should only deal with a minimal set of errors.

- MetaAPI and KVApi returns MetaError instead of ErrorCode.
  MetaError includes more error variants than just application errors(`AppError`),
  such as network error or raft error such as ForwardToLeader etc.

  The caller is free to just treat all these non-applicationn error as
  unhandleable fatal error.

  To extract application error to the old fashion ErrorCode, just `ErrorCode::from(meta_err)`,
  application error will be mapped to ErrorCode.

- Some error types in return values are narrowed down to a smaller set by replacing ErrorCode
  with more specific errors.

  Such as when serializing request, only a serde_json::Error will be
  emitted. In such cases, we do not need to use ErrorCode.

- ErrorCode returned from a dependency mod is mapped to MetaError one by
  one by its code number.

- MetaError is serializable thus it does not have to depend on other
  type such as SerializedError to transmit an error form peer to peer.

- Some error messages has changed.

- `ErrorCode` is removed from `MetaError` and some other type specific
  errors are added: `SerdeError`, `EncodeError`, `AppError` and `Fatal`.

- If an error is caused by another error, try to use `AnyError` as a
  sub error wrapper to keep as much info as possible.


ariesdevil (approved):
LGTM

Issue #4151, opened 2022-02-14
Let DNSResolver returns structured errors


drmingdrmer:
Using ErrorCode in a mod forces all depending mods to use ErrorCode as an error return value.
It is sometimes harmful to a mod that needs an error type with more than one variant, or to a mod that does not need to deal with every ErrorCode.

To relieve the depending mod from this trouble, introduce a typed error for DNSResolver and impl a `From<DNSError>` for ErrorCode to let the old code run smoothly.

PR #4156, opened 2022-02-15
[grpc] ConnectionFactory: returns structured error so that a depending mod does not need to deal with every error defined in ErrorCode

drmingdrmer:

##### [grpc] ConnectionFactory: returns structured error so that a depending mod does not need to deal with every error defined in ErrorCode
- Remove conversion from ErrorCode to MetaError, which is a potential bug maker:
  It can not find out missing matching branch at compile time.

- Remove MetaNetworkError::CannotConnectNode, it should be a sub error of ConnectionError

- Remove MetaError:TLSConfigurationFailure, it should be a sub error of MetaNetworkError

- Return Infallible if a method does not return any error

- fix: #4151

## Related Issues

### DIFF
--- a/common/meta/grpc/src/grpc_client.rs
+++ b/common/meta/grpc/src/grpc_client.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::convert::Infallible;
 use std::fmt::Debug;
 use std::sync::Arc;
 use std::time::Duration;
@@ -22,7 +23,9 @@ use common_containers::ItemManager;
 use common_containers::Pool;
 use common_exception::Result;
 use common_grpc::ConnectionFactory;
+use common_grpc::GrpcConnectionError;
 use common_grpc::RpcClientTlsConfig;
+use common_meta_types::anyerror::AnyError;
 use common_meta_types::protobuf::meta_service_client::MetaServiceClient;
 use common_meta_types::protobuf::HandshakeRequest;
 use common_meta_types::protobuf::RaftReply;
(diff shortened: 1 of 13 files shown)


lichuang (approved):

common/meta/grpc/src/grpc_client.rs
```diff
@@ -87,7 +101,9 @@ pub struct MetaGrpcClient {
 const AUTH_TOKEN_KEY: &str = "auth-token-bin";
 
 impl MetaGrpcClient {
-    pub async fn try_new(conf: &MetaGrpcClientConf) -> Result<MetaGrpcClient> {
+    pub async fn try_new(
+        conf: &MetaGrpcClientConf,
+    ) -> std::result::Result<MetaGrpcClient, Infallible> {
```
ariesdevil, line 106: Why not just rename `try_new` to `new` and change the return type from `Result<T, Infallible>` to `T` if a method does not return any error.

query/src/catalogs/backends/meta_backend.rs
```diff
@@ -70,7 +70,10 @@ impl MetaBackend {
         F: Send + Sync + 'static,
         T: Send + Sync + 'static,
     {
-        let cli = self.meta_api_provider.try_get_meta_client().await?;
+        let res: Result<_, std::convert::Infallible> =
+            self.meta_api_provider.try_get_meta_client().await;
+        let cli = res.unwrap();
```
ariesdevil, line 75: If we do the above return type modification, we do not need unwrap here.

common/meta/grpc/src/grpc_client.rs
drmingdrmer: I'm not very sure about whether in the future an error will return.
IMHO, it is very likely to be.

query/src/catalogs/backends/meta_backend.rs
drmingdrmer: There are two different ways to make codes clean:
- To minimize the logic involved,
- or to follow a universal code style, e.g. a method that creates an instance always returns a `Result`.

In this case, either to minimize logic or to keep the return type of every `new_xxx()` method a `Result`. Both feel good to me.

It does not bother me. Then I'll take your word, sir:D

ariesdevil: Convinced by you :)

PR #4159, opened 2022-02-15
QUERY-4157: Improve storage accessor initate logic

### LINKED ISSUE #4157: query: Improve storage accessor initate logic
## Backgroud

We are using `get_storage_accessor` to get a storage accessor for underlying storage operation. However, unlike indicated as the name `get_xxx`, it's a heavy operation that will check and init storage clients.

Even worse, databend-query have multiple `query`s in the same `session`, every time we call `get_storage_accessor`, we are initiating the storage client.

## Changes

So it's better for us to do the following enhancement:

- [ ] Split `get_storage_accessor` into `init_storage_accessor` and `get_storage_accessor`
  - `init_storage_accessor` is used to init a storage accessor
  - `get_storage_accessor` is used to get (a.k.a, `clone`) a storage accessor from query ctx.
- [ ] Move the `init_storage_accessor` logic to databend-query app init stage, so that we can share the same client between different sessions.

## Benefits

After all those changes have been done, we will:

- Reduce the whole init times. (For the best, only once during databend-query startup)
- Move the init logic out of the critical business path. 

## Quick Notes

- AWS SDK `load_env` will try to connect their metadata services with 1s timeout and 3 timers retry. This improvement will fix https://github.com/datafuselabs/opendal/issues/16

BohuTANG:

## Summary

Summary about this PR:
1. **Move the storage accessor init to the session manager**
2. **Re-order some private functions order**

## Changelog

- Improvement
- Performance Improvement

## Related Issues

Fixes #4157

## Test Plan

Unit Tests

Stateless Tests


query/src/sessions/session_mgr.rs
```diff
@@ -51,12 +57,14 @@ pub struct SessionManager {
     pub(in crate::sessions) max_sessions: usize,
     pub(in crate::sessions) active_sessions: Arc<RwLock<HashMap<String, Arc<Session>>>>,
     pub(in crate::sessions) storage_cache_manager: Arc<CacheManager>,
+    storage_accessor: Arc<dyn Accessor>,
```
Xuanwo, line 60: It's better to host an `Operator` here.

BohuTANG: We can do that here, reasons:
operator needs storage ctx, and the storage ctx is binding to query context for get metrics for each query.

BohuTANG: The ABI is `pub async fn get_storage_accessor(&self) -> **Result<Operator>**`, and `Operator::new(accessor)` almost no cost here, so I put them here.

BohuTANG: Update:
Change accessor to operator.

BohuTANG:
@mergify update

Xuanwo (approved):

BohuTANG:
Developing / test_stateless_cluster (ubuntu-latest, x86_64-unknown-linux-gnu) failure is not related to this PR.
I have check it on my local pc(ubuntu), it works well.
Let's merge!

PR #4160, opened 2022-02-15
refactor if to get better performence

### LINKED ISSUE #4120: Feature:   Refactor  `if` function to use macros instead of column viewer for better performance
**Summary**

Description for this feature.

Function `if` is widely used, so we should take the performance seriously.

`if(a,b,c)`

1. if `a` is constant, then result is Column b or c, this is the fastest path.
2. if `b` is constant, c is constant or not ...
3. if `b` is nullable, c is nullable or not...

Veeupup:

## Summary

refactor if to get better performence:
* improve performance when condition column when cond is const or nullable or null column
* improve performance when lhs or rhs is const
* improve performance when lhs and rhs are both scalar column

## Changelog

- New Feature

## Related Issues

Fixes #4120 

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/conditionals/if.rs
```diff
@@ -41,102 +42,330 @@ impl IfFunction {
...
+                            }
+                        });
+                        let col = <$T as Scalar>::ColumnType::from_iterator(iter);
+                        return Ok(col.arc());
+                    }
+                };
+            }
+
+            match type_id {
+                TypeID::Boolean => {
```
sundy-li, line 211: There is binding from `Scalar::ColumnType --->   Column`

Veeupup: resolved

sundy-li (approved):

BohuTANG:
@merfigy update

sundy-li:
SQL: `SELECT sum( if (number > number * 3 , number, number + 3 ) ) FROM numbers_mt(100000000000) ;`
Before: 30.35 GB/s
After: 47.10 GB/s.

BohuTANG:
stateless test failed, I have re-run the job but not work, cc @Veeupup

Veeupup:
> stateless test failed, I have re-run the job but not work, cc @Veeupup

seems wired, let me check

PsiACE:
I tried to remove the changes except `if` locally and it passed the stateless test normally

Veeupup:
> I tried to remove the changes except `if` locally and it passed the stateless test normally

@PsiACE  yes, something buggy happened in nested column limit.. it will be resolved in another pr

PsiACE (approved):

Issue #4009, opened 2022-01-27
bug: databend-bot usually give duplicate comments in PR


Veeupup:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

newest

### What's Wrong?

databend-bot sometimes gives duplicate comments in PR
![image](https://user-images.githubusercontent.com/26339429/151399936-f9ecc1f0-a255-4259-9772-2a842edf457e.png)

### What You Expected?

databend-bot give only one unique comment

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Xuanwo:
Occurred at https://github.com/datafuselabs/databend/pull/4146 too, I will fix it.

/assignme

PR #4162, opened 2022-02-15
ci: Use mergify for PR labels

Xuanwo:

## Changelog

- Build/Testing/CI

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4009

## Test Plan

Unit Tests

Stateless Tests


Xuanwo:
For preview: https://github.com/Xuanwo/mergify-test/pull/1

.github/mergify.yml
```diff
@@ -45,3 +43,71 @@ pull_request_rules:
...
+        add:
+          - pr-bugfix
+  - name: label on Improvement
+    conditions:
+      - body~=Improvement
+    actions:
+      label:
+        add:
+          - pr-improvement
+  - name: label on Improvement
```
BohuTANG, line 79: It seems this label will never be set since the above:
```
  conditions:
      - body~=Improvement
  ```

Xuanwo: They will be applied at the same time. Maybe we can adopt more strict rules for the behavior.

Xuanwo: image

Xuanwo: Fixed by checking `- Improvement` instead of `Improvement`

Xuanwo:
Take https://github.com/Xuanwo/mergify-test/pull/2 for preview.

Issue #3931, opened 2022-01-21
integrate new processor framework


zhang2014:
- [x] integrate standalone select query with new processor
    - [x] integrate numbers table with new processor
    - [x] integrate systems table with new processor
    - [x] #4345
- [x] #4471
- [x] #4799

xudong963:
Hi @zhang2014 , I've read through the processor code in general, If you don't mind, I'll take the final task (integrate insert query with new processor). I think the last task can be developed in parallel with #4471, right?

sundy-li:
Actually, I just want to start to work on it(not ready yet).   You can take the task~

BohuTANG:
This is ready, let's close it.

PR #4166, opened 2022-02-16
Integrate system tables with new processor

zhang2014:

## Summary

Integrate system tables with new processor

## Changelog

- New Feature

## Related Issues

part #3931 

## Test Plan

Unit Tests

Stateless Tests


zhang2014:
after this pr. all system tables can work in the new processor framework.

sundy-li (approved):

query/src/pipelines/new/executor/pipeline_runtime_executor.rs
```diff
@@ -24,48 +24,60 @@ use crate::pipelines::new::pipeline::NewPipeline;
...
+                        }
                     }
                 }
-            });
+            }));
         }
+
+        while let Some(join_handle) = threads.pop() {
+            // flatten error.
+            match join_handle.await {
```
sundy-li, line 71: However, Rust futures won't do any work until they're actively .await !

BohuTANG:
> after this pr. all system tables can work in the new processor framework.
Cool!
Seems many of the interpreter logic should be moved to a transform(or processor), then we can track and profile them under the new processor framework.
For example: insert(... select) and delete etc, then we can visualize the profile easily.

PR #4173, opened 2022-02-16
refactor(functions/maths): migrate abs/degress/radians to function2

PsiACE:

## Summary

- [x] refactor(maths/angle): migrate degress/radians to function2
- [x] refactor(maths/abs): migrate abs to function2

## Changelog

- Improvement

## Related Issues

Part of #3972 

## Test Plan

Unit Tests

Stateless Tests


common/functions/src/scalars/maths/abs.rs
```diff
@@ -47,87 +47,69 @@ impl AbsFunction {
...
+            TypeID::Int32 => i32::to_data_type(),
+            TypeID::Int64 => i64::to_data_type(),
+            TypeID::UInt8 => u8::to_data_type(),
+            TypeID::UInt16 => u16::to_data_type(),
+            TypeID::UInt32 => u32::to_data_type(),
+            TypeID::UInt64 => u64::to_data_type(),
+            TypeID::Float32 => f32::to_data_type(),
+            TypeID::Float64 => f64::to_data_type(),
+            _ => unreachable!(),
+        };
```
PsiACE, line 76: Unlike before, the results are no longer cast here. I think it is reasonable behavior for the result type to be the same as the parameter type.

Issue #4098, opened 2022-02-10
Add test: MetaAPI: add table operation on absent db.


drmingdrmer:
A `UnknownDatabase` error is expected.

`common/meta/api/src/meta_api_test_suite.rs`

ClSlaid:
BTW, I found a strange `-`(`u+2010` instead of `u+002d`) in this [line](https://github.com/datafuselabs/databend/blob/cb1557caca976ae10d55ef848275b5540daf609d/common/meta/api/src/meta_api_test_suite.rs#L637), is this intention?
![图片](https://user-images.githubusercontent.com/44747719/154252146-f6ee84b8-ac2f-461b-ac69-1a54efaa3056.png)

drmingdrmer:
Good catch!
Of cause not intentional. :DDD

PR #4175, opened 2022-02-16
[Test]MetaAPI: add table opration on absent DB

ClSlaid:

## Summary

Add tests for MetaAPI in `common/meta/api/src/meta_api_test_suite.rs`.

## Changelog

- Added tests

## Related Issues

issue #4098 

## Test Plan

Unit Tests

Stateless Tests


ClSlaid:
Hope the CI could run tests for me...
My little laptop cannot hold such a heavy burden with its 14GB RAM.

ariesdevil (approved):

BohuTANG:
failures:
[784](https://github.com/datafuselabs/databend/runs/5216628993?check_suite_focus=true#step:3:784)

[785](https://github.com/datafuselabs/databend/runs/5216628993?check_suite_focus=true#step:3:785)
---- meta_api_impl::test_meta_embedded_table_create_get_drop stdout ----
[786](https://github.com/datafuselabs/databend/runs/5216628993?check_suite_focus=true#step:3:786)
thread 'meta_api_impl::test_meta_embedded_table_create_get_drop' panicked at 'assertion failed: `(left == right)`
[787](https://github.com/datafuselabs/databend/runs/5216628993?check_suite_focus=true#step:3:787)
  left: `1003`,
[788](https://github.com/datafuselabs/databend/runs/5216628993?check_suite_focus=true#step:3:788)
 right: `1020`',

common/meta/api/src/meta_api_test_suite.rs
```diff
@@ -427,10 +428,57 @@ impl MetaApiTestSuite {
...
+
+            let got = mt.upsert_table_option(plan).await;
+            tracing::debug!("upsert table on unknow db got: {:?}", got);
+
+            assert!(got.is_err());
+            let err = got.unwrap_err();
+            let err = ErrorCode::from(err);
+
+            assert_eq!(unknown_database_code, err.code());
+        }
```
ClSlaid, line 461: I brainlessly wrote similar tests for every operation (`create`, `get`, `upsert` and `drop`). However, the `upsert` operation is seen to operate on tables by `table_id`, regardless of DBs.
So removing this test should solve the problem.

ClSlaid:
Strange, I can't reproduce locally on my laptop...

图片

drmingdrmer (approved):
Nice shot!

ClSlaid:
> Nice shot!

the problem is more complex than I think... It came out to be `MetaServiceError` instead.
I'll take a look tomorrow(this morning).

drmingdrmer:
> > Nice shot!
> 
> the problem is more complex than I think... After deletion, a `MetaServiceError` will be thrown from the drop-on-unknown DB test instead. The implementation of `create` and `drop` has a lot in common, but why `create` returns an `UnknowDataBase` while `drop` returns a `MetaServiceError`? This problem has only shown up in tests of `metasrv`, a further looking into is needed... I'll take a look tomorrow (this morning).

That should be incorrect error type that is used when a database is not found. Let me have a look.

ariesdevil:
> 



> > > Nice shot!
> > 
> > 
> > the problem is more complex than I think... After deletion, a `MetaServiceError` will be thrown from the drop-on-unknown DB test instead. The implementation of `create` and `drop` has a lot in common, but why `create` returns an `UnknowDataBase` while `drop` returns a `MetaServiceError`? This problem has only shown up in tests of `metasrv`, a further looking into is needed... I'll take a look tomorrow (this morning).
> 
> That should be incorrect error type that is used when a database is not found. Let me have a look.

Fixed at: [`d7d8f0b` (#4181)](https://github.com/datafuselabs/databend/pull/4181/commits/d7d8f0b2712238610365019cb53d57062b480470)

ariesdevil:
> > > Nice shot!
> > 
> > 
> > the problem is more complex than I think... After deletion, a `MetaServiceError` will be thrown from the drop-on-unknown DB test instead. The implementation of `create` and `drop` has a lot in common, but why `create` returns an `UnknowDataBase` while `drop` returns a `MetaServiceError`? This problem has only shown up in tests of `metasrv`, a further looking into is needed... I'll take a look tomorrow (this morning).
> 
> That should be incorrect error type that is used when a database is not found. Let me have a look.

@ClSlaid plz wait for #4182 merged then rebase it and I think all tests will be passed.

BohuTANG:
@mergify update

Issue #4137, opened 2022-02-14
Bug: metasrv should not resolve the address when initializing 


drmingdrmer:
Raft config `raft_api_host` can be a hostname that will be resolved by DNS when using it.
This makes it easy to deploy a raft node in k8s when the ip address of a container changes.

Currently, raft resolves the hostname and store the result ip address in its store and will never change.
This just invalidates the dynamic address mechanism.

Solution:
- Do not resolve address when initializing a node.
- Only resolve address for listening, i.e., in `MetaNode::start_grpc()` and when connecting, i.e., in `metasrv::*::Network::make_client()`

https://github.com/datafuselabs/databend/blob/86fab7e1e46ced7806a5a20669b45369f6ec6668/metasrv/src/meta_service/raftmeta.rs#L274-L288

ariesdevil:
> As I implemented those parts, I found that in kubernetes environment, our DNS resolver has limits(which generally the hardest part to debug...), I think we could refer to the dns resolver code in linkerd proxy for more robust solution [https://github.com/linkerd/linkerd2-proxy/tree/main/linkerd/dns](https://github.com/linkerd/linkerd2-proxy/tree/main/linkerd/dns?rgh-link-date=2022-02-14T16%3A44%3A23Z)

Let's make another issue for this.

PR #4199, opened 2022-02-18
Refactor: use listen addr when start grpc

lichuang:

## Summary

1. Refactor: use listen addr when start grpc.
2. resolve host when start_grpc
3. save advertise addr in cluster node config.
4. assign raft_{listen|advertise}_host in test config, so if something wrong in raft meta nodes communication we will catch bug in unit tests.

## Changelog

- Bug Fix

## Related Issues

Fixes [#issue](https://github.com/datafuselabs/databend/issues/4137)

## Test Plan

Unit Tests

Stateless Tests


metasrv/src/meta_service/raftmeta.rs
```diff
@@ -265,33 +292,22 @@ impl MetaNode {
         let sto = Arc::new(sto);
 
         let mut builder = MetaNode::builder(&config).sto(sto.clone());
+        // config.id only used for the first time
+        let self_node_id = if is_open { sto.id } else { config.id };
```
drmingdrmer, line 296: Sweet!

metasrv/src/meta_service/raftmeta.rs
```diff
@@ -265,33 +292,22 @@ impl MetaNode {
...
-                    .raft_api_addr()
-                    .await
-                    .map_err(|e| MetaError::Fatal(AnyError::new(&e)))?,
-            );
-        }
-
+        // use ip:port to start grpc listening
+        builder = builder
+            .node_id(self_node_id)
+            .addr(config.raft_api_listen_host_string());
```
drmingdrmer, line 301: What about just passing in a tuple `(hostname, port)` into `addr()`? So that you do not need to parse the `host:port` string again in `start_grpc()`.

common/meta/types/src/endpoint.rs
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt;
+
+use serde::Deserialize;
+use serde::Serialize;
+
+#[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq)]
+pub struct Endpoint {
```
ariesdevil, line 21: Why not just use [SocketAddr](https://doc.rust-lang.org/std/net/struct.SocketAddrV4.html)

drmingdrmer: Endpoint need to be serializable to be stored in raft log

drmingdrmer (approved):
LGTM :DDD

ariesdevil (approved):
Look great to me:)

PR #4212, opened 2022-02-21
bump arrow2 to main

### LINKED ISSUE #3746: Bump datafuse-extras/arrow2

youngsofun:

## Summary

- [x]  the first commit bump to arrow2 0.9.1 with parquet2 0.9, main changes

1. replace `RecordBatch` by `Chunk`
2. new traits for simd8 for compare ops

- [x] the second commit (todo) bump to HEAD of main, to use new API [FileWriter/FileStreamer](https://github.com/jorgecarleitao/arrow2/pull/795), based on which I am working on the implementation of encryption for parquet2.

is there any other consideration given test is passed? 

## Changelog

## Related Issues

Fixes #3746


youngsofun:
Some Note:

first, the new async reader interface [read_columns_many_async](https://github.com/jorgecarleitao/arrow2/blob/3d528c99589e96f0539de4c07b11843fa22f23ac/examples/parquet_read_async.rs)use reader factory has 2 improvements:
    1. read the whole column chunk in one `read_exact` instead of page by page
    2. allow reading column chunks in a row group 
    
however, current ParquetSource brings a reader with it, thus cannot use the function, so I wrote a similar version of `read_columns_many_async` for this situation.

maybe we can adopt the reader factory pattern later.
  
second, Some arrow2 functions return Chunk<Box<dyn Array>> while we need Chunk<Arc<dyn Array>>,  I wrote a translation function for it `
box_chunk_to_arc_chunk`. Any better ways?
    
however, current ParquetSource brings a reader with it, thus cannot use the function, so I wrote a similar version of `read_columns_many_async` for this situation.

maybe we can adopt the reader factory pattern later.
  
second, Some arrow2 functions return Chunk<Box<dyn Array>> while we need Chunk<Arc<dyn Array>>,  I wrote a translation function for it `
box_chunk_to_arc_chunk`. Any better ways?

youngsofun:
Some Note:

first, the new async reader interface [read_columns_many_async](https://github.com/jorgecarleitao/arrow2/blob/3d528c99589e96f0539de4c07b11843fa22f23ac/examples/parquet_read_async.rs)use reader factory has 2 improvements:
    1. read the whole column chunk in one `read_exact` instead of page by page
    2. allow reading column chunks in a row group 
    
however, current ParquetSource brings a reader with it, thus cannot use the function, so I wrote a similar version of `read_columns_many_async` for this situation.

maybe we can adopt the reader factory pattern later.
  
second, Some arrow2 functions return Chunk<Box<dyn Array>> while we need Chunk<Arc<dyn Array>>,  I wrote a translation function for it `
box_chunk_to_arc_chunk`. Any better ways?
    
however, current ParquetSource brings a reader with it, thus cannot use the function, so I wrote a similar version of `read_columns_many_async` for this situation.

maybe we can adopt the reader factory pattern later.
  
second, Some arrow2 functions return Chunk<Box<dyn Array>> while we need Chunk<Arc<dyn Array>>,  I wrote a translation function for it `
box_chunk_to_arc_chunk`. Any better ways?

dantengsky:
> second, Some arrow2 functions returnChunk<Box<dyn Array>> while we needChunk<Arc>, I wrote a translation function for it
box_chunk_to_arc_chunk`. Any better ways?

hi, there is a rough idea, FYI

https://github.com/dantengsky/fuse-query/commit/71e3fefd76054a34723569425d7dde5845ea23c0

youngsofun:
looks good to me，翻墙从中午一直挂，晚点再来改

common/arrow/src/parquet_read.rs
```diff
@@ -0,0 +1,81 @@
...
+
+async fn read_columns_async<'a, R: AsyncRead + AsyncSeek + Send + Unpin>(
+    reader: &mut R,
+    columns: &'a [ColumnChunkMetaData],
+    field_name: &str,
+) -> Result<Vec<(&'a ColumnChunkMetaData, Vec<u8>)>> {
+    let col_metas = get_field_columns(columns, field_name);
+    let mut cols = Vec::with_capacity(col_metas.len());
+    for meta in col_metas {
+        cols.push((meta, _read_single_column_async(reader, meta).await?))
```
sundy-li, line 58: How to make IO run parallel?

Xuanwo: How about accepting an `Operator` or `Object` here to create different readers for different columns?

youngsofun: my version of read_columns_many_async is not parallel. 

the one in arrow2 is  parallel  

https://github.com/jorgecarleitao/arrow2/blob/3d528c99589e96f0539de4c07b11843fa22f23ac/src/io/parquet/read/row_group.rs#L166

query/tests/it/interpreters/interpreter_factory_interceptor.rs
```diff
@@ -106,7 +106,7 @@ async fn test_interpreter_interceptor_for_insert() -> Result<()> {
...
             "| 1        | TestSession  | 8         | 0         | 0          | 0               | 0            | 0             | 0           | 0            | CreateTablePlan | create table t as select number from numbers_mt(1) | root     | UserQuota { max_cpu: 0, max_memory_in_bytes: 0, max_storage_in_bytes: 0 } |",
-            "| 2        | TestSession  | 8         | 1         | 8          | 0               | 1            | 1090          | 0           | 0            | CreateTablePlan | create table t as select number from numbers_mt(1) | root     | UserQuota { max_cpu: 0, max_memory_in_bytes: 0, max_storage_in_bytes: 0 } |",
+            "| 2        | TestSession  | 8         | 1         | 8          | 0               | 1            | 1330          | 0           | 0            | CreateTablePlan | create table t as select number from numbers_mt(1) | root     | UserQuota { max_cpu: 0, max_memory_in_bytes: 0, max_storage_in_bytes: 0 } |",
```
youngsofun, line 109: written_bytes changed from 1090 to 1330

not known why

query/src/storages/fuse/io/block_reader.rs
```diff
@@ -85,12 +86,20 @@ impl BlockReader {
...
         let read_buffer_size = self.read_buffer_size;
+        let parquet_fields = metadata.schema().fields();
 
+        // read_columns_many_async use field name to filter columns
         let fields_to_read = self
             .projection
             .clone()
             .into_iter()
-            .map(|idx| arrow_fields[idx].clone())
+            .map(|idx| {
```
youngsofun, line 96: the test `test_fuse_table_normal_case`
 failed because parquet is written with field names 'a', 'b' ...
while the block reader later read with field name 'id'

query/src/storages/fuse/io/block_reader.rs
```diff
@@ -88,67 +83,50 @@ impl BlockReader {
...
+                let origin = arrow_fields[idx].clone();
+                Field {
+                    name: parquet_fields[idx].name().to_string(),
+                    ..origin
+                }
+            })
+            .collect();
 
-        let stream = futures::stream::iter(cols).map(|(col_meta, idx)| {
+        let factory = || {
```
youngsofun, line 105: we use the arrow2 parallel read_columns_many_async here

common/arrow/src/parquet_read.rs
sundy-li: So let's export `_read_single_column_async` as pub, we can use it  in `block_reader` ?

query/tests/it/interpreters/interpreter_factory_interceptor.rs
sundy-li: I think it's related to `HashMap --> BTreeMap `.

query/src/storages/fuse/io/block_reader.rs
youngsofun: @Xuanwo  (data_accessor, path) ?

common/arrow/src/parquet_read.rs
youngsofun: no,  block_reader already used it, I mean maybe SourceFactory can accept something like (data_accessor, path).  even use factory  justlike arrow2, and  struct(data_accessor, path) an impl of it

query/src/storages/fuse/io/block_reader.rs
dantengsky: some thoughts here:

1. shall we control the number of futures that simultaneously read the data of columns?
2. about the `try_join_all` used in `read_columns_may_async`
    at least, "join_all will switch to the more powerful [FuturesOrdered](https://docs.rs/futures/0.3.21/futures/stream/struct.FuturesOrdered.html) for performance reasons" according to 
     https://docs.rs/futures/0.3.21/futures/future/fn.join_all.html#see-also
     not sure if it's worth bothering to replace it.
     
 3. read column chunk in one go might be a huge benefit (as @youngsofun already mentioned)

   now the `read_page_header` scattered read bytes in an in-memory `Cursor`
 https://github.com/jorgecarleitao/parquet2/blob/85d1f01597907c0cc30a234a6d6209c3d7ef17cf/src/read/page_iterator.rs#L65-L68
 
  maybe we can eliminate the `BufReader` used in the reading of column
    
    https://github.com/datafuselabs/databend/blob/b41d39b05dcf281db921377d8b027477d333f8b6/query/src/storages/fuse/io/block_reader.rs#L108-L111

sundy-li (approved):

query/src/storages/fuse/io/block_reader.rs
youngsofun: cursor is only for file metadata https://github.com/jorgecarleitao/parquet2/search?q=cursor

but we can eliminate the BufReader because "we execute exactly 1 seek and 1 read on them."  https://github.com/jorgecarleitao/arrow2/blob/3d528c99589e96f0539de4c07b11843fa22f23ac/examples/parquet_read_async.rs#L31

dantengsky: > cursor is only for file metadata

Oh, I mean the `Cursor` used in arrow2 `to_deserializer`

https://github.com/jorgecarleitao/arrow2/blob/3d528c99589e96f0539de4c07b11843fa22f23ac/src/io/parquet/read/row_group.rs#L173-L185

> pr https://github.com/datafuselabs/databend/pull/4230

:+1:

Issue #4218, opened 2022-02-22
bug: In K8S,if meta node use K8S pod host name as advertize host, the meta cannot join a cluster  success.


lichuang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

0.70

### What's Wrong?

bug: In K8S,if meta node use K8S pod host name as advertize host, the meta cannot join a cluster  success.

### What You Expected?

join cluster.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #4219, opened 2022-02-22
refactor join cluster process

lichuang:

## Summary

1. join cluster after start all service, this can make K8S readiness probe happy, thus the node host name can be resolved by DNS.
2. use node id as node name in add_node,instead of an empty string.

## Changelog

- Bug Fix

## Related Issues

Fixes   [4218](https://github.com/datafuselabs/databend/issues/4218)

## Test Plan

Unit Tests

Stateless Tests


metasrv/src/meta_service/raftmeta.rs
```diff
@@ -412,6 +412,47 @@ impl MetaNode {
         Ok(mn)
     }
 
+    pub async fn join_cluster(&self, conf: &RaftConfig) -> MetaResult<()> {
+        if conf.join.is_empty() {
+            return Ok(());
```
ariesdevil, line 417: Should we log a warning message when this config is empty?

metasrv/src/meta_service/raftmeta.rs
```diff
@@ -412,6 +412,47 @@ impl MetaNode {
...
+        // Try to join a cluster only when this node is just created.
+        // Joining a node with log has risk messing up the data in this node and in the target cluster.
+        if self.is_opened() {
+            return Ok(());
+        }
+
+        let addrs = &conf.join;
+        // Join cluster use advertise host instead of listen host
+        let raft_api_advertise_host_endpoint = conf.raft_api_advertise_host_endpoint();
+        #[allow(clippy::never_loop)]
```
ariesdevil, line 429: Why not just use `addrs[0]` ?

metasrv/src/meta_service/raftmeta.rs
drmingdrmer: I think the elaborated name of this method should be `try_to_join_cluster_if_cli_arg_join_is_specified()`.

`conf.join` being empty means the user did not pass `--join` arguments to metasrv. It's totally OK.

metasrv/src/meta_service/raftmeta.rs
drmingdrmer: It's an unfinished code snippet... There should be a retry loop... :((

drmingdrmer (approved):
Looks pretty good to me:DDD

ZhiHanZ (approved):
tested and works, thanks @lichuang

BohuTANG (approved):

PR #4227, opened 2022-02-23
Move passthrough_null to FunctionFeatures

zhyass:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/functions/src/scalars/arithmetics/arithmetic_div.rs
+++ b/common/functions/src/scalars/arithmetics/arithmetic_div.rs
@@ -57,7 +57,6 @@ impl ArithmeticDivFunction {
             FunctionFeatures::default()
                 .deterministic()
                 .monotonicity()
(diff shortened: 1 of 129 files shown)


common/functions/src/scalars/arithmetics/arithmetic_div.rs
```diff
@@ -57,6 +57,7 @@ impl ArithmeticDivFunction {
             FunctionFeatures::default()
                 .deterministic()
                 .monotonicity()
+                .passthrough_null()
```
sundy-li, line 60: why not make this default ?

zhyass: > why not make this default ?

Fixed

junli1026 (approved):

PR #4228, opened 2022-02-23
Feature: Delete From

dantengsky:

## Summary

WIP: Statement "delete from"

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/operations/delete_from.rs
```diff
@@ -0,0 +1,155 @@
...
+                        } else {
+                            // Maybe there are rows should be deleted (false-positive of pruner)
+                            // use `delete_rows` to do the real data filtering.
+                            let new_segment = self.delete_rows(ctx.clone(), blocks, &expr).await?;
+                            segments_new.push(new_segment);
+                        }
+                    }
+
+                    let new_snapshot = self.merge_segments(&segments_kept, &segments_new);
+                    self.commit_deletion(new_snapshot).await?
```
sundy-li, line 74: So `deletion` will block `inserts` & `updates`?

dantengsky: `inserts` does not block other mutations. In this pr, a rather simple rule will be applied: when there are write-write conflicts,  first-committer-win.

sundy-li: So continuous small inserts will cause the heavy deletion retry ...

dantengsky:
transfered to PR #5691

query/src/storages/fuse/operations/delete_from.rs
dantengsky: > So continuous small inserts will cause the heavy deletion retry ...

er... too awkward... to catch this message months later. 

- yes, OCC implies the possibility of retrying
- deletion might be heavy, but the retrying of deletion might not be

PR #4203, opened 2022-02-18
opendal: Bump to v0.0.5 to address performance regression and much UE enhancement

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

This PR intends to adapt the opendal API changes, Fix https://github.com/datafuselabs/databend/issues/4197

More details could be found in https://github.com/datafuselabs/opendal/pull/35

## Changelog

- Performance Improvement

## Test Plan

Unit Tests

Stateless Tests

PR #4230, opened 2022-02-23
remove buffer in block_reader of fuse store

youngsofun:

## Summary

do not need buffer since  "we execute exactly 1 seek and 1 read on them."
https://github.com/jorgecarleitao/arrow2/blob/3d528c99589e96f0539de4c07b11843fa22f23ac/examples/parquet_read_async.rs#L31

a buffer with column chunk length is allocated each time.
https://github.com/jorgecarleitao/arrow2/blob/3d528c99589e96f0539de4c07b11843fa22f23ac/src/io/parquet/read/row_group.rs#L126

thanks to @dantengsky 
https://github.com/datafuselabs/databend/pull/4212#discussion_r812863275

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/storages/fuse/operations/read.rs
+++ b/query/src/storages/fuse/operations/read.rs
@@ -63,7 +63,6 @@ impl FuseTable {
 
         let part_stream = futures::stream::iter(iter);
 
-        let read_buffer_size = ctx.get_settings().get_storage_read_buffer_size()?;
         let stream = part_stream
             .map(move |part| {
                 let da = operator.clone();
@@ -81,7 +80,6 @@ impl FuseTable {
                         table_schema,
                         projection,
                         part_len,
-                        read_buffer_size,
                         reader,
                     );
                     block_reader.read().await.map_err(|e| {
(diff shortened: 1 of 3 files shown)


youngsofun:
@Xuanwo ok, this is a small pr, I will wait for you to merge first

BohuTANG:
> Hi, as #4203 is going to merge, how about wait a bit so that we don't need to resolve the conflict?

Merged!

query/src/storages/fuse/operations/read.rs
```diff
@@ -63,7 +63,6 @@ impl FuseTable {
 
         let part_stream = futures::stream::iter(iter);
 
-        let read_buffer_size = ctx.get_settings().get_storage_read_buffer_size()?;
```
BohuTANG, line 66: Does ctx.get_storage_read_buffer_size no used anymore?

youngsofun: yes, still in use

sundy-li (approved):

BohuTANG:
These codes also need clean?
https://github.com/datafuselabs/databend/blob/f9971bdf335333ffc2253b60b0842b2a3c8ca6cc/query/src/storages/fuse/io/meta_readers.rs#L182-L184

cc @youngsofun @dantengsky @Xuanwo

youngsofun:
@BohuTANG not needed for the current parquet2 implementation,   but this is function is also used to read our JSON metafile.  need a little refactor.

PR #4244, opened 2022-02-25
ISSUE-4241: version of storage layout

### LINKED ISSUE #4243: Add `version` to storage layout for upgrade

dantengsky:

## Summary

Issues trying to be addressed in the PR:

-  Table may have multiple format versions of snapshots (time travel)
-  TableSnapshot may have multiple format versions of Segment
-  Segment may also have ... 

---------------

- Introduce table option FUSE_OPT_KEY_SNAPSHOT_VER
  which marks the current format version of the snapshot inside meta-server.
- Enable format version fields in SegmentInfo and TableSnapshot 

--------------

- no observable performance degradation in the benchmark of the "ontime" data set

## Changelog

- Improvement

## Related Issues

Fixes #4243

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/meta/v1/block.rs
```diff
@@ -1,4 +1,4 @@
-//  Copyright 2021 Datafuse Labs.
+//  Copyright 2022 Datafuse Labs.
```
BohuTANG, line 1: See that we make a new `v1` directory for the v1 version, if we only add a new field of the block layout, how about the new directory?
v2/block.rs or ?

dantengsky: yeah, there is another versioned mod (but is v0 :D). will be pushed later

query/src/storages/fuse/fuse_table.rs
```diff
@@ -172,27 +173,48 @@ impl Table for FuseTable {
...
     }
 
+    pub fn snapshot_format_version(&self) -> Result<u64> {
+        Self::parse_snaphost_format_version(&self.table_info)
+    }
+
+    pub fn parse_snaphost_format_version(table_info: &TableInfo) -> Result<u64> {
+        table_info
+            .options()
+            .get(FUSE_OPT_KEY_SNAPSHOT_VER)
```
BohuTANG, line 190: Do we have the default version? If so, where the default defined?

query/src/storages/fuse/meta/versions.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::marker::PhantomData;
+
+use common_datablocks::DataBlock;
+use common_exception::ErrorCode;
+
+use crate::storages::fuse::meta::common::Versioned;
+use crate::storages::fuse::meta::v0;
+use crate::storages::fuse::meta::v1;
+
+pub const CURRNET_SEGMETN_VERSION: u64 = v1::SegmentInfo::VERSION;
```
BohuTANG, line 24: Guess `CURRNET_*` should be `CURRENT_*`

query/src/storages/fuse/io/locations.rs
```diff
@@ -34,24 +39,51 @@ impl TableMetaLocationGenerator {
...
+}
+
+trait SnapshotLocationCreator {
+    fn create(&self, id: &Uuid, prefix: impl AsRef<str>) -> String;
+}
+
+impl SnapshotLocationCreator for SnapshotVersion {
+    fn create(&self, id: &Uuid, prefix: impl AsRef<str>) -> String {
+        match self {
+            SnapshotVersion::V0(_) => {
```
BohuTANG, line 70: :+1: 
Thank you!

query/src/storages/fuse/fuse_table.rs
dantengsky: Yes, we should have a default value of snapshot version, which is 0, or `v0::Snapshot::VERSION`. let me fix it.

format-versions other than v0::Snapshot should have a table option explicitly defined in table options.

query/src/storages/fuse/meta/versions.rs
dantengsky: Technically, these constants could be removed. these are the legacy code to ensure that for each type of table meta type (e.g. SegmentInfo), there is only one numeric number has been bound to it (the `impl Version<n> for SegmentInfo`)

query/src/storages/fuse/fuse_table.rs
BohuTANG: We can set a current default version, upgrade it by hardcode if we changed the format, why need the table options to defined here?

dantengsky: > Do we have the default version?

Some basic principles of this PR (should have been written in the PR's summary:- )   :

-  users other than `VersionedReader`  should always be working on the data structure exported by `meta::current`
 -  `io::meta_readers` always return the CURRENT version of the table metadata
    - conversion between versions is explicitly executed inside the reader.
    - backward compatibility if enforced by the implementation of `From <Old> for New`
      and of course, depends on that the converters are implemented correctly
 -  writers always write the table metadata of the CURRENT format-version
 - with the following exceptions
    - During the merge of TableSnaphost, versions of segments are kept (one snapshot may refer to different versions of segments)
    - The previous TableSnaphost of a given table snapshot may have a different format version.

query/src/storages/fuse/operations/commit.rs
```diff
@@ -221,18 +235,23 @@ impl FuseTable {
...
+
+        let req = UpsertTableOptionReq {
+            table_id,
+            seq: MatchSeq::Exact(tbl_id.version),
+            options: [
+                (
+                    FUSE_OPT_KEY_SNAPSHOT_LOC.to_owned(),
+                    Some(new_snapshot_location),
+                ),
+                (FUSE_OPT_KEY_SNAPSHOT_VER.to_owned(), Some("1".to_owned())),
```
BohuTANG, line 248: I think we should hard code some default value for the current like:
CURRENT_SNAPSHOT_VER = 1
...

dantengsky: Right & thanks!!

BohuTANG (approved):
:rocket:

PR #4255, opened 2022-02-26
Improvement: Use case insensitive table option keys

dantengsky:

## Summary

Use case insensitive table option keys

## Changelog

- Improvement

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/storages/fuse/constants.rs
+++ b/query/src/storages/fuse/constants.rs
@@ -13,12 +13,12 @@
 //  limitations under the License.
 //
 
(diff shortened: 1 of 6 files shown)


dantengsky:
> Is it possible to add a stateless test for `create table ... snapshot_loc='<ss>'`? Hard part is how to get the snapshot id and set to the create statement.

let' me try it

BohuTANG:
/lgtm

I think the failure is not related to this PR, let's merge.

Issue #4258, opened 2022-02-26
bug: `SELECT number FROM numbers_mt(10) WHERE number > 8 LIMIT 2` returns 0 rows


platoneko:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

databend v0.6.74-nightly

### What's Wrong?

```
databend :) SELECT number FROM numbers_mt(10) WHERE number > 8 LIMIT 2

SELECT number
FROM numbers_mt(10)
WHERE number > 8
LIMIT 2

Query id: f69fb15f-4bf8-49c9-b79b-669d1138a71d

Ok.

0 rows in set. Elapsed: 0.034 sec.
```

### What You Expected?

```
┌─number─┐
│      9 │
└────────┘
```

### Anything Else?

```
databend :) SELECT number FROM numbers_mt(10) WHERE number > 8 LIMIT 10

SELECT number
FROM numbers_mt(10)
WHERE number > 8
LIMIT 10

Query id: 574d132e-964b-47bb-998e-1dbaa6350df1

┌─number─┐
│      9 │
└────────┘

1 rows in set. Elapsed: 0.034 sec. 
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

BohuTANG:
Good catch！
I guess(haven't test on other engines e.g. Memory or Fuse yet) it's the numbers table bug.

If so, this will be helpful:
https://github.com/datafuselabs/databend/pull/4215

We can c

sundy-li:
Limit pushdown should be rejected if filter is Some.

PR #4261, opened 2022-02-26
fix numbers_mt limit pushdown bug

platoneko:

## Summary

Summary about this PR

## Changelog

- Bug Fix

## Related Issues

Fixes [#4258](https://github.com/datafuselabs/databend/issues/4258)

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
It would be better to add a unit test to check that our patch meets expectations  :)

BohuTANG:
@mergify update

query/tests/it/table_functions/numbers_table.rs
```diff
@@ -57,3 +57,46 @@ async fn test_number_table() -> Result<()> {
...
+    ];
+
+    for test in tests {
+        let ctx = crate::tests::create_query_context()?;
+        let plan = PlanParser::parse(ctx.clone(), test.query).await?;
+        let mut optimizer = Optimizers::without_scatters(ctx);
+
+        let optimized_plan = optimizer.optimize(&plan)?;
+        let actual = format!("{:?}", optimized_plan);
+        assert_eq!(test.expect, actual, "{:#?}", test.name);
```
BohuTANG, line 99: We should check the final block result not only the optimized plan :)
Please reference:
https://github.com/datafuselabs/databend/blob/6eb7ca72c29428c5d24d1552d0908bcea32d9d6d/query/tests/it/table_functions/numbers_table.rs#L42-L56

sundy-li (approved):

PR #4269, opened 2022-02-27
Zero extra cost of async trait

sundy-li:

## Summary

Summary about this PR

Use rust's GAT feature to have zero extra async trait instead of `#[async_trait]` in some hot path functions.
This is an interesting idea from: https://www.skyzh.dev/posts/articles/2022-01-31-gat-async-trait/
 
cc @zhang2014  

cc @Xuanwo  @drmingdrmer   Maybe this can also help in  opendal, openraft and openkv ?

## Changelog

- Improvement
 

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
Great.
I think @drmingdrmer is interested in this for OpenRaft :D

Xuanwo:
There are the following questions for `opendal` to adapt to this change:

- Is this trait `dyn safe`? opendal highly depends on `Arc<dyn Accessor>`. I'm still finding a better way for that.
- `opendal` expect to work on `stable` rust. I will try to measure the difference in performance to decide whether or not we give up this limitation.

BohuTANG:
If you mind `Arc<dyn Accessor>` there is a easy way to measure:
writing a bench test for Arc<T> and T.

If we know that some place are performance bottlenecks, rather than optimizing ahead of time :)

sundy-li:
I think it's suitable to use it when we call async trait functions inside a hot loop (This did not have `Boxed<dyn Future>` cost).

Eg:

-  Async Iterator: async next function
-  Async consume/generate function

For other not hot path async functions, it's useless to use it.

PsiACE (approved):
lgtm.

I think we should consider maintaining some benchmark suites for hot paths.

Xuanwo (approved):

PR #4270, opened 2022-02-27
chore(metactl): move to tools/metactl

PsiACE:

## Summary

chore(metactl): move to tools/metactl

It's just a simple move that doesn't look like it will break anything.

## Changelog

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


Xuanwo (approved):

BohuTANG:
Does the `metactl` is build and in our release package?

PsiACE:
> Does the `metactl` is build and in our release package?

It is currently not built with the `make` cmd and is not in our release.

BohuTANG:
> > Does the `metactl` is build and in our release package?
> 
> It is currently not built with the `make` cmd and is not in our release.

I think we should built metactl to our release, could you do it in this PR? @PsiACE 
Thank you.

PsiACE:
> I think we should built metactl to our release, could you do it in this PR?

ok

PsiACE:
@BohuTANG @Xuanwo PTAL, I tried to add an asset called databend-tools, including metactl and benchmark.

Issue #4264, opened 2022-02-27
Feature:  Simd Selection of column filter


sundy-li:
**Summary**

Description for this feature.

Now filter works with memcpy one by one, we can use `Vec.extend` function to copy a batch of  at most 64 items one time.

https://github.com/datafuselabs/databend/blob/dc058c9d22baa9e61763661f77cd10ec62c87c48/common/datavalues/src/columns/primitive/mod.rs#L187-L204

**Some useful tools:**

- Bitmap's Chunk function:
```
impl Bitmap {
    /// Returns an iterator over bits in chunks of `T`, which is useful for
    /// bit operations.
    pub fn chunks<T: BitChunk>(&self) -> BitChunks<T> {
        BitChunks::new(&self.bytes, self.offset, self.length)
    }
}
```

-  Function:  `__builtin_ctz`
```
  int __builtin_ctz (unsigned int x)
Returns the number of trailing 0-bits in x, starting at the least significant bit position. If x is 0, the result is undefined.
```

Rust's  https://doc.rust-lang.org/std/primitive.u32.html#method.trailing_zeros

 - Blog: https://lemire.me/blog/2018/02/21/iterating-over-set-bits-quickly/

PR #4271, opened 2022-02-28
ISSUE-4264: simd selection

platoneko:

## Summary

Summary about this PR

## Changelog

- New Feature
- Performance Improvement

## Related Issues

Fixes [#4264](https://github.com/datafuselabs/databend/issues/4264)

## Test Plan

Unit Tests

Stateless Tests


common/datavalues/src/lib.rs
```diff
@@ -18,6 +18,7 @@
 
 #![feature(generic_associated_types)]
 #![feature(trusted_len)]
+#![feature(core_intrinsics)]
```
sundy-li, line 21: use https://doc.rust-lang.org/std/primitive.u32.html#method.trailing_zeros   instead

sundy-li:
Almost Looks good to me.

I just posted  [a pr in arrow2](https://github.com/jorgecarleitao/arrow2/pull/871), you can have a look, but you don't have to follow that pr's behavior.

common/datavalues/src/columns/primitive/mod.rs
```diff
@@ -189,17 +189,32 @@ impl<T: PrimitiveType> Column for PrimitiveColumn<T> {
...
+        let values = self.values();
+
+        const MASK_BITS: usize = 64;
+        for mut mask in filter.values().chunks::<u64>() {
+            if mask == u64::MAX {
+                res.extend(&values[offset..offset + MASK_BITS]);
+            } else {
+                while mask != 0 {
+                    let n = std::intrinsics::cttz(mask) as usize;
+                    res.push(values[offset + n]);
```
sundy-li, line 204: maybe better to use `ptr.write` to have better performance.

common/datavalues/src/columns/primitive/mod.rs
```diff
@@ -189,17 +189,32 @@ impl<T: PrimitiveType> Column for PrimitiveColumn<T> {
...
-        let values: Vec<T> = iter.collect();
-        let col = PrimitiveColumn {
-            values: values.into(),
-        };
+        let mut res = Vec::<T>::with_capacity(length);
+        let mut offset = 0;
+        let values = self.values();
+
+        const MASK_BITS: usize = 64;
+        for mut mask in filter.values().chunks::<u64>() {
```
sundy-li, line 198: should care about the remaining in chunks.

platoneko: I  have checked that `Bitmap::chunks()` won't yield chunk less than 64

platoneko: In my unittest, I set N = 1000 to make Column not aligned to 64.

common/datavalues/src/lib.rs
platoneko: LGTM

common/datavalues/src/columns/primitive/mod.rs
```diff
@@ -189,17 +189,35 @@ impl<T: PrimitiveType> Column for PrimitiveColumn<T> {
...
+        }
+
+        for (&v, _) in values
             .iter()
             .zip(filter.values().iter())
+            .skip(offset)
             .filter(|(_, f)| *f)
-            .map(|(v, _)| *v);
+        {
+            res.push(v);
```
platoneko, line 217: Processing the remaining in chunks is here, maybe not elegant :(

common/datavalues/src/columns/primitive/mod.rs
platoneko: > maybe better to use `ptr.write` to have better performance.

I'd like to have a try

common/datavalues/src/columns/primitive/mod.rs
sundy-li: It's not elegant. `skip` may cause extra iterating.

common/datavalues/src/columns/primitive/mod.rs
```diff
@@ -190,34 +191,49 @@ impl<T: PrimitiveType> Column for PrimitiveColumn<T> {
...
-                while mask != 0 {
-                    let n = std::intrinsics::cttz(mask) as usize;
-                    res.push(values[offset + n]);
-                    mask = mask & (mask - 1);
+        let mut new = Vec::<T>::with_capacity(length);
+        let mut dst = new.as_mut_ptr();
+
+        const CHUNK_SIZE: usize = 64;
+        let mut chunks = self.values().chunks_exact(CHUNK_SIZE);
+        let mut mask_chunks = filter.values().chunks::<u64>();
```
sundy-li, line 199: There is a faster path for `mask_chunks`, if the filter has no offsets, it could use `chunks_exact` to have better performance.  See: https://github.com/jorgecarleitao/arrow2/blob/main/src/compute/filter.rs#L142-L147

There are two approaches:

1.  Use generic dispatch like `BitChunkIterExact`

2. if offset > 0, use an extra loop to consume the offset, this is called `Header Loops`, then we can continue the  `Main Loops` && `Tail Loops`.

The second one may have better performance because `chunks.iter()` must merge two u64 values to generate one merged u64, see:  merge_reversed in `arrow2`

sundy-li:
Seems this is slower:

Main:

```

MySQL [(none)]> select count() from numbers_mt(10000000000) where rand()  > 0.90;
+------------+
| count()    |
+------------+
| 1000011915 |
+------------+
1 row in set (1.942 sec)

```

Now:
```


MySQL [(none)]> select count() from numbers_mt(10000000000) where rand()  > 0.90;
+------------+
| count()    |
+------------+
| 1000049955 |
+------------+
1 row in set (2.689 sec)
```

sundy-li:
@mergify update

sundy-li:
@platoneko  LGTM now. Would you want this pr to be merged? Or maybe you want to continue to add this into `BooleanColumn` && `StringColumn` in this pr?

sundy-li:
@mergify update

platoneko:
> 
I want this pr to be merged :)

Issue #4256, opened 2022-02-26
query_log/bug:  scan_byte_cost_ms is 0


BohuTANG:
**Summary**

Ontime dataset, query is:
`SELECT DayOfWeek, count(*) AS c FROM ontime WHERE Year >= 2000 AND Year <= 2008 GROUP BY DayOfWeek ORDER BY c DESC`

query_log:
```
           log_type: 2
       handler_type: MySQL
          tenant_id: test_tenant
         cluster_id: test_cluster
           sql_user: root
     sql_user_quota: UserQuota { max_cpu: 0, max_memory_in_bytes: 0, max_storage_in_bytes: 0 }
sql_user_privileges: UserGrantSet { entries: [GrantEntry { user: "root", host_pattern: "127.0.0.1", object: Global, privileges: BitFlags<UserPrivilegeType>(0b1111111101111, Usage | Select | Insert | Update | Delete | Create | Drop | Alter | S
uper | CreateUser | CreateRole | Grant) }], roles: {} }
           query_id: 14fb2884-230a-4608-855b-0e4b2bbf19d7
         query_kind: SelectPlan
         query_text: SELECT DayOfWeek, count(*) AS c FROM ontime WHERE Year >= 2000 AND Year <= 2008 GROUP BY DayOfWeek ORDER BY c DESC
         event_date: 2022-02-26
         event_time: 1645872355531
   current_database: default
          databases: 
             tables: 
            columns: 
        projections: 
       written_rows: 0
      written_bytes: 0
          scan_rows: 195662213
         scan_bytes: 586986639
  scan_byte_cost_ms: 0
         scan_seeks: 0
  scan_seek_cost_ms: 0
    scan_partitions: 196
   total_partitions: 196
        result_rows: 7
       result_bytes: 63
          cpu_usage: 8
       memory_usage: 2672
        client_info: 
     client_address: Some(127.0.0.1:39130)
     exception_code: 0
     exception_text: 
        stack_trace: 
     server_version: 
              extra: 

```

BohuTANG:
Just found that the `scan_byte_cost_ms` metric we do not count it, where the place to add?
```
#[async_trait]
impl Accessor for DalContext {
    async fn read(&self, args: &OpRead) -> DalResult<BoxedAsyncReader> {
        let metrics = self.metrics.clone();

        // TODO(xuanwo): Maybe it's better to move into metrics.
        self.inner.as_ref().unwrap().read(args).await.map(|reader| {
[...]

Issue #3586, opened 2021-12-22
Copy INTO <table> from external location 


BohuTANG:
**Summary**
Load all files in a s3 bucket and insert into table:

```
copy into mytable
  from 's3://mybucket/data/files'
  credentials=(aws_key_id='$AWS_ACCESS_KEY_ID' aws_secret_key='$AWS_SECRET_ACCESS_KEY')
  encryption=(master_key = 'eSxX0jzYfIamtnBKOEOwq80Au6NbSgPH5r4BDDwOaO8=')
  file_format = (type = csv field_delimiter = '|' skip_header = 1);
```

Tasks:
- [x] https://github.com/datafuselabs/databend/pull/4170 (works with minio)
- [x] #4277
- [ ] #4308
- [x] Copy from S3 with Anonymous access (OpenDAL done, TODO: need add a stateful test for it)
- [x] Copy from S3 pattern name files (#4328)
- [x] Copy from S3 all directory files (#4373)
- [x] Copy from with Parquet file (#4373)

BohuTANG:
We need some data accessor apis to help the file reading, maybe we can address it in:
https://github.com/datafuselabs/databend/pull/3575 

cc @Xuanwo

Xuanwo:
I'm working on #3436 now.

I believe we will have a more elegant API design to support those feature requests.

GrapeBaBa:
need wait #3436 ？

BohuTANG:
> need wait #3436 ？

That's right!

GrapeBaBa:
@Xuanwo can this task start right now?

Xuanwo:
@GrapeBaBa, sorry for the late, I think we still need to wait for https://github.com/datafuselabs/databend/issues/3874

BohuTANG:
Plan to experience the new dal design, I take this issue.
Feel free to comments  @GrapeBaBa

BohuTANG:
This feature need openDAL list: https://github.com/datafuselabs/opendal/issues/12 cc @Xuanwo

BohuTANG:
Ready to close, and #4308 rolling to next version(Nightly v0.8)

PR #4277, opened 2022-02-28
External source with new processor

BohuTANG:

## Summary

Summary about this PR:
1. Bump opendal to v0.1.3.
2. Make external source(like S3) as table engine:
    * We can load data as `COPY INTO t1 FROM 's3://'`, use `read2()` api
    * We can unload data from table to s3 as `COPY INTO s3:// FROM t1`, use `append_data()` api
    * We can purge the s3 file after load data, use `truncate()` api
    * We can query from s3 file directly, for some file format we can do the predicates/limit push down.

## Changelog

- Improvement

## Related Issues

Part of #3586 

## Test Plan

Unit Tests

Stateless Tests


common/planners/src/plan_read_datasource.rs
```diff
@@ -16,18 +16,45 @@ use std::collections::BTreeMap;
...
+use crate::S3ExternalTableInfo;
 use crate::Statistics;
 
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq)]
+pub enum SourceInfo {
+    // Normal table source, 'fuse/system'.
+    TableSource(TableInfo),
+
+    // S3 external source, 's3://'.
+    S3ExternalSource(S3ExternalTableInfo),
```
sundy-li, line 35: Why not make `s3` as a special table? 

Such as a table function that holds the temporary table.

BohuTANG: Make s3 as table engine, we can do the unloading as `copy from table into s3://`

Xuanwo: It seems not extensible enough, considering we will support `azblob`, `gcs`, and other locations.

query/src/storages/s3/s3_external_table.rs
```diff
@@ -0,0 +1,108 @@
...
+    ) -> Result<SendableDataBlockStream> {
+        unimplemented!()
+    }
+
+    async fn truncate(
+        &self,
+        _ctx: Arc<QueryContext>,
+        _truncate_plan: TruncateTablePlan,
+    ) -> Result<()> {
+        Ok(())
```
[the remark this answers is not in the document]
BohuTANG: Addressed.

common/planners/src/plan_read_datasource.rs
sundy-li: But table function can also support insert.

`copy from table into s3://`

  `select * from s3://`

1. `as_table` to generate a temporary table.
2.  Select or insert works with the table.
3.

BohuTANG: I have add a desc for why make it as a table engine.

BohuTANG:
> To fix #4256, I start a PR #4298 to address it. Let's merge the `main` branch after the PR gets merged. 😆

I will rebase this PR later :)

common/planners/src/plan_read_datasource.rs
BohuTANG: BTW, for `select * from @stage ...`, we will convert the `@stage` alias to the external table info which get from the metasrv(after we support stage) in the sql parser phase, it will be same as `select * from s3://` did.

BohuTANG:
@zhang2014 @sundy-li,  PTAL
@Xuanwo ,PTAL the part of OpenDAL
Thanks

sundy-li (approved):

PR #4291, opened 2022-03-02
refactor: clearer table_functions layout

PsiACE:

## Summary

- move number_table into storages/memory
- move table_functions into functions
- rename fuse/table_functions to fuse/history_table

## Changelog

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
Well,
1. table_functions is not same as functions(scalar or aggregate), it's related to the table, general it's a special funcion
2. number_table is a table function not a storage
3. history_table is one of the table function, we should not need move

So, I close this PR.

Issue #4294, opened 2022-03-02
CALL: Parser for CALL Syntax


flaneur2020:
**Tracking Issue**

https://github.com/datafuselabs/databend/issues/4267

Issue #4295, opened 2022-03-02
CALL: Trait for system function


flaneur2020:
**Tracking**

https://github.com/datafuselabs/databend/issues/4267

**Summary**

add a trait definition for system functions, allowing add system functions easily. 

some requirements:

- allow passing arbitrary arguments
- allows access control with privileges
- allows hide the function if current user do not have permission to this function

ZhiHanZ:
I think it could be closed? cc  @Junnplus

Junnplus:
@ZhiHanZ I think so. cc @flaneur2020

PR #4315, opened 2022-03-03
[Query] Add CALL CMD

Junnplus:
Signed-off-by: ye.sijun <junnplus@gmail.com>

## Summary

- Impl CALL syntax parser.
- Add Trait for system function.
- Add `fuse_history` function for CALL.

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4294 https://github.com/datafuselabs/databend/issues/4295

## Test Plan

Unit Tests

Stateless Tests


wangzhen11aaa:
Excellent, small neat code block.

BohuTANG:
Nice work,  will take a look :)

common/planners/src/plan_node_display_indent.rs
```diff
@@ -304,4 +306,9 @@ impl<'a> PlanNodeIndentFormatDisplay<'a> {
     fn format_copy(f: &mut Formatter, plan: &CopyPlan) -> fmt::Result {
         write!(f, "{:?}", plan)
     }
+
+    fn format_call(f: &mut Formatter, _plan: &CallPlan) -> fmt::Result {
+        // TODO
+        write!(f, "Call")
```
BohuTANG, line 312: Maybe we can print the args,like `CALL <name>(arg1, arg2, ...)`

query/src/interpreters/interpreter_call.rs
```diff
@@ -0,0 +1,68 @@
...
+    ) -> Result<SendableDataBlockStream> {
+        let plan = &self.plan;
+        plan.validate()?;
+        // TODO: fine-grained permissions
+        self.ctx
+            .get_current_session()
+            .validate_privilege(&GrantObject::Global, UserPrivilegeType::Super)
+            .await?;
+        let name = plan.name.clone();
+        let func = FunctionFactory::instance().get(name)?;
```
BohuTANG, line 60: See that we place the call function to `common/functions/system`, how can we do a function like `call fuse_history('db','table')` in the CallInterpreter?
Since all the fuse_history logic is in the query.

Junnplus: > Since all the fuse_history logic is in the query.

It looks like we need to move the call function to `query/src/functions`.

common/planners/src/plan_node_display_indent.rs
Junnplus: SGTM

query/src/interpreters/interpreter_call.rs
BohuTANG: That's right.

query/src/interpreters/interpreter_call.rs
```diff
@@ -0,0 +1,84 @@
...
+        mut _input_stream: Option<SendableDataBlockStream>,
+    ) -> Result<SendableDataBlockStream> {
+        let plan = &self.plan;
+        self.validate()?;
+        // TODO: fine-grained permissions
+        self.ctx
+            .get_current_session()
+            .validate_privilege(&GrantObject::Global, UserPrivilegeType::Super)
+            .await?;
+        let name = plan.name.clone();
```
BohuTANG, line 75: We can use the `fuse_history` as the `call` demo, will address the way work or not.
https://github.com/datafuselabs/databend/blob/7335ea73dce65483ba298c6527fdc9a383530055/query/src/storages/fuse/table_functions/fuse_history_table.rs

flaneur2020:
nice job！

query/src/interpreters/interpreter_call.rs
Junnplus: > We can use the fuse_history as the call demo.

Good idea, I added an implementation of `fuse_history`, PTAL : )

BohuTANG: Add @dantengsky to review :)

sundy-li (approved):

query/src/functions/systems/fuse_history.rs
```diff
@@ -0,0 +1,75 @@
...
+        let tbl = ctx
+            .get_catalog()
+            .get_table(
+                tenant_id.as_str(),
+                database_name.as_str(),
+                table_name.as_str(),
+            )
+            .await?;
+
+        let tbl = tbl.as_any().downcast_ref::<FuseTable>().ok_or_else(|| {
```
dantengsky, line 62: `FuseTable::try_from_table`  be reused here?

https://github.com/datafuselabs/databend/blob/953abd39cece82d26c9c8ac5ebb267ae2574611c/query/src/storages/fuse/fuse_table.rs#L183-L190

query/src/functions/systems/fuse_history.rs
```diff
@@ -0,0 +1,75 @@
...
+            ErrorCode::BadArguments(format!(
+                "expecting fuse table, but got table of engine type: {}",
+                tbl.get_table_info().meta.engine
+            ))
+        })?;
+
+        Ok(FuseHistory::new(ctx, tbl.clone()).get_history().await?)
```
dantengsky, line 69: not sure if a reference (instead of `tbl.clone()`) is possible here, `FuseHistory` owns a `FuseTable`, but seems that a reference of the table is enough to fulfill the works.

query/src/functions/systems/fuse_history.rs
Junnplus: They have different error codes.

query/src/functions/systems/fuse_history.rs
dantengsky: `tbl.clone()` is not that heavy,  let's move on

query/src/functions/systems/fuse_history.rs
dantengsky: > They have different error codes.

got it

PR #4339, opened 2022-03-05
polish wall_time_ms in http handler

youngsofun:

## Summary

1. return wall_time_ms only after execute finished, before that it is None/nil
2. use float instead of int
3. include parse and analyze time

## Changelog

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


youngsofun:
\review @Chasen-Zhang

ZhiHanZ:
overall LGTM, could you clarify the wall time in the documentation and add examples with instructions for checking on query logs for profiling time such as planner time etc

ZhiHanZ (approved):

youngsofun:
> overall LGTM, could you clarify the wall time in the documentation and add examples with instructions for checking on query logs for profiling time such as planner time etc

ok

PR #4158, opened 2022-02-15
Draft RFC for support nullable_number_type.

wangzhen11aaa:
Signed-off-by: wang.zhen <wangzhenaaa7@gmail.com>

I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary
Draft RFC for support nullable_number_type.

## Changelog

- Bug Fix

## Related Issues

https://github.com/datafuselabs/databend/issues/4079

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

PR #4340, opened 2022-03-05
Support nullable group by

### LINKED ISSUE #4079: Feature: full support of group by nullable column
**Summary**

Description for this feature.

## Currently

we ignore the nulls to generate the hash key of the group by columns.

```
  pub fn fixed_hash(column: &ColumnRef, ptr: *mut u8, step: usize) -> Result<()> {
        let column = column.convert_full_column();
        // TODO support nullable
        let column = Series::remove_nullable(&column);
        ...
    }
```

So the results are not correct in this query:
```
MySQL [(none)]> create table t(a UInt64, b UInt32) Engine = Fuse;
Query OK, 0 rows affected (0.003 sec)

MySQL [(none)]>
MySQL [(none)]> insert into t(a,b)  select if (number % 3 = 1, null, number) as a, number + 3 as b from numbers(10);
Query OK, 0 rows affected (0.006 sec)

MySQL [(none)]> select a%3 as c, count(1) from t group by c;
+------+----------+
| c    | count(1) |
+------+----------+
|    0 |        7 |
|    2 |        3 |
+------+----------+
2 rows in set (0.005 sec)
```

## How:

1. `a % 3 as c`  is `Nullable(UInt8)`. If `c` is `UInt8`, we can hash column `c` into `UInt8` Column using `fixed_hash`. But we should use extra bits to identify the null value,  so `UInt16` is used:   [ `8bit(value)`  --  `8bits(just use 1 bit to identify null) ]`

2. If there are multiple nullable columns, say: N,  `N * 8` bits will be used.

3. The final bits must be rounded into primitive types： [ `Nullabe(UInt8)`, `UInt8` ]   -->  `24bits` ---> round into `32bits`, UInt32Column will be used to store the hashkeys.

This task can be taken after #4074  is merged.

Related:
- DataBlock::choose_hash_method
- Series::fixed_hash
- Series::serialize

sundy-li:

## Summary

Summary about this PR

## Changelog

 
- Improvement
 

## Related Issues

Continue: #4158
Fixes #4079 

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/common/datavalues/src/columns/group_hash.rs
+++ b/common/datavalues/src/columns/group_hash.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 use std::fmt::Debug;
 
+use common_arrow::arrow::bitmap::Bitmap;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_io::prelude::*;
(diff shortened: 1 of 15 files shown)


common/datavalues/src/columns/group_hash.rs
```diff
@@ -39,15 +73,40 @@ impl Series {
     /// Apply binary mode function to each element of the column.
     /// WARN: Can't use `&mut [Vec<u8>]` because it has performance drawback.
     /// Refer: https://github.com/rust-lang/rust-clippy/issues/8334
-    pub fn serialize(column: &ColumnRef, vec: &mut Vec<Vec<u8>>) -> Result<()> {
+    /*
+     * not nullable col1                 nullable col2 (first byte to indicate null or not)
```
BohuTANG, line 77: :100:

sundy-li:
Ontime datasets benchmark (the second query with metacache)：

Before:

```
MySQL [(none)]> SELECT DayOfWeek, count(*) AS c FROM ontime GROUP BY DayOfWeek ORDER BY c DESC;
+-----------+----------+
| DayOfWeek | c        |
+-----------+----------+
|         3 | 23513895 |
|         2 | 23384295 |
|         4 | 23113350 |
|         7 | 21270195 |
|         1 | 20861550 |
|         6 | 19360215 |
|         5 | 18896085 |
+-----------+----------+
7 rows in set (2.578 sec)
```

After:

```
MySQL [(none)]> SELECT DayOfWeek, count(*) AS c FROM ontime GROUP BY DayOfWeek ORDER BY c DESC;
+-----------+----------+
| DayOfWeek | c        |
+-----------+----------+
|         3 | 23513895 |
|         2 | 23384295 |
|         4 | 23113350 |
|         7 | 21270195 |
|         1 | 20861550 |
|         6 | 19360215 |
|         5 | 18896085 |
+-----------+----------+
7 rows in set (0.896 sec)
```

zhang2014 (approved):

BohuTANG:
Also cc @wubx @dantengsky for the performance improvement.
 
It's time to change the subject/ontime table schema:
https://github.com/datafuselabs/databend/discussions/4342

Issue #3792, opened 2022-01-06
Add escape for management key


BohuTANG:
## Summary

Cluster already has the escape for the key:
https://github.com/datafuselabs/databend/blob/b80c460048cbf64eb4dd2f5c7d905df4fc9b677f/common/management/src/cluster/cluster_mgr.rs#L62-L83

stage/udf/user also need.

BohuTANG:
Hello, @BlackSmith96 
Any progress for this issue? We add this issue to v0.7(plan to release in the mid of this month).
Thanks.

BlackSmith96:
@BohuTANG Sorry for delay.  I'll create a PR these two days.

BohuTANG:
Thank you.

PR #4363, opened 2022-03-08
Add escape for management key

BlackSmith96:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

#3792 

## Test Plan

Unit Tests

Stateless Tests


BohuTANG:
@mergify update

BohuTANG:
We can move the common function `escape_for_key` from `common/management/` to `common/base`

BlackSmith96:
Hi, @zhang2014 Could I have a question about why we need unescape_for_key function in cluster_mgr.rs?
https://github.com/datafuselabs/databend/blob/7bbafdc4c58c724019465a88298fe92f6b77cbc4/common/management/src/cluster/cluster_mgr.rs#L165-L178
Will the `id` field in `node_info` change after kv write & read?

zhang2014:
> Hi, @zhang2014 Could I have a question about why we need unescape_for_key function in cluster_mgr.rs?

unescape_for_key for `SELECT id FROM system.clusters`.

> Will the id field in node_info change after kv write & read?

No, it's immutable.

BohuTANG:
@mergify update

common/base/src/string_func.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    let bytes = key.as_bytes();
+
+    let mut index = 0;
+    while index < bytes.len() {
+        match bytes[index] {
+            b'%' => {
+                let mut num = unhex(bytes[index + 1]) * 16;
```
ariesdevil, line 60: Should we check the index out of range if the index is already the last byte?

ariesdevil: For me I prefer to use iterator here so that we just call `next()` and `peek()` do stuff, and it's idiomatic rust way.

common/base/src/string_func.rs
```diff
@@ -0,0 +1,73 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::string::FromUtf8Error;
+
+pub fn escape_for_key(key: &str) -> Result<String, FromUtf8Error> {
```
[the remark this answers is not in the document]
lichuang: Also,some unit test cases for this `string_func` module is prefered.

BlackSmith96:
@mergify update

common/base/src/string_func.rs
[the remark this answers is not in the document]
BlackSmith96: Have add comment and unit test.

common/base/src/string_func.rs
BlackSmith96: > hould we check the index out of range if the index is already the last byte? If we can guarantee `%` does not appear in the last position, we should leave a comment here.

This function is only used with reverse case of escape_for_key. I think there is no need to check the input. I have leave a comment in the code. 

> For me I prefer to use iterator here so that we just call `next()` and `peek()` do stuff, and it's idiomatic rust way.

In this function, the iterator sometimes needs to move 3 steps. I think use index would be more easy.

common/base/tests/it/string_func.rs
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common_base::*;
+use common_exception::Result;
+
+#[test]
+fn test_progress() -> Result<()> {
```
BohuTANG, line 19: How about adding the slash case?
@drmingdrmer expected your suggestions on the escape case :)

drmingdrmer: @BohuTANG 
Looks good to me.
And it does not look like affecting me at all. 🤔

BlackSmith96: @BohuTANG Thanks for your advise. I have changed the test case.

ariesdevil (approved):
LGTM, thanks for your contribution and patience

BohuTANG:
Thank you @BlackSmith96 !

PR #4367, opened 2022-03-09
[http handler] rename response fields and return elapsed time when running.

youngsofun:

## Summary

1.  rename progress to scan_progress.  context also provide result_progress, but I guess client does not need it now.
3.  return elapsed time when running,  useful for client to calculate speed from progress
4.  rename wall_time_ms to running_time_ms. 
5.  polish description about running_time_ms in doc

@Chasen-Zhang @flaneur2020 

## Changelog

## Related Issues

no issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/docs/user/03-api/http-handler.md
+++ b/docs/user/03-api/http-handler.md
@@ -23,7 +23,7 @@ example:
 
 ```
 {
-   "sql": "select * from numbers(10)"
+   "sql": "select * from numbers(10)"[
 }
 ```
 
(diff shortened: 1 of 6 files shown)


docs/user/03-api/http-handler.md
```diff
@@ -23,7 +23,7 @@ example:
 
 ```
 {
-   "sql": "select * from numbers(10)"
+   "sql": "select * from numbers(10)"[
```
flaneur2020, line 26: typo?

PR #4372, opened 2022-03-09
Feature: embed more column meta data inside SegmentInfo

### LINKED ISSUE #4371: Feature: embed more column meta data inside SegmentInfo
**Summary**

As suggested by @zhang2014 :

It may be necessary/better if we could embed more metadata of columns inside the `SegmentInfo`, 

e.g.

~~~rust 
pub struct ColumnMeta {
    /// where the data of column start
    pub offset: u64,
    /// the length of the column
    pub len: u64,
    /// num of "rows"
    pub num_values: u64,
   /// and more...
}
~~~

such that

- during the `read plan` phase, after block pruning, more metadata of columns could be baked into `PartInfo`, without touching the data of blocks
- in the subsequent `read` phase, the block readers may 
    - extract the data of columns without touching the parquet FileMetaData
    - thus, the query node which executes `Table::read` no longer suffers from the cold meta caches

dantengsky:

## Summary

implementation of the feature request 
- #4371 

@zhang2014 let me know if there are any other requirements/suggestions

if the shape of `ColumnMeta` is OK, I'd like to merge the modifications of `arrows` and `parquet2`  into repos of `datafuse-extras` first.

## Changelog

- New Feature
- Improvement

## Related Issues

Fixes #4371

## Test Plan

Unit Tests

Stateless Tests


zhang2014 (approved):

dantengsky:
@BohuTANG  Pls take a look

PR #4373, opened 2022-03-09
feat(copy): support copy csv/parquet files from s3 directory

sundy-li:

## Summary

Summary about this PR

1. Bump opendal to 0.2.0
2. Support copy from directory
```
copy into ontime200 from 's3://testbucket/admin/data/' credentials=(aws_key_id='minioadmin' aws_secret_key='minioadmin') PATTERN = '.*ontime.*' FILE_FORMAT = (type = 'CSV' field_delimiter = ','  record_delimiter = '\n' skip_header = 1);
```

3. Support copy from parquet files

now works

## Changelog

- New Feature

## Related Issues

Part of  #3586

## Test Plan

Unit Tests

Stateless Tests


sundy-li:
S3::list returns abs path rather than path, cc @Xuanwo 
```

async fn test_s3_list() -> Result<()> {
    let acc = s3::new().await?;
    if acc.is_none() {
        println!("OPENDAL_S3_TEST not set, ignore");
        return Ok(());
    }
    let operator = Operator::new(acc.unwrap());
    let mut objects = operator.objects("/admin/data/");
    while let Some(object) = objects.next().await {
        let meta = object?.metadata().await?;
        println!("{:?}", meta);
    }
    Ok(())
}

```

returns:

```
running 1 test
Metadata { complete: true, path: "admin/data/logs/", mode: Some(DIR), content_length: Some(0) }
Metadata { complete: true, path: "admin/data/alltypes_plain.parquet", mode: Some(FILE), content_length: Some(1851) }
Metadata { complete: true, path: "admin/data/nyctaxi.csv", mode: Some(FILE), content_length: Some(1063) }
Metadata { complete: true, path: "admin/data/ontime_200.csv", mode: Some(FILE), content_length: Some(91007) }
Metadata { complete: true, path: "admin/data/ontime_200.parquet", mode: Some(FILE), content_length: Some(37753) }
Metadata { complete: true, path: "admin/data/ontime_200_v1.csv", mode: Some(FILE), content_length: Some(91007) }
Metadata { complete: true, path: "admin/data/sample.csv", mode: Some(FILE), content_length: Some(96) }
test s3::test_s3_list ... ok
```

Xuanwo:
It must be a bug, I will take it a look.

Xuanwo:
> It must be a bug, I will take it a look.

Oh, my mistake. The behavior is expected, we don't need to handle the `root` related logic.

It's Ok to construct a new object by `meta.path()`:

```rust
let (src, dst) = (s3::Backend, fs::Backed);
let objects = src.objects("dir");
for Some(o) = objects.next().await {
    let mut o = o?;
    let meta = o.metadata_cahed().await;
    let w = dst.object(meta.path());
    w.write_bytes(xxxx)
}
```

I will add a `copy` example under opendal.

query/src/interpreters/interpreter_copy.rs
```diff
@@ -111,7 +112,7 @@ impl CopyInterpreter {
     //  COPY should deal with the file one by one and do some error handler on the OnError strategy.
 
     #[tracing::instrument(level = "debug", name = "copy_one_file_to_table", skip(self), fields(ctx.id = self.ctx.get_id().as_str()))]
-    async fn copy_one_file_to_table(&self, file_name: Option<String>) -> Result<()> {
+    async fn copy_one_file_to_table(&self, file_name: Option<String>) -> Result<Vec<DataBlock>> {
```
BohuTANG, line 115: Not sure that if a file is 100GB if we return the Vec<DataBlock> is it OOM?

tests/suites/1_stateful/00_copy/00_0000_copy_from_s3_location.sh
```diff
@@ -11,18 +11,32 @@ echo "Test copy from file"
...
+echo "select count(1), avg(Year), sum(DayOfWeek)  from ontime200" | $MYSQL_CLIENT_CONNECT
+echo "truncate table ontime200" | $MYSQL_CLIENT_CONNECT
+
+
+## Copy from parquet
+echo "copy into ontime200 from 's3://testbucket/admin/data/' credentials=(aws_key_id='minioadmin' aws_secret_key='minioadmin') PATTERN = 'ontime.*parquet' FILE_FORMAT = (type = 'PARQUET')
+" | $MYSQL_CLIENT_CONNECT
+echo "select count(1), avg(Year), sum(DayOfWeek)  from ontime200" | $MYSQL_CLIENT_CONNECT
+echo "truncate table ontime200" | $MYSQL_CLIENT_CONNECT
+
```
BohuTANG, line 40: Cool.

In my previous work, there was another case that was not added here:
anonymous access

It seems we need the minio client, @Xuanwo has done some work related it:
https://github.com/datafuselabs/opendal/commit/e2cef593d12ce53749d9b31eb186dfa535782ee5

But it's ok we add it in the later PR.

query/src/interpreters/interpreter_copy.rs
sundy-li: It's the append entry result(with metadata: append id, ..), not the actual data block.

PR #4404, opened 2022-03-12
ISSUE-4403:  dedicated runtime for io tasks

### LINKED ISSUE #4403: Consider using dedicated io thread pool
**Summary**

Use dedicated thread pool to execute IO tasks.

dantengsky:

## Summary

- dedicated runtime for io tasks

  A dedicated tokio runtime is created during startup. 
  `cnfig.storage.storage_num_cpus` or `num_cpus() / 2` number of threads will be created for the pool.

- tasks of reading column chunks are spawned into io-runtime
  - only the "pure" IO tasks, like reading a range of data from S3, are spawned into io-runtime
  - relatively CPU intensive tasks like deserialization, decompression etc are kept in the query runtime
  - the tasks of reading table snapshot and segment are not issued to io-runtime yet, will be addressed in upcoming PRs
  
`Ontime` benchmark  (EC2 + S3)

| Command | Mean [ms] (branch-main) | Mean(dedicated-io-runtime-enabled) [ms] |
|:---|---:|---:|
| bench_0.sql  | 253.2 ± 18.7   |  193.2 ± 6.8    |
| bench_1.sql  | 402.8 ± 35.0   |  286.9 ± 33.4   |
| bench_2.sql  | 571.2 ± 145.2  |  372.9 ± 11.3   |
| bench_3.sql  | 231.2 ± 17.8   |  162.6 ± 30.5   |
| bench_4.sql  | 263.9 ± 35.1   |  184.2 ± 21.0   |
| bench_5.sql  | 603.0 ± 26.6   |  482.3 ± 4.9    |
| bench_6.sql  | 645.6 ± 43.6   |  474.4 ± 6.4    |
| bench_7.sql  | 744.2 ± 95.4   |  536.6 ± 37.5   |
| bench_8.sql  | 413.7 ± 130.3  |  344.0 ± 5.3    |
| bench_9.sql  | 781.4 ± 109.7  |  531.3 ± 23.7   |
| bench_10.sql | 637.6 ± 42.7   |  470.9 ± 28.1   |
| bench_11.sql | 3604.9 ± 119.7 |  3045.5 ± 214.6 |
| bench_12.sql | 1766.9 ± 43.2  |  1572.6 ± 18.8  |

## Changelog

- Improvement

## Related Issues

Fixes #4403

## Test Plan

Unit Tests

Stateless Tests


query/Cargo.toml
```diff
@@ -88,6 +88,7 @@ opendal = "0.2.0"
 parquet-format-async-temp = "0.2.0"
 paste = "1.0.6"
 petgraph = "0.6.0"
+pin-project = "1.0.10"
```
dantengsky, line 91: no longer need this

query/Cargo.toml
```diff
@@ -102,6 +103,7 @@ threadpool = "1.8.1"
 tokio-rustls = "0.23.2"
 tokio-stream = { version = "0.1.8", features = ["net"] }
 toml = "0.5.8"
+tokio-util = "0.7.0"
```
dantengsky, line 106: no longer need this

dantengsky:
@zhang2014 
If this PR conflicts with other ongoing refactorings, please let me know. I'd like to withdraw or postpone it. :-)

query/src/sessions/session_mgr.rs
```diff
@@ -69,6 +71,14 @@ impl SessionManager {
         // Cluster discovery.
         let discovery = ClusterDiscovery::create_global(conf.clone()).await?;
 
+        let storage_runtime = {
+            let mut storage_num_cpus = conf.storage.storage_num_cpus as usize;
+            if storage_num_cpus == 0 {
+                storage_num_cpus = std::cmp::max(1, num_cpus::get() / 2)
```
sundy-li, line 77: why it defaults to half of the CPU, I think it might be higher than worker threads num?

dantengsky: > I think it might be higher than worker threads num?

Not sure.

if underlying io tasks are executed in a blocking way, or some CPU-intensive tasks, let say, decompression, deserialization is also executed in these threads, it might be better to spawn more threads (when the physical resource is available).

In our case, IO operations are non-block, typically io threads just issue connect/read/write/etc, and wait them to be completed. Spawning lots of threads may not help.

Even worse, If the total num of io and computing threads is too larger(than the num of cores),  io and computing tasks are more likely to be scheduled in the same physical core -- which is this PR trying  to avoid.

> why it defaults to half of the CPU

This is an arbitrary number :D, during the benchmark(settings.max_threads = num_cpu), it seems to be a good/balanced num for the cases (but not the best for all the cases).  it would be better if this num could be chosen adaptively though。

sundy-li (approved):

BohuTANG:
/lgtm

Nice work!

query/src/storages/fuse/io/block_reader.rs
```diff
@@ -44,13 +50,15 @@ pub struct BlockReader {
...
     parquet_schema_descriptor: SchemaDescriptor,
+    ctx: Arc<QueryContext>,
 }
 
 impl BlockReader {
     pub fn create(
         operator: Operator,
         schema: DataSchemaRef,
         projection: Vec<usize>,
+        ctx: Arc<QueryContext>,
```
BohuTANG, line 61: Put the ctx as the first parameter of the function?
That's OK to fix it in the later PR :)

dantengsky: :ok_hand:

PR #4428, opened 2022-03-13
ISSUE-4292: limit pushdown in read plan

### LINKED ISSUE #4292: Feature:  Offset/Limit Pushdown for fuse engine
**Summary**

Description for this feature.

```
select * from ontime limit 10;
```

This SQL has not filters and it just need first 10 rows, so it's a simple and common query.  But it will fetch all partition data now.

dantengsky:

## Summary

WIP
Fixe issue: #4292

## Changelog

- New Feature
- Improvement

## Related Issues

Fixes #4292

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/operations/read_partitions.rs
```diff
@@ -94,32 +95,71 @@ impl FuseTable {
...
-            statistics.read_rows += block_meta.row_count as usize;
+            statistics.read_rows += rows;
             statistics.read_bytes += block_meta.block_size as usize;
+
+            if remaining > rows {
+                remaining -= rows;
+            } else {
+                // the last block we shall take
+                if remaining != rows {
+                    statistics.is_exact = false;
```
sundy-li, line 119: is_exact should be always true?

Because the Statistics are for IO cost before the filter.

dantengsky: not sure about the "semantic" of `statistics.is_exact`,  according to the method `is_exact`

https://github.com/datafuselabs/databend/blob/fb6ea13630bf9e2ccd113d05ec74a02238d8934b/query/src/storages/fuse/operations/read_partitions.rs#L89-L95

It seems that if the plan returned might scan more rows, `is_exact` should be false.

cc @zhang2014, need your help , please take a look

BohuTANG: I think the `exact` only used for the count:
https://github.com/datafuselabs/databend/blob/fb6ea13630bf9e2ccd113d05ec74a02238d8934b/query/src/optimizers/optimizer_statistics_exact.rs#L55-L77

If the statistics is exact, the count will return from the statictics instead of the underlying table.

sundy-li: is_exact represents the IO statistics is exact or not.

I think is_exact should be always true in fuse engine even with filter.

zhang2014: `Select COUNT() ...` will use statistics. In this case, is_exact is false

BohuTANG: > is_exact represents the IO statistics is exact or not.
> 
> I think is_exact should be always true in fuse engine even with filter.

If the statistics is approximately, we set `is_exact` to true, `SELECT COUNT from fuse_table` will be incorrect(it's  approximately count).

zhang2014 (approved):

query/src/storages/fuse/operations/read_partitions.rs
sundy-li: > SELECT COUNT from fuse_table will be incorrect.

We will not use the statistics if the SQL filter exists. But the io statistics are exactly true.

is_exact means we generate blocks with exact statistics, it's storage level statistics.

Filter is query level.

dantengsky: > We will not use the statistics if the SQL filter exists

@BohuTANG  I think that @sundy-li  means the following case:


1. `select count(*) from t limit N`
    **NO** limit pushdown will be passed to `read_plan`
    
please correct me if I'm mistaken @sundy-li     

Some other common cases for your consideration:
  
2. `select * from t limit N`
    "limit N" will be passed to `read_plan`
3. `select c from t where c> X limit N` (no aggregation, no order) 
    "limit N" will be passed to `read_plan`
4. `select count(*), c from t group by c limit N` 
    "limit N" will **NOT* be passed to `read_plan`

BohuTANG: `is_exact means we generate blocks with exact statistics, it's storage level statistics.`

Now, the `is_exact` usage is not like this(the only usage is for `count` in optimizer), maybe we can address other features in the future :/

BohuTANG: > > We will not use the statistics if the SQL filter exists
> 
> @BohuTANG I think that @sundy-li means the following case:
> 
> 1. `select count(*) from t limit N`
>    **NO** limit pushdown will be passed to `read_plan`
> 
> please correct me if I'm mistaken @sundy-li
> 
> Some other common cases for your consideration:
> 
> 2. `select * from t limit N`
>    "limit N" will be passed to `read_plan`
> 3. `select c from t where c> X limit N` (no aggregation, no order)
>    "limit N" will be passed to `read_plan`
> 4. `select count(*), c from t group by c limit N`
>    "limit N" will *_NOT_ be passed to `read_plan`

Oh, I miss this comments.
The `is_exact` flag is not enough? Seems we should add some other like 'count_is_exact' and '[other]_is_exact'?

sundy-li: @dantengsky , yes, that's what I mean.

`limit N` will passed into ReadPlan only it have meaningful information.

- `Group By` limit will not pass, because limit is meaningful to the data after group by | order by, not after read.
- `Filter`  limit will pass, because the engine may apply the filter by index and apply limit

dantengsky: > The is_exact flag is not enough? Seems we should add some other...

Seems enough for now. let's add it or refactor it on demand.

PR #4432, opened 2022-03-14
 support for super user

youngsofun:

## Summary

temporary support for super user 

1.  only  auth  by JWT 
2.  temporary bypass current Privilege system (fixed Privileges, no need to store/get in meta)

@flaneur2020 @ZhiHanZ 

by adding is_superuser field,  we have a sep user namespace for  superusers.
this is the way with minimal code change. 

btw, we already has an UserPrivilegeType called `Super`

```
    // Privilege to Kill query, Set global configs, etc.
    Super = 1 << 9,
```

- [ ] add unit test

## Changelog

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/users/auth/auth_mgr.rs
+++ b/query/src/users/auth/auth_mgr.rs
@@ -56,11 +56,15 @@ impl AuthMgr {
     pub async fn auth(&self, credential: &Credential) -> Result<UserInfo> {
         match credential {
             Credential::Jwt { token: t } => {
-                let user_name = match &self.jwt {
+                let (user_name, is_supper) = match &self.jwt {
                     Some(j) => j.get_user(t.as_str())?,
                     None => return Err(ErrorCode::AuthenticateFailure("jwt auth not configured.")),
                 };
-                self.users.get_user(&self.tenant, &user_name, "%").await
+                if is_supper {
+                    Ok(UserInfo::new_super(user_name))
+                } else {
+                    self.users.get_user(&self.tenant, &user_name, "%").await
+                }
             }
             Credential::Password {
                 name: n,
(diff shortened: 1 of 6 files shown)


query/src/users/auth/auth_mgr.rs
```diff
@@ -56,11 +56,15 @@ impl AuthMgr {
...
         match credential {
             Credential::Jwt { token: t } => {
-                let user_name = match &self.jwt {
+                let (user_name, is_supper) = match &self.jwt {
                     Some(j) => j.get_user(t.as_str())?,
                     None => return Err(ErrorCode::AuthenticateFailure("jwt auth not configured.")),
                 };
-                self.users.get_user(&self.tenant, &user_name, "%").await
+                if is_supper {
+                    Ok(UserInfo::new_supper(user_name))
```
flaneur2020, line 64: maybe the super user's granted privileges would be empty in this case 🤔

some unit tests may help us making the semantics of superuser more clear

youngsofun: > maybe the super user's granted privileges would be empty in this case 🤔
it was default(), I just changed to empty()

flaneur2020: maybe the superuser has more privileges than root?  q.q

if it's root user, it'd get all the privileges

youngsofun: I though superuser do not use privileges at all.

flaneur2020:
how about limit the scope of the super marker, instead add an user option like `can modify tenant`?

ZhiHanZ:
Additionally, this PR is related to daily ops, and need some example documentations about JWT and how to access databend-query as super user with JWT Authentication

Issue #4354, opened 2022-03-08
Feature: add /v1/status uri to show real time tenant, cluster, S3, and latest query endtime status


ZhiHanZ:
**Summary**
It would be nice to have it for management and debugging

ZhiHanZ:
cc @youngsofun , do you want to take this issue?

flaneur2020:
we can also consider adding a field into the /api/v1/health handler:

https://github.com/datafuselabs/databend/blob/main/query/src/api/http/v1/health.rs#L20

flaneur2020:
we can also consider adding a field into the /api/v1/health handler:

https://github.com/datafuselabs/databend/blob/main/query/src/api/http/v1/health.rs#L20

youngsofun:
cluster_id is directly for from config.
tenant also from config, can only change in for  a session

can wait for they can be update？
maybe we need a global setting/context, instead of construct everything from conf directly

---

 last queried time is ok
 can store it as an atomic in global session_manager (maybe in global setting/context later)
 so /api/v1/health is will not fail or block.

and I can return durion since ` last queried time`  to avoid problem of client diff clock time
@flaneur2020

PR #4454, opened 2022-03-15
query add http api /v1/status

youngsofun:

## Summary

related issue https://github.com/datafuselabs/databend/issues/4354

add /v1/status which return  `running_queries_count, last_query_started_at, last_query_finished_at`, may add other fields later.

```
pub struct Status {
    pub running_queries_count: u64,
    // secs since epoch
    pub last_query_started_at: Option<u64>,
    // secs since epoch
    pub last_query_finished_at: Option<u64>,
}
```

to make it lightweight, maintain the info in sessionManager and update on each query start/finish.

to ignore queries issued by admin api, change session type from string to enum. making it more useful.

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4354

## Test Plan

Unit Tests

Stateless Tests


query/src/api/http/v1/status.rs
```diff
@@ -0,0 +1,57 @@
...
+use poem::web::Data;
+use poem::web::Json;
+use poem::IntoResponse;
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::sessions::SessionManager;
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Debug)]
+pub struct Status {
```
ZhiHanZ, line 27: should we also cover tenant_id and cluster_id?  since those things update dynamically

youngsofun: ok I will check it

youngsofun: @ZhiHanZ 

> since those things update dynamically

do you mean reload_config,  so you can get  tenant_id and cluster_id through /v1/config?

query/src/api/http/v1/status.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub running_queries_count: u64,
+    // secs since epoch
+    pub last_query_started_at: Option<u64>,
+    // secs since epoch
+    pub last_query_finished_at: Option<u64>,
```
flaneur2020, line 32: how about add a instance_started_at timestamp to record when the query instance started.

if a query instances started but not getting any query, the last_query_started_at and last_query_finished_at would both be empty, but it's also worth triggering the auto suspend operator

youngsofun: ok

query/src/api/http/v1/status.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+fn secs_since_epoch(t: SystemTime) -> u64 {
+    t.duration_since(SystemTime::UNIX_EPOCH)
+        .expect("Time went backwards")
```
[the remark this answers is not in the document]
youngsofun: we use this in multi place in data bend.  resolve it as another issue?

query/src/api/http/v1/status.rs
youngsofun: done

query/src/api/http/v1/status.rs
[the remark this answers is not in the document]
flaneur2020: ok to me, we may replace it with a default value on Err to replace the `expect`

ZhiHanZ (approved):
nice work

Issue #4458, opened 2022-03-15
Feature: separate log file for query log


everpcpc:
**Summary**

We need a separate log file for better aggregating query logs.

PR #4462, opened 2022-03-15
feat: add query logger to session manager

everpcpc:

## Summary

Add a rolling file query logger to session manager.

## Changelog

- Improvement

## Related Issues

Fixes [#4458](https://github.com/datafuselabs/databend/issues/4458)

## Test Plan

Unit Tests

Stateless Tests


common/tracing/src/logging.rs
```diff
@@ -113,6 +113,29 @@ pub fn init_global_tracing(app_name: &str, dir: &str, level: &str) -> Vec<Worker
     guards
 }
 
+pub fn init_query_logger(
+    log_name: &str,
+    dir: &str,
+) -> (Vec<WorkerGuard>, Arc<dyn Subscriber + Send + Sync>) {
+    let mut guards = vec![];
+
+    let rolling_appender = RollingFileAppender::new(Rotation::HOURLY, dir, log_name);
```
ZhiHanZ, line 122: I think meta service also needs the rolling appender, maybe add it in further PR

query/src/sessions/session_mgr.rs
```diff
@@ -90,6 +95,15 @@ impl SessionManager {
         let active_sessions = Arc::new(RwLock::new(HashMap::with_capacity(max_sessions)));
         let status = Arc::new(RwLock::new(Default::default()));
 
+        let (_guards, query_logger) = if conf.log.log_query_enabled {
+            let query_log_name = format!("query-detail-{}", conf.query.cluster_id,);
```
ZhiHanZ, line 99: what would be happened during reload when cluster id got changed.
I think file name could some thing not related to cluster id, tenant id, just relate to timestamp? since we already provide cluster and tenant inside each logging message

ZhiHanZ (approved):

BohuTANG (approved):

PR #4473, opened 2022-03-16
ISSUE-4275, metactl: dump data from a running metasrv

### LINKED ISSUE #4275: Feature: metactl: dump data from a running metasrv
`databend-metactl --export --raft-dir <path>` is used to export meta data for backup when the metasrv is closed.
Running this command with a running metasrv just results in a locking error.

New feature request: `databend-metactl --export --port <port>` to export meta data from a running metasrv on local host.

This can be done by calling the grpc API `export`:
https://github.com/datafuselabs/databend/blob/3713168845ecd80e02d52db386623806a10982d1/metasrv/src/api/grpc/grpc_service.rs#L150-L163

light4:

## Summary

1. `metactl`: dump data from a running `metasrv`
2. fix export panic if data length is smaller than chunk
3. add reflection to grpc service, so we can test it using tools like https://github.com/fullstorydev/grpcurl, will remove it if not needed

## Changelog

- New Feature
- Bug Fix
- Improvement
- Build/Testing/CI
- Other 

## Related Issues

Fixes #4275

## Test Plan

Unit Tests

Stateless Tests


tools/metactl/src/main.rs
```diff
@@ -72,6 +79,14 @@ async fn main() -> anyhow::Result<()> {
 
     eprintln!("raft_config: {}", pretty(raft_config)?);
 
+    // export from grpc api if raft is running
+    let addr = raft_config.raft_api_addr().await?.to_string();
+    if config.export && service_is_running(&addr).await? {
+        eprintln!("export meta from: {}", &config.grpc_api_address);
+        grpc::export_meta(&config.grpc_api_address).await?;
```
drmingdrmer, line 86: Why not just check `grpc_api_address` but `raft_api_addr()`?

BTW it exports from a running metasrv if and only if `grpc_api_address` is specified.

tools/metactl/src/main.rs
```diff
@@ -169,3 +184,12 @@ fn print_meta() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+// if port is open, service is running
+async fn service_is_running(addr: &str) -> anyhow::Result<bool> {
```
drmingdrmer, line 189: `anyhow::Result` will spread everywhere. Only use it at the top-level entry.

Here a specific error type would be better.

tools/metactl/src/grpc.rs
```diff
@@ -0,0 +1,40 @@
...
+    let mut lines = vec![];
+    while let Some(chunk_res) = stream.next().await {
+        let chunk = chunk_res?;
+
+        lines.extend_from_slice(&chunk.data);
+    }
+
+    for line in lines {
+        println!("{}", line);
+    }
```
drmingdrmer, line 37: The lines could be just printed without caching them.

drmingdrmer (commented):
Thanks man!

Reflection server is an interesting feature!

There should be a test to ensure this new API works as expected.
Other test about metactl is in `tests/metactl/test-metactl.sh`.

tools/metactl/src/grpc.rs
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common_meta_grpc::MetaGrpcClient;
+use common_meta_types::protobuf::Empty;
+use tokio_stream::StreamExt;
+
+pub async fn export_meta(addr: &str) -> anyhow::Result<()> {
+    let client = MetaGrpcClient::try_create(addr, "root", "xxx", None, None).await?;
```
BohuTANG, line 20: password need parameter setting?

tools/metactl/src/main.rs
light4: Should use `grpc_api_address` indeed.
What if `metasrv` is running and `grpc_api_address` is not specified.

tools/metactl/src/grpc.rs
light4: I find this use case in other codes. Wondering it needs an empty user/password or this(root/xxx) if connected from localhost.

tools/metactl/src/main.rs
drmingdrmer: If  `grpc_api_address` is invalid metasrv won't run at all.

Besides, if `grpc_api_address` is invalid, there is no way to export data from it.

tools/metactl/src/main.rs
light4: Updated

tools/metactl/src/main.rs
light4: Updated, and a basic test.

tests/metactl/test-metactl.sh
```diff
@@ -16,3 +16,14 @@ cat $meta_json \
...
+# test export from grpc
+chmod +x ./target/debug/databend-meta
+./target/debug/databend-meta --single &
+METASRV_PID=$!
+echo $METASRV_PID
+sleep 0.5
+
+./target/debug/databend-metactl --export
+
+kill $METASRV_PID
```
drmingdrmer, line 29: The output of `./target/debug/databend-metactl --export` should be redirected to a file and the content of the file should be checked.

light4: I wonder if `databend-meta --single` exported data is identical, if so I'll use that.

drmingdrmer: I can not tell if it is identical. But there should be at least a record like this:
```
["test-29000-raft_state",{"RaftStateKV":{"key":"Id","value":{"NodeId":0}}}]
```

light4: I find it's not identical. So just check `["raft_state",{"RaftStateKV":{"key":"Id","value":{"NodeId":0}}}]` in exported file.
Wondering if it's ok.

ariesdevil (approved):

PR #4481, opened 2022-03-17
refactor: Adapt reqsign and separate IO runtime

Xuanwo:

## Summary

This is a draft to address hyper panic at `dispatch dropped`

Major changes:

- uses [reqsign](https://github.com/Xuanwo/reqsign) to replace AWS SDK.
- use separate IO runtime for all opendal operations.

Works that to be done:

- [x] use tagged version of opendal
- [x] make sure all test cases passed
- [x] make sure `dispatch dropped` not happen again
- [x] make sure no performance regression.

This PR is a draft now, I will fix all other issues after hyper dispatch dropped panic is verified to be gone.

## Changelog

- Bug Fix

## Test Plan

Unit Tests

Stateless Tests


Xuanwo:
Ping @dantengsky to make sure there is no performance regression.

query/src/storages/fuse/operations/commit.rs
```diff
@@ -90,7 +90,7 @@ impl FuseTable {
                     match backoff.next_backoff() {
                         Some(d) => {
                             let name = tbl.table_info.name.clone();
-                            tracing::error!(
+                            tracing::warn!(
```
Xuanwo, line 93: @dantengsky Is it a good idea to make this log a `warning` instead of `error`?

dantengsky: sounds good to me

Xuanwo:
All concerns have been addressed, let's take a review and merge it!

BohuTANG:
We can test the parallelism insert based on this patch again, cc @wubx @TCeason

common/contexts/Cargo.toml
```diff
@@ -10,4 +10,6 @@ test = false
 
 [dependencies]
 async-trait = "0.1.52"
-opendal = "0.2.3"
+opendal = "0.2.4"
+time = "0.3.7"
+tokio = "1.17"
```
BohuTANG, line 15: I think we should use common/base/runtime tokio instead of the tokio, then we can only maintain one tokio version.

query/src/sessions/query_ctx.rs
```diff
@@ -394,7 +390,7 @@ impl QueryContextShared {
     pub(in crate::sessions) fn destroy_context_ref(&self) {
         if self.ref_count.fetch_sub(1, Ordering::Release) == 1 {
             std::sync::atomic::fence(Acquire);
-            tracing::debug!("Destroy QueryContext");
+            tracing::warn!("Destroy QueryContext");
```
dantengsky, line 393: is it a temporary "upgrade" of log level? for the testing of this PR only? : )

Xuanwo: Let's me roll it back.

Xuanwo:
@BohuTANG @dantengsky All comments have been addressed.

zhang2014 (approved):

PR #4497, opened 2022-03-18
ISSUE-4413: impove performance of insert-into literal values

### LINKED ISSUE #4413: Feature: Improve insert into performance
**Summary**

Description for this feature.

Query:

```
INSERT INTO `books` VALUES ('Transaction Processing','Jim Gray', 1992),('Readings in Database Systems','Michael Stonebraker', 2004);
```

The entire SQL goes into SQL parser and expression-executor, because we can have expression inside VALUES, eg:

```
INSERT INTO `books` VALUES ('Transaction Processing','Jim Gray', now() - 3),('Readings in Database Systems','Michael Stonebraker', 2004);
```

To support expressions, this insert SQL works really slow. We can improve the performance by:

1. Use [source_values](https://github.com/datafuselabs/databend/blob/v0.6.26-nightly/common/streams/src/sources/source_values.rs) by default.
2. If error happens, fall back to current way.

ygf11:

## Summary

1. Using `ValueSource` to parse the values first.
2. When the first failed, try to use `Expression Executor`. 

- [ ] 1. Add skip values option to `parse_insert` in sqlparser.  
- [ ] 2. Return datablocks generated by `ValuesSource` or return `Vec<Vec<Expr>>` in `DfParser.parse_insert()`.

## Changelog

- Improvement
- Performance Improvement

## Related Issues

Fixes #4413 

## Test Plan

Unit Tests

Stateless Tests


ygf11:
I do a simple bench in my pc. It seems improve a lot.

`create table`:
```
CREATE TABLE `User` (name VARCHAR(64), age UInt8, country VARCHAR(64), born_time DateTime);
```

`insert(all is literal)`:
```
INSERT INTO `User` VALUES ('name', 10 , 'China', '2022-03-15 00:00:00');
```

In standalone mode(Memory Table):
1. Insert `10000` records, this pr costs `0.12s` when main costs `0.23s`
2. Insert `100000` records, this pr costs `0.91s` when main costs `2.13s`

In cluster-3-node mod(Fuse Table):
1. Insert `10000` records, this pr costs `1.12s` when main costs `3.76s`
2. Insert `100000` records, this pr costs `10.88s` when main costs `38.94s`

sundy-li:
> I do a simple bench in my pc. It seems improve a lot.

Good job, but it seems not enough.

1. Why cluster mode works slower than standalone ? @dantengsky 

2. The values big string goes into the SQL parser and we cloned the whole String to parser it in values source. 
We can find a way to avoid the values SQL parser and pass the string slice into values source.

query/src/sql/statements/statement_insert.rs
```diff
@@ -164,7 +164,11 @@ impl DfInsertStatement {
             value_exprs.push(exprs);
         }
 
-        Ok(InsertInputSource::Expressions(value_exprs))
+        let values = format!("{}", values);
```
sundy-li, line 167: `values` can be  very large, it's expensive to parse the values string to Expr::Values and use format to convert it into String back.

ygf11:
> Why cluster mode works slower than standalone ? 

@sundy-li  Sorry, I made a mistake, it's the result of debug build.

After I change to release build, the result is the same as standalone.

ygf11:
> The values big string goes into the SQL parser and we cloned the whole String to parser it in values source.
We can find a way to avoid the values SQL parser and pass the string slice into values source.

How about parsing the total insert-sql string in value source? 

1. We can get sql via `QueryContext.get_query_str` in `InsertInterpreter`, since it has been cloned before.
2. We avoid clone or format in analyze stage.
3. We just parse simple insert-sql here, so it is fast.

I also think about get values-string-slice from sqlparser when parsing values expr, 
But it will introduce lifetime here, which will broadcast in PlanNode and Interpreter.
```
pub enum InsertInputSource<'a> {
    SelectPlan(Box<PlanNode>),
    Expressions(&'a str, Vec<Vec<Expression>>),
    StreamingWithFormat(String),
}
```

sundy-li:
> How about parsing the total insert-sql string in value source?

Yes, it can work , but it needs the parser to depend on `ValueSource`, but let's try this way if we did not found better idea.

dantengsky:
> 1. Why cluster mode works slower than standalone ? @dantengsky

Insertion of literal values is executed in a non-distributed way (by the query-node which accepts the insert statement), it should not work much slower than standalone mode.

dantengsky (approved):
LGTM

BohuTANG:
@mergify update

ygf11:
@sundy-li @dantengsky @BohuTANG I add tasks that will do next about this PR, please take a look.

BohuTANG:
@mergify update

ygf11:
Maybe we should move to lifetime mark again. 

Current plan is not good enough which do `ValueSource` parse in `DfParser`, 
1. `ValueSource` depends on table schema, we need introduce `QueryContext` to `DfParser`. 
2. All DfParse's codebase is `sync`, and `ctx.get_table()` is `async`.
3. Geting and checking table info in `analyze` stage is more reasonable.

When Moving to lifetime mark, the tasks will be like:

1. Get `values-str` reference in `parser` stage, and save in `DfInsertStatement<'a>`.
2. Parse `valuesSource` or `fallback` to expression in `analyze` stage(Depends on DfParser if `ValuseSource` failed).
3. Append datablocks directly or do `expresion-transform`.

Also meets problem.... 

Parser will parse multiple statment in `parser.parse_sql_with_dialect()`.
As adding `sql-life-time` in DfStatement, say `DfStatement<'a>`, `rustc` will report `immutable-borrow after muttable-borrow` error, it is reasonable. 
```rust
loop {
            // ignore empty statements (between successive statement delimiters)
            while parser.parser.consume_token(&Token::SemiColon) {// --- immutable borrow of Self in next loop
                expecting_statement_delimiter = false;
            } 

            if parser.parser.peek_token() == Token::EOF {
                break;
            }
            if expecting_statement_delimiter {
                return parser.expected("end of statement", parser.parser.peek_token());
            }

            let statement = parser.parse_statement(sql)?; // --- mutable borrow of Self
            stmts.push(statement);// hold reference here
            expecting_statement_delimiter = true;
        }
```
This is because `DfStatement<'a>` is derived from `DfParser.sql` in complier's view. 

I found a way to resolve it. We can simple mark `DfStatement<'a>` is derived from `sql-str`, like:
```
// new parse function
fn parse_xxx(&self, sql:&'a str) -> DfStatement<'a> {
  ....
}

// old parse function
fn parse_xxx(&self) -> DfStatement<'a> {
   ... 
}
```  

This works, however it will introduce a lot unused function args(`sql`) in `DfParser`. I do not find a better way about it.

Do you have any suggestion?

Issue #4518, opened 2022-03-21
Consider populates meta caches during append data to fuse table


dantengsky:
**Summary**

pros:

 better performance of accessing the fresh data 

cons:

  may evict cache items unnecessarily

dantengsky:
addressed by PR #4519

PR #4514, opened 2022-03-20
fix(clickhouse): to_clickhouse_block always convert to full column if constant

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR

to_clickhouse_block always convert to full column if constant

## Changelog
 
- Bug Fix
 
## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests

PR #4519, opened 2022-03-21
ISSUE-4518: Feat populate cache while writing

dantengsky:

## Summary

- populates meta caches during append data to fuse table
   - issue #4518 
-  case insensitive comparison of op name in the `optimizer_statistics_exact.rs`
   - which enables case-insensitive `select CoUnT`, based on PR #4514,  
## Changelog

- Improvement

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/operations/append.rs
```diff
@@ -59,6 +60,8 @@ impl FuseTable {
         .await;
 
         let locs = self.meta_locations().clone();
+        let segment_info_cache = ctx.get_storage_cache_manager().get_table_segment_cache();
```
BohuTANG, line 63: What's the cache path?

dantengsky: later, at line num 78 (of the append.rs), the path of the segment info cache item is given :
   ~~~
                        if let Some(ref cache) = segment_info_cache {
                            let cache = &mut cache.write().await;
                            cache.put(seg_loc, seg);
                        }
  ~~~

PR #4537, opened 2022-03-22
query: Add region config back

Xuanwo:

## Summary

Some users may have restrictions to regions that we can't auto-detect. This adds the region back so that users can specify it.

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


query/src/sessions/session_mgr.rs
```diff
@@ -341,6 +341,11 @@ impl SessionManager {
                     builder.endpoint(&s3_conf.endpoint_url);
                 }
 
+                // Region
+                {
+                    builder.region(&s3_conf.region);
```
BohuTANG, line 346: It would be better to check the region is empty or not:
```
!if se_config.region.is_empty(){
  builder.region(&s3_conf.region);
}
```

Xuanwo: > It would be better to check the region is empty or not:
> 
> ```
> if !s3_config.region.is_empty(){
>   builder.region(&s3_conf.region);
> }
> ```

opendal will handle all those work.

PR #4538, opened 2022-03-22
support server-side-session for HTTP handler

youngsofun:

## Summary

support server-side-session for HTTP handler

it is useful for `use` or `set` commands.

to use it
1. for the first query, specify `session.max_idle_time` in secs
2. for the following query, use `session_id` from response to specify `session.id` .  Some Notes:
    1. when `id` field is used, other fields in `session` is invaid.
    2. currently not support queueing in server side, make sure last query response status is Succeeded or Failed before post, otherwise thHTTPtp request will fail with error message `last query on the session not finished`
    
    
Notes:

1.  exist code not affected
1.  timeout for sessions is also needed for clickhouse/MySQL, 
  but they can be implemented more directly(maybe part of `mysql-srv`), 
  so I try to minimize intrusiveness with `Expirable` trait and limit most code to the http mod,

Summary about this PR

## Changelog

- New Feature

No issue

## Test Plan

Unit Tests

Stateless Tests

### DIFF
--- a/query/src/interpreters/interpreter_factory_interceptor.rs
+++ b/query/src/interpreters/interpreter_factory_interceptor.rs
@@ -81,6 +81,7 @@ impl Interpreter for InterceptorInterpreter {
     async fn finish(&self) -> Result<()> {
         let session = self.ctx.get_current_session();
         let now = SystemTime::now();
+        session.get_status().write().query_finish();
         if session.get_type().is_user_session() {
             session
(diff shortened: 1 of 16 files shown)


youngsofun:
@flaneur2020 if this pr is ok,  we can merge it first, I may add docs for it later in another pr, since the whole HTTP handler doc is missing now.

PR #4606, opened 2022-03-28
ISSUE-4558: Add `check_json` function

### LINKED ISSUE #4458: Feature: separate log file for query log

### LINKED ISSUE #4558: Feature: Semi-structured data function check_json
**Summary**

Add support Semi-structured Data Function `CHECK_JSON` to check the validity of a JSON document.

https://docs.snowflake.com/en/sql-reference/functions/check_json.html

kevinw66:

## Summary

Add `check_json` function

## Changelog

- New Feature

## Related Issues

Fixes #4558 

## Test Plan

Unit Tests

Stateless Tests


kevinw66:
@b41sh I don't know if the testcase for `NullColumn` is correct or not, please help review it

kevinw66:
Still has some problem, Please wait

kevinw66:
> Still has some problem, Please wait

The console output for `null` has different case.
```
mysql> select check_json('123.12');
+----------------------+
| check_json('123.12') |
+----------------------+
| null                 |
+----------------------+
```

```
mysql> select check_json(v), v from t1;
+---------------+-----------+
| check_json(v) | v         |
+---------------+-----------+
| NULL          | NULL      |
| NULL          | null      |
| NULL          | true      |
| NULL          | 123       |
| NULL          | "abc"     |
| NULL          | [1,2,3]   |
| NULL          | {"a":"b"} |
+---------------+-----------+
```

common/functions/src/scalars/semi_structureds/check_json.rs
```diff
@@ -0,0 +1,151 @@
...
+        &*self.display_name
+    }
+
+    fn return_type(&self, args: &[&DataTypePtr]) -> Result<DataTypePtr> {
+        if args[0].data_type_id() == TypeID::Null {
+            return Ok(NullType::arc());
+        }
+
+        if args[0].is_nullable() {
+            return Ok(Arc::new(NullableType::create(VariantType::arc())));
```
b41sh, line 60: Regardless of whether the input value `is_nullable` or not, the type of return value should be a nullable `StringType`, not `VariantType'. When the value is `Null` or a valid JSON return `Null`, otherwise return an error message string

b41sh (changes requested):

b41sh:
> > Still has some problem, Please wait
> 
> The console output for `null` has different case.
> 
> ```
> mysql> select check_json('123.12');
> +----------------------+
> | check_json('123.12') |
> +----------------------+
> | null                 |
> +----------------------+
> ```
> 
> ```
> mysql> select check_json(v), v from t1;
> +---------------+-----------+
> | check_json(v) | v         |
> +---------------+-----------+
> | NULL          | NULL      |
> | NULL          | null      |
> | NULL          | true      |
> | NULL          | 123       |
> | NULL          | "abc"     |
> | NULL          | [1,2,3]   |
> | NULL          | {"a":"b"} |
> +---------------+-----------+
> ```
@kevinw66  `select check_json('123.12');` return `null` is because you check `is_nullable()` and return return a no nullable column, so we can't get a `NULL` value. Modify function `return_type` to always return nullable column.

common/functions/tests/it/scalars/mod.rs
```diff
@@ -23,7 +23,7 @@ mod maths;
 mod nullables;
 mod others;
 mod scalar_function2_test;
-mod semi_structureds;
+mod semi_structures;
```
b41sh, line 26: Better to use `semi_structureds` here, as this is `Semi-structured Data Types`

common/functions/src/scalars/semi_structureds/check_json.rs
```diff
@@ -0,0 +1,139 @@
...
+            column = nullable_column.inner();
+        }
+
+        if column.data_type_id() == TypeID::Null {
+            return NullType::arc().create_constant_column(&DataValue::Null, input_rows);
+        }
+
+        let mut builder = NullableColumnBuilder::<Vu8>::with_capacity(input_rows);
+
+        with_match_physical_primitive_type!(column.data_type_id().to_physical_type(), |$T| {
```
b41sh, line 80: We should use `with_match_primitive_type_id` here, as both `Uint32` and `Date32` have some physical type.
Use `columns[0].field().data_type_id()` instead of `column.data_type_id()` for the same reason.

b41sh (changes requested):

common/functions/src/scalars/semi_structureds/check_json.rs
```diff
@@ -0,0 +1,139 @@
...
+                        if let JsonValue::String(s) = v {
+                            match serde_json::from_str::<JsonValue>(s.as_str()) {
+                                Ok(_v) => builder.append_null(),
+                                Err(e) => builder.append(e.to_string().as_bytes(), true),
+                            }
+                        } else {
+                            builder.append_null()
+                        }
+                    }
+                }
```
b41sh, line 122: Date16, Date32, DateTime32, DateTime64, Interval should convert to string to check json

kevinw66: will date be an invalid json string?

b41sh: the date is not a valid json string, `serde_json::from_str` will return error

b41sh: The date is originally a string, and we convert this string into a number when we store it, so we need to convert it back to a string here. As this string is unquoted, so converting to JSON will always fail.
In order to simplify the processing, I think it is ok not to convert to string, just append a fixed error message directly. something like `Date32 is not a valid JSON`

kevinw66: got it.

kevinw66: Should we append error messages or return Err.

b41sh: append error messages

b41sh (approved):

PR #4608, opened 2022-03-28
Feature: add state machine range and subscribe API

lichuang:

## Summary

Feature: add state machine range and subscribe API

## Changelog

- New Feature

## Related Issues

Fixes #issue https://github.com/datafuselabs/databend/issues/4607

## Test Plan

Unit Tests

Stateless Tests


drmingdrmer:
Why do we need range APIs for meta datas?

lichuang:
> Why do we need range APIs for meta datas?

in etcd watch API, see https://etcd.io/docs/v3.2/learning/api/#watch-api, they have the key range:

`
message WatchCreateRequest {
  bytes key = 1;
  bytes range_end = 2;
  // ...
}
`

lichuang:
> > in etcd watch API, see https://etcd.io/docs/v3.2/learning/api/#watch-api, they have the key range:
> > `message WatchCreateRequest { bytes key = 1; bytes range_end = 2; // ... }`
> 
> On the doc page: `Key, Range_End - The key range to watch.`.
> 
> If several given events happened as the following:
> 
> ```
> z:       z=3    
> y:                     y=2
> x: x=1         x=0
> ---+-----+-----+-------+------------------------------> time(revision)
>    t1    t2    t3      t4
> ```
> 
> A watch request for key range `[y,z]`, from `t3`, should only return event `y=2` at `t4`. In another word, to collect response data for `watch`, metasrv iterates all events in time order, then filters them by key range. It does not need to iterate over the key range.

I wonder if or not return the current event at the first time when a watcher has created, but in etcd doc and code, seems they do not return event when create watcher.ok, i delete the range API.

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -90,6 +90,42 @@ const SEQ_DATABASE_META_ID: &str = "database_meta_id";
 // const TREE_META: &str = "meta";
 const TREE_STATE_MACHINE: &str = "state_machine";
 
+pub enum DataChangeType {
+    Create,
+    Update,
+    Drop,
+}
```
drmingdrmer, line 97: It may need a `Rename` operation type in it.

A rename is a bit different because it relates to two identifiers of an instance.

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -90,6 +90,42 @@ const SEQ_DATABASE_META_ID: &str = "database_meta_id";
...
+/// StateMachine subscriber trait
+#[async_trait::async_trait]
+pub trait StateMachineSubscriber: Debug + Sync + Send {
+    async fn database_changed(
+        &self,
+        change_type: DataChangeType,
+        key: &DatabaseLookupKey,
+        prev: &Option<SeqV<DatabaseMeta>>,
+        meta: &Option<SeqV<DatabaseMeta>>,
+    );
```
drmingdrmer, line 108: If possible, a subscriber may not need to be a `trait`, IMHO.

A concrete type may be easier to use.

Just an idea:
If it has to be a `trait`. A trait with type parameters will be more generalized and extensible.

```rust
trait Subscriber<K, M> {
    async fn changed(&self,
        typ: DataChangeType,
        key: K,
        prev: &Option<SeqV<M>>,
        meta: &Option<SeqV<M>>);
}
```

lichuang: > 

In this case, may be need three instance for `Subscriber ` for database\table\kv？Why these three Subscriber unify into one trait impl instance?

drmingdrmer: > ```
> trait Subscriber<K, M> {
>     async fn changed(&self,
>         typ: DataChangeType,
>         key: K,
>         prev: &Option<SeqV<M>>,
>         meta: &Option<SeqV<M>>);
> }
> ```
> 
> In this case, may be need three instance for `Subscriber ` for database\table\kv？
> ```
> pub db_subscriber: Option<Box<dyn Subscriber<DatabaseLookupKey, DatabaseMeta>>>,
> pub table_subscriber: Option<Box<dyn Subscriber<TableUpdateKey, TableMeta>>>,
> pub kv_subscriber: Option<Box<dyn Subscriber<str, SeqV>>>,
> ```
> 
> Why these three Subscriber unify into one trait impl instance?

Right, it's not that obvious in this way.

drmingdrmer:
@lichuang 
Is this ready to review or are you gonna cleanup the commits?

drmingdrmer:
@lichuang 
Is this ready to review or are you gonna cleanup the commits?

lichuang:
> Is this ready to review or are you gonna cleanup the commits?

ready

drmingdrmer:
> I wonder if or not return the current event at the first time when a watcher has created, but in etcd doc and code, seems they do not return event when create watcher.ok, i delete the range API.

The range APIs are still there.

BTW, another issue is that range-API is not snapshot-based. The result may be inconsistent.

lichuang:
> > I wonder if or not return the current event at the first time when a watcher has created, but in etcd doc and code, seems they do not return event when create watcher.ok, i delete the range API.
> 
> The range APIs are still there.
> 
> BTW, another issue is that range-API is not snapshot-based. The result may be inconsistent.

sad, when rebase source it does not remove the range API...

drmingdrmer (approved):
LGTM

Issue #4568, opened 2022-03-25
Feature: Add ClickHouse REST handler


BohuTANG:
## Summary

Endpoint:
```
/v1/clickhouse
```

## Query
Uri:
```
/v1/clickhouse/?query=<sql>
```

Example:
```
echo 'SELECT 1' | curl 'http://localhost:8123/?query=' --data-binary @-
1
```
                                                                                                                      

## Insert

Uri:
```
/v1/clickhouse/?query=<insert into {db}.{table} format {foramt}> 

format:
1. Values
2. JSONEachRow
```

Example(`values` format):
```
echo '(7),(8),(9)' | curl 'http://localhost:8000/v1/clickhouse/?query=INSERT%20INTO%20t%20FORMAT%20Values' --data-binary @-
```

Example(`JSONEachRow` format):
```
echo '{"foo":"bar"}' | curl 'http://localhost:8000/v1/clickhouse/?query=INSERT%20INTO%20test%20FORMAT%20JSONEachRow' --data-binary @-
```

PR #4613, opened 2022-03-29
Basic clickhouse REST handler

youngsofun:

## Summary

Basic clickhouse  REST handler 

try to be compatible with [HTTP Interface | ClickHouse Documentation](https://clickhouse.com/docs/en/interfaces/http/)

currently only support:

1. select with format TSV
2. insert with format JSONEachRow. to support [vector/clickhouse.rs at master · vectordotdev/vector (github.com)](https://github.com/vectordotdev/vector/blob/master/src/sinks/clickhouse.rs)
3. insert ... values 

Like the original clickhouse HTTP handler, this handler support sending the beginning of the query in the 'query' parameter, and the rest in the POST. but with a limitition for parsing efficiency: `insert .. format JSONEachRow  <datas>` , send the `<data>`  in Body, and the `insert .. format JSONEachRow `  part in the 'query' parameter.

doc and usage will be added in next PR  if this is merged fast

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4568

## Test Plan

Unit Tests

Stateless Tests


youngsofun:
should we put this handler under /v1 or not?

query/src/servers/http/http_services.rs
```diff
@@ -49,13 +49,13 @@ impl HttpHandler {
...
 
-    // TODO(younsofun): add doc url after it`s ready
     pub fn usage(sock: SocketAddr) -> String {
+        let json = r#"{"foo": "bar"}"#;
         format!(
             r#" examples:
-curl --request POST '{:?}/v1/statement/' --header 'Content-Type: text/plain' --data-raw 'SELECT avg(number) FROM numbers(100000000)'
-curl --request POST '{:?}/v1/query/' --header 'Content-Type: application/json' --data-raw '{{"sql": "SELECT avg(number) FROM numbers(100000000)"}}'"#,
-            sock, sock
+curl --request POST '{:?}/v1/query/' --header 'Content-Type: application/json' --data-raw '{{"sql": "SELECT avg(number) FROM numbers(100000000)"}}'
```
BohuTANG, line 56: :+1:

query/src/servers/http/v1/clickhouse_handler.rs
```diff
@@ -0,0 +1,225 @@
...
+            return Ok(None);
+        }
+        if statements.len() > 1 {
+            return Err(ErrorCode::SyntaxException("Only support single query"));
+        }
+        let statement = &statements[0];
+        if let DfStatement::InsertQuery(ref insert) = statement {
+            if let Some(format) = &insert.format {
+                match format.as_str() {
+                    FORMAT_JSON_EACH_ROW => return Ok(Some((Format::TSV, statements))),
```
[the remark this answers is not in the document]
youngsofun: no, here it is input format, vector use JSON,  so we support is first.
select reply with TSV by default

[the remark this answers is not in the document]
youngsofun: fixed, thank you

query/src/servers/http/formats/tsv_output.rs
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+
+use common_datablocks::DataBlock;
+use common_exception::ErrorCode;
+use common_exception::Result;
+
+const FIELD_DELIMITER: u8 = b'\t';
+const ROW_DELIMITER: u8 = b'\n';
+
+pub fn block_to_tsv(block: &DataBlock) -> Result<Vec<u8>> {
```
sundy-li, line 22: Maybe we should move formats into a separate directory.

It can be addressed in another pr.

query/src/servers/http/v1/clickhouse_handler.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+    Ok(None)
+}
+
+// parse query params first to avoid tokenize the body part when it is a insert with format.
+async fn parse(
+    sql: &str,
+    body: Body,
+) -> PoemResult<(Option<(Format, Body)>, Cow<'_, str>, Vec<DfStatement>)> {
+    if let Some((format, statements)) = try_parse_insert_formatted(sql).map_err(BadRequest)? {
```
sundy-li, line 160: If it is not insert SQL, it will be parsed twice.
I think it's better to reuse the `statements`.

youngsofun: @sundy-li statements are already reused now, "parsed twice" only happens when need to append the body part.

query/src/servers/http/v1/clickhouse_handler.rs
```diff
@@ -0,0 +1,225 @@
...
+        .await
+        .map_err(InternalServerError)?;
+    session.set_current_user(user_info.0.clone());
+
+    let ctx = session
+        .create_query_context()
+        .await
+        .map_err(InternalServerError)?;
+
+    let plan = PlanParser::build_plan(statements, ctx.clone())
```
youngsofun, line 192: @sundy-li reused here

query/src/servers/http/formats/tsv_output.rs
youngsofun: yes, I am analyzing its requirements about it to get a better interface, maybe more thinking when adding other formats.  put it here tmp since it is simple for now.

youngsofun: for POST, parse the 'query' parameter first in case it is "insert ...  format".
if not (parse fail or not insert format ),  append the body and parse again.

query/src/servers/http/v1/clickhouse_handler.rs
youngsofun: for POST, parse the 'query' parameter first in case it is "insert ...  format".
if not (parse fail or not insert format ),  append the body and parse again.

youngsofun: I thought about reusing the tokens,  but since the  'query' parameter is not long most of the time, I wonder if it is worth it.

sundy-li: It's trivial.

Yet when the SQL is very long, users may post a select query inside the body, then it's parsed twice.

I suggest to call :

```
let (statements, _) = DfParser::parse_sql(&sql).map_err(BadRequest)?;
```

And match the statements to get the format.

youngsofun: > sql



> It's trivial.
> 
> Yet when the SQL is very long, users may post a select query inside the body, then it's parsed twice.
> 
> I suggest to call :
> 
> ```
> let (statements, _) = DfParser::parse_sql(&sql).map_err(BadRequest)?;
> ```
> 
> And match the statements to get the format.

what is the `sql`?   including the POST body part?

sundy-li: ClickHouse supported  post the sql into body instead of query params.

```
➜ curl http://127.0.0.1:8123 -d 'select * from numbers(10)'
0
1
2
3
4
5
6
7
8
9
```

youngsofun: we already support it now: https://github.com/datafuselabs/databend/pull/4613/files#diff-cf80e97711c9701ffe83aa5deded1afa6581db50518d4b25548fec8bb4d84507R79

youngsofun: the problem is that the current DfParser tokenize the whole query at first (get a vec<Token> instead of gen a token stream)
so it will be a bigger waste to parse`"query" + "body" ` if the body is a big data file.

sundy-li (approved):

PR #4632, opened 2022-03-30
ISSUE-4262: prohibits using reserved table option in create table statement.

### LINKED ISSUE #4624: `Show create table` prints the fuse engine internal information
**Summary**

create table:
```
    create table nginx_logs
    (
    RemoteAddr String,
    Client String,
    RemoteUser String,
    TimeLocal DateTime,
    RequestMethod String,
    Request String,
    HttpVersion String,
    Status Int32,
    BytesSent Int64,
    UserAgent String
    );
```

show create table:
```
mysql> show create table nginx_logs;
+------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table      | Create Table                                                                                                                                                                                                                                                                            |
+------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| nginx_logs | CREATE TABLE `nginx_logs` (
  `RemoteAddr` String,
  `Client` String,
  `RemoteUser` String,
  `TimeLocal` DateTime32,
  `RequestMethod` String,
  `Request` String,
  `HttpVersion` String,
  `Status` Int32,
  `BytesSent` Int64,
  `UserAgent` String,
) ENGINE=FUSE DATABASE_ID='1' |
```

Tasks:
1. `DATABASE_ID='1' ` is only used for fuse internal, we should not show it to the user
2. Can't allow the user to add the `DATABASE_ID` (or other internal attributes)  to the `create/alter table` statement

dantengsky:

## Summary

- `DATABASE_ID` and `SNAPSHOT_VER` are now reserved fuse table option keys
- they are not allowed to be specified in the CREATE TABLE statement
- the values of those options are hidden, for the result of SHOW CREATE TABLE statement
-----------------
MISC:

- unify dev stateless test with the corresponding CI tests
- ci-run-tests-embedded.sh does not enable table cache
  which is a better setting to discover more potential flaws
  
- no longer use table option key SNAPSHOT_VER

- infer the version of snapshot from snapshot location

  Only in the bootstrap phase, i.e. the version of snapshot that comes
  directly from TableInfo is deduced in this way.

  For the versions of history snapshots, the numeric values that kept
  in meta data file are used.

  Later, when zero-copy clone is implemented seriously, the above inconsistency
  should disappear.

## Changelog

- Bug Fix

## Related Issues

Fixes #4624

## Test Plan

Unit Tests

Stateless Tests


query/src/storages/fuse/constants.rs
```diff
@@ -27,3 +30,16 @@ pub const FUSE_TBL_SNAPSHOT_PREFIX: &str = "_ss";
 pub const DEFAULT_BLOCK_PER_SEGMENT: usize = 1000;
 pub const DEFAULT_ROW_PER_BLOCK: usize = 1000 * 1000;
 pub const DEFAULT_BLOCK_SIZE_IN_MEM_SIZE_THRESHOLD: usize = 100 * 1024 * 1024;
+
+lazy_static! {
+    pub static ref RESERVED_OPTS: HashSet<&'static str> = {
```
BohuTANG, line 35: Looks better move to sql crate? Let's keep sql crate not depends on the storage.

dantengsky: > Looks better move to sql crate? Let's keep sql crate not depends on the storage.

OK

dantengsky:
@BohuTANG 

Got a problem in the stateless test, `make stateless-test` gives positive, but `./scripts/ci/ci-run-tests-embedded-meta.sh` negative.

Seems there are some flaws in the arrangement of snapshot versioning, digging into it.

PR will be marked as ready after the CI passed.

dantengsky:
@BohuTANG PTAL:

besides  issue #4262 that this PR is trying to address:
 
- unify dev stateless test with the corresponding CI tests
  `make stateless-test` now uses `ci-run-tests-embedded.sh`
  which does not enable table cache, imo it is a more proper setting that may discover more potential flaws
  
- no longer use table option key SNAPSHOT_VER

  instead, inferring the versions of snapshots from the snapshot's location

  Only in the bootstrap phase, i.e. the version of snapshot that comes
  directly from TableInfo is deduced in this way.

  For the versions of history snapshots, the numeric values that kept
  in meta data file are used.

  Later, when zero-copy clone is implemented seriously, the above inconsistency
  should disappear.

- backward compatibility of this PR has also been verified manually in EC2 + S3

BohuTANG (approved):
lgtm

PR #4634, opened 2022-03-30
feat(io): refactor type deserialization

sundy-li:

## Summary

Summary about this PR

1. Refactor type deserialization
2. Add new settings:

```
empty_as_default  Format empty_as_default, default value: 1	UInt64
field_delimiter		Format field delimiter, default value: ,	String
record_delimiter Format record_delimiter, default value: \n	String
skip_header    Whether to skip the input header, default value: 0
```

## Changelog

- New Feature

## Related Issues

Related #4413

## Test Plan

Unit Tests

Stateless Tests


ygf11:
> This is a CheckpointReader for SourceValues. If any error happens in deserialization, it will use to_checkpoint_mode to reset the reader. I think it may be helpful to use it to parse expressions for this special row.

Great. With `CheckpointReader`, we can do further improvement:

1. Before parsing a row, we clear the buffer(`reader.checkpoint()`).
2. When `ValueSource` meets error, we move to checkpoint mod to parse expressions.
3. At last, blocks generated from 1 and 2 can merge.

ygf11:
> This is a CheckpointReader for SourceValues. If any error happens in deserialization, it will use to_checkpoint_mode to reset the reader. I think it may be helpful to use it to parse expressions for this special row.

So With `CheckpointReader`, we can do further improvement, like:

1. Before parsing a row, we clear the buffer(`reader.checkpoint()`).

3. When ValueSource meet error, we move to checkpoint mode to parse the error row as expression, and after that continue to parse remaining rows with the main mode.
4. At last, blocks and expressions can merge in interpreter.

sundy-li:
Yes, almost.

1. Default to use streaming parser, before parsing the values tuple, I will use `checkpoint ` to save the checkpoint.
2. If error happens, turn the reader to read util next ')'  to finish this tuple and use checkpoint mode to read and parse into the expression.  If we got expressions, we can have a single row block, this block can merged into the builder.

3. Continue to parse the rest tuple using streaming parser.

Learn it from: https://github.com/ClickHouse/ClickHouse/blob/8e9783388bbbf10a4fad6af47da627df5d909c6a/src/Processors/Formats/Impl/ValuesBlockInputFormat.cpp  

tryReadValue

sundy-li:
But I'm suffering from the  `ReaderBufer` now, it's hard to work with `CheckPointReader`.

query/src/sessions/session_settings.rs
```diff
@@ -137,6 +137,34 @@ impl Settings {
                 level: ScopeLevel::Session,
                 desc: "Enable new processor framework if value != 0, default value: 1",
             },
+
+            SettingValue {
```
BohuTANG, line 141: Why we add these parameters which only used by CSV  to the session settings?

sundy-li: 1. With this, users can specify their default format styles in settings. 
2. We can build `FormatSettings` from `QueryContext`, all serialization methods have `FormatSettings` args as [ClickHouse did](https://github.com/ClickHouse/ClickHouse/blob/ed424372199cc424fbe51672688796a58ec6d30c/src/DataTypes/Serializations/SimpleTextSerialization.h#L9-L60)

In the future, we can have a custom input/output format with settings.

For example:

```
echo "select * from ontime format CSV" | mysql -uroot --field_delimiter '\t' 

echo "insert into ontime format TabSeparated" |  mysql -uroot --field_delimiter ','     < ~/a.csv
```

Refer to [ClickHouse settings](https://github.com/ClickHouse/ClickHouse/blob/2e0d3efba0e22748507f204df444b72846849a8e/src/Core/Settings.h)

BohuTANG (approved):
lgtm

PR #4642, opened 2022-03-31
Feature: add range-map module

lichuang:

## Summary

Feature: add range-map module

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/range-map/src/range_wrapper.rs
```diff
@@ -0,0 +1,134 @@
...
+
+impl<T> PartialOrd for RangeWrapper<T>
+where T: Ord + std::fmt::Debug
+{
+    fn partial_cmp(&self, other: &RangeWrapper<T>) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+#[cfg(test)]
```
BohuTANG, line 90: Databend test style is `tests/it` ,  https://github.com/datafuselabs/databend/tree/main/common/streams/tests/it/

common/range-map/src/range_wrapper.rs
```diff
@@ -0,0 +1,134 @@
...
+                }
+                Ordering::Equal => {
+                    return Ordering::Greater;
+                }
+                Ordering::Greater => {
+                    return Ordering::Greater;
+                }
+            },
+        }
+    }
```
drmingdrmer, line 79: These lines are just a complex way of comparing `(self.range.end, self.range.start)` and `(other.end, other.start)`.

common/range-map/src/range_wrapper.rs
```diff
@@ -0,0 +1,134 @@
...
+impl<T> PartialEq for RangeWrapper<T>
+where T: Eq
+{
+    fn eq(&self, other: &RangeWrapper<T>) -> bool {
+        self.range.start == other.range.start && self.range.end == other.range.end
+    }
+}
+
+impl<T> Ord for RangeWrapper<T>
+where T: Ord + std::fmt::Debug
```
drmingdrmer, line 38: To manually implement a total order value, these properties have to be satisfied or a data structure such as btree or binary search won't work as expected:

https://en.wikipedia.org/wiki/Total_order

image

Thus tests that ensure these properties will be needed.

common/range-map/src/range_map.rs
```diff
@@ -0,0 +1,255 @@
...
+use super::range_wrapper::RangeWrapper;
+
+#[derive(Clone)]
+pub struct RangeMap<K, V> {
+    // Wrap ranges so that they are `Ord`.
+    // See `range_wrapper.rs` for explanation.
+    pub(crate) btm: BTreeMap<RangeWrapper<K>, V>,
+}
+
+impl<K, V> Default for RangeMap<K, V>
```
drmingdrmer, line 31: I do not quite get the design goal of this map:

- Does it allow to insert overlapping `range` with different values? If it does allow this to happen, what will a `get()` actually get?
- Is a query for a sub-range allowed? What if the range map contains two ranges `[1,3]` and `[3,5]` and a query for `[2,4]`? Will it return a positive result?

common/range-map/src/range_map.rs
```diff
@@ -0,0 +1,151 @@
...
+    V: Eq + Clone + std::fmt::Debug,
+{
+    /// Makes a new empty `RangeMap`.
+    pub fn new() -> Self {
+        RangeMap {
+            btm: BTreeMap::new(),
+        }
+    }
+
+    pub fn contain_key(&self, key: &K) -> Vec<(&RangeWrapper<K>, &V)> {
```
drmingdrmer, line 83: Normally `contain_key` returns a bool. If it returns some stored value, it should be `get_xxx()`.

common/range-map/src/range_map.rs
```diff
@@ -0,0 +1,151 @@
...
+        match old_value {
+            None => {
+                self.btm.insert(new_range_start_wrapper, new_value);
+            }
+            Some(old_value) => {
+                let new_value = f(old_value, &new_value);
+                self.btm.insert(new_range_start_wrapper, new_value);
+            }
+        };
+    }
```
drmingdrmer, line 128: To simplify the API, `F` could be a function `Fn(&Option<V>) -> V`.

This way the argument `value: V` won't be necessary. It will be the duty of `F` to produce a value if the argument is `None`.

common/range-map/src/range_wrapper.rs
```diff
@@ -0,0 +1,78 @@
...
+{
+    fn cmp(&self, other: &RangeWrapper<T>) -> Ordering {
+        // first handle the case one of the range has only one point(start == end)
+        if other.range.start == other.range.end {
+            return self.compare_to_point(other);
+        } else if self.range.start == self.range.end {
+            return Ordering::reverse(other.compare_to_point(self));
+        }
+        // use (start,end) tuple cmp result
+        (self.range.start, self.range.end).cmp(&(other.range.start, other.range.end))
```
drmingdrmer, line 68: Why do you need to compare a range with a point?
Isn't line 68 enough for your need?

drmingdrmer:
Another issue is that for a map, equal keys will override each other.
But in our case, two watchers have chance to watch on the same key(key range)s, and both of them should be stored in the map.

common/range-map/src/range_map.rs
lichuang: if pass to Fn the `Option`, then no matter or not there is existing key, the new value is decide by the function, by the design of this `upsert` function is using the fn to generate new value only when there is existing value.If there is no value exise, it will insert directly.

common/range-map/src/range_wrapper.rs
lichuang: emm, without consider the cases, if we have range [1,5],[2,4], and `get_by_key(2)`,  will not be return the range [1,5], because (2, 2).cmp(&(1, 5) return Greater.

common/range-map/src/range_map.rs
drmingdrmer: Right, in case the user just wants to override the existing value, the user does not need to write the `value` twice.

common/range-map/src/range_wrapper.rs
drmingdrmer: There should be two `cmp()` methods defined: 
- One of them is for comparing two `range`s. It is used by a BTreeMap to organize keys in it.
- The other should only be used for `get_by_key()`: it only compares a `range` with a point value. Because `range<K>` actually can not be made a total order value that is compatible with the `Ord` defined for `K`. `range` is more like a two-dimension value.

lichuang:
since merge from upstream/main make some mistakes, transfer to this pr:https://github.com/datafuselabs/databend/pull/4669

BohuTANG:
@mergify update

PR #4669, opened 2022-04-02
Feature: add range-map module

lichuang:

## Summary

Feature: add range-map module

## Changelog

- New Feature

## Related Issues

Fixes #issue

## Test Plan

Unit Tests

Stateless Tests


common/range-map/src/range_key.rs
```diff
@@ -0,0 +1,76 @@
...
+where
+    T: Eq,
+    K: Eq,
+{
+    fn eq(&self, other: &RangeKey<T, K>) -> bool {
+        self.range.start == other.range.start
+            && self.range.end == other.range.end
+            && self.key == other.key
+    }
+}
```
drmingdrmer, line 55: `Eq` can be derived. 🤔

common/range-map/src/range_key.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl<T, K> Ord for RangeKey<T, K>
+where
+    T: Ord + std::fmt::Debug + Copy,
+    K: Ord + std::fmt::Debug + Copy,
+{
+    fn cmp(&self, other: &RangeKey<T, K>) -> Ordering {
+        ((self.range.end, self.range.start), self.key)
+            .cmp(&((other.range.end, other.range.start), other.key))
+    }
```
drmingdrmer, line 65: This method deserves a doc comment explaining why the order is defined as such.

The definition of this `Ord` is critical to the correctness of the range-set.

common/range-map/src/range_set.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl<T, K> Default for RangeSet<T, K>
+where
+    T: Ord + Clone + std::fmt::Debug + Copy,
+    K: Ord + Clone + std::fmt::Debug + Copy + Default,
+{
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
drmingdrmer, line 37: Such a default can be derived. Save you several lines of code :DDD

common/range-map/src/range_set.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl<T: Debug, K: Debug> Debug for RangeSet<T, K>
+where
+    T: Ord + Clone + std::fmt::Debug + Copy,
+    K: Ord + Clone + std::fmt::Debug + Copy + Default,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_set().entries(self.iter()).finish()
+    }
+}
```
drmingdrmer, line 67: Isn't the derived implementation of `Debug` enough?

common/range-map/src/range_set.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    pub fn remove(&mut self, range: Range<T>, k: K) {
+        self.set.remove(&RangeKey::new(range, k));
+    }
+
+    pub fn iter(&self) -> Iter<'_, T, K> {
+        Iter {
+            inner: self.set.iter(),
+        }
+    }
```
drmingdrmer, line 114: Is this enough for you need?
```suggestion
    pub fn iter(&self) -> Iter<'_, T, K> {
            self.set.iter()
    }
```

common/range-map/src/range_set.rs
```diff
@@ -0,0 +1,115 @@
...
+    // 3. `get_by_key_range(5)` return [2,4],[2,6]
+    // Use the default key when construct `RangeKey::key` for search.
+    pub fn get_by_key_range(&self, key: &T) -> Vec<&RangeKey<T, K>> {
+        let key_as_start = RangeKey::new(*key..*key, K::default());
+
+        self.set
+            .range((Bound::Included(key_as_start), Bound::Unbounded))
+            .filter(|e| {
+                e.range.start.cmp(key) != Ordering::Greater
+                    && e.range.end.cmp(key) != Ordering::Less
```
drmingdrmer, line 101: I think `e.range.end >= key` will always hold. right?

Cargo.toml
```diff
@@ -23,6 +23,7 @@ members = [
     "common/meta/raft-store",
     "common/meta/sled-store",
     "common/meta/types",
+    "common/range-map",
```
drmingdrmer, line 26: It is renamed to `range-set` right?

Xuanwo:
This module seems not to depend on any crates in databend, how about moving it out as a separate project?

common/range-set/src/range_key.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl<T, K> PartialEq for RangeKey<T, K>
+where
+    T: Eq,
+    K: Eq,
+{
+    fn eq(&self, other: &RangeKey<T, K>) -> bool {
+        self.range.eq(&other.range) && self.key.eq(&other.key)
+    }
+}
```
drmingdrmer, line 53: Such a default trait impl can be derived too.

common/range-set/src/range_set.rs
```diff
@@ -0,0 +1,65 @@
...
+    // 1. `get_by_key_range(1)` return [1,5]
+    // 2. `get_by_key_range(2)` return [1,5],[2,4],[2,6]
+    // 3. `get_by_key_range(5)` return [2,4],[2,6]
+    // Use the default key when construct `RangeKey::key` for search.
+    pub fn get_by_key_range(&self, key: &T) -> Vec<&RangeKey<T, K>> {
+        let key_as_start = RangeKey::new(*key..*key, K::default());
+
+        self.set
+            .range((Bound::Included(key_as_start), Bound::Unbounded))
+            .filter(|e| e.range.start.cmp(key) != Ordering::Greater)
```
drmingdrmer, line 58:
```suggestion
            .filter(|e| e.range.start <= key)
```

common/range-set/tests/it/range_key_test.rs
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use core::cmp::Ordering;
+use std::collections::BTreeMap;
+use std::collections::BTreeSet;
+
+use common_range_set::RangeKey;
+
+#[cfg(test)]
```
drmingdrmer, line 21: Is this required?

drmingdrmer (commented):
The term `key` is used abusively in this crate:
It can be `(end, start, some_thing)` AKA, `RangeKey`, or the last value in `RangeKey`.

`RangeKey` in this crate is `a range and a key`, but the name sounds like `range as a key` or `the key of some range`. Especially when there is a `RangeSet`, which means `a set of ranges`.

BTW `get_by_key_range()` is actually a method to get by `key`.

lichuang:
> The term key is used abusively in this crate:
It can be (end, start, some_thing) AKA, RangeKey, or the last value in RangeKey.

RangeKey in this crate is a range and a key, but the name sounds like range as a key or the key of some range. Especially when there is a RangeSet, which means a set of ranges.

BTW get_by_key_range() is actually a method to get by key.



rename `get_by_key_range` to `get_by_point` and add more comments on `RangeKey`

lichuang:
> This module seems not to depend on any crates in databend, how about moving it out as a separate project?

good suggest! But now this mod is too small to be a a separate project, maybe i will do it sometime later to add more function in this module.

drmingdrmer (approved):
LGTM~

PR #4677, opened 2022-04-02
ISSUE-4761: Hide table opt SNAPSHOT_LOC in `show create`

### LINKED ISSUE #4761: bug: desc view not work
### Search before asking

### Version

main

### What's Wrong?

`desc INFORMATION_SCHEMA.COLUMN` returns empty results.

### What You Expected?

Works like mysql.

### How to Reproduce?

```
MySQL [(none)]> desc INFORMATION_SCHEMA.COLUMNS;
Empty set (0.023 sec)
```

### Anything Else?

_No response_

dantengsky:

## Summary

- table option SNAPSHOT_LOC will no longer be shown in `show create table`
- table option SNAPSHOT_LOC can still be specified in `create table`
- table function `fuse_history` and `system$fuse_history` now contains a new column `location` which indicates the location of the table snapshot

## Changelog

- Improvement

## Related Issues

Fixes #4761

## Test Plan

Unit Tests

Stateless Tests


tests/suites/0_stateless/09_fuse_engine/09_0011_create_table_with_opt_loc.sh
```diff
@@ -11,12 +11,14 @@ echo "show value of table being cloned"
...
-SNAPSHOT_LOC=$(echo "show create table default.t09_0011" | mysql -h127.0.0.1 -P3307 -uroot -s | sed 's/.*SNAPSHOT_LOC=\(.*\).*/\1/g')
+SNAPSHOT_LOC=$(echo "select location from fuse_history('default','t09_0011')" | mysql -h127.0.0.1 -P3307 -uroot -s)
+
 ## create a shallow clones of t09_0011 by using the table option 'snapshot_loc'
 ## using lower case option key
-echo "create table t09_0011_clone1(c int) snapshot_loc=$SNAPSHOT_LOC" | $MYSQL_CLIENT_CONNECT
+echo "create table t09_0011_clone1(c int) snapshot_loc='$SNAPSHOT_LOC'" | $MYSQL_CLIENT_CONNECT
 ## using upper case option key
-echo "create table t09_0011_clone2(c int) SNAPSHOT_LOC=$SNAPSHOT_LOC" | $MYSQL_CLIENT_CONNECT
+echo "create table t09_0011_clone2(c int) SNAPSHOT_LOC='$SNAPSHOT_LOC'" | $MYSQL_CLIENT_CONNECT
```
BohuTANG, line 20: Can we use the full name:
SNAPSHOT_LOC -> SNAPSHOT_LOCATION?

Let's make user not confused at 'LOC'.

dantengsky: OK

query/src/sql/table_option_keys.rs
```diff
@@ -17,15 +17,31 @@ use std::collections::HashSet;
 use lazy_static::lazy_static;
 
 pub const OPT_KEY_DATABASE_ID: &str = "database_id";
+pub const OPT_KEY_SNAPSHOT_LOCATION: &str = "snapshot_loc";
```
BohuTANG, line 20: "snapshot_loc" -> "snapshot_location" is better?

query/src/storages/fuse/fuse_history.rs
```diff
@@ -82,23 +89,25 @@ impl<'a> FuseHistory<'a> {
             current_snapshot_version = ver;
         }
 
-        DataBlock::create(FuseHistory::schema(), vec![
+        Ok(DataBlock::create(FuseHistory::schema(), vec![
             Series::from_data(snapshot_ids),
-            Series::from_data(prev_snapshot_ids),
+            Series::from_data(snapshot_locs),
```
BohuTANG, line 94: I think the column  name `snapshot_locs` > `snapshot_locations` is more clearly for the user.

dantengsky: OK

query/src/sql/table_option_keys.rs
dantengsky: If we should keep the backward compatibility, I am afraid the `"snapshot_loc"` have to be kept.

Another option key is being introduced `OPT_KEY_SNAPSHOT_LOCATION = "snapshot_location"`, which will replace the `"snapshot_loc"`. 

engine will try to read `snapshot_location` first, if failed, try to read `snapshot_loc`, thus we can read the existing tables (which have `snapshot_loc` as the location key).

All the new tables will use `"snapshot_location"`

BohuTANG (approved):
Nice work!

PR #4684, opened 2022-04-04
feat(function): Support Semi-structured function GET/GET_IGNORE_CASE/GET_PATH

b41sh:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Support `GET` function
Support `GET_IGNORE_CASE` function
Support `GET_PATH` function

## Changelog

- New Feature
- Build/Testing/CI
- Documentation

## Related Issues

Fixes #4665

## Test Plan

Unit Tests

Stateless Tests

PR #4733, opened 2022-04-07
fix(function): Const(Nullable(Object)) column downcast to Nullable failed

b41sh:

## Summary

If the column is `Const(Nullable(Object))`, it will panic when downcast to `Nullable(Object)` column.

There is one more issue to be concerned about. The column data type returned by the storage is a little different. If the engine is FUSE, it will return the `Nullable` column, but if the engine is Memory, it will return the `Const Nullable` column.

## Changelog

- Bug Fix

## Related Issues

[#4684](https://github.com/datafuselabs/databend/pull/4684)


common/functions/src/scalars/function_adapter.rs
```diff
@@ -180,7 +180,12 @@ impl Function for FunctionAdapter {
                 });
 
                 let col = if col.is_nullable() {
-                    let nullable_column: &NullableColumn = unsafe { Series::static_cast(&col) };
+                    let nullable_column: &NullableColumn = if col.is_const() {
+                        let const_column: &ConstColumn = Series::check_get(&col)?;
+                        Series::check_get(const_column.inner())?
```
sundy-li, line 185: Size may be different.

sundy-li (approved):

Xuanwo (approved):

Issue #4741, opened 2022-04-07
`unwrap` in `dal_contexts` is wired


Xuanwo:
The code style is wired, using make lint to do the format?

_Originally posted by @BohuTANG in https://github.com/datafuselabs/databend/pull/4737#discussion_r845028897_

PR #4750, opened 2022-04-08
refactor(contexts): Improve logic for getting inner Accessor

Xuanwo:

## Summary

Fix https://github.com/datafuselabs/databend/issues/4741

## Changelog

- Improvement

### DIFF
--- a/common/contexts/src/dal/dal_context.rs
+++ b/common/contexts/src/dal/dal_context.rs
@@ -49,6 +49,13 @@ impl DalContext {
         }
     }
 
+    fn get_inner(&self) -> Arc<dyn Accessor> {
+        match &self.inner {
+            None => panic!("dal context must init wrongly, inner accessor is empty"),
+            Some(inner) => inner.clone(),
+        }
+    }
+
     pub fn get_metrics(&self) -> Arc<DalMetrics> {
         self.metrics.clone()
     }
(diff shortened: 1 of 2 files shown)


Xuanwo:
@mergify update

sundy-li (approved):

common/contexts/src/dal/dal_context.rs
```diff
@@ -66,13 +73,13 @@ impl Layer for DalContext {
 #[async_trait]
 impl Accessor for DalContext {
     async fn create(&self, args: &OpCreate) -> Result<()> {
-        self.inner.as_ref().unwrap().create(args).await
+        self.get_inner().create(args).await
```
dantengsky, line 76: Since `async fn create(&self, args: &OpCreate)` returns `Result<()>`,  will it be better to make `get_inner`  return `Result<Arc<dyn Accessor>>`? such that the `panic!` can be avoided.

Panics can be caught and handled by the session gracefully though (hope I get it right)... 

@zhang2014  is `painc`  suggested being used here? (`get_inner` may panic)

Xuanwo: I use `panic` here because it never should happen unless we use the `DalContext` wrongly which should be fixed ASAP.

Xuanwo: > Panics can be caught and handled by the session gracefully though (hope I get it right)...

No, I think it never happens.

The only way to initiate the `DalContext` is `DalContext::new()` or `DalContext::layer()` . Maybe I should also remove the `Default` too?

Xuanwo: `QueryContextShared::try_create()` will init a `dal_context` with `Default::default()`.

@dantengsky I got your concerns. Let's return an error instead.

dantengsky: > Maybe I should also remove the `Default` too?

:+1:  I tried a little bit, but failed to figure out how to implement the `QueryContextShared::try_create`

dantengsky (approved):
LGTM

common/contexts/src/dal/dal_context.rs
Xuanwo: Yep, I met the same question so I realized that it's better to return errors instead.

PR #4751, opened 2022-04-08
Fix schema display on http api

### LINKED ISSUE #4753: bug: schema column empty on "SHOW DATABASES" statement in the HTTP query api
### Search before asking

### Version

latest

### What's Wrong?

origin_img_v2_494a087a-ad5c-4634-b2b0-9532e748991g

but it's ok to SELECT system.tables

### What You Expected?

it should display the column fields in the SHOW DATABASES statement.

### How to Reproduce?

_No response_

### Anything Else?

_No response_

flaneur2020:

## Summary

currently the schema is taken from plan.schema:

Screen Shot 2022-04-08 at 9 53 42 AM

but in the "SHOW xxx" statements, plan is tranformed into the SELECT query, so the plan struct can not really get the actual schema of the statement.

it might be more accurate to get the schema from the DataBlock, as it always binds the schema with the data inside it.

changes:

- wraps JsonBlock with a schema reference, which inherits from the original DataBlock.
- make JsonBlock owns by Page instead of Arc
- rename ExecuteStateName to ExecuteStateKind
- rename block_to_json.rs to json_block.rs

## Changelog

- Bug Fix

## Related Issues

Fixes #4753


query/tests/it/servers/http/http_query_handlers.rs
```diff
@@ -128,7 +136,7 @@ async fn test_async() -> Result<()> {
         assert!(result.next_uri.is_none());
         assert!(result.schema.is_none());
```
youngsofun, line 137: > assert!(result.schema.is_none()); 

change is_none to is_some()

youngsofun: before this change, only the first response contains schema.

now every response may contain or not contain schema,  only depends on if data is empty.

youngsofun:
only one question:

it may take some time to wait for the first block to be generated.
may be not what user (like UI @Chasen-Zhang ) expected?

query/tests/it/servers/http/http_query_handlers.rs
flaneur2020: fixed, thank you for remind

flaneur2020:
> only one question:
> 
> it may take some time to wait for the first block to be generated. may be not what user (like UI @Chasen-Zhang ) expected?

i guess it's ok to get the data & schema together but later, a schema without data maybe not very useful for the UI, in the most time, the UI would display a progressing view until the first block arrives, then it displays the table with data & schema together 🤔

PR #4760, opened 2022-04-08
query(mysql): information_schema case insenstive

sundy-li:

## Summary

Summary about this PR

## Changelog

- Improvement 
 

## Related Issues

Fixes #issue


BohuTANG:
How about to run this query?
```
select * from information_schema.TABLES ... 
```

query/src/sessions/query_ctx.rs
```diff
@@ -221,6 +221,12 @@ impl QueryContext {
     /// SELECT * FROM (SELECT * FROM db.table_name) as subquery_1, (SELECT * FROM db.table_name) AS subquery_2
     /// ```
     pub async fn get_table(&self, database: &str, table: &str) -> Result<Arc<dyn Table>> {
+        if database.eq_ignore_ascii_case("INFORMATION_SCHEMA") {
```
BohuTANG, line 224: With this ,so we don;t need `impl<const UPPER: bool> ColumnsTable<UPPER>`

sundy-li: You are right!

BohuTANG: It would be better to add a map `case_insensitive_databases` to mark these database's table is case insensitive?

sundy-li: How about keep these then we can have upper & lower case db/tables in `select * from system.tables` ?

BohuTANG: I think `INFORMATION_SCHEMA ` main for MySQL driver/BI usage, `select * from system.tables` no related to it?

BohuTANG: We can keep only one like `INFORMATION_SCHEMA` and its tables with upper-case too?

BohuTANG:
Oops, conflicting files:
query/src/interpreters/interpreter_show_tables.rs

BohuTANG (approved):
💯

PR #4764, opened 2022-04-08
feat: row granularity stream values insert

ygf11:

## Summary

Summary about this PR

## Changelog

- Improvement
- Performance Improvement


ygf11:
Performance improves a lot in some workload, like some rows contain expression-columns.

I do bench in my local pc.

Data: 100 rows which contains expressions when insert 100000 record. [bench-script](https://www.logpasta.com/paste/5314d32d-28ba-486a-ba2e-bee45d123d2c)

1. In main branch, the cost is 1.622s.
2. In current, the cost is 379ms.

ygf11:
But When every row contains  expression-columns, the performance will drop.

[bench-script](https://www.logpasta.com/paste/494e7f6d-e712-4d39-bf81-b6bccef3b19c)

1. in main branch, the cost is 2.283s.
2. in current, the cost is 5.172s.

It seems the bottleneck is creating too many `BackTrace` when deserialize fails.

query/src/sql/statements/value_source.rs
```diff
@@ -99,29 +89,128 @@ impl ValueSource {
...
+                        .await?;
+                deser.append_data_value(values[col].clone())?;
+                datavalues = Some(values);
+            }
+        }
+
+        Ok(())
+    }
+
+    fn try_deserialize_next_column(
```
sundy-li, line 141: Let's just use `deserializer.de_text_quoted(reader).`

If error happens we pop from `[0 - current_idx]`

query/src/sql/statements/value_source.rs
```diff
@@ -99,29 +89,128 @@ impl ValueSource {
...
+        &self,
+        reader: &mut CpBufferReader,
+        deserializer: &mut Box<dyn TypeDeserializer>,
+        col_end: u8,
+    ) -> Result<()> {
+        deserializer.de_text_quoted(reader).and_then(|_| {
```
sundy-li, line 147: if `deserializer.de_text_quoted(reader)` has error, maybe we should pop the whole row.

query/src/sql/statements/value_source.rs
ygf11: Addressed.

query/src/sql/statements/value_source.rs
ygf11: Addressed.

common/datavalues/src/types/deserializations/date_time.rs
```diff
@@ -69,9 +69,17 @@ where
     }
 
     fn de_text_quoted(&mut self, reader: &mut CpBufferReader) -> Result<()> {
-        reader.must_ignore_byte(b'\'')?;
+        reader
+            .must_ignore_byte(b'\'')
+            .map_err_to_code(ErrorCode::NoneBtBadBytes, || {
+                "Invalid start of datetime string"
+            })?;
         let datetime = reader.read_datetime_text(&self.tz)?;
```
sundy-li, line 77: This has an exception backtrace...

sundy-li:
@ygf11  I add `BacktraceGuard` in your commits, you can keep this guard as a field in `ValueSource`.

ygf11:
It seems that this guard does not work for multiple users, such as: 
1. user1 enable this guard. 
2. user2 disable this guard before user1 finish.

common/datavalues/src/types/deserializations/date_time.rs
ygf11: I also use `NoneBtBadBytes` in `read_datetime_text` and `read_date_text`.

sundy-li:
> It seems that this guard does not work for multiple users, such as:
> 
> 1. user1 enable this guard.
> 2. user2 disable this guard before user1 finish.

Yes， but I think `backtrace` is really trivial, it can be None if this case happens.

ygf11:
> Yes， but I think backtrace is really trivial, it can be None if this case happens.

How about using as a thread local here?

sundy-li:
> > Yes， but I think backtrace is really trivial, it can be None if this case happens.
> 
> How about using as a thread local here?

Yes， it can be better to make this a thread_local variable.

ygf11:
> Yes， it can be better to make this a thread_local variable.

Let me try it.

sundy-li (approved):

PR #4779, opened 2022-04-09
Feature: Watch api

lichuang:

## Summary

1. add watch API for metasrv
2. add unit tests for watch API.

## Changelog

- New Feature

## Related Issues

Fixes #issue


common/meta/types/proto/meta.proto
```diff
@@ -76,4 +135,8 @@ service MetaService {
   // Including raft hard state, logs and state machine.
   // The exported data is a list of json strings in form of `(tree_name, sub_tree_prefix, key, value)`.
   rpc Export(Empty) returns (stream ExportedChunk);
+
+  // Add watch key stream.
+  // Whenever the watch key data updated, client will be notified accross the stream.
+  rpc Watch(stream WatchRequest) returns (stream WatchResponse);
```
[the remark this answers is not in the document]
lichuang: emm, etcd watch API has binary c/s stream implementation, it can save some resource if the client create/cancel/ and then create another watch request again...

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,404 @@
...
+            self.watcher_streams
+                .insert(self.current_stream_id, watcher_stream);
+
+            self.current_stream_id += 1;
+
+            self.create_watcher(self.current_stream_id, create);
+        }
+    }
+
+    async fn watch_request(&mut self, stream_id: WatcherStreamId, req: WatchRequest) {
```
ariesdevil, line 305: We should report `create` and `cancel` metrics here.

common/meta/types/proto/meta.proto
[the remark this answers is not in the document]
drmingdrmer: hmm... I doubt that the resource reduction is measurable in this case...
But your code your decision:DD

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -574,7 +584,12 @@ impl StateMachine {
         tracing::debug!("applied UpsertKV: {} {:?}", key, result);
 
         if let Some(subscriber) = &self.subscriber {
-            let _ = subscriber.kv_changed(&key_str, &prev, &result);
+            let update = match value_op {
+                Operation::Update(_v) => UpdateType::UpInsert,
+                Operation::Delete => UpdateType::Delete,
+                Operation::AsIs => UpdateType::UpInsert,
+            };
+            let _ = subscriber.kv_changed(&key_str, update, &prev, &result);
```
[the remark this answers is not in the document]
drmingdrmer: I unresolved this comment, because if using `Some(x), None` to represent a `delete` operation, the `UpdateType` can be removed, but it is still there.

I do not quite get it. Are you gonna remove `UpdateType` or still use it?

common/meta/types/proto/meta.proto
```diff
@@ -48,6 +48,65 @@ message ExportedChunk {
...
+  string key_end = 2;
+}
+
+message WatchCancelRequest {
+  int64 watcher_id = 1;
+}
+
+message Event {
+  enum EventType {
+    UpInsert = 0;
```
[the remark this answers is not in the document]
drmingdrmer: Still, it is not changed but it has been marked as **resolved**.
I am confused:(

common/meta/types/proto/meta.proto
```diff
@@ -48,6 +48,65 @@ message ExportedChunk {
...
+
+  string key = 1;
+
+  EventType event = 2;
+
+  // current hold current value of key
+  bytes current = 3;
+
+  // current hold prev value of key(if any)
+  bytes prev = 4;
```
drmingdrmer, line 89: It seems it is possible to make it an `Option`:
https://github.com/tokio-rs/prost#field-modifiers

common/meta/types/proto/meta.proto
```diff
@@ -48,6 +48,65 @@ message ExportedChunk {
...
+  // The client should record the watch_id and expect to receive events for
+  // the created watcher from the same stream.
+  // All events sent to the created watcher will attach with the same watch_id.
+  bool created = 2;
+
+  // canceled is set to true if the response is for a cancel watch request.
+  // No further events will be sent to the canceled watcher.
+  bool canceled = 4;
+
+  repeated Event events = 3;
```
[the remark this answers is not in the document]
lichuang: Watcher may be notified many event at the same time, for example, if watch ["a", "z"], then event ["a"] and ["z"] come in a same internal, these two events can be merge into one response.

metasrv/src/watcher/watcher_config.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub watcher_notify_internal: u64,
+}
+
+impl Default for WatcherConfig {
+    fn default() -> Self {
+        Self {
+            watcher_notify_internal: 100,
+        }
+    }
+}
```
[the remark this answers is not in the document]
lichuang: Watcher may be notified many event at the same time, for example, if watch ["a", "z"], then event ["a"] and ["z"] come in a same internal, these two events can be merge into one response.

watcher_notify_internal is the batch response time.

metasrv/src/watcher/watcher_manager.rs
lichuang: maybe metrics can be added next pr?

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -584,12 +571,14 @@ impl StateMachine {
...
 
         if let Some(subscriber) = &self.subscriber {
-            let update = match value_op {
-                Operation::Update(_v) => UpdateType::UpInsert,
-                Operation::Delete => UpdateType::Delete,
-                Operation::AsIs => UpdateType::UpInsert,
+            match value_op {
+                Operation::Delete => {
+                    let _ = subscriber.kv_changed(&key_str, prev.clone(), None);
+                }
```
drmingdrmer, line 577: Won't the result be `None` if it is a successful delete?

common/meta/types/proto/meta.proto
```diff
@@ -49,44 +49,39 @@ message ExportedChunk {
...
+  optional string key_end = 2;
+
+  enum FilterType {
+    // recv all kind update event.
+    ALL = 0;    
+    // filter out update event.
+    NOUPDATE = 1;
+    // filter out delete event.
+    NODELETE = 2;
+  }
```
drmingdrmer, line 66: Using positive expression would be better: `UPDATA` is better than `NODELETE`.

metasrv/src/watcher/watcher_config.rs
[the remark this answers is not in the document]
drmingdrmer: But why does it need batching? The client should receive the event ASAP IMHO.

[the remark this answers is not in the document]
lichuang: well, it's a tradeoff of efficiency and performance, user can config the internal using the `watcher_notify_internal ` config.

if notify ASAP, the server may be too busy in notifying events; if internal too big, may be events came too late.so using `watcher_notify_internal ` to cache the events and batch notify all the events in this internal, and user can config the internal.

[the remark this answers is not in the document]
drmingdrmer: The batching can be left to be done in the network layer IMHO.
We do not have to worry too much about it.

If application-level batching is really necessary, trying the best to drain a message channel by a limit of the number of messages would be better than by a timer:

https://github.com/datafuselabs/openraft/blob/b50679dcc6eba4e44a7e2106a1919f350676eff4/openraft/src/replication/mod.rs#L483-L509

common/meta/types/proto/meta.proto
```diff
@@ -48,6 +48,56 @@ message ExportedChunk {
...
+  string key = 1;
+
+  EventType event = 2;
+
+  // current hold current value of key(if any)
+  optional bytes current = 3;
+
+  // prev value of key
+  optional bytes prev = 4;
+}
```
drmingdrmer, line 85: We do not need `EventType` anymore.
`(prev.is_some(), current.is_some())` is enough to describe the event type.

common/meta/types/proto/meta.proto
```diff
@@ -48,6 +48,56 @@ message ExportedChunk {
...
+message WatchResponse {
+  //ResponseHeader header = 1;
+  // watch_id is the ID of the watcher that corresponds to the response.
+  int64 watch_id = 1;
+
+  // created is set to true if the response is for a create watch request.
+  // The client should record the watch_id and expect to receive events for
+  // the created watcher from the same stream.
+  // All events sent to the created watcher will attach with the same watch_id.
+  optional bool created = 2;
```
drmingdrmer, line 96: `created` is not necessary anymore if there is no `cancel` response.

common/meta/types/proto/meta.proto
```diff
@@ -48,6 +48,56 @@ message ExportedChunk {
...
+  optional bytes current = 3;
+
+  // prev value of key
+  optional bytes prev = 4;
+}
+
+message WatchResponse {
+  //ResponseHeader header = 1;
+  // watch_id is the ID of the watcher that corresponds to the response.
+  int64 watch_id = 1;
```
drmingdrmer, line 90: Is this field still in use?

metasrv/src/meta_service/raftmeta.rs
```diff
@@ -809,4 +821,8 @@ impl MetaNode {
         let res: Result<ForwardResponse, MetaError> = raft_mes.into();
         res
     }
+
+    pub fn create_watcher_stream(&self, request: Request<WatchRequest>, tx: WatcherStreamSender) {
+        self.watcher.create_watcher_stream(request, tx)
+    }
```
drmingdrmer, line 827: Except for the service endpoint layer, these method does not need to know about tonic request.

A `WatchRequest` is quite enough here.

```suggestion
    pub fn create_watcher_stream(&self, request: WatchRequest, tx: WatcherStreamSender) {
        self.watcher.create_watcher_stream(request, tx)
    }
```

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,276 @@
...
+
+type CreateWatcherEvent = (Request<WatchRequest>, WatcherStreamSender);
+
+#[derive(Clone, Debug)]
+struct StateMachineKvData {
+    pub key: String,
+    pub event: EventType,
+    pub prev: Option<SeqV>,
+    pub current: Option<SeqV>,
+}
```
drmingdrmer, line 49: Isn't it a duplicate of `Event` defined in `meta.proto`?

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    sm_rx: mpsc::UnboundedReceiver<StateMachineKvData>,
+
+    /// A channel for sending stop stream request.
+    close_stream_tx: Arc<mpsc::UnboundedSender<CloseWatcherStreamReq>>,
+
+    /// A channel for receiving stop stream request.
+    close_stream_rx: mpsc::UnboundedReceiver<CloseWatcherStreamReq>,
+
+    shutdown_rx: mpsc::UnboundedReceiver<()>,
```
drmingdrmer, line 79: What about merging all `*_rx` into one mpsc channel?

Such an `rx` can be used to receive every kind of event.
Of cause it requires defining the event with an `enum`.
Managing several enum variants may be easier than managing several channels.

And when this `rx` is shutting down, `WatchManager` will be shut down.

It is similar to `close_stream_tx`: a `closing` event can be replaced with shutting down a channel in most cases.

common/meta/types/proto/meta.proto
lichuang: > Is this field still in use?

i think client still need a id to identify watcher.

metasrv/src/watcher/watcher_manager.rs
drmingdrmer: Why resolve this comment? 🤔

metasrv/src/watcher/watcher_manager.rs
drmingdrmer: I do not see why you always click `resolve` before making any update...

common/meta/types/proto/meta.proto
drmingdrmer: > i think client still need a id to identify watcher.

And what does a client need a watch id for?...

metasrv/src/watcher/watcher_stream.rs
```diff
@@ -0,0 +1,72 @@
...
+                "close watcher stream {:?} cause send err: {:?}",
+                self.id,
+                err
+            );
+            let close_req: CloseWatcherStream = (self.id, err.to_string());
+            let _ = self
+                .close_stream_tx
+                .send(WatcherEvent::CloseWatcherStreamEvent(close_req));
+        }
+    }
```
drmingdrmer, line 71: A simpler way for closing a stream is to return the error to its caller.

Then the caller could just drop a `WatchStream` to close it.

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -209,10 +205,8 @@ impl WatcherManagerCore {
             filter,
         });
 
-        let _ = watcher_stream.send(WatchResponse {
-            watch_id: watcher_id,
-            event: None,
-        });
+        // send None event to notify stream has been created.
+        let _ = watcher_stream.send(WatchResponse { event: None });
```
drmingdrmer, line 215: Since create/cancel is removed, the client may not need to be notified anymore?

lichuang: maybe notified at the first time the watcher been created?

drmingdrmer: I don't know.

Usually, when I create a watcher, I'll expect no message to return until something I'm interested in happens.

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,242 @@
...
+                continue;
+            }
+
+            if let Some(stream) = self.watcher_streams.get(&watcher_id) {
+                assert_eq!(stream.id, watcher_id);
+                let resp = WatchResponse {
+                    event: Some(Event {
+                        key: kv.key.clone(),
+                        current: current.clone(),
+                        prev: prev.clone(),
```
drmingdrmer, line 169: If the `seq` is discarded, when the client receives an event, it may need to retrieve the `seq` with another RPC. Since several meta operations rely on `seq` to guarantee consistency.

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,242 @@
...
+        WatcherManager {
+            event_tx: event_tx.clone(),
+            subscriber: WatcherStateMachineSubscriber { event_tx },
+        }
+    }
+
+    pub fn stop(&self) {
+        let _ = self.event_tx.send(WatcherEvent::ShutdownEvent(()));
+    }
+
```
drmingdrmer, line 98: Just dropping `self.event_tx` is quite enough to notify the core to shut down.

Then at the receiving end:
```rust
            tokio::select! {
                event = self.event_rx.recv() => {
                    if let Some(event) = event {...}
                    else { /* shutdown */ }
```

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,242 @@
...
+    #[tracing::instrument(level = "debug", skip(self))]
+    fn close_stream(&mut self, id: WatcherId) {
+        let watcher = self.watcher_streams.get(&id);
+        if let Some(watcher) = watcher {
+            let range = watcher.key.clone()..watcher.key_end.clone();
+            self.watcher_range_set.remove(range, WatcherKey {
+                id,
+                filter: FilterType::All,
+            });
```
drmingdrmer, line 134: It looks like the filter should be part of the `WatcherStream`, but not part of the `WatchKey`.

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                },
+            }
+        }
+    }
+
+    #[tracing::instrument(level = "debug", skip(self))]
+    fn close_stream(&mut self, id: WatcherId) {
+        let watcher = self.watcher_streams.get(&id);
+        if let Some(watcher) = watcher {
```
drmingdrmer, line 129: This may be more suitable in this case:
`let watcher = self.watcher_streams.remove(&id)`.
The removed item will just be returned.

metasrv/src/watcher/watcher_manager.rs
```diff
@@ -0,0 +1,250 @@
...
+                            WatcherEvent::StateMachineKvDataEvent(kv) => {self.notify_event(kv).await;},
+                            WatcherEvent::ShutdownEvent(_) => {
+                                tracing::info!("watcher manager has been shutdown");
+                                break;
+                            }
+                        }
+                    }
+                },
+            }
+        }
```
drmingdrmer, line 124: It does not need a `select!` anymore. :)

metasrv/tests/it/grpc/metasrv_grpc_watch.rs
```diff
@@ -0,0 +1,279 @@
...
+use common_meta_types::Operation;
+use common_meta_types::UpsertKVAction;
+
+use crate::init_meta_ut;
+
+async fn watcher_client_main(
+    addr: String,
+    watch: WatchRequest,
+    start_tx: UnboundedSender<()>,
+    mut events_rx: UnboundedReceiver<Vec<Event>>,
```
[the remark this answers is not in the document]
lichuang: `watcher_client_main` simulate the watcher client, `test_watch_main` simulate the client which update data to metasrv, one for watcher, one for updater.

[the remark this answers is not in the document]
drmingdrmer: Put what `watcher_client_main` does in the test function body and the simple data-updater in another task.
This will get rid of these several channels for communication.

metasrv/tests/it/grpc/metasrv_grpc_watch.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    let h = tokio::spawn(watcher_client_main(addr.clone(), watch, start_tx, events));
+
+    // wait for client stream start up
+    wait_notify(&mut start_rx, 2000);
+
+    // update some kv
+    for update in updates.iter() {
+        let _ = client.upsert_kv(update.clone()).await;
+    }
```
drmingdrmer, line 100: I mean, put these 3 lines in to a task, then you can just remove `watcher_client_main` along with the channels for communication: just move the body of `watcher_client_main` here.

lichuang: `client.upsert_kv` must wait for watcher has been created, so use the channel to notify the watcher has been created, only after that the `update_kv` has meaning in this case.

drmingdrmer: That's not a problem. You can just create the watcher, then spawn the task, then drain the response stream to check the events.

metasrv/tests/it/grpc/metasrv_grpc_watch.rs
```diff
@@ -0,0 +1,217 @@
...
+) -> anyhow::Result<()> {
+    let client = MetaGrpcClient::try_create(addr.as_str(), "root", "xxx", None, None).await?;
+
+    let mut grpc_client = client.make_client().await?;
+
+    let request = tonic::Request::new(watch);
+
+    let mut client_stream = grpc_client.watch(request).await?.into_inner();
+
+    let _h = tokio::spawn(upsert_kv_client_main(addr.clone(), updates));
```
drmingdrmer, line 53:
```suggestion
    tokio::spawn(upsert_kv_client_main(addr.clone(), updates));
```

metasrv/tests/it/grpc/metasrv_grpc_watch.rs
```diff
@@ -0,0 +1,217 @@
...
+
+                        if watch_events.is_empty() {
+                            break;
+                        }
+                    }
+
+                }
+            },
+        }
+    }
```
drmingdrmer, line 73: Waiting for a single channel does not need a `select!`

metasrv/tests/it/grpc/metasrv_grpc_watch.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    loop {
+        tokio::select! {
+            resp = client_stream.message() => {
+                if let Ok(Some(resp)) = resp {
+                    if let Some(event) = resp.event {
+                        assert!(!watch_events.is_empty());
+
+                        assert_eq!(watch_events.get(0), Some(&event));
+                        watch_events.remove(0);
```
drmingdrmer, line 63: This should be enough to guarantee what you expect.
```suggestion
                        assert_eq!(watch_events.remove(0), Some(&event));
```

lichuang: not a big deal

metasrv/tests/it/grpc/metasrv_grpc_watch.rs
lichuang: not a big deal

ariesdevil (approved):
LGTM, and after this PR merged we need a simple doc that let the cloud team know how to use this feature.

drmingdrmer:
/lgtm

trump-lgtm

BohuTANG:
The API adds/remove will be a compatible problem with old databend-metasrv?
If so, how the user upgrade the old databend-metasrv to the new? There is already some users using databend in the production now :100: 
cc @lichuang @drmingdrmer

drmingdrmer:
> The API adds/remove will be a compatible problem with old databend-metasrv? If so, how the user upgrade the old databend-metasrv to the new? There is already some users using databend in the production now 💯 cc @lichuang @drmingdrmer

This feature does not affect data format thus the upgrade would be smooth:)

PR #4781, opened 2022-04-10
docs: unify query http handler port

doki23:

## Summary

In the previous doc(deploy), the Query HTTP Handler listen on port `8081` (in the databend-query.toml). 
So, it's necessary to keep unified here.

## Changelog

- Documentation


BohuTANG (approved):
:+1:

PR #4786, opened 2022-04-10
ISSUE-4765: Show fuse engine table status

### LINKED ISSUE #4765: Show fuse engine table status
**Summary**

After https://github.com/datafuselabs/databend/pull/4757 checked in, we should fetch the status underlying from the fuse engine.
```
mysql> CREATE TABLE t(id int);
mysql> SHOW TABLE STATUS\G
*************************** 1. row ***************************
           Name: t
         Engine: FUSE
        Version: 0
     Row_format: NULL
           Rows: NULL
 Avg_row_length: NULL
    Data_length: NULL
Max_data_length: NULL
   Index_length: NULL
      Data_free: NULL
 Auto_increment: NULL
    Create_time: 2022-04-08 04:13:48.988 +0000
    Update_time: NULL
     Check_time: NULL
      Collation: NULL
       Checksum: NULL
        Comment:
```

dantengsky:

## Summary

- introduce method `statistics`  method for trait `Table`
  ~~~rust
  async fn statistics(&self, ctx: Arc<QueryContext>) -> Result<Option<TableStatistics>>
  ~~~
  for the time being,  `TableStatistics` contains only some basic items.
  ~~~rust
      pub struct TableStatistics {
        pub num_rows: Option<u64>,
        pub data_length: Option<u64>,
        pub data_length_compressed: Option<u64>,
        pub index_length: Option<u64>,
    }
  ~~~
-  impl `async fn statistics` for FuseTable
- exports `TableStatistics` of FuseTable to `System.Tables`

## Changelog

- Improvement

## Related Issues

Fixes #4765


tests/suites/0_stateless/06_show/06_0008_show_table_status.sql
```diff
@@ -11,15 +11,23 @@ SHOW TABLE STATUS LIKE 't%';
...
+SHOW TABLE STATUS WHERE name = 't%' AND 1 = 0;
+SHOW TABLE STATUS WHERE name = 't2' OR 1 = 1;
+SHOW TABLE STATUS WHERE name = 't2' AND 1 = 1;
 
 USE default;
-SHOW TABLE STATUS FROM showtabstat WHERE table_name LIKE 't%';
-SHOW TABLE STATUS FROM showtabstat WHERE table_name = 't%' AND 1 = 0;
-SHOW TABLE STATUS FROM showtabstat WHERE table_name = 't2' OR 1 = 1;
-SHOW TABLE STATUS FROM showtabstat WHERE table_name = 't2' AND 1 = 1;
+SHOW TABLE STATUS FROM showtabstat WHERE name LIKE 't%';
```
sundy-li, line 20: cc @TCeason 
Which one should we use?

`name` or `table_name` ?

Seems `name` is right.

https://www.db-fiddle.com/f/sxCehFV1BHqFbRentu1VRo/0

TCeason: But now we not support alias on select.

TCeason: So may be we should use table_name tmp.

dantengsky: @sundy-li @TCeason 

According to  https://dev.mysql.com/doc/refman/8.0/en/show-table-status.html, seems we should polish further the implementation of `show table status` stmt, such that, not only the schema of the resultset adheres to the spec, but also the columns that could be used in the where-clause.

pls take a look at  81544585

dantengsky: forget to mention: since the column name is case-sensitive, `name` has be changed to `Name`

TCeason: image

ACTUALLY in MySQL, if the var lower_case_table_names=0;  `name` is the same as `Name`, I think that's ok.

dantengsky: > ACTUALLY in MySQL; name is the same as Name, I think that's ok.

roger that

sundy-li (approved):

PR #4791, opened 2022-04-11
refactor(credits): try persist credits at build time

### LINKED ISSUE #4301: Improve: generate a Rust file to save the credits
**Summary**

We currently use [cargo-license](https://github.com/onur/cargo-license) to generate [credits table](https://databend.rs/user/sql-statement/system-tables/system-credits).

This solution requires `Cargo.lock`, so it will be lost during the release process. It may be better to generate a hard-coded file directly.

Related: https://github.com/datafuselabs/databend/issues/2541

PsiACE:

## Summary

After some investigation, we don't need to try to save/embed any credits file, otherwise we would have to parse it ourselves.

Instead, we will extract the relevant information at build time and save it. 

Looks a little dirty, but effective.

## Changelog

- Bug Fix
- Improvement

## Related Issues

Fixes #4301


common/building/src/utils.rs
```diff
@@ -0,0 +1,40 @@
...
+        if v.is_empty() {
+            return Ok(());
+        }
+
+        for s in &self.0[0..v.len() - 1] {
+            comma_separated.push_str(s);
+            comma_separated.push_str(", ");
+        }
+
+        comma_separated.push_str(&self.0[v.len() - 1]);
```
Xuanwo, line 36: How about using `self.0.iter().join(", ")`?

PsiACE: Makes sense. I just wanted a pretty print look at first. Let me try this.

Xuanwo (approved):

PR #4796, opened 2022-04-11
fix: `select *` shouldn't return results

### LINKED ISSUE #4728: Bug: `select *` should not return results
### What's Wrong?

`select *` returns results 
```
mysql> select *;
+-------+
| dummy |
+-------+
|     1 |
+-------+
1 row in set (0.03 sec)
```

### What You Expected?

`select *` reports error;
```
postgres=# select *;
2022-04-07 07:56:53.698 CST [17886] ERROR:  SELECT * with no tables specified is not valid at character 8
2022-04-07 07:56:53.698 CST [17886] STATEMENT:  select *;
ERROR:  SELECT * with no tables specified is not valid
LINE 1: select *;
```

xudong963:

## Summary

- fix `select *`
- fix nightly clippy

## Changelog

- Bug Fix

## Related Issues

Fixes #4728

### DIFF
--- a/query/tests/it/sql/plan_parser.rs
+++ b/query/tests/it/sql/plan_parser.rs
@@ -266,6 +266,12 @@ async fn test_plan_parser() -> Result<()> {
             sql: "DROP ROLE role1",
             expect: "Drop role role1 if_exists:false",
             error: "",
+        },
+        Test {
+            name: "select-without-table",
+            sql: "SELECT *",
+            expect: "",
+            error: "Code: 1015, displayText = SELECT * with no tables specified is not valid (while in analyze select projection)."
         }
     ];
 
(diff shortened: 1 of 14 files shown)


query/tests/it/sql/plan_parser.rs
```diff
@@ -266,6 +266,12 @@ async fn test_plan_parser() -> Result<()> {
             sql: "DROP ROLE role1",
             expect: "Drop role role1 if_exists:false",
             error: "",
+        },
+        Test {
+            name: "select-without-table",
+            sql: "SELECT *",
+            expect: "",
+            error: "Code: 1015, displayText = SELECT * with no tables specified is not valid (while in analyze select projection)."
```
sundy-li, line 274: How about adding a stateless test.  

select *;  -- {ErrorCode 1015}

xudong963: done

sundy-li: Just use:

```
select *; -- {ErrorCode 1015}
```

sundy-li (approved):

Issue #4814, opened 2022-04-12
tracking sharing


youngsofun:
[StateMachine](https://github.com/datafuselabs/databend/blob/0278dfd3514cd63f38d05bbfa4768ab046fd496e/common/meta/raft-store/src/state_machine/sm.rs#L379) key spaces:

1. shares/g_share_id -> share_info(name, option<db_name>, tables)
  1.   tables may use seperate key space when the designe is stable
2. share_lookup/provider/share_name -> g_share_id 
3. share_outbunds/g_share_id/consumer -> share_outbund_meta(created_on)
4. share_inbunds/consumer/g_share_id->  share_inbund_meta(created_on, option<database_name>)
5. databases /db_id ->  db_meta  ##  engine_option.share_id = g_share_id

add meta_api:
- [x] create/drop shares
- [ ] grant/revoke share to tenant
- [ ] grant/revoke privileges to share
- [ ] get share_ext (and more tests for)

 

affected api
- [ ] create/drop database from share
- [ ] get_tables 
- [ ] get_table
- [ ] grant privilege to user/role

SQLs:

- [ ] create/drop share
- [ ] grant/revoke privilige to share
- [ ] grant/revoke share to tenant
- [ ] create/drop database 'db_1' engine='share'（share='tenant_A'.'share_1'）
- [ ] desc share
- [ ] show shares

read 

- [ ] shared table with storage conf

tests:

- [ ] unit test

xudong963:
SM means storage meta? @youngsofun

youngsofun:
state machine @xudong963

PR #4816, opened 2022-04-12
feat(meta/share) add to MetaApi: CreateShare DropShare.

youngsofun:

## Summary

part of https://github.com/datafuselabs/databend/issues/4814

to avoid big pr,  add  CreateShare DropShare first.
also hope @drmingdrmer can point out early if there are problem of my understanding to meta api.

## Changelog

- Not for changelog

## Related Issues

part of https://github.com/datafuselabs/databend/issues/4814


common/meta/api/src/meta_api_test_suite.rs
```diff
@@ -955,6 +957,36 @@ impl MetaApiTestSuite {
...
+
+        tracing::info!("--- drop share1");
+        {
+            mt.drop_share(DropShareReq {
+                if_exists: false,
+                tenant: tenant1.to_string(),
+                share_name: share_name1.to_string(),
+            })
+            .await?;
+        }
```
drmingdrmer, line 986: It has to ensure that a created `share` can be got with `get_share()` and a dropped `share` must not be got anymore. Since the name of this case is `create_get_drop` :)

common/meta/raft-store/src/sled_key_spaces.rs
```diff
@@ -147,6 +154,38 @@ impl SledKeySpace for TableLookup {
...
+}
+
+pub struct ShareInbounds {}
+impl SledKeySpace for ShareInbounds {
+    const PREFIX: u8 = 16;
+    const NAME: &'static str = "share-inbounds";
+    type K = ShareInboundKey;
+    type V = SeqV<ShareInboundValue>;
+}
+
```
drmingdrmer, line 188: What about using an exclusive prefix range for `share` related metas? E.g. `[20,30)`

To ease debugging :DD

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -1140,21 +1157,118 @@ impl StateMachine {
...
+            &share_lookup_tree,
+            &share_lookup_key,
+            &MatchSeq::Exact(0),
+            Operation::Update(ShareLookupValue(share_id)),
+            None,
+        )?;
+
+        // if it is just created
+        if prev.is_some() {
+            let share_id = prev.unwrap().data;
```
drmingdrmer, line 1194: If you need to unwrap it, `if let Some(x) = prev {}` would be better, or using `match` to extract the inner data. Avoid using `unwrap()`

common/meta/raft-store/src/state_machine/sm_meta_api_impl.rs
```diff
@@ -352,6 +359,61 @@ impl MetaApi for StateMachine {
...
+        let tenant = &req.tenant;
+        let share_name = &req.share_name;
+        let if_not_exists = req.if_not_exists;
+
+        tracing::info!("create share: {:}: {:?}", &tenant, &share_name);
+
+        let cr = Cmd::CreateShare {
+            tenant: tenant.clone(),
+            share_name: share_name.clone(),
+        };
```
drmingdrmer, line 372: For a historical reason, `Cmd` does not utilize `XXXReq`. But instead in our codebase it manually converts `XXXReq` to a corresponding `Cmd`.

It would be convenient to let `Cmd` have a variant that includes an `XXXReq` in it. E.g.: 
```rust
enum Cmd {
    CreateShare(CreateShareReq),
}
```

This would save a lot of trouble. :D

common/meta/api/src/meta_api_test_suite.rs
```diff
@@ -955,6 +957,36 @@ impl MetaApiTestSuite {
...
+            let req = CreateShareReq {
+                if_not_exists: false,
+                tenant: tenant1.to_string(),
+                share_name: share_name1.to_string(),
+            };
+
+            let res = mt.create_share(req).await;
+            tracing::info!("create share res: {:?}", res);
+            let res = res.unwrap();
+            assert_eq!(1, res.share_id, "first share id is 1");
```
drmingdrmer, line 975: `Create` needs a test case that creates a `share` when the `share` already exists.

Similar for `DropShare` test.

drmingdrmer (commented):
Nice shot~

common/meta/raft-store/src/sled_key_spaces.rs
youngsofun: good  idea

common/meta/raft-store/src/state_machine/sm.rs
youngsofun: thanks

common/meta/raft-store/src/state_machine/sm_meta_api_impl.rs
youngsofun: is it time to do it now? can I do the refactor by the way?

drmingdrmer: Next time will be a good time:)

common/meta/api/src/meta_api_test_suite.rs
drmingdrmer: If a simple `get_share()` won't bring in too much work, a new feature that is fully tested  would be better:D

drmingdrmer (commented):
The update looks good to me!
162908875-16321925-392f-4451-b6f0-ca15a0c3290a

common/meta/api/src/meta_api_test_suite.rs
youngsofun: added 4 commits for get_share and more tests

common/meta/api/src/meta_api_test_suite.rs
```diff
@@ -90,7 +91,7 @@ impl MetaApiTestSuite {
         tracing::info!("--- create db1 again with if_not_exists=true");
         {
             let req = CreateDatabaseReq {
-                if_not_exists: false,
+                if_not_exists: true,
```
drmingdrmer, line 94: Oh my... Thanks!

common/meta/types/src/share.rs
```diff
@@ -12,11 +12,29 @@
...
+// id used to distinguish share with same name (but never the same time)
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq, Default)]
 pub struct ShareInfo {
-    share_name: String,
-    database: Option<String>,
-    tables: Vec<String>,
+    pub id: u64,
+    pub name: String,
+    pub meta: ShareMeta,
+}
```
drmingdrmer, line 22: Is an `id` enough to identify a `ShareMeta`?

As in `TableInfo`, a tuple of `id, ver`(`TableIdent`) is required to uniquely identify a `TableMeta`. Because a `TableMeta` can be updated and the `TableIdent.ver` will be increased.
https://github.com/datafuselabs/databend/blob/fbba0fca848662eb1e9044d71fc2caf33e38c948/common/meta/types/src/table.rs#L82-L99

youngsofun: I'm not sure: table version refer to things like snapshot?  need it because a query/transaction  need to refer the same version of data.

for shareInfo, user just use the newest updated share  to find the available object inside (a list of ref to tables)
so I think it is enough

youngsofun: @drmingdrmer share is at the database level, maybe has similar conditions, but not very clear to me now.
I also saw SEQ_DATABASE_META_ID, but not found it usage.

currently only think of it as a role object,  a white list of the database/tables behind it.

BohuTANG:
> The update looks good to me! 162908875-16321925-392f-4451-b6f0-ca15a0c3290a

Approved? Man

drmingdrmer:
> Approved? Man

I think there are still some ongoing updates on this PR. 🤔

common/meta/types/src/share.rs
drmingdrmer: Anything that supports modification should have a `version` to identify a specific version of it, inside metasrv.

For a client outside metasrv, without a `version`, one update operation may override another update operation, which may be not expected.

drmingdrmer: Anything that supports modification should have a `version` to identify a specific version of it, inside metasrv.

For a client outside metasrv, without a `version`, one update operation may override another update operation, which may be not expected.

youngsofun: got it,  the `version`  just play the role of seqV.seq in KV api. thank you!

youngsofun: > For a client outside metasrv, without a version, one update operation may override another update operation, which may be not expected.

does unexpected override happen even if all update is done in txn.

drmingdrmer: > got it, the `version` just play the role of seqV.seq in KV api. thank you!

Right. They are the same thing with different names :(



> does unexpected override happen even if all update is done in txn.

Yes. an update request may not specify an expected version to update, i.e., it tries to override existing value without considering the version, by using the argument `MatchSeq::Any`.

```rust
/// Describes what `seq` an operation must match to take effect.
/// Every value written to meta data has a unique `seq` bound.
/// Any conditioned or non-conditioned write operation can be done through the corresponding MatchSeq.
#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq)]
pub enum MatchSeq {
    /// Any value is acceptable, i.e. does not check seq at all.
    Any,

    /// To match an exact value of seq.
    /// E.g., CAS updates the exact version of some value,
    /// and put-if-absent adds a value only when seq is 0.
    Exact(u64),

    /// To match a seq that is greater-or-equal some value.
    /// E.g., GE(1) perform an update on any existent value.
    GE(u64),
}
```

drmingdrmer: > what if the value is never update as a whole, instead api finally call share info.add_table(tablename) in txn , the func check tablename existence
> 
> in this case, no need for seq?

If all operations on a `share` are [commutative](https://en.wikipedia.org/wiki/Commutative_property), no `seq` is required. 

The simplest example of a commutative value is an integer, and only accept addition and substaction operation: `x+1+3` and `x+3+1` result in the same value.

An integer that accepts `+-*/` is not commutative: `x+1*3` may not equal `x*3+1`.

A set of integer that provides methods `Set::insert(i)` and `Set::remove(i)` is not commutative: `{0,1,2}.insert(1).remove(1)` does not equal `{0,1,2}.remove(1).insert(1)`.

A non-commutative value needs a `seq` to serialize every operation on it.

> another choice is make share immutalbe store sub tables in another keyspace

It sounds like a **pointer**, right? Unfortunately, updating a pointer is not a commutative operation. `ptr = a; ptr = b` and `ptr = b; ptr = a` will result in different values.

drmingdrmer: If you can be sure that executing any set of operations in any permutation always results in a valid state, e.g., no lost-update issue, etc, a `seq` is not required.

drmingdrmer (approved):

PR #4841, opened 2022-04-13
support parse select expr as `alias_name`

### LINKED ISSUE #4840: Feature: Support select expr as `alias name`
**Summary**

Link to this issue #[26](https://github.com/datafuse-extras/sqlparser-rs/issues/26)

TCeason:

## Summary

After this pr,  this SQL will return err.

```SQL
SELECT TRUE as `_`  FROM `t1` WHERE 1 <> 1 LIMIT 0;

ERROR 1105 (HY000): Code: 1005, displayText = sql parser error: Expected an identifier after AS, found: `_`.
```
The err occurs in sqlparser, the err has already been fixed [there](https://github.com/datafuse-extras/sqlparser-rs/pull/27).

## Changelog

- New Feature

## Related Issues

Fixes #4840


common/ast/src/udfs/udf_transformer.rs
```diff
@@ -127,9 +127,12 @@ impl UDFTransformer {
                     op: op.clone(),
                     expr: Box::new(Self::clone_expr_with_replacement(&**expr, replacement_fn)?),
                 }),
-                Expr::Cast { expr, data_type } => Ok(Expr::Cast {
+                Expr::Cast {
+                    expr, data_type, ..
+                } => Ok(Expr::Cast {
```
TCeason, line 132: @b41sh  Hi the pg_style can be set as false?

BohuTANG:
This need a stateless test, it is easy to find the bug when the new sqlparser is introduced someday.

TCeason:
> This need a stateless test, it is easy to find the bug when the new sqlparser is introduced someday.

Yep, you are right. Added.

common/ast/src/parser/expr/expr_visitor.rs
```diff
@@ -57,7 +57,11 @@ pub trait ExprVisitor: Sized + Send {
             Expr::Subquery(subquery) => self.visit_subquery(subquery),
             Expr::Function(function) => self.visit_function(function).await,
             Expr::TryCast { expr, data_type } => self.visit_try_cast(expr, data_type).await,
-            Expr::Cast { expr, data_type } => self.visit_cast(expr, data_type).await,
+            Expr::Cast {
+                expr,
+                data_type,
+                pg_style: _pg_style,
```
BohuTANG, line 63: Does the new version of sqlparser introduced `pg_style` params?

common/ast/src/udfs/udf_transformer.rs
BohuTANG: We don't used `pg_style`, i think it's OK here.

common/ast/src/parser/expr/expr_visitor.rs
TCeason: https://github.com/datafuse-extras/sqlparser-rs/commit/b0571e8f91cac092fc8f8873e8babcc05fb91974

Yes in this commit.

BohuTANG: O, @b41sh should know it better :D

common/ast/src/udfs/udf_transformer.rs
b41sh: we can set `pg_style` as false, it is used to distinguish the field names of two different casts.
```sql
select cast(a as int8) from test;
select a::int8 from test;
```
I am working on this modification, you can set it as `false` first.

PR #4844, opened 2022-04-13
chore: delete useless unsafe

xudong963:

## Summary

Delete useless unsafe to reduce confuse

## Changelog

- Other 

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/pipelines/new/executor/executor_tasks.rs
+++ b/query/src/pipelines/new/executor/executor_tasks.rs
@@ -51,7 +51,7 @@ impl ExecutorTasksQueue {
     /// # Safety
     ///
     /// Method is thread unsafe and require thread safe call
-    pub unsafe fn steal_task_to_context(&self, context: &mut ExecutorWorkerContext) {
+    pub fn steal_task_to_context(&self, context: &mut ExecutorWorkerContext) {
         {
             let mut workers_tasks = self.workers_tasks.lock();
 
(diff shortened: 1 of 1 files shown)


query/src/pipelines/new/executor/executor_tasks.rs
```diff
@@ -51,7 +51,7 @@ impl ExecutorTasksQueue {
     /// # Safety
```
zhang2014, line 51: remove safety comment

query/src/pipelines/new/executor/executor_tasks.rs
```diff
@@ -232,10 +230,16 @@ impl ExecutorTasks {
         ExecutorTask::None
     }
 
-    pub unsafe fn push_task(&mut self, worker_id: usize, task: ExecutorTask) {
+    pub fn push_task(&mut self, worker_id: usize, task: ExecutorTask) {
         self.tasks_size += 1;
+        assert!(worker_id < self.workers_sync_tasks.len(), "out of index");
```
zhang2014, line 235: maybe assert_debug is better?

xudong963: Do you think release is not needed because users use our binary directly and will not be aware of it at the code level?

zhang2014 (approved):

query/src/pipelines/new/executor/executor_tasks.rs
zhang2014: Yes.

PsiACE (approved):
LGTM

Issue #4651, opened 2022-04-01
Feature: Adapt to different countries support user change timezone


wubx:
**Summary**

Support databend-query setting the global timezone， To adapt to different countries.

if the user not setting the timezone make it follow the os system.

BohuTANG:
This is a global session setting(work in progress, https://github.com/datafuselabs/databend/issues/3933).
Default should be UTC not the system timezone.
See https://docs.snowflake.com/en/sql-reference/parameters.html#timezone

wubx:
Learn from Snowflake that their timezone is just a session setting.

```
Type
Session — Can be set for Account » User » Session

Data Type
String (Constant)
```
Maybe session setting is enough.

Veeupup:
This feature may have a lot of changes with functions and query files (hundreds of files), because we need to pass the parameter `timezone` from `query` to `functions`. And in the same time, we need to change the logical of date functions, so we need to change it into smaller tasks to make it.

- [ ] pass parameter from `QueryContext` to `FunctionContext` (for now, only for`timezone`)
- [ ] add session setting of timezone
- [ ] modify date functions logic and add more stateless test

PR #4852, opened 2022-04-14
feat: add timezone session settings

Veeupup:

## Summary

this is part of #4651 ,
* add `timezone` session settings.
* make `FunctionContext` be be the first parameter in `Function::eval`

## Changelog

- New Feature

## Related Issues

Fixes #issue


common/functions/src/scalars/function.rs
```diff
@@ -50,7 +50,7 @@ pub trait Function: fmt::Display + Sync + Send + DynClone {
         &self,
         _columns: &ColumnsWithField,
         _input_rows: usize,
-        _func_ctx: FunctionContext,
+        _func_opts: FunctionOptions,
```
b41sh, line 53: I think it's better to name it `FunctionContext` or `Context`, which can later be used to place some data that needs to be passed between functions, such as `warnings`, `hints`, etc. And This parameter should be the first parameter.

sundy-li: I also think `FunctionOptions` is bad name.  Let's keep it `FunctionContext `

Veeupup: ok, let keep `FunctionContext` and make it be the first parameter

Veeupup: and another point, should we make `FunctionContext` in this func be `Arc<FunctionContext>`cc @sundy-li @b41sh

sundy-li: > and another point, should we make `FunctionContext` in this func be `Arc<FunctionContext>`cc @sundy-li @b41sh

It's trivial.

b41sh: Better to use `Arc<FunctionContext>`, there may be a lot of data later.

sundy-li: I do think it's better to use `&mut FunctionContext ` to avoid copy, but it's trivial, we can refactor it later.

common/functions/src/scalars/dates/week_date.rs
```diff
@@ -201,14 +201,22 @@ where
 
         let func = FunctionAdapter::create(func, true);
         let left_val = func
-            .eval(&[args[0].left.clone().unwrap()], 1, FunctionContext {
-                tz: None,
-            })?
+            .eval(
+                FunctionContext {
```
sundy-li, line 205: why not pass the `_func_ctx` to func?

Veeupup: Oh. Miss It. Will be fixed

Veeupup: `timezone` will not affect the function's monotonicity, and we do not have `FunctionContext` parameter in `get_monotonicity` func. 
so maybe a default value will be ok. @sundy-li

sundy-li (approved):

wubx:
Please add a stateless test: how to use it.

sundy-li:
> Please add a stateless test: how to use it.

It's part of, not prod ready.

Issue #4834, opened 2022-04-13
For a historical reason, `Cmd` does not utilize `XXXReq`. But instead in our codebase it manually converts `XXXReq` to a corresponding `Cmd`.


drmingdrmer:
For a historical reason, `Cmd` does not utilize `XXXReq`. But instead in our codebase it manually converts `XXXReq` to a corresponding `Cmd`.

It would be convenient to let `Cmd` have a variant that includes an `XXXReq` in it. E.g.: 
```rust
enum Cmd {
    CreateShare(CreateShareReq),
}
```

This would save a lot of trouble. :D

_Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/pull/4816#discussion_r848332518_

youngsofun:
is refactor like this enough? 

https://github.com/youngsofun/databend/commit/0e844ab05001a49c43752d7bc893053abf2a5b5f

drmingdrmer:
> is refactor like this enough （apply on all cmds that has XXXReq ）?
> 
> [youngsofun@0e844ab](https://github.com/youngsofun/databend/commit/0e844ab05001a49c43752d7bc893053abf2a5b5f)

It looks good:D.
Feel free to fire a **draft** PR instead of a single commit. A PR would be much easier to review and discuss.

PR #4859, opened 2022-04-14
refactor(meta): Cmd utilize XXXReq.

youngsofun:

## Summary

Summary about this PR

## Changelog

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4834


youngsofun:
why ci run only 3 checks ?

BohuTANG:
Oops, not sure.
Here should ping the dad of the CI @Xuanwo

Xuanwo:
> why ci run only 3 checks ?

I'm not sure, maybe something internal error happens so GHA missing our commit events.

Re-run all checks seems a useful workaround.

youngsofun:
>  Re-run all checks seems a useful workaround.

yes, it return to normal after I push again

drmingdrmer (approved):
Oh my... I love nothing more than the reduction of LOC!!!

drmingdrmer:
It seems the metactl test should be updated in accord.

youngsofun:
@drmingdrmer 

It should be fixed by replace field name in meta.json.
I`m writing a simple python to do the translation.

but here we can stop a while to choose the better field name style:

from this file we can see the field name of `XXXCmd`  and `XXXReq`, which do you prefer:
https://github.com/datafuselabs/databend/blob/main/common/meta/types/src/cmd.rs

drmingdrmer:
> `db` or `table` is used to represent a table instance, such as `Arc<dyn Table>`.
> In the files about `category`, there are some name conflict IMHO.
> Thus I prefer a suffix `_name` to let readers know it is more like a `String`.

But Indeed, naming is not a critical issue. Choose a style you feel comfortable with.:)

youngsofun:
not `_meta*` dir, after `RUST_TEST_THREADS=2 cargo test --package databend-meta `.

and the following 2 always fail

```
    meta_node::meta_node_all::test_meta_node_snapshot_replication
    meta_node::meta_node_kv_api::test_meta_node_kv_api
```

drmingdrmer:
Sorry my mistake. It should be `make stateless-cluster-test` to produce a `_meta*` dir.

Let have a look at the failure of the two case.

youngsofun:
```


---- meta_node::meta_node_all::test_meta_node_snapshot_replication stdout ----
Error: timeout after 10s when non-voter replicated all logs .last_log_index -> Some(15) latest: Metrics{id:1,Learner, term:1, last_log:None, last_applied:None, leader:Some(0), membership:{log_id:0-0 membership:[{1}]}, snapshot:None, replication:
thread 'meta_node::meta_node_all::test_meta_node_snapshot_replication' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/test/src/lib.rs:186:5

---- meta_node::meta_node_all::test_meta_node_write_to_local_leader stdout ----
Error: timeout after 10s when add :2 .last_log_index -> Some(4) latest: Metrics{id:2,Learner, term:1, last_log:None, last_applied:None, leader:Some(0), membership:{log_id:0-0 membership:[{2}]}, snapshot:None, replication:
thread 'meta_node::meta_node_all::test_meta_node_write_to_local_leader' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/test/src/lib.rs:186:5

---- meta_node::meta_node_all::test_meta_node_add_database stdout ----
Error: timeout after 10s when  .last_applied -> Some(7) latest: Metrics{id:1,Follower, term:1, last_log:Some(7), last_applied:Some(LogId { term: 1, index: 6 }), leader:Some(0), membership:{log_id:1-7 membership:[{0, 1, 2}]}, snapshot:None, replication:
thread 'meta_node::meta_node_all::test_meta_node_add_database' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/test/src/lib.rs:186:5

---- grpc::metasrv_grpc_kv_api::test_metasrv_kv_api stdout ----
Error: raft stopped

Caused by:
    raft stopped
thread 'grpc::metasrv_grpc_kv_api::test_metasrv_kv_api' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/test/src/lib.rs:186:5

```

drmingdrmer:
Right. These tests are somehow flaky...

youngsofun:
@drmingdrmer  



```
pub struct TableMeta {
    pub schema: Arc<DataSchema>,
    pub engine: String,
    pub engine_options: HashMap<String, String>,
    pub options: HashMap<String, String>,
    pub created_on: DateTime<Utc>,
}
```

the order of  items in HashMap is random, should I change it to BTreeMap?

or  use this method: https://stackoverflow.com/questions/42723065/how-to-sort-hashmap-keys-when-serializing-with-serde


```
       "options": {
            "snapshot_location": "13/56/_ss/9bb26c300a1d4936a6d03de844938ece_v1.json",
            "database_id": "13"
          },
```

drmingdrmer:
> @drmingdrmer
> 
> ```
> pub struct TableMeta {
>     pub schema: Arc<DataSchema>,
>     pub engine: String,
>     pub engine_options: HashMap<String, String>,
>     pub options: HashMap<String, String>,
>     pub created_on: DateTime<Utc>,
> }
> ```
> 
> the order of items in HashMap is random, should I change it to BTreeMap?
> 
> or use this method: https://stackoverflow.com/questions/42723065/how-to-sort-hashmap-keys-when-serializing-with-serde
> 
> or make the check not depend on it?
> 
> ```
>        "options": {
>             "snapshot_location": "13/56/_ss/9bb26c300a1d4936a6d03de844938ece_v1.json",
>             "database_id": "13"
>           },
> ```

I'll always prefer a `BTreeMap` to `HashMap` if the performance is not a concern. :)

It makes tests much easier to be done.

Xuanwo:
> It makes tests much easier to be done.

I'm curious about why. Can you explain it in a bit more detail? Thanks.

Xuanwo:
> or make the check not depend on it?

`HashMap` implements `Eq`. How about compare with a `HashMap` directly instead the raw content?

drmingdrmer:
> > or make the check not depend on it?
> 
> `HashMap` implements `Eq`. How about comparing with a `HashMap` directly instead of the raw content?

In this case, the test is trying to compare two serialized strings :(

youngsofun:
> HashMap implements Eq. How about comparing with a HashMap directly instead of the raw content?

`make metactl-test` compare dumped json log files

youngsofun:
after change xxx_options from hashmap to btreemap.  
seems `make metactl-test` result in error in  log entry seqs

```
killall databend-meta 
killall databend-query
rm -rf _logs*
rm -rf _meta*
make stateless-cluster-test
killall databend-meta 
./target/debug/databend-metactl --export --raft-dir "_meta2" > tests/metactl/meta.json

```

```
make metactl-test

```

only sees Info logs

PR #4845, opened 2022-04-13
fix(executor): fix dead loop for tasks

zhang2014:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

fix(executor): fix dead loop for tasks

## Changelog

- Bug Fix

## Related Issues

Fixes #issue

PR #4862, opened 2022-04-14
Introduce new processor for insert interpreter

### LINKED ISSUE #4799: integrate insert query with new processor
###  Introduce `append2` function like `read2`
- [ ] Implement for Memory Engine
- [ ] Implement for Null Engine
- [ ] Implement for Fuse Engine
    -  Introduce `TransformCompact` to compact the blocks before inseration.
    -  Introduce `AsyncSink` to insert the block into storage and convert the stats seg info into `DataBlock` and pushed the block into QueryContext.
    - Consume the `Vec<DataBlock>` from QueryContext and commit the seg infos.

sundy-li:

## Summary

Summary about this PR

## Changelog

 
- Improvement
 

## Related Issues

Fixes #4799 

Waiting for #4845


common/datavalues/src/columns/series.rs
```diff
@@ -56,6 +56,7 @@ impl Series {
     }
 
     pub fn check_get<T: 'static + Column>(column: &ColumnRef) -> Result<&T> {
+        dbg!("xxx: {}", column.clone());
```
xudong963, line 59: sorry -- I forgot to delete it.

query/src/pipelines/new/processors/sources/stream_source.rs
```diff
@@ -0,0 +1,66 @@
...
+    const NAME: &'static str = "stream source";
+    type BlockFuture<'a> = impl Future<Output = Result<Option<DataBlock>>> where Self: 'a;
+
+    fn generate(&mut self) -> Self::BlockFuture<'_> {
+        async move {
+            match self
+                .stream
+                .take()
+                .ok_or_else(|| ErrorCode::EmptyData("input stream not exist or consumed"))?
+                .next()
```
sundy-li, line 57: This is not right.

1. The Option is  taken and call next.
2. So it will not be taken again, it'll generate at most one block.

xudong963: Oh, got it

will fix by
```rust
          match self
                .stream.as_mut()
                .ok_or_else(|| ErrorCode::EmptyData("input stream not exist or consumed"))?
                .next()
```

xudong963: > Better to Wrap `Box<dyn Source>` to be a StreamSource

Not particularly understood, currently,`StreamSource` is created by `input_stream` in the clickhouse handler, so I wrapped `Option<SendableDataBlockStream>` directly in "StreamSource".

xudong963:
@mergify update

xudong963:
CI still failed, but local is fine. -- I'll deep into the `optimize table` related code to find the reason :D

xudong963:
@mergify update

xudong963:
@mergify update

xudong963:
Wow, ready！！！

PR #4878, opened 2022-04-15
feat(query): support timezone  

### LINKED ISSUE #4651: Feature: Adapt to different countries support user change timezone

Veeupup:

## Summary

this is part of #4651 , make databend work with timezone.

And now, we only store UTC timestamp in microseconds in databend, and then we will display with timezone.
and all functions with timestamp will be the corresponding logic with timezone.
such as in `UTC`, timestamp is `2022-04-30 22:00:00` and then in `Asia/Shanghai`, it should be `2022-05-01 06:00:00`. and when we call `tostartofmonth`, the first will output `2022-04-01` and the latter will be `2022-05-01`. Others will be the same logic.

- [x] serialization with timezone
- [x] deserialization with timezone
- [x] display with timezone
    - [x] mysql
- [x] datetime function
    - [x] number function
    - [x] round function
    - [x] week function
    - [x] cast function
    - [x] interval function 
- [x] stateless test with timezone

## Changelog

- New Feature

## Related Issues

Fixes #4651


Veeupup:
@sundy-li  @b41sh  there are many details to deal with, so comments are welcome if u can help me to check it

common/datavalues/src/types/data_type.rs
```diff
@@ -125,7 +126,15 @@ pub trait DataType: std::fmt::Debug + Sync + Send + DynClone {
 
     fn create_mutable(&self, capacity: usize) -> Box<dyn MutableColumn>;
     fn create_serializer(&self) -> TypeSerializerImpl;
+    /// work only for timestamp serializer
+    fn create_serializer_with_tz(&self, _tz: Tz) -> TypeSerializerImpl {
+        unimplemented!()
+    }
     fn create_deserializer(&self, capacity: usize) -> TypeDeserializerImpl;
+    /// work only for timestamp deserializer
+    fn create_deserializer_with_tz(&self, _capacity: usize, _tz: Tz) -> TypeDeserializerImpl {
```
sundy-li, line 135: Can we pass tz into `FormatSettings` ? Thus we don't need this function.

common/functions/src/scalars/dates/interval_function.rs
```diff
@@ -141,12 +143,15 @@ where
...
     fn eval(
         &self,
-        _func_ctx: FunctionContext,
+        func_ctx: FunctionContext,
         columns: &ColumnsWithField,
         _input_rows: usize,
     ) -> Result<ColumnRef> {
         // Todo(zhyass): define the ctx out of the eval.
-        let mut ctx = EvalContext::new(self.factor, self.precision, None);
+        let tz = func_ctx.tz.parse::<Tz>().map_err(|_| {
```
sundy-li, line 151: Let's use  `Tz` type inside `FunctionContext`.

common/io/src/format_settings.rs
```diff
@@ -36,6 +37,7 @@ impl Default for FormatSettings {
             empty_as_default: false,
             skip_header: false,
             compression: Compression::None,
+            timezone: vec![b'U', b'T', b'C'],
```
b41sh, line 40: How about using `"UTC".as_bytes().to_vec()`? better readability

common/functions/src/scalars/expressions/cast_with_type.rs
```diff
@@ -66,23 +69,27 @@ pub fn cast_column_field(
     column_with_field: &ColumnWithField,
     from_type: &DataTypeImpl,
     target_type: &DataTypeImpl,
+    func_ctx: &FunctionContext,
```
b41sh, line 72: How about putting `func_ctx` as the first argument?

common/datavalues/src/types/data_type.rs
Veeupup: we can not pass tz into `FormatSettings` because `FormatSettings` has derived `Serialize` and `Deserialize` trait and `chrono_tz::Tz` do not implement the trait.

and another way is to pass `FormatSettings` when we call `deserialize_xxxx`, and then we'll need to parse `tz: vec<u8>` to `tz: Tz` every time we call `deserialize_xxxx`

common/functions/src/scalars/dates/interval_function.rs
Veeupup: make sense

common/io/src/format_settings.rs
Veeupup: make sense

sundy-li (approved):

BohuTANG (approved):
:100:

common/functions/src/scalars/expressions/cast_with_type.rs
BohuTANG: We can do this little refactor in another PR

leiysky:
I'd like to know if we support DST(Daylight Saving Time) or not?

As soon as timezone is introduced, the implementation of datetime functions(e.g. `date add`) should take care of DST.

Veeupup:
@leiysky I did not add stateless-test about  DST now. We use `chrono` and `chnoro_tz` crate to deal with datetime convert between different timezones. And it seems that `chrono_tz` has done with it : D

Issue #4882, opened 2022-04-16
Better function name style


BohuTANG:
**Summary**

In Databend, most function names are defined like:
```
avgMax
avgIF
```

For the name of the function, most database use `a_b`(a and b are single word):
```
AVG_MAX
AVG_IF
```
Snowflake: https://docs.snowflake.com/en/sql-reference/functions-all.html
SQLServer: https://docs.microsoft.com/en-us/sql/t-sql/functions/checksum-agg-transact-sql?view=sql-server-ver15

The names of Databend functions should be standardized to make it easier for users to use.

PR #4884, opened 2022-04-16
chore(function): better aggregate function name

BohuTANG:

## Summary

better aggregate function name: xxYy -> xx_yy

## Changelog

- Improvement

## Related Issues

Part of #4882


sundy-li:
1. So this will be not compatible with ClickHouse's function name style, I agree it's better to use snake-case.
2. I think it's better not to add `*_if.md`  to document files, it's hard to cover all combinations, eg: `window_funnel_if`.
    We can add combinators `if.md`, `distinct.md` [documents](https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators/)

BohuTANG:
> 1. So this will be not compatible with ClickHouse's function name style, I agree it's better to use snake-case.
> 2. I think it's better not to add `*_if.md`  to document files, we'll add more combinators & agg functions,  it's hard to cover all combinations(cross join), eg: `window_funnel_if`, `retention_default`.
>    We can add combinators `if.md`, `distinct.md` [documents](https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators/)

Good idea, this PR will keep it then will refactor them in another PR.

PR #4893, opened 2022-04-17
fixes(processor): replace global mutex with atomic

zhang2014:

## Summary

fixes(processor): replace global mutex with atomic

## Changelog

- Performance Improvement

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/pipelines/new/executor/executor_tasks.rs
+++ b/query/src/pipelines/new/executor/executor_tasks.rs
@@ -53,16 +53,15 @@ impl ExecutorTasksQueue {
 
         if !workers_tasks.is_empty() {
             let task = workers_tasks.pop_task(context.get_worker_num());
-            let is_async_task = matches!(&task, ExecutorTask::Async(_));
+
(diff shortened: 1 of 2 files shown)


query/src/pipelines/new/executor/executor_tasks.rs
```diff
@@ -53,16 +53,15 @@ impl ExecutorTasksQueue {
 
         if !workers_tasks.is_empty() {
             let task = workers_tasks.pop_task(context.get_worker_num());
-            let is_async_task = matches!(&task, ExecutorTask::Async(_));
+
+            if matches!(&task, ExecutorTask::Async(_)) {
+                workers_tasks.waiting_async_tasks += 1;
```
BohuTANG, line 58: Does it need atomic operation?

zhang2014: workers_tasks is locked.

zhang2014:
Wrong solution.

Issue #3747, opened 2022-01-04
Migrate to new planner framework


leiysky:
The new planner framework was introduced in #2983 .

Here's the tracking list of migration tasks from original planner to new planner:
- #3748 
- #3749 
- Scalar expression binding: #1537 #1538
- #5134
- Add a setting item to switch between planners #4989 
- #5163
- #5210
- #5203
- #5202
- #5235
- https://github.com/datafuselabs/databend/issues/5240
- #5259 
- #5343
- #5350
- #5353 
- #5360
- #5367
- #5351 
- https://github.com/datafuselabs/databend/issues/5536
- https://github.com/datafuselabs/databend/issues/5647
- https://github.com/datafuselabs/databend/issues/5651
- #5773

xudong963:
Seems we can close the issue

leiysky:
> Seems we can close the issue

I'll close this after new planner is enabled by default.

Xuanwo:
@leiysky The migration has been done, let's go further.

PR #4895, opened 2022-04-17
Refine new planner framework

leiysky:

## Summary

Part of #3747

Refine new planner framework.

Use trait object to represent plans to make it more extensible.

## Changelog

- New Feature


query/src/sql/exec/util.rs
```diff
@@ -28,3 +34,35 @@ pub fn check_physical(expression: &SExpr) -> bool {
...
+    let result = FIELD_NAME_RE.captures(field_name);
+    match result {
+        Some(res) => {
+            if res.len() != 3 {
+                Err(ErrorCode::LogicalError(format!(
+                    "Invalid field name: {}",
+                    field_name
+                )))
+            } else {
+                let name = res.get(1).unwrap().as_str().to_string();
```
zhang2014, line 58: replace `unwrap` with `?`

leiysky: https://docs.rs/regex/latest/regex/struct.Captures.html#method.get

`Captures::get` actually returns a `Option<Match>`.

Since the length of `res` has been checked here, it's safe to `unwrap` the `Option`.

Xuanwo: If the length has been checked, how about use `res[1]` instead?

Xuanwo: If the length has been checked, how about using `res[1]` instead?

Xuanwo: If the length has been checked, how about use `res[1]` instead?

Xuanwo: If the length has been checked, how about using `res[1]` instead?

Xuanwo: If the length has been checked, how about using `res[1]` instead?

Xuanwo: If the length has been checked, how about use `res[1]` instead?

query/src/sql/exec/mod.rs
```diff
@@ -12,124 +12,109 @@
...
+        let output_schema = schema_builder.build_project(project, input_schema.clone())?;
 
-        let mut exprs = vec![];
-        for item in project.items.iter() {
-            let alias = self.metadata.column(item.index).name.clone();
-            let builder = ExpressionBuilder::create(&self.metadata);
-            let expr = builder.build(&item.expr)?;
-            let expr = Expression::Alias(alias, Box::new(expr));
-            exprs.push(expr);
+        let mut expressions = vec![];
```
Xuanwo, line 89: If the length of `project.items` is already known, how about using `Vec::with_capacity()` here?

query/src/sql/exec/mod.rs
```diff
@@ -12,124 +12,109 @@
...
+        let expr_builder = ExpressionBuilder::create(&self.metadata);
+        for expr in project.items.iter() {
+            let scalar = expr.expr.as_any().downcast_ref::<Scalar>().unwrap();
```
Xuanwo, line 92: How about using `expect` so that we can know more about this panic?

query/src/sql/exec/util.rs
```diff
@@ -28,3 +34,35 @@ pub fn check_physical(expression: &SExpr) -> bool {
...
+}
+
+/// Decode a field name into display name and index
+pub fn decode_field_name(field_name: &str) -> Result<(String, IndexType)> {
+    let result = FIELD_NAME_RE.captures(field_name);
+    match result {
+        Some(res) => {
+            if res.len() != 3 {
+                Err(ErrorCode::LogicalError(format!(
+                    "Invalid field name: {}",
```
Xuanwo, line 54:
```suggestion
                    "Invalid field name: {field_name}",
```

query/src/sql/exec/util.rs
```diff
@@ -28,3 +34,35 @@ pub fn check_physical(expression: &SExpr) -> bool {
...
+                    field_name
+                )))
+            } else {
+                let name = res.get(1).unwrap().as_str().to_string();
+                let index = res.get(2).unwrap().as_str().parse::<IndexType>()?;
+                Ok((name, index))
+            }
+        }
+        None => Err(ErrorCode::LogicalError(format!(
+            "Invalid field name: {}",
```
Xuanwo, line 64:
```suggestion
            "Invalid field name: {field_name}",
```

query/src/sql/optimizer/rule/transform_state.rs
```diff
@@ -14,20 +14,21 @@
...
 use crate::sql::optimizer::s_expr::SExpr;
 
+#[derive(Clone, Default)]
 pub struct TransformState {
     results: Vec<SExpr>,
 }
 
 impl TransformState {
-    pub fn create() -> Self {
+    pub fn new() -> Self {
```
Xuanwo, line 23: If there is no inputs for `new`, how about using `default()` instead?

query/src/sql/planner/binder/bind_context.rs
```diff
@@ -135,8 +106,10 @@ impl BindContext {
         let mut result = vec![];
         if let Some(table) = table {
             for column_binding in self.columns.iter() {
-                if column_binding.table_name.eq(&table) && column_binding.column_name.eq(&column) {
-                    result.push(column_binding.clone());
+                if let Some(table_name) = &column_binding.table_name {
+                    if table_name.eq(&table) && column_binding.column_name.eq(&column) {
```
Xuanwo, line 110: How about using `==` directly?

BTW, in which situation `table_name` will `!= table`?

query/src/sql/planner/metadata.rs
```diff
@@ -84,15 +89,10 @@ pub struct Metadata {
...
-        Self::default()
-    }
-
-    fn next_table_index(&self) -> IndexType {
-        self.tables.len()
-    }
-
-    fn next_column_index(&self) -> IndexType {
-        self.columns.len()
+        Self {
```
Xuanwo, line 92: How about implementing `Default` for `Metadata`?

query/src/sql/planner/plans/physical_scan.rs
```diff
@@ -12,39 +12,51 @@
...
+impl BasePlan for PhysicalScan {
+    fn plan_type(&self) -> PlanType {
+        PlanType::PhysicalScan
     }
-}
 
-impl PhysicalPlan for PhysicalScan {
-    fn compute_physical_prop(&self, _expression: &SExpr) -> PhysicalProperty {
-        PhysicalProperty::default()
+    fn is_physical(&self) -> bool {
```
Xuanwo, line 37: Is there a plan that both `physical` or `logical`？ Or neither `physical` and `logical`？

query/src/sql/exec/util.rs
leiysky: > If the length has been checked, how about using `res[1]` instead?

Good idea. I haven't notice that `Captures` implemented `Index` trait.

query/src/sql/exec/mod.rs
leiysky: I was considering introducing a `assert_downcast` function later.

query/src/sql/optimizer/rule/transform_state.rs
leiysky: I'm not sure that it won't have any input. Thus I prefer to keep this.

query/src/sql/planner/binder/bind_context.rs
leiysky: In the case that there are multiple tables in current context. For example, `SELECT * FROM t, t1`.

query/src/sql/planner/plans/physical_scan.rs
leiysky: Yes, `ProjectPlan` is both physical and logical.

And `PatternPlan` is neither physical nor logical.

query/src/sql/exec/mod.rs
leiysky: Fixed.

query/src/sql/exec/util.rs
leiysky: Fixed.

query/src/sql/exec/util.rs
leiysky: Fixed.

query/src/sql/planner/metadata.rs
leiysky: Ditto, the `Default` trait doesn't seem reasonable to me.

andylokandy (approved):

query/src/interpreters/interpreter_select_v2.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::sessions::QueryContext;
+use crate::sql::Planner;
+
+/// Interpret SQL query with new SQL planner
+pub struct SelectInterpreterV2 {
+    ctx: Arc<QueryContext>,
+    query: String,
+}
+
+impl SelectInterpreterV2 {
```
xudong963, line 34: Maybe we can directly use the current `SelectInterpreter`

xudong963 (commented):
Will the refactoring of the directory structure in `common/ast` be done later?

query/src/interpreters/interpreter_select_v2.rs
BohuTANG: Seems we can add a new trait like `execute2` in SelectInterpreter for the new planner, this what new processor did.

BohuTANG: And `InterpreterFactory` should add a new factory method like `get2` too.

BohuTANG:
Cool, let's merge first.

leiysky:
> Will the refactoring of the directory structure in `common/ast` be done later?

Yes, would you like to refine it?

PR #4897, opened 2022-04-17
Set interpolateParams in golang dev doc

july2993:

## Summary
since do not support prepared statements
set interpolateParams in golang development doc.

or user needs to avoid using placeholders and may take care to escape the value itself.

## Changelog

- Documentation

## Related Issues

Fixes #issue


BohuTANG (approved):
:+1:

Issue #4890, opened 2022-04-17
bug: metadata  error 


oslet:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

main: 
commitid: d6fe2d1e17ae4ebc984eb44be48201115350c725
tag: v0.7.20-nightly

### What's Wrong?

````
2022-04-17T00:07:53.557980Z ERROR openraft::core: quit RaftCore::main on error err=StorageError(IO { source: StorageIOError { subject: Logs, verb: Read, source: common_meta_types::meta_storage_errors::MetaStorageError: serde json se/de error: Error("missing field `if_not_exists`", line: 1, column: 228), backtrace: "<disabled>" } })
````

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

BohuTANG:
Hi, @oslet 

Thank you for the information.
When your metadata created first time?

Is there any way to get the metadata version and time in the disk? cc @drmingdrmer

drmingdrmer:
It must be upgraded and the data format has been changed in:
- #4816

drmingdrmer:
This upgrade added field `if_xxx_exist` to metadata.
Let me find a way to solve this issue 🤔

oslet:
It is recommended to add scheduled backup functions, such as hourly backup configuration to the object storage directory

drmingdrmer:
If you are using a single-node metasvr a backup may be necessary.

But a backup should be considered only for data durability. It may not help in case data format is changed.:(

BohuTANG:
@oslet 

Agreed.
The metadata is small now(only table schema, user and fuse engine root snapshot), but it's very very important, I think the metasrv cluster for HA and Data Reliability, but this is not enough, the data maybe corruption by the metasrv bug.

I agree with you to backup metadata to object or other places with version, this is a solid last resort for the metasrv.

BohuTANG:
@drmingdrmer 
metasrv also need to add some tests to check the version compatibly for the current version with the previous version of metadata changed.
cc @drmingdrmer

BohuTANG:
Hi @oslet 

We have made databend-meta forward compatible from v0.7.22, please use https://github.com/datafuselabs/databend/releases/tag/v0.7.22-nightly

PR #4901, opened 2022-04-17
feature: metasrv has to be compatible with 20220413-34e89c99e4e35632718e9227f6549b3090eb0fb9

drmingdrmer:

##### feature: metasrv has to be compatible with 20220413-34e89c99e4e35632718e9227f6549b3090eb0fb9
Let metasvr be able to load the log-entry format before commit
34e89c99e4e35632718e9227f6549b3090eb0fb9,
in which the `Cmd` format changed.

To solve th compatible issue, an intermedia type is introduced, which is
a super set of the old and new type.
When loading a record, first load it into the superset type, then reduce
to the latest type.

- fix: #4890

## Changelog

- New Feature

## Related Issues

### DIFF
--- a/common/meta/types/src/cmd.rs
+++ b/common/meta/types/src/cmd.rs
@@ -191,25 +191,16 @@ impl<'de> Deserialize<'de> for Cmd {
                 table_name,
                 table_meta,
             } => {
-                if let Some(x) = if_not_exists {
-                    // latest
-                    Cmd::CreateTable(CreateTableReq {
-                        if_not_exists: x,
-                        tenant,
-                        db_name,
-                        table_name,
-                        table_meta,
-                    })
-                } else {
-                    // 20220413
-                    Cmd::CreateTable(CreateTableReq {
-                        if_not_exists: false,
-                        tenant,
-                        db_name,
-                        table_name,
-                        table_meta,
-                    })
-                }
+                // since 20220413 there is an `if_exists` field.
+                let if_not_exists = if_not_exists.unwrap_or_default();
+
+                Cmd::CreateTable(CreateTableReq {
+                    if_not_exists,
+                    tenant,
+                    db_name,
+                    table_name,
+                    table_meta,
+                })
             }
             cmd_00000000_20220413::Cmd::DropTable {
                 if_exists,
@@ -217,23 +208,15 @@ impl<'de> Deserialize<'de> for Cmd {
                 db_name,
                 table_name,
             } => {
-                if let Some(x) = if_exists {
-                    // latest
-                    Cmd::DropTable(DropTableReq {
-                        if_exists: x,
-                        tenant,
-                        db_name,
-                        table_name,
-                    })
-                } else {
-                    // 20220413
-                    Cmd::DropTable(DropTableReq {
-                        if_exists: false,
-                        tenant,
-                        db_name,
-                        table_name,
-                    })
-                }
+                // since 20220413 there is an `if_exists` field.
+                let if_exists = if_exists.unwrap_or_default();
+
+                Cmd::DropTable(DropTableReq {
+                    if_exists,
+                    tenant,
+                    db_name,
+                    table_name,
+                })
             }
             cmd_00000000_20220413::Cmd::RenameTable {
                 if_exists,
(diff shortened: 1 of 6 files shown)


BohuTANG:
Is it possible set a version for the metadata, default is version_0, if the metadata changed upgrade it to version_1, and we handle the versions?
This time we can force the user to upgrade, next time will compatibly with the old version.

lichuang:
i wonder if there is more elegant way to do compatible in Rust enum type, like protobuf `optional` type? 

In this PR's way, the next time Cmd upgrade, it will has a new cmd_xxxx type to compatible, it is boring....

lichuang:
> i wonder if there is more elegant way to do compatible in Rust enum type, like protobuf `optional` type?
> 
> In this PR's way, the next time Cmd upgrade, it will has a new cmd_xxxx type to compatible, it is boring....

or, IMHO, if the CMD type may be upgrade in the future and it is the key data of meta API, we can define it as a protobuf type, to make it compatible easily?

flaneur2020:
how about reduce the business logics like CreateDatabase / CreateShare into the raft command, instead make them depends on some generic K/V operations like PutWriteBatch / CommitTx?

this may reduce the possbility of changes about low level raft commands

lichuang:
> how about reduce the business logics like CreateDatabase / CreateShare into the raft command, instead make them depends on some generic K/V operations like PutWriteBatch / CommitTx?
> 
> this may reduce the possbility of changes about low level raft commands

emm, good idea.

drmingdrmer:
> Is it possible set a version for the metadata, default is version_0, if the metadata changed upgrade it to version_1, and we handle the versions? This time we can force the user to upgrade or removed the if-exists from the metadata, next time will compatibly with the old version.

Right. It is the second part of the plan.
Next time to upgrade, we need to:
1. detect the version changes, as you mentioned above,
2. and load the old version data then convert them to the new version. This is what this PR does.

drmingdrmer:
> i wonder if there is more elegant way to do compatible in Rust enum type, like protobuf `optional` type?

The compatible intermediate struct `cmd_20220413.Cmd` loads the data in this way,
and convert to a struct without `Option` fields.

Using `Option` directly in the latest `Cmd` is not considered because it may confuse the user by implying that a field is optional.

> In this PR's way, the next time Cmd upgrade, it will has a new cmd_xxxx type to compatible, it is boring....

It does not have to be, if we can limit the oldest version that is backward compatible with, e.g.: the publish metasvr version could be:
- version A;
- version B that is compatible with A
- version B that is **NOT** compatible with A
- version C that is compatible with B
- ...

And there probably is not another `cmd_xxx`, to be compatible with two older version, the max compatible `Cmd` has to be able to load both old version A and old version B.
Thus the max compatible `Cmd` will just become a bit larger...

drmingdrmer:
> or, IMHO, if the CMD type may be upgrade in the future and it is the key data of meta API, we can define it as a protobuf type, to make it compatible easily?

Compatibility is not an issue in such a case. Actually, rust serde provides more flexible ser/de ability. Such as [Field attributes](https://serde.rs/field-attrs.html)

The dirty job is how to avoid using `Option` in a struct, to prevent abuse of  `unwrap()` etc.

drmingdrmer:
> how about reduce the business logics like CreateDatabase / CreateShare into the raft command, instead make them depends on some generic K/V operations like PutWriteBatch / CommitTx?
> 
> this may reduce the possbility of changes about low level raft commands

This solution is considered before but we were facing the fact that CreatDatabase and else are not pure **batch** write but conditional updates.
It requires the metasrv to provide a small language with which a complete tx can be expressed as this demo illustrated(current solution and the tx-language solution):
```rust
fn create_table() {
    let xx = dbs.get((args.db_name, args.table_name));
    if xx.is_ok() {
        return Ok();
    } else {
        let id = new_id();
        dbs.insert((args.db_name, args.table_name), id);
        tables.insert((args.db_name, id), args.table);
        return Ok();
    }
}
```

```rust
pub enum Cmd {

    /// Conditional evaluate: 
    /// - first evaluate `condition`, it has to be a bool value.
    /// - then evaluate then_cmd or else_cmd
    Cmd::IfThenElse {
        condition: Cmd,
        then_cmd: Cmd,
        else_cmd: Cmd,
    }

    /// Assign a Cmd and save the output into a temporary variable with name `var`.
    ///
    /// 
    Cmd::Assign {
        var: String, 
        cmd: Cmd
    }

    /// A series of Cmd that will be evaluated one by one.
    Cmd::Statements(Vec<Cmd>),

    /// Return true if a variable is `None`.
    Cmd::IsNull(String), 

    /// Return the value of a variable, the value type is `AppliedState`.
    Cmd::Var(String), 


    /// Generate a new unique id in a specified name space.
    Cmd::NewId(String), 

    // Tree operations

    /// alias of BTreeMap.get()
    ///
    Cmd::GetKV{
        /// The tree to acces.
        /// Only predefined trees are provided, such as:
        /// table, db, 
        tree: String, 
        key: impl IntoTreeKey, 
    }
}

///

Statements(vec![

    Assign {
        var: "curr",
        cmd: GetKV{
            tree: "db_tbl_id",
            key: (args.db_name, args.table_name)
        }
    },
    IfThenElse {
        condition: IsNull("curr"),
        then_cmd: Var("curr"),
        else_cmd: Statements(vec![
            Assign {
                var: "id",
                cmd: NewId("table"),
            },
            Upsert {
                tree: "db_tbl_id",
                key: (args.db_name, args.table_name),
                value: Var("id"),
            },
            Upsert {
                tree: "db_tbl_t",
                key: (args.db_name, Var("id")),
                value: args.table,
            },
            Var(id)
        ])
    }

])
```

drmingdrmer:
Converted to draft because several test should be added before merge.:(

drmingdrmer:
@lichuang @ariesdevil anything need to improve?

ariesdevil (approved):
lgtm

Issue #4875, opened 2022-04-15
Feature: create user if not exists on JWT authenticate


Junnplus:
**Summary**

Create user if not exists on JWT authenticate.

Issue #4572, opened 2022-03-25
Feature: Disable create/drop of root


junnplus:
**Summary**

Creation and deletion of root users should be disabled.

PR #4924, opened 2022-04-18
feat: create user if not exists on JWT authenticate

Junnplus:

## Summary

Create user if not exists on JWT authenticate.

Specify the `create_user` struct in custom claims to create a user.
```rs
#[derive(Default, Deserialize, Serialize)]
pub struct CreateUser {
    pub tenant_id: Option<String>,
    pub roles: Vec<String>,
}

#[derive(Default, Deserialize, Serialize)]
pub struct CustomClaims {
    pub create_user: Option<CreateUser>,
}
```

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4875

### DIFF
--- a/query/src/users/auth/auth_mgr.rs
+++ b/query/src/users/auth/auth_mgr.rs
@@ -19,8 +19,6 @@ use common_exception::Result;
 use common_meta_types::AuthInfo;
 use common_meta_types::UserIdentity;
 use common_meta_types::UserInfo;
-use serde_json::json;
-use serde_json::Value;
 
 pub use crate::configs::Config;
 use crate::users::auth::jwt::JwtAuthenticator;
@@ -62,22 +60,18 @@ impl AuthMgr {
         match credential {
             Credential::Jwt { token: t } => {
                 let jwt = match &self.jwt {
-                    Some(j) => j.get_jwt(t.as_str())?,
+                    Some(j) => j.parse_jwt_claims(t.as_str())?,
                     None => return Err(ErrorCode::AuthenticateFailure("jwt auth not configured.")),
                 };
                 let user_name = jwt.subject.unwrap();
-                let tenant = jwt
-                    .custom
-                    .get("tenant")
-                    .unwrap_or(&json!(self.tenant))
-                    .to_string();
-                let mut user_info = UserInfo::new(&user_name, "%", AuthInfo::JWT);
-                if let Some(Value::Array(roles)) = jwt.custom.get("roles") {
-                    for role in roles {
-                        user_info.grants.grant_role(role.to_string());
+                if let Some(create_user) = jwt.custom.create_user {
+                    let tenant = create_user.tenant_id.unwrap_or_else(|| self.tenant.clone());
+                    let mut user_info = UserInfo::new(&user_name, "%", AuthInfo::JWT);
+                    for role in create_user.roles {
+                        user_info.grants.grant_role(role);
                     }
+                    self.user_mgr.add_user(&tenant, user_info, true).await?;
                 }
-                self.user_mgr.add_user(&tenant, user_info, true).await?;
                 self.user_mgr
                     .get_user(&self.tenant, UserIdentity::new(&user_name, "%"))
                     .await
--- a/query/src/users/auth/auth_mgr.rs
+++ b/query/src/users/auth/auth_mgr.rs
@@ -66,8 +66,12 @@ impl AuthMgr {
                     None => return Err(ErrorCode::AuthenticateFailure("jwt auth not configured.")),
                 };
                 let user_name = jwt.subject.unwrap();
-                let tenant = jwt.custom.get("tenant").unwrap_or(&json!(self.tenant)).to_string();
-                let mut user_info = UserInfo::new(user_name.to_string(), "%".to_string(), AuthInfo::JWT);
+                let tenant = jwt
+                    .custom
+                    .get("tenant")
+                    .unwrap_or(&json!(self.tenant))
+                    .to_string();
+                let mut user_info = UserInfo::new(&user_name, "%", AuthInfo::JWT);
                 if let Some(Value::Array(roles)) = jwt.custom.get("roles") {
                     for role in roles {
                         user_info.grants.grant_role(role.to_string());
--- a/query/src/users/auth/auth_mgr.rs
+++ b/query/src/users/auth/auth_mgr.rs
@@ -19,6 +19,8 @@ use common_exception::Result;
 use common_meta_types::AuthInfo;
 use common_meta_types::UserIdentity;
 use common_meta_types::UserInfo;
+use serde_json::json;
+use serde_json::Value;
 
 pub use crate::configs::Config;
 use crate::users::auth::jwt::JwtAuthenticator;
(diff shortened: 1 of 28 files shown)


query/src/users/auth/auth_mgr.rs
```diff
@@ -42,28 +42,37 @@ pub enum Credential {
...
             Credential::Jwt { token: t } => {
-                let user_name = match &self.jwt {
-                    Some(j) => j.get_user(t.as_str())?,
+                let jwt = match &self.jwt {
+                    Some(j) => j.get_jwt_claims(t.as_str())?,
                     None => return Err(ErrorCode::AuthenticateFailure("jwt auth not configured.")),
                 };
-                self.users
+                let user_name = jwt.subject.unwrap();
+                if let Some(create_user) = jwt.custom.create_user {
```
youngsofun, line 67: 1. `add_user`  return error if user exists. does the cloud platform when it should carry the create_user field?

2. for an existed user,  what  if  the existed user info diff with info in create_user?

query/src/users/auth/auth_mgr.rs
```diff
@@ -42,28 +42,37 @@ pub enum Credential {
...
+                    let tenant = create_user.tenant_id.unwrap_or_else(|| self.tenant.clone());
+                    let mut user_info = UserInfo::new(&user_name, "%", AuthInfo::JWT);
```
youngsofun, line 69: what to do with special names like "root"

query/src/users/auth/auth_mgr.rs
```diff
@@ -42,28 +42,37 @@ pub enum Credential {
...
+                    for role in create_user.roles {
+                        user_info.grants.grant_role(role);
+                    }
+                    self.user_mgr.add_user(&tenant, user_info, true).await?;
+                }
+                self.user_mgr
```
youngsofun, line 75: if add_user does return ok,  no need to get_user here?

query/src/users/auth/jwt/authenticator.rs
```diff
@@ -44,15 +64,15 @@ impl JwtAuthenticator {
         Ok(Some(JwtAuthenticator { key_store }))
     }
 
-    pub fn get_user(&self, token: &str) -> Result<String> {
+    pub fn get_jwt_claims(&self, token: &str) -> Result<JWTClaims<CustomClaims>> {
```
flaneur2020, line 67: how about naming as `parse_jwt_claims` ?

query/src/users/auth/auth_mgr.rs
Junnplus: > add_user return error if user exists. does the cloud platform when it should carry the create_user field?

User will be created only if user does not exist. `is_not_exists` is true for `add_user `.

> for an existed user, what if the existed user info diff with info in create_user?

`create_user` just used to create users, not update.

query/src/users/auth/auth_mgr.rs
Junnplus: If the user already exists, we still need to get the user.

query/src/users/auth/jwt/authenticator.rs
Junnplus: SGTM.

query/src/users/auth/auth_mgr.rs
Junnplus: Should be dealt with with https://github.com/datafuselabs/databend/issues/4572 ?

flaneur2020: is root can be accessed with JWT? i guess it'd be safer to restrict the root access by %'127.0.0.1' q.q

Junnplus: > i guess it'd be safer to restrict the root access by %'127.0.0.1'

Agree, so we should restrict the creation of root. see https://github.com/datafuselabs/databend/issues/4572

youngsofun: > Should be dealt with with #4572 ?

I thought #4572 mean to restrict it in the parser.  Ok to deal with it later.

query/src/users/auth/auth_mgr.rs
youngsofun: > User will be created only if user does not exist

do you mean  cloud platform do this check  to determine whether or not to add`create_user` in jwt?  cc @flaneur2020   


`add_user` may ErrorCode::UserAlreadyExists

query/src/users/auth/auth_mgr.rs
youngsofun: why not return the UserInfo just created?

query/src/users/auth/auth_mgr.rs
flaneur2020: > do you mean cloud platform do this check to determine whether or not to addcreate_user in jwt?

i guess the cloud platform could ensure user exists by an jwt with create_user enabled when the user successfully logged in, and then pass normal jwt for the normal operations to interact with databend-query.

a jwt is used to authenticate an user, if the jwt verified successfully and the user already existed, it can pass the authentication happily, it might a bit weird to get a userAlreadyExists in this case. 🤔

flaneur2020: btw, how about making the `CreateUser` claim into a bool?

the infomation about tenant and role may already existed some another claims 🤔

Junnplus: > `add_user` may ErrorCode::UserAlreadyExists

@youngsofun `UserAlreadyExists` will be ignored if the user already exists by `is_not_exists` is true.

query/src/users/auth/auth_mgr.rs
Junnplus: > why not return the UserInfo just created?

User will not be created if user already exists, so we need to `get_user`.

query/src/users/auth/auth_mgr.rs
Junnplus: > btw, how about making the create_user claim into a bool?

like this?
```
{
    "create_user": true,
    "tenant_id": "",
    "roles": []
}
```

I'm not sure if the same fields would be in custom claims, it's better together.

youngsofun: @Junnplus sorry, what I  saw is UserApi::add_user.

query/src/users/auth/auth_mgr.rs
youngsofun: so, put it together,  Jwt only create a user if not exist.
the user may be altered later as normal.

lgtm

youngsofun (approved):

query/src/users/auth/auth_mgr.rs
youngsofun: > i guess the cloud platform could ensure user exists by an jwt with create_user enabled when the user successfully logged in, and then pass normal jwt for the normal operations to interact with databend-query.


maybe we can call it  ```ensure_user```?  I saw it a lot in puppet.  or "create_user_if_not_exist".

> I'm not sure if the same fields would be in custom claims, it's better together

seems currently this `create_user` provides only a `basic` template for a newly-seen user, and it will not override behavior specified by an existing userInfo, I prefer separate from other fields (which may later be used to override some behavior of the session)

Junnplus: `ensure_user` SGTM.

flaneur2020: cool, `ensure_user` is much better than `create_user_if_not_exist`.

I guess jwt in some manner favors shorter names, as it'd like to avoids getting too big in the transport, likewise "aud" over "audience", "sub" over "subject" etc.

PR #4934, opened 2022-04-19
chore(datatype): change all the test to new data type and refine the doc

BohuTANG:

## Summary

* Change `typeof` to display the new data type name
* Change all the docs to new data type name
* Change the codeblock style 

## Changelog

- Improvement
- Documentation

## Related Issues

Fixes #issue


common/functions/src/scalars/others/type_of.rs
```diff
@@ -61,7 +59,8 @@ impl Function for TypeOfFunction {
         columns: &common_datavalues::ColumnsWithField,
         input_rows: usize,
     ) -> Result<common_datavalues::ColumnRef> {
-        let type_name = format!("{:?}", columns[0].data_type());
+        let non_null_type = remove_nullable(columns[0].data_type());
+        let type_name = format_data_type_sql(&non_null_type);
```
sundy-li, line 63: Shall we return the nullable sign?

Eg: `Nullable(VARCHAR)`

If yes, we should we `columns[0].data_type().sql_name()`.

BohuTANG: Only type name is better here?
I think the user don't understand what's `Nullable(VARCHAR)` mean :thinking:

sundy-li: Ok

PR #4944, opened 2022-04-19
feature: support REGEXP_REPLACE function

### LINKED ISSUE #3057: REGEXP_REPLACE() | Replace substrings matching regular expression

nange:

## Summary

* Support REGEXP_REPLACE function
* Make the logic of arguments validating more reasonable and consistent for `regexp_*` functions (the same as snowflake)
* Mini fix for how-to-write-scalar-functions doc

## Changelog

- New Feature
- Improvement
- Documentation

## Related Issues

Fixes #3057


common/functions/src/scalars/strings/regexp_like.rs
```diff
@@ -100,6 +100,10 @@ impl Function for RegexpLikeFunction {
             mt,
         )?))
     }
+
+    fn passthrough_constant(&self) -> bool {
```
sundy-li, line 104: I think it's incorrect.

`passthrough_constant` only works when all arguments are constant. So it's ok to `check_get` ConstColumn.

common/functions/src/scalars/strings/regexp_instr.rs
```diff
@@ -69,7 +74,7 @@ impl Function for RegexpInStrFunction {
     }
 
     fn return_type(&self) -> DataTypePtr {
-        u64::to_data_type()
+        NullableType::arc(u64::to_data_type())
```
sundy-li, line 77: Why do we return `NullableType` and `disable_passthrough_null`?

It makes the code more complex.

sundy-li: Oh, it's same as snowflake did.

common/functions/src/scalars/strings/regexp_like.rs
nange: Thanks, fixed.  And I also have simplified some of the code for regexp_like function.

sundy-li (approved):

common/functions/src/scalars/strings/regexp_instr.rs
nange: Yes. If we enable the `passthrough_null` , we can't know original of the argument(eg. pos) is `0` or `null` in `eval` function.

Issue #4362, opened 2022-03-08
bug: config overwrite when specify `--config` and any other command line args.


DCjanus:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

git commit ID: a6b2d4db98492ebfa871f277d57616e7f0335110

### What's Wrong?

For example, with [those lines](https://github.com/datafuselabs/databend/blob/a6b2d4db98492ebfa871f277d57616e7f0335110/query/src/bin/databend-query.rs#L37-L49), if we specify command line arguments like this `--config config.toml --log_level DEBUG`, the log level would not be `DEBUG`, if not specified from other source.

### What You Expected?

Merge command line arguments with other configuration sources.

### Anything Else?

There is a [crate](https://docs.rs/config/latest/config/) provide multiple configuration sources support, but to combine with `clap`, there would be a lot of job to be done.

To resolve this issue, maybe we should discuss about whole config system.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

DCjanus:
If someone trying to fix this, feel free to assign yourself.

BohuTANG:
Currently, `Config::load_from_args` and `Config::load_from_toml` are exclusive, for this issue, we should do some merge between them.

DCjanus:
Sorry for delay, there is so many boring jobs in my official job to lose energy. I've noticed that my friend @Xuanwo  has been took this issue, that's really great.

About this issue, there is draft in my mind: keep config types that drived with `clap::Parser` or `clap::Args`, and implement `config::Source` to `databend_query::config::Config`, and then, code would be like this:

```rust
impl Config {
    pub fn load() -> anyhow::Result<Self> {
        let c = Config::parse();

        let mut cfg_builder = ::config::Config::builder();

        let clap_source = ::config::Config::try_from(&c)?;
        cfg_builder = cfg_builder.add_source(clap_source);

        if !c.config_file.is_empty() {
            let config_file = Path::new(&c.config_file);
            let file_source = ::config::File::from(config_file);
            cfg_builder = cfg_builder.add_source(file_source);
        }

        let c = cfg_builder.build()?.try_deserialize()?;

        Ok(c)
    }
}
```

And boom, now we've able to load config from env, cli and local config file, with order.

Xuanwo:
I happen to have similar thoughts: https://github.com/mehcode/config-rs/issues/315

PR #4971, opened 2022-04-20
refactor: Rewrite config parser to respect the order

Xuanwo:

## Summary

This PR will fix https://github.com/datafuselabs/databend/issues/4362

To make our config parsed in a more understandable way, I built the following two libs:

- [serde-bridge](https://github.com/Xuanwo/serde-bridge): A bridge between different serde implementations.
- [serfig](https://github.com/Xuanwo/serfig): Layered configuration system built upon serde

Now, our config is parsed in the following ways:

- Load from files (powered by toml and so on)
- Load from env (powered by [serde-env](https://github.com/Xuanwo/serde-env/pull/1))
- Load from clap args

All of them are bridged by `serde-bridge`.

With this PR, we will:

- Avoid args conflict via the new style of args naming style
- Remove `env-helper!` as we are loading by [envy](https://github.com/softprops/envy)

Our config will be more **consistent** and **extensible**.

### args naming

- From `access-key-id` to `storage-s3-access-key-id`
- From `region` to `storage-s3-region`
- Remove `load_from_args`, `load_from_env`, `load_from_file` from `Config`

All storage service-related args will be prefixed by `storage-<service>-` instead.

### env naming

- From `S3_STORAGE_ACCESS_KEY_ID` to `STORAGE_S3_ACCESS_KEY_ID`
- Every `env` will map to an arg directly.

## Preview

image

## Current Status

PR is ready for review now!

Works are listed here

- [ ] Make sure all test cases passed
- [x] Merge with main to adapt to new changes
- [x] Make a release for `serde-bridge`, `serfig` and `serde-env`.
- [x] Update all docs related to this change.

## Changelog

- Improvement

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4362


ZhiHanZ:
I prefer load order like:
1. from argument
2. from configfile
3. from environment variable

to treat environment variable as first-class citizen and respect the configuration from configfile, and it would also cause painful breaking change if config from configfile could overlap settings from environment variables

Xuanwo:
> I prefer load order like:
> 
> 1. from argument
> 2. from configfile
> 3. from environment variable
> 
> to treat environment variable as first-class citizen and respect the configuration from configfile, and it would also cause painful breaking change if config from configfile could overlap settings from environment variables

Yep, this proposal does implement this order, next layer will overwrite the same value for the current layer. That means we will take env first, and `overwrite` by `config` and `argument`.

query/Cargo.toml
```diff
@@ -106,6 +106,8 @@ tonic = "=0.6.2"
 typetag = "0.1.8"
 uuid = { version = "0.8.2", features = ["serde", "v4"] }
 walkdir = "2.3.2"
+serfig = {path="../../../xuanwo/serfig"}
+serde-bridge = {path="../../../xuanwo/serde-bridge"}
```
drmingdrmer, line 110: Can not find the file 🤔 ?

Xuanwo: Yep, this PR is a demo to show my idea and is not workable.

I expect to make it works tomorrow :rocket:

Xuanwo: The PR is buildable now!

Xuanwo:
envy is not suitable for databend as it can't deserialize value in nested structs. I build a new lib [serde-env](https://github.com/Xuanwo/serde-env) to fix it.

Now, our PR is buildable. I'm going to move to the next step: make PR ready for review :rocket:

Xuanwo:
image

Oh, I love deleting code (

Xuanwo:
I added a test for override: https://github.com/datafuselabs/databend/pull/4971/files#diff-a8af230cc9f832a19df365f4ddd5cf49cc3bd752705c27445008c928d7f2041dR193-R295

@ZhiHanZ Please help check whether this behavior is expected.

docs/doc/10-deploy/11-cluster-k8s-minio.md
```diff
@@ -129,4 +129,6 @@ SELECT count(*) FROM t1;
 +----------+
 | 10000000 |
 +----------+
-```
\ No newline at end of file
+1 row in set (0.02 sec)
```
BohuTANG, line 132: These two lines should removed:
```
1 row in set (0.02 sec)
Read 1 rows, 1 B in 0.001 sec., 749.34 rows/sec., 749.34 B/sec.
```

flaneur2020:
I guess we can take a versioned config approach like: https://github.com/k3d-io/k3d/tree/main/pkg/config

on introducing incompatibility config changes, we can bounce up the config version with a migration logic, like https://github.com/flaneur2020/k3d/blob/412950bd84bcee0050c42515ef19356689152328/pkg/config/v1alpha3/migrations.go#L41

thus the older configs can keep compatible until the annual deprecation cycle, allowing users to migrate the configs after reading our deprecation news but before the actual api got break.

Xuanwo:
> I guess we can consider a versioned config approach like: https://github.com/k3d-io/k3d/tree/main/pkg/config

I will add a version id first in this PR.

Xuanwo:
`test_stateless_standalone_macos` seems not very stable.

Xuanwo:
ping @BohuTANG @flaneur2020 for another approve.

BohuTANG (approved):
lgtm

PR #4982, opened 2022-04-21
improvement: work with the new date/datetime type

### LINKED ISSUE #4968: Feature: Make parser suppport datatype with options like `DateTime(3)`/`DateTime(6)...`
**Summary**

After #4921 finished, we need to make `datetype` with options work, like `DateTime(3)/DateTime(6)` (`3` means precision, millisecond, `6` means microseconds).

For now, parser do not support datatype with options.

`DateTime(n)`    n between [ 0, 9]

cc @sundy-li @leiysky

Veeupup:

## Summary

* bump `sqlparser-rs` to the latest `49f4ec2`
* work with `datetime` work with precision

`TIMESTAMP` equals `DATETIME(0)`, with  second precision. If we want more precision, we can use `DATETIME(n)`, n ranges from 0~9

## Changelog

- Improvement

## Related Issues

Fixes #4968


tests/suites/0_stateless/02_function/02_0001_function_type_of.sql
```diff
@@ -31,5 +31,5 @@ select 'UInt64 OP UInt64', typeof(65536 * 65536 + 65536 * 65536), typeof(65536 *
 
 select '=== TEST_datetimes';
 
-select typeof(now()) = 'DateTime_0';
+select typeof(now()) = 'DateTime(0)';
```
sundy-li, line 34: if it's zero, better to display `DateTime` ?

Veeupup: sounds reasonable

common/datavalues/src/types/type_datetime.rs
```diff
@@ -36,10 +36,10 @@ pub struct DateTimeType {
     /// tz indicates the timezone, if it's None, it's UTC.
     tz: Option<String>,
 }
-const DATETIME_0: &str = "DateTime_0";
-const DATETIME_3: &str = "DateTime_3";
-const DATETIME_6: &str = "DateTime_6";
-const DATETIME_9: &str = "DateTime_9";
+const DATETIME_0: &str = "DateTime(0)";
```
sundy-li, line 39: Seems we can remove these variables

Veeupup: yes, should be removed,, will be fixed

b41sh:
almost ok for me, please fix the test.

tests/suites/0_stateless/02_function/02_0001_function_type_of.sql
Veeupup: resolved

common/datavalues/src/types/type_datetime.rs
Veeupup: fixed

sundy-li (approved):

BohuTANG (approved):
:100:

BohuTANG:
@sundy-li @Veeupup 

What's the databend final datetime data type is？
In some database systems(snow, [bigquery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type), duckdb, [cockroachdb](https://www.cockroachlabs.com/docs/stable/timestamp.html)), they are introduce only one type: timestamp(or datetime), and the range:
```
0001-01-01 00:00:00 to 9999-12-31 23:59:59.999999 UTC
```
Supports precision levels from 0 (seconds) to 6 (microseconds) .

I suggest the default is TIMESTAMP(6), what do you think?

PR #4985, opened 2022-04-21
feat(meta-types): make meta type versioned and backward compatible with all the old

BohuTANG:

## Summary

* Try to make meta type versioned, and change-free for all the meta type

## Changelog

- Improvement

## Related Issues

Fixes #issue


common/meta/types-v1/src/example.rs
```diff
@@ -0,0 +1,122 @@
...
+    V2(ExampleV2),
+}
+
+// V1(please don't modify this struct).
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct ExampleV1 {
+    pub tenant: String,
+}
+
+// From V1 to string.
```
dantengsky, line 52: This converter is for demonstration only, right?
Seems that, in practice the conversions should  be 
- `ExampleVn ~> Result<Vec<u8>>` 
-  `Vec<u8> ~> Result<ExampleVn>`

sundy-li:
So if we modify meta types (even one field), we should create a separated version and directory?

BohuTANG:
> So if we modify meta types (even one field), we should create a separated version and directory?

See the example: https://github.com/datafuselabs/databend/pull/4985/files#diff-63b6925b40dfd88714de0bbbd90e89e401d2411e2b468d8f75be4834c674ea4fR50

We should add a new Version to do that -- like ExampleV2, each version is a snapshot.

Plan to add the test for:
1. All the version ser<Vx> sample data
2. Deserialize all the old version sample data with the latest Version struct to check  backward compatible

This will make the backward compatible more clearly and strongly.

PR #4987, opened 2022-04-21
store endpoints to metasrv and use balance endpoints grpc connection channel

### LINKED ISSUE #4865: Feature: allow passing a endpoint list to the metasrv client in a HA manner
**Summary**

it may resembles with the etcd-client:

https://github.com/etcd-io/etcd/tree/main/client/v3#get-started

an endpoint list can ensure the high availability of metasrv without any intermediate load balancer or virtual ip. when an endpoint dies, the client may simply retry another living endpoint.

ariesdevil:

## Summary

Store endpoints to metasrv and use a balanced channel to connect metasrv nodes.

## Changelog

- Improvement

## Related Issues

Fixes #4865


query/src/configs/config_meta.rs
```diff
@@ -37,6 +46,10 @@ pub struct MetaConfig {
     #[serde(alias = "meta_address")]
     pub address: String,
 
+    #[clap(long, env = META_ENDPOINTS, default_value = "", help = "MetaStore peers endpoints")]
```
Xuanwo, line 49: - `env = META_ENDPOINTS` is not needed anymore.
- And it perfer to use `default_value_t` instead.
- To make `clap` works correctly, we need `long="meta-endpoints"`

query/src/configs/config_meta.rs
```diff
@@ -37,6 +46,10 @@ pub struct MetaConfig {
...
+    #[serde(alias = "meta_endpoints")]
```
Xuanwo, line 50: `alias` is used for compatibility, new fields don't need to set `alias` here.

query/src/configs/config_meta.rs
```diff
@@ -111,6 +126,7 @@ impl fmt::Debug for MetaConfig {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("MetaConfig")
             .field("address", &self.address)
+            .field("endpoints", &self.endpoints.join(","))
```
Xuanwo, line 129: I think `&self.endpoints` is enough as `Vec<String>` implements `Debug`.

ariesdevil:
@Xuanwo The `endpoints` seem overwritten by the later builder.
```
// Then, load from env.
builder = builder.collect(from_env());

// Finally, load from args.
builder = builder.collect(from_self(arg_conf));
```

Xuanwo:
Some quick ideas: this may be caused by `clap`.

The default value `""` here may be parsed in `vec![""]` instead of `vec![]`. 

- Please try add a `parse(from_str=xxx)` for `endpoints`: xxx is a function `fn(&str) -> Vec<String>` implemented by `strings.Split(",")`.
- Or just remove the `default_value`?

References Docs: https://github.com/clap-rs/clap/blob/v3.1.7/examples/derive_ref/README.md#arg-types

BohuTANG:
If the `endpoints` are not configured in the config file, how is it work?

ariesdevil:
> If the `endpoints` are not configured in the config file, how is it work?

If `endpoints` is empty, `make_conn` will [return an error](https://github.com/datafuselabs/databend/pull/4987/files#diff-757d3d716b4424b8e86d968f8a68c94b6bc9cd938ac9ee3caa9440ef35bb7707R192), users can call `set_endpoints` manualy.

ariesdevil:
I find the bug that stuck `auto_sync_endpoints` finally, working on it.

BohuTANG:
> > If the `endpoints` are not configured in the config file, how is it work?
> 
> If `endpoints` is empty, `make_conn` will [return an error](https://github.com/datafuselabs/databend/pull/4987/files#diff-757d3d716b4424b8e86d968f8a68c94b6bc9cd938ac9ee3caa9440ef35bb7707R192), users can call `set_endpoints` manualy.

But we already have a `meta_address` , if the endpoints are empty(not configured in the config file), using the `meta_address` as a seed looks more make sense?
endpoints feature is an enhancement, not a required.

ariesdevil:
> > > If the `endpoints` are not configured in the config file, how is it work?
> > 
> > 
> > If `endpoints` is empty, `make_conn` will [return an error](https://github.com/datafuselabs/databend/pull/4987/files#diff-757d3d716b4424b8e86d968f8a68c94b6bc9cd938ac9ee3caa9440ef35bb7707R192), users can call `set_endpoints` manualy.
> 
> But we already have a `meta_address` , if the endpoints are empty(not configured in the config file), using the `meta_address` as a seed looks more make sense? endpoints feature is an enhancement, not a required.

IMO, `meta_address` should be removed when `endpoints` feature is landed.
But as you say, for compatible reason, we can keep both, let me add it back.

scripts/ci/deploy/config/databend-query-node-1.toml
```diff
@@ -47,6 +47,7 @@ query_enabled = false
 # To enable embedded meta-store, set meta_address to ""
 meta_embedded_dir = "_meta_embedded_1"
 meta_address = "0.0.0.0:9191"
+endpoints = ["0.0.0.0:9191"]
```
BohuTANG, line 50: Here should remove  `endpoints` to make sure the old config works.

query/src/configs/config_meta.rs
```diff
@@ -22,6 +22,15 @@ use common_meta_grpc::MetaGrpcClientConf;
 use serde::Deserialize;
 use serde::Serialize;
 
+// Meta env.
+pub const META_ADDRESS: &str = "META_ADDRESS";
```
BohuTANG, line 26: Hmm, why add these env back?

BohuTANG: I think not use them anymore since this PR a few days ago:
https://github.com/datafuselabs/databend/pull/4971/files#diff-51b91c219c223ffd4cf157ac0a9f3e44bf2f6ec489dcfbe3682a816d9a337dbe

common/meta/grpc/src/grpc_client.rs
```diff
@@ -94,59 +114,91 @@ impl ItemManager for MetaChannelManager {
...
-        Ok(Self {
+
+        let (tx, rx) = oneshot::channel();
+
+        let addr = conf.meta_service_config.address.to_string();
+        let endpoints = if !conf.meta_service_config.endpoints.is_empty() {
+            conf.meta_service_config.endpoints.clone()
+        } else {
+            vec![addr]
+        };
```
drmingdrmer, line 143: Looks like it's time to remove `address`. Maybe in next PR.

common/meta/grpc/src/grpc_client.rs
```diff
@@ -94,59 +114,91 @@ impl ItemManager for MetaChannelManager {
...
         MetaError,
     > {
-        let channel = self.conn_pool.get(&self.addr).await?;
-        tracing::debug!("connecting to {}, channel: {:?}", &self.addr, channel);
+        let eps = self.endpoints.read().await;
+        if (*eps).is_empty() {
+            return Err(MetaError::InvalidConfig("endpoints is empty".to_string()));
+        }
+        let channel = self.conn_pool.get(&*eps).await?;
+        tracing::info!("connecting with channel: {:?}", channel);
```
drmingdrmer, line 201: Release the lock on `eps` ASAP.

common/meta/grpc/src/grpc_client.rs
BohuTANG: Hmm, if we remove `address` this PR, please let @ZhiHanZ know, the cloud platform will not work if we remove it.
I will add the databend query configuration to databend.rs soon.

lichuang (approved):

BohuTANG:
Conflicting files @ariesdevil

common/meta/grpc/src/grpc_client.rs
BohuTANG: Looks we can use the block to do that:
```
let channel = ...
{
 let eps = self.endpoints.read().await;
        if (*eps).is_empty() {
            return Err(MetaError::InvalidConfig("endpoints is empty".to_string()));
        }
        let channel = self.conn_pool.get(&*eps).await?;
        tracing::info!("connecting with channel: {:?}", channel);
}
```

common/meta/raft-store/src/sled_key_spaces.rs
```diff
@@ -109,6 +109,15 @@ impl SledKeySpace for ClientLastResps {
...
+/// Key-Value types for meta service addrs in sled::Tree:
+pub struct MetaSrvAddrs {}
+impl SledKeySpace for MetaSrvAddrs {
+    const PREFIX: u8 = 11;
+    const NAME: &'static str = "metasrv-addr";
+    type K = String;
+    type V = String;
+}
+
 /// Enum of key-value pair types of all key spaces.
```
drmingdrmer, line 121: Storing the client-grpc address in the existing keyspace `Nodes` would be better.

Next PR!

drmingdrmer (approved):
Looks good:D

common/meta/grpc/src/grpc_client.rs
```diff
@@ -165,6 +218,51 @@ impl MetaGrpcClient {
...
+    }
+
+    async fn auto_sync_endpoints(_client: Arc<MetaGrpcClient>, mut cancel_rx: Receiver<()>) {
+        loop {
+            select! {
+                _ = &mut cancel_rx => {
+                    return;
+                }
+                _ = tokio::time::sleep(Duration::from_secs(10)) => {
+                    // let r = client.sync_endpoints();
```
BohuTANG, line 255: Not used anymore?

ariesdevil: Comment out temporarily, will be used in the next PR

Issue #4936, opened 2022-04-19
Feature: introducting tenant_id in JWT CustomClaims


Junnplus:
**Summary**

Consider introducing an `InjectContext` in JWT `CustomClaims`:
```rs
pub struct InjectContext {
    pub tenant_id: String,
}

pub struct CustomClaims {
    pub inject_context: Option<InjectContext>,
}
```

Now the query node in management mode cannot switch tenant, we can achieve it through `InjectContext`.

Junnplus:
Any ideas? @flaneur2020 @BohuTANG @ZhiHanZ

flaneur2020:
is it feasible to put a plain `tenant_id` field into the `CustomClaims` instead of nested inside an InjectContext? 

(sort of Law of Demeter?  >v<)

Junnplus:
> is it feasible to put a plain tenant_id field into the CustomClaims directly instead of introduce an new struct InjectContext?

Like `EnsureUser` in JWT `CustomClaims`, it is necessary to introduce a separate structure.

flaneur2020:
is it the restriction from jwt library? q.q

Junnplus:
> is it the restriction from jwt library? q.q

We discussed this issue offline, it is determined that the `tenant_id` is directly placed in CustomClaims without introducing a new structure.

PR #4990, opened 2022-04-21
improvement: introduce tenant_id in JWT CustomClaims

Junnplus:

## Summary

Introduce tenant_id in JWT CustomClaims

## Changelog

- Improvement

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4936

### DIFF
--- a/query/src/servers/http/v1/http_query_handlers.rs
+++ b/query/src/servers/http/v1/http_query_handlers.rs
@@ -12,19 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::sync::Arc;
-
 use common_base::ProgressValues;
 use common_datavalues::DataSchemaRef;
 use common_exception::ErrorCode;
-use common_meta_types::UserInfo;
 use common_tracing::tracing;
 use poem::error::Error as PoemError;
 use poem::error::Result as PoemResult;
 use poem::get;
 use poem::http::StatusCode;
 use poem::post;
-use poem::web::Data;
 use poem::web::Json;
 use poem::web::Path;
 use poem::web::Query;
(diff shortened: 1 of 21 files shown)


youngsofun (approved):

query/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -142,12 +138,11 @@ pub(crate) struct CancelParams {
 
 #[poem::handler]
 async fn query_cancel_handler(
-    sessions_extension: Data<&Arc<SessionManager>>,
+    ctx: &HttpQueryContext,
```
flaneur2020, line 141: 6666, nice abstraction!

query/src/sessions/session_ctx.rs
```diff
@@ -77,10 +79,21 @@ impl SessionContext {
         *lock = db
     }
 
-    pub fn get_tenant(&self) -> String {
+    pub fn get_current_tenant(&self) -> String {
+        if self.conf.query.management_mode {
```
flaneur2020, line 83: add a comment about the difference between the current tenant logic of management_mode and normal case?

query/src/sessions/session_settings.rs
```diff
@@ -300,7 +300,7 @@ impl Settings {
     }
 
     fn set_to_global(&self, setting: &mut SettingValue) -> Result<()> {
-        let tenant = self.session_ctx.get_tenant();
```
flaneur2020, line 303: how about remove the `get_tenant()` method of session_ctx struct? 

newcomers may miss the different uses between the two methods 🤔

query/src/users/auth/jwt/authenticator.rs
```diff
@@ -35,18 +35,26 @@ pub struct JwtAuthenticator {
...
 impl CustomClaims {
     pub fn new() -> Self {
-        CustomClaims { ensure_user: None }
+        CustomClaims {
+            tenant_id: None,
+            ensure_user: None,
+        }
+    }
+
+    pub fn with_tenant_id(mut self, tenant_id: &str) -> Self {
```
flaneur2020, line 55: how about move a `tenant_id: String` here? maybe this could reduce a string copy

Junnplus: This is for testing only.

Issue #4874, opened 2022-04-15
Datetime type `attempt to multiply with overflow` from chrono


BohuTANG:
**Summary**

```
create table t1(a datetime);
insert into t1 values('9999-12-31 23:59:59.999999');

ERROR 2013 (HY000): Lost connection to MySQL server during query
```

Server error log:
```
2022-04-15T08:14:41.711788Z ERROR common_tracing::panic_hook: panicked at 'attempt to multiply with overflow', /home/bohu/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.19/src/naive/datetime.rs:351:21 backtrace=   0: common_tracing::panic_hook::set_panic_hook::{{closure}}
             at common/tracing/src/panic_hook.rs:25:25
   1: std::panicking::rust_panic_with_hook
             at /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/std/src/panicking.rs:702:17
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/std/src/panicking.rs:586:13
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/std/src/sys_common/backtrace.rs:138:18
   4: rust_begin_unwind
             at /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/std/src/panicking.rs:584:5
   5: core::panicking::panic_fmt
             at /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/core/src/panicking.rs:143:14
   6: core::panicking::panic
             at /rustc/76d770ac21d9521db6a92a48c7b3d5b2cc535941/library/core/src/panicking.rs:48:5
   7: chrono::naive::datetime::NaiveDateTime::timestamp_nanos
             at /home/bohu/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.19/src/naive/datetime.rs:351:21
   8: chrono::datetime::DateTime<Tz>::timestamp_nanos
             at /home/bohu/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.19/src/datetime.rs:164:9
   9: <common_datavalues::types::deserializations::date_time::DateTimeDeserializer<T> as common_datavalues::types::deserializations::TypeDeserializer>::de_text_quoted
[...]

PR #4997, opened 2022-04-22
feat(datatype): datatype timestamp with precision

### LINKED ISSUE #4994: What will the datetime data type
@sundy-li @Veeupup 

What's the databend final datetime data type is？
In some database systems(snow, [bigquery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type), [duckdb](https://duckdb.org/docs/sql/data_types/timestamp), [cockroachdb](https://www.cockroachlabs.com/docs/stable/timestamp.html)), they are introduce only one type: timestamp(datetime may be a aliases name in some of them), and the range:
```
0001-01-01 00:00:00 to 9999-12-31 23:59:59.999999 UTC
```
Supports precision levels from 0 (seconds) to 6 (microseconds) .

I suggest the default is TIMESTAMP(6), what do you think?

_Originally posted by @BohuTANG in https://github.com/datafuselabs/databend/pull/4982#issuecomment-1105896600_

Veeupup:

## Summary

* alias datatype `timestamp` alias to `datetime` and `timestamp(n)` can have precision like `datetime(n)`
* default `timestamp` precision is `6`.

## Changelog

- New Feature

## Related Issues

Fixes #4994


BohuTANG:
Cool.
It would be great to add some datetime cast or bound check in the file(it's empty now):
https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/11_data_type/11_0001_data_type_date_time.sql

sundy-li (approved):

Veeupup:
> Cool.
> It would be great to add some datetime cast or bound check in the file(it's empty now):
> https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/11_data_type/[11_0001_data_type_date_time.sql
](https://github.com/datafuselabs/databend/issues/4874)
Maybe this can be fixed with this

BohuTANG:
> > Cool.
> > It would be great to add some datetime cast or bound check in the file(it's empty now):
> 
> [https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/11_data_type/[11_0001_data_type_date_time.sql](https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/11_data_type/%5B11_0001_data_type_date_time.sql) ]
> 
> (#4874) Maybe this can be fixed with this issue

Ok, let's merge :D

b41sh:
@Veeupup  The default precision of `datetime` has changed, so the documentation also needs to be modified.
https://github.com/datafuselabs/databend/blob/main/docs/doc/30-reference/10-data-types/20-data-type-time-date-types.md

BohuTANG:
> @Veeupup The default precision of `datetime` has changed, so the documentation also needs to be modified. https://github.com/datafuselabs/databend/blob/main/docs/doc/30-reference/10-data-types/20-data-type-time-date-types.md

I will do that, not only the datatype also some examples.

BohuTANG:
DATETIME should be an alias for TIMESTAMP, now databend TIMESTAMP is an alias for DATETIME.
See:
https://docs.snowflake.com/en/sql-reference/data-types-datetime.html#datetime
https://duckdb.org/docs/sql/data_types/timestamp
https://www.cockroachlabs.com/docs/stable/timestamp.html (CRDB has no DATETIME type)

cc @Veeupup @sundy-li

PR #5003, opened 2022-04-22
minor(planner): add `context.get_cluster().is_empty()` condition to solid the code

xudong963:

## Summary

Currently `cluster select with new processor` is not ready, so add a determination of whether the cluster is empty to solid the code
## Changelog

- others

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/servers/clickhouse/interactive_worker_base.rs
+++ b/query/src/servers/clickhouse/interactive_worker_base.rs
@@ -42,7 +42,6 @@ use tokio_stream::wrappers::ReceiverStream;
 
 use super::writers::from_clickhouse_block;
 use crate::interpreters::InterpreterFactory;
-use crate::interpreters::SelectInterpreterV2;
 use crate::pipelines::new::processors::port::OutputPort;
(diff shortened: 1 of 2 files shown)


xudong963:
Currently, [opensrv](https://github.com/datafuselabs/opensrv) clickhouse doesn't support `set enable_new_processor_framework & enable_planner_v2`, I'll file a ticket to add them.

BohuTANG:
Hmm, actually we add the setting for newplanner with MySQL handler is enough, like the newprocessor did:
All the tests are using mysql client/driver

xudong963:
> Hmm, actually we add the setting for newplanner with MySQL handler is enough, like the newprocessor did: All the tests are using mysql client/driver
> 
> When all is ready, these settigns will be removed then.

Oh, got it. Let the ticket just add a `context.get_cluster().is_empty()`.

BohuTANG:
Please add more info about this PR to the summay :D

BohuTANG (approved):

PR #5013, opened 2022-04-23
feat(ast): make nullable defaults to false in new AST

sundy-li:

## Summary

Summary about this PR

## Changelog

1. Make datatype nullable defaults to false( special case in databend)
2. Support offset expr

- Improvement

## Related Issues

Fixes #issue


common/ast/src/ast/query.rs
```diff
@@ -413,6 +415,12 @@ impl Display for Query {
             }
         }
 
+        // TODO: We should validate if offset exists, limit should be empty or just one element
```
andylokandy, line 418: I prefer to perform such a check in the optimizer.

common/ast/src/ast/query.rs
```diff
@@ -413,6 +415,12 @@ impl Display for Query {
...
+        if let Some(offset) = &self.offset {
+            write!(f, " OFFSET ")?;
+            write!(f, "{}", offset)?;
```
andylokandy, line 421:
```suggestion
            write!(f, " OFFSET {}", offset)?;
```

andylokandy (approved):

PR #5017, opened 2022-04-23
fix(in): fix function in datatypes bug and add err info

### LINKED ISSUE #4995: Feature:  Function In support other datatypes
**Summary**

Description for this feature.

https://github.com/datafuselabs/databend/blob/0768992f646a1ddb834caec1f789420486556ed9/common/functions/src/scalars/conditionals/in_basic.rs

fkuner:

## Summary

fix function in datatypes bug when date or datetime column in nullable and in values contain NULL.

## Changelog

- Bug Fix

## Related Issues

Fixes #4995


common/functions/src/scalars/conditionals/in_basic.rs
```diff
@@ -127,26 +127,23 @@ impl<const NEGATED: bool> Function for InFunction<NEGATED> {
...
+        let data_type = remove_nullable(columns[0].field().data_type());
+        let least_super_dt = if data_type.data_type_id().is_date_or_date_time() {
             match columns[1..]
                 .iter()
                 .map(|column| column.field().data_type().data_type_id())
-                .all(|t| t.is_string() || t.is_date_or_date_time())
+                .all(|t| t.is_string() || t.is_date_or_date_time() || t.is_null())
             {
-                true => columns[0].field().data_type().clone(),
+                true => wrap_nullable(columns[0].field().data_type()),
```
sundy-li, line 137: Why it need to be wrapped nullable?

fkuner: Column datatype may be nullable and the values in (*, *) may be NULL. In this case, the datatype should be wrapped nullable. But `wrap_nullable` is compatible without nullable, so i use wrap_nullable. For performance, it should add the judgement to choose. I will fix it to add nullable judgement.

Xuanwo:
@fkuner Hi, it's better to leave a comment to explain why this PR is closed. It will help the community to understand what happened here.

fkuner:
> 

Ok, thanks!

fkuner:
The function `in` will be refactored, which will contain more simple and understandable implementation. So this PR can be closed.

PR #5027, opened 2022-04-24
feat(planner): implement aggregate operator in new planner framework

### LINKED ISSUE #3749: Implement aggregate operator in new planner framework
Implement aggregate operator in new planner.

The `GROUP BY` clause is used to group rows together with specified group key. In standard SQL, the group key can only be column, but in databend we allow the group key to be scalar expression, e.g. `a + 1`, `a + b`.

With `GROUP BY`, we cannot reference any non-grouped columns in `SELECT` clause directly(`select b from t group by a` is not allowed), since after grouping the non-grouped columns are reduced with aggregate functions and cannot appear in result set.

Aggregate functions are the functions that takes groups as input, and produce a single accumulated value as result. Aggregated functions can only work with `GROUP BY`. If there is a query with aggregate function(e.g. `select count(*) from t`) but no group key is specified, it will be treated as an aggregate query has only one group which consists of all the input rows.

If we want to filter the aggregated result, a `HAVING` clause can be utilized with. You can reference aggregate functions and group keys in `HAVING` clause(which is pretty similar with `WHERE` clause).

Here's an example of aggregate:
```sql
SELECT a, sum(b + c) + 1 FROM t GROUP BY a, b + 1 HAVING b + 1 > 1;
```

`a` is one of the group keys, so it could appears in the result. Although `b` and `c` are not group keys, they can be referenced by an aggregate function `sum`. 

`b + 1` is one of the group keys, so we can use it in `HAVING` clause. But if we replace the `b + 1 > 1` with `b > 0`, then the query won't work.

xudong963:

## Summary

- implement aggregate operator in the new planner framework

## Changelog

- New Feature

## Related Issues

Fixes #3749 https://github.com/datafuselabs/databend/issues/5066


query/src/sql/planner/binder/select.rs
```diff
@@ -88,7 +98,7 @@ impl Binder {
                     scan_fields: None,
                     parts,
                     statistics,
-                    description: "".to_string(),
+                    description: format!("read source from table {}", table),
```
Xuanwo, line 101:
```suggestion
                    description: format!("read source from table {table}"),
```

xudong963:
Currently, the ticket can't produce results due to two known reasons:
1. https://github.com/datafuselabs/databend/issues/5044
2. https://github.com/datafuselabs/databend/discussions/5053

But, it's time to review, ensure me on the right road. And if the ticket is ok, I think we can merge it and make it work in next tickets.

query/src/sql/planner/binder/project.rs
```diff
@@ -79,7 +79,9 @@ impl Binder {
     ) -> Result<BindContext> {
         let mut output_context = BindContext::create();
         output_context.expression = input_context.expression.clone();
-
```
leiysky, line 82: The needed column bindings are added later, this may cause duplication.

query/src/sql/planner/binder/select.rs
```diff
@@ -55,6 +60,10 @@ impl Binder {
         // Output of current `SELECT` statement.
         let mut output_context = self.normalize_select_list(&stmt.select_list, &input_context)?;
 
+        if !stmt.group_by.is_empty() {
```
leiysky, line 63: The `output_context` is `BindContext` generated by current query, which is actually the state after executing `GROUP BY`.

This is a tricky part in SQL. The aggregate functions look to be defined in `SELECT` clause, but they are actually implicitly defined with `GROUP BY` and get referenced in `SELECT` clause. 

To be more specific, an aggregage query:
```sql
select sum(a) from t group by a;
```
Can be represented in a more understandable way:
```sql
SELECT sum(a) FROM t GROUP BY a AGGREGATE INTO sum(a); 
```

I think what you would like to do is:
1. Extract aggregate functions from `output_context`
2. Resolve `GROUP BY` clause with `input_context` and the aggregate information

My suggestion is that you can add another method like `analyze_aggregate(input_context, output_context)`, which will extract aggregate functions from `output_context` and add them to input_context. Then you can `bind_group_by` with the `input_context` with aggregate information.

query/src/sql/exec/mod.rs
```diff
@@ -199,4 +208,46 @@ impl PipelineBuilder {
 
         Ok(output_schema)
     }
+
+    fn build_aggregate(
+        &mut self,
+        aggregate: &AggregatePlan,
+        input_schema: DataSchemaRef,
+    ) -> Result<DataSchemaRef> {
+        let output_schema = input_schema.clone();
```
leiysky, line 217: Does this `output_schema` work? It looks not that reasonable to me.

query/src/sql/planner/binder/project.rs
xudong963: Yes, after applying your following suggestion, these codes are useless.

query/src/sql/planner/binder/select.rs
xudong963: > I think what you would like to do is:

Yes, you are right!

> My suggestion is that you can add another method like `analyze_aggregate(input_context, output_context)`, which will extract aggregate functions from `output_context` and add them to input_context. Then you can `bind_group_by` with the `input_context` with aggregate information.

Nice suggestion~

> This is a tricky part in SQL. The aggregate functions look to be defined in `SELECT` clause, but they are actually implicitly defined with `GROUP BY` and get referenced in `SELECT` clause.

I printed out ast from parser and tried to see where the aggregation function appears, but I didn't find aggregation functions in group by. At which stage does this implicit conversion occur?

query/src/sql/exec/mod.rs
xudong963: It seems to just need to be consistent with the input schema? Isn't it? 🤔

Such as: `select sum(a) from t group by a;` (table t only has a column `a`)
The schemas from bottom to up are the followings:

table_scan
```
    fields: [
        DataField {
            name: "\"a\"_0",
            data_type: Int32,
            nullable: false,
        },
    ],
```

aggregation plan
```
    fields: [
        DataField {
            name: "\"a\"_0",
            data_type: Int32,
            nullable: false,
        },
    ],
```

projection plan
```
    fields: [
        DataField {
            name: "\"a\"_0",
            data_type: Int32,
            nullable: false,
        },
        DataField {
            name: "\"?column?\"_1",
            data_type: Int64,
            nullable: true,
        },
    ],
```

query/src/sql/planner/binder/select.rs
leiysky: > I printed out ast from parser and tried to see where the aggregation function appears, but I didn't find aggregation functions in group by. At which stage does this implicit conversion occur?

Well, it's exactly what you are working on. 🤣

query/src/sql/exec/mod.rs
leiysky: The aggregation functions should be added to `DataSchema` so it can be accessed by projection plan later.

Besides, aggregate operator would be splited into 2 stages: `AggregatePartial` and `AggregateFinal`, you may refer to: https://github.com/datafuselabs/databend/blob/ec0acd34ae8c058dcb59f60c5d596d3b2e48f29a/query/src/pipelines/new/pipeline_builder.rs#L101-L136

query/src/sql/exec/mod.rs
```diff
@@ -228,4 +239,80 @@ impl PipelineBuilder {
...
+            group_expressions.push(expr);
+        }
+        let mut final_exprs = agg_expressions.to_owned();
+        final_exprs.extend_from_slice(group_expressions.as_slice());
+        let final_schema = Arc::new(DataSchema::new(RewriteHelper::exprs_to_fields(
+            final_exprs.as_slice(),
+            &input_schema,
+        )?));
+
+        dbg!(final_schema.clone());
```
BohuTANG, line 275: I think this is only for debug, please tracing::debug! or removed the dbg probe.

query/src/sql/plan_parser.rs
```diff
@@ -76,12 +76,14 @@ impl PlanParser {
 
     pub fn build_query_plan(data: &QueryAnalyzeState) -> Result<PlanNode> {
         let from = Self::build_from_plan(data)?;
```
BohuTANG, line 78: ditto

query/src/sql/exec/mod.rs
xudong963: Yes, will remove

query/src/sql/exec/mod.rs
xudong963: > The aggregation functions should be added to `DataSchema` so it can be accessed by projection plan later.

The reason I didn't add the aggregation function to the data schema here before is because I see that `ProjectPlan` gets the aggregation function from its `Vec<ProjectItem>` when it processes schema:
https://github.com/datafuselabs/databend/blob/43af01f26652f745f8b4dbc256b8d21c0d37eabe/query/src/sql/exec/data_schema_builder.rs#L43

In the case: `select sum(a) from t group by a`, `ProjectPlan`'s output schema is the following though I didn't add aggregate functions to `AggregatePlan`'s output schema. `?column?` is the aggregate expr.
```
    fields: [
        DataField {
            name: "\"a\"_0",
            data_type: Int32,
            nullable: false,
        },
        DataField {
            name: "\"?column?\"_1",
            data_type: Int64,
            nullable: false,
        },
    ],
```

In fact, I think the right output schema for `ProjectPlan` should be the following:
```
    fields: [
        DataField {
            name: "sum(a)",
            data_type: Int64,
            nullable: false,
        },
    ],
```
Now I add aggregate functions to `DataSchema`, and the output schema for each operator is the following:

`PhysicalScan`
```
    fields: [
        DataField {
            name: "\"a\"_0",
            data_type: Int32,
            nullable: false,
        },
    ],
```

`Aggregate`
```
    fields: [
        DataField {
            name: "sum(\"a\"_0)",
            data_type: Int64,
            nullable: false,
        },
        DataField {
            name: "\"a\"_0",
            data_type: Int32,
            nullable: false,
        },
    ],
```

`Project`
```
    fields: [
        DataField {
            name: "sum(\"a\"_0)",
            data_type: Int64,
            nullable: false,
        },
        DataField {
            name: "\"a\"_0",
            data_type: Int32,
            nullable: false,
        },
        DataField {
            name: "\"?column?\"_1",
            data_type: Int64,
            nullable: true,
        },
    ],
```

I think `PhysicalScan` and `Aggregate` output schemas are right, but for `Project`, it should be 
```
    fields: [
        DataField {
            name: "sum(a)",
            data_type: Int64,
            nullable: false,
        },
    ],
```

query/src/sql/planner/binder/scalar.rs
```diff
@@ -48,6 +52,71 @@ impl ScalarBinder {
...
+
+                        let scalar_binder = ScalarBinder::new();
+                        let mut scalar_exprs = Vec::with_capacity(args.len());
+                        for arg in args.iter() {
+                            scalar_exprs.push(
+                                scalar_binder
+                                    .bind_expr(arg, bind_context)?
+                                    .as_any()
+                                    .downcast_ref::<Scalar>()
+                                    .unwrap()
```
zhang2014, line 84: remove unwrap.

query/src/sql/planner/binder/scalar.rs
```diff
@@ -48,6 +52,71 @@ impl ScalarBinder {
...
+                        for col_binding in col_bindings.iter() {
+                            fields.push(DataField::new(
+                                col_binding.column_name.as_str(),
+                                col_binding.data_type.clone(),
+                            ))
+                        }
+
+                        let agg_func_ref = AggregateFunctionFactory::instance()
+                            .get(name.clone(), data_values.clone(), fields)
+                            .unwrap();
```
zhang2014, line 106: remove unwrap

query/src/sql/planner/binder/scalar.rs
```diff
@@ -48,6 +52,71 @@ impl ScalarBinder {
...
+                        Ok(Arc::new(Scalar::AggregateFunction {
+                            func_name: name.clone(),
+                            distinct: *distinct,
+                            params: data_values,
+                            args: scalar_exprs,
+                            data_type: agg_func_ref.return_type().unwrap(),
+                            nullable: false,
+                        }))
+                    }
+                    false => todo!(),
```
zhang2014, line 117: maybe Err(ErrorCode::Unimplemented) is better.

query/src/sql/exec/mod.rs
xudong963: Oops, `align_data_schema` eliminates my worries

query/src/sql/exec/expression_builder.rs
```diff
@@ -49,16 +56,35 @@ impl<'a> ExpressionBuilder<'a> {
...
+    pub fn build_aggr_function(
+        &self,
+        op: String,
+        distinct: bool,
+        params: Vec<DataValue>,
+        args: &Vec<Scalar>,
+    ) -> Result<Expression> {
+        let mut arg_exprs = Vec::with_capacity(args.len());
+        for arg in args.iter() {
+            arg_exprs.push(self.build(arg).unwrap());
```
zhang2014, line 81: unwrap

query/src/sql/exec/mod.rs
```diff
@@ -136,6 +141,11 @@ impl PipelineBuilder {
                 let input_schema = self.build_pipeline(&expression.children()[0])?;
                 self.build_filter(filter, input_schema)
             }
+            PlanType::Aggregate => {
+                let aggregate = plan.as_any().downcast_ref::<AggregatePlan>().unwrap();
```
zhang2014, line 145: unwrap

query/src/sql/planner/binder/project.rs
```diff
@@ -166,12 +165,29 @@ impl Binder {
     }
 }
 
-pub fn get_expr_display_string(expr: &Expr) -> String {
+fn create_function_display_name(fun: &str, distinct: &bool, args: &[Expr]) -> Result<String> {
+    let names: Vec<String> = args
+        .iter()
+        .map(get_expr_display_string)
+        .collect::<Result<_>>()?;
+    let distinct_str = match distinct {
```
zhang2014, line 173:
```suggestion
    match distinct {
        true => Ok(format!("{}(distinct {})", fun, names.join(","))),
        false => Ok(format!("{}({})", fun, names.join(","))),
    }
```

query/src/sql/planner/binder/scalar.rs
```diff
@@ -48,6 +52,74 @@ impl ScalarBinder {
...
+                            })
+                        }
+
+                        let scalar_binder = ScalarBinder::new();
+                        let mut scalar_exprs = Vec::with_capacity(args.len());
+                        for arg in args.iter() {
+                            scalar_exprs.push(
+                                scalar_binder
+                                    .bind_expr(arg, bind_context)?
+                                    .as_any()
```
zhang2014, line 80: maybe move `.as_any().downcast_ref.ok_or_else` into `ScalarExprRef` is better?

```
impl ScalarExpr {
    pub fn safe_cast<To>(&self) -> Result<To> {
        self.as_any().downcast_ref::<To>().ok_or_else(|| {...})
    }
}
```

query/src/sql/planner/binder/scalar.rs
```diff
@@ -48,6 +52,74 @@ impl ScalarBinder {
...
+                        Ok(Arc::new(Scalar::AggregateFunction {
+                            func_name: name.clone(),
+                            distinct: *distinct,
+                            params: data_values,
+                            args: scalar_exprs,
+                            data_type: agg_func_ref.return_type()?,
+                            nullable: false,
+                        }))
+                    }
+                    false => Err(ErrorCode::UnImplement(format!(
```
zhang2014, line 118: What's happen with `SELECT sleep(1);`?

query/src/sql/planner/binder/scalar_common.rs
```diff
@@ -0,0 +1,101 @@
...
+pub fn find_aggregate_scalars_from_bind_context(bind_context: &BindContext) -> Result<Vec<Scalar>> {
+    let scalars = bind_context
+        .all_column_bindings()
+        .iter()
+        .filter(|col_binding| col_binding.scalar.is_some())
+        .map(|col_binding| {
+            col_binding
+                .scalar
+                .as_ref()
+                .unwrap()
```
zhang2014, line 94: unwrap.

query/src/sql/planner/binder/select.rs
```diff
@@ -138,4 +184,27 @@ impl Binder {
...
         Ok(())
     }
+
+    pub(super) fn bind_group_by(
+        &mut self,
+        group_by_expr: &[Expr],
+        input_context: &mut BindContext,
+    ) -> Result<()> {
+        let scalar_binder = ScalarBinder::new();
+        let group_expr = group_by_expr
```
zhang2014, line 194:
```suggestion
        let group_expr = group_by_expr
        .iter()
            .map(|expr| scalar_binder.bind_expr(expr, input_context))
            .collect::<Result<Vec<ScalarExprRef>>>()?;
```

query/src/sql/planner/binder/select.rs
```diff
@@ -55,11 +60,52 @@ impl Binder {
         // Output of current `SELECT` statement.
         let mut output_context = self.normalize_select_list(&stmt.select_list, &input_context)?;
 
+        self.analyze_aggregate(&output_context, &mut input_context)?;
+
+        if !stmt.group_by.is_empty() {
```
zhang2014, line 65: What's happen with `SELECT COUNT() FROM table`.

In this case, we don't seem to create an AggregatePlan

query/src/sql/planner/binder/select.rs
```diff
@@ -55,11 +60,52 @@ impl Binder {
...
+                    nullable,
+                } => agg_expr.push(Arc::new(Scalar::AggregateFunction {
+                    func_name,
+                    distinct,
+                    params,
+                    args,
+                    data_type,
+                    nullable,
+                })),
+                _ => {
```
zhang2014, line 98: What's happen with `SELECT SUM(column_a + 1) FROM table`.

xudong963:
Wow, it works in my local ! 
```sql
Read 0 rows, 0 B in 0.002 sec., 0 rows/sec., 0 B/sec.

mysql> select sum(a)  from t group by a;
+----------+
| sum(a_0) |
+----------+
|        1 |
|        2 |
+----------+
2 rows in set (0.04 sec)
Read 2 rows, 8 B in 0.008 sec., 262.05 rows/sec., 1.05 KB/sec.
```

query/src/sql/planner/binder/select.rs
xudong963: `Literal` isn't supported currently ~

query/src/sql/planner/binder/scalar.rs
xudong963: Nice idea!

query/src/sql/planner/binder/select.rs
xudong963: Yes, in the ticket I don't consider the case.

query/src/sql/planner/binder/scalar.rs
xudong963: The case will fail in the parser phase:
```
  --> SQL:1:16
  |
1 | SELECT sleep(1)
  | ------         ^ expected token <FROM>
  | |
  | while parsing `SELECT ...`

```
cc @andylokandy

query/src/sql/planner/binder/select.rs
zhang2014: Or `SELECT SUM(column_a + column_b) FROM table`?

query/src/sql/exec/mod.rs
sundy-li: It's addressed in #5077

query/src/sql/planner/binder/select.rs
xudong963: Currently, binary operator only supports `Eq`.  Will add more scalar exprs support in the next PRs.

BohuTANG:
Some conflicts that must be resolved :D

xudong963:
> Some conflicts that must be resolved :D

Ok, now start to solve  conflicts and use `enum_dispatch` 😆

zhang2014 (approved):

sundy-li (approved):

Issue #4918, opened 2022-04-18
Transaction API. High prio


drmingdrmer:
metasrv provides a small language with which a complete tx can be expressed as this demo illustrated(embedded tx support solution and the tx-language solution).

- [ ] Define a struct(maybe `Statements` to describe transaction). This can be considered a small script language.
- [ ] Define a runtime that interprets and executes the transaction defined with `Statements`.
- [ ] Define format of transaction execution result.

Workflow:
A caller(metasrv client) creates a `Statements`, and sends it to metasrv.
metasrv replicates it through raft protocol, commits it, and executes it. Then respond the result to the client(metasrv client).

```rust
// embedded transaction
fn create_table() {
    let xx = dbs.get((args.db_name, args.table_name));
    if xx.is_ok() {
        return Ok();
    } else {
        let id = new_id();
        dbs.insert((args.db_name, args.table_name), id);
        tables.insert((args.db_name, id), args.table);
        return Ok();
    }
}
```

```rust
// user defined transaction

/// List of commands available for user.
pub enum Cmd {

    /// Conditional evaluate: 
    /// - first evaluate `condition`, it has to be a bool value.
    /// - then evaluate then_cmd or else_cmd
    Cmd::IfThenElse {
        condition: Cmd,
        then_cmd: Cmd,
        else_cmd: Cmd,
    }

    /// Assign a Cmd and save the output into a temporary variable with name `var`.
    ///
    /// 
    Cmd::Assign {
        var: String, 
        cmd: Cmd
    }

    /// A series of Cmd that will be evaluated one by one.
    Cmd::Statements(Vec<Cmd>),

    /// Return true if a variable is `None`.
    Cmd::IsNull(String), 

    /// Return the value of a variable, the value type is `AppliedState`.
    Cmd::Var(String), 


    /// Generate a new unique id in a specified name space.
    Cmd::NewId(String), 

    // Tree operations

    /// alias of BTreeMap.get()
    ///
    Cmd::GetKV{
        /// The tree to acces.
        /// Only predefined trees are provided, such as:
        /// table, db, 
        tree: String, 
        key: impl IntoTreeKey, 
    }
}

// Demo of user defined transaction which does the same as the embedded transaction above.
Statements(vec![

    Assign {
        var: "curr",
        cmd: GetKV{
            tree: "db_tbl_id",
            key: (args.db_name, args.table_name)
        }
    },
    IfThenElse {
        condition: IsNull("curr"),
[...]

PR #5030, opened 2022-04-24
Feature: transaction api

lichuang:

## Summary

Feature: transaction api

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4918


common/meta/types/proto/meta.proto
```diff
@@ -16,10 +16,12 @@ syntax = "proto3";
...
 package meta;
 
+import "request.proto";
+
 message Empty {}
-message GetRequest { string key = 1; }
+message RaftGetRequest { string key = 1; }
 
-message GetReply {
+message RaftGetReply {
```
flaneur2020, line 24: could this rename break the backwards compatibility of grpc api? q.q

common/meta/types/proto/meta.proto
```diff
@@ -87,10 +83,72 @@ message WatchResponse {
...
+message TxnOpResponse {
+  oneof response {
+    GetResponse get = 1;
+    PutResponse put = 2;
+    DeleteResponse delete = 3;
+  }
+}
+
+message TxnRequest {
+  TxnCondition cond = 1;
```
flaneur2020, line 134: add a `repeated` for conditions?

i remembers that it's allowed to pass multiple conditions in an atomic manner in the etcd api

common/meta/types/proto/request.proto
```diff
@@ -0,0 +1,63 @@
...
+  TxnKey key = 1;
+
+  optional SeqV value = 2;
+}
+
+// Put request and response
+message PutRequest {
+  TxnKey key = 1;
+  bytes value = 2;
+  bool prev_kv = 3;
```
drmingdrmer, line 45: What does this `bool` mean?

common/meta/types/proto/request.proto
```diff
@@ -0,0 +1,63 @@
...
+
+message PutResponse {
+  TxnKey key = 1;
+  optional SeqV prev_value = 2;
+}
+
+// Delete request and response
+message DeleteRequest {
+  TxnKey key = 1;
+  bool prev_kv = 2;
```
drmingdrmer, line 56: And this ...

common/meta/types/proto/request.proto
```diff
@@ -0,0 +1,63 @@
...
+}
+
+message DeleteResponse {
+  TxnKey key = 1;
+  bool success = 2;
+  optional SeqV prev_value = 3;
+}
```
drmingdrmer, line 63: Will there be an unsuccess delete? If the `txn` conditions are not satisfied, the txn should not run at all. Otherwise, delete should always succeed.

common/meta/types/src/seq_value.rs
```diff
@@ -92,3 +93,19 @@ impl<T> SeqV<T> {
...
     }
 }
+
+impl<T> Ord for SeqV<T>
+where T: Ord
+{
+    fn cmp(&self, other: &SeqV<T>) -> Ordering {
+        self.data.cmp(&other.data)
+    }
+}
```
drmingdrmer, line 103: Why does SeqV need to compare by data? The data could be an arbitrary byte sequence.

drmingdrmer (commented):
Not yet reviewed the execution part. As having some question about the datatype definition.

common/meta/types/proto/meta.proto
BohuTANG: This patch is a breaking change by design.

common/meta/types/src/seq_value.rs
lichuang: > Why does SeqV need to compare by data? The data could be an arbitrary byte sequence.

TxnCondition has some compare by value

common/meta/types/proto/request.proto
lichuang: > What does this `bool` mean?

if or not return prev value

common/meta/types/proto/meta.proto
lichuang: > could this rename break the backwards compatibility of the grpc api? q.q

ok, i will change it back.

common/meta/types/proto/meta.proto
lichuang: ok

common/meta/types/src/transaction.rs
```diff
@@ -0,0 +1,350 @@
...
+        meta: None,
+    })
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq)]
+pub struct TransactionKey {
+    pub key: String,
+    pub tenant: String,
+    pub tree: String,
+}
```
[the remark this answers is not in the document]
lichuang: all the struct in transaction.rs is de/ser by protobuf, as protobuf cannot Copy.

common/meta/types/proto/request.proto
```diff
@@ -0,0 +1,63 @@
...
+message SeqV {
+  uint64 seq = 1;
+  bytes data = 2;
+}
+
+message TxnKey {
+  string key = 1;
+  string tenant = 2;
+  string tree = 3;
+}
```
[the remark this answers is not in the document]
flaneur2020: is it possible that we may introduce some features like tenant isolation inside the metasrv in the future? if so, a metadata about tenent inside each key maybe helpful. 

likewise we have an JWT inside the user session, each JWT is associated with a limitted scope of JWT, we may pass the JWT alongside to the metasrv, and metasrv can restrict the session can only access the data inside one tenant.

common/meta/types/proto/meta.proto
flaneur2020: > This patch is a breaking change by design.

is this PR is mostly adding a new api? if so, i guess making break changes on the existed api may not be really required.

i think we can avoid the accidental breaking changes like fix naming styles if possible, but free to make break changes if it can not make the feature possible without a necessary architecture change 🤔

common/meta/types/proto/request.proto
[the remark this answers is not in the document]
drmingdrmer: >is it possible that we may introduce some features like tenant isolation inside the metasrv in the future? if so, a metadata about tenent inside each key maybe helpful.

It can be done. But if to add this feature is still up to the design principle:
Whether to keep the metasrv a pure kv service like etcd or it has to understand some of the business logic.

Currently `tenant` is embedded into the `key` thus metasrv do not need to understand `tenant`. Are we gonna break this rule?

common/meta/types/src/transaction.rs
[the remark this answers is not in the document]
drmingdrmer: Then why does it need to be `Copy`?

BTW if you do need `Copy`, there is a simpler way to derive `Copy`:
In the `build.rs`, to make `TxnOpResponse` a `Copy`:
```
    tonic_build::configure()
        .type_attribute("TxnOpResponse", "#[derive(Eq)]")
        .compile(...)

```

common/meta/types/src/seq_value.rs
drmingdrmer: It's a little bit misleading, to determine the order of `SeqV` without comparing the field `seq`. 

It's hard for a user to know that `seq_v_1 < seq_v_2` only compares the `data`.
And this `Ord` will break the integrity: `!(a < b) && !(b < a)` does not imply `a == b`, unless the `a == b` is determined by only `SeqV.data`, which is still weird.

common/meta/types/proto/request.proto
drmingdrmer: Maybe always returning the previous kv would be better?
The application does not need to think about whether it needs it or not.
And the response handler in the application does not need to know whether the method that sends the request has set this field.

common/meta/types/proto/request.proto
lichuang: may be user `get` in condition use another key, so i  think notify if or not success if fine:)

common/meta/types/proto/request.proto
lichuang: i just follow etcd api define, how about let transaction API user to judge these param

drmingdrmer: OK. your code your decision.

common/meta/types/proto/meta.proto
```diff
@@ -87,10 +83,75 @@ message WatchResponse {
...
+    // only used for key compare
+    KeyExist = 0;
+    KeyNotExist = 1;
+
+    // only used for value compare
+    ValueEqual = 2;
+    ValueGreater = 3;
+    ValueLess = 4;
+    ValueNotEqual = 5;    
+  }
```
drmingdrmer, line 105: metasrv is a bit different from etcd:
- `KeyExist` can be expressed with `SeqMatch::GE(0)`.
- `KeyNotExist` can be expressed with `SeqMatch::Exact(0)`.

Thus maybe using an expected seq range is more appropriate in our case.

And besides key and value, we need the ability to assert the version(seq) matching an expected version value.

common/meta/types/src/cmd.rs
```diff
@@ -19,8 +19,9 @@ use serde::de;
 use serde::Deserialize;
 use serde::Deserializer;
 use serde::Serialize;
+use LatestVersionCmdMod::Cmd as LatestVersionCmd;
 
-use crate::compatibility::cmd_00000000_20220413;
+use crate::compatibility::cmd_00000000_20220425 as LatestVersionCmdMod;
```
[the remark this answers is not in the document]
lichuang: next time there is a new cmd mod, just change `LatestVersionCmdMod ` and add changed type, otherwise will also change mod name in function.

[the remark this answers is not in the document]
drmingdrmer: I mean, `LatestVersionCmdMod` should be `latest_version_cmd`, since it is a mod but not a struct.

common/meta/types/proto/meta.proto
drmingdrmer: Again, my comment is marked as **resolved** as I can not see any changes to this piece of code, and there is no following comment explaining why it is marked **resolved**.

ariesdevil (approved):

common/meta/types/proto/meta.proto
lichuang: i think change `KeyExist` to a seq compare is not intuition for user, but fine, i refactor the Condition for just Value or Seq compare.

drmingdrmer: It will always be OK to accept a suggestion or not, by updating the code or explaining why it does not need to be solved.

But in this case, the most important thing is the ability to assert the condition of a `seq`, which is not negligible.

lichuang: `seq` is a type of uint64

drmingdrmer: > `seq` is a type of uint64

Right, so?

common/meta/types/proto/meta.proto
```diff
@@ -87,6 +83,71 @@ message WatchResponse {
...
+
+  string key = 1;
+  ConditionTarget target = 2;
+
+  oneof target_union {
+    // used when compare value
+    bytes value = 3;
+    // used when compare seq
+    uint64 seq = 4;
+  }
```
drmingdrmer, line 110: There is already `enum ConditionTarget`. It looks `enum ConditionTarget` can be removed?

common/meta/types/proto/request.proto
```diff
@@ -0,0 +1,59 @@
...
+message TxnDeleteRequest {
+  string key = 1;
+  // if or not return the prev value
+  bool prev_kv = 2;
+}
+
+message TxnDeleteResponse {
+  string key = 1;
+  bool success = 2;
+  optional SeqV prev_value = 3;
```
drmingdrmer, line 58: In the request, the flag is `prev_kv`, which looks like it is expecting a key-value pair. But the response has only the field `prev_value`.

Maybe rename `Request.prev_kv` to `prev_value`?

common/meta/types/proto/request.proto
drmingdrmer: Is there any progress?

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,265 @@ impl StateMachine {
...
+                None => TxnGetResponse {
+                    key: get.key.clone(),
+                    value: None,
+                },
+            },
+            Err(_v) => TxnGetResponse {
+                key: get.key.clone(),
+                value: None,
+            },
+        };
```
drmingdrmer, line 698: Never ignore an error

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,265 @@ impl StateMachine {
...
+                    },
+                }
+            }
+            Err(err) => {
+                tracing::debug!("txn_execute_put_operation fail: {} {:?}", key, err);
+                TxnPutResponse {
+                    key: put.key.clone(),
+                    prev_value: None,
+                }
+            }
```
drmingdrmer, line 742: Just logging an error is not enough. Storage level error has to be dealt with.

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,265 @@ impl StateMachine {
...
+        delete: &TxnDeleteRequest,
+        resp: &mut TxnReply,
+    ) {
+        let key = delete.key.clone();
+        let sub_tree = txn_tree.key_space::<GenericKV>();
+        let key_str = key.to_string();
+
+        let del_resp = if let Ok(result) =
+            self.txn_sub_tree_upsert(&sub_tree, &key_str, &MatchSeq::Any, Operation::Delete, None)
+        {
```
drmingdrmer, line 762: Deal with storage error

common/meta/types/proto/meta.proto
```diff
@@ -87,6 +83,71 @@ message WatchResponse {
...
+    Seq = 1;
+  }
+
+  // condition result
+  enum ConditionResult {
+    Equal = 0;
+    Greater = 1;
+    Less = 2;
+    NotEqual = 3; 
+  }
```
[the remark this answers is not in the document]
lichuang: add `ge` and `le` support.

common/meta/types/proto/meta.proto
lichuang: remove `ConditionTarget` enum.

lichuang: done

common/meta/types/src/cmd.rs
```diff
@@ -14,13 +14,14 @@
 
 use std::fmt;
 
+use latest_version_cmd_mod::Cmd as latest_version_cmd;
```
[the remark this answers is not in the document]
lichuang: done

common/meta/types/src/seq_value.rs
```diff
@@ -17,6 +17,8 @@ use std::convert::TryInto;
 use serde::Deserialize;
 use serde::Serialize;
 
+use crate::protobuf::SeqV as pb_seqv;
```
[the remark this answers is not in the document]
lichuang: done

common/meta/types/src/seq_value.rs
```diff
@@ -92,3 +94,18 @@ impl<T> SeqV<T> {
         self
     }
 }
+
+pub fn convert_seqv_to_pb(seqv: Option<SeqV>) -> Option<pb_seqv> {
+    seqv.as_ref().map(|seqv| pb_seqv {
+        seq: seqv.seq,
+        data: seqv.data.clone(),
+    })
+}
```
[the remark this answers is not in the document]
lichuang: done

common/meta/types/src/seq_value.rs
```diff
@@ -92,3 +94,18 @@ impl<T> SeqV<T> {
...
+
+pub fn convert_to_pb_to_seqv(seqv: Option<pb_seqv>) -> Option<SeqV> {
+    seqv.as_ref().map(|seqv| SeqV {
```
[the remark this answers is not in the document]
lichuang: done

metasrv/src/api/grpc/grpc_service.rs
```diff
@@ -182,6 +184,21 @@ impl MetaService for MetaServiceImpl {
...
+    async fn transaction(
+        &self,
+        request: Request<TxnRequest>,
+    ) -> Result<Response<TxnReply>, Status> {
+        self.check_token(request.metadata())?;
+        common_tracing::extract_remote_span_as_parent(&request);
+
+        let request = request.into_inner();
+
+        tracing::info!("Receive txn_action: {:?}", request);
```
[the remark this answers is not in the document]
lichuang: done

common/meta/raft-store/src/state_machine/sm.rs
lichuang: done

common/meta/raft-store/src/state_machine/sm.rs
lichuang: done

common/meta/raft-store/src/state_machine/sm.rs
lichuang: done

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,265 @@ impl StateMachine {
...
+        txn_tree: &TransactionSledTree,
+        cond: &TxnCondition,
+    ) -> bool {
+        let key = cond.key.clone();
+
+        let sub_tree = txn_tree.key_space::<GenericKV>();
+        let sv = sub_tree.get(&key);
+        if sv.is_err() {
+            return false;
+        }
```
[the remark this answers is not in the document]
lichuang: done

common/meta/types/proto/request.proto
lichuang: done

common/meta/api/src/kv_api_test_suite.rs
```diff
@@ -532,6 +547,347 @@ impl KVApiTestSuite {
...
+
+            assert!(resp.success);
+            let expected: Vec<TxnOpResponse> = vec![TxnOpResponse {
+                response: Some(txn_op_response::Response::Put(TxnPutResponse {
+                    key: txn_key.clone(),
+                    prev_value: convert_seqv_to_pb(Some(SeqV::with_meta(1, None, val1.clone()))),
+                })),
+            }];
+
+            self.check_transaction_responses(&resp.responses, &expected);
```
drmingdrmer, line 615: Need to check `resp.success`

common/meta/api/src/kv_api_test_suite.rs
```diff
@@ -532,6 +547,347 @@ impl KVApiTestSuite {
...
+            let expected: Vec<TxnOpResponse> = vec![
+                // change k1
+                TxnOpResponse {
+                    response: Some(txn_op_response::Response::Put(TxnPutResponse {
+                        key: txn_key1.clone(),
+                        prev_value: convert_seqv_to_pb(Some(SeqV::with_meta(
+                            4,
+                            None,
+                            val1.clone(),
+                        ))),
```
drmingdrmer, line 770: There is a method `new()`, which does not need `meta`:

If you'd like to save several lines of code:D

```rust
impl<T> SeqV<T> {
    pub fn new(seq: u64, data: T) -> Self {
        Self {
            seq,
            meta: None,
            data,
        }
    }

    pub fn with_meta(seq: u64, meta: Option<KVMeta>, data: T) -> Self {
        Self { seq, meta, data }
    }
```

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,241 @@ impl StateMachine {
         Ok(Change::new(prev, result).into())
     }
 
+    fn return_value_condition_result(
+        &self,
+        expected: i32,
+        target_value: &Vec<u8>,
+        value: SeqV,
```
drmingdrmer, line 583: This method does not consume `value`, using a reference would be better: `value: &SeqV`

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,241 @@ impl StateMachine {
...
+            Some(ConditionResult::Le) => value.data <= *target_value,
+            _ => false,
+        }
+    }
+
+    pub fn return_seq_condition_result(
+        &self,
+        expected: i32,
+        target_seq: &u64,
+        value: SeqV,
```
drmingdrmer, line 600: So is this `value: &SeqV`

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,241 @@ impl StateMachine {
...
+            TxnGetResponse {
+                key: get.key.clone(),
+                value: convert_seqv_to_pb(Some(sv)),
+            }
+        } else {
+            TxnGetResponse {
+                key: get.key.clone(),
+                value: None,
+            }
+        };
```
drmingdrmer, line 680:
```suggestion
        let get_resp = TxnGeetResponse { key: get.key.clone(), value: convert_seqv_to_pb(sv)};
```

common/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -559,6 +576,241 @@ impl StateMachine {
...
+                    None
+                },
+            }
+        } else {
+            TxnDeleteResponse {
+                key: delete.key.clone(),
+                success: false,
+                prev_value: None,
+            }
+        };
```
drmingdrmer, line 753: Similar to the above, this block can be reduced to one statement

drmingdrmer (approved):
LGTM

BohuTANG (approved):
:100:

PR #5046, opened 2022-04-25
feat(query): attach short sql in context

sundy-li:

## Summary

Summary about this PR

## Changelog

 
- Improvement
 

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/sql/sql_common.rs
+++ b/query/src/sql/sql_common.rs
@@ -87,8 +87,8 @@ impl SQLCommon {
     }
 
     pub fn short_sql(query: &str) -> String {
-        if query.len() > 64 && query[0..=6].eq_ignore_ascii_case("INSERT") {
-            format!("{}...", query[0..64].to_string())
+        if query.len() >= 64 && query[..=6].eq_ignore_ascii_case("INSERT") {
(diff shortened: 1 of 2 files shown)


query/src/sql/sql_common.rs
```diff
@@ -85,4 +85,12 @@ impl SQLCommon {
             ))),
         }
     }
+
+    pub fn short_sql(query: &str) -> String {
+        if query.len() >= 64 && query[..=6].eq_ignore_ascii_case("INSERT") {
```
zhang2014, line 90: Maybe need trim for `.    INSERT INTO table()`

sundy-li: I don't want to use parser/regexp here, so this it's the simple way to store short insert sqls.

zhang2014: Okay.

zhang2014:
`mysql: Failed to download resource "protobuf_bottle_manifest"` restarted failure test

BohuTANG:
> `mysql: Failed to download resource "protobuf_bottle_manifest"` restarted failure test

Looks not related to this PR :/

Issue #4955, opened 2022-04-20
Feature: the same level of e2e test coverage in http api handler


ZhiHanZ:
**Summary**
During platform integration, we heavily rely on http handler, but test coverage is not enough comparing to mysql case.
Need the following parts:
1. DDL related, CREATE, DROP, ALTER on DATABASE, TABLE etc
2. DSL related on different database or table
3. Security related, especially for mTLS and JWT

- [ ] http api stateless test
- [ ] http api stateful test
- [ ] http api mTLS test
- [ ] http api JWT test

BohuTANG:
Better way is to make the `databend-test` support http client and send the stateless sql and check result:
https://github.com/datafuselabs/databend/blob/077b349822578676854a64e6844d5e58ef008934/tests/databend-test#L637-L641

BohuTANG:
Does this issue related to #4958?

flaneur2020:
hmm, seems duplicated

sundy-li:
> Better way is to make the databend-test support http client and send the stateless sql and check result

It might be incorrect. Http client can't send multi-query at once. But we need multi-query at once to respect the settings inside the SQL file.

ZhiHanZ:
> > Better way is to make the databend-test support http client and send the stateless sql and check result
> 
> It might be incorrect. Http client can't send multi-query at once. But we need multi-query at once to respect the settings inside the SQL file.

I think we could split sql to single query array and aggregate result file?

Xuanwo:
The new sql logic test framework has covered this. Let's close.

PR #5048, opened 2022-04-25
feat(test): new sql logic test framework RFC

ZhiHanZ:

## Summary

RFC of the new sql logic test framework comply with sqllite sql logic test with some extension.
Write down basic PoC in python

## Changelog

- New Feature
- Build/Testing/CI
- Documentation

## Related Issues

Part of #4955
Part of https://github.com/datafuselabs/databend/issues/2647


flaneur2020 (approved):

sundy-li (approved):

sundy-li:
For `http handler` and `mysql handler`.   I think it's not good to show two separate results. (HTTP handler tests should convert the results into it's format)

PsiACE (approved):

ZhiHanZ:
> For `http handler` and `mysql handler`.   I think it's not good to show two separate results. (HTTP handler tests should convert the results into it's format)

agreed， http handler test should be as compatible as possible to MySQL handler test, the label override only apply if the format is hard to convert but correct.
For example, under nested json value case.

PsiACE:
> For `http handler` and `mysql handler`. I think it's not good to show two separate results. (HTTP handler tests should convert the results into it's format)

The current idea is to do an aggregation of tags. This way only separate tests need to be maintained for those whose results do differ.

Issue #4876, opened 2022-04-15
http reset api not log the error


BohuTANG:
**Summary**

1) Start the databend-query
2) Send a  request
```
curl --request POST '127.0.0.1:8000/v1/query/' --header 'Content-Type: application/json' --data-raw '{"sql": "create table t1(a smallint unsigned)"}'
```
Response:
```
{"id":"b9fd8222-cb5e-4b16-bf04-083db76ec814","session_id":null,"schema":null,"data":[],"state":"Failed","error":{"code":1005,"message":"sql parser error: Expected ',' or ')' after column definition, found: unsigned","backtrace":"   0: common_exception::exception_code::<impl common_exception::exception::ErrorCode>::SyntaxException\n             at 
```

The server log:
```
2022-04-15T10:51:14.757363Z  INFO databend_query::servers::http::v1::http_query_handlers: receive http query: HttpQueryRequest { session: New(HttpSessionConf { database: None, max_idle_time: None }), sql: "create table t1(a smallint unsigned)", pagination: PaginationConf { wait_time_secs: 1 } }
```

We should log the error to databend-query log not return it to the client(maybe the client only need the error message not the stacktrace), then it will be analyzed by our monitor system.

youngsofun:
@BohuTANG 

>  client only need the error message not the stacktrace

agree

----
> log the error to databend-query log 

it may overlap with `system.query_log`

currently,  for http handler,  mainly 3 type of errors:

##  bad http request (bad json, auth fail ...),   

http not return 200
there will be a WARN log， should I turn it to ERROR?

```
2022-04-16T02:59:30.619276Z  WARN databend_query::servers::http::middleware: http request error: status=400 Bad Request, msg=parse: key must be a string at line 1 column 2
```

## interperter error

we already have error info (including backtrace) in `system.query_log`

## others( most sql parse error)

currently neither in error log nor  `system.query_log`


# suggest

what about add another type of  `system.query_log` (other than 1.start, 2.end) that  record parser error.
no matter the handle it used.

BohuTANG:
@youngsofun 
* system.query_log has error column but we no used it yet, we can set the error to this field.
* Each http request if the status is not 200 we should log it.

youngsofun:
> ` Each HTTP request if the status is not 200 we should log it.`   

already done ( SQL parse/execute error will still return 200.)

- [ ]  `4.end -- error during the query execution.`
- [ ] `3.start -- error before the query execution`

PR #5057, opened 2022-04-26
feat(query_log): set error fields in system.query_log.

youngsofun:

## Summary

- record error in query_ctx.
- add 3 log_type:  
   - ErrorBeforeStart = 3,   (not implemented) cannot use Interpreter,  will be implemented in next pr
   - ErrorExecute = 4,  two cases: 
       - error in block stream. (implemented, not tested) (need to construct such a test case, like read source error?)
       - return directly by  Interpreter.exeute() , including most DDL (tested)
   - Aborted = 5,   ( tested ) abstract it because in nature, it is not a `error`. and it not really need stack,  better with some `reason`  int text;  
- not covering SelectInterpreterV2 in this pr. 

## Changelog

## Related Issues

part of  https://github.com/datafuselabs/databend/issues/4876


youngsofun:
rebase only to fmt common/streams/Cargo.toml, no other chages

query/src/interpreters/interpreter_query_log.rs
```diff
@@ -31,7 +32,9 @@ use crate::sessions::QueryContext;
 pub enum LogType {
     Start = 1,
     Finish = 2,
-    Error = 3,
+    ErrorBeforeStart = 3,
```
junnplus, line 35: Is there a better name? or add some comment?

youngsofun: ErrorExecute means  and `interpreter.exeute()`
ErrorBeforeStart means error before it:  parse, binding, and any other error.

maybe not need to distinguish the two?  @BohuTANG 
or distinguish by `parse_error` and `other_error` sounds more reasonable to me. because for ddl SQLs, some checks may be done in either/both planer or interpreter?

BohuTANG: ErrorExecute is ok :D

BohuTANG: One error makes sense

youngsofun: done

junnplus (approved):
LGTM

PR #5058, opened 2022-04-26
fix(parser): allow to omit semicolon

andylokandy:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

- make the semicolon on the statements optional

## Changelog

- Bug Fix
- Improvement

## Related Issues

closes https://github.com/datafuselabs/databend/issues/5044

PR #5059, opened 2022-04-26
feat(planner): Support select operator in new planner framework

leiysky:

## Summary

Support `WHERE` clause in new planner.

Should be merged after #5058 

## Changelog

- New Feature

## Related Issues

Related to #3747


query/src/sql/planner/binder/scalar.rs
```diff
@@ -43,7 +45,34 @@ impl ScalarBinder {
...
+    fn bind_binary_op(
+        &self,
+        op: &BinaryOperator,
+        left_child: &Expr,
+        right_child: &Expr,
+        bind_context: &BindContext,
+    ) -> Result<ScalarExprRef> {
+        let left_scalar = self.bind_expr(left_child, bind_context)?;
+        let right_scalar = self.bind_expr(right_child, bind_context)?;
+        match op {
```
xudong963, line 67: Why do we need to match the `op` here, can't we just leave it to `Expression`?
```rust
--- a/query/src/sql/planner/binder/scalar.rs
+++ b/query/src/sql/planner/binder/scalar.rs
@@ -64,16 +64,10 @@ impl ScalarBinder {
     ) -> Result<ScalarExprRef> {
         let left_scalar = self.bind_expr(left_child, bind_context)?;
         let right_scalar = self.bind_expr(right_child, bind_context)?;
-        match op {
-            BinaryOperator::Eq => Ok(Arc::new(Scalar::Equal {
-                left: left_scalar,
-                right: right_scalar,
-            })),
-            _ => Err(ErrorCode::UnImplement(format!(
-                "Unsupported binary operator: {}",
-                op.to_string()
-            ))),
-        }
+        Ok(Arc::new(Scalar::Binary {
+            left: left_scalar,
+            right: right_scalar,
+        }))
     }
 }
```
Then other `op`s will be supported
```sql
Connection id:    8

Query OK, 0 rows affected (0.56 sec)
Read 0 rows, 0 B in 0.537 sec., 0 rows/sec., 0 B/sec.

mysql> select * from t;
+------+
| a    |
+------+
|    1 |
|    2 |
+------+
2 rows in set (0.06 sec)
Read 2 rows, 8 B in 0.017 sec., 115.6 rows/sec., 462.41 B/sec.

mysql> select a from t where a <= a;
+------+
| a    |
+------+
|    1 |
|    2 |
+------+
2 rows in set (0.04 sec)
Read 2 rows, 8 B in 0.005 sec., 378.62 rows/sec., 1.51 KB/sec.
```

andylokandy (approved):

xudong963 (commented):
👍，it works for some cases
```sql
mysql> select * from t;
+------+
| a    |
+------+
|    1 |
|    2 |
+------+
2 rows in set (0.07 sec)
Read 2 rows, 8 B in 0.019 sec., 103.73 rows/sec., 414.93 B/sec.

mysql> select a from t where a = 1;
ERROR 1105 (HY000): Code: 1002, displayText = Unsupported expr: Literal(Number("1")).
mysql> select a from t where a = a;
+------+
| a    |
+------+
|    1 |
|    2 |
+------+
2 rows in set (0.04 sec)
Read 2 rows, 8 B in 0.006 sec., 346.2 rows/sec., 1.38 KB/sec.
```

query/src/sql/planner/binder/scalar.rs
leiysky: Some binary operators have useful property, I'd like to make them typed.

More scalar expressions would be supported later.

xudong963: The useful property will be used in the optimizer?

leiysky: Yes, for instance cardinality estimation.

PR #5069, opened 2022-04-27
Feature: add leave node API

lichuang:

## Summary

add leave node API in metasrv

## Changelog

- New Feature

## Related Issues

Fixes [Add API to remove a node from metasrv cluster · Issue #4312 · datafuselabs/databend](https://github.com/datafuselabs/databend/issues/4312)


metasrv/src/meta_service/meta_leader.rs
```diff
@@ -151,6 +156,40 @@ impl<'a> MetaLeader<'a> {
...
+        membership.remove(&node_id);
+
+        let ent = LogEntry {
+            txid: None,
+            cmd: Cmd::RemoveNode { node_id },
+        };
+
+        self.write(ent.clone()).await?;
+
+        self.change_membership(membership).await
```
drmingdrmer, line 190: If the first entry that removes a node is committed and applied but the second membership entry is not, the the network will find a absent node, which will cause a network error.

metasrv/src/meta_service/meta_leader.rs
```diff
@@ -184,10 +184,11 @@ impl<'a> MetaLeader<'a> {
...
         };
-
-        self.write(ent.clone()).await?;
-
-        self.change_membership(membership).await
+        self.change_membership(membership).await?;
+        match self.write(ent.clone()).await {
+            Ok(_) => Ok(()),
+            Err(err) => Err(err),
+        }
```
drmingdrmer, line 191: Does `ent` need a `clone()`?


And a simple way is :
```
self.write(ent.clone()).await?;
Ok(())
```

drmingdrmer (approved):
Everything else look good to me :DDD

BohuTANG (approved):

PR #5075, opened 2022-04-28
feat(group_by): support two-level hashmap

### LINKED ISSUE #4602: two-level hashmap optimization
Yes, as the number of stored keys increases we did not have the two-level hashmap optimization like ClickHouse.
It can be enhanced in the future.

_Originally posted by @sundy-li in https://github.com/datafuselabs/databend/discussions/4600#discussioncomment-2446491_

fkuner:

## Summary

Two level hashmap

## Changelog

- New Feature
- Improvement
- Performance Improvement

## Related Issues

Fixes #4602


fkuner:
@sundy-li

query/src/common/hashtable/hash_table.rs
```diff
@@ -55,7 +57,7 @@ impl<Key: HashTableKeyable, Entity: HashTableEntity<Key>> Drop for HashTable<Key
 
 impl<Key: HashTableKeyable, Entity: HashTableEntity<Key>> HashTable<Key, Entity> {
     pub fn create() -> HashTable<Key, Entity> {
-        let size = (1 << 8) * mem::size_of::<Entity>();
+        let size = (1 << 12) * mem::size_of::<Entity>();
```
zhang2014, line 60: Why need change from 8 to 12?

query/src/common/hashtable/two_level_hash_table.rs
```diff
@@ -0,0 +1,90 @@
...
+static MAX_BUCKECT: usize = NUM_BUCKETS - 1;
+
+pub struct TwoLevelHashTable<Key: HashTableKeyable, Entity: HashTableEntity<Key>> {
+    // hash_tables: vec![HashTable<Key, Entity>],
+    hash_tables: Vec<HashTable<Key, Entity>>,
+}
+
+impl <Key: HashTableKeyable, Entity: HashTableEntity<Key>> TwoLevelHashTable<Key, Entity> {
+    pub fn create() -> TwoLevelHashTable<Key, Entity> {
+        let mut hash_tables: Vec::<HashTable<Key, Entity>> = Vec::new();
```
zhang2014, line 39: Vec::with_capacity()

query/src/common/hashtable/hash_table.rs
fkuner: It's test code. I change the bit size to observer the proportion of resize. It wll be 8 finally.

query/src/common/hashtable/hash_table.rs
```diff
@@ -119,7 +135,8 @@ impl<Key: HashTableKeyable, Entity: HashTableEntity<Key>> HashTable<Key, Entity>
             let grower = &self.grower;
 
             let mut place_value = grower.place(hash_value);
-            while !self.entities.offset(place_value).is_zero()
+
+            while std::intrinsics::likely(!self.entities.offset(place_value).is_zero())
```
zhang2014, line 139: need remove likely.

fkuner: Yes, I forgot it.

query/src/pipelines/new/processors/transforms/aggregator/aggregator_final.rs
```diff
@@ -128,6 +130,12 @@ impl<Method: HashMethod + PolymorphicKeysHelper<Method> + Send> Aggregator
         let keys_column = block.column(aggregate_function_len);
         let keys_iter = self.method.keys_iter_from_column(keys_column)?;
 
+        let group_by_two_level_threshold =
+            self.ctx.get_settings().get_group_by_two_level_threshold()? as usize;
+        if !self.state.is_two_level() && block.num_rows() >= group_by_two_level_threshold {
```
zhang2014, line 135: maybe we should use `map.size` to replace block.num_rows ?

tests/suites/0_stateless/06_show/06_0003_show_settings.result
```diff
@@ -3,6 +3,7 @@ enable_new_processor_framework	1	1	SESSION	Enable new processor framework if val
 enable_planner_v2	0	0	SESSION	Enable planner v2 by setting this variable to 1, default value: 0	UInt64
 field_delimiter	,	,	SESSION	Format field delimiter, default value: ,	String
 flight_client_timeout	60	60	SESSION	Max duration the flight client request is allowed to take in seconds. By default, it is 60 seconds	UInt64
+group_by_two_level_threshold	10000	10000	SESSION	The threshold of keys to open two-level aggregation, default value: 10000	UInt64
```
sundy-li, line 6: Need another tests  to enable this `group_by_two_level_threshold`.

Example:

```
set group_by_two_level_threshold = 10;

select ....from numbers(...)   group by ...
```

fkuner: Ok!

query/src/pipelines/new/processors/transforms/aggregator/aggregator_final.rs
fkuner: make sense. I will test it.

fkuner: If change to `self.state.len()`, I also need to change `convert_to_two_level` to copy the old map. Because I create a new empty `two_level_hash_table` now. I will change it.

tests/suites/0_stateless/03_dml/03_0003_select_group_by.sql
```diff
@@ -41,3 +41,11 @@ select created_at, sum(count) from t_datetime group by created_at order by creat
 select created_time, sum(count) from t_datetime group by created_time order by created_time;
 
 drop table t_datetime;
+
+--
+SELECT count(*) FROM numbers_mt(10000000) group by number limit 5;
+set group_by_two_level_threshold=10;
+SELECT count(*) FROM numbers_mt(10000000) group by number limit 5;
+set group_by_two_level_threshold=1000000000;
+SELECT count(*) FROM numbers_mt(10000000) group by number limit 5;
```
sundy-li, line 50: add `order by number ` to keep the result stable.

query/src/pipelines/new/processors/transforms/aggregator/aggregator_partial.rs
```diff
@@ -224,6 +227,15 @@ impl<Method: HashMethod + PolymorphicKeysHelper<Method> + Send> Aggregator
         let group_columns = Self::group_columns(&self.params.group_columns_name, &block)?;
         let group_keys = self.method.build_keys(&group_columns, block.num_rows())?;
 
+        let group_by_two_level_threshold =
+            self.ctx.get_settings().get_group_by_two_level_threshold()? as usize;
+        if !self.state.is_two_level()
+            && (block.num_rows() >= group_by_two_level_threshold
```
sundy-li, line 233: we just need `self.state.len() >= group_by_two_level_threshold`

query/src/pipelines/new/processors/transforms/aggregator/aggregator_final.rs
```diff
@@ -128,6 +130,15 @@ impl<Method: HashMethod + PolymorphicKeysHelper<Method> + Send> Aggregator
         let keys_column = block.column(aggregate_function_len);
         let keys_iter = self.method.keys_iter_from_column(keys_column)?;
 
+        let group_by_two_level_threshold =
+            self.ctx.get_settings().get_group_by_two_level_threshold()? as usize;
+        if !self.state.is_two_level()
+            && (block.num_rows() >= group_by_two_level_threshold
```
sundy-li, line 136: just `self.state.len() >= group_by_two_level_threshold`

tests/suites/0_stateless/03_dml/03_0003_select_group_by.sql
fkuner: the result seems stable without order by.

sundy-li: Yes, it happens, but if we run the query in different env(max_thread is different), it may be not stable

BohuTANG:
How about the group by performance gained with this PR?

fkuner:
> How about the group by performance gained with this PR?

It denpends the cardinal number. In low cardinal situlation, single level map is better than two-level hashmap. In higher cardinal situation, the two-level hashmap is better than single-level hashmap because it decreased the cost of resize which gained abount 6%~8%. But When the cardinal number is higher, the single-level hashmap is better again. This problem happens in clickhouse also when I test the performance of single-level and two-level hashmap.

The pr gains low performance because we only done the first step. We need to open parallel merge in the next step. After do this, the performance will increase 2~3 times. I have discussed this implementation with @sundy-li and @zhang2014. Parallel merge needs to expand the pipeline dynamically which is a little bit hard for me. But we will do it soon because it can 
 bring obvious benifits really.

cc @BohuTANG @sundy-li @zhang2014

sundy-li (approved):

PR #5080, opened 2022-04-28
fix(functions): manually drop state in function eval_aggr function

sundy-li:

## Summary

Summary about this PR

## Changelog

- Bug Fix
- 
## Related Issues

Continue  #5056


BohuTANG:
@mergify update

BohuTANG (approved):

common/functions/src/aggregates/aggregator_common.rs
```diff
@@ -89,16 +89,30 @@ pub fn eval_aggr(
...
+        let mut builder = data_type.create_mutable(1024);
+        func.merge_result(addr, builder.as_mut())?;
+
+        Ok(builder.to_column())
+    };
 
-    Ok(builder.to_column())
+    let result = apply();
+
+    if f.need_manual_drop_state() {
```
zhang2014, line 111: need `exit_guard`

sundy-li: `apply` function is used to catch the whole errors.

`exit_guard ` can not catch the `Bump`.

zhang2014: when panic?

sundy-li: I think `exit_guard`  can't catch the panic either.

zhang2014: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a80480b9211961b420a2ce55f0f58cf2

PR #5096, opened 2022-04-28
[common/meta] impl create-db with protobuf and kv-txn API

drmingdrmer:

## Related Issues

### DIFF
--- a/common/meta/grpc/src/meta_client_on_kv.rs
+++ b/common/meta/grpc/src/meta_client_on_kv.rs
@@ -1,0 +1,107 @@
+// Copyright 2021 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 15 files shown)


common/meta/grpc/src/meta_client_on_kv.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Convert to a string as a key used in KVApi
+pub trait ToKVMetaKey: Debug {
+    fn to_key(&self) -> Result<String, FromUtf8Error>;
+}
+
+impl ToKVMetaKey for DatabaseNameIdent {
+    fn to_key(&self) -> Result<String, FromUtf8Error> {
+        let k = format!(
+            "__fd_database/{}/by_name/{}",
```
lichuang, line 75: prefix like "__fd_database" should define as CONST with a name if prefered.

drmingdrmer: It feels a bit weird by using a const.
`impl ToKVMetaKey for DatabaseNameIden` should be the only place using this literal string.

I'd keep this, unless there is a second place using this literal.

BohuTANG (approved):

PR #5100, opened 2022-04-29
[WIP] feat(query): support show cluster information

zhyass:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues

Fixes #issue


BohuTANG:
cluster information is not for the user, so we should use the `system$cluster_information` call, not a SQL statement.
See `system$fuse_hisory(db, table)`.

PR #5108, opened 2022-04-29
Introduce a helper ExpressionEvaluator to simplify expression evaluation

### LINKED ISSUE #5105: Introduce a helper to evaluate scalar expression
So far, we don't have a effective way to evaluate scalar expressions into `ColumnRef`.

The `ExpressionExecutor` is too complicated for normal use.

leiysky:

## Summary

Introduce a helper `ExpressionEvaluator` which evaluates scalar expression in a recursive fashion.

Compare to `ExpressionExecutor`, `ExpressionEvaluator` doesn't rely on `DataSchema`, and won't produce intermidiate result. It's useful when you want to evaluate an expression on the fly.

## Changelog

- New Feature

## Related Issues

Fixes #5105


query/src/common/expression_evaluator.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use crate::sessions::QueryContext;
+
+pub struct ExpressionEvaluator;
+
+impl ExpressionEvaluator {
+    pub fn eval(
+        expression: &Expression,
+        block: &DataBlock,
+        ctx: &QueryContext,
```
BohuTANG, line 36: 'ctx' is generally used as the first parameter of the function

query/src/common/expression_evaluator.rs
```diff
@@ -0,0 +1,172 @@
...
+    ) -> Result<ColumnRef> {
+        let func_ctx = FunctionContext {
```
sundy-li, line 38: Better to add `fn get_function_ctx` inside `ctx`, thus we no longer need ctx.

query/src/common/expression_evaluator.rs
```diff
@@ -0,0 +1,172 @@
...
+            tz: String::from_utf8(ctx.get_settings().get_timezone()?).map_err(|_| {
+                ErrorCode::LogicalError("Timezone has been checked and should be valid.")
+            })?,
+        };
+        match expression {
+            Expression::Column(name) => block.try_column_by_name(name.as_str()).cloned(),
+
+            Expression::Alias(_, _) => Err(ErrorCode::LogicalError(
```
zhang2014, line 46: ErrorCode::Unimplemented is better.

query/src/common/expression_evaluator.rs
leiysky: Fixed.

query/src/common/expression_evaluator.rs
leiysky: Fixed.

query/src/common/expression_evaluator.rs
leiysky: Fixed.

sundy-li (approved):

BohuTANG (approved):

Issue #5106, opened 2022-04-29
bug: `select 1 * 1.0` return wrong result


cadl:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

8061a4725723ab0f6d3d1a9fd1efaeb3281aa5ee

### What's Wrong?

On mysql cli:
```
mysql> SELECT 1 * 1.0;
+---------+
| (1 * 1) |
+---------+
|       0 |
+---------+
1 row in set (0.04 sec)
Read 1 rows, 1 B in 0.005 sec., 217.72 rows/sec., 217.72 B/sec.
```

Query server info log:
```
2022-04-29T10:58:43.491152Z  INFO databend_query::servers::mysql::mysql_interactive_worker: Normal query: SELECT 1 * 1.0
```

```
mysql> SELECT 1 + 1.0;
+---------+
| (1 + 1) |
+---------+
|       1 |
+---------+
```

### How to Reproduce?

run `select 1 * 1.0` on MySQL cli

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

DCjanus:
If we are trying to query `select 1+1.0`, [in those lines](https://github.com/datafuselabs/databend/blob/f202108a9477a06659f17b382d93bf0a92ae8aa1/common/functions/src/scalars/expressions/binary.rs#L53-L54), type of `left` is `PrimitiveColumn<u8>`, but type of `l` is `PrimitiveColumn<f64>`. with [this function](https://github.com/datafuselabs/databend/blob/505e93a286f8ab5b25dd6708ae54194f70e77e4e/common/datavalues/src/columns/series.rs#L38-L41), value of `left` would be `0`

sundy-li:
Here we generate a constant column `1.0`, which name is `1`, so it overrides another column.

https://github.com/datafuselabs/databend/blob/239d393d2e8de7c47af73c30f91c7c84ba5e9c0a/common/datavalues/src/data_value.rs#L379

PR #5111, opened 2022-04-29
tiny improve: add a debug_assert to check unsafe operation

DCjanus:

## Summary

Add a debug_assert to check unsafe operation, which would expose the problem in #5106 

## Changelog

## Related Issues

#5106

### DIFF
--- a/common/datavalues/src/columns/series.rs
+++ b/common/datavalues/src/columns/series.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::any::Any;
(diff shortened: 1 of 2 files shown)


DCjanus:
I was trying to fix #5106, I've found something which cause the BUG, but I'm not familiar with whole codebase, I'm not sure if I can fix it in right way, so before that, I think the thing that I can do is add this check.

sundy-li:
Thanks, two test bugs are detected by this new `assert!`.

Finally, we will remove this `static_cast`, because it's really unsafe!

PR #5077, opened 2022-04-28
feat(plan): add enum-dispatch support to BasePlan to avoid downcast_ref

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR

## Changelog

 
- Improvement
 
## Related Issues

Fixes #issue

PR #5115, opened 2022-04-30
feat(planner): support more aggregate syntax

xudong963:

## Summary

Support the following syntax
- `select sum(a) from t;`
- `select count(a) from t;`
- `select count() from t`;
- `select count(1) from t`;
-  `select count(*) from t`;
- `select count() from t group by a`;
- `select count(1) from t group by a`;

## Changelog

- New Feature

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/sql/planner/binder/scalar.rs
+++ b/query/src/sql/planner/binder/scalar.rs
@@ -26,6 +26,7 @@ use common_exception::Result;
 use common_functions::aggregates::AggregateFunctionFactory;
 
 use crate::sql::planner::binder::BindContext;
+use crate::sql::planner::metadata::optimize_remove_count_args;
 use crate::sql::plans::Scalar;
 
 /// Helper for binding scalar expression with `BindContext`.
@@ -49,6 +50,20 @@ impl ScalarBinder {
                     nullable: column_binding.nullable,
                 }))
             }
+            Expr::Literal(literal) => Ok(match literal {
+                Literal::Number(val) => Arc::new(Scalar::Literal {
+                    data_value: DataValue::try_from_literal(val, None)?,
+                }),
+                Literal::String(val) => Arc::new(Scalar::Literal {
+                    data_value: DataValue::String(val.clone().into_bytes()),
+                }),
+                Literal::Boolean(val) => Arc::new(Scalar::Literal {
+                    data_value: DataValue::Boolean(*val),
+                }),
+                Literal::Null => Arc::new(Scalar::Literal {
+                    data_value: DataValue::Null,
+                }),
+            }),
             Expr::BinaryOp { op, left, right } => {
                 self.bind_binary_op(op, left.as_ref(), right.as_ref(), bind_context)
             }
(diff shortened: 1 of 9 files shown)


query/src/sql/planner/binder/scalar.rs
```diff
@@ -160,6 +132,80 @@ impl ScalarBinder {
...
+                scalar_binder
+                    .bind_expr(arg, bind_context)?
+                    .as_any()
+                    .downcast_ref::<Scalar>()
+                    .ok_or_else(|| ErrorCode::UnImplement("Can't downcast to Scalar"))?
+                    .clone(),
+            );
+        }
+
+        let col_pairs = bind_context.result_columns();
```
leiysky, line 169: Maybe use `BindContext::all_column_bindings()`?

query/src/sql/planner/binder/scalar.rs
```diff
@@ -160,6 +132,80 @@ impl ScalarBinder {
...
+        } else {
+            Scalar::AggregateFunction {
+                func_name: func_name.name.clone(),
+                distinct,
+                params: data_values,
+                args: scalar_exprs,
+                data_type: agg_func_ref.return_type()?,
+                nullable: false,
+            }
+        };
```
leiysky, line 206:
```suggestion
        let agg_scalar = Scalar::AggregateFunction {
            func_name: func_name.name.clone(),
            distinct,
            params: data_values,
            args: if optimize_remove_count_args(func_name.name.as_str(), distinct, args) { vec![] } else { scalar_exprs },
            data_type: agg_func_ref.return_type()?,
            nullable: false,
        };
```

xudong963: Nice, It is surprisingly possible to put `if else` here 😲

query/src/sql/planner/binder/scalar.rs
```diff
@@ -160,6 +132,68 @@ impl ScalarBinder {
...
+        }
+
+        let scalar_binder = ScalarBinder::new();
+        let mut scalar_exprs = Vec::with_capacity(args.len());
+        for arg in args.iter() {
+            scalar_exprs.push(
+                scalar_binder
+                    .bind_expr(arg, bind_context)?
+                    .as_any()
+                    .downcast_ref::<Scalar>()
```
BohuTANG, line 163: We can use `enum-dispatch` instead of `downcast_ref`, see https://github.com/datafuselabs/databend/pull/5077
That's ok to do it in another PR :D

xudong963: I had tried in my local, but a blocking point is encountered due to https://github.com/datafuselabs/databend/blob/218821b4120e511ab9afacf25d0601a8ef90b340/query/src/sql/planner/binder/project.rs#L108

xudong963:
The [commit](https://github.com/datafuselabs/databend/pull/5115/commits/bb277628476c4f6fd85afed730e5b0b22ec8fdad) fixed a bug:
```sql
mysql> select * from t1;
+------+------+
| a    | b    |
+------+------+
|    1 |    2 |
|    2 |    3 |
|    3 |    4 |
+------+------+
3 rows in set (0.05 sec)
Read 3 rows, 24 B in 0.021 sec., 140.25 rows/sec., 1.12 KB/sec.

mysql> SELECT sum(a) FROM t1 group by a;
ERROR 1105 (HY000): Code: 1028, displayText = sum expect to have single arguments, but got 2.
mysql>
```
Now it normally works!

---

Btw, the following is todos for aggregator in new planner:

- [x] count(*), count(col), count(1)
- select sum(a) from t;
- select count(a) from t;
- select count() from t group by a;
- select count() from t;
- select count(1) from t;
- select count(1) from t group by a;
- [x] There is a bug when the table contains multiple columns
- [ ] support having
- [ ] support scalar expression, such as a + 1, b + 1 in group by
- [ ] https://dev.mysql.com/doc/refman/8.0/en/group-by-handling.html,  the aggregation function supports the inclusion of columns that are not in the group by: low pri

I will open new tickets for the rest todos ~

leiysky:
What about moving the aggregation related code to a new file `aggregate.rs`?

xudong963:
> What about moving the aggregation related code to a new file `aggregate.rs`?

done

query/src/sql/planner/binder/scalar.rs
sundy-li: Yes, I tries but failed.

sundy-li (approved):

PR #5119, opened 2022-04-30
improve(function): Decouple `Function` with `DataField`

leiysky:

## Summary

Decouple `Function` with `DataField`.

## Changelog

- Improvement


leiysky:
I've just found that `ColumnRef::data_type` returns physical type of the column. For `Date` or `Timestamp` we still need an extra logical type.

PR #5130, opened 2022-05-02
improve(function): Specialize `NumberFunction` with input type

leiysky:

## Changelog

- Improvement

### DIFF
--- a/common/functions/src/scalars/dates/number_function.rs
+++ b/common/functions/src/scalars/dates/number_function.rs
@@ -39,6 +39,7 @@ use crate::scalars::RoundFunction;
 #[derive(Clone, Debug)]
 pub struct NumberFunction<T, R> {
     display_name: String,
+    input_type: DataTypeImpl,
     t: PhantomData<T>,
     r: PhantomData<R>,
 }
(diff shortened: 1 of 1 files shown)


common/functions/src/scalars/dates/number_function.rs
```diff
@@ -39,6 +39,7 @@ use crate::scalars::RoundFunction;
 #[derive(Clone, Debug)]
 pub struct NumberFunction<T, R> {
     display_name: String,
```
BohuTANG, line 41: Add the input_type is for?

leiysky: https://github.com/datafuselabs/databend/blob/b4c65cdfb47a59ad1d3bc300987527a75c4dcc5b/common/functions/src/scalars/dates/number_function.rs#L361-L369

For now the input type is infered during evaluation, which should be determined when instantiating the function.

sundy-li: Currently, it's useless because we do not need it. But we may need it in the future.

sundy-li (approved):

PR #5131, opened 2022-05-02
feat(planner): support `having` and `scalar expression in group by` for new planner

### LINKED ISSUE #5120: Todos for aggregator in new planner:
The following is todos for aggregator in new planner:

- [x] count(*), count(col), count(1)
- select sum(a) from t;
- select count(a) from t;
- select count() from t group by a;
- select count() from t;
- select count(1) from t;
- select count(1) from t group by a;
- [x] There is a bug when the table contains multiple columns
- [ ] support having
- [ ] support scalar expression, such as a + 1, b + 1 in group by
- [ ] https://dev.mysql.com/doc/refman/8.0/en/group-by-handling.html,  the aggregation function supports the inclusion of columns that are not in the group by: low pri

_Originally posted by @xudong963 in https://github.com/datafuselabs/databend/issues/5115#issuecomment-1114008619_

xudong963:

## Summary

support having and scalar expression in group by for new planner
```sql
Read 0 rows, 0 B in 0.002 sec., 0 rows/sec., 0 B/sec.

mysql> select sum(a) from t group by a having sum(a) = 1;
+------------+
| sum("a"_0) |
+------------+
|          1 |
+------------+
1 row in set (0.05 sec)
Read 3 rows, 12 B in 0.021 sec., 144.59 rows/sec., 578.35 B/sec.

mysql> select sum(a) from t1 group by a + 1, b;
+------------+
| sum("a"_0) |
+------------+
|          1 |
|          3 |
|          2 |
+------------+
3 rows in set (0.09 sec)
Read 3 rows, 24 B in 0.032 sec., 92.74 rows/sec., 741.91 B/sec.
```

## Changelog

- New Feature

## Related Issues

Fixes #5120


xudong963:
The [commit](https://github.com/datafuselabs/databend/pull/5131/commits/e25c955cb5ce884df5f837aa645e7a43ff9c716d) supports scalar expression, such as a + 1, b + 1 in group by : )

query/src/sql/planner/binder/aggregate.rs
```diff
@@ -68,13 +69,24 @@ impl Binder {
...
         let scalar_binder = ScalarBinder::new();
-        let group_expr = group_by_expr
+        let group_exprs = group_by_expr
             .iter()
             .map(|expr| scalar_binder.bind_expr(expr, input_context))
-            .collect::<Result<Vec<ScalarExprRef>>>();
+            .collect::<Result<Vec<ScalarExprRef>>>()?;
+        let mut need_expression_plan = false;
+        for group_expr in group_exprs.iter() {
+            if let Scalar::ColumnRef { .. } = group_expr.safe_cast_to_scalar()? {
```
zhang2014, line 78: maybe if !matches() is better

zhang2014:
Need some test:
``` sql
SELECT sum(a) FROM t GROUP BY a HAVING sum(b) = 1;
-- Does this calculate `a + 1` twice?
SELECT COUNT() FROM t GROUP BY a + 1, b + 2 HAVING a + 1 = 3;
SELECT COUNT() FROM t GROUP BY a + 1, b + 2 HAVING b + 2 = 3;
SELECT COUNT() FROM t GROUP BY a + 1, b + 2 HAVING c = 4; -- {error }
```

BohuTANG:
In some cases, we should integrate the stateless test for the new planner (like this PR). Do we have the plan?

BohuTANG:
> In some cases, we should integrate the stateless test for the new planner (like this PR). Do we have the plan?

Oh, already working on it: https://github.com/datafuselabs/databend/pull/5133

xudong963:
> > In some cases, we should integrate the stateless test for the new planner (like this PR). Do we have the plan?
> 
> Oh, already working on it: #5133

Yes, @leiysky is preparing for integrating the stateless, after he is ready, I'll open a new ticket to make aggregate-related tests pass. cc @zhang2014

BohuTANG:
@mergify update

query/src/sql/exec/data_schema_builder.rs
```diff
@@ -80,6 +81,27 @@ impl<'a> DataSchemaBuilder<'a> {
...
+        exprs: &[Expression],
+        input_schema: DataSchemaRef,
+    ) -> Result<DataSchemaRef> {
+        let mut fields = input_schema.fields().clone();
+        for expr in exprs.iter() {
+            let expr_name = expr.column_name().clone();
+            if input_schema.has_field(expr_name.as_str()) {
+                continue;
+            }
+            let field = if expr.nullable(&input_schema)? {
```
zhang2014, line 95: Maybe better to use match when we care about all branches and the get results from branches.

zhang2014 (approved):

query/src/sql/exec/mod.rs
```diff
@@ -267,7 +283,7 @@ impl PipelineBuilder {
 
         let mut group_expressions = Vec::with_capacity(aggregate.group_expr.len());
         for scalar_expr in aggregate.group_expr.iter() {
-            let scalar = scalar_expr.as_any().downcast_ref::<Scalar>().unwrap();
```
zhang2014, line 270: maybe we need to check aggregate functions for group by exprs?

xudong963: good catch!

query/src/sql/planner/plans/expression.rs
```diff
@@ -0,0 +1,67 @@
...
+use crate::sql::optimizer::RelationalProperty;
+use crate::sql::optimizer::SExpr;
+use crate::sql::plans::BasePlan;
+use crate::sql::plans::LogicalPlan;
+use crate::sql::plans::PhysicalPlan;
+use crate::sql::plans::PlanType;
+use crate::sql::ScalarExprRef;
+
+#[derive(Clone)]
+pub struct ExpressionPlan {
```
leiysky, line 27: This is the same as `Project` operator, you should use that instead.

By the way, I don't think there is requirement for this. Aggregate plan should have produced the aggregate functions in its output, which can be referenced by the following `HAVING` and `SELECT` clauses.

BohuTANG:
There is a panic:
```
}] panic.file="common/functions/src/rdoc/function_doc_asset.rs" panic.line=58 panic.column=65
2022-05-04T02:07:30.669735Z ERROR databend_query::servers::mysql::writers::query_result_writer: OnQuery Error: Code: 1068, displayText = Cannot join handle from context's runtime, cause: panic.

<disabled>
[ FAIL ] - result differs with:
--- /Users/runner/work/databend/databend/tests/suites/0_stateless/06_show/06_0005_show_functions.result	2022-05-04 02:02:59.000000000 +0000
+++ /Users/runner/work/databend/databend/tests/suites/0_stateless/06_show/06_0005_show_functions.stdout	2022-05-04 02:07:30.000000000 +0000
@@ -1,8 +1,2 @@
-today	1	0		Returns current date.
-todayofmonth	1	0		Converts a date or date with time to a UInt8 number containing the number of the day of the month (1-31).
-todayofweek	1	0		Converts a date or date with time to a UInt8 number containing the number of the day of the week (Monday is 1, and Sunday is 7).
-todayofyear	1	0		Converts a date or date with time to a UInt16 number containing the number of the day of the year (1-366).
-today	1	0		Returns current date.
-todayofmonth	1	0		Converts a date or date with time to a UInt8 number containing the number of the day of the month (1-31).
-todayofweek	1	0		Converts a date or date with time to a UInt8 number containing the number of the day of the week (Monday is 1, and Sunday is 7).
-todayofyear	1	0		Converts a date or date with time to a UInt16 number containing the number of the day of the year (1-366).
+ERROR 1105 (HY000) at line 1: Code: 1068, displayText = Cannot join handle from context's runtime, cause: panic.
+ERROR 1105 (HY000) at line 2: Code: 1068, displayText = Cannot join handle from context's runtime, cause: panic.
```

https://github.com/datafuselabs/databend/runs/6283239742?check_suite_focus=true#step:3:7814

sundy-li:
@Mergify update

query/src/sql/planner/plans/expression.rs
BohuTANG: As leiysky suggested, will we do it in this PR or another? @xudong963

xudong963: Sorry for the late reply, I took a day off yesterday.

I use `ExpressionPlan` to process scalar expression in group by, such as group by a+1. Its role is to bridge the schema of scan plan and aggregate plan. I didn't use the direct project operator because it's a bit strange to have two project operators in the whole plan tree, and there is no need to record column index in Expression Plan.

BohuTANG:
Oops, conflicts need to be resolved

xudong963:
I'll close the ticket because basic aggregator can't work now in the main branch.

query/src/sql/planner/plans/expression.rs
leiysky: > Sorry for the late reply, I took a day off yesterday.
> 
> I use `ExpressionPlan` to process scalar expression in group by, such as group by a+1. Its role is to bridge the schema of scan plan and aggregate plan. I didn't use the direct project operator because it's a bit strange to have two project operators in the whole plan tree, and there is no need to record column index in Expression Plan.


I got you.

The point is, every expression can be evaluated on the fly with a context, so we don't need a `ExpressionPlan` to materialize it in the context.

A special case is projection. In projection, an expression can be assigned to a variable through `expr AS "var"`, which makes it can be referenced by other expressions. Canonically, all the variable `"var"` are directly replaced by `expr`. But we don't like that, it's meaningless.

Our solution is, for the expressions(aliased expression in projection or aggregate function produced by `GROUP BY`) that can be explicitly referenced by other expressions, we will store it in the context as a column. In particular, we give the `Scalar` a unique identifer(i.e. `column_index`) and store it in `BindContext` as a `ColumnBinding`.

For the rest expressions, since most of them are not reuseable, we just store them inside the `Plan`s where they are needed.

In your case, there is no need to pre-evaluate some expressions for `GROUP BY`, since they are already involved in the `AggregatePlan`.

After all, there is no `schema` in the planning phase. It's all about logical representation of relational algebra.

If you are confused by `schema`, I'm betting you encountered a execution issue. Try to fix it in `PipelineBuilder`.

PR #5138, opened 2022-05-03
improve(function): Decouple Function with DataField

leiysky:

## Summary

Summary about this PR

## Changelog

- Improvement

## Related Issues

Fixes #issue

### DIFF
--- a/common/functions/src/scalars/comparisons/comparison.rs
+++ b/common/functions/src/scalars/comparisons/comparison.rs
@@ -131,10 +131,12 @@ impl<T: ComparisonImpl> ComparisonFunctionCreator<T> {
 
         let lhs_id = args[0].data_type_id();
         let rhs_id = args[1].data_type_id();
+        let lhs_type = args[0].clone();
+        let rhs_type = args[1].clone();
 
         if args[0].eq(args[1]) {
             return with_match_physical_primitive_type!(lhs_id.to_physical_type(), |$T| {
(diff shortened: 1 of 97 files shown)


sundy-li:
Column's datatype did not cover the logical datatype. 
So we still need the `DataField`.

common/functions/src/scalars/comparisons/comparison.rs
```diff
@@ -131,10 +131,12 @@ impl<T: ComparisonImpl> ComparisonFunctionCreator<T> {
 
         let lhs_id = args[0].data_type_id();
```
sundy-li, line 132: column is physical, so it's data_type is always physical.

For example:
A column with `timestamp` values, it's UInt64Column physically.

If we use `ColumnRef` to replace `ColumnsWithField`, we should keep DataFields inside the function.

leiysky: @sundy-li  Yes, I noticed it. The arguments types of a `Function` should be determined during instantiating, which will reduce many bugs.

But I found most of the functions are not fully specialized, then it's a hard work to decouple `DataField` with `Function`.

In new planner, we will perform strict type checking. Available type cast will be added, and the type errors will be raised during planning. Therefore I'm going to continue the work as soon as new planner is enabled by default.

sundy-li: Ok.

> Available type cast will be added, and the type errors will be raised during planning.

Now we rejected the datatypes inside the function's creator fn (arg type validation is inside creator) and call `return_type` to have the return datatype

leiysky:
Closed due to #5438

PR #5145, opened 2022-05-03
Add cluster key statistics in block meta

zhyass:

## Summary

Add cluster key statistics in block meta

## Changelog

- New Feature

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/storages/fuse/fuse_table.rs
+++ b/query/src/storages/fuse/fuse_table.rs
@@ -152,7 +152,7 @@ impl Table for FuseTable {
         ctx: Arc<QueryContext>,
         stream: SendableDataBlockStream,
     ) -> Result<SendableDataBlockStream> {
-        let log_entry_stream = self.append_trunks(ctx, stream).await?;
+        let log_entry_stream = self.append_chunks(ctx, stream).await?;
         let data_block_stream =
             log_entry_stream.map(|append_log_entry_res| match append_log_entry_res {
                 Ok(log_entry) => DataBlock::try_from(log_entry),
--- a/query/src/storages/fuse/meta/common.rs
+++ b/query/src/storages/fuse/meta/common.rs
@@ -34,6 +34,7 @@ pub struct Statistics {
     pub compressed_byte_size: u64,
 
     pub col_stats: HashMap<ColumnId, ColumnStatistics>,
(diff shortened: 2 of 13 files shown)


query/src/storages/fuse/fuse_table.rs
```diff
@@ -152,7 +152,7 @@ impl Table for FuseTable {
         ctx: Arc<QueryContext>,
         stream: SendableDataBlockStream,
     ) -> Result<SendableDataBlockStream> {
-        let log_entry_stream = self.append_trunks(ctx, stream).await?;
+        let log_entry_stream = self.append_chunks(ctx, stream).await?;
```
dantengsky, line 155: thanks :)

query/src/storages/fuse/meta/common.rs
```diff
@@ -34,6 +34,7 @@ pub struct Statistics {
     pub compressed_byte_size: u64,
 
     pub col_stats: HashMap<ColumnId, ColumnStatistics>,
+    pub cluster_stats: HashMap<ColumnId, ColumnStatistics>,
```
dantengsky, line 37: Just a reminder:

we need a `#[serde(default)]` attr for the new field `cluster_stats` to keep backward compatible in the final PR

query/src/storages/fuse/meta/v1/segment.rs
```diff
@@ -48,6 +48,7 @@ pub struct BlockMeta {
     pub file_size: u64,
     pub col_stats: HashMap<ColumnId, ColumnStatistics>,
     pub col_metas: HashMap<ColumnId, ColumnMeta>,
+    pub cluster_stats: HashMap<ColumnId, ColumnStatistics>,
```
dantengsky, line 51: Just a reminder:

we need a `#[serde(default)]` attr for the new field `cluster_stats` to keep backward compatible in the final PR

query/src/storages/fuse/fuse_table.rs
zhyass: > thanks :)

https://github.com/datafuselabs/databend/pull/5194

dantengsky: Got it

PR #5149, opened 2022-05-04
[metasrv] impl list-db with kv and txn

drmingdrmer:

##### [metasrv] impl list-db with kv and txn
Add several asssertion to list-db test.

KVApiKey provides deserializing key into structure KVApi key

## Related Issues


BohuTANG (approved):
:rocket:

PR #5162, opened 2022-05-04
feat(planner): Refine `Scalar` with enum_dispatch and support more scalar expressions

leiysky:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Changelog

- New Feature

Close #1537 
Close #1538

PR #5165, opened 2022-05-05
fix(planner): make aggregator work and add simple stateless tests

xudong963:

## Summary
Fix degradation due to https://github.com/datafuselabs/databend/pull/5162/ 
- Make aggregation function work properly
- Add simple stateless

## Changelog

- Bug Fix
- Build/Testing/CI

## Related Issues

Fixes #issue


query/src/sql/exec/expression_builder.rs
```diff
@@ -39,7 +39,12 @@ impl<'a> ExpressionBuilder<'a> {
     }
 
     pub fn build_and_rename(&self, scalar: &Scalar, index: IndexType) -> Result<Expression> {
-        let expr = self.build(scalar)?;
+        let mut expr = self.build(scalar)?;
+        if let Expression::AggregateFunction { .. } = expr {
```
BohuTANG, line 43: If we want to deal with some other case in the future, `match` is better?

xudong963: I'm not sure if we'll add more special cases like `AggregateFunction`. Even if it does, https://github.com/datafuselabs/databend/blob/main/common/planners/src/plan_expression_chain.rs#L73 will prompt us to take care of it

leiysky: Will this work with `count(*) + 1`?

xudong963: Hold on a second, it's almost ready 🤣

leiysky (approved):
LGTM

PR #5186, opened 2022-05-06
[metasrv] create/drop/rename table has to update the seq of its belonging db.

drmingdrmer:

##### [metasrv] create/drop/rename table has to update the seq of its belonging db.
By updating the seq, a list-then-drop-all-table operation will be
blocked if new table are created just after listing-tables and before
deleting them.
This way an atomic drop-db-and-its-tables can be done.

- Fix: #5178

## Related Issues


drmingdrmer:
@mergify merge

drmingdrmer:
@Mergifyio requeue

PR #5194, opened 2022-05-06
Add cluster key statistics in block meta

zhyass:

## Summary

Add cluster key statistics in block meta.

## Changelog

- New Feature

## Related Issues

Fixes #issue


common/datavalues/src/data_value.rs
```diff
@@ -258,6 +273,62 @@ impl DataValue {
     }
 }
 
+impl Eq for DataValue {}
+
+impl Ord for DataValue {
```
dantengsky, line 278: Not sure if we can define a meaningful total order relation between `DataValue`s. 
@sundy-li any comments about this? thanks

query/src/storages/fuse/operations/append.rs
```diff
@@ -116,15 +113,21 @@ impl FuseTable {
...
             for expr in &self.order_keys {
                 let cname = expr.column_name();
-                if !merged.iter().any(|x| x.name() == &cname) {
-                    merged.push(expr.to_data_field(&input_schema)?);
-                }
+                let index = merged
+                    .iter()
+                    .position(|x| x.name() == &cname)
+                    .unwrap_or_else(|| {
+                        merged.push(expr.to_data_field(&input_schema).unwrap());
```
dantengsky, line 127: seems that `unwrap` can be avoided here ... (at least if fall back to imperative style)

query/src/storages/fuse/statistics/reducers.rs
```diff
@@ -72,38 +66,39 @@ pub fn reduce_block_stats<T: Borrow<ColumnsStatistics>>(
...
-                max_column.len(),
-            )?;
-
-            if mins.len() > 0 {
-                min = mins.get(0);
-            }
-            if maxs.len() > 0 {
-                max = maxs.get(0);
-            }
+            min_stats
```
dantengsky, line 75: @sundy-li need your help.  I am not sure if replacing `eval_aggr` with "manually comparison" is proper here

sundy-li: I think it's ok.

dantengsky: got it

common/datavalues/src/data_value.rs
dantengsky: @sundy-li, shall we reuse the `impl<S> ChangeIf<S> for CmpMin/Max ` for the comparession of `DataValue`

https://github.com/datafuselabs/databend/blob/a5748357bdbe5e5ee694ac6c1d750d1b0fe78e3e/common/functions/src/aggregates/aggregate_scalar_state.rs#L45-L73

sundy-li: `aggregate_scalar_state` is generic comparision impls without matching each `DataValue`, so it's optimized by performance.


In `cluster key`, we already have min/max indexes, so we don't need to copy this indexes into a Column and apply `eval_aggr`.

BohuTANG:
There is a CI error:
```
---- servers::http::clickhouse_handler::test_insert_format_ndjson stdout ----
thread 'servers::http::clickhouse_handler::test_insert_format_ndjson' panicked at 'Cannot compare different types with non-numeric type', common/datavalues/src/data_value.rs:314:13
```
https://github.com/datafuselabs/databend/runs/6318429474?check_suite_focus=true#step:3:2198

common/datavalues/src/data_value.rs
dantengsky: > `aggregate_scalar_state` is generic comparision impls without matching each `DataValue`, so it's optimized by performance...

got it.

Two other concerns:
1.  Although we can define our own total order for DataValue (even for weird things like NaN...)
     but a PartialOrder may be easier to define (or just use rust's derive mechanism)
2. The min/max logic is also implemented by the aggregate_min/max
     It would be better if these two keep in coherent.     

A suggestion:

- derive `PartialOrd` for DataValue      
- implements the min/max like the `aggregate_scalar_state` does 
   maybe something like this (hope it is doing the same thing as aggregate min max does:-) 
   
   ~~~
           let min = min_stats
                 .iter()
                 .filter(|x| !x.is_null())
                 .min_by(|x, y| min.partial_cmp(&r).unwrap_or(Ordering::Equal))
                 .unwrap_or(DataValue::Null);

            let max = max_stats
                 .iter()
                 .filter(|x| !x.is_null())
                 .min_by(|x, y| max.partial_cmp(&r).unwrap_or(Ordering::Equal))
                 .unwrap_or(DataValue::Null);
~~~

dantengsky: The logics want to follow:

https://github.com/datafuselabs/databend/blob/6623a328b70d793c6f00bb97807ff9df2d4fc1f2/common/functions/src/aggregates/aggregate_scalar_state.rs#L126-L135

https://github.com/datafuselabs/databend/blob/6623a328b70d793c6f00bb97807ff9df2d4fc1f2/common/functions/src/aggregates/aggregate_scalar_state.rs#L67-L71

Issue #5192, opened 2022-05-06
Rename db


drmingdrmer:
Add a method to `MetaApi`: `rename_database`.

- To support #4839

drmingdrmer:
There won't be anything blocking this issue after the following is done:
- #5206

PR #5197, opened 2022-05-06
Feature: impl rename_database with kv-txn

lichuang:

## Summary

Feature: impl rename_database with kv-txn

## Changelog

- New Feature

## Related Issues

Fixes [Issue #5192 ](https://github.com/datafuselabs/databend/issues/5192)

### DIFF
--- a/common/meta/grpc/src/meta_client_on_kv_impl.rs
+++ b/common/meta/grpc/src/meta_client_on_kv_impl.rs
@@ -51,6 +51,8 @@ use common_meta_types::MatchSeqExt;
 use common_meta_types::MetaError;
 use common_meta_types::MetaId;
 use common_meta_types::Operation;
+use common_meta_types::RenameDatabaseReply;
+use common_meta_types::RenameDatabaseReq;
 use common_meta_types::RenameTableReply;
 use common_meta_types::RenameTableReq;
 use common_meta_types::ShareInfo;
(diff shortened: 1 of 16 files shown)


metasrv/src/executor/meta_handlers.rs
```diff
@@ -143,6 +146,35 @@ impl RequestHandler<DropDatabaseReq> for ActionHandler {
     }
 }
 
+#[async_trait::async_trait]
+impl RequestHandler<RenameDatabaseReq> for ActionHandler {
```
BohuTANG, line 150: This seems not a KV API. Is it the old-way meta API?

common/meta/grpc/src/meta_client_on_kv_impl.rs
```diff
@@ -194,6 +196,64 @@ impl MetaApi for MetaClientOnKV {
...
+
+            {
+                let new_tenant_dbname = DatabaseNameIdent {
+                    tenant: tenant_dbname.tenant.clone(),
+                    db_name: req.new_db_name.clone(),
+                };
+                let txn_req = TxnRequest {
+                    condition: vec![
+                        self.txn_cond_seq(tenant_dbname, Eq, db_id_seq)?,
+                        self.txn_cond_seq(&db_id_key, Eq, db_meta_seq)?,
```
drmingdrmer, line 232: Maybe a `rename` does not need to be blocked by `db_meta` changes.

drmingdrmer (commented):
When renaming a db, it has to assert the **dest** db-name does not exist.

BTW, this is not a emergency task for now, it is in the **postponed task** list. Meanwhile the first to do is to replace the old MetaApi impls with the kv-txn based impl. After that, this PR will be simpler to impl too

metasrv/src/executor/meta_handlers.rs
lichuang: > This seems not a KV API. Is it the old-way meta API?

After all meta API has impl by KV api, all the meta handlers in old way will be removed. But now, in order to pass compile, it has to impl with an old way API.

lichuang:
> this is not a emergency task for now, it is in the postponed task list. 

Ok.

Issue #5140, opened 2022-05-03
 Integrate the stateless test for the new planner's aggregation


xudong963:
- #5202 
- #5203

PR #5204, opened 2022-05-07
feat(planner):integrate the stateless test for the new planner's aggregation

xudong963:

## Summary

integrate the stateless test for the new planner's aggregation

## Changelog

- New Feature
- Build/Testing/CI

## Related Issues

part of https://github.com/datafuselabs/databend/issues/5140


query/src/sql/exec/mod.rs
```diff
@@ -320,6 +337,7 @@ impl PipelineBuilder {
                 )
             })?;
 
+        self.pipeline.resize(1)?;
```
xudong963, line 340: Very critical, but previously missed, otherwise the final result is not aggregated

tests/suites/0_stateless/03_dml/03_0001_select_aggregator.sql
```diff
@@ -1,5 +1,7 @@
+set enable_planner_v2 = 1;
```
xudong963, line 1: I'm not sure if I should enable planner v2 directly in this test suit, or copy these tests to https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/20%2B_others/20_0001_planner_v2.sql 🤔

zhang2014: Maybe we should add new test check for planner v2(ignore some errors for merge)?

query/src/sql/exec/data_schema_builder.rs
```diff
@@ -134,4 +135,44 @@ impl<'a> DataSchemaBuilder<'a> {
...
+        &self,
+        input_schema: DataSchemaRef,
+        exprs: &[Expression],
+    ) -> Result<(DataSchemaRef, Vec<Expression>)> {
+        let mut fields = input_schema.fields().clone();
+        let mut agg_inner_expressions = vec![];
+        for arg_expr in exprs.iter() {
+            match arg_expr {
+                Expression::AggregateFunction { args, .. } => {
+                    for arg_inner_expr in args.iter() {
```
zhang2014, line 149: maybe need check args is not aggregate function? for example: `SUM(SUM(1))`;

xudong963: Nice!

tests/suites/0_stateless/03_dml/03_0001_select_aggregator.sql
xudong963: My goal is that each pr can make test cases in a test file or multiple test files pass.

There are two more test files related to aggregation plan, they are blocked by group by and subquery table reference https://github.com/datafuselabs/databend/issues/5140, I am going to use two PRs to solve them. 

So we don't need to ignore errors for the sake of merging?

sundy-li: I think directly is ok, thus we can cover old behavior if we enable this setting.

After the new planner is stable, we can remove this setting.

leiysky: I suggest to make a copy of this test for planner v2. As soon as the planner v2 is enabled by default, it will be removed.

query/src/sql/exec/mod.rs
leiysky: I've added this for you in a previous PR, but it seems missed in some rebase operation. 😢

query/src/sql/exec/data_schema_builder.rs
leiysky: I thought this has been checked in planner? If not then you should also take care of the case `sum(func(sum(a)))`.

xudong963: It'll be checked in expression_chain, `Action must be a non-aggregated function`, but I think it's reasonable to check earlier

leiysky: > It'll be checked in expression_chain, `Action must be a non-aggregated function`, but I think it's reasonable to check earlier and give more meaningful error hint

You are right. Maybe it's better to fix this in another PR.

leiysky (approved):

tests/suites/0_stateless/03_dml/03_0001_select_aggregator.sql
sundy-li: I thought it again, maybe @leiysky  is right, we should keep both tests stable during refactoring.

So let's make a copy or separate tests like `03_0001_select_aggregator_v2.sql`.

When planner v2 is enabled by default, we could merge the tests into `03_0001_select_aggregator.sql`

sundy-li: If we have separate tests, we can remove `set enable_planner_v2 = 0;` at the end of the file

xudong963: Yes, separating tests won't influence old planner development.

xudong963:
@mergify update

PR #5222, opened 2022-05-07
bugfix(planner): Fix wrong result of hash join when join keys have different types

leiysky:

## Summary

The root cause is comparison between expressions with different types and nullability.

## Changelog

- Bug Fix


query/src/sql/planner/binder/join.rs
```diff
@@ -229,7 +231,13 @@ impl<'a> JoinConditionResolver<'a> {
                 },
             );
 
-            // TODO(leiysky): bump types of left conditions and right conditions
+            // Bump types of left conditions and right conditions
+            let left_type = left.data_type();
+            let right_type = right.data_type();
+            let least_super_type = merge_types(&left_type, &right_type)?;
```
leiysky, line 237: @sundy-li Seems `compare_coercion` doesn't work with `Nullable` types, is that by design?

By the way, it looks have the same function with `merge_types` except the nullability check.

sundy-li: Yes， it's by design.  I'll document it that we should use `merge_types` if we should care about nullable.

query/src/sql/planner/binder/scalar_common.rs
```diff
@@ -115,3 +118,12 @@ pub fn split_equivalent_predicate(scalar: &Scalar) -> Option<(Scalar, Scalar)> {
         _ => None,
     }
 }
+
+pub fn wrap_cast(scalar: Scalar, target_type: &DataTypeImpl) -> Scalar {
+    let cast = CastExpr {
```
sundy-li, line 123: Add if `scalar.data_type() != target_type` special case

leiysky: Is `DataType` comparable? 

At the beginning, I was trying to use `DataTypeID` to compare the two types, but found it doesn't work with nested data types.

Maybe I should compare them with type names?

sundy-li: Yes, it's comparable.

```
impl PartialEq for DataTypeImpl {
    fn eq(&self, that: &DataTypeImpl) -> bool {
        equal(self, that)
    }
}
```

query/src/sql/planner/binder/join.rs
leiysky: > Yes， it's by design.  I'll document it that we should use `merge_types` if we should care about nullable.

IMHO, `get_least_super_type` is a reasonable name.

query/src/sql/planner/binder/scalar_common.rs
leiysky: @sundy-li Fixed.

sundy-li (approved):

BohuTANG (approved):

leiysky:
@mergify requeue

Issue #5236, opened 2022-05-08
bug: `select *` can't work due to duplicate name in new planner 


xudong963:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

main

### What's Wrong?

```sql
mysql> select * from numbers(1000) as t inner join numbers(1000) as t1 on t.number = t1.number;
ERROR 1105 (HY000): Code: 1002, displayText = Duplicate alias name :number (while in processor thread 0).
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

leiysky:
@xudong963  The root cause of this issue is `ExpressionExecutor` doesn't allow duplicated names.

I'm going to refactor that after migrated to new planner.

xudong963:
> I'm going to refactor that after migrated to new planner.

Nice!

PR #5242, opened 2022-05-09
bugfix(pipeline): Fix state machine of hash join

### LINKED ISSUE #5234: Bug: `COUNT(*)` produce wrong result with joins in new planner
```sql
MySQL [(none)]> select count(*) from numbers(1000) as t inner join numbers(1000) as t1 on t.number = t1.number;
Query OK, 0 rows affected (0.139 sec)
Read 2000 rows, 15.62 KiB in 0.110 sec., 18.19 thousand rows/sec., 142.11 KiB/sec.
```

The expected result is:
```
+----------+
| count(*) |
+----------+
|     1000 |
+----------+
```

While the query `select t.number from numbers(1000) as t inner join numbers(1000) as t1 on t.number = t1.number;` works well.

leiysky:

## Summary

Fix the bug that hash join probe transform won't emit `Finished` event.

## Changelog

- Bug Fix

## Related Issues

Close #5234


BohuTANG:
LGTM.
Do we have any plan to move some logic to some files from `SQL/exec/mod.rs`? The codes in it are getting more and more trivial :D

leiysky:
> LGTM. Do we have any plan to move some logic to some files from `sql/exec/mod.rs`? The codes in it are getting more and more trivial :D

It's just somewhat glue code. I'm going to do some decouple work on the concepts `PlanNode`, `Expression` and `Transform` after migrating to new planner. After that, we will deal with these files.

sundy-li (approved):

tests/suites/0_stateless/20+_others/20_0001_planner_v2.sql
```diff
@@ -65,4 +65,6 @@ drop table t;
 drop table t1;
 drop table t2;
 
+select count(*) from numbers(1000) as t inner join numbers(1000) as t1 on t.number = t1.number;
```
BohuTANG, line 68: Not sure why this is not working:
```
mysql> select * from numbers(1000) as t inner join numbers(1000) as t1 on t.number = t1.number;
ERROR 1105 (HY000): Code: 1002, displayText = Duplicate alias name :number (while in processor thread 0).
```

BohuTANG: On, this is OK:
```
mysql> select t.number from numbers(1000) as t inner join numbers(1000) as t1 on t.number = t1.number;
```

sundy-li: same problem in #5236

PR #5244, opened 2022-05-09
feat(function): Support generic Array access elements by index

### LINKED ISSUE #5224: Feature: support get item by index to generic Array column
**Summary**

Description for this feature.

```
:) create table a (number Array(UInt8) );

:) insert into a values( [1,2]), ([3,4]);
Read 2 rows, 28.00 B in 0.011 sec., 174.21 rows/sec., 2.38 KiB/sec.

:) select * from a;
+--------+
| number |
+--------+
| [1, 2] |
| [3, 4] |
+--------+
2 rows in set (0.03 sec)
Read 2 rows, 28.00 B in 0.004 sec., 490.61 rows/sec., 6.71 KiB/sec.

:) select number[0], number[1] from a;
ERROR 1105 (HY000): Code: 1007, displayText = Invalid argument types for function 'GET_PATH': (Array, String) (while in select before projection).
:) select number[0], number[1] from a;
ERROR 1105 (HY000): Code: 1007, displayText = Invalid argument types for function 'GET_PATH': (Array, String) (while in select before projection).
:) select get(number,0) from a;
ERROR 1105 (HY000): Code: 1007, displayText = Invalid argument types for function 'GET': (Array, UInt8) (while in select before projection).
:)
```

b41sh:

## Summary

Support accessing elements of generic `Array` by index

## Changelog

- New Feature
- Build/Testing/CI
- Documentation

## Related Issues

Fixes #5224


common/functions/src/scalars/semi_structureds/array_get.rs
```diff
@@ -0,0 +1,154 @@
...
+            return Err(ErrorCode::IllegalDataType(format!(
+                "Invalid argument types for function '{}': ({:?}, {:?})",
+                display_name.to_uppercase(),
+                data_type.data_type_id(),
+                path_type.data_type_id()
+            )));
+        }
+
+        Ok(Box::new(ArrayGetFunctionImpl::<BY_PATH> {
+            data_type: data_type.clone(),
```
sundy-li, line 57: data_type should be saved as `ArrayType`.

common/functions/src/scalars/semi_structureds/array_get.rs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            return Err(ErrorCode::IllegalDataType(format!(
+                "Invalid argument types for function '{}': ({:?}, {:?})",
+                display_name.to_uppercase(),
+                data_type.data_type_id(),
+                path_type.data_type_id()
+            )));
+        }
+
+        let array_type = if let DataTypeImpl::Array(array_type) = data_type {
```
sundy-li, line 56: using `TryFrom`  is better.

common/functions/src/scalars/semi_structureds/array_get.rs
```diff
@@ -0,0 +1,148 @@
...
+                                len, index
+                            )));
+                        } else {
+                            builder.append(inner_column.get_data(offset + index), true);
+                        }
+                        offset += len;
+                    }
+                } else {
+                    return Err(ErrorCode::IllegalDataType(format!(
+                        "Array column only support accessed by index, but got {:#?}",
```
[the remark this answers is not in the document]
b41sh: The purpose of using `path_keys` is to reuse the `parse_path_keys` function. It seems that the access method of the array column and variant column is somewhat different. We can rewrite a function to generate an index path key.

fkuner:
What's the purpose of get_path of array? It seems same as normal get.

b41sh:
> What's the purpose of get_path of array? It seems same as normal get.

[`get_path`](https://docs.snowflake.com/en/sql-reference/functions/get_path.html) is a snowflake semistructured function, we can use it to access element variant by multiple index path, such as '[0][1][2]', while [`get`](https://docs.snowflake.com/en/sql-reference/functions/get.html) can only access by one index. But it's a bit too complicated for access the data of array, I'll simplify it.

b41sh:
I have removed `ArrayGetPathFunction`, array map access will be converted to the recursive get function.
for example: 
```sql
select arr[0][1] from test;
```
will be converted to
```sql
select get(get(arr, 0), 1) from test;
```
PTAL @sundy-li @fkuner

common/functions/src/scalars/semi_structureds/array_get.rs
```diff
@@ -0,0 +1,135 @@
...
+        } else {
+            Series::check_get(columns[0].column())?
+        };
+
+        let inner_type = self.array_type.inner_type().data_type_id();
+        with_match_scalar_types_error!(inner_type.to_physical_type(), |$T| {
+            let inner_column: &<$T as Scalar>::ColumnType = Series::check_get(array_column.values())?;
+            let mut builder = NullableColumnBuilder::<$T>::with_capacity(input_rows);
+
+            for index in indexes.iter() {
```
sundy-li, line 89: I think it's incorrect, what if index's size is not 1 ?

tests/suites/0_stateless/02_function/02_0051_function_semi_structureds_get.sql
```diff
@@ -68,4 +72,10 @@ select id, json_extract_path_text(str, '["a"]') from t3;
 select id, json_extract_path_text(str, 'b.c') from t3;
 select id, json_extract_path_text(str, '["b"]["c"]') from t3;
 
+select '==get from array table==';
+select id, get(arr, 0) from t4;
```
sundy-li, line 76: get(arr, number  % 2 ) from t4;

common/functions/src/scalars/semi_structureds/array_get.rs
b41sh: if the index is from a table column, we will have multi indexes.
for example:
```sql
mysql> create table t (index uint64);
mysql> insert into t values(0),(1),(2);
mysql> select get([1,2,3,4], index) from t;
+--------------------------+
| get([1, 2, 3, 4], index) |
+--------------------------+
|                        1 |
|                        2 |
|                        3 |
+--------------------------+
3 rows in set (0.05 sec)
```

fkuner:
LGTM

common/functions/src/scalars/semi_structureds/array_get.rs
```diff
@@ -0,0 +1,132 @@
...
+                for i in 0..input_rows {
+                    let len = array_column.size_at_index(i);
+                    let _ = check_index(index, len)?;
+                    builder.append(inner_column.get_data(offset + index), true);
+                    offset += len;
+                }
+            } else {
+                let mut offset = 0;
+                for i in 0..input_rows {
+                    let index = columns[1].column().get(i).as_u64()? as usize;
```
sundy-li, line 106: virtual call here.

b41sh: fixed

common/functions/src/scalars/semi_structureds/array_get.rs
```diff
@@ -0,0 +1,143 @@
...
+    array_type: ArrayType,
+    display_name: String,
+}
+
+impl ArrayGetFunction {
+    pub fn try_create(display_name: &str, args: &[&DataTypeImpl]) -> Result<Box<dyn Function>> {
+        let data_type = args[0];
+        let path_type = args[1];
+
+        if !data_type.data_type_id().is_array() || !path_type.data_type_id().is_integer() {
```
fkuner, line 39: What if index is negative?

b41sh: currently, we don't support index by a negative number, a error will return.
```sql
mysql> select get(v, -1) from arr;
ERROR 1105 (HY000): Code: 1010, displayText = Unexpected type:Int64 to get u64 number (while in processor thread 0).
```

fkuner: Sorry, I have not seen `as_u64` method.

common/functions/src/scalars/semi_structureds/array_get.rs
```diff
@@ -0,0 +1,143 @@
...
+                        let len = array_column.size_at_index(i);
+                        let _ = check_index(index, len)?;
+                        builder.append(inner_column.get_data(offset + index), true);
+                        offset += len;
+                    }
+                } else {
+                    let index_column: &PrimitiveColumn<$T2> = Series::check_get(columns[1].column())?;
+                    let mut offset = 0;
+                    for i in 0..input_rows {
+                        let index = index_column.get(i).as_u64()? as usize;
```
sundy-li, line 116: we should not use `get(i)` to construct a datavalue and convert it into `as_u64`


```
for (row, index) in index_column.iter().enumerate() {
      if index < 0 { ...}
      
      ....

}
```

b41sh: ok

BohuTANG:
@mergify update

BohuTANG:
@mergify update

sundy-li (approved):

Issue #5215, opened 2022-05-07
Tracking issues of adopting features added in OpenDAL v0.6


Xuanwo:
OpenDAL v0.6 added the following features

- retry able operations
- hdfs services support

> Read [OpenDAL v0.6 Release](https://github.com/datafuselabs/opendal/discussions/265) for more information

This issue intends to track the adoption of OpenDAL v0.6:

- [x] https://github.com/datafuselabs/databend/pull/5216
- [x] https://github.com/datafuselabs/databend/pull/5245
- [x] https://github.com/datafuselabs/databend/pull/5247

PR #5245, opened 2022-05-09
feat: Add HDFS support

Xuanwo:

## Summary

This PR will allow databend-query to use hdfs as storage backend.

Part of #5215 

## Changelog

- New Feature


Xuanwo:
@dantengsky Maybe I need to borrow some work from your PR to get JAVA setup correctly?

BohuTANG:
From I understand, this PR makes the HDFS a normal storage backend like AWS S3, it's not related to @dantengsky work on hive?

Xuanwo:
> From I understand, this PR makes the HDFS as a normal storage backend like AWS S3, it's not related to @dantengsky work on hive?

Yes. Hive's integration should happen in another PR.

query/src/configs/config_storage.rs
```diff
@@ -161,6 +161,19 @@ impl fmt::Debug for AzblobStorageConfig {
     }
 }
 
+#[derive(Clone, PartialEq, Serialize, Deserialize, Default, Args, Debug)]
+#[serde(default)]
+pub struct HdfsConfig {
+    #[clap(long = "storage-hdfs-name-node", default_value_t)]
+    pub name_node: String,
```
BohuTANG, line 168: What's the name_node URI look like? Is there any spec for it in OpenDAL.

Xuanwo: > What's the name_node URI look like? Is there any spec for it in OpenDAL.

Docs: https://docs.rs/opendal/0.6.1/opendal/services/hdfs/struct.Builder.html

They are the same as how hdfs java clients use.

query/src/configs/config_storage.rs
```diff
@@ -185,6 +198,10 @@ pub struct StorageConfig {
     // azure storage blob config.
     #[clap(flatten)]
     pub azblob: AzblobStorageConfig,
+
+    // hdfs storage backend config
+    #[clap(flatten)]
+    pub hdfs: HdfsConfig,
```
BohuTANG, line 204: If we don't add the `HdfsConfig` item to *.toml, the config deserialize looks will crash?

BohuTANG: This is another problem, we don't want unused config items explicitly:
```
[storage]
# fs|s3
type = "s3"

[storage.fs] -- this config

[storage.s3]
bucket = "databend"
endpoint_url = "https://s3.amazonaws.com"
access_key_id = "<your-key-id>"
secret_access_key = "<your-access-key>"

[storage.azblob] -- this config
```

Is it possible to configure the item only we used, like `storage.s3`?

dantengsky:
> @dantengsky Maybe I need to borrow some work from your PR to get JAVA setup correctly?

Yeah, currently, hive PR's ut/it is not integrated with the github workflows yet. Just a local  hadoop + hive cluster for the testings. there are no build-time dependencies on the JDK/jar files.  A docker image seems to be able to cover it.

But for this PR, a docker image may not be enough. hope I get it right:

 - to enable feature `storage-hdfs` at compiling 
    we need a JDK(for libjvm.so and some header files of jvm)
 -  to enable feature `storage-hdfs` at runtime (or ut/it)
    an HDFS cluster (for the specified version) is needed

query/src/configs/config_storage.rs
Xuanwo: > If we don't add the `HdfsConfig` item to *.toml, the config deserialize looks will crash?

If `HdfsConfig` is not added in `*.toml`, we will use the default value instead.

I tested this behavior locally: `query` is able to start without adding hdfs-related staff.

BohuTANG: That's great, I am out, thank you.
I will remove all the unused config items from the documents.

Xuanwo:
> > @dantengsky Maybe I need to borrow some work from your PR to get JAVA setup correctly?
> 
> Yeah, currently, hive PR's ut/it is not integrated with the github workflows yet. Just a local hadoop + hive cluster for the testings. there are no build-time dependencies on the JDK/jar files. A docker image seems to be able to cover it.
> 
> But for this PR, a docker image may not be enough. hope I get it right:
> 
> * to enable feature `storage-hdfs` at compiling
>   we need a JDK(for libjvm.so and some header files of jvm)
> * to enable feature `storage-hdfs` at runtime (or ut/it)
>   an HDFS cluster (for the specified version) is needed

Thanks for the advice! Maybe we can make databend-query compilable in this PR and test it in the next PR.

BohuTANG (approved):
:+1:

PR #5253, opened 2022-05-09
feat(planner): support  `order by` in new planner

### LINKED ISSUE #5202: Feature: `order by` for new planner
**Summary**
The test suit contains many order by related sqls.
https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/03_dml/03_0003_select_group_by.sql

xudong963:

## Summary

Support `order by` in new planner

## Changelog

- New Feature

## Related Issues

Fixes #5202


query/src/sql/planner/binder/bind_context.rs
```diff
@@ -43,13 +43,18 @@ pub struct ColumnBinding {
...
     /// The relational operator in current context
     pub expression: Option<SExpr>,
 
     /// Aggregation scalar expression
     pub agg_scalar_exprs: Option<Vec<Scalar>>,
+
+    /// Order by columnBinding, consider the sql: select a as c from t1 order by c, a,
+    /// Order by requires not just the columns in the selection,
+    /// but the columns of the entire table as well as the columns of the selection
+    pub group_by_columns: Vec<ColumnBinding>,
```
leiysky, line 57: This is not part of standard SQL. We'd better not to support it to keep our code clear.

query/src/sql/planner/binder/bind_context.rs
```diff
@@ -43,13 +43,19 @@ pub struct ColumnBinding {
...
     /// The relational operator in current context
     pub expression: Option<SExpr>,
 
     /// Aggregation scalar expression
     pub agg_scalar_exprs: Option<Vec<Scalar>>,
+
+    /// Order by columnBinding, consider the sql: select sum(a) from t group by a,b order by b;,
+    /// Order by requires not just the columns in the selection,
+    /// but the columns of the entire table as well as the columns of the selection
+    ///
```
xudong963, line 57: I updated the annotation @leiysky 

The `bind_select_stmt` method returns a `BindContext` whose columns only contain colums of selection, but for the following case:
```sql
mysql> create table t(a int, b int);

mysql> insert into t values(1, 2),(2,3)
    -> ;
Read 2 rows, 16.00 B in 0.011 sec., 184.19 rows/sec., 1.44 KiB/sec.
mysql> select sum(a) from t group by a,b order by b;
+--------+
| sum(a) |
+--------+
|      1 |
|      2 |
+--------+
2 rows in set (0.02 sec)
Read 2 rows, 16.00 B in 0.006 sec., 310.92 rows/sec., 2.43 KiB/sec.
```
`Order by` alao need the column of `a`, `b`, so in the case `order_by_columns` is used to store `a`, `b`, `sum(a)`

leiysky: I think it's better to pass a `has_order_by` hint to `bind_select_stmt`, and associate it with related code.

query/src/sql/planner/binder/select.rs
```diff
@@ -47,15 +47,26 @@ impl Binder {
...
             _ => Err(ErrorCode::UnImplement("Unsupported query type")),
         }?;
 
         if !query.order_by.is_empty() {
             let bind_context_cols = bind_context.columns.clone();
-            bind_context.columns = bind_context.order_by_columns.clone();
+            bind_context.columns = bind_context
+                .order_by_columns
+                .as_ref()
+                .ok_or_else(|| ErrorCode::LogicalError("Order by should have order by columns"))?
```
leiysky, line 68: Better to use `SemanticError`.

tests/suites/0_stateless/20+_others/20_0001_planner_v2.sql
```diff
@@ -60,11 +60,15 @@ select * from t1 inner join t on t.a = t1.b;
 select * from t2 inner join t on t.a = t2.c;
 select * from t2 inner join t on t.a = t2.c + 1;
 select * from t2 inner join t on t.a = t2.c + 1 and t.a - 1 = t2.c;
+select count(*) from numbers(1000) as t inner join numbers(1000) as t1 on t.number = t1.number;
 
+-- order by
+SELECT number%3 as c1, number%2 as c2 FROM numbers_mt (10) order by c1 desc, c2 asc;
+SELECT number, null from numbers(3) order by number desc;
+SELECT number%3 as c1, number%2 as c2 FROM numbers_mt (10) order by c1, number desc;
```
zhang2014, line 68: Maybe need test `SELECT COUNT() FROM numbers_mt(10) GROUP BY number ORDER BY SUM(number)`;

zhang2014: `SELECT SUM(number) AS s FROM numbers_mt(10) GROUP BY number ORDER BY s`;

query/src/sql/planner/binder/project.rs
```diff
@@ -121,8 +127,17 @@ impl Binder {
...
-                        scalar: Some(Box::new(bound_expr)),
+                        scalar: Some(Box::new(bound_expr.clone())),
                     };
+                    if has_order_by
+                        && !matches!(bound_expr, Scalar::BoundColumnRef(BoundColumnRef { .. }))
+                    {
+                        output_context
+                            .order_by_columns
+                            .as_mut()
+                            .unwrap()
```
zhang2014, line 138: maybe return result::err is better?

tests/suites/0_stateless/20+_others/20_0001_planner_v2.sql
xudong963: > Maybe need test `SELECT COUNT() FROM numbers_mt(10) GROUP BY number ORDER BY SUM(number)`;

Nice, the sort expression(s) can be any expression that would be valid in the query's select list, so I need to process its input schema and produce its output schema. Thank you!

query/src/sql/planner/binder/project.rs
xudong963: In fact, `order_by_columns` is always `Some`, because it's initialized by `Option` with empty vector, so `unwrap` won't panic

BohuTANG: It's better to handle the unwrap(Though it may not happen) here :D

tests/suites/0_stateless/20+_others/20_0001_planner_v2.sql
xudong963: I haven't figured out how to support aggregate functions in order by, but now that other expressions than aggregate functions are supported, I'll note an issue.

xudong963:
Let's merge it first? Then I can do other tasks based on the ticket.

PR #5262, opened 2022-05-09
feat(format): add scan progress values

zhang2014:

## Summary

feat(format): add scan progress values

## Changelog

- Bug Fix

## Related Issues

Fixes #issue


query/src/servers/http/v1/multipart_format.rs
```diff
@@ -128,6 +131,7 @@ impl MultipartWorker {
 impl MultipartFormat {
     pub fn input_sources(
         name: &str,
+        ctx: Arc<QueryContext>,
```
BohuTANG, line 134: It would be better to put the `ctx` as the first argument, we can refine it in another PR.

BohuTANG (approved):

Issue #5257, opened 2022-05-09
Remove unauthenticated behavior


junnplus:
https://github.com/datafuselabs/databend/blob/5fdce0da2207cd72a0c03bfce0ce3e79f0d777c6/query/src/servers/http/middleware.rs#L98

PR #5263, opened 2022-05-09
Remove unauthenticated behavior

junnplus:

## Summary

Remove unauthenticated behavior

## Changelog

- Improvement

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/5257


query/src/servers/http/middleware.rs
```diff
@@ -38,14 +39,22 @@ pub struct HTTPSessionMiddleware {
...
         let msg = &format!("Multiple {} headers detected", AUTHORIZATION);
         return Err(ErrorCode::AuthenticateFailure(msg));
     }
     if auth_headers.is_empty() {
-        return Ok(None);
+        return Err(ErrorCode::AuthenticateFailure(
+            "No authorization header detected",
+        ));
+    }
+    let mut hostname = headers.get(HOST).unwrap().to_str().unwrap_or("%");
```
BohuTANG, line 53: Do we need to handle the unwrap ?

query/tests/it/servers/http/http_query_handlers.rs
```diff
@@ -424,15 +424,15 @@ async fn test_query_log() -> Result<()> {
...
-    assert!(
-        result.data[0][3]
-            .as_str()
-            .unwrap()
-            .to_lowercase()
-            .contains("backtrace"),
-        "{:?}",
-        result
-    );
+    // assert!(
```
BohuTANG, line 427: stale codes or temporary fix?

youngsofun:
I think we should use https://docs.rs/poem/latest/poem/web/struct.RemoteAddr.html

query/tests/it/servers/http/http_query_handlers.rs
```diff
@@ -543,13 +546,16 @@ async fn post_json_to_endpoint(
...
     let body = serde_json::to_vec(&json)?;
+    let basic = headers::Authorization::basic("root", "");
 
     let response = ep
         .call(
             Request::builder()
                 .uri(uri.parse().unwrap())
                 .method(Method::POST)
                 .header(header::CONTENT_TYPE, content_type)
+                .header(header::HOST, "localhost")
```
youngsofun, line 557: not safe  if client can set it

query/tests/it/servers/http/http_query_handlers.rs
junnplus: Sorry, this is debugging locally.

query/tests/it/servers/http/http_query_handlers.rs
junnplus: I understand that the client cannot set it, but I removed it.

query/src/servers/http/middleware.rs
junnplus: I changed to use `remote_addr` of request.

query/src/servers/http/middleware.rs
```diff
@@ -57,21 +64,21 @@ fn get_credential(headers: &HeaderMap) -> Result<Option<Credential>> {
                 let c = Credential::Password {
                     name,
                     password,
-                    hostname: None,
+                    hostname: client_ip,
```
youngsofun, line 67: can u1@<some_ip> fallback  to  "u1"@"%"` too?

youngsofun: do we always use the IP/hostname as a part of  user_id?

or

just may localhost special? and let other addrs map to "%" directly?

junnplus: > can u1@<some_ip> fallback to "u1"@"%"` too?

Yes. see https://github.com/datafuselabs/databend/blob/main/query/src/users/user_mgr.rs#L69-L75

query/src/servers/http/middleware.rs
```diff
@@ -38,15 +38,22 @@ pub struct HTTPSessionMiddleware {
...
     }
     if auth_headers.is_empty() {
-        return Ok(None);
+        return Err(ErrorCode::AuthenticateFailure(
+            "No authorization header detected",
+        ));
     }
+    let client_ip = match req.remote_addr().0 {
+        Addr::SocketAddr(addr) => Some(addr.ip().to_string()),
+        Addr::Custom(..) => Some("127.0.0.1".to_string()),
```
youngsofun, line 54: where is this case used

youngsofun (approved):

query/src/servers/http/middleware.rs
junnplus: For testing,  `ep.call` will not have remote addr

PR #5266, opened 2022-05-10
feat(doc): add Array(T) docs

b41sh:

## Summary

add `Array(T)` docs

We can provide two types of arrays.
- `Array(T)` is used to store data with the pre-defined inner data type, similar to [ClickHouse's `Array(T)`](https://clickhouse.com/docs/en/sql-reference/data-types/array). 
- The semi-structured `Array` is used to store JSON-type data.

## Changelog

- Documentation

## Related Issues

Fixes #issue


BohuTANG:
Cool, we should document the difference with semi-struct array, such as performance etc.

BohuTANG (approved):

PR #5270, opened 2022-05-10
refactor: Move configs to common so we can reuse it

Xuanwo:

## Summary

refactor: Move configs to common so we can reuse it

For example:

- Use configs in `common_grpc`
- Use configs in `common_meta_grpc`
- TODO: Use configs in stage storage, so we don't need to declare `S3StorageConfig` twice.

## Changelog

- Improvement


Xuanwo:
> * TODO: Use configs in stage storage, so we don't need to declare `S3StorageConfig` twice.

@sundy-li Is it a good idea for you?

BohuTANG:
This is a good idea, but `query.rs` and `config.rs` are only used by the query, it's not common, preferred to keep them in `query/src/configs`.

Xuanwo:
> but `query.rs` and `config.rs` are only used by the query, it's not common, preferred to keep them in `query/src/configs`.

`common_grpc` and `common_meta` will also depend on them to build their own conf struct.

Comparing `cfg.query.to_grpc_client_config()`, maybe it's better to use `impl From<&MetaConfig> for MetaGrpcClientConf`?

BohuTANG:
> > but `query.rs` and `config.rs` are only used by the query, it's not common, preferred to keep them in `query/src/configs`.
> 
> `common_grpc` and `common_meta` will also depend on `config` to build their own conf struct.
> 
> Comparing `cfg.query.to_grpc_client_config()`, maybe it's better to use `impl From<&MetaConfig> for MetaGrpcClientConf`?

Ok, `config.rs` can be kept in `query/src/configs`?

Xuanwo:
> Ok, `config.rs` can be kept in `query/src/configs`?

LGTM, I will address that.

drmingdrmer (approved):
LGTM:DDD

BohuTANG (approved):

PR #5272, opened 2022-05-10
docs: Algolia enforce using databend-rs as prefix

Xuanwo:

## Summary

The new upgrade from Algolia enforce us to use databend-rs as prefix instead of name.

So we will migrate to `databend-rs-docs`.

## Changelog

- Documentation


Xuanwo:
Sorry for my mistake, this PR should be revered. @BohuTANG

PR #5279, opened 2022-05-10
feat(planner): support subquery table reference type for new planner

### LINKED ISSUE #5203: Feature: support  table reference type: subquery  for new planner
**Summary**

To make the following tests pass
```sql
SELECT a,b,count() from (SELECT cast((number%4) AS bigint) as a, cast((number%20) AS bigint) as b from numbers(100)) group by a,b order by a,b limit 3 ;
```

xudong963:

## Summary

support table reference type: subquery for new planner

## Changelog

- New Feature

## Related Issues

Fixes #5203


query/src/sql/planner/binder/select.rs
```diff
@@ -260,7 +271,10 @@ impl<'a> Binder {
                 Ok((s_expr, bind_context))
             }
             TableReference::Join(join) => self.bind_join(bind_context, join).await,
-            _ => Err(ErrorCode::UnImplement("Unsupported table reference type")),
+            TableReference::Subquery { subquery, alias } => {
+                self.bind_query_with_alias(subquery, alias, bind_context)
+                    .await
+            }
```
leiysky, line 277: I think you can just call `apply_table_alias` here.

sundy-li (approved):

query/src/sql/planner/binder/select.rs
xudong963: Do you mean directly processing columns of current `bind_context`?
```rust
        if let Some(table_alias) = table_alias {
            for column_binding in bind_context.clone().all_column_bindings().iter() {
                if let Some(origin_table_name) = &column_binding.table_name {
                    bind_context.apply_table_alias(origin_table_name, table_alias)?;
                }
            }
        }
```

query/src/sql/planner/binder/project.rs
```diff
@@ -74,9 +75,17 @@ impl<'a> Binder {
     pub(super) async fn normalize_select_list(
         &mut self,
         select_list: &[SelectTarget<'a>],
-        input_context: &BindContext,
+        table_alias: &Option<TableAlias<'a>>,
```
zhang2014, line 78: What happen if this query:
```
SELECT * FROM system.columns, (SELECT 1) AS `system.columns`;
```

query/src/sql/planner/binder/select.rs
leiysky: Yes, this work has been done by `apply_table_alias`. Actually, the argument `original_name` is useless now, you can remove it safely.

query/src/sql/planner/binder/project.rs
xudong963: Duplicated table names will result in an error.

BohuTANG:
Oops, conflicts need solved

query/src/sql/planner/binder/project.rs
zhang2014: \`system.table\` should not be equal to system.table

https://www.db-fiddle.com/f/k2vZTW2y8Vm9AjUSTPkVGv/1

query/src/sql/planner/binder/project.rs
```diff
@@ -74,7 +74,7 @@ impl<'a> Binder {
     pub(super) async fn normalize_select_list(
         &mut self,
         select_list: &[SelectTarget<'a>],
-        input_context: &BindContext,
+        input_context: &mut BindContext,
```
leiysky, line 77: This `mut` can be removed.

query/src/sql/planner/binder/project.rs
leiysky: > `system.table` should not be equal to system.table
> 
> https://www.db-fiddle.com/f/k2vZTW2y8Vm9AjUSTPkVGv/1

The "system.table" should be parsed into identifier. It won't be an issue.

query/src/sql/planner/binder/project.rs
BohuTANG: Make the `BindContext` as the first parameter of the function is more make sense.

leiysky: > Make the `BindContext` as the first parameter of the function is more make sense.

I've just opened a pr for this https://github.com/datafuselabs/databend/pull/5371.

leiysky (approved):

query/src/sql/planner/binder/project.rs
xudong963: I see it, hope they won't conflict :D

BohuTANG:
Hmm, Conflicting files after another merged.

PR #5280, opened 2022-05-10
refactor: Reuse StorageConfig in stage

Xuanwo:

## Summary

refactor: Reuse StorageConfig in stage

This PR will reuse `StorageConfig` in the stage so that we don't need to maintain two sets of storage parse and handle logic.

## Changelog

- Improvement


BohuTANG (approved):
:100:

BohuTANG:
Stateful CI failure:
```
00_0000_copy_from_s3_location:                                          2022-05-10T13:02:19.818118Z ERROR databend_query::servers::mysql::writers::query_result_writer: OnQuery Error: Code: 3001, displayText = entity not found (object error: (op: stat, path: admin/data/ontime_200.csv/admin/data/ontime_200.csv, source: response part: Parts { status: 404, version: HTTP/1.1, headers: {"accept-ranges": "bytes", "content-length": "0", "content-security-policy": "block-all-mixed-content", "server": "MinIO", "strict-transport-security": "max-age=31536000; includeSubDomains", "vary": "Origin", "vary": "Accept-Encoding", "x-amz-request-id": "16EDBFFEE5BBF315", "x-content-type-options": "nosniff", "x-xss-protection": "1; mode=block", "date": "Tue, 10 May 2022 13:02:19 GMT"} }, body: "")).
```

Xuanwo:
@mergify update

common/proto-conv/src/config_from_to_protobuf_impl.rs
```diff
@@ -0,0 +1,46 @@
...
+            root: p.root,
+            master_key: p.master_key,
+        })
+    }
+
+    fn to_pb(&self) -> Result<pb::S3StorageConfig, Incompatible> {
+        Ok(pb::S3StorageConfig {
+            region: self.region.clone(),
+            endpoint_url: self.endpoint_url.clone(),
+            access_key_id: self.access_key_id.clone(),
```
Xuanwo, line 39: @drmingdrmer Is it safe to store ak/sk in meta directly? Are there some consideration here?

BohuTANG:
@mergify update

common/proto-conv/src/config_from_to_protobuf_impl.rs
drmingdrmer: hmm... metasrv would be a nice place to store it if with some strong encryption. 🤔

Xuanwo:
@BohuTANG All check passed.

common/protos/proto/config.proto
```diff
@@ -12,6 +12,16 @@
...
 
-mod file_s3;
+syntax = "proto3";
 
-pub use file_s3::S3File;
+package databend_proto;
+
+message S3StorageConfig {
+  string region = 1;
+  string endpoint_url = 2;
```
drmingdrmer, line 21: There should be a `ver` to identify the `proto-conv` version if this message has chances to be used alone. Otherwise, without a `ver`, it implies `proto-conv` will always have to be able to deal with any old version `S3StorageConfig` message.

Xuanwo: I agree with your concern.

Configs will have their own version flags (our binary also relies upon the same struct). I'm going to add them in another PRs.

Xuanwo: Tracked at https://github.com/datafuselabs/databend/issues/5297

BohuTANG: @Xuanwo  I think it would be nice to do it in this PR, or it would be a compatibility issue again.

Xuanwo: I'm using the default `VER` for now and will have a complete solution after #5297 is finished.

drmingdrmer (approved):

Xuanwo:
After 4 PRs merged, my PR is still in the queue :sob:

Xuanwo:
Cheers! :1st_place_medal:

PR #5294, opened 2022-05-11
chore(query): introduce meta Runtime

sundy-li:

## Summary

Summary about this PR

## Changelog

 
- Improvement
 

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/sessions/query_ctx.rs
+++ b/query/src/sessions/query_ctx.rs
@@ -395,6 +395,10 @@ impl QueryContext {
         self.shared.session.session_mgr.get_storage_runtime()
     }
 
+    pub fn get_meta_runtime(&self) -> Arc<Runtime> {
+        self.shared.session.session_mgr.get_meta_runtime()
+    }
+
     pub async fn reload_config(&self) -> Result<()> {
         self.shared.reload_config().await
     }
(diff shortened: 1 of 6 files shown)


query/src/sessions/query_ctx.rs
```diff
@@ -424,6 +430,30 @@ impl QueryContext {
         })?;
         Ok(FunctionContext { tz })
     }
+
+    pub fn block_on_meta<F>(&self, future: F) -> Result<F::Output>
+    where
```
sundy-li, line 435: With this function, `show function ` can work well in Mac Osx.


I do not know whether this code is right or not.  cc @drmingdrmer  @dantengsky

query/src/sessions/query_ctx.rs
```diff
@@ -424,6 +430,30 @@ impl QueryContext {
...
+            tx.send(res).unwrap();
+        })?;
+        match rx.recv() {
+            Ok(v) => Ok(v),
+            Err(cause) => Err(ErrorCode::LogicalError(format!(
+                "Logical error, receive error. {:?}",
+                cause
+            ))),
+        }
+    }
```
drmingdrmer, line 456: One concern: this method can not be called in a tokio-runtime, right? This method itself can not provide any guarantee on this. It's totally up to the caller.

Except this, everything else looks great to me:DDD

drmingdrmer (approved):

query/src/sessions/query_ctx.rs
sundy-li: I think it can, because `QueryNormalizer::normalize` is inside tokio-runtime.

sundy-li:
hmm, this approach may block the runtime, I revert it first.

PR #5318, opened 2022-05-12
fix: hashtable

usamoi:

## Summary

A trivial typo in query/src/common/hashtable/hash_table_entity.rs. There is no reference to this function so it has not caused problems before.

## Changelog

- Not for changelog


zhang2014 (approved):
Thank you contribution. @usamoi

PR #5319, opened 2022-05-12
[meta/schema-api]: rename-table need to check table exist before returning OK

drmingdrmer:

## Related Issues

### DIFF
--- a/.github/actions/test_stateful_standalone/action.yml
+++ b/.github/actions/test_stateful_standalone/action.yml
@@ -60,3 +60,4 @@ runs:
           metasrv/_logs*/
           query/_logs*/
           store/_logs*/
+          nohup.out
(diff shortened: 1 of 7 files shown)


drmingdrmer:
Thanks to @TCeason for reminding me about this bug:DDD

TCeason:
LGTM. By the way maybe we can delete the “if exists” token.

drmingdrmer:
> LGTM. By the way maybe we can delete the “if exists” token.

delete it from where? 🤔

TCeason:
> > LGTM. By the way maybe we can delete the “if exists” token.
> 
> delete it from where? 🤔

Like MySQL or Oracle rename table not support if exists.

drmingdrmer:
> > > LGTM. By the way maybe we can delete the “if exists” token.
> > 
> > 
> > delete it from where? 🤔
> 
> Like MySQL or Oracle rename table not support if exists.

I do not know, forward this question to @BohuTANG  :DD

BohuTANG:
I think it's Ok here. Let's merge!

PR #5324, opened 2022-05-12
RFC: Config Backward Compatibility

Xuanwo:

## Summary

[rendered](https://github.com/Xuanwo/databend/blob/versioned-config/docs/doc/60-contributing/03-rfcs/20220512-config-backward-compatibility.md)

## Changelog

- New Feature


Xuanwo:
Request review from:

- @flaneur2020 @ZhiHanZ for config files/envs/args
- @drmingdrmer @lichuang for work related to `proto-conv`

Other reviews are also welcome.

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
```diff
@@ -0,0 +1,141 @@
...
+
+# Guide-level explanation
+
+With this RFC, users will have a new field called `version` at the top of the config file. The value of the version is an integer and defaults to `0`. If the version is not specified, we will load it as` 0`.
+
+All config files, env, and args will be supported:
+
+- config file: `version=42`
+- config env: `export CONFIG_VERSION=42`
+- args: `--config-version=42`
```
flaneur2020, line 42: the versioned config approach is commonly seen in many projects, but `--version` args approach is very quite familiar, can we have some similar examples in the other projects?

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
```diff
@@ -0,0 +1,146 @@
...
+
+The outer config of the query will be like this:
+
+```rust
+#[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize, Parser)]
+#[clap(about, version, author)]
+#[serde(default)]
+pub struct ConfigV1 {
+    #[clap(long, short = 'c', default_value_t)]
+    pub config_file: String,
```
sundy-li, line 84: Do we need to create a new version if we just add new fields into the old config ?

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
Xuanwo: To be honest, I didn't find a project accepting args like `config-version` :smile_cat: 

`--version` is different from `--config-version` we used here. `--version` has often been used as a flag which prints app versions and `--config-version` intends to specify the config version we are trying to load.

Although we support the usage like:

```shell
databend-query -c ./old-version.toml --config-version=42
```

But I think we don't need to mention the behavior inside docs. Because the args parser always uses the latest version (

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
Xuanwo: No, we only need to bump versions if we are doing incompatible changes like `config removed/renamed/changed`

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
```diff
@@ -0,0 +1,146 @@
...
+With this RFC, our users will upgrade their deployments without breaking. Old configs should always work along with the new implementations.
+
+# Guide-level explanation
+
+With this RFC, users will have a new field called `version` at the top of the config file. The value of the version is an integer and defaults to `0`. If the version is not specified, we will load it as` 0`.
+
+All config files, env, and args will be supported:
+
+- config file: `version=42`
+- config env: `export CONFIG_VERSION=42`
```
flaneur2020, line 41: what'd happen when the versions between the config env and config file got conflicts?

Xuanwo: There are no conflicts. As they are the same field, `env` will always overwrite `file`. (The same rules we apply for all fields.)

Xuanwo: Well, I came up with some ideas like config file and env are using a different version, can they by loaded at the same time.

It's possible, but a bit complex. I expect to support the overwrite behavior first.

Will be added in `Future possibilities` sections.

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
Xuanwo: > Because the args parser always uses the latest version (

Well, it's also possible to support different versions of args...

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
flaneur2020: > Well, I came up with some ideas like config file and env are using a different version, can they by loaded at the same time?

oh i'd hope this not happen, it's a bit anti-intuitive, and hard to reason about which field really acts in effect

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
flaneur2020: how about taking the args in a different manner? IMO args is harder to envolve, and less chance to change. 

etcd marks the obsoleted args with a DEPRECATED flag, so it give users some freedom to transist his/her args in production: https://etcd.io/docs/v3.2/op-guide/configuration/

flaneur2020: oh, also in etcd, it has an env var called ETCDCTL_API=3 to envolve the args, FYI

```
export ETCDCTL_API=3
```

Xuanwo: As described in RFC, command UX is another topic, and I think we should discuss it in a new RFC. In this RFC, we will keep the current behavior that all config fields map to a unique arg.

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
flaneur2020: we can also consider the DEPRECATED approach when we need removing fields. just leave some time to users to remove the deprecated fields.

IMO, the config version change mostly happens with some developer-experience-driven redesign.

Xuanwo: Add a new `Config Maintenance` section to describe the actions we should take while updating config.

Xuanwo:
Our community members have great concern about the complexity of introducing a versioned config at this stage(no stable release, no production users). It seems better to split `outer` and `inner` configs now. Leave the decision of versioned config till the future.

I have updated this RFC and moved content about the versioned config to `Future possibilities` part. Please review again.

@sundy-li @flaneur2020 @BohuTANG

flaneur2020:
LGTM :+1:

docs/doc/60-contributing/03-rfcs/20220512-versioned-config.md
Xuanwo: Addressed via https://github.com/datafuselabs/databend/pull/5324#issuecomment-1125794431

flaneur2020 (approved):

sundy-li (approved):

PR #5381, opened 2022-05-14
feat(function): Support connection_id function

b41sh:

## Summary

support MySQL  function [connection-id](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_connection-id)
It is used by [mycli](https://www.mycli.net/) client

## Changelog

- New Feature
- Documentation

## Related Issues

Fixes #issue


query/src/common/context_function.rs
```diff
@@ -46,6 +46,9 @@ impl ContextFunction {
             "current_user" => vec![Expression::create_literal(DataValue::String(
                 ctx.get_current_user()?.identity().to_string().into_bytes(),
             ))],
```
BohuTANG, line 48: New planner also need add the case.

BohuTANG: https://github.com/datafuselabs/databend/pull/5369/files#diff-37b754ad52b168f02e969eb2486df14bde8efb8c819a218053a1dcf79bff6b50R591

b41sh: ok

BohuTANG (approved):

Issue #4311, opened 2022-03-03
Feature: metasrv: add API to return metrics


drmingdrmer:
Add API to return raft metrics in:
https://github.com/datafuselabs/databend/blob/16e06e414c4680f0d640abada631af89369be877/metasrv/src/api/http_service.rs#L46-L57

The metrics can be retrieved from `HttpService.meta_node.raft.metrics()`

RinChanNOWWW:
Hello @LiuYuHui , are you still working on this issue?

LiuYuHui:
Hi @RinChanNOWWW , thanks for your reminder, please go ahead, sorry for the inconvenience.

drmingdrmer:
Thanks man!

drmingdrmer:
@lichuang 
Does the following PR fix this issue?
- #5376

PR #5389, opened 2022-05-15
feat(metasrv): add metrics to http service

RinChanNOWWW:

## Summary

Add new http api "/metrics" which returns the raft metrics.

Like:

```json
{
    "running_state": {
        "Ok": null
    },
    "id": 1,
    "state": "Leader",
    "current_term": 1,
    "last_log_index": 2,
    "last_applied": {
        "term": 1,
        "index": 2
    },
    "current_leader": 1,
    "membership_config": {
        "log_id": {
            "term": 0,
            "index": 0
        },
        "membership": {
            "configs": [
                [
                    1
                ]
            ],
            "all_nodes": [
                1
            ]
        }
    },
    "snapshot": null,
    "leader_metrics": {
        "replication": {}
    }
}
```

image

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/4311

### DIFF
--- a/metasrv/tests/it/api/http/metrics.rs
+++ b/metasrv/tests/it/api/http/metrics.rs
@@ -1,0 +1,64 @@
+// Copyright 2022 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common_base::base::tokio;
+use databend_meta::api::http::v1::metrics::metrics_handler;
+use databend_meta::meta_service::MetaNode;
+use poem::get;
+use poem::http::Method;
+use poem::http::StatusCode;
+use poem::http::Uri;
+use poem::Endpoint;
+use poem::EndpointExt;
+use poem::Request;
+use poem::Route;
+use pretty_assertions::assert_eq;
+
+use crate::init_meta_ut;
+use crate::tests::service::MetaSrvTestContext;
+
+#[tokio::test]
+async fn test_metrics() -> common_exception::Result<()> {
+    let (_log_guards, ut_span) = init_meta_ut!();
+    let _ent = ut_span.enter();
+
+    let tc = MetaSrvTestContext::new(0);
+
+    let meta_node = MetaNode::start(&tc.config.raft_config).await?;
+
+    let cluster_router = Route::new()
+        .at("/v1/metrics", get(metrics_handler))
+        .data(meta_node.clone());
+    let response = cluster_router
+        .call(
+            Request::builder()
+                .uri(Uri::from_static("/v1/metrics"))
+                .method(Method::GET)
+                .finish(),
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let body = response.into_body().into_vec().await.unwrap();
+    let metrics =
+        serde_json::from_str::<serde_json::Value>(String::from_utf8_lossy(&body).as_ref())?;
+
+    metrics["has_leader"].as_bool().unwrap();
+
+    meta_node.stop().await?;
+
+    Ok(())
+}
--- a/metasrv/src/api/http/v1/metrics.rs
+++ b/metasrv/src/api/http/v1/metrics.rs
@@ -1,0 +1,38 @@
+// Copyright 2022 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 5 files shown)


RinChanNOWWW:
It just simply returns the result of `meta_node.raft.metrics()`, please confirm if it meets the requirement @drmingdrmer . And I only wrote a simple test.

metasrv/tests/it/api/http/metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+        .call(
+            Request::builder()
+                .uri(Uri::from_static("/v1/metrics"))
+                .method(Method::GET)
+                .finish(),
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(response.status(), StatusCode::OK);
```
drmingdrmer, line 39: A test has to assert the JSON format.

E.g., in our case, assert expected keys present in the response JSON and the value format(integer, string) is as expected

drmingdrmer (commented):
Good PR!

metasrv/tests/it/api/http/metrics.rs
RinChanNOWWW: I have a problem about thrift when running tests on my own pc. Can you help me to figure out what's wrong?

```
error[E0432]: unresolved import `thrift::protocol::TSerializable`
  --> path/to/databend/target/debug/build/common-hive-meta-store-2361c9c96a485357/out/hms_patched.rs:16:142
   |
16 | ...ssageType, TInputProtocol, TOutputProtocol, TSerializable, TSetIdentifier, TStructIdentifier, TType};
   |                                                ^^^^^^^^^^^^^ no `TSerializable` in `protocol`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `common-hive-meta-store` due to previous error
```

BohuTANG: Try to run 'make setup -d'

lichuang:
we now implement a metrics API in `http://metric_api_address:metrics`, see `metrics_service.rs`, but it is in Prometheus style, not json style.

The metrics you implement here is raft metrics, may be is not suitable to metasrv users, such as current_term, etc, the metasrv user may not care about all the raft metrics.

See the struct `MetaMetrics`, it is the metrics which metasrv user care about, may be can add HTTP api of this struct in json?

RinChanNOWWW:
OK, I'll take a look at `MetaMetrics`.

RinChanNOWWW:
I found there is only one field `has_leader`  in `MetaMetrics`. So I just return this?

```json
{
    "has_leader":  true # or false
}
```

or something like Prometheus format:

```json
{
    "has_leader": {
        "type": "gauge",
        "namespace": "metasrv",
        "subsystem": "serer",
        "help": "Whether or not a leader exists.",
        "value": 1.0 # or 0.0 or true or false
    }
}
```

PTAL @lichuang

metasrv/tests/it/api/http/metrics.rs
```diff
@@ -0,0 +1,102 @@
...
+        },
+        \"id\": 0,
+        \"state\": \"Leader\",
+        \"current_term\": 1,
+        \"last_log_index\": 2,
+        \"last_applied\": {
+            \"term\": 1,
+            \"index\": 2
+        },
+        \"current_leader\": 0,
```
drmingdrmer, line 72: The values in a json are volatile.

The test needs to decode the json and check the keys one by one.
And for the values, only check the type of them.

RinChanNOWWW: > The values in a json are volatile.
> 
> The test needs to decode the json and check the keys one by one. And for the values, only check the type of them.

OK. Waiting for confirming the content of the repsonse, or there will be a lot of useless work.

lichuang:
> I found there is only one field `has_leader` in `MetaMetrics`. So I just return this?
> 
> ```json
> {
>     "has_leader":  true # or false
> }
> ```
> 
> or something like Prometheus format:
> 
> ```json
> {
>     "has_leader": {
>         "type": "gauge",
>         "namespace": "metasrv",
>         "subsystem": "server",
>         "help": "Whether or not a leader exists.",
>         "value": 1.0 # or 0.0 or true or false
>     }
> }
> ```
> 
> PTAL @lichuang

i think the first way is good enough:)

RinChanNOWWW:
Updated:

```json
{
    "has_leader": true
}
```

image

Like `node_handler` and `state_handler`, the json fields is hard-coded by `Json(serde_json:json!()))`, since it's not convenient to convert `MetaMetrics` to `serde_json::Value`.

metasrv/tests/it/api/http/metrics.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let body = response.into_body().into_vec().await.unwrap();
+    let metrics =
+        serde_json::from_str::<serde_json::Value>(String::from_utf8_lossy(&body).as_ref())?;
+
+    metrics["has_leader"].as_bool().unwrap();
+
+    meta_node.stop().await?;
```
drmingdrmer, line 61: You do not need to manually stop it.

It will be dropped when this test function returned.

drmingdrmer (approved):

metasrv/tests/it/api/http/metrics.rs
RinChanNOWWW: I referred to the test codes in `cluster_state_test.rs` 😂.

drmingdrmer: Hmm...Then that is unnecessary too...

BohuTANG:
Thanks for the contribution!

PR #5395, opened 2022-05-16
Feat(httphandler): result download

youngsofun:

## Summary

part of https://github.com/datafuselabs/databend/issues/5398

## 1. result_table 

`result table` is used in multi places:

- downloading with formats
- table function like [result_scan(<query_id>)](https://docs.snowflake.com/en/sql-reference/functions/result_scan.html)
- internal cache for other query

this pr only implement the `download` feature.
but to better support the latter two, I use `fuse segment` as storage format. so with regard to read，
the `result table` has the same ability as `fuse table`. 
another consideration is that we can reuse a lot of code of fuse table.

### 1.1 write to result_table

supported first in http hanlder.

when it pull a block from interpreter:

1. the block is written to cache table first，**no longer blocked by client (because limited buffer)**，
2. the block or its `part_info` (if `max_rows_in_buffer` is reached) is appended to `block_buffer` in memory.
3. client get block/`part_info` from `block_buffer`,  for `part_info`, reread datablock from `result_table`

### 1.2 download result 

 download data via new endpoint `/v1/query/<query-id>/download`
 it read from the `result table` and transform to desired format. 
 
 - tsv is supported first，csv and json will be support in next pr，
 - the formated output may be cached too later.

## 2. other related change to http handler 

1. page size(in rows) is decoupled from blocksize. 
2. add `kill_uri` in response to distingush if from `final_uri`,
   1. `kill_uri` is like sql `KILL QUERY 'query_id' `
   2. `final_uri` is used to tell server to auto release alive state in memory (paging, cache..) right after the query is done.

## Changelog

- New Feature

## Related Issues
part of https://github.com/datafuselabs/databend/issues/5398


BohuTANG:
Can we add a stateful test for this download?

query/src/storages/result/download.rs
```diff
@@ -0,0 +1,93 @@
...
+    type Err = String;
+    fn from_str(s: &str) -> std::result::Result<Self, String> {
+        match s.to_uppercase().as_str() {
+            "TSV" => Ok(DownloadFormatType::Tsv),
+            _ => Err("Unknown file format type, must be one of { TSV }".to_string()),
+        }
+    }
+}
+
+fn serialize_block(block: DataBlock, fmt: DownloadFormatType) -> Result<Vec<u8>> {
```
sundy-li, line 55: maybe it's time to move this util function into `query/src/formats/format_xxx.rs`

youngsofun: ok

put input and output logics of `csv` in the same file `format_csv.rs`?  cc @zhang2014 
output interface is very diff from output.


other related:

- query::formats::format::{FormatFactory, InputFormat}
  -  input only
- comon_io::format_settings::FormatSettings 
  -  output only
- common::meta::types::user_stage{StageFileFormatType, FileFormatOptions}
  -  input/output only

do we need a unified enum for FileFormatType?

Xuanwo: In https://github.com/datafuselabs/databend/pull/5399, I moved the storage-related config into `common-io` and removed all stage's own storage config. I think it's OK to add a new FileFormatType here.

Only one note: `FileFormatType` is part of our contract to meta API, we should make it will not break `proto-conv`.

zhang2014 (approved):

Issue #5297, opened 2022-05-11
Tracking issue of RFC Config Backward Compatibility


Xuanwo:
## Background

There should be a `ver` to identify the `proto-conv` version if this message has chances to be used alone. Otherwise, without a `ver`, it implies `proto-conv` will always have to be able to deal with any old version `S3StorageConfig` message.

_Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/pull/5280#discussion_r869913827_

## RFC

[Config Backward Compatibility](https://github.com/datafuselabs/databend/blob/4aab38cc0ab46a3f718788d402ef53554c62f2c3/docs/doc/60-contributing/03-rfcs/20220512-config-backward-compatibility.md)

## Roadmap

- [x] https://github.com/datafuselabs/databend/pull/5324
- [x] https://github.com/datafuselabs/databend/pull/5399
- [x] https://github.com/datafuselabs/databend/pull/5409
- [x] https://github.com/datafuselabs/databend/pull/5421
- [x] Add tests for both inner and outer config
- [x] Add detailed docs for config maintain

## Unresolved Problems

- serde alias may not work as expected, we need to handle the backward compatibility by hand

Xuanwo:
Most work have been done. Let's close.

PR #5399, opened 2022-05-16
refactor(query/config/storage): Adapt RFC Config Backward Compatibility

Xuanwo:

## Summary

Part of https://github.com/datafuselabs/databend/issues/5297

This PR adds an outer config for `query` Config: `OuterConfigV0`.

We will make sure that:

- Implement `TryFrom OuterConfigV0 Into Config`
- Implement `From Config Into OuterConfigV0`
- All internal functions are implemented against on inner `Config`.
- All public layers will use `config.into_outer()` to expose OuterConfigV0
  - User Config / Env / Args
  - system.configs Table
  - HTTP handler

In this PR, we refactor `StorageConfig` in an enum, so that we can use storage-related configs with confidence.

Only `StorageConfig` has been adapted, other config fields will be updated in the following PRs.

## Changelog

- Improvement


sundy-li (approved):

flaneur2020 (approved):

BohuTANG (approved):
:+1:

Xuanwo:
Some stage-related tests is broken, I'm working on it.

PR #5410, opened 2022-05-16
Feature: Support DISTINCT in new planner 

### LINKED ISSUE #5343: Feature: Support `DISTINCT` in new planner
In SQL, we can use `DISTINCT` to deduplicate the result set of a query, and produce a multi-set.

```SQL
SELECT DISTINCT a, b FROM t; -- The output won't contain duplicated (a, b)
```

This is essentially equivalent with `GROUP BY a, b`.

ygf11:

## Summary

Support base `DISTINCT` and `DISTINCT` with group by.
## Changelog

- New Feature

## Related Issues

Fixes #5343


query/src/sql/exec/mod.rs
```diff
@@ -362,6 +363,12 @@ impl PipelineBuilder {
             })?;
         }
 
+        // Since transform has added or did, making group expressions as column expr is safe.
+        group_expressions = group_expressions
```
xudong963, line 367: The group expression may have been a column, so there may be duplicate conversions here.

xudong963, line 367: The group expression may have been a column, so there may be duplicate conversions here.

ygf11: Yes, if a group expression is a column, it is unnecessary here.

The desitnation of adding this code is to make distinct with group-by work.
Two ways to make it better:
1. We do this only when aggregation is generated by distinct.
2. Move this when bind distinct.

query/src/sql/exec/mod.rs
```diff
@@ -364,6 +366,12 @@ impl PipelineBuilder {
             })?;
         }
 
+        // Since transform has been added, making group expressions as column expr is safe.
+        group_expressions
+            .iter_mut()
+            .filter(|expr| matches!(expr, Expression::Column(_)).not())
+            .for_each(|expr| *expr = Expression::Column(expr.column_name()));
+
```
leiysky, line 374: Maybe we can use a `iter().filter().map().collect()` instead?

ygf11: If the original group expression is a column, then it will not be in the collect result.

leiysky (approved):

PR #5416, opened 2022-05-17
feat(test): Sql logic test framework improve

ZeaLoVe:

## Summary

Improve the sqllogictest framework
Add http_connector and http_runner implement.
Add generated test suites which will be use to test both  http handler and mysql handler.

## Changelog

- Improvement
- Build/Testing/CI
- Documentation

## Related Issues

Part of https://github.com/datafuselabs/databend/issues/4955
Part of https://github.com/datafuselabs/databend/issues/2647
RFC see pr: https://github.com/datafuselabs/databend/pull/5048


tests/logictest/requiresments.txt
```diff
@@ -0,0 +1,4 @@
+mysql-connector>=2.2.9
```
Xuanwo, line 1: We are using poetry to maintain python deps.

See https://github.com/datafuselabs/databend/blob/main/tests/pyproject.toml for more details.

tests/logictest/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Usage
+
+## prepare
+1. Change to the scripts dir, cd tests/logictest/
+2. Make sure python3 is installed
+3. Using pip to install dependency, pip install -r ./requirements.txt  (use -i for private repo)
+
+## Need to Known
+1. Cases from **tests/suites/0_stateless/**  to  **tests/logictest/suites/gen/**
```
sundy-li, line 14: If we already gen cases from stateless dir, should we keep the old stateless tests ?

ZeaLoVe: Unless the logictest framework is perfect, before that, the two mechanisms are parallel

ZeaLoVe: Unless the logictest framework is perfect, before that, the two mechanisms are parallel. Better keep those.

tests/logictest/requiresments.txt
ZeaLoVe: OK，need a commit to use poetry instead of pip

tests/logictest/README.md
sundy-li: Ok. Almost LGTM.  

If this pr is merged, which test style should we use?

BohuTANG:
Can we skip the CI build-check if only the `tests/logictest` files are changed? cc @Xuanwo

tests/logictest/README.md
ZeaLoVe: sqllogictest is better, developers can see every single case results or error(code or msg). But it's a long term to make it good enough.

Xuanwo:
> Can we skip the CI build-check if only the `tests/logictest` files are changed? cc @Xuanwo

We can use `path-ignores`： https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

```yaml
on:
  push:
    paths-ignore:
      - 'docs/**'
```

tests/logictest/README.md
sundy-li: Ok, since two version tests will make developers confuse and poor maintenance. To keep fewer duplicated tests I suggest:

1. Delete the `stateless` tests directory if this pr is merged.
2. Or we generate `gen` directories on the fly and keep this directory git ignore.

BohuTANG: We should do it, or the stateless will add new tests and need to sync the test files with logictest. @ZeaLoVe

PR #5409, opened 2022-05-16
refactor(query/config): Adapt RFC Config Backward Compatibility

Xuanwo:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Part of https://github.com/datafuselabs/databend/issues/5297

Followup of https://github.com/datafuselabs/databend/pull/5399

After this PR, `databend-query` will maintain config backward compatibility.

## Changelog

- Improvement

PR #5417, opened 2022-05-17
fix(test): fix httphandler flaky test.

youngsofun:

## Summary

fix https://github.com/datafuselabs/databend/pull/5409#issuecomment-1128379748

@Xuanwo 
the first HTTP request return as long as a page of 2 rows is generated, maybe before the SQL finish running and  `succeed` state is set, depending on the task scheduling,  although `select * from numbers(10)`  should run very fast.

## Changelog


Xuanwo (approved):

Xuanwo:
Thanks for the quick fix! Impressive.

PR #5419, opened 2022-05-17
feat(function): support date_add for new parser

### LINKED ISSUE #5076: Combine datetime interval add&subsruct  to one DATE_ADD
**Summary**

Syntax:
```
​​DATE_ADD('<unit>', <interval>, <date_or_time_expr>)
```

`<unit>`: `SECOND`, `MINUTE`, `HOUR`, `DAY`, `WEEK`, `MONTH`, `YEAR`
`<interval>`: Number, this can be negative for substract
`<date_or_time_expr>`:  The expression evaluate to a date, time, or timestamp

Example:
```
SELECT toDate(18875), DATE_ADD('YEAR', 2, toDate(18875));
+---------------+-----------------------------+
| 2021-09-05    | 2023-09-05                  |
+---------------+-----------------------------+

SELECT toDate(18875), DATE_ADD('YEAR', -2, toDate(18875));
+---------------+-----------------------------+
| 2021-09-05    | 2019-09-05                  |
+---------------+-----------------------------+
```

fkuner:

## Summary

support sql:

```
set enable_planner_v2=1;
select date_add(to_date(18321), 1, YEAR);
select date_add(to_date(18321), -1, YEAR);
select date_add(to_date(18321), 1, SECOND);

create table t(a int);
insert into t values(1), (2);
select date_add(to_date(18321), a, YEAR) from t;
drop table t;

set enable_planner_v2=0;
```

## Changelog

- New Feature

## Related Issues

Fixes #5076


common/functions/src/scalars/dates/date.rs
```diff
@@ -124,5 +133,16 @@ impl DateFunction {
...
+        factory.register("toIntervalYear", ToIntervalYearFunction::desc());
+        factory.register("toIntervalMonth", ToIntervalMonthFunction::desc());
+        factory.register("toIntervalDay", ToIntervalDayFunction::desc());
+        factory.register("toIntervalHour", ToIntervalHourFunction::desc());
+        factory.register("toIntervalMinute", ToIntervalMinuteFunction::desc());
+        factory.register("toIntervalSecond", ToIntervalSecondFunction::desc());
+        factory.register("toIntervalDoy", ToIntervalDoyFunction::desc());
+        factory.register("toIntervalDow", ToIntervalDowFunction::desc());
+
+        factory.register("dateAdd", DateAddFunction::desc());
```
sundy-li, line 146: Let's use better snake styles.

```
to_interval_xxx
date_add
```

fkuner: ok

fkuner:
We don't support grammer like `interval '1' year` now. But we can use `interval 1 year` or `interval 1+2 year` after we convert to the new parser.

fkuner:
Ths cluster mode seems not to use the new parser. Should we ignore it?

common/ast/src/ast/expr.rs
```diff
@@ -689,6 +693,17 @@ impl<'a> Display for Expr<'a> {
                 write_comma_separated_list(f, exprs)?;
                 write!(f, "]")?;
             }
+            Expr::Interval { expr, unit, .. } => {
+                write!(f, "INTERVAL({expr} {unit})")?;
```
sundy-li, line 697: seems we should not need parentheses.

common/ast/src/ast/expr.rs
```diff
@@ -689,6 +693,17 @@ impl<'a> Display for Expr<'a> {
...
+            }
+            Expr::DateAdd {
+                date,
+                interval,
+                unit,
+                ..
+            } => {
+                write!(f, "DATEADD({date}, INTERVAL({interval} {unit}))")?;
```
sundy-li, line 705: ditto

common/functions/src/scalars/dates/to_interval_function.rs
```diff
@@ -0,0 +1,85 @@
...
+
+macro_rules! impl_to_interval_function {
+    ($FUNCTION_NAME:ident, $INTERVAL_KIND: ident) => {
+        #[derive(Clone)]
+        pub struct $FUNCTION_NAME {
+            display_name: String,
+        }
+
+        impl $FUNCTION_NAME {
+            pub fn try_create(
```
sundy-li, line 35: Wow, I don't think it's good practice to use macros here.

Why not use :

```
IntervalFunction {
   pub try_create(display_name: &str,  interval_kind: IntervalKind) -> Result<Box<dyn Function>> {}
}

```

See [cast](https://github.com/datafuselabs/databend/blob/b9ea557c4a060b4622c1287bc7af71e126e609be/common/functions/src/scalars/expressions/cast.rs)

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -553,6 +567,114 @@ impl<'a> TypeChecker<'a> {
         }
     }
 
+    pub async fn resolve_interval(
+        &mut self,
+        arg: &Expr<'a>,
+        interval_kind: &IntervalKind,
+        _required_type: Option<DataTypeImpl>,
+    ) -> Result<(Scalar, DataTypeImpl)> {
+        match interval_kind {
```
sundy-li, line 576: Too long, the match should be inside the functions.

common/functions/src/scalars/dates/to_interval_function.rs
fkuner: make sense

query/src/sql/planner/semantic/type_check.rs
fkuner: seems can't do it. the function can't receive `IntervalKind`

common/ast/src/ast/expr.rs
fkuner: ok. actually parentheses is more beautiful

query/src/sql/planner/semantic/type_check.rs
fkuner: emm..seems can't do this. If we do this must, we will do some repeat work. I think it is unnecessary.

common/functions/src/scalars/dates/to_interval_function.rs
```diff
@@ -0,0 +1,78 @@
...
+        IntervalType::new_impl(self.interval_kind)
+    }
+
+    fn eval(
+        &self,
+        _func_ctx: FunctionContext,
+        columns: &ColumnsWithField,
+        _input_rows: usize,
+    ) -> Result<ColumnRef> {
+        IntervalType::new_impl(self.interval_kind).create_column(&columns[0].column().to_values())
```
sundy-li, line 62: should not use `create_column`.

Just case the columns[0] into Int64Column.

common/functions/src/scalars/dates/to_interval_function.rs
```diff
@@ -0,0 +1,78 @@
...
+    interval_kind: IntervalKind,
+}
+
+impl ToIntervalFunction {
+    pub fn try_create(
+        display_name: String,
+        interval_kind: IntervalKind,
+        _args: &[&DataTypeImpl],
+    ) -> Result<Box<dyn Function>> {
+        Ok(Box::new(ToIntervalFunction {
```
sundy-li, line 40: Must check that args[0] is numeric datatype.

common/ast/src/parser/expr.rs
```diff
@@ -750,6 +818,8 @@ pub fn expr_element(i: Input) -> IResult<WithSpan> {
             | #binary_op : "<operator>"
             | #unary_op : "<operator>"
             | #cast : "`CAST(... AS ...)`"
+            | #date_add: "`DATEADD(..., ..., (YEAR| MONTH | DAY | HOUR | MINUTE | SECOND | DOY | DOW))`"
```
b41sh, line 821: what do `DOY` and `DOW` mean?

sundy-li (approved):

common/ast/src/parser/expr.rs
sundy-li: DAY OF WEEK/YEAR

b41sh: ok

common/ast/src/parser/token.rs
```diff
@@ -788,7 +790,17 @@ impl TokenKind {
...
+            | TokenKind::WITH
+            | TokenKind::DATEADD
+            | TokenKind::YEAR
+            | TokenKind::MONTH
+            | TokenKind::DAY
+            | TokenKind::HOUR
+            | TokenKind::MINUTE
+            | TokenKind::SECOND
+            | TokenKind::DOY
+            | TokenKind::DOW
```
andylokandy, line 802: If these keywords are not reserved, will there be any ambiguity?

common/ast/src/ast/expr.rs
```diff
@@ -689,6 +693,17 @@ impl<'a> Display for Expr<'a> {
                 write_comma_separated_list(f, exprs)?;
                 write!(f, "]")?;
             }
+            Expr::Interval { expr, unit, .. } => {
```
andylokandy, line 696: I prefer to move it to `Literal` since there is already `CurrentTimestamp`.

common/ast/src/parser/token.rs
fkuner: I remeber that it will be recognitized as a column.

common/ast/src/ast/expr.rs
fkuner: We can discuss with @sundy-li to find better way to dowith it.

common/ast/src/parser/token.rs
fkuner: Is the syntax I defined  maybe incorrect？I will check it.

andylokandy: If it's recongized as a column, it should be in is_reserved_ident rather than is_reserved_function_name

andylokandy: i think reserving only the date_add as function namd should solve the problem.

common/ast/src/ast/expr.rs
sundy-li: It might not be literal if we want to have column-based Interval.

So we choose to have `interval expr unit`, surely if it's literal it may make life easier.

refer: [Doris](https://doris.apache.org/sql-manual/sql-functions/date-time-functions/date_add.html#date-add)

sundy-li: Will it take a bad effect on the query `select Year, count(1) from ontime group by Year `  ?

fkuner: If we put token like YEAR, MONTH.. as reserved indent, it will take a bad effect. In fact, I do this at beigin which can't pass the stateful-test. So, I recoginize them as reserved function.

common/ast/src/parser/token.rs
fkuner: Ok

common/ast/src/ast/expr.rs
fkuner: If we set `YEAR` to reserved ident, it will take a bad effect. 

Now I submit a new pr #5490 to delete `YEAR` from reserved function. It is not unnecessary.

PR #5423, opened 2022-05-17
feat(planner): support `using` and `natural` for join

xudong963:

## Summary

- support `using` and `natural` for join in new planner

## Changelog

- New Feature

## Related Issues

Fixes #issue


query/src/sql/planner/binder/bind_context.rs
```diff
@@ -44,6 +44,11 @@ pub struct ColumnBinding {
     /// Another example is aggregation. In a `GROUP BY` context, aggregate funtions
     /// will be extracted and be added to `BindContext` as a `ColumnBinding`.
     pub scalar: Option<Box<Scalar>>,
+
+    /// Consider the sql: `select * from t join t1 using(a)`.
+    /// The result should only contain one `a` column.
+    /// So we need mark `t.a` or `t1.a` duplicated.
+    pub duplicated: bool,
```
leiysky, line 51: `duplicated` is a bit confusing. Maybe it's rename to `visibleInUnqualifiedWildcard`.

leiysky (approved):

query/src/sql/planner/binder/join.rs
```diff
@@ -252,44 +265,122 @@ impl<'a> JoinConditionResolver<'a> {
...
+            }
+            if left_scalars.is_empty() {
+                return Err(ErrorCode::SemanticError(format!(
+                    "column {} specified in USING clause does not exist in left table",
+                    join_key_name
+                )));
+            }
+            assert_eq!(left_scalars.len(), 1);
+            let mut right_scalars = vec![];
+            for col_binding in self.right_context.columns.iter() {
```
sundy-li, line 300: maybe better use `find`

BohuTANG: We can do it in another PR :)

PR #5424, opened 2022-05-17
fixes(insert): fix drop dispatcher when commit insert query

zhang2014:

## Summary

fixes(insert): fix drop dispatcher when commit insert query

## Changelog

- Bug Fix

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/interpreters/interpreter_insert.rs
+++ b/query/src/interpreters/interpreter_insert.rs
@@ -16,6 +16,7 @@ use std::collections::VecDeque;
 use std::sync::Arc;
 
 use chrono_tz::Tz;
+use common_base::base::TrySpawn;
 use common_base::infallible::Mutex;
 use common_datavalues::DataType;
 use common_exception::ErrorCode;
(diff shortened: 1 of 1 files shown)


query/src/interpreters/interpreter_insert.rs
```diff
@@ -160,15 +161,24 @@ impl InsertInterpreter {
...
         let append_entries = self.ctx.consume_precommit_blocks();
-        table
-            .commit_insertion(
-                self.ctx.clone(),
-                &self.plan.catalog_name,
-                append_entries,
-                self.plan.overwrite,
-            )
-            .await?;
+        let handler = self.ctx.get_storage_runtime().spawn(async move {
```
BohuTANG, line 168: We should add some comments to tell why we need a `spawn` here :D

dantengsky (approved):
LGTM

query/src/interpreters/interpreter_insert.rs
zhang2014: Sorry, I have added and committed comment, but I forgot to push : (

BohuTANG: We can do it in another PR :)

Issue #5379, opened 2022-05-14
`record_delimiter` breaking the show settings


BohuTANG:
**Summary**

```
+--------------------------------+---------+---------+---------+----------------------------------------------------------------------------------------------------+--------+
| name                           | value   | default | level   | description                                                                                        | type   |
+--------------------------------+---------+---------+---------+----------------------------------------------------------------------------------------------------+--------+
| empty_as_default               | 1       | 1       | SESSION | Format empty_as_default, default value: 1                                                          | UInt64 |
| enable_new_processor_framework | 1       | 1       | SESSION | Enable new processor framework if value != 0, default value: 1                                     | UInt64 |
| enable_planner_v2              | 0       | 0       | SESSION | Enable planner v2 by setting this variable to 1, default value: 0                                  | UInt64 |
| field_delimiter                | ,       | ,       | SESSION | Format field delimiter, default value: ,                                                           | String |
| flight_client_timeout          | 60      | 60      | SESSION | Max duration the flight client request is allowed to take in seconds. By default, it is 60 seconds | UInt64 |
| group_by_two_level_threshold   | 10000   | 10000   | SESSION | The threshold of keys to open two-level aggregation, default value: 10000                          | UInt64 |
| max_block_size                 | 10000   | 10000   | SESSION | Maximum block size for reading                                                                     | UInt64 |
| max_threads                    | 16      | 16      | SESSION | The maximum number of threads to execute the request. By default, it is determined automatically.  | UInt64 |
| record_delimiter               | 
       | 
       | SESSION | Format record_delimiter, default value: 
                                                          | String |
| skip_header                    | 0       | 0       | SESSION | Whether to skip the input header, default value: 0                                                 | UInt64 |
| storage_read_buffer_size       | 1048576 | 1048576 | SESSION | The size of buffer in bytes for buffered reader of dal. By default, it is 1MB.                     | UInt64 |
| timezone                       | UTC     | UTC     | SESSION | Timezone, default value: UTC,                                                                      | String |
+--------------------------------+---------+---------+---------+----------------------------------------------------------------------------------------------------+--------+

```

https://github.com/datafuselabs/databend/blob/e31a5f341cc720ee8e602ed29ab59679eefa2f88/query/src/sessions/session_settings.rs#L112-L116

Need escape.

PR #5452, opened 2022-05-18
stype(functions): format record_delimiter show settings

hantmac:
Signed-off-by: hanshanjie <hanshanjie@meituan.com>

## Summary

Format `record_delimiter` show settings.

## Changelog
- Improvement

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/5379


Xuanwo:
Thanks for contributing!

Maybe it's better to escape the \n into "\n"?

query/src/sessions/session_settings.rs
```diff
@@ -110,10 +110,10 @@ impl Settings {
                 desc: "Enable planner v2 by setting this variable to 1, default value: 0",
             },
             SettingValue {
-                default_value: DataValue::String("\n".as_bytes().to_vec()),
-                user_setting: UserSetting::create("record_delimiter", DataValue::String("\n".as_bytes().to_vec())),
+                default_value: DataValue::String(" ".as_bytes().to_vec()),
+                user_setting: UserSetting::create("record_delimiter", DataValue::String(" ".as_bytes().to_vec())),
                 level: ScopeLevel::Session,
-                desc: "Format record_delimiter, default value: \n",
+                desc: "Format record_delimiter, default value:  ",
```
Xuanwo, line 116: We do use `\n` as `record_delimiter`. Changing it into space is wrong.

I prefer to escape it while in displaying

```suggestion
                desc: "Format record_delimiter, default value: \\n",
```

hantmac: I got it, this default value of this field is `\n` itself. 
I will fix it. Thank you for your review.

sundy-li: Now we escaped `record_delimiter` from "\n"  to be "\\n", but it will make `record_delimiter ` not work. Because some codes depend on this settings, like [format_csv](https://github.com/datafuselabs/databend/blob/3b34c0f6a79bfaa2dedfc3622f09d7aa73902163/query/src/formats/format_csv.rs#L88)

hantmac: > Now we escaped `record_delimiter` from "\n" to be "\n", but it will make `record_delimiter ` not work. Because some codes depend on this settings, like [format_csv](https://github.com/datafuselabs/databend/blob/3b34c0f6a79bfaa2dedfc3622f09d7aa73902163/query/src/formats/format_csv.rs#L88)

Got it, so I will close this pr and read the code carefully.

PR #5461, opened 2022-05-19
feat(function): support object_keys function

### LINKED ISSUE #4667: Feature: Semi-structured data function object_keys
**Summary**

Add support Semi-structured Data Function `OBJECT_KEYS` to get keys from Object.

https://docs.snowflake.com/en/sql-reference/functions/object_keys.html

fkuner:

## Summary

support object_keys function

## Changelog

- New Feature

## Related Issues

Fixes #4667


common/functions/src/scalars/semi_structureds/object_keys.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+impl Function for ObjectKeysFunction {
+    fn name(&self) -> &str {
+        &*self.display_name
+    }
+
+    fn return_type(&self) -> DataTypeImpl {
+        ArrayType::new_impl(StringType::new_impl())
```
b41sh, line 58: we‘d better return `VariantArray` type, compatible with Snowflake [`object_keys`](https://docs.snowflake.com/en/sql-reference/functions/object_keys.html#returns)

sundy-li (approved):

common/functions/src/scalars/semi_structureds/object_keys.rs
fkuner: Make sense. But Arrat<T> seems gain better performance?

b41sh: ok

PR #5471, opened 2022-05-19
chore: remove unneccesary user creation during tenant boot call

ZhiHanZ:

## Summary

Use hardcoded user with password is not the best practice and depreciated, we should only use `call admin$bootrap_tenant(tenantName)` to create necessary roles
## Changelog

- Improvement

## Related Issues

Fixes #issue


flaneur2020 (approved):

junnplus (approved):

BohuTANG:
02_0000_kill_query.py failure is not related to this PR. Let's merge.

PR #5476, opened 2022-05-19
feat(table statistics): add statistics to `TableMeta`

dantengsky:

## Summary

- Add statistic to `TableMeta`
- Commit whole new table meta to meta server, instead of upsert table option only
 
NOTE:
  - **For existing tables -- even if none-empty, the statistics will be zeros
     after an insertion, correct statistics will be populated** 
 - tables populated after this PR checked in, shall generate statistics in time
  - `index_size` is None, stats of this attribute have not been collected yet.
## Changelog

- New Feature


flaneur2020 (approved):
thanks!

BohuTANG (approved):

PR #5493, opened 2022-05-20
Fix the documents avgIf to avg_if

hanyisong:

##Summary
Change the document function avgIf to avg_if


BohuTANG:
Thanks for your contribution!

Xuanwo:
Hi, thanks for contributing!

There is something we should take care about:

- Please don't rename the file path into Chinese, as we rely on the file path to generate correct docs.

Xuanwo:
ping @BohuTANG, please roll this PR back.

image

Xuanwo:
Addressed by https://github.com/datafuselabs/databend/pull/5494

Issue #5323, opened 2022-05-12
[metasrv] unify kv_service and meta_service


ariesdevil:
**Summary**
Since we now use kv-api for all meta data, there is no need to distinguish kv-api and meta-api. cc @Xuanwo @drmingdrmer @lichuang 
ref: https://github.com/datafuselabs/databend/discussions/5320

drmingdrmer:
There is only one `try_get_meta_store()` method now in `MetaStoreProvider`

PR #5495, opened 2022-05-20
[improving] store grpc addr to node info and auto refresh backends addrs for grpc client

### LINKED ISSUE #5308: [metasrv] store grpc_addr automatically when a new metasrv node launching.
**Summary**
Store grpc_addr automatically when a new metasrv node launching and remove it when node shutdown, so the grpc client could fetch all available meta-grpc backend addrs for connection.

ariesdevil:

## Summary

* [x] Store grpc addr to node info
* [x] auto-refresh backends addrs for grpc client
* [x] tests

## Changelog

- New Feature
- Improvement

## Related Issues

Fixes #5308


common/meta/types/src/meta_errors.rs
```diff
@@ -82,7 +82,7 @@ pub enum MetaError {
     Fatal(AnyError),
 }
 
-pub type MetaResult<T> = std::result::Result<T, MetaError>;
+pub type MetaResult<T> = Result<T, MetaError>;
```
Xuanwo, line 85: :+1:

metasrv/src/bin/metasrv.rs
```diff
@@ -75,14 +75,16 @@ async fn main(_global_tracker: Arc<RuntimeTracker>) -> common_exception::Result<
...
+            conf.grpc_api_address.clone()
         );
         srv.start().await.expect("Databend meta service error");
         stop_handler.push(Box::new(srv));
     }
 
     // join raft cluster after all service started
-    meta_node.join_cluster(&conf.raft_config).await?;
+    meta_node
+        .join_cluster(&conf.raft_config, conf.grpc_api_address)
```
Xuanwo, line 86: I'm confused about the `raft_config`.

Will we use `raft_config` outside `metasrv`? How about merging them into `MetaConfig` directly?

query/src/common/meta/meta_client.rs
```diff
@@ -46,7 +46,7 @@ impl MetaClientProvider {
 
     /// Get kv async client, operations trait defined in KVApi.
     pub async fn try_get_kv_client(&self) -> Result<Arc<dyn KVApi>> {
-        let local = self.grpc_conf.meta_service_config.address.is_empty();
+        let local = self.grpc_conf.kv_service_config.address.is_empty();
```
Xuanwo, line 49: Considering https://github.com/datafuselabs/databend/issues/5323

Will we adopt `kv_service_config` always and remove `meta_service_config`?

Xuanwo:
`test_metactl` failed:

```shell
Error: missing field `grpc_api_addr` at line 1 column 195
```

Seems we are introducing changes that are not compatible.

metasrv/src/bin/metasrv.rs
ariesdevil: `raft_config` is all about raft related config, `MetaConfig` includes other configs that nothing to do with raft. So I think I think it's better to separate.

drmingdrmer: Right...It's a bit dirty:

RaftConfig is also used by `common/meta/embedded`, which is an embedded meta server for testing purposes.

drmingdrmer:
This looks like a breaking change: An old metasrv will crash loading the data written by a new metasrv that has `grpc_api_addr` in a `Node` record.

What about making it an `Option`?
```rust
pub struct Node {
    pub name: String,
    pub endpoint: Endpoint,
    pub grpc_api_addr: Option<String>, // <--- 🤔
}
```

drmingdrmer:
Everything else looks good to me :D

drmingdrmer:
If this PR is still under development, convert it to draft?

common/meta/grpc/src/grpc_client.rs
```diff
@@ -346,7 +366,18 @@ impl MetaGrpcClient {
...
-                    resp.map(message::Response::MakeClient)
+                    match resp {
+                        Ok(r) => Ok(message::Response::MakeClient(r)),
+                        Err(e) => {
+                            tracing::warn!("make client error, try to remake. {:?}", e);
+                            let res = self.make_client().await;
+                            res.map(message::Response::MakeClient)
+                        }
+                    }
+                }
```
drmingdrmer, line 377: Isn't it better to let `self.make_client()` do the retry-loop when failed to build a client?

common/meta/grpc/src/grpc_client.rs
```diff
@@ -491,25 +474,124 @@ impl MetaGrpcClient {
...
+            *current_endpoint = None;
+        }
+
+        let mut eps = self.get_endpoints().await;
+        debug_assert!(!eps.is_empty());
+
+        // remove unhealthy endpoints
+        {
+            let ues = self.unhealthy_endpoints.lock().unwrap();
+            eps.retain(|e| !ues.contains_key(e));
```
drmingdrmer, line 489: Is `retain()` a method of `TtlHashMap`?

common/meta/grpc/src/grpc_client.rs
```diff
@@ -491,25 +474,124 @@ impl MetaGrpcClient {
...
     pub async fn set_endpoints(
         &self,
         endpoints: Vec<String>,
     ) -> std::result::Result<(), MetaError> {
         Self::endpoints_non_empty(&endpoints)?;
 
+        let distinct_cnt = endpoints.iter().filter(|n| !(*n).is_empty()).count();
+        if distinct_cnt < endpoints.len() / 2 + 1 {
+            tracing::warn!(
+                "distinct endpoints small than majority of meta cluster nodes {}<{}, endpoints: {:?}",
```
drmingdrmer, line 550: It may need a comment explaining why it does so

common/meta/grpc/src/grpc_client.rs
```diff
@@ -491,25 +474,124 @@ impl MetaGrpcClient {
...
+        Ok(())
+    }
+
+    async fn auto_sync_endpoints(self: Arc<Self>, mut cancel_tx: OneSend<()>) {
+        loop {
+            select! {
+                _ = cancel_tx.closed() => {
+                    return;
+                }
+                _ = sleep(Duration::from_secs(5)) => {
```
drmingdrmer, line 584: `endpoints` changes rarely, hence the sleep time could be relatively longer than several seconds IMHO.

common/meta/grpc/src/grpc_client.rs
```diff
@@ -775,10 +848,24 @@ impl MetaGrpcClient {
...
+        {
+            let ca = self.current_endpoint.lock().unwrap();
+            let mut ue = self.unhealthy_endpoints.lock().unwrap();
+            ue.insert((*ca).as_ref().unwrap().clone(), ());
+        }
+        {
+            let mut token = self.token.write().await;
+            *token = None;
+        }
+    }
```
drmingdrmer, line 861: Naming a method with what it does would be better than with what it does for.

E.g., to me, (maybe unthoughtful name) `mark_as_unhealthy()` would be better than `preapre_retry_req()`

common/meta/grpc/src/grpc_client.rs
ariesdevil: nop, it's a `Vec`'s method

common/meta/grpc/src/grpc_client.rs
ariesdevil: Good idea

common/meta/grpc/src/grpc_client.rs
```diff
@@ -158,6 +164,10 @@ pub struct MetaGrpcClient {
 pub struct ClientHandle {
     /// For sending request to meta-client worker.
     pub(crate) req_tx: Sender<message::ClientWorkerRequest>,
+    /// Notify auto sync to stop.
+    /// `oneshot::Receiver` impl `Drop` by sending a closed notification to the `Sender` half.
+    #[allow(dead_code)]
+    cancel_auto_sync_rx: OneRecv<()>,
```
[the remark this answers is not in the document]
ariesdevil: it's still a dead_code since no explicitly used

common/meta/types/src/cluster.rs
```diff
@@ -38,11 +38,21 @@ pub struct Slot {
...
 }
 
 impl fmt::Display for Node {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}={}", self.name, self.endpoint)
+        let grpc_addr_display = if self.grpc_api_addr.is_some() {
+            self.grpc_api_addr.clone().unwrap()
+        } else {
+            "".to_string()
+        };
```
drmingdrmer, line 50: `let x = if let Some(v) = ... { A } else { B }`

common/meta/grpc/src/grpc_client.rs
```diff
@@ -775,10 +878,24 @@ impl MetaGrpcClient {
...
+    async fn mark_as_unhealthy(&self) {
+        {
+            let ca = self.current_endpoint.lock();
+            let mut ue = self.unhealthy_endpoints.lock();
+            ue.insert((*ca).as_ref().unwrap().clone(), ());
+        }
+        {
+            let mut token = self.token.write().await;
+            *token = None;
+        }
```
drmingdrmer, line 890: There is a concurrency issue that before 888, another task created a client, then the current task reset the token to None, which belongs to another connection.

common/meta/grpc/src/grpc_client.rs
```diff
@@ -498,18 +559,69 @@ impl MetaGrpcClient {
...
+            .await?;
+        let result: Vec<String> = endpoints.into_inner().data;
+        self.set_endpoints(result).await?;
+        Ok(())
+    }
+
+    async fn auto_sync_endpoints(self: Arc<Self>, mut cancel_tx: OneSend<()>) {
+        if self.auto_sync_interval == Duration::from_secs(0) {
+            return;
+        }
```
drmingdrmer, line 609:
```
if let Some(interval) = self.auto_sync_interval {
}
```

common/meta/grpc/src/grpc_client.rs
```diff
@@ -418,77 +446,110 @@ impl MetaGrpcClient {
...
+    }
 
-            item?
+    async fn make_channel(&self, addr: Option<&String>) -> std::result::Result<Channel, MetaError> {
+        let addr = if addr.is_some() {
+            addr.unwrap().clone()
+        } else {
+            let eps = self.endpoints.read().await;
+            eps.first().unwrap().clone()
         };
```
drmingdrmer, line 529: Use `if let Some(...)`

drmingdrmer (approved):

metasrv/src/meta_service/raftmeta.rs
```diff
@@ -503,17 +504,17 @@ impl MetaNode {
                 body: ForwardRequestBody::Join(JoinRequest {
                     node_id: conf.id,
                     endpoint: raft_api_advertise_host_endpoint.clone(),
+                    grpc_api_addr: grpc_api_addr.clone(),
```
lichuang, line 507: grpc_api_addr seems no more used, so can remove .clone()?

ariesdevil: This is an assignment in a for loop, so need to clone here.

metasrv/src/meta_service/raftmeta.rs
```diff
@@ -691,7 +702,18 @@ impl MetaNode {
...
         let sm = self.sto.state_machine.read().await;
-        sm.get_metasrv_addrs()
+        let nodes = sm.get_nodes()?;
+        let endpoints: Vec<String> = nodes
+            .iter()
+            .map(|n| {
+                if let Some(addr) = n.grpc_api_addr.clone() {
+                    addr
+                } else {
+                    "".to_string()
```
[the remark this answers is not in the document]
ariesdevil: Here is for compatibility with older versions, the client will handle null values. I'll add some comments here.

BohuTANG:
We need a benchmark test for this patch, and a check will it works well or not. cc @ariesdevil @drmingdrmer

lichuang (approved):

PR #5496, opened 2022-05-21
(WIP)stage: show stages

### LINKED ISSUE #4488: stage: show stages
## Summary

```
show stages
```

Show all the stage names of a tenant.

LiuYuHui:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues

Fixes #4488


common/protos/proto/user.proto
```diff
@@ -178,4 +178,6 @@ message UserStageInfo {
   CopyOptions copy_options = 5;
 
   string comment = 6;
+
+  string created_on = 7;
```
BohuTANG, line 182: To keep backwards compatible,it seems  need 'Option' here

Xuanwo: I don't like the idea to store a time string in metasrv.

How about converting to [Timestamp](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto) or UNIX timestamp as u64.

LiuYuHui: Good suggestion, will try it :D

LiuYuHui:
close this PR, as #5581 has better implementation than me :(.

Issue #5489, opened 2022-05-20
bug: wrong results in hash join


xudong963:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

main

### What's Wrong?

```sql
mysql> select t3.number from numbers(1000000) as t3 join (select * from numbers(100000)) as t4 on t3.number = t4.number;
Empty set (0.90 sec)
Read 200000 rows, 1.53 MiB in 0.882 sec., 226.78 thousand rows/sec., 1.73 MiB/sec.
```
The output shouldn't be empty

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

xudong963:
I'm optimizing hash join, the issue doesn't appear in my optimized version

Issue #5259, opened 2022-05-09
Improvement: `Hash Join` efficiency  improve


Veeupup:
**Summary**

For now, hash join implementation in time and space needs to improve,  we need to improve it.
cc @leiysky @sundy-li

sundy-li:
1. Velox used  unroll the loops to improve 

https://github.com/facebookincubator/velox/blob/5f55bbb38efe9bf0b342919c335a42b951cdf5fa/velox/exec/HashTable.cpp#L459-L502

2. Paper of photon used vectorized lookup 

https://www-cs.stanford.edu/~matei/papers/2022/sigmod_photon.pdf

PR #5497, opened 2022-05-21
improvement(planner): hash join ~6x performance improvement

xudong963:

## Summary

Optimize hash join

**case 1**: 
`select t3.number from numbers(10000) as t3 join (select * from numbers(1000)) as t4 on t3.number  = t4.number;` 
In the case, build table(right table) size < probe table size, there is ~6x performance improvement(from 0.434s to 0.074s).

**case 2**:
`select t3.number from numbers(10000) as t3 join (select * from numbers(1000)) as t4 on t3.number % 1000 = t4.number;`
In the case, probe expression is binary expression, there is ~1.7 performance improvement(from 0.693s to 0.4s).

**case 3**:
`select t3.number from numbers(10000) as t3 join (select * from numbers(10000)) as t4 on t3.number = t4.number;`
In the case, build table(right table) size = probe table size, there is ~1.6 performance improvement(from 0.645s to 0.411s).

In the future, continued optimization can be done in the process of supporting other join types (e.g. outer join)

For case2, the optimized version's flame graph is the following:
image

## Changelog

- Improvement
- bug fix

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/5489 https://github.com/datafuselabs/databend/issues/5259


leiysky:
Is the flame graph generated from a release build?

query/src/pipelines/new/processors/transforms/hash_join/row.rs
```diff
@@ -155,8 +152,7 @@ pub fn compare_and_combine(
 
     let data_block = DataBlock::create(Arc::new(DataSchema::new(data_fields)), columns);
 
-    let filter =
-        ExpressionEvaluator::eval(ctx.try_get_function_context()?, &predicate, &data_block)?;
+    let filter = ExpressionEvaluator::eval(func_ctx, &predicate, &data_block)?;
```
leiysky, line 155: There is no need to filter the result anymore, since the hash keys should have been compared when probing the hash table.

query/src/common/hashtable/hash_table.rs
```diff
@@ -35,6 +35,16 @@ pub struct HashTable<Key: HashTableKeyable, Entity: HashTableEntity<Key>, Grower
...
 
+unsafe impl<Key: HashTableKeyable, Entity: HashTableEntity<Key>, Grower: HashTableGrower> Send
+    for HashTable<Key, Entity, Grower>
+{
+}
+
+unsafe impl<Key: HashTableKeyable, Entity: HashTableEntity<Key>, Grower: HashTableGrower> Sync
+    for HashTable<Key, Entity, Grower>
+{
+}
```
leiysky, line 46: Maybe we should remove this?

query/src/pipelines/new/processors/transforms/hash_join/row.rs
xudong963: Nice!!! I need to update the performance result 😆

xudong963:
> Is the flame graph generated from a release build?

No, from debug build.  The performance results are from release build

query/src/common/hashtable/hash_table.rs
xudong963: It seems not, because `HashJoinState` implements send & sync

BohuTANG:
The result is not correct in my case:
```
create table t1(a int, b int);
create table t2(a int, b int);
create table t3(a int, b int);

insert into t1 select number, number+1 from numbers(100000000);
insert into t2 select number, number+1 from numbers(20000000);
insert into t3 select number, number+1 from numbers(10000000);

select count(*) from t1 inner join t2 on t1.a=t2.a inner join t3 on t2.b=t3.b;

+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
1 row in set (6.45 sec)
```

Result should be:
```
┌──count()─┐
│ 10000000 │
└──────────┘
```

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
```diff
@@ -63,113 +62,75 @@ impl ChainingHashTable {
...
+            let keys_ref = KeysRef::create(key.as_ptr() as usize, key.len());
+            let probe_result_ptr = hash_table.find_key(&keys_ref);
+            if probe_result_ptr.is_none() {
                 // No matched row for current probe row
                 continue;
             }
+            let probe_result_ptrs = probe_result_ptr.unwrap().get_value();
+            // `result_block` is the block of build table
             let result_block = self.row_space.gather(probe_result_ptrs)?;
+            let mut probe_block = DataBlock::block_take_by_indices(input, &[i as u32])?;
```
sundy-li, line 123: `result_block` and `probe_block` seems have different size.

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
```diff
@@ -44,7 +43,7 @@ pub struct ChainingHashTable {
     ctx: Arc<QueryContext>,
 
     /// A shared big hash table stores all the rows from build side
-    hash_table: RwLock<Vec<Vec<RowPtr>>>,
+    hash_table: RwLock<HashMap<KeysRef, Vec<RowPtr>>>,
```
zhang2014, line 46: KeysRef maybe memory leak?

query/src/common/hashtable/hash_table.rs
leiysky: `HashJoinState` is required to be `Sync + Send`, which is satisfied by `Arc<Mutex<HashTable>>`. It's not safe to impl `Send` and `Sync` for `HashTable` since it can not be shared between threads without a mutex guard.

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
xudong963: Yes, but I don't get your concerned point in the comment.

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
xudong963: The usage of `KeysRef` in hash join without unsafe. Could you please explain why you think `KeysRef` may cause memory leak? Thanks~

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
sundy-li: so `probe_block.add_column(col.clone(), field.clone())` may be incorrect if they are in different size.

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
zhang2014: We referenced the data by ptr, but the data has been destroyed with the iteration of the data block.

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
xudong963: got it

sundy-li:
This SQL can reproduce the bug with a smaller data set, you can use this to debug.
```
select count(*) from (select to_int32(a) as a from t1 order by a limit 10) t1 inner join (select to_int32(a) as a from t2 order by a limit 10) t2 on t1.a=t2.a
```

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
sundy-li: we should use `replicated` to expand the probe_bock to the same size as result_block;

The probe_block may not be in just one size.

For example:

```
 select count(*) from (select 1 as a from t1 order by a limit 10) t1 inner join (select 1 as a from t2 order by a limit 10) t2 on t1.a=t2.a
```
Expect to have 100 output, but just 10;

sundy-li:
```
duckdb> select count(*) from (select 1 as a from range(0,10)) t1 inner join (select 1 as a from range(0, 10)) t2 on t1.a=t2.a;
┌──────────────┐
│ count_star() │
╞══════════════╡
│          100 │
└──────────────┘
Elapsed: 4 ms

duckdb> 
```

xudong963: > Expect to have 100 output, but actually it's 10;

I tested in my local, the output is 100.
```sql
mysql>  select count(*) from (select 1 as a from t1 order by a limit 10) t1 inner join (select 1 as a from t2 order by a limit 10) t2 on t1.a=t2.a
    -> ;
+----------+
| COUNT(*) |
+----------+
|      100 |
+----------+
1 row in set (9.82 sec)
Read 30000000 rows, 228.88 MiB in 9.789 sec., 3.06 million rows/sec., 23.38 MiB/sec.

```

sundy-li: That's interesting
```
:) select count(*) from (select 1 as a from t1 order by a limit 10) t1 inner join (select 1 as a from t2 order by a limit 10) t2 on t1.a=t2.a;
+----------+
| COUNT(*) |
+----------+
|       10 |
+----------+
1 row in set (0.67 sec)
Read 30000000 rows, 228.88 MiB in 0.666 sec., 45.04 million rows/sec., 343.63 MiB/sec.
```

xudong963: t1 and t2 are the following?
```
create table t1(a int, b int);
create table t2(a int, b int);

insert into t1 select number, number+1 from numbers(100000000);
insert into t2 select number, number+1 from numbers(20000000);
```

sundy-li: Oh, with your latest commit, it's ok now

BohuTANG:
> This SQL can reproduce the bug with a smaller data set, you can use this to debug.
> 
> ```
> select count(*) from (select to_int32(a) as a from t1 order by a limit 10) t1 inner join (select to_int32(a) as a from t2 order by a limit 10) t2 on t1.a=t2.a
> ```

Will this PR fix this bug?

xudong963:
> > This SQL can reproduce the bug with a smaller data set, you can use this to debug.
> > ```
> > select count(*) from (select to_int32(a) as a from t1 order by a limit 10) t1 inner join (select to_int32(a) as a from t2 order by a limit 10) t2 on t1.a=t2.a
> > ```
> 
> Will this PR fix this bug?

I think we can fix in the following PR, the bug was not introduced by this pr.

sundy-li:
Ok  /LGTM

BohuTANG (approved):

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
```diff
@@ -63,113 +62,79 @@ impl ChainingHashTable {
...
+            let result_blocks = self.row_space.gather(probe_result_ptrs)?;
             let probe_block = DataBlock::block_take_by_indices(input, &[i as u32])?;
-            let mut replicated_probe_block = DataBlock::empty();
-            for (i, col) in probe_block.columns().iter().enumerate() {
-                let replicated_col = ConstColumn::new(col.clone(), result_block.num_rows()).arc();
 
-                replicated_probe_block = replicated_probe_block
-                    .add_column(replicated_col, probe_block.schema().field(i).clone())?;
+            for result_block in result_blocks.iter() {
+                assert_eq!(result_block.clone().num_rows(), 1);
```
leiysky, line 126: Why does result_block only have 1 row of data?

xudong963: I changed the logic of `gather` so that instead of having `concat_blocks` in gather, it just returns `Vec<DataBlock>`. So now `result_block` has only one line, the same number of rows as `probe_block`.

Xuanwo:
> I think we can fix in the following PR, the bug was not introduced by this pr.

LGTM. Please create an issue and linked here, so that we will not lose this context.

xudong963:
> LGTM. Please create an issue and linked here, so that we will not lose this context.

done!

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
leiysky: It looks wierd, take this example:

```
build side:
column1 column2
1       2
1       3
1       4

probe side:
column1
1
```
There is only one block of build side, i.e. chunks number is 1.

When you are probing with the probe key `1`, it should return all the 3 rows of data. In this case, how to make sure that `result_block` has only one row?

xudong963: I think the following code can split the single chunk to three chunks
```rust
            for row_ptr in row_ptrs.iter() {
                assert!((row_ptr.chunk_index as usize) < chunks.len());
                let block = self.gather_single_chunk(&chunks[row_ptr.chunk_index as usize], &[
                    row_ptr.row_index,
                ])?;
                if !block.is_empty() {
                    data_blocks.push(block);
                }
            }
```
The debug result of your case
```
[query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs:99] input =
[
    "+-------+-------+",
    "| \"c\"_0 | \"d\"_1 |",
    "+-------+-------+",
    "| 1     | 2     |",
    "+-------+-------+",
]

[query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs:125] result_blocks.clone() = [

    [
        "+-------+-------+",
        "| \"b\"_3 | \"a\"_2 |",
        "+-------+-------+",
        "| 2     | 1     |",
        "+-------+-------+",
    ]
    ,

    [
        "+-------+-------+",
        "| \"b\"_3 | \"a\"_2 |",
        "+-------+-------+",
        "| 3     | 1     |",
        "+-------+-------+",
    ]
    ,

    [
        "+-------+-------+",
        "| \"b\"_3 | \"a\"_2 |",
        "+-------+-------+",
        "| 4     | 1     |",
        "+-------+-------+",
    ]
    ,
]
[query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs:127] result_block.clone() =
[
    "+-------+-------+",
    "| \"b\"_3 | \"a\"_2 |",
    "+-------+-------+",
    "| 2     | 1     |",
    "+-------+-------+",
]

[query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs:127] result_block.clone() =
[
    "+-------+-------+",
    "| \"b\"_3 | \"a\"_2 |",
    "+-------+-------+",
    "| 3     | 1     |",
    "+-------+-------+",
]

[query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs:127] result_block.clone() =
[
    "+-------+-------+",
    "| \"b\"_3 | \"a\"_2 |",
    "+-------+-------+",
    "| 4     | 1     |",
    "+-------+-------+",
]
```

PR #5510, opened 2022-05-22
feature(planner): Translate subquery into apply operator

leiysky:

## Summary

Translate subqueries into apply operators, so we can represent all the subqueries(correlated or uncorrelated) with relational algebra.

Besides, this PR also refines the property derivation mechanism with `RelExpr`.

## Changelog

- New Feature

## Related Issues

Part of #5210

### DIFF
--- a/query/src/sql/planner/mod.rs
+++ b/query/src/sql/planner/mod.rs
@@ -13,11 +13,11 @@
 // limitations under the License.
 
 use std::sync::Arc;
-use std::sync::RwLock;
 
 use common_ast::parser::error::Backtrace;
 use common_ast::parser::parse_sql;
 use common_ast::parser::tokenize_sql;
+use common_base::infallible::RwLock;
 use common_exception::ErrorCode;
 use common_exception::Result;
 pub use plans::ScalarExpr;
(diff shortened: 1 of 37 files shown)


query/src/sql/planner/mod.rs
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::sync::Arc;
+use std::sync::RwLock;
```
BohuTANG, line 16: It's better to use the common/base/core/infallible rwlock https://github.com/datafuselabs/databend/blob/main/common/base/src/infallible/rwlock.rs , so we no need do any unwrap.

leiysky: That's good. 

By the way, I'd like to know if it is the canonical `Mutex` of databend?

BohuTANG: Infallible rwlock is a simple wrapper around the lock() function of a std::sync::RwLock The only difference is that you don’t need to call unwrap() on it.

Infallible Mutex https://github.com/datafuselabs/databend/blob/main/common/base/src/infallible/mutex.rs

leiysky: Fixed.

BohuTANG (approved):
:100:

BohuTANG:
Let's merge it first, the review remains open.

PR #5512, opened 2022-05-23
feature(planner): Common tree structure formatter for plan display

### LINKED ISSUE #5491: Feature: Format `SExpr` into string
To implement `EXPLAIN`, we need a mechanism to format a plan into string.

I propose a tree-like structure to store the formated string of each plan node, and provide serialization interface to satisfy different need.

leiysky:

## Summary

Introduce a common formatting mechanism for tree stucture, so we can exploit it to imlement `EXPLAIN` and so on.

A indent format is implemented, here's an example:
```
PhysicalHashJoin: build keys: [plus(col1, 123)], probe keys: [col2]
    Filter: [true]
        PhysicalScan: catalog.database.table
    PhysicalScan: catalog.database.table
```

## Changelog

- New Feature

## Related Issues

Close #5491

### DIFF
--- a/query/src/sql/exec/mod.rs
+++ b/query/src/sql/exec/mod.rs
@@ -31,7 +31,7 @@ use common_planners::RewriteHelper;
 pub use util::decode_field_name;
 pub use util::format_field_name;
 
-use super::plans::Operator;
+use super::plans::RelOperator;
 use super::MetadataRef;
 use crate::pipelines::new::processors::port::InputPort;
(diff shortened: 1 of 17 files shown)


BohuTANG:
It would be better to add a stateless test for the new `explain`.

leiysky:
> It would be better to add a stateless test for the new `explain`.

`EXPLAIN` isn't involved in this PR. @xudong963 will support it later.

xudong963 (approved):

PR #5539, opened 2022-05-23
bugfix(executor): Fix incorrect context usage

### LINKED ISSUE #5522: Bug(planner): wrong output in hash join
The result is not correct in my case:
```
create table t1(a int, b int);
create table t2(a int, b int);
create table t3(a int, b int);

insert into t1 select number, number+1 from numbers(100000000);
insert into t2 select number, number+1 from numbers(20000000);
insert into t3 select number, number+1 from numbers(10000000);

select count(*) from t1 inner join t2 on t1.a=t2.a inner join t3 on t2.b=t3.b;

+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
1 row in set (6.45 sec)
```

Result should be:
```
┌──count()─┐
│ 10000000 │
└──────────┘
```

_Originally posted by @BohuTANG in https://github.com/datafuselabs/databend/issues/5497#issuecomment-1133793121_

leiysky:

## Summary

Summary about this PR

## Changelog

- Bug Fix

## Related Issues

Close #5522


andylokandy (approved):

query/src/sql/exec/mod.rs
```diff
@@ -161,7 +161,7 @@ impl PipelineBuilder {
 
         match plan {
             RelOperator::PhysicalScan(physical_scan) => {
-                self.build_physical_scan(physical_scan, pipeline)
+                self.build_physical_scan(context, physical_scan, pipeline)
```
zhang2014, line 164: We should also check `build_project`, `build_eval_scalar`, `build_filter`, `build_aggregate`, `build_order_by`, `build_limit`.

zhang2014 (changes requested):
.

query/src/sql/exec/mod.rs
leiysky: Fixed.

PR #5541, opened 2022-05-24
Feature: Support TRIM function in new planner

### LINKED ISSUE #5367: Feature: Support `TRIM` function in new planner

ygf11:

## Summary

Support:
1. `trim(leading 'c' from 'cbc')`.
2. `trim(trailing 'c' from 'cbc')`.
3. `trim(both 'c' from 'cbc')`.
4. `trim(' abc ')`, which is equivalent to `trim(both ' ' from ' abc ')`.

## Changelog

- New Feature

## Related Issues

Fixes #5367


common/functions/src/scalars/strings/string.rs
```diff
@@ -68,6 +71,9 @@ impl StringFunction {
         factory.register("from_base64", Base64DecodeFunction::desc());
         factory.register("rtrim", RTrimFunction::desc());
         factory.register("trim", TrimFunction::desc());
+        factory.register("trim_leading", TrimLeadingFunction::desc());
```
sundy-li, line 74: what's the difference between `trim_leading` and `ltrim`.

sundy-li (approved):

common/functions/src/scalars/strings/string.rs
sundy-li: I get it

leiysky (approved):

BohuTANG:
@mergify update

PR #5557, opened 2022-05-24
fix(scripts):deploy meta-service failed because of no pv definition

hantmac:
Signed-off-by: hantmac <hantmac@outlook.com>

## Summary

- fix(scripts):deploy meta-service failed because of no pv definition

## Changelog

- Bug Fix: deploy meta-service failed because of no pv definition


scripts/kubernetes/meta-standalone.yaml
```diff
@@ -1,4 +1,19 @@
...
+  labels:
+    type: local
+spec:
+  storageClassName: local-storage
+  capacity:
+    storage: 10Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: "/mnt/data"
```
Xuanwo, line 15: Is it safe to share the same host path with other pods?

hantmac: > Is it safe to share the same host path with other pods?

I'm afraid not safe, change it to another maybe better.

Xuanwo: I have noticed that other services have similar problems. Are you willing to change them at the same time?

For naming, I prefer:

- meta: `/mnt/databend/meta`
- minio: `/mnt/databend/minio`
- component_name: `/mnt/databend/component_name`

Does this look good to you?

hantmac: > I have noticed that other services have similar problems. Are you willing to change them at the same time?
> 
> For naming, I prefer:
> 
> * meta: `/mnt/databend/meta`
> * minio: `/mnt/databend/minio`
> * component_name: `/mnt/databend/component_name`
> 
> Does this look good to you?

Yeah,mostly LGMT. BTW, where is the `component_name ` service? Or what is the `component_name ` mean?

PR #5570, opened 2022-05-25
[meta/client] let MetaGrpcClient run in a dedicated Runtime

drmingdrmer:

##### [meta/client] let MetaGrpcClient run in a dedicated Runtime
As @zhang2014 suggested a solution to the cross-runtime blocking issue,
by isolating grpc client in a single tokio-runtime:

Create a dedicated `Runtime` to run meta-client `worker`, every time
MetaGrpcClient is created. And a `handle` is returned to its caller to
communicate to the `worker`.

So that all background tasks spawned by `tower` or `hyper` will run in
the worker runtime and won't be affected by other tasks that may block a
tokio-task.

TODO:

- [ ] This way it's recommended to create just one singleton instance of
  MetaGrpcClient in one process. To avoid creating a runtime too often.

Changes:

- `MetaGrpcClient::try_new()` returns a `ClientHandle` instead of a
  `MetaGrpcClient`. `ClientHandle` impl `KVApi` and can be used the same
  way as before.

- Revert to using conn pool.

- `ClientHandle::make_client()` can still be used to retrieve an
  internal Grpc-client, which is supposed to run in the worker-runtime.

  Using it outside may leaks background tasks to other runtime and should be avoid.

- Add `watch()` and `export()` to `ClientHandle`.

Refactors:

- Add `#[track_caller]` to methods of `Runtime`, to track the caller when
  tracing.

- Add thread name to customized tracing event formatter.

## Related Issues


lichuang (approved):

common/meta/raft-store/src/state_machine/sm_kv_api_impl.rs
```diff
@@ -21,15 +21,15 @@ use common_meta_types::MetaError;
...
+use common_meta_types::UpsertKVReply;
 use common_meta_types::UpsertKVReq;
 use common_tracing::tracing;
 
 use crate::state_machine::StateMachine;
 
 #[async_trait::async_trait]
 impl KVApi for StateMachine {
-    async fn upsert_kv(&self, act: UpsertKVReq) -> Result<UpsertKVActionReply, MetaError> {
+    async fn upsert_kv(&self, act: UpsertKVReq) -> Result<UpsertKVReply, MetaError> {
```
BohuTANG, line 32: Does this rename will break the computable between metasrv and query?

drmingdrmer: No. It is just a type alias and we have `tests/compat/test-compat` to ensure an old query talks happily to the latest metasrv.

PR #5588, opened 2022-05-25
fix(query): Deny the root login from others host

### LINKED ISSUE #5555: Deny the root login from others host(not localhost).
**Summary**

Let's deny the root login from another host(not localhost).

chowc:

## Summary

I change `query/src/servers/mysql/mysql_interactive_worker.rs`  so that mysql worker and clickhouse worker both use `auth_mgr.rs:auth` to do auth and this mades it easier to add root login check in `auth_mgr.rs`;

## Changelog

- Bug Fix

## Related Issues

Fixes #5555


query/src/users/auth/auth_mgr.rs
```diff
@@ -33,6 +33,7 @@ pub struct AuthMgr {
 pub enum Credential {
     Jwt {
         token: String,
+        hostname: Option<String>,
```
junnplus, line 36: Why JWT also needs hostname?

query/src/servers/mysql/mysql_interactive_worker.rs
```diff
@@ -232,16 +233,33 @@ impl<W: std::io::Write> InteractiveWorkerBase<W> {
...
-            self.session.set_current_user(user_info);
-        }
-        Ok(authed)
+        let auth_manager = ctx.get_auth_manager();
+        let credential = Credential::Password {
+            name: user_name.to_string(),
+            password: Some(salt.to_vec()),
+            hostname: Some(client_ip.to_string()),
+        };
+        return match auth_manager.auth(&credential).await {
```
junnplus, line 242: Nice job!

query/src/users/auth/auth_mgr.rs
```diff
@@ -85,6 +98,15 @@ impl AuthMgr {
...
             } => {
+                let identity = UserIdentity {
+                    username: n.to_string(),
+                    hostname: h.as_ref().unwrap_or(&"%".to_string()).to_string(),
+                };
+                if identity.is_root() && !identity.is_localhost() {
+                    return Err(ErrorCode::AuthenticateFailure(
+                        "root user can only login from localhost",
+                    ));
+                }
```
junnplus, line 109: I think we should check it on https://github.com/datafuselabs/databend/blob/main/query/src/users/user_mgr.rs#L31-L41

Xuanwo:
I have updated the title to make it more readable. Feel free to update again to meet your needs.

common/meta/types/src/user_identity.rs
```diff
@@ -31,6 +31,10 @@ impl UserIdentity {
     pub fn is_localhost(&self) -> bool {
         &self.hostname.to_lowercase() == "localhost" || &self.hostname == "127.0.0.1"
     }
+
+    pub fn is_root(&self) -> bool {
+        self.username.eq_ignore_ascii_case("root")
```
youngsofun, line 36: should not ignore case I think.

query/src/users/auth/auth_mgr.rs
chowc: If use jwt to do auth, it seems we need `client_ip` to check it's from localhost or not, so I add hostname field.

query/src/users/auth/auth_mgr.rs
chowc: `user_mgr:get_user` method will also be called by `ShowGrantsInterpreter` and `DfAlterUser`, and these two seems to be irrelevant to auth logic so I put the host check in `auth_mgr`.

junnplus: Why not return unknown user directly in get_user?

query/src/servers/mysql/mysql_interactive_worker.rs
```diff
@@ -232,16 +233,33 @@ impl<W: std::io::Write> InteractiveWorkerBase<W> {
...
-
-        let authed = user_info.auth_info.auth_mysql(&info.user_password, salt)?;
-        if authed {
-            self.session.set_current_user(user_info);
-        }
-        Ok(authed)
+        let auth_manager = ctx.get_auth_manager();
+        let credential = Credential::Password {
+            name: user_name.to_string(),
+            password: Some(salt.to_vec()),
```
youngsofun, line 239: why assign salt to the password field?

query/src/users/auth/auth_mgr.rs
```diff
@@ -64,6 +68,15 @@ impl AuthMgr {
...
                     .unwrap_or_else(|| self.tenant.clone());
+                let identity = UserIdentity {
+                    username: user_name.to_string(),
+                    hostname: h.as_ref().unwrap_or(&"%".to_string()).to_string(),
+                };
+                if identity.is_root() && !identity.is_localhost() {
+                    return Err(ErrorCode::AuthenticateFailure(
+                        "root user can only login from localhost",
+                    ));
+                }
```
junnplus, line 79: We should try to avoid special logic, if user 'root'@'other_host' is not found, it will not be able to login

query/src/servers/mysql/mysql_interactive_worker.rs
youngsofun: adding a stateless  https://github.com/datafuselabs/databend/pull/5600
you can start servers using ./scripts/ci/deploy/databend-query-standalone.sh and test with mysql client first

BohuTANG:
@mergify update

common/meta/types/src/user_identity.rs
chowc: updated.

query/src/servers/mysql/mysql_interactive_worker.rs
chowc: `mysql_interactive_worker` seems to be using different auth algorithm from `auth_mgr` so I rollback the change here.

query/src/users/user_mgr.rs
```diff
@@ -36,6 +36,11 @@ impl UserApiProvider {
                         UserPrivilegeSet::available_privileges_on_global(),
                     );
                     user_info.option.set_all_flag();
+                } else if user.is_root() {
```
youngsofun, line 39: @BohuTANG should we accept `default@localhost` too for the convenience of clickhouse client? 

choices:

1.  accept,  `default@localhost` and `root@localhost` works as root for all handlers
2. deny `default@localhost`,  prompt the clickhouse user to use `root` 
3.  accept  `default@localhost`  only for clickhouse handler

query/src/users/user_mgr.rs
```diff
@@ -36,6 +36,11 @@ impl UserApiProvider {
...
+                    return Err(ErrorCode::UnknownUser(format!(
+                        "'{}'@'{}'",
```
youngsofun, line 41: 1.  better remove the  `"default" | "" | "root"`  code
2.  a little more error message like `only accept root from localhost ` ?

query/src/users/user_mgr.rs
BohuTANG: I think 2 or 3 is ok, which one is easy to do?

youngsofun: > which one is easy to do?

almost the same： just need a small refactor to pass an enum indicating the handler for the auth-related methods,   (for 2 only if we need `prompt the clickhouse user to use root` )

for 2 also need to deny create user called `default  `

BohuTANG: Got, looks 1 is the better way, we allow `default` and `root` users only from the local host.
BTW, no special considerations here, this is to restrict these default users(`default`/`root`) to local login only.

youngsofun: @chowc so what about change `is_root()` to include `default` and use it  for `if`

chowc: So we accept 'default' and 'root' username as root user and only allow them to login in localhost? Am i right? 
Also empty login name will be denied.

chowc: > @chowc so what about change `is_root()` to include `default` and use it for `if`

ok, I will update it.

youngsofun (approved):

BohuTANG:
Thanks for your contribution! @chowc

query/tests/it/users/user_mgr.rs
```diff
@@ -347,14 +349,16 @@ async fn test_user_manager_with_root_user() -> Result<()> {
...
+        let res = user_mgr
             .get_user(tenant, UserIdentity::new(username2, hostname3))
-            .await?;
-        assert_eq!(user.name, username2);
-        assert_eq!(user.hostname, hostname3);
-        assert!(user.grants.entries().is_empty());
+            .await;
+        assert!(res.is_err());
+        assert_eq!(
+            "Code: 2201, displayText = only accept root from localhost 'root'@'otherhost'.",
```
Xuanwo, line 359: This error message seems confusing.

How about changing them into:

```shell
"Code: 2201, displayText = only accept root from localhost, current: 'root'@'otherhost'."
```

PR #5608, opened 2022-05-26
fix(query): cluster keys take effect in cluster mode

### LINKED ISSUE #5473: bug: clustering key does not take effect in cluster mode
**Summary**

```
mysql> create table t(a int, b int) cluster by(b,a);
Read 0 rows, 0.00 B in 0.050 sec., 0 rows/sec., 0.00 B/sec.

mysql> insert into t values(4,4);
Read 1 rows, 8.00 B in 0.049 sec., 20.47 rows/sec., 163.78 B/sec.
```

```
...
"col_stats":{"1":{"min":{"Int64":4},"max":{"Int64":4},"null_count":0,"in_memory_size":4},"0":{"min":{"Int64":4},"max":{"Int64":4},"null_count":0,"in_memory_size":4}},"cluster_stats":null},
...
```

zhyass:

## Summary

Summary about this PR

## Changelog

- Bug Fix

## Related Issues

Fixes #5473


common/protos/proto/metadata.proto
```diff
@@ -155,7 +155,7 @@ message TableMeta {
   map<string, string> options = 5;
 
   // Keys to sort rows in table.
-  optional string order_keys = 9;
+  optional string cluster_keys = 9;
```
sundy-li, line 158: will this make meta incompatible?

sundy-li (approved):

common/protos/proto/metadata.proto
sundy-li: Maybe it's not time to be version compatible.

PR #5616, opened 2022-05-26
[metasrv] Return an effective connection when make_client

ariesdevil:

## Summary

Return an effective connection when make_client to avoid connect to a dropped meta node.

## Changelog

- Improvement

## Related Issues

Fixes #issue


common/meta/grpc/src/grpc_client.rs
```diff
@@ -95,20 +96,61 @@ impl MetaChannelManager {
...
-            .iter()
-            .map(|a| self.build_endpoint(a))
-            .collect();
-        let channel_eps = channel_eps?;
-        let ch = Channel::balance_list(channel_eps.into_iter());
-        Ok(ch)
+        let mut n = endpoints.len();
+        let addrs = RoundRobinAddr::new(endpoints.clone());
+        let item = loop {
+            if let Some(addr) = addrs.next().await {
```
lichuang, line 130: if all the addr is invalid, it may cause a dead loop?

common/grpc/src/dns_resolver.rs
```diff
@@ -153,7 +153,8 @@ impl ConnectionFactory {
         inner_connector.set_keepalive(None);
         inner_connector.enforce_http(false);
```
lichuang, line 155: add `innner.set_connect_timeout` may be more suitable

common/meta/grpc/src/grpc_client.rs
drmingdrmer: Right, a for-loop looks good enough

common/grpc/src/dns_resolver.rs
ariesdevil: timeout is set here https://github.com/datafuselabs/databend/pull/5616/files#diff-50b6563436352a0bf2e50ea49d02a3883b8a9478637144963bae177bd359d49cL197

common/meta/grpc/src/grpc_client.rs
ariesdevil: @lichuang if all addrs are invalid, will break loop at [here](https://github.com/datafuselabs/databend/pull/5616/files#diff-757d3d716b4424b8e86d968f8a68c94b6bc9cd938ac9ee3caa9440ef35bb7707R143) with last error.

drmingdrmer (commented):
This feature needs tests to ensure it works with unreachable endpoints and it returns an error if no endpoint is online.

lichuang (approved):

drmingdrmer (approved):

common/meta/grpc/src/grpc_client.rs
```diff
@@ -346,7 +340,37 @@ impl MetaGrpcClient {
...
 
-            self.conn_pool.get(&*eps).await?
+            let num_eps = eps.len();
+            let mut start = rand::thread_rng().gen_range(0..10);
+            let end = start + num_eps;
+            let item = loop {
+                if start == end {
+                    break Err(MetaError::InvalidConfig("endpoints is empty".to_string()));
+                } else {
+                    let addr = eps.get(start % num_eps).unwrap();
```
BohuTANG, line 350: What happens if the endpoints is empty?

common/meta/grpc/src/grpc_client.rs
```diff
@@ -346,7 +340,37 @@ impl MetaGrpcClient {
             let eps = self.endpoints.read().await;
             debug_assert!(!eps.is_empty());
 
-            self.conn_pool.get(&*eps).await?
+            let num_eps = eps.len();
+            let mut start = rand::thread_rng().gen_range(0..10);
```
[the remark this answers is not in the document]
lichuang: same, i think there is no need a random to loop the endpoints.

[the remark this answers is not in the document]
ariesdevil: Using random start to avoid pick up same endpoint everytime, and we just loop `endpoints.len()` times here to avoid repeating create connections all the time.

PR #5617, opened 2022-05-26
feat: data time travel "select at"

### LINKED ISSUE #5516: Feature: support basic `select at` statement
**Summary:**
  
  Enhance the select query, so that a time point can be specified, sth like this
  
  `select * from t at(offset => -2) as t where ...`
  
  https://docs.snowflake.com/en/sql-reference/constructs/at-before.html
  
  A basic idea of impl (not a strict requirement, any other functionality equivalent impls are ok)
  - enhance the corresponding parser components
     encode the timepoint specification (the `at(offset...)` parts) into `DfQueryStatement`
  - carries the timepoint spec all the way to `ToReadDataSourcePlan::read_plan_with_catalog`   
     - detects if the underlying table instance support time travel.
     - if it does, navigate there, and construct the `ReadDataSourcePlan` accordingly
     
     also see https://github.com/datafuselabs/databend/issues/5514

dantengsky:

## Summary

Basic `select at` syntax supports, **Legacy Parser ONLY**

`select * from t12_0004 at (snapshot => "98ec833d7806463ba71cfb6a13fa12c3");`

~~~
mysql> create table t12_0004(c int);
Read 0 rows, 0.00 B in 0.045 sec., 0 rows/sec., 0.00 B/sec.

mysql> insert into t12_0004 values(1),(2);
Read 2 rows, 8.00 B in 0.028 sec., 71.33 rows/sec., 285.32 B/sec.

mysql> insert into t12_0004 values(3);
Read 1 rows, 4.00 B in 0.029 sec., 34.72 rows/sec., 138.89 B/sec.

mysql> select * from t12_0004;
+------+
| c    |
+------+
|    1 |
|    2 |
|    3 |
+------+
3 rows in set (0.04 sec)
Read 3 rows, 12.00 B in 0.006 sec., 486.07 rows/sec., 1.90 KiB/sec.

mysql> select * from fuse_snapshot('default', 't12_0004');
+----------------------------------+-----------------------------------------------------+----------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
| snapshot_id                      | snapshot_location                                   | format_version | previous_snapshot_id             | segment_count | block_count | row_count | bytes_uncompressed | bytes_compressed |
+----------------------------------+-----------------------------------------------------+----------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
| 1e046191e45848b5b59e414811930d95 | 1/1583/_ss/1e046191e45848b5b59e414811930d95_v1.json |              1 | 98ec833d7806463ba71cfb6a13fa12c3 |             2 |           2 |         3 |                 12 |              372 |
| 98ec833d7806463ba71cfb6a13fa12c3 | 1/1583/_ss/98ec833d7806463ba71cfb6a13fa12c3_v1.json |              1 | NULL                             |             1 |           1 |         2 |                  8 |              188 |
+----------------------------------+-----------------------------------------------------+----------------+----------------------------------+---------------+-------------+-----------+--------------------+------------------+
2 rows in set (0.03 sec)
Read 2 rows, 366.00 B in 0.007 sec., 305.18 rows/sec., 54.54 KiB/sec.

mysql> select * from t12_0004 at (snapshot => "98ec833d7806463ba71cfb6a13fa12c3");
+------+
| c    |
+------+
|    1 |
|    2 |
+------+
2 rows in set (0.03 sec)
Read 2 rows, 8.00 B in 0.005 sec., 427.27 rows/sec., 1.67 KiB/sec.
~~~

## Changelog

- New Feature
- Documentation

## Related Issues

Fixes #5516

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6014,6 +6014,7 @@ dependencies = [
 [[package]]
 name = "sqlparser"
 version = "0.13.1-alpha.0"
+source = "git+https://github.com/datafuse-extras/sqlparser-rs?rev=7d74a3c#7d74a3c5fc78f8d0c0f7e4e7e14d7058f94d8841"
 dependencies = [
  "hashbrown 0.12.0",
  "log",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6014,7 +6014,6 @@ dependencies = [
 [[package]]
 name = "sqlparser"
 version = "0.13.1-alpha.0"
-source = "git+https://github.com/datafuse-extras/sqlparser-rs?rev=45ead90#45ead90755aa86d5ff3f0faaa6c112d057b3494a"
 dependencies = [
  "hashbrown 0.12.0",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,6 @@ members = [
     "tools/fuzz",
     "tools/metactl",
 ]
-exclude = ["patch/sqlparser-rs"]
 
 [profile.release]
 debug = 1
@@ -67,6 +66,3 @@ rustc-demangle = { opt-level = 3 }
 [patch.crates-io]
 parquet2 = { version = "0.12", optional = true, git = "https://github.com/datafuse-extras/parquet2", rev = "03c108b" }
 chrono = { git = "https://github.com/datafuse-extras/chrono", rev = "279f590" }
-
-[patch.'https://github.com/datafuse-extras/sqlparser-rs']
-sqlparser = { path = './patch/sqlparser-rs' }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ members = [
     "tools/fuzz",
     "tools/metactl",
 ]
+exclude = ["patch/sqlparser-rs"]
 
 [profile.release]
 debug = 1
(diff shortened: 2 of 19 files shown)


BohuTANG:
Do we plan to support
 `SELECT ...
FROM ...
  {
   AT (TIMESTAMP|)
  }
`
?

dantengsky:
> Do we plan to support `SELECT ... FROM ... { AT (TIMESTAMP|) } ` ?

yes,  storage layer is ready. 

https://github.com/datafuselabs/databend/blob/e164aa23933781abb757b0666b655461481dcdbb/query/src/storages/fuse/operations/navigate.rs#L32-L36

in this pr, syntax that is most easily to be implemented  and tested is chose :)

Xuanwo:
@dantengsky 's fork name (`fuse-query`) represent a history :1st_place_medal:

BohuTANG:
How about run this statement in the new planner?  If it not working we should adapter it in the new planner.

dantengsky:
> How about run this statement in the new planner? If it not working we should adapter it in the new planner.

How about trying it in another PR? seems it is much more involved than simple syntax adaption like `drop all`

BohuTANG (approved):

PR #5618, opened 2022-05-26
feat(planner): support position function

### LINKED ISSUE #5536: Feature: Support `POSITION` expression in new planner
The reference https://databend.rs/doc/reference/functions/string-functions/position

xudong963:

## Summary

as title

## Changelog

- New Feature

## Related Issues

Fixes #5536

### DIFF
--- a/query/src/sql/planner/semantic/type_check.rs
+++ b/query/src/sql/planner/semantic/type_check.rs
@@ -420,6 +420,12 @@ impl<'a> TypeChecker<'a> {
 
             Expr::Array { exprs, .. } => self.resolve_array(exprs).await,
 
+            Expr::Position {
+                substr_expr,
+                str_expr,
+                ..
+            } => self.resolve_position(substr_expr, str_expr).await,
+
             _ => Err(ErrorCode::UnImplement(format!(
                 "Unsupported expr: {:?}",
                 expr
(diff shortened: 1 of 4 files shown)


query/src/sql/planner/semantic/type_check.rs
```diff
@@ -899,4 +905,47 @@ impl<'a> TypeChecker<'a> {
...
+                value: index.clone(),
+                data_type: index.data_type(),
+            }
+            .into(),
+            index.data_type(),
+        ))
+    }
+
+    fn match_expr(&self, expr: &'_ Expr<'a>) -> Result<String> {
+        let expr_str = match expr {
```
xudong963, line 931: For the sql: `SELECT POSITION("xbar" IN "xbar0");`
`"xbar"` will be identified as `Expr::ColumnRef`, is this the correct result for parser? @andylokandy  If so, I need to handle the `ColumnRef` case here.

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -899,4 +905,47 @@ impl<'a> TypeChecker<'a> {
             data_type,
         ))
     }
+
+    async fn resolve_position(
+        &self,
+        substr_expr: &'_ Expr<'a>,
+        str_expr: &'_ Expr<'a>,
```
leiysky, line 912: Maybe it's possible to remove the lifetime annotation `'_`?

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -899,4 +905,47 @@ impl<'a> TypeChecker<'a> {
...
+            .into(),
+            index.data_type(),
+        ))
+    }
+
+    fn match_expr(&self, expr: &'_ Expr<'a>) -> Result<String> {
+        let expr_str = match expr {
+            Expr::Literal { lit, .. } => {
+                let (value, data_type) = self.resolve_literal(lit, None)?;
+                if !data_type.eq(&DataTypeImpl::String(StringType {})) {
```
leiysky, line 934:
```suggestion
                if data_type != &StringType::new_impl() {
```

tests/suites/0_stateless/20+_others/20_0001_planner_v2.sql
```diff
@@ -240,4 +240,9 @@ select '====Correlated Subquery====';
 select * from numbers(10) as t where exists (select * from numbers(2) as t1 where t.number = t1.number);
 select (select number from numbers(10) as t1 where t.number = t1.number) from numbers(10) as t order by number;
 
+-- position function
+select '===Position Function===';
+SELECT POSITION('bar' IN 'foobarbar');
```
leiysky, line 245: Shall we allow it to take variable as argument? Like `position('bar' in column)`

xudong963: I checked the old planner, it supports the case.

query/src/sql/planner/semantic/type_check.rs
leiysky: You can replace double quote `"` with single quote `'`.

tests/suites/0_stateless/20+_others/20_0001_planner_v2.sql
andylokandy: This works in pg:

```SQL
drop table if EXISTS T;
create table t (a varchar, B varchar);
insert into t values ('FOO', 'barfoo');
select position(UPPER(A) in UPPER(B)) from T;
```

sundy-li (approved):

andylokandy (approved):

PR #5628, opened 2022-05-27
Feat: `select ... at` for new parser & planner

dantengsky:

## Summary

- new parser: parse `select ...  at ` 
- new planner : execute `select ... at`

## Changelog

- Improvement

## Related Issues

Fixes #issue


query/src/sql/planner/binder/mod.rs
```diff
@@ -18,9 +18,11 @@ pub use aggregate::AggregateInfo;
 pub use bind_context::BindContext;
 pub use bind_context::ColumnBinding;
 use common_ast::ast::Statement;
+use common_ast::ast::TimeTravelPoint;
 use common_datavalues::DataTypeImpl;
 use common_exception::ErrorCode;
 use common_exception::Result;
+use sqlparser::ast::Instant;
```
leiysky, line 25: Maybe it's better to introduce a internal `Instant` struct.

And is it possible to use an exists struct instead?

query/src/sql/planner/binder/mod.rs
```diff
@@ -114,10 +116,16 @@ impl<'a> Binder {
         catalog_name: &str,
         database_name: &str,
         table_name: &str,
+        travel_point: &Option<TimeTravelPoint>,
```
leiysky, line 119:
```suggestion
        travel_point: Option<&TimeTravelPoint>,
```

common/ast/src/parser/query.rs
```diff
@@ -153,11 +153,21 @@ pub fn aliased_table(i: Input) -> IResult<TableReference> {
...
+                travel_point,
             }
         },
     )(i)
 }
 
+pub fn travel_point(i: Input) -> IResult<TimeTravelPoint> {
+    map(
+        rule! {
+            AT ~  "(" ~ SNAPSHOT ~ "=>" ~ #literal_string ~ ")"
```
andylokandy, line 165:
```suggestion
            AT ~ "(" ~ SNAPSHOT ~ "=>" ~ #literal_string ~ ")"
```

andylokandy (approved):

query/src/sql/planner/binder/mod.rs
dantengsky: `sqlparser::ast::Instant` is an new struct that we introduced to `sqlparser-rs`

https://github.com/datafuse-extras/sqlparser-rs/blob/7d74a3c5fc78f8d0c0f7e4e7e14d7058f94d8841/src/ast/query.rs#L321-L325

Shall we introduce another intermediate data structure that  `TimeTravelPoint` and `sqlparser::ast::Instant` can both convert into?

query/src/sql/planner/binder/mod.rs
dantengsky: oops, there are compilation errors after this suggestion is applied. so I revert it.

travel point is captured here:
https://github.com/datafuselabs/databend/blob/e2a4d789b12805d5d50eda755828ca2c872c2251/common/ast/src/parser/query.rs#L139
Which is of type `Option<TimeTravelPoint>`.

did I missed something here?  

Or  `Option<&T>` style is recommended in new parser/planner?

query/src/sql/planner/binder/mod.rs
leiysky: It's good to have one. We will deprecate the sqlparser crate eventually.

dantengsky: OK

query/src/sql/planner/binder/mod.rs
leiysky: It's recommended, so you don't need to write a `&None`.

query/src/sql/planner/binder/mod.rs
dantengsky: pls have look at commit `Introduce NavigationPoint` [#5546572a36e36e03473f945fc31b555374c957dc](https://github.com/datafuselabs/databend/pull/5628/commits/5546572a36e36e03473f945fc31b555374c957dc)

leiysky (approved):

PR #5637, opened 2022-05-27
test(meta): check db's seq after create/drop/rename table

### LINKED ISSUE #5220: Test: create/drop/rename table has to increment the seq of db_meta
Add standalone tests to ensure that when create/drop/rename a table, the seq of the belonging db_meta will be increased.

Reference:
https://github.com/datafuselabs/databend/blob/eb07b393c769b152ef742c0b8f91032de5c98562/common/meta/api/src/schema_api_test_suite.rs#L946-L960

The tests should be added as part of `schema_api_test_suite.rs` and should be called somewhere else a `SchemaAPI` is implemented, such as 

https://github.com/datafuselabs/databend/blob/eb07b393c769b152ef742c0b8f91032de5c98562/metasrv/tests/it/grpc/metasrv_grpc_schema_api.rs#L104-L114

RinChanNOWWW:

## Summary

Check db's seq after create/drop/rename table, requiring the db's seq before changing is lower than the new one after changing.

## Changelog

- Build/Testing/CI

## Related Issues

Fixes #5220

### DIFF
--- a/common/meta/api/src/schema_api_test_suite.rs
+++ b/common/meta/api/src/schema_api_test_suite.rs
@@ -838,6 +838,15 @@ impl SchemaApiTestSuite {
         Ok(())
     }
 
+    fn req_get_db(tenant: impl ToString, db_name: impl ToString) -> GetDatabaseReq {
+        GetDatabaseReq {
+            inner: DatabaseNameIdent {
+                tenant: tenant.to_string(),
+                db_name: db_name.to_string(),
+            },
+        }
+    }
+
     pub async fn table_create_get_drop<MT: SchemaApi>(&self, mt: &MT) -> anyhow::Result<()> {
         let tenant = "tenant1";
         let db_name = "db1";
@@ -955,7 +964,10 @@ impl SchemaApiTestSuite {
         };
         let tb_ident_2 = {
             let tb_ident_2 = {
+                let old_db = mt.get_database(Self::req_get_db(tenant, db_name)).await?;
                 let res = mt.create_table(req.clone()).await?;
+                let cur_db = mt.get_database(Self::req_get_db(tenant, db_name)).await?;
+                assert!(old_db.ident.seq < cur_db.ident.seq);
                 assert!(res.table_id >= 1, "table id >= 1");
 
                 let tb_id = res.table_id;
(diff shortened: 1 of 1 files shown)


common/meta/api/src/schema_api_test_suite.rs
```diff
@@ -2465,7 +2596,16 @@ impl SchemaApiTestSuite {
                 },
             };
 
+            let get_db_req = GetDatabaseReq {
+                inner: DatabaseNameIdent {
+                    tenant: tenant.to_string(),
+                    db_name: db_name.to_string(),
+                },
+            };
```
drmingdrmer, line 2604: Consider creating a function for building a `GetDatabaseReq`, to reduce lines of codes.

Something like `fn reg_get_db(tenant: impl ToString, db_name: impl ToString) {...}`

common/meta/api/src/schema_api_test_suite.rs
```diff
@@ -2049,14 +2158,27 @@ impl SchemaApiTestSuite {
...
+
             let tb_ids = {
+                let old_db = mt.get_database(get_db_req.clone()).await?;
```
drmingdrmer, line 2169: This line does not need to be inside this block

drmingdrmer (commented):
Looks good to me.

Please try to minimize the lines of code.

common/meta/api/src/schema_api_test_suite.rs
RinChanNOWWW: How about adding a new function or macro like:

```rust
ASSERT_SEQ_INCR(mt.rename_table(..), mt, tenant1, db1, ...);
```

drmingdrmer: > How about adding a new function or macro like:
> 
> ```rust
> ASSERT_SEQ_INCR(mt.rename_table(..), tenant1, db1, ...);
> ```

It will encapsulate too many things in one function and make it unclear what it tries to assert.

RinChanNOWWW:
Codes upated. But I'm wondering what is the best place to put `req_get_db`.

drmingdrmer:
> Codes upated. But I'm wondering what is the best place to put `req_get_db`.

I do not know either. I find it a difficult task to organize methods and files too.

In this case, putting the util functions at bottom of this file would be OK.

drmingdrmer (approved):
Looks good to me

PR #5638, opened 2022-05-27
feat(parser): implment string unescape

### LINKED ISSUE #5631: Parser cannot handle string literal with quote character correctly
If we are trying to parse a string literal with single quote character(e.g. `'I'm a human'`), we have to escape the single quote with two consecutive single quote. For example, `'I''m a human'` will be displayed as `I'm a human`.

But in databend, it will produce a wierd result
```sql
MySQL [(none)]> select 'I''m a human';
+----------------+
| 'I''m a human' |
+----------------+
| I''m a human   |
+----------------+
1 row in set (0.030 sec)

```

andylokandy:

## Summary

- Support string literal unescape

## Changelog

- New Feature
- Bug Fix

## Related Issues

Fixes #5631


common/ast/src/parser/unescape.rs
```diff
@@ -0,0 +1,80 @@
+use std::char;
```
leiysky, line 1: Seems you need to add a license header.

Issue #5627, opened 2022-05-27
test-compat should only run cases for the old-query


drmingdrmer:
> Seems this PR break the compat test:
> 
> ```
> The failure tests:
>     /workspace/tests/suites/0_stateless/05_ddl/05_0000_ddl_drop_tables.sql
> 
> --- /workspace/tests/suites/0_stateless/05_ddl/05_0000_ddl_drop_tables.result	2022-05-26 23:47:40.000806565 +0000
> +++ /workspace/tests/suites/0_stateless/05_ddl/05_0000_ddl_drop_tables.stdout	2022-05-26 23:50:47.354907500 +0000
> @@ -0,0 +1,3 @@
> +ERROR 1105 (HY000) at line 12: Code: 1005, displayText = sql parser error: Expected end of statement, found: ALL.
> +ERROR 1105 (HY000) at line 14: Code: 2302, displayText = Table 't' already exists.
> +ERROR 1105 (HY000) at line 15: Code: 1005, displayText = sql parser error: Expected end of statement, found: ALL.
> ```
> 
> How to deal with this case: The test file is for the new version query, and the old-version query to do the compact test? cc @drmingdrmer

Hmm...test-compat should run the test cases only valid for the old query...

Let me find a way to filter them out

_Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/issues/5619#issuecomment-1139263115_

PR #5520, opened 2022-05-23
[ci] add compatibility test with old metasrv and latest query

drmingdrmer:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

##### [ci] add compatibility test with old metasrv and latest query

- fix: #4949

## Related Issues

PR #5645, opened 2022-05-28
[metasrv] feature: exchange protocol version with client

drmingdrmer:

##### [metasrv] feature: exchange protocol version with client
Before transfer any data, `MetaGrpcClient` must ensure it is compatible
with the metasrv it is connecting to, in the handshake RPC.

Add compatibility cross test:

```
latest query -- latest metasrv
old    query -- latest metasrv
latest query -- old    metasrv
```

Features

- Compatibility test: download test suite for query.

  When testing an old query with latest metasrv,
  download the test suite for that version of query instead of using
  the latest suite, which may not pass if new feature test is added
  into the latest suite but not yet impl-ed in the old query.

- Compatibility test:

  Test the compatibility between:
  - The latest query and the latest metasrv.
  - The oldest compatible query with latest metasrv.
  - The latest query with oldest compatible  metasrv.

- Protocol version exchange when meta-client `handshake` with metasrv.

  meta-client(query) keeps track of the min-metasrv version it is compatible
  with.

  metasrv keeps track of the min-meta-client version it is compatible
  with.

  Client handshake check these versions to ensure compatibility.
  Otherwise an error is returned and no further data will be transferred.

- Add admin CLI command `--cmd ver`.

  `databend-query --cmd ver` outputs its build version and the minimal
  compatible metasrv version, such as:

  ```
  version: 0.7.61-nightly
  min-compatible-metasrv-version: 0.7.59
  ```

  `databend-meta --cmd ver` outputs its build version and the minimal
  compatible meta-client(query) version, such as:

  ```
  version: 0.7.61-nightly
  min-compatible-client-version: 0.7.57
  ```

Refactors

- Add a new error `InvalidArgument` to indicate protocol-error

- `MetaGrpcClient::try_new()` does not need to be `async`.

- Metasrv config does not `check()` if `--cmd` is non-empty:

  When `cmd` is specified, it does not mean to run it.

Docs

- Add doc explain how to check versions and how to upgrade.

---

- Fix: #5627
- Fix: #4948

## Changelog

- New Feature

## Related Issues


BohuTANG:
Cool.
If we build from source with out git information, what happenes during metasrv and query?

lichuang (approved):

drmingdrmer:
> Cool. If we build from source with out git information, what happenes during metasrv and query?

The commit-version will be missing and the query won't be able to connect to metasrv.

BohuTANG (approved):

drmingdrmer:
> Cool. If we build from source with out git information, what happenes during metasrv and query?

BTW I've once considered using the version defined in `Cargo.toml` for building a release.

```
[package]
name = "databend-query"
version = "0.7.0"
```

E.g., if the version in Cargo is `0.7.0`, then a nightly build version could be `0.7.0+20220528`.

Then the git-tag triggered release action only is used for formal releases.

BohuTANG:
> > Cool. If we build from source with out git information, what happenes during metasrv and query?
> 
> The commit-version will be missing and the query won't be able to connect to metasrv.

I think this is OK.

drmingdrmer:
> > > Cool. If we build from source with out git information, what happenes during metasrv and query?
> > 
> > 
> > The commit-version will be missing and the query won't be able to connect to metasrv.
> 
> I think this is OK.

Hmm... looks like CI failed due to this issue. In the CI it just checkout one last commit without any tags. The version can not be determined.

It can be worked around by using a special protocol-version == 0, which I do not like. 🤔

BohuTANG:
> > > > Cool. If we build from source with out git information, what happenes during metasrv and query?
> > > 
> > > 
> > > The commit-version will be missing and the query won't be able to connect to metasrv.
> > 
> > 
> > I think this is OK.
> 
> Hmm... looks like CI failed due to this issue. In the CI it just checkout one last commit without any tags. The version can not be determined.
> 
> It can be worked around by using a special protocol-version == 0, which I do not like. 🤔

@everpcpc fixed it, may help here

.github/actions/test_compat/action.yml
```diff
@@ -15,6 +15,10 @@ runs:
     - name: Maximize build space
       uses: ./.github/actions/cleanup
 
+    - name: Rust setup
```
everpcpc, line 18: `dev_setup` is only needed on macos.
on linux ci,  `setup_build_tool` already made it working.

everpcpc:
> > > > > Cool. If we build from source with out git information, what happenes during metasrv and query?
> > > > 
> > > > 
> > > > The commit-version will be missing and the query won't be able to connect to metasrv.
> > > 
> > > 
> > > I think this is OK.
> > 
> > 
> > Hmm... looks like CI failed due to this issue. In the CI it just checkout one last commit without any tags. The version can not be determined.
> > It can be worked around by using a special protocol-version == 0, which I do not like. 🤔
> 
> @everpcpc fixed it, may help here

should be fixed with: https://github.com/datafuselabs/databend/pull/5520/files#r884142502

.github/actions/test_compat/action.yml
drmingdrmer: > > @everpcpc fixed it, may help here
> 
> should be fixed with: https://github.com/datafuselabs/databend/pull/5520/files#r884142502

Is it appropriate to fetch the full history? 
The fetching time would be longer then.

drmingdrmer:
> should be fixed with: https://github.com/datafuselabs/databend/pull/5520/files#r884142502

Is it appropriate to fetch the full history?
The fetching time would be longer then.

.github/actions/test_compat/action.yml
```diff
@@ -15,6 +15,10 @@ runs:
...
+      run: |
+        bash ./scripts/setup/dev_setup.sh -yb
+
     - name: Setup Build Tool
       uses: ./.github/actions/setup_build_tool
```
everpcpc, line 24: installing python libs is also not need, and already done by build tool image

.github/actions/test_compat/action.yml
drmingdrmer: It complains in `test-compat.sh` that `wget` is not found.
What did I do wrongly?

I was trying to run it with build-tool:
`build-tool bash ./tests/compat/test-compat.sh`

`wget` should be already provided by github-action.

everpcpc:
> 

we are already doing this with release build, and it seems checkout takes around 10s which seems fine.

everpcpc:
> > should be fixed with: https://github.com/datafuselabs/databend/pull/5520/files#r884142502
> 
> Is it appropriate to fetch the full history? The fetching time would be longer then.

we are already doing this with release build, and the checkout takes around 10s which seems fine.

.github/actions/test_compat/action.yml
everpcpc: ah, it's not in the build-tool image now. 
maybe you can use curl instead.

drmingdrmer: That's all right! Thanks!

PR #5646, opened 2022-05-28
feat(query): add call stats functions

everpcpc:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues

Fixes #issue


BohuTANG:
Oops, `.github/actions/check/action.yml` Conflicting

everpcpc:
ask @drmingdrmer for advice, if there are any needs to limit request to meta?

drmingdrmer:
> ask @drmingdrmer for advice, are there any needs to limit request to meta?

Yes, if it's possible. The estimated max throughput is about 30,000 writes per second for now.

everpcpc:
> Yes, if it's possible. The estimated max throughput is about 30,000 writes per second for now.

The stats call is just read, no write at all. Would that make any difference?

drmingdrmer:
> > Yes, if it's possible. The estimated max throughput is about 30,000 writes per second for now.
> 
> The stats call is just read, no write at all. Would that make any difference?

A read does not need to replicate a raft-log or to persist anything on a disk. It would be cheaper.

If the total RPS is under 5,000 there should not be a concern.

everpcpc:
> > > Yes, if it's possible. The estimated max throughput is about 30,000 writes per second for now.
> > 
> > 
> > The stats call is just read, no write at all. Would that make any difference?
> 
> A read does not need to replicate a raft-log or to persist anything on a disk. It would be cheaper.
> 
> If the total RPS is under 5,000 there should not be a concern.

thx~
currently the `list_tables` operation would result in a RPS of total table count, but we only need the table count, would a `list_id_value` or `list_keys` call different from common read calls?
Moreover, would it be ok for meta to provide some api for stats with grouping prefix (such as tenant id) ?

drmingdrmer:
> Would a `list_id_value` or `list_keys` call different from common read calls? 

`list_id_value` can be implemented with a `prefix-list` RPC to metasrv.
`list_keys` can not be expressed with KVApi yet.

> Moreover, would it be ok for meta to provide some api for stats grouped by prefix (such as tenant id) ?

metasrv does not know what the delimiter is for an application, although traditionally it is a slash `/`.
And maintaining the counts of every prefix will result in unnecessary write-amplification.

Thus IMHO, `count by the prefix` should be implemented on the application side:

When creating/dropping/undropping/renaming a table, also update a counter record.

Issue #5660, opened 2022-05-29
Formal specification of `HAVING` clause


leiysky:
## Specification of `HAVING` clause
According to the specification of `HAVING` in SQL standard, a `HAVING` clause will perform filtering on grouped tables.

There are three situations:
- `SELECT ... FROM ... HAVING ...`. This `HAVING` will treat the whole table in `FROM` clause as a single group.
- `SELECT ... FROM ... WHERE ... HAVING ...`. This `HAVING` will take the data filtered by `WHERE` as input, and treat it as a single group.
- `SELECT ... FROM ... GROUP BY ... HAVING ... `. This `HAVING` will filter the groups produced by `GROUP BY` with specified condition.

~~`HAVING` must work in aggregate context, that is, at least one aggregate function is needed to be defined. Otherwise we will raise an error for it. For example:~~
```sql
select a from t having a = 1; -- Error

select count(a) from t having count(a) = 1; -- Ok
```

The usage of a `HAVING` clause is much similar to `WHERE`, the difference is:
- `WHERE` clause is executed before grouping, while `HAVING` clause is executed after grouping
- It is allowed to use an aggregate function as condition in `HAVING` clause, which is impossible in a `WHERE` clause

In most scenarios, put the condition into `WHERE` clause is much more efficient than using `HAVING`. So the the only situation that we have to use a `HAVING` clause is that we have aggregate functions in filter condition.

The restriction of columns in `HAVING` clause is the same as `SELECT` clause in a grouping context:
- Column reference must be a single value for per group, which means non-grouping columns cannot appear outside aggregate functions
- Columns must be unambiguous

PR #5662, opened 2022-05-29
improve(planner): Support use alias group item in `HAVING`

leiysky:

## Summary

Support use alias group item in having clause, for example:
```sql
select a+1 as b from t group by b having b = 1;
```

## Changelog

- Improvement

Related to #5660


leiysky:
@mergify update

andylokandy (approved):

PR #5665, opened 2022-05-30
RFC-5665: User Friendly Releases

PsiACE:

## Summary

Releasing packages for different systems is unnecessary and can get trapped in other complications.

After a brief survey, we do not need to substantially change the existing release process. Just some additions needed.

[rendered](https://github.com/PsiACE/databend/blob/rfc-release/docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md)

cc @Xuanwo @wubx @soyeric128 for discussion

## Changelog

## Related Issues

Related #5563

### DIFF
--- a/docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
+++ b/docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
@@ -1,0 +1,71 @@
+---
+title: User-friendly Releases
+description: RFC for user-friendly releases
+---
+
+# Summary
+
+A simple user-friendly release solution.
+
+# Motivation
+
+In the past, only specific binaries were included in the Databend release. Understanding how to operate and maintain Databend often requires extensive discussion.
+
+This proposal aims to improve this process by providing some simple tips and default configs to help make the process painless for users.
+
+# Detailed design
+
+The release directory structure is roughly as follows:
+
+```rust
+.
+├── bin
+│   ├── databend-query
+│   ├── databend-meta
+│   └── databend-metactl (?)
+├── configs
+│   ├── databend-query.toml
+│   └── databend-meta.toml
+├── docs 
+│   ├── basic-guide.md
+│   ├── faq.md
+│   └── mini-tutorial.md (?)
+├── data/ (?)
+├── scripts/ (?)
+│   ├── bootstrap.sh
+│   └── benchmark.sh
+├── Makefile (?)
+└── README.md
+
+- `(?)` means optional.
+```
+
+## Required
+
+- `README.md`, basic information and important tips, useful links.
+- `bin`, of course, our binaries.
+- `configs`, the basic configs, to prompt for configuration usage and start the Databend service.
+- `docs`, helpful documents, frequently asked questions and basic user guide.
+
+## Optional
+
+- `databend-metactl`, perhaps we no longer need to release databend-tools.
+- `docs/mini-tutorial.md`, like a worksheet, it will help to learn how to use Databend.
+- `data`, if we have a `mini-tutorial` in release, here is a copy of the data.
+- `scripts`, for easy installation and deployment.
+- `Makefile`, one-line commands, `install`, `run` and more.
+
+# Rationale and alternatives
+
+## Why not use a directory structure closer to a package like `/etc`, `/usr` ?
+
+Just for a quick glance, users can find the content they care about very intuitively.
+
+It would also be easy to switch to a more package-like directory structure, but we don't have enough motivation to do so at the moment.
+
+## What are other good practices?
+
+The releases of most open source databases are no different with us. However, some good examples can still be found.
+
+- [`clickhouse`](https://github.com/ClickHouse/ClickHouse/releases), package-like directory structure, includes configs and utilities, the documentation is mainly a collection of useful links.
+- [`cayley`](https://github.com/cayleygraph/cayley/releases/), a notable highlight is the inclusion of data and queries to help users quickly explore linked data.
--- a/docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
+++ b/docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
@@ -26,34 +26,24 @@ The release directory structure is roughly as follows:
 ├── configs
 │   ├── databend-query.toml
 │   └── databend-meta.toml
-├── docs 
-│   ├── basic-guide.md
-│   ├── faq.md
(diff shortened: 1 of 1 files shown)


wubx:
change configs to etc?

BohuTANG:
> change configs to etc?

I don't think we need it, let's the user put it anywhere they want.

PsiACE:
> change configs to etc?

The usual practice might be to put it in `etc/configs`, but that doesn't seem necessary unless we need to add something else, such as `init.d`.

BohuTANG:
The default releases are for the user who wants to experience the databend, so we should not make it complex here.

> > change configs to etc?
> 
> The usual practice might be to put it in `etc/configs`, but that doesn't seem necessary unless we need to add something else, such as `init.d`.

The default releases are only for the user who wants to experience the databend locally, so we should not make it complex. here.

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
```diff
@@ -0,0 +1,71 @@
...
+```rust
+.
+├── bin
+│   ├── databend-query
+│   ├── databend-meta
+│   └── databend-metactl (?)
+├── configs
+│   ├── databend-query.toml
+│   └── databend-meta.toml
+├── docs
```
Xuanwo, line 29: Doesn't make sense to me.

- It's hard to read locally.
- It's hard to update and get a quick response (compared to databend.rs)
- It's hard to maintian.

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
```diff
@@ -0,0 +1,71 @@
...
+│   ├── databend-meta
+│   └── databend-metactl (?)
+├── configs
+│   ├── databend-query.toml
+│   └── databend-meta.toml
+├── docs 
+│   ├── basic-guide.md
+│   ├── faq.md
+│   └── mini-tutorial.md (?)
+├── data/ (?)
```
Xuanwo, line 33: Putting data in releases is not a good idea for me.

Even if we put them in data, users without tutorials will not know how to use them. But if we have tutorials, why not let them download directly...

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
```diff
@@ -0,0 +1,71 @@
...
+├── scripts/ (?)
+│   ├── bootstrap.sh
+│   └── benchmark.sh
+├── Makefile (?)
```
Xuanwo, line 37: Users could be confused to have a `Makefile` in releases.

soyeric128:
> The default releases are for the user who wants to experience the databend, so we should not make it complex here.
> 
> > > change configs to etc?
> > 
> > 
> > The usual practice might be to put it in `etc/configs`, but that doesn't seem necessary unless we need to add something else, such as `init.d`.
> 
> The default releases are only for the user who wants to experience the databend locally, so we should not make it complex here.

Im a little confused here. Do you have other releases (except the default ones) developed for production environments?

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
PsiACE: > It's hard to read locally.

We have plans to add some comments to make it readable. like clickhouse, but simpler.

> It's hard to update and get a quick response (compared to databend.rs).

Yep, but it seems a little more difficult for the user to write from scratch.

> It's hard to maintian.

This completely reuses [the configs we use for testing](https://github.com/datafuselabs/databend/tree/main/scripts/ci/deploy/config). In fact, I don't think the maintenance problem is caused by this proposal. We have always lacked a more automated solution for document/config generation.

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
PsiACE: I agree with you, I will remove it.

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
Xuanwo: Hi, I'm commenting on `docs`.

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
PsiACE: It's actually more like some actual runnable query examples.

It appears in  [cayley](https://github.com/cayleygraph/cayley/releases/) and is listed in the proposal only because it looks interesting. For Databend, there may indeed be no value.

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
soyeric128: I think it would be very helpful if the release package contains both the toml files with detailed explanations for each parameter in each. So users can change the files accordingly when deploying instead of starting from the ground up. 

I prefer the way Snowflake does for the doc part. Can we add the doc links in the readme file? then we don't need additional tutorials for the package. We update the docs only on our website.

PsiACE: > Hi, I'm commenting on `docs`.

Oh, sorry. I'm really not sure what to do with Docs, and markdown is definitely not a good idea. So let's cut it down and just provide some useful links in txt format. How do you feel about it?

PS: github should give me a comment highlight, I did miss it.

PsiACE: > Can we add the doc links in the readme file?

I agree. So we can just keep a readme file as doc in release.

PsiACE:
> Im a little confused here. Do you have other releases (except the default ones) developed for production environments?

Now we only have nightly releases. We may later have a stable channel for production environments. https://databend.rs/doc/releases/release-channels

docs/doc/60-contributing/03-rfcs/20220530-user-friendly-releases.md
PsiACE: I have removed the description of `data/`.

soyeric128:
Now it looks good to me for the doc part. So after this structure is approved, shall we add the parameter descriptions to the toml files?

Xuanwo (approved):
LGTM, let's create a tracking issue for this RFC!

PR #5689, opened 2022-05-30
feature(executor): Introduce `ScalarEvaluator` to evaluate `Scalar`

leiysky:

## Summary

Introduce a reusable `ScalarEvaluator` to evaluate `Scalar` with a given context.

This is part of refator of execution framework.

## Changelog

- New Feature


query/src/common/evaluator/eval_context.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn insert_vector(&mut self, id: Self::VectorID, vector: TypedVector) -> Result<()>;
+
+    fn tuple_count(&self) -> usize;
+}
+
+impl EvalContext for DataBlock {
+    type VectorID = IndexType;
+
+    fn get_vector(&self, id: &IndexType) -> Result<TypedVector> {
+        let name = id.to_string();
```
sundy-li, line 38: IndexType is aliased to usize. Why we use it as name ?

query/src/common/evaluator/scalar.rs
```diff
@@ -0,0 +1,181 @@
...
+                    .iter()
+                    .map(Self::build_eval_tree)
+                    .collect::<Result<_>>()?;
+                let arg_types: Vec<&DataTypeImpl> = func.arg_types.iter().collect();
+                let func = FunctionFactory::instance().get(func.func_name.as_str(), &arg_types)?;
+                Ok(EvalNode::Function { func, args })
+            }
+            Scalar::Cast(cast) => {
+                let arg = Self::build_eval_tree(&cast.argument)?;
+                let func = CastFunction::create_try(
```
sundy-li, line 108: Do we have `Scalar::TryCast` ?

1. `cast`  will throw exceptions if an error happens, so its return type is not null.
2. `try_cast`will default to return null if an error happens, its return type is nullable.

sundy-li: 1. https://docs.snowflake.com/en/sql-reference/functions/try_cast.html
2. https://docs.firebolt.io/sql-reference/functions-reference/try-cast.html

query/src/common/evaluator/eval_context.rs
leiysky: It's actually an unique ID of each column that can be referenced in a query. `ColumnIndex` is bound with `Metadata` by binder, so we can just utilize with it in executor.

A `EvalContext` is an unordered set of variables(i.e. `TypedVector`), each variable is assigned with a unique ID(`IndexType` for `ScalarExpr`). What we want here is a `HashMap<IndexType, TypedVector>` rather than a `DataBlock`, but we can keep it for now.

leiysky: And it's possible to use a string instead, but I prefer to use `IndexType` here.

query/src/common/evaluator/scalar.rs
leiysky: We are going to put `Function` object inside `Scalar` later, so every function can be represented with `Scalar::FunctionCall`, including `Cast` and `TryCast`(they can be handled in type checking). 

For now we only keep `TryCast`, since it can satisfy most of the scenarios.

sundy-li (approved):

query/src/common/evaluator/eval_context.rs
```diff
@@ -0,0 +1,106 @@
...
+        let column = self.try_column_by_name(name.as_str())?;
+        let field = self.schema().field_with_name(name.as_str())?;
```
xudong963, line 40: It seems that `column.data_type()` and `field.data_type()` are the same?

sundy-li: `field.data_type()` is logical datatype.
`column.data_type()` is physical datatype.

Timestamp is logical datatype  as its physical datatype is int64.

xudong963 (approved):

PR #5691, opened 2022-05-30
Feat: statement `delete from... `

### LINKED ISSUE #5146: DELETE statement
**Summary**

Each segment file(parquet) in the underlying storage is immutable; it is not suited for mutable operations (e.g., Delete, Update).
There are many ways to improve that, such as adding a hidden column for each row: `1` mark for deleted, `0` is normal.
The parquet file also needs to read the old and write the new mark flag to the new, the parquet file is `compact` format(vs. `wide` format), so it is not well suited for any modification-related operations.
For Databend deletion, the first version must work, then the performance. The initial plan is:
1. Parse the delete statement
2. Check which segments are hit by the statement
3. Create a row-level bitmap in BlockInfo or a single delete file?
5. Commit the new snapshot
6. Done

There is already a draft PR in #4228, but it still needs work.

dantengsky:

## Summary

- Simple delete statement (**preview edition**)
` DELETE FROM table WHERE ...`

~~~
mysql> -- both planners v0/v1 are supported
mysql> create table t as select * from numbers(10);

mysql> delete from t where number % 2 = 0;

mysql>
mysql> select * from t;
+--------+
| number |
+--------+
|      1 |
|      3 |
|      5 |
|      7 |
|      9 |
+--------+
5 rows in set (0.03 sec)
Read 5 rows, 40.00 B in 0.005 sec., 978.99 rows/sec., 7.65 **KiB/sec.**
~~~
- Main known limitations
  - `USING` clause is not supported yet
  - Non-distributed execution
  - no conflicts resolution yet
    in the case of mutation conflicts, the delete operation will abort

-----------

## Changelog

- New Feature

## Related Issues

Fixes #5146


Xuanwo:
This PR could be affected by a huge refactor I did on AST statement, FYI: https://github.com/datafuselabs/databend/pull/6176

query/src/storages/fuse/operations/delete.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::storages::fuse::meta::TableSnapshot;
+use crate::storages::fuse::operations::mutation::delete_from_block;
+use crate::storages::fuse::operations::mutation::mutations_collector::Deletion;
+use crate::storages::fuse::operations::mutation::mutations_collector::DeletionCollector;
+use crate::storages::fuse::pruning::BlockPruner;
+use crate::storages::fuse::FuseTable;
+use crate::storages::Table;
+
+impl FuseTable {
+    pub async fn do_delete(&self, ctx: Arc<QueryContext>, plan: &DeletePlan) -> Result<()> {
```
BohuTANG, line 33: What will happen if many threads do the same delete query?

BohuTANG: Will we add more unit tests and span tracing for the delete-related functions?

query/src/storages/fuse/operations/read_partitions.rs
```diff
@@ -37,6 +37,7 @@ impl FuseTable {
         ctx: Arc<QueryContext>,
         push_downs: Option<Extras>,
     ) -> Result<(Statistics, Partitions)> {
+        eprintln!("push down R {:?}", push_downs);
```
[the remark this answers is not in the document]
BohuTANG: Hmm, to avoid the stale codes left in the PR, it's better to use `tracing::debug!()` to instead.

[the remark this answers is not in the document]
dantengsky: thanks~

query/src/sql/planner/plans/mod.rs
```diff
@@ -287,6 +290,7 @@ impl Plan {
             Plan::RevokePriv(_) => Arc::new(DataSchema::empty()),
             Plan::RevokeRole(_) => Arc::new(DataSchema::empty()),
             Plan::Insert(plan) => plan.schema(),
+            Plan::Delete(_) => todo!(),
```
zhang2014, line 293:
```suggestion
            Plan::Delete(_) => Arc::new(DataSchema::empty()),
```

query/src/storages/fuse/operations/mutation/block_filter.rs
```diff
@@ -0,0 +1,101 @@
...
+use common_exception::Result;
+use common_planners::Expression;
+
+use crate::pipelines::transforms::ExpressionExecutor;
+use crate::sessions::QueryContext;
+use crate::storages::fuse::meta::BlockMeta;
+use crate::storages::fuse::operations::mutation::mutations_collector::Deletion;
+use crate::storages::fuse::FuseTable;
+
+pub async fn delete_from_block(
```
zhang2014, line 28: Why can't we reuse 'insert into overwrite table SELECT * FROM table WHERE delete_selector ' code? 
maybe we can support `INSERT INTO OVERWRITE_BLOCK table SELECT * FROM table WHERE delete_selector`.

zhang2014: got it. please ignore

BohuTANG:
cc @soyeric128 for documentation

soyeric128:
> cc @soyeric128 for documentation

Please review this PR for doc updates: https://github.com/datafuselabs/databend/pull/6260

PR #5692, opened 2022-05-30
improve(parser): parse integer to u64

andylokandy:

## Summary

- Previously, the integer in AST was a string. But now, it's a real u64.
- Same for floats.

## Changelog

- Improvement


leiysky (approved):

sundy-li (approved):

sundy-li:
@andylokandy Maybe we should support negative literal in the params field of the `FunctionCall`.

```
f(-3, 4)(a,b)
```
is parsed incorrectly now.

PR #5710, opened 2022-05-31
improve(test): Minor improvement of sqllogic test

leiysky:

## Summary

- Add a simple CLI tool.
- Support filtering test cases with regex

## Changelog

- Improvement

### DIFF
--- a/tests/logictest/main.py
+++ b/tests/logictest/main.py
@@ -10,17 +10,19 @@ from config import mysql_config, http_config
 import fire
 
 
-def run(pattern=".*", enable_mysql=True, enable_http=True):
+def run(pattern=".*"):
     """
     Run tests
     """
-    if enable_mysql:
+    disable_mysql_test = os.getenv("DISABLE_MYSQL_LOGIC_TEST")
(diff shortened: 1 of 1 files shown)


tests/logictest/main.py
```diff
@@ -7,17 +7,25 @@
...
-    if disable_http_test is None:
-        http = TestHttp("http")
+import fire
+
+
+def run(pattern=".*", enable_mysql=True, enable_http=True):
+    """
+    Run tests
+    """
+    if enable_mysql:
```
ZhiHanZ, line 17: I think we still need to push down the feature flags by some flags or environment variables

ZeaLoVe: You can use config.py to keep default value, make sure always use the config of environment variables. All current environment variables must keep.

leiysky: Ok, I just restore the env variables.

leiysky:
@ZhiHanZ @ZeaLoVe PTAL

BohuTANG (approved):

PR #5715, opened 2022-06-01
feat(planner): support cross join

### LINKED ISSUE #5499: Feature: support cross join
**Summary**

Description for this feature.

xudong963:

## Summary

1. Minor refactor for hash join
2. support cross join

## Changelog

- New Feature

## Related Issues

Fixes #5499


query/src/sql/planner/plans/logical_join.rs
```diff
@@ -23,15 +23,27 @@ use crate::sql::plans::PhysicalPlan;
...
+    AntiJoin,
+    CrossJoin,
+}
+
 #[derive(Clone, Debug)]
-pub struct LogicalInnerJoin {
+pub struct LogicalJoin {
     pub left_conditions: Vec<Scalar>,
     pub right_conditions: Vec<Scalar>,
+    pub join_type: JoinType,
```
leiysky, line 41: It's better to introduce another struct `CrossJoin`.

The predicate of inner join and outer join are different. 

We can split conditions into `left_conditions` and `right_conditions` for inner join, but for outer join, we have to keep the `other_conditions` too.

And for `CrossJoin`, we will lift the predicates into a `Filter` for push down.

leiysky:
You can implement this as `select * from t inner join t1 on 1 = 1`, so you don't need to write cross join related logic.

leiysky (approved):
LGTM

sundy-li (approved):

BohuTANG:
@mergify update

PR #5740, opened 2022-06-02
Add cluster name to metasrv and meta-client

### LINKED ISSUE #5606: Add cluster name to metasrv and meta-client
A client will only be allowed to metasrv cluster with the same cluster name.

TODO:
- [ ] elaborate what needs to be done.

devillove084:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues
Fixes #5606

### DIFF
--- a/common/meta/raft-store/src/config.rs
+++ b/common/meta/raft-store/src/config.rs
@@ -98,6 +98,10 @@ pub struct RaftConfig {
 
     /// For test only: specifies the tree name prefix
     pub sled_tree_prefix: String,
+
+    ///  The node name. If the user specifies a name,
+    /// the user-supplied name is used, if not, the default name is used.
+    pub cluster_name: String,
 }
 
 pub fn get_default_raft_advertise_host() -> String {
@@ -127,6 +131,7 @@ impl Default for RaftConfig {
             join: vec![],
             id: 0,
             sled_tree_prefix: "".to_string(),
+            cluster_name: "foo_cluster".to_string(),
         }
     }
 }
--- a/metasrv/src/configs/outer_v0.rs
+++ b/metasrv/src/configs/outer_v0.rs
@@ -184,6 +184,7 @@ pub struct ConfigViaEnv {
     pub metasrv_join: Vec<String>,
     pub kvsrv_id: u64,
     pub sled_tree_prefix: String,
+    pub cluster_name: String,
 }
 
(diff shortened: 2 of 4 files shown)


common/meta/raft-store/src/config.rs
```diff
@@ -98,6 +98,10 @@ pub struct RaftConfig {
 
     /// For test only: specifies the tree name prefix
     pub sled_tree_prefix: String,
+
+    ///  The node name. If the user specifies a name,
+    /// the user-supplied name is used, if not, the default name is used.
+    pub name: String,
```
drmingdrmer, line 104: This should be the cluster name, not a name of a node in a cluster, IMHO

metasrv/src/configs/outer_v0.rs
```diff
@@ -334,6 +337,11 @@ pub struct RaftConfig {
     /// For test only: specifies the tree name prefix
     #[clap(long, default_value = "")]
     pub sled_tree_prefix: String,
+
+    /// Tne node name. If the user specifies a name, the user-supplied name is used,
+    /// if not, the default name is used
+    #[clap(long, default_value = "foo_cluster")]
+    pub name: String,
 }
```
drmingdrmer, line 345: All of the default value for a field in inner config, outer config and env config has to be the same. Because the `serfig` crate use the default value to determine if a field is set by user or not.

drmingdrmer (commented):
It would be better using a elaborate field name `cluster_name`

drmingdrmer:
The following task to enable cluster-name check:

- [ ] The implementation of `RaftNetwork` has to send cluster-name in every request.
- [ ] Raft RPC handler has to check the cluster-name before handling an RPC.
- [ ] Non-raft RPC should send cluster-name to the remote peer too.
- [ ] Non-raft RPC has to check the cluster-name before handling an PRC.

common/meta/raft-store/src/config.rs
devillove084: Done.

metasrv/src/configs/outer_v0.rs
devillove084: Because the default values in the code are hard-coded, when the user modifies the cluster_name in the configuration file or parameter, serfig will see inconsistencies with the default. The default name in the code is now foo_cluster

Issue #5647, opened 2022-05-28
Feature: Implement UDF in new planner


leiysky:
In databend, we support user-defined function(aka UDF)

To support UDF in new planner, we have to:
- Support parse UDF definition with new SQL parser
- Support reolve UDF when doing type check

PR #5751, opened 2022-06-02
feat(planner): support udf

xudong963:

## Summary

as the tile

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/5647

### DIFF
--- a/common/ast/src/parser/mod.rs
+++ b/common/ast/src/parser/mod.rs
@@ -24,6 +24,7 @@ use common_exception::ErrorCode;
 use common_exception::Result;
 
 use self::error::DisplayError;
+use crate::ast::Expr;
 use crate::ast::Statement;
 use crate::parser::error::Backtrace;
 use crate::parser::statement::statements;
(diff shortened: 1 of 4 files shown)


common/ast/src/parser/mod.rs
```diff
@@ -52,3 +53,20 @@ pub fn parse_sql<'a>(
         Err(nom::Err::Incomplete(_)) => unreachable!(),
     }
 }
+
+/// Parse udf function into Expr
+pub fn parse_udf<'a>(
```
leiysky, line 58: I think `parse_expr` is better.

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -446,9 +451,12 @@ impl<'a> TypeChecker<'a> {
     pub async fn resolve_function(
         &mut self,
         func_name: &str,
-        arguments: &[&Expr<'a>],
+        arguments: &[&Expr<'_>],
         _required_type: Option<DataTypeImpl>,
     ) -> Result<(Scalar, DataTypeImpl)> {
+        if !is_builtin_function(func_name) {
+            return self.resolve_udf(func_name, arguments).await;
+        }
```
leiysky, line 459: We should move this check to https://github.com/datafuselabs/databend/blob/83884cd152d43c5d77907262a4a58f2d8e0be083/query/src/sql/planner/semantic/type_check.rs#L259-L268

common/ast/src/parser/mod.rs
xudong963: done

query/src/sql/planner/semantic/type_check.rs
xudong963: done

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -933,4 +940,284 @@ impl<'a> TypeChecker<'a> {
...
+            })?;
+            self.resolve(&udf_expr, None).await
+        } else {
+            Err(ErrorCode::SemanticError(
+                "No function matches the given name.",
+            ))
+        };
+    }
+
+    fn clone_expr_with_replacement<F>(
```
leiysky, line 988: It's better to do this with a unified visitor, but we can open another issue for this.

leiysky (approved):

query/src/sql/planner/semantic/type_check.rs
xudong963: Ok

PR #5752, opened 2022-06-02
feat(meta): Output [binary] in debug message instead

Xuanwo:

## Summary

This will make developers' live easier:

### Before

image

image

### After

image

image

We will output `[binary]` in logs instead so that our logs will not be polluted by non-sense binaries, for example:

image

## Changelog

- New Feature


lichuang:
LGTM

PR #5770, opened 2022-06-04
feat: undrop database

### LINKED ISSUE #5701: Feature:  `undrop database` statement
**Summary**

As schema api already supports `undrop_database` operation

https://github.com/datafuselabs/databend/blob/main/common/meta/api/src/schema_api_impl.rs#L268

`undrop database` statement can be implemented now.

- the syntax: 

`UNDROP DATABASE <name>` 

- the "sementics"
   undrop the dropped table ... : -), e.g.
~~~
  create database test;
  create table test.tbl (a int);
  drop database test;
  select * from test.tbl; -- expects error 1003, `UnknownDatabase`
  undrop database test;
  select * from test.tbl; -- back to normal
~~~

reference: 
- https://github.com/datafuselabs/databend/pull/5562
- NOTE, unlike the previous PR
  - do not need  to implement the `show databases history` statement
  - and   legacy parser(`sqlparser-rs`) only is OK

LiuYuHui:

## Summary
Implement `undrop database` statement.

## Changelog

- New Feature

## Related Issues

Fixes #5701


dantengsky:
@LiuYuHui 

Nice work!

As you might already noticed that `Hive` catalog should also impl the method `async fn undrop_database`.

https://github.com/datafuselabs/databend/runs/6737371920?check_suite_focus=true#step:3:1368

The feature `hive` is not enabled by default,  to eable it we need something like this:

`cargo build --bin databend-query --features hive`

then we can check if it compiles.

dantengsky (approved):
LGTM

BohuTANG:
Nice job!
Thanks for your contribution.

PR #5783, opened 2022-06-05
fix(query): Don't load credential while reading stage

Xuanwo:

## Summary

This PR will disable the credential loader while the reading stage so that the following SQL works:

```sql
CREATE TABLE books
(
    title VARCHAR,
    author VARCHAR,
    date VARCHAR
);


COPY INTO books
  FROM 's3://databendcloud/loaddata/books.csv'
  file_format = (type = 'CSV' field_delimiter = ','  record_delimiter = '\n' skip_header = 0);
```

That means we can load data from a public repo directly!

## Changelog

- New Feature


common/protos/proto/config.proto
```diff
@@ -26,6 +26,7 @@ message S3StorageConfig {
   string bucket = 5;
   string root = 6;
   string master_key = 7;
+  bool disable_credential_loader = 8;
```
BohuTANG, line 29: I think here should add option to backwards compatible

Xuanwo: > If you add new fields, any messages serialized by code using your "old" message format can still be parsed by your new generated code. You should keep in mind the [default values](https://developers.google.com/protocol-buffers/docs/proto3#default) for these elements so that new code can properly interact with messages generated by old code. Similarly, messages created by your new code can be parsed by your old code: old binaries simply ignore the new field when parsing. See the [Unknown Fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns) section for details.

In my humble opinion, using the default value `false` here is enough.

reference: https://developers.google.com/protocol-buffers/docs/proto3#updating

BohuTANG: Seems protobobuf can handle a new field with the default value?
cc @drmingdrmer

drmingdrmer: AFAIK, protobuf always fills in a default value for an absent field in the message:)

With prost, this explains the behavior of the modifier:
https://github.com/tokio-rs/prost#field-modifiers

<p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(36, 41, 47); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Protobuf scalar value and enumeration message fields can have a modifier depending on the Protobuf version. Modifiers change the corresponding type of the Rust field:</p>

.proto Version | Modifier | Rust Type
-- | -- | --
proto2 | optional | Option<T>
proto2 | required | T
proto3 | default | T for scalar types, Option<T> otherwise
proto3 | optional | Option<T>
proto2/proto3 | repeated | Vec<T>

BohuTANG (approved):

Issue #5784, opened 2022-06-05
I found there is a flaw in the proto-compatibility:


drmingdrmer:
I found there is a flaw in the compatibility:

The updated protobuf `TableMeta` has changed. The protobuf types should be serialized with a new `ver`.
But the data written by a query with new `ver` can not be read by a query with the old `ver`,
when only some of the query-nodes in a cluster are upgraded.

The crate `proto-conv` has to write the `OLDEST_COMPATIBLE_VER` into protobuf message, so that an old query will be able to find out if it is safe to read the data.

https://github.com/datafuselabs/databend/blob/25531a724d38fd242a03155fbfea63e443b68a03/common/proto-conv/src/util.rs#L17-L18

_Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/issues/5718#issuecomment-1146746600_

PR #5785, opened 2022-06-05
[meta] feature: protobuf message has to persist `MIN_COMPATIBLE_VER` in it

drmingdrmer:

##### [meta] feature: protobuf message has to persist `MIN_COMPATIBLE_VER` in it

**Non-breaking** meta-protobuf change.

This way to let old query be able to decide if it is safe to load data
written by a newer query executable.

- Fix: #5784

## Changelog

- New Feature

## Related Issues

### DIFF
--- a/common/protos/proto/config.proto
+++ b/common/protos/proto/config.proto
@@ -18,6 +18,7 @@ package databend_proto;
 
 message S3StorageConfig {
   uint64 version = 100;
+  uint64 min_compatible = 101;
 
   string region = 1;
   string endpoint_url = 2;
@@ -30,6 +31,7 @@ message S3StorageConfig {
 
 message FsStorageConfig {
   uint64 version = 100;
+  uint64 min_compatible = 101;
 
   string root = 1;
 }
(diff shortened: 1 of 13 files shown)


BohuTANG:
This is a breaking change?

drmingdrmer:
> This is a breaking change?

No:)

This PR only affects the query:
It just added a new field into protobuf message. Old query will just ignore this new field when loading new data. New query will have a default value `0` when loading the old data.

common/protos/proto/config.proto
```diff
@@ -18,6 +18,7 @@ package databend_proto;
 
 message S3StorageConfig {
   uint64 version = 100;
+  uint64 min_compatible = 101;
```
zhyass, line 21: Why `min_compatible` is 101 bigger then `version`?

drmingdrmer: This is the index of a field. not the value of it

zhyass: 😂, My fault.

zhyass (approved):
LGTM

PR #5791, opened 2022-06-06
row base serializer

youngsofun:

## Summary

Conclusions of prev bench results:
Compared with row base impl in this pr:
1. Return vec<string> is very slow
2. Return something like StringColumn is faster but still slow(2 times) and has more memory overhead
3. Return streamin_iter is a bit complex and a litter slower

with row base impl
1.   holding &[T] in Serializer make it  30%  faster, compared with passing ColumnRef each time and cast
2.   enum_dispath is 10% faster than dyn dispatch

## Changelog

- Performance Improvement

## Related Issues

Fixes #issue


youngsofun:
enum_dispath is not IDE friendly, both do not support associate const( can be used in `quote`),  not sure if enum_dispath can be able to support it later, is there any other drawback on it， since it is only 10% faster?

@sundy-li

common/datavalues/src/types/serializations_new/boolean.rs
```diff
@@ -0,0 +1,51 @@
...
+use common_io::prelude::FormatSettings;
+
+use crate::prelude::*;
+use crate::types::serializations_new::TypeSerializerNew;
+
+const TRUE_STR: &[u8] = &[b'1'];
+const FALSE_STR: &[u8] = &[b'0'];
+
+#[derive(Clone)]
+pub struct BooleanSerializerNew<'a> {
```
sundy-li, line 14: Seems we do not need 'a here

sundy-li:
> enum_dispath is not IDE friendly, both do not support associate const( can be used in quote), not sure if enum_dispath can be able to support it later, is there any other drawback on it， since it is only 10% faster?

Let's make life easier, enum_dispatch is not needed.

youngsofun:
some notes:

1.   add ConstSerializer,  it can be optimized a lot later by caching the serialized value
2.   MySQL writer writes string for float, to make float formats consistent， its the only place need to change the test. see  the 2nd commit

youngsofun:
> > enum_dispath is not IDE friendly, both do not support associate const( can be used in quote), not sure if enum_dispath can be able to support it later, is there any other drawback on it， since it is only 10% faster?
> 
> Let's make life easier, enum_dispatch is not needed.

the serializer is used in limited places, not too bothering,  I prefer keep it for now

youngsofun:
- [ ]  use options of lexical

youngsofun:
@BohuTANG @sundy-li  is it acceptable to see `1.0` instead of `1` for float?

sundy-li:
> table to see `1.0` instead of `1` f

https://www.db-fiddle.com/f/abKfoifJD29jCXVgBNnZap/0

MySQL outputs 1, but I think we do not need to follow MySQL strictly.

youngsofun:
snowflake

image

youngsofun:
I prefer  always use `1.0`

1.  uniform format
2.  easier to know the type.
3.  to use `1`  for MySQL , need `lexical_core::write_with_options`, instead of  `lexical_core::write`,  maybe a little slow 

`trim_float`  is not used for any prepared FORMATS in lib lexical, including MySQL 

https://github.com/Alexhuszagh/rust-lexical/blob/main/lexical-write-float/src/options.rs#L1373


> > table to see `1.0` instead of `1` f
> 
> https://www.db-fiddle.com/f/abKfoifJD29jCXVgBNnZap/0
> 
> MySQL outputs 1, but I think we do not need to follow MySQL strictly.

youngsofun:
have some problem to use WriteFloatOptions


`N::Options == WriteFloatOptions` only holds for f32 and f64

```
pub fn write_with_options<'a, N: ToLexicalWithOptions, const FORMAT: u128>(
    n: N,
    bytes: &'a mut [u8],
    options: &N::Options,
) -> &'a mut [u8] {
    n.to_lexical_with_options::<FORMAT>(bytes, options)

```

common/datavalues/src/types/data_type.rs
```diff
@@ -125,7 +128,26 @@ pub trait DataType: std::fmt::Debug + Sync + Send + DynClone {
 
     fn create_mutable(&self, capacity: usize) -> Box<dyn MutableColumn>;
 
-    fn create_serializer(&self) -> TypeSerializerImpl;
+    fn create_serializer_const<'a>(&self, col: &'a ColumnRef) -> Result<TypeSerializerImpl<'a>> {
```
zhang2014, line 131: Maybe we can merge `create_serializer_const` into `create_serializer`?

youngsofun: oh, I got you ..., you are right, I will refator it

common/datavalues/src/types/data_type.rs
```diff
@@ -125,7 +128,26 @@ pub trait DataType: std::fmt::Debug + Sync + Send + DynClone {
...
+    }
+    fn create_serializer<'a>(&self, col: &'a ColumnRef) -> Result<TypeSerializerImpl<'a>> {
+        if col.is_const() {
+            self.create_serializer_const(col)
+        } else {
+            self.create_serializer_inner(col)
+        }
+    }
+
+    fn create_serializer_inner<'a>(&self, _col: &'a ColumnRef) -> Result<TypeSerializerImpl<'a>> {
```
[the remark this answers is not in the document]
youngsofun: create_serializer_inner use for impl
create_serializer as interface

common/datavalues/src/types/data_type.rs
youngsofun: done, thanks

youngsofun:
oops,   SQL logic tests take effect,  need to fix for float too

common/datavalues/src/types/serializations/array.rs
```diff
@@ -12,86 +12,91 @@
...
         }
 
-        let inner_data = self
-            .inner
-            .serialize_clickhouse_format(column.values(), format)?;
+        let inner_data = self.inner.serialize_clickhouse_column(format)?;
         Ok(Arc::new(ArrayColumnData::create(inner_data, offsets)))
     }
+
+    fn serialize_json(&self, format: &FormatSettings) -> Result<Vec<Value>> {
```
sundy-li, line 90: So row based api not works on `serialize_json` ?

youngsofun: serializer can has API for both row based and col-based(clickhouse,  json::Value)

youngsofun: format choose to use which

sundy-li:
```
pub fn block_to_json_value(
    block: &DataBlock,
    format: &FormatSettings,
) -> Result<Vec<Vec<JsonValue>>> {
    let cols = block_to_json_value_columns(block, format)?;
    Ok(transpose(cols))
}
```

Since we need to transpose, so `serialize_json ` is row based ?

sundy-li: But it's not zero alloc in the transpose function. So row based is faster in this case.

youngsofun: > But it's not zero alloc in the transpose function. So row based is faster in this case.

make sense

youngsofun:
@sundy-li @zhang2014  rebase only to include new test cases, the original commits are not changed

common/datavalues/src/types/serializations/array.rs
youngsofun: may I refactor it later in another pr later,  this pr is large and need to change new-added tests case, hope it to merge early.

youngsofun: also need add formats based on this

youngsofun:
@sundy-li @zhang2014  all tests passed (except a build),  please review

PR #5806, opened 2022-06-06
hotfix: lz4raw compression of zero len buffer

dantengsky:

## Summary

fix bug, unexpected compression errors if 
  - column is nullable 
  - and all rows of that column are null

or put it another way, "unexpected compression error" when the length of input buffer is zero.

modificatios of [de]compression: 

https://github.com/datafuse-extras/parquet2/commit/07176f1fd7899f5603420c5d87460c7e132b4f2e

The case to reproduce the bug:

~~~
CREATE TABLE t ( `id` BIGINT UNSIGNED NULL, `business_id` BIGINT UNSIGNED NULL, `col3` BIGINT UNSIGNED NULL  );
insert into t(id) select * from numbers(20000);  -- the original case that discover this issue
insert into t(id) select * from numbers(112); -- below 112 rows, no error, BUT, it is just a accident 
insert into t(id) select * from numbers(113); -- 113 rows or above, unexpected compression error occurs
~~~

## Changelog

- Bug Fix

## Related Issues

Fixes #issue


Cargo.toml
```diff
@@ -65,5 +65,5 @@ object = { opt-level = 3 }
 rustc-demangle = { opt-level = 3 }
 
 [patch.crates-io]
-parquet2 = { version = "0.12", optional = true, git = "https://github.com/datafuse-extras/parquet2", rev = "03c108b" }
+parquet2 = { version = "0.12", optional = true, git = "https://github.com/datafuse-extras/parquet2", rev = "07176f1f" }
```
BohuTANG, line 68: This is a temporary fix?
https://github.com/datafuse-extras/parquet2/pull/3/commits/ea7820dc4f96bce29bae4d2def57b8ba67fd2f50

Issue #5825, opened 2022-06-07
Feature: support `SetOperator` in new planner


xudong963:
**Summary**

Support `UNION`, `EXCEPT`, `INTERSECT` in new planner

PR #5833, opened 2022-06-08
feat(planner): support semi, anti join and part of set operators

### LINKED ISSUE #5794: Feature: support semi join and anti join
**Summary**

Description for this feature.

xudong963:

## Summary

This PR contains three parts
1. Support set operators in new parser
2. Support **intersect distinct** and **except distinct** in new planner
3. Support **semi and anti** join in new planner

## Changelog

- New Feature

## Related Issues

Fixes #5794 and part of #5825


query/src/servers/mysql/mysql_interactive_worker.rs
```diff
@@ -309,6 +311,14 @@ impl<W: std::io::Write> InteractiveWorkerBase<W> {
                         let (plan, _) = PlanParser::parse_with_hint(query, context.clone()).await;
                         plan.and_then(|v| InterpreterFactory::get(context.clone(), v))
                     };
+                } else {
+                    // If old parser failed, try new planner
```
leiysky, line 315: Should check if new planner is enabled here.

common/ast/src/parser/query.rs
```diff
@@ -305,3 +271,68 @@ pub fn order_by_expr(i: Input) -> IResult<OrderByExpr> {
...
+                    .map(|(_, _, group_by)| group_by)
+                    .unwrap_or_default(),
+                having: opt_having_block.map(|(_, having)| having),
+            }))
+        },
+    );
+
+    map(
+        rule!(#select ~ (UNION | EXCEPT | INTERSECT)? ~ ALL? ~ #set_expr?),
+        |(select, set_operator, all_kind, set_expr)| {
```
leiysky, line 311: There is execution precedence of set operators.

For example, 
```sql
(select * from t union select * from t1) intersect select * from t2
```

The SQL will be interpreted into:
```
   ∩
  / \
 ∪  t2
/ \
t t1
```

So the rule should be like `#set_expr ~ SET_OP ~ #set_expr`.

query/src/sql/planner/binder/select.rs
```diff
@@ -193,4 +213,122 @@ impl<'a> Binder {
...
+
+    fn bind_intersect_or_except(
+        &mut self,
+        left_context: BindContext,
+        right_context: BindContext,
+        left_expr: SExpr,
+        right_expr: SExpr,
+        join_type: JoinType,
+    ) -> Result<(SExpr, BindContext)> {
+        let left_expr = self.bind_distinct(
```
leiysky, line 298: Better not to reuse the `bind_distinct` here, since it's not actually a `DISTINCT` clause. Construct an `Aggregate` operator here is fine.

query/src/sql/planner/binder/select.rs
```diff
@@ -193,4 +213,122 @@ impl<'a> Binder {
         let new_expr = SExpr::create_unary(filter_plan.into(), child);
         Ok(new_expr)
     }
+
+    pub(super) async fn bind_set_operator(
+        &mut self,
+        bind_context: &BindContext,
+        left: &'_ SetExpr<'_>,
+        right: &'_ SetExpr<'_>,
```
leiysky, line 221:
```suggestion
        left: &SetExpr<'_>,
        right: &SetExpr<'_>,
```

query/src/sql/planner/binder/select.rs
```diff
@@ -193,4 +213,122 @@ impl<'a> Binder {
...
+            return Err(ErrorCode::SemanticError(
+                "SetOperation must have the same number of columns",
+            ));
+        } else {
+            for (left_col, right_col) in left_bind_context
+                .columns
+                .iter()
+                .zip(right_bind_context.columns.iter())
+            {
+                if !left_col.data_type.eq(&right_col.data_type) {
```
leiysky, line 238:
```suggestion
                if &left_col.data_type == &right_col.data_type {
```

leiysky: We should coerce the result columns of inputs operators.

For example:
```sql
create table t(a int8 null);
create table t1(b int64 not null);

select * from t union select * from t1; -- this is allowed
```

common/ast/src/parser/query.rs
xudong963: Yes, I have thought the case and wanted to test the current rule if satisfy it, but I got the following error
```
mysql> (select * from t1 intersect select * from t2) intersect select * from t1;
  --> SQL:1:1
  |
1 | (select * from t1 intersect select * from t2) intersect select * from t1
  | ^ expected `EXPLAIN`, `SELECT`, `SHOW`, `CREATE`, `DROP`, `ALTER`, or 10 more ...

.
```



> #set_expr ~ SET_OP ~ #set_expr

This should be the rule of `SetOperation`, for `SetExpr`, the rule may be `#select | #set_expr`, but this will cause dead loop.

query/src/sql/planner/binder/select.rs
xudong963: Yes, I knew `bind_distinct` just wrapped a `Aggregate` plan for current `SExpr`.

Here my aim is to remove duplicates for `left_expr` to implement `intersect` or `except` **distinct**, so I think it's ok to reuse `bind_distinct`, which fits the context.

common/ast/src/parser/query.rs
leiysky:
```ebnf
paren_select := ( '(', paren_select, ')' ) | select;

set_operation := set_expr, SET_OP, set_expr;

set_expr :=  set_operation | subquery | paren_select;
```

You may need this.

xudong963:
I'll fix parser related tests after reviewers think my change in parser is ok.

common/ast/src/parser/query.rs
```diff
@@ -305,3 +271,131 @@ pub fn order_by_expr(i: Input) -> IResult<OrderByExpr> {
...
+                having: opt_having_block.map(|(_, having)| having),
+            }))
+        },
+    )(i)
+}
+
+pub fn parenthesized_select(i: Input) -> IResult<SetExpr> {
+    map(
+        rule! (
+            "(" ~ (#parenthesized_select  | #select ) ~ ")"
```
leiysky, line 316: You can write this as:

```
( "(" ~ #parenthesized_select ~ ")" ) | #select
```

Both parenthesized `SELECT` statement and `SELECT` statement can be reduced into `SetExpr` with rule.

So you don't have to write `(#select | #parenthesized_select)` in https://github.com/datafuselabs/databend/blob/1bdac9f4eaee2eeee13df184bec6206569ad918d/common/ast/src/parser/query.rs#L325

leiysky:
Rest LGTM, @andylokandy PTAL

common/ast/src/parser/query.rs
xudong963: > ( "(" ~ #parenthesized_select ~ ")" ) | #select

`( "(" ~ #parenthesized_select ~ ")" )` and `#select` don't have same type, so can we combine them with `|`?

xudong963:
Make the pr as draft

After aligning with @leiysky and @andylokandy , we need to do more for parser to support `Set operator`:

Nom can't solve the left combination problem, so we need to introduce `Pratt Parser` to process `Set operator` .
In the future, we can also use `Pratt Parser` to process `Join` and make it more regular.

common/ast/src/parser/query.rs
```diff
@@ -305,3 +280,229 @@ pub fn order_by_expr(i: Input) -> IResult<OrderByExpr> {
...
+    let group = map(
+        rule! {
+           "("
+           ~ ^#sub_set_operation()
+           ~ ^")"
+        },
+        |(_, set_expr, _)| SetOperationElement::Group(set_expr),
+    );
+
+    map(rule!(#group |#set_operator | #select_stmt), |elem| elem)(i)
```
andylokandy, line 371:
```suggestion
    rule!(
        #group
        | #set_operator
        | #select_stmt
    )(i)
```

common/ast/src/parser/query.rs
```diff
@@ -305,3 +280,229 @@ pub fn order_by_expr(i: Input) -> IResult<OrderByExpr> {
...
+        };
+        Ok(set_expr)
+    }
+
+    fn prefix(
+        &mut self,
+        _op: Self::Input,
+        _rhs: Self::Output,
+    ) -> Result<Self::Output, Self::Error> {
+        todo!()
```
andylokandy, line 469:
```suggestion
        unreachable!()
```

common/ast/src/parser/query.rs
```diff
@@ -305,3 +280,229 @@ pub fn order_by_expr(i: Input) -> IResult<OrderByExpr> {
...
+    }
+
+    fn postfix(
+        &mut self,
+        _lhs: Self::Output,
+        _op: Self::Input,
+    ) -> Result<Self::Output, Self::Error> {
+        todo!()
```
andylokandy, line 477:
```suggestion
        unreachable!()
```

common/ast/src/parser/query.rs
```diff
@@ -311,3 +284,190 @@ pub fn order_by_expr(i: Input) -> IResult<OrderByExpr> {
...
+    let iter = &mut set_operation_elements.into_iter();
+    let mut iter = iter.peekable();
+    let set_expr = SetOperationParser
+        .parse_input(&mut iter, Precedence(0))
+        .map_err(|err| Error::from_error_kind(rest.slice(..1), ErrorKind::from(err)))
+        .map_err(nom::Err::Error)?;
+    if let Some(_) = iter.peek() {
+        // Rollback parsing footprint on unused expr elements.
+        i.1.clear();
+        Ok((i.slice(1..), set_expr))
```
andylokandy, line 386: The slice index should be starting from the span of the first unused element. Seems that `WithSpan` is still necessary.

xudong963: Yes, fixed

common/ast/src/parser/query.rs
```diff
@@ -285,76 +286,79 @@ pub fn order_by_expr(i: Input) -> IResult<OrderByExpr> {
     )(i)
 }
 
+#[derive(Debug, Clone)]
+pub struct SetOperationWithSpan<'a> {
+    span: Input<'a>,
+    elem: SetOperationElement<'a>,
+}
```
andylokandy, line 293: We can extract it to util.rs:

```suggestion
#[derive(Debug, Clone)]
pub struct WithSpan<'a, T> {
    span: Input<'a>,
    elem: T,
}
```

xudong963: Lol,  we think the same. I am doing it

common/ast/src/parser/util.rs
```diff
@@ -319,3 +323,43 @@ impl<'a> nom::Slice<RangeFull> for Input<'a> {
...
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct WithSpan<'a, T> {
+    pub(crate) span: Input<'a>,
+    pub(crate) elem: T,
+}
+
+pub fn pratt_parser<'a, I, P, E>(
```
andylokandy, line 333:
```suggestion
pub fn run_pratt_parser<'a, I, P, E>(
```

andylokandy (approved):
rest LGTM

query/src/sql/planner/binder/select.rs
```diff
@@ -193,4 +217,133 @@ impl<'a> Binder {
...
+            {
+                let left_data_type = remove_nullable(&left_col.data_type);
+                let right_data_type = remove_nullable(&right_col.data_type);
+                let left_id = left_data_type.data_type_id();
+                let right_id = right_data_type.data_type_id();
+                if left_id.is_numeric() && right_id.is_numeric() {
+                    let coercion_type =
+                        numerical_coercion(&left_data_type, &right_data_type, false)?;
+                    left_col.data_type = coercion_type.clone();
+                    right_col.data_type = coercion_type;
```
leiysky, line 251: Maybe we should keep the nullability of the coercion type?

query/src/sql/planner/binder/select.rs
```diff
@@ -139,20 +147,36 @@ impl<'a> Binder {
...
+        set_expr: &SetExpr,
+        order_by: &[OrderByExpr],
     ) -> Result<(SExpr, BindContext)> {
-        let (mut s_expr, bind_context) = match &query.body {
-            SetExpr::Select(stmt) => {
-                self.bind_select_stmt(bind_context, stmt, &query.order_by)
-                    .await
-            }
+        match set_expr {
+            SetExpr::Select(stmt) => self.bind_select_stmt(bind_context, stmt, order_by).await,
```
leiysky, line 157: The semantic of `ORDER BY` will become much different if we support set operations.

We should add some tests about this, for example:

```sql
create table t(a int);

select a as b from t order by a; -- allowed
select a as b from t union select b from t order by a; -- not allowed
```

query/src/sql/planner/binder/select.rs
xudong963: I'm not sure about that, I think both are ok. cc @sundy-li

leiysky: I think it's fine to use `merge_types` here, maybe @sundy-li would answer this.

sundy-li: Let's use `merge_types`.

query/src/sql/planner/binder/select.rs
xudong963: added

query/src/sql/planner/binder/select.rs
xudong963: done

tests/suites/0_stateless/20+_others/20_0002_planner_v2_display_error.sql
```diff
@@ -3,3 +3,5 @@ set enable_planner_v2 = 1;
 create table t(a int, b int);
 select aa from t;
 select *;
+create table t1(a int);
+select a as b from t1 intersect select b from t1 order by a;
```
leiysky, line 7: This is not exactly what I mean.

In current implementation, to support `ORDER BY` extra columns that doesn't defined in `SELECT` clause, we will pass the `ORDER BY` into `bind_select`.

But if there is set operator, then the `ORDER BY` cannot access the extra columns any more.

This test case should be rewritten as:
```sql
select a as b from t1 intersect select a as b from t1 order by a; -- column a doesn't exists, ORDER BY
```

And the implementation of order by should be fixed as well.

xudong963: fixed

leiysky (approved):

PR #5840, opened 2022-06-08
support function timezone()

### LINKED ISSUE #5820: Feature: Support function timezone
**Summary**

use like this:

```SQL
select timezone();
```

return a string val.

TCeason:

## Summary

Support function timezone.

this function will return a string value.

And I think no need to check if the timezone is valid. Because we already check the value is valid when we set it.

## Changelog

- New Feature

## Related Issues

Fixes #5820


b41sh:
almost ok, please docs for this function.

PR #5849, opened 2022-06-08
fix: http_connector will fetch all results

ZeaLoVe:

## Summary

http_connector gets a single response now and will fail the logic test when the HTTP handler returns empty data with a valid next_uri.
In order to fix this case, http_connector must fetch all results and combine them.
cc @youngsofun 

## Changelog

## Related Issues

Fixes #issue

### DIFF
--- a/tests/logictest/http_connector.py
+++ b/tests/logictest/http_connector.py
@@ -169,7 +169,7 @@ class HttpConnector():
         response = self.query(statement, current_session)
         log.info("response content: {}".format(response))
         response_list.append(response)
-        for i in range(3):
+        for i in range(6):
             if response['next_uri'] is not None:
                 try:
                     resp = requests.get(url="http://{}:{}{}".format(self._host,self._port,response['next_uri']), headers=self.make_headers())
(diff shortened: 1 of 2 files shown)


tests/logictest/http_connector.py
```diff
@@ -181,16 +185,20 @@ def query_with_session(self, statement):
...
-        return response
+        return response_list
 
     def fetch_all(self, statement):
-        # TODO use next_uri to get all results
-        resp = self.query_with_session(statement)
-        if resp is None:
+        resp_list = self.query_with_session(statement)
+        if len(resp_list) == 0 :
             log.warning("fetch all with empty results")
```
youngsofun, line 193: sometimes it is expect to get empty results？ does it happen frequently in tests?

youngsofun: add `after 6 secs` in the log to remind others it may be a slow/blocked query?

youngsofun: I suggest :

1.  warning if next URL is not  None after 6 secs
2. interval from  2 secs to 1 secs or 0.5 secs to make the whole test faster?

youngsofun (approved):

PR #5850, opened 2022-06-08
feat(query): move global configs to settings

### LINKED ISSUE #5845: Feature: move global config of async insert to session settings
**Summary**

In async insert mode, we can only load relevant config when starting which is inconvenient. Therefore, we should move the global config `wait_for_async_insert` and `wait_for_async_insert_timeout` to session settings. 

In addition to this, we should also add session setting `client_enable_async_insert` so that each client can control whether uses async insert.

cc @sundy-li @BohuTANG

fkuner:

## Summary

In async insert mode, we can only load relevant config when starting which is inconvenient. Therefore, we should move the global config `enable_async_insert`, `wait_for_async_insert` and `wait_for_async_insert_timeout` to session settings.

In addition to this, we should also add session setting `enable_async_insert` so that each client can control whether uses async insert.

After this pr, we can use like this

server:
```bash
./databen-query [--async-insert-busy-timeout=100] [--async-insert-stale-timeout=50]
```

client:
```bash
curl -u "default:" --request POST '127.0.0.1:8000/v1/query/' --header 'Content-Type: application/json' --data-raw '{"sql": "", "session": {"settings": {"enable_async_insert": "1", "wait_for_async_insert":"1", "wait_for_async_insert_timeout": "100"}}}'
```

## Changelog

- New Feature
- Improvement

## Related Issues

Fixes #5845


Xuanwo:
@mergify update

query/tests/it/storages/system/settings_table.rs
```diff
@@ -34,6 +34,7 @@ async fn test_settings_table() -> Result<()> {
         "| name                           | value   | default | level   | description                                                                                        | type   |",
         "+--------------------------------+---------+---------+---------+----------------------------------------------------------------------------------------------------+--------+",
         "|                                |         |         |         |                                                                                                    |        |",
+        "| client_enable_async_insert     | 0       | 0       | SESSION | Whether the client open async insert mode, default value: 0                                        | UInt64 |",
```
sundy-li, line 37: Maybe we should just take `enable_async_insert` as the client setting, server defaults to open async insert.

sundy-li (approved):

PR #5856, opened 2022-06-08
Improve: add schema api txn max retry times

lichuang:

## Summary

Improve: add schema api txn max retry times

## Changelog

- Improvement

## Related Issues

Fixes #issue

### DIFF
--- a/common/exception/src/exception_code.rs
+++ b/common/exception/src/exception_code.rs
@@ -192,6 +192,7 @@ build_exceptions! {
     DropTableWithDropTime(2314),
     DropDbWithDropTime(2315),
     UndropDbWithNoDropTime(2316),
+    TxnRetryMaxTimes(2317),
 
     // Cluster error codes.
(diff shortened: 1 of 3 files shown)


lichuang:
> Just for curiosity: is there any problem without limiting the retry time?

In some bad cases(like modifying the DB, Table in a high concurrency way, etc), it is possible to retry many times without success, so I think adding max retry time is more suitable in these cases.

drmingdrmer:
> > Just for curiosity: is there any problem without limiting the retry time?
> 
> In some bad cases(like modifying the DB, Table in a high concurrency way, etc), it is possible to retry many times without success, so I think adding max retry time is more suitable in these cases.

Right... this transaction model is just to serialize everything...

flaneur2020:
I think this retry limit is a good pratice, we can make this default a BIG value if the case is really rare, but the limit can save us from a dead loop if any bug runs into the loop q.q

PR #5857, opened 2022-06-08
list/remove stage files from meta api

junnplus:

## Summary

list/remove stage files from meta api.

Add a `StageFile` struct for storage stage file:
```
pub struct StageFile {
    pub path: String,
    pub size: u64,
    pub md5: Option<String>,
    pub last_modified: DateTime<Utc>,
    pub creator: Option<UserIdentity>,
}
```

## Changelog

- Improvement

## Related Issues

Fixes #issue


common/meta/types/src/user_stage.rs
```diff
@@ -227,18 +229,29 @@ pub struct UserStageInfo {
...
+                // It's internal, so we should prefix with stage name.
+                format!("/stage/{}/", self.stage_name)
+            }
         }
     }
 }
+
+#[derive(Default, Clone)]
+pub struct StageFile {
+    pub name: String,
```
Xuanwo, line 252: How about using `path` to make the behavior clear?

common/meta/types/src/user_stage.rs
```diff
@@ -227,18 +229,29 @@ pub struct UserStageInfo {
...
+    pub size: u64,
+    pub md5: String,
```
Xuanwo, line 254: The file may lack `md5` sometimes:

- On underlying storage like `fs`, `hdfs`, and so on.
- Large objects (over 5GB) on s3, azblob, and so on.

Should we take this into consideration?

query/src/storages/system/stages_table.rs
```diff
@@ -49,13 +49,38 @@ impl AsyncSystemTable for StagesTable {
         let mut copy_options: Vec<Vec<u8>> = Vec::with_capacity(stages.len());
         let mut file_format_options: Vec<Vec<u8>> = Vec::with_capacity(stages.len());
         let mut comment: Vec<Vec<u8>> = Vec::with_capacity(stages.len());
+        let mut number_of_files: Vec<Vec<u8>> = Vec::with_capacity(stages.len());
```
Xuanwo, line 52: It's hard to calculate the number of files for external stages, what value should the external stage use?

query/src/sql/statements/statement_create_user_stage.rs
```diff
@@ -61,6 +62,7 @@ impl AnalyzableStatement for DfCreateUserStage {
                     &self.location,
                     &self.credential_options,
                     &self.encryption_options,
+                    &self.endpoint_url,
```
Xuanwo, line 65: Here is a consideration of accepting the user's endpoint.

Users could input endpoints like `127.0.0.1` to try to access our internal infra. We need to detect and avoid those things.

common/meta/types/src/user_stage.rs
junnplus: It should be an optional.

common/meta/types/src/user_stage.rs
junnplus: SGTM.

query/src/storages/system/stages_table.rs
junnplus: `number_of_file`s should be optional, it is a `None/NULL` for external stages.

drmingdrmer:
It seems like it breaks the compatibility...

junnplus:
> It seems like it breaks the compatibility...

What is compatible with?

junnplus:
> The failure is emitted by the case `tests/suites/0_stateless/05_ddl/*stage_v2.*`.

I think you are referred to https://github.com/datafuselabs/databend/runs/6831188211?check_suite_focus=true , i have fixed it.

common/management/src/stage/stage_mgr.rs
```diff
@@ -110,19 +123,170 @@ impl StageApi for StageMgr {
...
-        if res.prev.is_some() && res.result.is_none() {
-            Ok(())
-        } else {
-            Err(ErrorCode::UnknownStage(format!("Unknown stage {}", name)))
+    async fn drop_stage(&self, name: &str) -> Result<()> {
+        let mut retry = 0;
+        while retry < TXN_MAX_RETRY_TIMES {
+            retry += 1;
+
+            let stage_key = format!("{}/{}", self.stage_prefix, escape_for_key(name)?);
```
lichuang, line 131: stage_key could be init outside the loop?

common/management/src/stage/stage_mgr.rs
```diff
@@ -110,19 +123,170 @@ impl StageApi for StageMgr {
...
+            let file_key_prefix = format!("{}/{}/", self.stage_file_prefix, escape_for_key(name)?);
```
lichuang, line 132: same as file_key_prefix

common/management/src/stage/stage_mgr.rs
```diff
@@ -110,19 +123,170 @@ impl StageApi for StageMgr {
...
+        while retry < TXN_MAX_RETRY_TIMES {
+            retry += 1;
+
+            let stage_key = format!("{}/{}", self.stage_prefix, escape_for_key(name)?);
+            let file_key = format!(
+                "{}/{}/{}",
+                self.stage_file_prefix,
+                escape_for_key(name)?,
+                escape_for_key(&file.path)?
+            );
```
lichuang, line 176: same as before

drmingdrmer (approved):

PR #5871, opened 2022-06-09
chore: publish logic test image

ZeaLoVe:

## Summary

Add workflow of public logic test image, including python script and use cases.

## Changelog

## Related Issues

Fixes #issue


.github/workflows/test-tool-release.yml
```diff
@@ -0,0 +1,45 @@
...
+          dockerhub_token: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: |
+            ${{ steps.login.outputs.dockerhub_repo }}:latest
+            ${{ steps.login.outputs.dockerhub_repo }}:${{ steps.get_version.outputs.VERSION }}
```
everpcpc, line 39: Maybe there's no need to publish an image for every nightly build?
A image with test dependencies would be ok, and test suites should be volumed when running.

everpcpc (approved):

ZeaLoVe:
@mergify update

Issue #5634, opened 2022-05-27
Implement project push down


leiysky:

AngleNet:
It seems that #5778 has been taken. Could you check it again for me? @leiysky

leiysky:
Closed by #5877

PR #5877, opened 2022-06-09
feature(optimizer): Refine optimizer framework

leiysky:

## Summary

Support some common optimizer rules.

## Changelog

- New Feature

## Related Issues

Part of #5634 #5633

### DIFF
--- a/query/src/sql/optimizer/s_expr.rs
+++ b/query/src/sql/optimizer/s_expr.rs
@@ -15,6 +15,8 @@
 use common_exception::ErrorCode;
 use common_exception::Result;
 
+use crate::sql::optimizer::rule::AppliedRules;
+use crate::sql::optimizer::rule::RuleID;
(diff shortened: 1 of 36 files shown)


query/src/sql/optimizer/s_expr.rs
```diff
@@ -27,6 +29,10 @@ pub struct SExpr {
     children: Vec<SExpr>,
 
     original_group: Option<IndexType>,
+
+    /// A bitmap to record applied rules on current SExpr, to prevent
+    /// redundant transformations.
+    pub(super) applied_rules: AppliedRules,
```
leiysky, line 35: @andylokandy I'd like to use a interior mutability pattern here, since it's a interior state. But can not find a proper way. How do you think of this?

Xuanwo:
@mergify update

BohuTANG (approved):

PR #5870, opened 2022-06-09
feature(query): fix read quoted string

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR

## Changelog

- Bug Fix

## Related Issues

Fixes #issue

PR #5892, opened 2022-06-10
fix(query): fix parser for incomplete data on fallback

b41sh:

## Summary

continue with pr #5870. If we deserialize `dataValue` failed and fall back to using SQL parser, the data may be incomplete.

## Changelog

- Bug Fix

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/sql/statements/value_source.rs
+++ b/query/src/sql/statements/value_source.rs
@@ -131,6 +131,8 @@ impl ValueSource {
                 for deser in desers.iter_mut().take(pop_count) {
                     deser.pop_data_value()?;
                 }
+                // rollback to get full data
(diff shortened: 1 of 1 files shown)


query/src/sql/statements/value_source.rs
```diff
@@ -131,6 +131,8 @@ impl ValueSource {
                 for deser in desers.iter_mut().take(pop_count) {
                     deser.pop_data_value()?;
                 }
+                // rollback to get full data
+                reader.rollback_to_checkpoint()?;
                 skip_to_next_row(reader, 1)?;
```
sundy-li, line 136: skip_to_next_row should be modified

b41sh: ok

PR #5896, opened 2022-06-10
feat(query): new table engine: RANDOM.

### LINKED ISSUE #5811: Feature: random Table Engine
**Summary**

This is very useful for testing.

```
create table t ( a int, b varchar, ....) Engine = Random;

select * from t limit 10; -- will generate ten rows random data .
```

Thus we can use this engine to generate random data and use them to test other features.

RinChanNOWWW:

## Summary

Implemented a new table engine: RANDOM.

select with limit:

image

select without limit:

image

## Changelog

- New Feature

## Related Issues

Fixes #5811


query/src/storages/random/random_table.rs
```diff
@@ -0,0 +1,188 @@
...
+                }
+                match push_downs.limit {
+                    Some(limit) => limit,
+                    None => block_size,
+                }
+            }
+            None => block_size,
+        };
+
+        SyncSourcer::create(ctx, output, RandomSource {
```
RinChanNOWWW, line 138: Should I implement a multi sourcer? Like:

https://github.com/datafuselabs/databend/blob/652bb413c7e5277928d7c9110ca7198bb9098023/query/src/storages/memory/memory_table.rs#L225-L238

query/src/storages/random/random_table.rs
```diff
@@ -0,0 +1,188 @@
...
+        loop {
+            for (col, pack) in packs.iter_mut().enumerate() {
+                let col_data_type = self.schema.field(col).data_type();
+                pack.append_data_value(col_data_type.random_value(), &format)?;
+            }
+
+            rows += 1;
+            self.rows += 1;
+
+            // Only generate at most `block_size` rows in one turn.
```
RinChanNOWWW, line 175: I don't know how to generate rows with exactly `block_size` of physical memory size, so I simply generate `block_size` rows one time. Is this right?

RinChanNOWWW:
Besides the comment above, I have other questions:

- I only implemented `read2`, and how should I write unit test for it?
- I don't implement `random_value` for semi-structured data types, is it OK?

common/datavalues/src/types/type_date.rs
```diff
@@ -68,6 +69,12 @@ impl DataType for DateType {
         DataValue::Int64(0)
     }
 
+    fn random_value(&self) -> DataValue {
```
sundy-li, line 72: Better to have `random_column(length: usize) -> ColumnRef`.

sundy-li: This can have better performance.

sundy-li: let it = (0..length).map(...);
let col = I32::from_own_iterator(it);

query/src/storages/random/random_table.rs
sundy-li: Yes,better to have.

RinChanNOWWW: If so, how many rows should one source generate one time? Will the sources generate rows parallelly or concurrently?

sundy-li: You can split the total rows into n Parts where n is the num of threads.

refer:  read_partitions  function  in https://github.com/datafuselabs/databend/blob/9261d9cc06d959fdca5516de7cd1381f5c9509df/query/src/table_functions/numbers_table.rs

query/src/storages/random/random_table.rs
```diff
@@ -0,0 +1,190 @@
...
+        plan: &ReadDataSourcePlan,
+        pipeline: &mut NewPipeline,
+    ) -> Result<()> {
+        let settings = ctx.get_settings();
+        let block_size = settings.get_max_block_size()? as usize;
+        let mut output_schema = self.table_info.schema();
+        let push_downs = plan.push_downs.clone();
+        // If extras.push_downs is None or extras.push_down.limit is None,
+        // set limit to `max_block_size`.
+        let limit = match push_downs {
```
sundy-li, line 91: This is incorrect, we should add these codes into `read_partitions`, thus we will have `Vec<Partitions>`.

Then we can have a `RandomSource`  like `NumbersSource`

RinChanNOWWW: OK

sundy-li (approved):

BohuTANG:
Good job, thank you! @RinChanNOWWW

PR #5530, opened 2022-05-23
test: use fuse engine instead of memory engine in test

b41sh:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

use fuse engine instead of memory engine in the test

## Changelog

- Build/Testing/CI

## Related Issues

Fixes #issue

PR #5897, opened 2022-06-10
feat(planner): migrate insert statement

fkuner:

## Summary

Summary about this PR

## Changelog

- New Feature
- Bug Fix
- Improvement
- Performance Improvement
- Build/Testing/CI
- Documentation

## Related Issues

Fixes #issue


sundy-li (approved):
Great!

tests/suites/0_stateless/03_dml/03_0016_insert_into_values_v2.sql
```diff
@@ -0,0 +1,19 @@
+set enable_planner_v2 = 1;
+
+DROP DATABASE IF EXISTS db1;
+CREATE DATABASE db1;
+USE db1;
+
+CREATE TABLE IF NOT EXISTS t1(a Int8 null, b UInt32 null, c DateTime null, d String null) Engine = Memory;
```
BohuTANG, line 7: Hmm, don't use `Engine = Memory` for the test, use `Fuse` engine by default.

fkuner: Ok, I just copy the original test file. The original test file has changed in #5530

fkuner:
The async_insert_queue will be entirely migrated after open `enable_planner_v2` default;

Xuanwo:
> The async_insert_queue will be entirely migrated after open `enable_planner_v2` default; #6124

It's so nice to create issues for the remaining work!

query/src/servers/http/clickhouse_handler.rs
```diff
@@ -207,15 +303,46 @@ pub async fn clickhouse_handler_post(
...
+            } => bind_context.format.clone(),
+            _ => None,
+        };
+
+        ctx.attach_query_str(&sql);
+        execute_v2(ctx, plan, format, None, params)
+            .await
+            .map_err(InternalServerError)
+    } else {
+        sql.push_str(body.into_string().await?.as_str());
```
youngsofun, line 336: `sql.push_str(body.into_string().await?.as_str()) `  before if else?  otherwise it will lead to conflict with a pr I am working on.

query/src/servers/http/clickhouse_handler.rs
```diff
@@ -207,15 +303,46 @@ pub async fn clickhouse_handler_post(
...
-        .await
-        .map_err(InternalServerError)
+    let (stmts, _) = DfParser::parse_sql(sql.as_str(), ctx.get_current_session().get_type())
+        .unwrap_or_else(|_| (vec![], vec![]));
+
+    let settings = ctx.get_settings();
+    if settings.get_enable_new_processor_framework().unwrap() != 0
+        && !ctx.get_config().query.management_mode
+        && ctx.get_cluster().is_empty()
+        && settings.get_enable_planner_v2().unwrap() != 0
```
leiysky, line 314:
```suggestion
        && settings.get_enable_planner_v2()? != 0
```

query/src/servers/http/clickhouse_handler.rs
fkuner: Ok.

query/src/servers/http/clickhouse_handler.rs
fkuner: Because the function returns `PoemResult<impl IntoResponse>`, we should add `map_err()?` to replace `?` 

It seems ok because `unwrap()` doesn't failed here?

leiysky: I think it's better not to use a `unwrap`, since it may fail one day, and we won't know that until the process panicked...

fkuner: Ok.

query/src/servers/http/clickhouse_handler.rs
youngsofun: ignore me leave it as it is now,  my pr is small, I will wait for this pr merged and rebase myself.

fkuner: No problem. I have modified  with `unwrap()` together.

query/src/servers/http/clickhouse_handler.rs
Xuanwo: Maybe it's better to implement `Into<PoemError> for ErrorCode`?

Xuanwo: > Maybe it's better to implement `Into<PoemError> for ErrorCode`? (not related to this PR)

Created an issue for it: https://github.com/datafuselabs/databend/issues/6137

PR #5898, opened 2022-06-10
clickhouse http handler support TsvWithNamesAndTypes.

youngsofun:

## Summary

basic support for TsvWithNamesAndTypes,

1. use FormatFactory only for `name-> enum OutputFormatType` , more freedom and less code to handle the following:
  - case_insensitive
  - alias (TSV, TabSeperated), (NDJSON, JSONEachrow)
   - "withName and withNameAndFormat" suffix
2. OutputFormatType is used as a unique id (no alias) to carry around(after sql parse) and ease ops like match, cmp, define capabilities
    -  it also make it very clear to see all supported formats in the code.
4.  use OutputFormatType to create dyn OutputFormat
5.  TsvWithNamesAndTypes implemented with constant generics,  make the code more uniform thanks to @zhang2014 

## Changelog

- New Feature

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/formats/output_format.rs
+++ b/query/src/formats/output_format.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fmt;
 use std::str::FromStr;
 
 use common_datablocks::DataBlock;
@@ -19,12 +20,18 @@ use common_datavalues::DataSchemaRef;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_io::prelude::FormatSettings;
+use strum_macros::EnumIter;
 
-use super::output_format_ndjson::NDJsonOutputFormat;
+use super::output_format_ndjson::JsonEachRowOutputFormat;
 use super::output_format_parquet::ParquetOutputFormat;
 use super::output_format_values::ValuesOutputFormat;
 use crate::formats::output_format_csv::CSVOutputFormat;
+use crate::formats::output_format_csv::CSVWithNamesAndTypesOutputFormat;
+use crate::formats::output_format_csv::CSVWithNamesOutputFormat;
 use crate::formats::output_format_csv::TSVOutputFormat;
+use crate::formats::output_format_csv::TSVWithNamesAndTypesOutputFormat;
+use crate::formats::output_format_csv::TSVWithNamesOutputFormat;
+use crate::formats::FormatFactory;
 pub trait OutputFormat: Send {
     fn serialize_block(
         &mut self,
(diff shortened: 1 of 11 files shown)


query/src/servers/http/clickhouse_handler.rs
```diff
@@ -85,13 +86,15 @@ async fn execute(
...
     let mut data_stream = ctx.try_create_abortable(data_stream)?;
     let format_setting = ctx.get_format_settings()?;
-    let mut fmt = OutputFormatType::Tsv;
+    let mut fmt = OutputFormatType::Tsv(HeaderConfig::new(false, false));
     if let Some(format) = format {
         fmt = OutputFormatType::from_str(format.as_str())?;
     }
 
     let mut output_format = fmt.create_format(plan.schema());
+    let header = output_format.serialize_header(&format_setting);
```
sundy-li, line 95: There are other places we need to call serialize_header.

Eg: download.rs, stage_table.rs

youngsofun: get the name from clickhouse
also csv used to call it header

any suggestion?

query/src/formats/output_format.rs
```diff
@@ -49,8 +68,12 @@ pub enum OutputFormatType {
 impl OutputFormatType {
     pub fn create_format(&self, schema: DataSchemaRef) -> Box<dyn OutputFormat> {
         match self {
-            OutputFormatType::Tsv => Box::new(TSVOutputFormat::create(schema)),
-            OutputFormatType::Csv => Box::new(CSVOutputFormat::create(schema)),
+            OutputFormatType::Tsv(header_config) => {
+                Box::new(TSVOutputFormat::create(schema, *header_config))
```
zhang2014, line 72: maybe use constant generics  is better

query/src/formats/output_format.rs
```diff
@@ -60,16 +83,21 @@ impl OutputFormatType {
...
 
 impl FromStr for OutputFormatType {
     type Err = ErrorCode;
     fn from_str(s: &str) -> std::result::Result<Self, ErrorCode> {
+        // todo(youngsofun):  refactor with factory
         match s.to_uppercase().as_str() {
-            "TSV" => Ok(OutputFormatType::Tsv),
-            "CSV" => Ok(OutputFormatType::Csv),
+            "TSV" => Ok(OutputFormatType::Tsv(HeaderConfig::new(false, false))),
+            "TSVWITHNAMES" => Ok(OutputFormatType::Tsv(HeaderConfig::new(true, false))),
```
zhang2014, line 96: How about new enum type? 
```
enum OutputFormatType {
    TSV,
    TSVWithName,
    TSVWithNameAndType,
    CSV,
    CSVWithName,
    CSVWithNameAndType
}
```

youngsofun: good idea!

youngsofun: @zhang2014 maybe it is time to use your FormatFactory,  more freedom and less code to handle the following:

1.   case_insensitive
2.  alias (TSV, TabSeperated), (NDJSON, JSONEache)
3.  "withName and withNameAndFormat"

query/src/servers/http/clickhouse_handler.rs
youngsofun: 'm mistaken, clickhouse use word `priefix` here,  maybe  it  is more general

query/src/formats/output_format.rs
youngsofun: my design:

1.  to use factory only for name-> OutputFormatType
2.  OutputFormatType  is used as a unique id (no alias) to carry around(after sql parse) and ease ops like match, cmp, in..
3. one function from OutputFormatType to create dyn OutputFormat

query/src/servers/http/clickhouse_handler.rs
youngsofun: renamed to prefix

youngsofun:
@sundy-li @zhang2014 refactor done

sundy-li (approved):

PR #5921, opened 2022-06-12
support ifnull feature

### LINKED ISSUE #5852: Feature: Support ifnull
**Summary**

select a, b, ifnull(a,b), ifnull(b,a) from i;

   a    |   b    | ifnull(a,b) | ifnull(b,a) |
--------|--------|-------------|-------------
 0      | 5      | 0           | 5           |
 0      | [NULL] | 0           | 0           |
 [NULL] | 5      | 5           | 5           |
 [NULL] | [NULL] | [NULL]      | [NULL]      |

https://docs.snowflake.com/en/sql-reference/functions/ifnull.html

yuuch:

## Summary

Support `IfNull` function.  

Only support `ifNull` function in the new planner mode.
## Changelog

- New Feature

## Related Issues

Close #5852


sundy-li:
I think we should not add a new function, we can transform the expression in new planner like `null_if` (https://github.com/datafuselabs/databend/blob/6254ce170f5aa454bd8f4682d3eff973bb923d2d/query/src/sql/planner/semantic/type_check.rs)

BohuTANG:
I think the function name `ifnull` is ok, not `if_null`, this will be compilable with other databases(Snow/MySQL).

common/ast/tests/it/testdata/expr.txt
```diff
@@ -2598,3 +2598,75 @@ NullIf {
 }
 
 
+---------- Input ----------
+ifnull(1, 1)
+---------- Output ---------
+IFNULL(1, 1)
+---------- AST ------------
+NullIf {
```
PsiACE, line 2606:
```suggestion
IfNull {
```

common/ast/tests/it/testdata/expr.txt
```diff
@@ -2598,3 +2598,75 @@ NullIf {
...
+        ],
+        database: None,
+        table: None,
+        column: Identifier {
+            name: "b",
+            quote: None,
+            span: Ident(10..11),
+        },
+    },
+}
```
PsiACE, line 2672: Need to add more newline as before

PsiACE (changes requested):
Also note that `or 50 more ...` in expr-error.txt needs to be replaced with `or 51 more ...` because of the addition of a new statement.

PsiACE:
Although the goldenfile output is somewhat unfriendly, it is still possible to scroll up see the difference.

a simple

There is an alternative process may help to troubleshoot such problems:

1. `REGENERATE_GOLDENFILES=1 cargo test -p common-ast --test it`
2. `git diff`
3. Compare and check carefully for any errors that are not expected

yuuch:
Screen Shot 2022-06-15 at 11 40 00 PM

PsiACE:
@mergify update

tests/suites/0_stateless/02_function/02_0058_function_ifnull.sql
```diff
@@ -0,0 +1,28 @@
+SET enable_planner_v2 = 1;
```
sundy-li, line 1: Now plan v2 not work on cluster mode, we can comment these tests temporary.

PsiACE: we can edit scripts/ci/ci-run-stateless-tests-cluster.sh to skip it

ref: https://github.com/datafuselabs/databend/pull/5958/files

sundy-li: Like `./databend-test '^0[^4]_' --mode 'cluster' --run-dir 0_stateless --skip '02_0057_function_nullif'  `

yuuch: Yes. I read the python script.  It will be: `./databend-test '^0[^4]_' --mode 'cluster' --run-dir 0_stateless --skip '02_0057_function_nullif' '02_0058_function_ifnull'`

PsiACE (approved):
lgtm

PR #5945, opened 2022-06-13
handler(clickhouse): ck http handler supports settings

### LINKED ISSUE #5832: Feature: http/ck handler support inject settings from url params
**Summary**

Description for this feature.

fkuner:

## Summary

Clickhouse http handler support settings.

## Changelog

- New Feature

## Related Issues

Fixes #5832


fkuner:
In order to support settings, I think it is better to use json?

What do you think? @youngsofun @sundy-li

youngsofun:
@fkuner  clickhouse http handler must be compatible with clickhouse  https://clickhouse.com/docs/en/interfaces/http/

fkuner:
> 

Got it. Thank you!

youngsofun:
@flaneur2020 compare with query-params, is http header better if we may inject it in the gateway?

query/src/servers/http/clickhouse_handler.rs
```diff
@@ -123,6 +124,14 @@ pub async fn clickhouse_handler_get(
         .await
         .map_err(InternalServerError)?;
 
+    if let Some(settings) = params.settings {
+        let session_settings: Vec<&str> = settings.split_terminator(',').collect();
+        for session_setting in session_settings.into_iter() {
+            let s: Vec<&str> = session_setting.split('=').collect();
+            session.get_settings().set_settings(s[0].to_string(), s[1].to_string(), false).map_err(BadRequest)?;
+        }
```
youngsofun, line 132: some suggestion: 

1.  ` map_err(BadRequest)?;`   desc the wrong setting
2.  strip of key/value strings (remove spaces)

fkuner: Good idea. What do you mean `strip of key/value strings (remove spaces)?` We should better use ' ' or '-' as separator？

fkuner:
Do you think which is better? @youngsofun 

query=***&settings=a:1,b:2,c:3

query=***&settings=a%201,b%202,c%203

query/src/servers/http/clickhouse_handler.rs
youngsofun: e.g. allow `%20` after `, `

fkuner: Got it.

fkuner:
> 

Ok.

query/src/servers/http/clickhouse_handler.rs
sundy-li: Why not follow the param rule as ClickHouse ?

`http://localhost:8123/?max_result_bytes=4000000&buffer_size=3000000&wait_end_of_query=...?

sundy-li: Why not follow the param rule as ClickHouse ?

`http://localhost:8123/?max_result_bytes=4000000&buffer_size=3000000&wait_end_of_query=...?

fkuner: This means that we put all settings in params? It seems not to maintain easily?

sundy-li: https://github.com/ClickHouse/ClickHouse/blob/8b4a6a24161c701ff7255bc14f1e39b5f896049d/src/Server/HTTPHandler.cpp

See: `SettingsChanges settings_changes;`

fkuner: Yes, I need to check whether poem can receive any undefined key-value. Current method can't do this.

youngsofun: https://serde.rs/attr-flatten.html#capture-additional-fields

query/src/servers/http/clickhouse_handler.rs
```diff
@@ -141,7 +145,17 @@ pub async fn clickhouse_handler_get(
         .await
         .map_err(InternalServerError)?;
 
+    let settings = session.get_settings();
+    for (k, v) in params.extra.iter() {
+        if settings.has_setting(k.as_str()) {
+            settings
+                .set_settings(k.to_string(), v.to_string(), false)
+                .map_err(BadRequest)?;
```
youngsofun, line 153: to not affect usage of clickhouse clients, I suggest for now not return error when meeting an unkown setting, we can log first

for 2 reasons:

1. clickhouse has many other params not included in current StatementHandlerParams, [I will add them later]
2.  settings names of clickhouse and data bend are not the same.  [may need  to modify our setting or add a mapping/translation fn later]

query/src/servers/http/clickhouse_handler.rs
```diff
@@ -164,6 +178,15 @@ pub async fn clickhouse_handler_post(
         .await
         .map_err(InternalServerError)?;
 
+    let settings = session.get_settings();
```
youngsofun, line 181: what about extract the dup code into a funtion.

query/src/servers/http/clickhouse_handler.rs
```diff
@@ -46,10 +48,12 @@ use crate::sessions::QueryContext;
 use crate::sessions::SessionType;
 use crate::sql::PlanParser;
 
-#[derive(Deserialize)]
+#[derive(Serialize, Deserialize)]
 pub struct StatementHandlerParams {
     query: Option<String>,
     compress: Option<u8>,
+    #[serde(flatten)]
+    extra: HashMap<String, String>,
```
youngsofun, line 56: directly call it settings is ok

query/src/servers/http/clickhouse_handler.rs
fkuner: make sense

query/src/servers/http/clickhouse_handler.rs
fkuner: OK

query/src/servers/http/clickhouse_handler.rs
fkuner: Ok

query/src/servers/http/clickhouse_handler.rs
youngsofun: you can add comment and test for it

fkuner: I think about it. Unkown settings can't occur error. Only the setting value is wrong, the error occurs. I think it should throw error.

fkuner: We first judge by 'has_setting()' function.

youngsofun: make sense

query/src/sessions/session_settings.rs
```diff
@@ -444,6 +444,20 @@ impl Settings {
...
         Ok(())
     }
+
+    pub fn set_batch_settings(
+        &self,
+        settings: &HashMap<String, String>,
+        is_global: bool,
+    ) -> Result<()> {
+        for (k, v) in settings.iter() {
+            if self.has_setting(k.as_str()) {
```
youngsofun, line 454: each has_setting， set_settings try get the lock
what about doing the batch upate with  one lock  cc @sundy-li

fkuner: Make sense.

fkuner: A trade-off between simple code and performance. But in this situation, it seems not be bottlelock? Generally speaking, there are no too many settings. In my opinion, I prefer keep this. What do you think?

youngsofun: lgtm

youngsofun (approved):

sundy-li (approved):

PR #5947, opened 2022-06-13
bugfix(planner): Fix correlated subquery with joins

leiysky:

## Changelog

- Bug Fix

### DIFF
--- a/query/src/sql/planner/binder/bind_context.rs
+++ b/query/src/sql/planner/binder/bind_context.rs
@@ -68,6 +68,13 @@ impl BindContext {
         }
     }
 
+    /// Create a new BindContext with self's parent as its parent
+    pub fn replace(&self) -> Self {
+        let mut bind_context = BindContext::new();
+        bind_context.parent = self.parent.clone();
+        bind_context
+    }
+
     /// Generate a new BindContext and take current BindContext as its parent.
     pub fn push(self) -> Self {
         Self::with_parent(Box::new(self))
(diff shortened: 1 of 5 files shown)


query/src/sql/planner/binder/bind_context.rs
```diff
@@ -68,6 +68,13 @@ impl BindContext {
         }
     }
 
+    /// Create a new BindContext with self's parent as its parent
+    pub fn replace(&self) -> Self {
```
leiysky, line 72: @andylokandy I'm not a big fan of this method name. Do you have a better idea?

query/src/sql/planner/binder/join.rs
```diff
@@ -49,13 +49,12 @@ impl<'a> Binder {
...
-        let (right_child, right_context) = self
-            .bind_table_reference(&left_context, &join.right)
-            .await?;
+        let (right_child, right_context) =
+            self.bind_table_reference(bind_context, &join.right).await?;
 
         check_duplicate_join_tables(&left_context, &right_context)?;
 
-        let mut bind_context = BindContext::new();
+        let mut bind_context = bind_context.replace();
```
xudong963, line 57: Seems you can directly use `with_parent`, then you don't need to think of a better function name to replace `replace`.
```rust
let mut bind_context = BindContext::with_parent(self.parent.clone());
```

leiysky: > Seems you can directly use `with_parent`, then you don't need to think of a better function name to replace `replace`.
> 
> ```rust
> let mut bind_context = BindContext::with_parent(self.parent.clone());
> ```

I was considering about that. The problem is `with_parent` takes a `Box<BindContext>` as input, while parent is a `Option<Box<BindContext>>`.

PR #4544, opened 2022-03-23
ISSUE-4471 integrate cluster select query with new processor

zhang2014:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

- Refine query fragment code
- integrate cluster with new processor
- Support push base model data exchange
- Preparing to remove the flight API(initial stage)
- Resource control (multiple query fragments will share the same resource(cpu and network) and refer to max_threads)
- Support multiple types of data transfer between cluster nodes(e.g., logs, metrics, progress, commit_entity, initial stage only support progress)

## Changelog

- New Feature

## Related Issues

Fixes #4471

## Test Plan

Unit Tests

Stateless Tests

PR #5952, opened 2022-06-13
ISSUE-5829: Support field comment.

### LINKED ISSUE #5829: Feature: Support table & field  comment
**Summary**
```
MySQL [(none)]> create table t1(a INT comment 'acomment');
ERROR 1105 (HY000): Code: 1022, displayText = COMMENT 'acomment' column option is not supported, please do not specify them in the CREATE TABLE statement.
MySQL [(none)]> select version();
+-----------------------------------------------------------------------------------------+
| @@version()                                                                             |
+-----------------------------------------------------------------------------------------+
| 8.0.26-v0.7.74-nightly-a556243-simd(rust-1.63.0-nightly-2022-06-06T16:24:09.098868778Z) |
+-----------------------------------------------------------------------------------------+
1 row in set (0.002 sec)
```
Support show create table display comment.

https://github.com/datafuselabs/databend/issues/4925

RinChanNOWWW:

## Summary

- Table comment is already supported.
- Support field comment for table under the new planner.

image

Currently, the comment is stored in `DataField`:

```rust
pub struct DataField {
    name: String,
    /// default_expr is serialized representation from PlanExpression
    default_expr: Option<String>,
    #[ignore_malloc_size_of = "insignificant"]
    data_type: DataTypeImpl,
    comment: Option<String>,
}
```

@youngsofun suggested:

> just store column comments in tableInfos in catelog . don`t need to carry it in Schema. https://github.com/datafuselabs/databend/discussions/5939#discussioncomment-2937707

I don't know which one is better. I think the size in storage and the retrieving efficiency is the same.

And I only implemented it under the new parser and planner, should I implement it in the old one? If so, it seems need to modify sqlparser-rs.

==================

2022-06-14 12:54 Update:

Put field comments in `TableMeta`:

```rust
pub struct TableMeta {
    pub schema: Arc<DataSchema>,
    pub engine: String,
    pub engine_options: BTreeMap<String, String>,
    pub options: BTreeMap<String, String>,
    pub cluster_key: Option<String>,
    // The vector of cluster keys.
    pub cluster_keys: Vec<String>,
    // The default cluster keys id.
    pub default_cluster_key_id: Option<u32>,
    pub created_on: DateTime<Utc>,
    pub updated_on: DateTime<Utc>,
    pub comment: String,
    pub field_comments: Vec<String>,

    // if used in CreateTableReq, this field MUST set to None.
    pub drop_on: Option<DateTime<Utc>>,
    pub statistics: TableStatistics,
}
```

Like MySQL, if comment is an empty string(''), ignore it:

image

## Changelog

- New Feature

## Related Issues

Fixes #5829


youngsofun:
for example, a schema is used not only for tables but also for results and intermedia results of query and used in the plan, block etc.
but the user-defined tables need comments.
even for user-defined tables,   comments are not related to any computing.
it is only used when desc table.

RinChanNOWWW:
> for example, a schema is used not only for tables but also for results and intermedia results of query and used in the plan, block etc. but the user-defined tables need comments. even for user-defined tables, comments are not related to any computing. it is only used when desc table.

I see. I will change it later.

tests/suites/0_stateless/05_ddl/05_0000_ddl_create_tables_v2.sql
```diff
@@ -78,3 +79,8 @@ DROP DATABASE db1;
 DROP DATABASE db2;
 
 CREATE TABLE system.test; -- {ErrorCode 1002}
+
+-- create table with column comment
+-- SELECT '====CREATE TABLE WITH COLUMN COMMENT====';
+-- CREATE TABLE column_comment_test (a INT COMMENT 'comment for a', b FLOAT NULL DEFAULT 0 COMMENT 'comment for b');
```
BohuTANG, line 85: Why disable the test here? If we need `enable_planner_v2` , please add a new test file for it :)

RinChanNOWWW: Or the test_stateless_cluster_* will failed.

BohuTANG: We should deep dive why the cluster has errors

RinChanNOWWW: I don't know if it is because https://github.com/datafuselabs/databend/pull/4544 is not finished.

BohuTANG: Not sure why the create table need cluster, what's the error is?

RinChanNOWWW:
```
ERROR 1105 (HY000) at line 61: Code: 1005, displayText = error: 
  --> SQL:1:14
  |
1 | create table with column comment
  | ------       ^^^^ expected <Ident>, <QuotedString>, or `IF`
  | |             
  | while parsing `CREATE TABLE [IF NOT EXISTS] [<database>.]<table> [<source>] [<table_options>]`

.
ERROR 1105 (HY000) at line 63: Code: 1022, displayText = COMMENT 'comment for a' column option is not supported, please do not specify them in the CREATE TABLE statement.
ERROR 1105 (HY000) at line 64: Code: 1025, displayText = Unknown table 'column_comment_test'.
```

Do cluster and standalone use the different parser?

RinChanNOWWW:
```
ERROR 1105 (HY000) at line 62: Code: 1022, displayText = COMMENT 'comment for a' column option is not supported, please do not specify them in the CREATE TABLE statement.
ERROR 1105 (HY000) at line 63: Code: 1025, displayText = Unknown table 'column_comment_test'.
```

This will not happen in standalone test.

Do standalone and cluster use different parsers?

RinChanNOWWW: If it is because the sqls will not be executed on the same node, and `set enable_planner_v2 = 1` don't take effect on every node?

And `CREATE TABLE column_comment_test (a INT COMMENT 'comment for a', b FLOAT NULL DEFAULT 0 COMMENT 'comment for b')` is parsed on a node with planner_v1?

leiysky: @RinChanNOWWW I think the reason is that we won't enable new planner in cluster mode for now, see https://github.com/datafuselabs/databend/blob/04a6c54fb958e424d2cacdb06eec09e0360525eb/query/src/servers/mysql/mysql_interactive_worker.rs#L301-L302

It is because new planner haven't implemented distributed query yet, which we are working on recently.

To resolve the test issue, you can create a particular result file for cluster mode, just as https://github.com/datafuselabs/databend/blob/main/tests/suites/0_stateless/04_explain/04_0000_explain_cluster.result

RinChanNOWWW:
image

It seems that the locations of error messages are different between `test_stateless_cluster_linux` and `test_stateless_cluster_macos`...

leiysky:
> If it is better to comment the tests at this time...

I think so. @BohuTANG What do you think of this?

BohuTANG:
There are conflicts files

> > If it is better to comment the tests at this time...
> 
> I think so. @BohuTANG What do you think of this?

Ok.
Conflict files :)  @RinChanNOWWW

RinChanNOWWW:
Conflicts solved.

query/src/interpreters/interpreter_table_show_create.rs
```diff
@@ -59,26 +59,37 @@ impl Interpreter for ShowCreateTableInterpreter {
...
                     Some(expr) => {
                         let expression = PlanParser::parse_expr(expr)?;
                         format!(" DEFAULT {}", expression.column_name())
                     }
                     None => "".to_string(),
                 };
+                // compatibility: creating table in the old planner will not have `fields_comments`
+                let comment = if field_comments.len() == n_fields && !field_comments[idx].is_empty()
+                {
+                    format!(" COMMENT '{}'", &field_comments[idx])
```
drmingdrmer, line 81: Does the `field_comments[idx]` need to be escaped?

common/proto-conv/tests/it/proto_conv.rs
```diff
@@ -114,6 +114,7 @@ fn new_table_info() -> mt::TableInfo {
             created_on: Utc.ymd(2014, 11, 28).and_hms(12, 0, 9),
             updated_on: Utc.ymd(2014, 11, 29).and_hms(12, 0, 10),
             comment: s("table_comment"),
+            field_comments: vec![],
```
drmingdrmer, line 117: The test has to add some field comments and check if it encodes/decodes the table meta correctly.

common/meta/app/src/schema/table.rs
```diff
@@ -183,6 +183,7 @@ pub struct TableMeta {
     pub created_on: DateTime<Utc>,
     pub updated_on: DateTime<Utc>,
     pub comment: String,
+    pub field_comments: Vec<String>,
```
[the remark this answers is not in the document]
RinChanNOWWW: > Is field_comments a vec of exactly the same number of elements as schema?

Yes.

> If it is, does it mean it does not distinguish between an empty string comment and no comment at all?

Empty string comment will be ignore, just like MySQL.

image

common/proto-conv/tests/it/proto_conv.rs
RinChanNOWWW: Ok.

query/src/interpreters/interpreter_table_show_create.rs
drmingdrmer: E.g. when there is a single quote `'` in the comment

RinChanNOWWW: Oh, I didn't consider this. I'll take a look.

RinChanNOWWW:
```sql
create table t (a int comment 'aaa\'bbb\'ccc');
show create table t;
```

MySQL's result:

```
+-------+--------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                         |
+-------+--------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` int DEFAULT NULL COMMENT 'aaa''bbb''ccc'
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+--------------------------------------------------------------------------------------------------------------------------------------+
```

MySQL split the string.

This PR's result:

```
+-------+------------------------------------------------------------------+
| Table | Create Table                                                     |
+-------+------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` INT COMMENT 'aaa'bbb'ccc'
) ENGINE=FUSE |
+-------+------------------------------------------------------------------+
```

Should we be the same with MySQL's behavior?

RinChanNOWWW:
CI failed after rebase main branch.

query/src/interpreters/interpreter_table_show_create.rs
sundy-li: It's simple to just replace `'` with `\'` ?

RinChanNOWWW: There will not be `'` in the comment, only `\'`. Because we cannot write `'aaa'bbb'ccc'` in SQL, but only can write `'aaa\'bbb\'ccc'`. The problem now is if we should follow MySQL style, or just keep the style above.

sundy-li: I think we do not need to follow MySQL style,  `'aaa''bbb''ccc'` is hard to read.

sundy-li: But we should be able to create a new table with this format SQL.

RinChanNOWWW: You means `create table t (a int comment 'aaa''bbb''ccc');` ? This will fail during the parse phase.

common/proto-conv/tests/it/proto_conv.rs
```diff
@@ -236,37 +237,48 @@ fn test_load_old() -> anyhow::Result<()> {
...
-            10, 14, 118, 97, 114, 105, 97, 110, 116, 95, 111, 98, 106, 101, 99, 116, 26, 9, 162, 1,
-            3, 160, 6, 1, 160, 6, 1, 160, 6, 1, 10, 26, 10, 8, 105, 110, 116, 101, 114, 118, 97,
-            108, 26, 11, 170, 1, 5, 8, 2, 160, 6, 1, 160, 6, 1, 160, 6, 1, 18, 6, 10, 1, 97, 18, 1,
-            98, 160, 6, 1, 42, 10, 10, 3, 120, 121, 122, 18, 3, 102, 111, 111, 50, 2, 52, 52, 58,
-            10, 10, 3, 97, 98, 99, 18, 3, 100, 101, 102, 74, 10, 40, 97, 32, 43, 32, 50, 44, 32,
-            98, 41, 162, 1, 23, 50, 48, 49, 52, 45, 49, 49, 45, 50, 56, 32, 49, 50, 58, 48, 48, 58,
-            48, 57, 32, 85, 84, 67, 170, 1, 23, 50, 48, 49, 52, 45, 49, 49, 45, 50, 57, 32, 49, 50,
-            58, 48, 48, 58, 49, 48, 32, 85, 84, 67, 178, 1, 13, 116, 97, 98, 108, 101, 95, 99, 111,
-            109, 109, 101, 110, 116, 160, 6, 1, 160, 6, 1,
+            10, 10, 8, 5, 16, 6, 160, 6, 1, 168, 6, 1, 18, 3, 102, 111, 111, 26, 3, 98, 97, 114,
```
drmingdrmer, line 240: These data is for compatibility test that new program can load old data. Do not modify it.

RinChanNOWWW: Fine

drmingdrmer (approved):
The meta data part looks good to me.

BTW, you can increase the `VER` to indicate it's a new version of meta data. It is not mandatory: not changing it means it has to be compatible with all changes happened in this `VER=1`:
https://github.com/datafuselabs/databend/blob/400fd52134cba2cc09eacad4d6a8d18f914b8410/common/proto-conv/src/util.rs#L17

PR #5957, opened 2022-06-13
Create dml-join.md

soyeric128:

## Summary

added doc for JOINs

## Changelog

- Documentation

## Related Issues

Fixes #issue

### DIFF
--- a/docs/doc/30-reference/30-sql/20-query-syntax/dml-join.md
+++ b/docs/doc/30-reference/30-sql/20-query-syntax/dml-join.md
@@ -1,0 +1,129 @@
+---
+title: JOIN
+---
+
+A *join* allows you to combine columns from two or more tables into a single result set. Databend supports the following *join* types:
+
+* Inner Join
+* Natural Join
+* Cross Join
+
+## Inner Join
+
+An *inner join* returns the rows that meet the join conditions in the result set.
+
+### Syntax
+
+```sql    
+SELECT select_list
+FROM table_a
+	[INNER] JOIN table_b
+		ON join_condition_1
+	[[INNER] JOIN table_c
+		ON join_condition_2]...
+```
+:::tip
+The INNER keyword is optional.
+:::
+
+When you join two tables on a common column with the equal operator, you can use the USING keyword to simplify the syntax.
+
+```sql    
+SELECT select_list
+FROM table_a
+	JOIN table_b
+		USING join_column_1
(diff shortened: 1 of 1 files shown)


BohuTANG:
join is a more complex operation. We'd better find a more realistic example to demonstrate them.

docs/doc/30-reference/30-sql/20-query-syntax/dml-join.md
```diff
@@ -0,0 +1,129 @@
+---
+title: JOIN
+---
+
+A *join* allows you to combine columns from two or more tables into a single result set. Databend supports the following *join* types:
+
```
xudong963, line 6: Better to have a general structure, like:
```sql
SELECT <expr_list>
FROM <left_table>
[GLOBAL] [INNER|LEFT|RIGHT|FULL|CROSS] [OUTER|SEMI|ANTI|ANY|ASOF] JOIN <right_table>
(ON <expr_list>)|(USING <column_list>) ...
```
From clickhouse doc

xudong963:
> join is a more complex operation. We'd better find a more realistic example to demonstrate them.

I agree. realistic table + realistic query. Do you have a dataset you prefer?

PsiACE:
> I agree. realistic table + realistic query. Do you have a dataset you prefer?

I think it's fine to keep using ontime?

docs/doc/30-reference/30-sql/20-query-syntax/dml-join.md
soyeric128: Can we have this structure until when we support them all ?

BohuTANG:
> > join is a more complex operation. We'd better find a more realistic example to demonstrate them.
> 
> I agree. realistic table + realistic query. Do you have a dataset you prefer?

`Orders` and `Customers`  is ok, there are many cases:
https://www.w3schools.com/sql/sql_join.asp#:~:text=Different%20Types%20of%20SQL%20JOINs&text=(INNER)%20JOIN%20%3A%20Returns%20records,records%20from%20the%20left%20table

BohuTANG:
> > I agree. realistic table + realistic query. Do you have a dataset you prefer?
> 
> I think it's fine to keep using ontime?

Please don't use that :D

soyeric128:
Examples improved in the latest commit.

xudong963 (approved):
Would you like to take another look?

PR #5968, opened 2022-06-14
feat: transient fuse table

dantengsky:

## Summary

The new statement `CREATE TRANSIENT TABLE...`:
- for FUSE table, during commit, no historical data will be kept
  or more precisely: GC will be triggered immediately after the commit
- for table engine other than fuse, nothing different

example: 

~~~
mysql> use default;
Database changed
mysql> CREATE TRANSIENT TABLE t (c bigint);
Read 0 rows, 0.00 B in 0.006 sec., 0 rows/sec., 0.00 B/sec.

Read 1 rows, 8.00 B in 0.011 sec., 91.64 rows/sec., 733.14 B/sec.

Read 1 rows, 8.00 B in 0.013 sec., 78.42 rows/sec., 627.38 B/sec.

Read 1 rows, 8.00 B in 0.013 sec., 77.94 rows/sec., 623.53 B/sec.

mysql> select count(*) from fuse_snapshot('default', 't');
+---------+
| count() |
+---------+
|       1 |    <-- ***instead of 3 snapshots, only one snapshot is kept***
+---------+
1 row in set (0.06 sec)
Read 1 rows, 211.00 B in 0.008 sec., 129.64 rows/sec., 26.71 KiB/sec.
----
~~~

- [x] transient table commit
- [x] parser, planner (legacy and v2)
- [x] ut and it
- [x] concurrent insertion test
   shamefully done manually

## Changelog

- New Feature
- Documentation

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/storages/fuse/operations/gc.rs
+++ b/query/src/storages/fuse/operations/gc.rs
@@ -17,6 +17,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use common_cache::Cache;
+use common_exception::ErrorCode;
 use common_exception::Result;
 use futures::TryStreamExt;
 use opendal::Operator;
@@ -29,7 +30,15 @@ use crate::storages::fuse::FuseTable;
 
 impl FuseTable {
     pub async fn do_gc(&self, ctx: &Arc<QueryContext>, keep_last_snapshot: bool) -> Result<()> {
-        let snapshot_opt = self.read_table_snapshot(ctx.as_ref()).await?;
+        let r = self.read_table_snapshot(ctx.as_ref()).await;
+        let snapshot_opt = match r {
+            Err(e) if e.code() == ErrorCode::storage_not_found_code() => {
+                // concurrent gc: someone else has already collected this snapshot, ignore it
+                return Ok(());
+            }
+            Err(e) => return Err(e),
+            Ok(v) => v,
+        };
 
         let last_snapshot = if let Some(s) = snapshot_opt {
             s
@@ -130,8 +139,16 @@ impl FuseTable {
         let reader = MetaReaders::segment_info_reader(ctx);
         for l in segments {
             let (x, ver) = l;
-            let res = reader.read(x, None, *ver).await?;
-            for block_meta in &res.blocks {
+            let r = reader.read(x, None, *ver).await;
+            let segment_info = match r {
+                Err(e) if e.code() == ErrorCode::storage_not_found_code() => {
+                    // concurrent gc: someone else has already collected this segment, ignore it
+                    continue;
+                }
+                Err(e) => return Err(e),
+                Ok(v) => v,
+            };
+            for block_meta in &segment_info.blocks {
                 result.insert(block_meta.location.0.clone());
             }
         }
@@ -163,6 +180,20 @@ impl FuseTable {
     }
 
     async fn remove_location(
+        &self,
+        data_accessor: &Operator,
+        location: impl AsRef<str>,
+    ) -> Result<()> {
+        match self.do_remove_location(data_accessor, location).await {
+            Err(e) if e.code() == ErrorCode::storage_not_found_code() => Ok(()),
+            Err(e) => Err(e),
+            Ok(_) => Ok(()),
+        }
+    }
+
+    // make type checker happy
+    #[inline]
+    async fn do_remove_location(
         &self,
         data_accessor: &Operator,
         location: impl AsRef<str>,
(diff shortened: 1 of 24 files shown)


sundy-li:
Why not have a new table options?

dantengsky:
> Why not have a new table options?

the new keyword  `transient`? just a syntax sugar. underlying, it IS a new table option

https://github.com/datafuselabs/databend/blob/ac3d9554c6f1a208450b930c9743c906b7cf9fa8/query/src/sql/parsers/parser_table.rs#L79-L81

BohuTANG:
> > Why not have a new table options?
> 
> the new keyword `transient`? just a syntax sugar. underlying, it IS a new table option
> 
> https://github.com/datafuselabs/databend/blob/ac3d9554c6f1a208450b930c9743c906b7cf9fa8/query/src/sql/parsers/parser_table.rs#L79-L81

Can we also create a transient fuse table with the options with the old planner?

query/src/storages/fuse/operations/gc.rs
```diff
@@ -29,7 +30,15 @@ use crate::storages::fuse::FuseTable;
 
 impl FuseTable {
     pub async fn do_gc(&self, ctx: &Arc<QueryContext>, keep_last_snapshot: bool) -> Result<()> {
-        let snapshot_opt = self.read_table_snapshot(ctx.as_ref()).await?;
+        let r = self.read_table_snapshot(ctx.as_ref()).await;
+        let snapshot_opt = match r {
+            Err(e) if e.code() == ErrorCode::storage_not_found_code() => {
+                // concurrent gc: someone else has already collected this snapshot, ignore it
```
BohuTANG, line 36: We can have a warning log here to know this is a concurrent gc.

query/src/storages/fuse/operations/gc.rs
```diff
@@ -130,8 +139,16 @@ impl FuseTable {
         let reader = MetaReaders::segment_info_reader(ctx);
         for l in segments {
             let (x, ver) = l;
-            let res = reader.read(x, None, *ver).await?;
-            for block_meta in &res.blocks {
+            let r = reader.read(x, None, *ver).await;
+            let segment_info = match r {
+                Err(e) if e.code() == ErrorCode::storage_not_found_code() => {
+                    // concurrent gc: someone else has already collected this segment, ignore it
```
BohuTANG, line 145: ditto

query/src/storages/fuse/operations/gc.rs
```diff
@@ -166,6 +183,20 @@ impl FuseTable {
         &self,
         data_accessor: &Operator,
         location: impl AsRef<str>,
+    ) -> Result<()> {
+        match self.do_remove_location(data_accessor, location).await {
+            Err(e) if e.code() == ErrorCode::storage_not_found_code() => Ok(()),
```
BohuTANG, line 188: ditto

query/src/storages/fuse/operations/gc.rs
dantengsky: > Can we also create a transient fuse table with the options with the old planner?


> Can we also create a transient fuse table with the options with the old planner?

yeah. both planners are all supported

https://github.com/datafuselabs/databend/blob/ac3d9554c6f1a208450b930c9743c906b7cf9fa8/tests/suites/0_stateless/05_ddl/05_0000_ddl_create_tables.sql#L78

https://github.com/datafuselabs/databend/blob/ac3d9554c6f1a208450b930c9743c906b7cf9fa8/tests/suites/0_stateless/05_ddl/05_0000_ddl_create_tables_v2.sql#L78

BohuTANG (approved):

dantengsky:
> Can we also create a transient fuse table with the options with the old planner?

Yes, we do

soyeric128:
Hi @dantengsky 
Do the transient tables support time travel? 
It seems not as historical snapshots are not kept.

dantengsky:
> Hi @dantengsky Do the transient tables support time travel? It seems not as historical snapshots are not kept.

yes, exactly! 
- Technically (but meaningless), it supports time travel, but can only access the latest snapshot (by snapshot_id or timestamp)
- `undrop` do work for transient tables

soyeric128:
> > Hi @dantengsky Do the transient tables support time travel? It seems not as historical snapshots are not kept.
> 
> yes, exactly!
> 
> * Technically (but meaningless), it supports time travel, but can only access the latest snapshot (by snapshot_id or timestamp)
> * `undrop` do work for transient tables

Thanks, pls review this PR for doc: https://github.com/datafuselabs/databend/pull/6608

Issue #5501, opened 2022-05-21
Feature: support right outer join 


xudong963:
**Summary**

Description for this feature.

PR #5972, opened 2022-06-14
feature(planner): support left outer join and right outer join

### LINKED ISSUE #5500: Feature: support left outer join
**Summary**

Description for this feature.

xudong963:

## Summary

support left outer join and right outer join for equi-join

## Changelog

- New Feature

## Related Issues

Fixes #5500 #5501


common/datablocks/src/data_block.rs
```diff
@@ -211,8 +211,7 @@ impl DataBlock {
         let mut serializers = vec![];
         for col_index in 0..columns_size {
             let column = self.column(col_index);
-            let field = self.schema().field(col_index);
-            let data_type = field.data_type();
+            let data_type = column.data_type();
```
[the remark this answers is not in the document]
xudong963: During I implemented outer join, I found the mysql service will hang, but `do_query` already returns data_block.

The root reason is in data block when column data type is nullable, but schema's field data type isn't nullable, then mysql service will hang, so we can just use the column data type to create `TypeSerializerImpl`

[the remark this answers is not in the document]
leiysky: The problem is sometime `DataField`'s data type doesn't fit with the column's data type, which may cause it choose the wrong serializer. 

While the mismatching behaviour is expected in some cases, since the data type of a column represents physical type, e.g. for `Date` columns the logical type is `Date` and the physical type is `Int32`.

Fortunately, this patch works well because it will always choose the right serializer when using physical type.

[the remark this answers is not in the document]
xudong963: > While the mismatching behaviour is expected in some cases, since the data type of a column represents physical type, e.g. for `Date` columns the logical type is `Date` and the physical type is `Int32`.

Yes, outer join also needs the mismatching behavior, in databend, when creating a table, the table fields are not nullable by default. But in outer join, they may need null data values in column.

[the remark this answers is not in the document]
leiysky: > > While the mismatching behaviour is expected in some cases, since the data type of a column represents physical type, e.g. for `Date` columns the logical type is `Date` and the physical type is `Int32`.
> 
> Yes, outer join also needs the mismatching behavior, in databend, when creating a table, the table fields are not nullable by default, but in outer join, they may need null data values in columns.

It's better to cast the columns to nullable.

query/src/sql/planner/binder/join.rs
```diff
@@ -105,15 +105,27 @@ impl<'a> Binder {
...
+                right_child,
+                left_child,
+            ),
+            JoinOperator::FullOuter => self.bind_join_with_type(
+                JoinType::Full,
+                left_join_conditions,
+                right_join_conditions,
+                left_child,
+                right_child,
+            ),
```
leiysky, line 128: Seems the full outer join hasn't been implemented yet?

common/datablocks/src/data_block.rs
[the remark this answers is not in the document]
xudong963: > It's better to cast the columns to nullable.

I have cast columns to nullable.

query/src/sql/planner/binder/join.rs
xudong963: Yes

leiysky: Maybe we should return an error here, otherwise it will panic in `probe_join` if user writes down a `FULL OUTER JOIN`.

query/src/sql/planner/metadata.rs
```diff
@@ -138,6 +138,12 @@ impl Metadata {
         column_index
     }
 
+    pub fn nullable_data_type(&mut self, index: IndexType) {
```
leiysky, line 141: We need to generate a new `ColumnEntry`. This may cause the data type of a not null column is mistake changed to nullable.

For example:
```sql
select t1.number from numbers(1) t  left join (select * from numbers(1) where number is null) t1 using(number)
```

The `t1.number` should be not null inside subquery `t1`. But after applying this patch, it'll be changed to nullable, which may potentially bring some bugs.

xudong963: > We need to generate a new `ColumnEntry`

If we add a new `ColumnEntry`, how to decide the new `ColumnEntry`'s name? I think it can't be the same as the old name, otherwise it'll be result in name ambiguous

leiysky: > I think it can't be the same as the old name, otherwise it'll be result in name ambiguous

I think it's fine to use the same name, since the outer join operator will eliminate old columns and produce new columns instead. It won't cause ambiguity.

You can do this as:
1. Check if the data type of a column changes after outer join. If it doesn't,  just pass it through outer join. If it does, go to 2.
2. Use `Metadata::add_column` to generate a new `ColumnEntry` for the column with original name and changed data type. Go to 3
3. Replace the column bindings of the inner table(e.g. right table in `LEFT JOIN`) in output `BindContext` with the column entries generated in step 2.

xudong963:
@mergify update

query/src/sql/planner/metadata.rs
leiysky: I see. IMO, we can store the output columns inside `OuterJoin` struct, and produce the corresponding output schema with hash join processor.

But whatever, it's time to refactor the executor, let's just merge this PR with your patch.

leiysky (approved):

PR #5998, opened 2022-06-15
fix: Check all proto files if changed

Xuanwo:

## Summary

fix: Check all proto files if changed

Before, we will not rebuild if `config.proto` changed.

## Changelog

- Bug Fix


drmingdrmer (approved):
1 LOC reduced:D

Issue #5912, opened 2022-06-12
Prepared TPCH data set for databend


leiysky:
Support TPCH dbgen for databend, and find a way to ingest massive test data efficiently.

PR #6024, opened 2022-06-16
feat(script): generate tpch data set

xudong963:

## Summary

Generate tpch date set for databend

## Changelog

- New Feature

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/5912


PsiACE:
Do we need a workflow to publish the docker image for tpch? Alternatively, is there a readily available image, which would reduce maintenance work?

xudong963:
> Do we need a workflow to publish the docker image for tpch? Alternatively, is there a readily available image, which would reduce maintenance work?

We don't need a workflow to publish. The image is used to generate the date set in local, the run tpch related benchmark(The related code is developing).

everpcpc:
where would this tool be used?

everpcpc:
We could install this tool to our build-tool image if it is need in CI.

xudong963:
> Where would this tool be used?

Used by generating tpch data set. Then run tpch benchmark with the data set.
image

xudong963:
> We could install this tool to our build-tool image if it is need in CI.

I think it doesn't need in CI. cc @leiysky Do we have a plan to run tpch in CI to detect performance regression?

Xuanwo:
How about putting them under `scripts`?

Xuanwo:
> Do we have a plan to run tpch in CI to detect performance regression?

Please consider do bench via [databend-perf](https://github.com/datafuselabs/databend-perf/). Our CI resources are very limited and not suitable for those workload.

xudong963:
> How about putting them under `scripts`?

I thought about putting it under `scripts` before, but in the future, I want to put tpch benchmark related code under `tpch`

tpch/run-tpch-dbgen.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
```
Xuanwo, line 2: Please add a README under `tpch/` to describe how to execute those tests.

tpch/tpchdata.dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:22.04
```
Xuanwo, line 1: We must run those tests in docker?

xudong963: Use a uniform env to generate data and avoid all kinds of problems. For example, It will not success to run `make` for https://github.com/databricks/tpch-dbgen.git in my macos : D

Xuanwo: > Use a uniform env to generate data and avoid all kinds of problems.

This is the problem `dev-setup.sh` want to address. Please consider adding them in `dev-setup.sh` so that we can reuse the docker file, image, tools...

tpch/tpchdata.dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN apt-get update && \
+    apt-get install -y git build-essential
+
+# Use https://github.com/databricks/tpch-dbgen to generate data
+RUN git clone https://github.com/databricks/tpch-dbgen.git && cd tpch-dbgen && make
```
leiysky, line 7: Will it generate parquet files?

tpch/run-tpch-dbgen.sh
```diff
@@ -0,0 +1,5 @@
...
+cd /tpch-dbgen
+./dbgen -vf -s 1
```
leiysky, line 4: Maybe we can allow user to specify scale factor?

xudong963: yes👍

tpch/tpchdata.dockerfile
leiysky: Ok, I thought databricks has supported generate parquet data, but it seems they didn't...

xudong963:
Updated, please take another look @Xuanwo @leiysky @everpcpc

benchmark/tpch/README.md
```diff
@@ -0,0 +1,11 @@
+# DataBend TPCH-Benchmark
```
BohuTANG, line 1:
```suggestion
# Databend TPCH-Benchmark
```

leiysky (approved):
LGTM, maybe we can add this to https://github.com/leiysky/tpch-databend as well.

PsiACE (approved):

PR #5922, opened 2022-06-12
Add funciton coalesce 

TianLangStudio:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Add funciton coalesce  WIP

## Changelog

- New Feature

## Related Issues

Fixes #5768

PR #6039, opened 2022-06-17
feature(query): add multi_if function

### LINKED ISSUE #5951: Feature: Support `CASE ... WHEN` expression
`CASE` expression is a conditional expression with multiple branches.

You can specify branches with `WHEN ... THEN ...` subexpressions, and end the `CASE` expression with `END`.

Besides, you can define a default branch with `ELSE`, otherwise the default value is `NULL`.

For example:
```sql
select case cond when 1 then 1 when 2 then 2 end;
```

cc @BohuTANG @sundy-li

sundy-li:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues

Fixes #5951

Related #5922


tests/suites/0_stateless/20+_others/20_0001_planner_v2.sql
```diff
@@ -178,6 +178,13 @@ insert into temp values (1);
 select a from temp;
 drop table temp;
 
+
+-- CASE WHEN
+select '=== Test CASE-WHEN ===';
+select count_if(a = '1'), count_if(a = '2'), count_if(a = '3'), count_if(a is null) from (
+	SELECT (CASE WHEN number % 4 = 1 THEN '1' WHEN number % 4 = 2 THEN '2' WHEN number % 4 = 3 THEN '3' END) as a FROM numbers(100)
+);
```
leiysky, line 186: Maybe we can add some test cases for `ELSE`?

For example:
```sql
select case 1 when 2 then 3 else 4 end;
```

leiysky: By the way, we should add some test cases for the situation that data types of braches are different(numeric, string, nullability) as well.

common/functions/src/scalars/conditionals/multi_if.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        if !(args.len() >= 3 && args.len() % 2 == 1) {
+            return Err(ErrorCode::NumberArgumentsNotMatch(
+                "Invalid number of arguments for function multi_if".to_string(),
+            ));
+        }
+
+        let mut type_of_branches = Vec::with_capacity(args.len() / 2);
+        // branch
```
b41sh, line 49: useless comment.

common/functions/src/scalars/conditionals/multi_if.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    // multi_if(c1, a1, c2, a2, c3, a3, c4, a4, d)
+    fn eval(
+        &self,
+        func_ctx: FunctionContext,
+        columns: &ColumnsWithField,
+        input_rows: usize,
+    ) -> Result<ColumnRef> {
+        let mut condition_columns = Vec::with_capacity(columns.len() / 2);
+        // branch
```
b41sh, line 161: Ditto

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -265,6 +265,32 @@ impl<'a> TypeChecker<'a> {
...
             }
 
+            Expr::Case {
+                span,
+                conditions,
+                results,
+                else_result,
+                ..
+            } => {
+                let mut arguments = Vec::with_capacity(conditions.len() * 2);
```
b41sh, line 275: `arguments` capacity should be `conditions.len() * 2 + 1`

docs/doc/30-reference/20-functions/11-conditional-functions/multi_if.md
```diff
@@ -0,0 +1,49 @@
+---
+title: MULTI_IF
```
leiysky, line 2: Shall we add a documentation for `CASE WHEN` as well?

sundy-li: It could be done after we enable new planner.

leiysky (approved):

docs/doc/30-reference/20-functions/11-conditional-functions/multi_if.md
BohuTANG: We can add `CASE WHEN` to the doc with the settings `SET enable_planner_v2 = 1;`, like https://databend.rs/doc/reference/functions/conditional-functions/nullif

cc @soyeric128

soyeric128: Hi @BohuTANG and @sundy-li 
Pls review this PR: https://github.com/datafuselabs/databend/pull/6133

Issue #6038, opened 2022-06-17
[bug]: schema mismatch in new planner 


sundy-li:
**Summary**

```
databend > select  count(a = '3' )    from  (   select concat(null, 'a')  as a from numbers(10)   );
+------------------+
| count((a = '3')) |
+------------------+
|                0 |
+------------------+
1 row in set (0.036 sec)

databend > set enable_planner_v2 = 1;
Query OK, 0 rows affected (0.031 sec)
Read 0 rows, 0.00 B in 0.003 sec., 0 rows/sec., 0.00 B/sec.

databend > select  count(a = '3' )    from  (   select concat(null, 'a')  as a from numbers(10)   );
ERROR 1105 (HY000): Code: 1104, displayText = assertion failed: schema.fields().iter().zip(columns.iter()).all(|(f, c)|\n        f.data_type().data_type_id().to_physical_type() ==\n            c.data_type().data_type_id().to_physical_type()).
databend >

```

xudong963:
I beat the bug! **NULL is the root of evil**

**The following is bug tracking.**
Let's see the explanation of the example query:
```sql
mysql> explain select  count(a = '3' )    from  (   select null  as a from numbers(10)   );
+---------------------------------------------------------------------------+
| explain                                                                   |
+---------------------------------------------------------------------------+
| Project: [count(a = '3')]                                                 |
|     EvalScalar: [count(a = '3')]                                          |
|         Aggregate: group items: [], aggregate functions: [count(a = '3')] |
|             EvalScalar: [a = 3]                                           |
|                 Project: [a]                                              |
|                     EvalScalar: [NULL]                                    |
|                         Scan: default.system.numbers                      |
+---------------------------------------------------------------------------+
7 rows in set (0.05 sec)
Read 0 rows, 0.00 B in 0.001 sec., 0 rows/sec., 0.00 B/sec.
```

`Project: [a]`'s input schema and output schema datatype is **NULL**

`EvalScalar: [a = 3]`'s input schema is *NULL**, but what should be the datatype of its output shcema?

Let's debug the information:
```
[common/datablocks/src/data_block.rs:37] schema.clone() = DataSchema {
    fields: [
        DataField {
            name: "\"a\"_1",
            data_type: Null,
            nullable: false,
        },
        DataField {
            name: "\"count_arg_0\"_4",
            data_type: Boolean,
            nullable: false,
        },
    ],
    metadata: {},
}
[...]

PR #6043, opened 2022-06-17
fix(planner): consider NULL  for binary op in type checker

xudong963:

## Summary

As the title

## Changelog

- Bug Fix

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/6038


query/src/sql/planner/plans/scalar.rs
```diff
@@ -410,7 +411,13 @@ pub struct ComparisonExpr {
 
 impl ScalarExpr for ComparisonExpr {
     fn data_type(&self) -> DataTypeImpl {
-        BooleanType::new_impl()
+        let mut data_type = BooleanType::new_impl();
+        if self.left.data_type() == DataTypeImpl::Null(NullType {})
```
sundy-li, line 415: But we should also have `nullable(BooleanType)`

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -650,15 +651,20 @@ impl<'a> TypeChecker<'a> {
...
+                }
                 Ok((
                     ComparisonExpr {
                         op,
                         left: Box::new(left),
                         right: Box::new(right),
                     }
                     .into(),
-                    BooleanType::new_impl(),
+                    data_type,
```
xudong963, line 667: I took a general look and this return value should not be needed. So https://github.com/datafuselabs/databend/pull/6043/files#diff-37b754ad52b168f02e969eb2486df14bde8efb8c819a218053a1dcf79bff6b50R654-R659 will not be needed

sundy-li:
A test to cover : 

```
set enable_planner_v2 = 1;
create table n( a int null, b int null) ;
insert into n select  if (number % 3, null, number), if (number % 2, null, number) from numbers(10);

select a + b, a and b, a - b, a or b from n;

set enable_planner_v2 = 0;
 select a + b, a and b, a - b, a or b  from n;
```

xudong963:
All binary operators need to consider `NULL`, such as BinaryOperator::Plus, BinaryOperator::Minus, and so on. But part of them will be converted to `function`, so functions also need to consider `NULL`.

query/src/sql/planner/plans/scalar.rs
xudong963: For expr, do we need to consider `nullable`? Could you please give a specific case?

query/src/sql/planner/semantic/type_check.rs
```diff
@@ -650,15 +651,20 @@ impl<'a> TypeChecker<'a> {
                 let op = ComparisonOp::try_from(op)?;
                 let (left, _) = self.resolve(left, None).await?;
                 let (right, _) = self.resolve(right, None).await?;
-
+                let mut data_type = BooleanType::new_impl();
+                if left.data_type() == DataTypeImpl::Null(NullType {})
+                    || right.data_type() == DataTypeImpl::Null(NullType {})
+                {
+                    data_type = NullType::new_impl();
+                }
```
leiysky, line 659: You can write as:
```rust
let func = FunctionFactory::instance().get(op, args)?;
let result_type = func.return_type();
```
And use the `result_type` as the `data_type`.

query/src/sql/planner/semantic/type_check.rs
leiysky: Not exactly, this is part of the original type check algorithm. But since we decided to refactor the whole type system, so we didn't finish this.

It's better to keep that for now.

query/src/sql/planner/plans/scalar.rs
leiysky: You can add a `return_type` field for `ComparisonExpr` to store the resolved return type as `FunctionCall`.

query/src/sql/planner/semantic/type_check.rs
xudong963: done

sundy-li (approved):

PR #6056, opened 2022-06-18
feat(query): support `SELECT ... FROM ... { AT TIMESTAMP }` on planner_v2

### LINKED ISSUE #5733: SELECT ... FROM ... { AT (TIMESTAMP|) }
> Do we plan to support `SELECT ... FROM ... { AT (TIMESTAMP|) } ` ?

yes,  storage layer is ready. 

https://github.com/datafuselabs/databend/blob/e164aa23933781abb757b0666b655461481dcdbb/query/src/storages/fuse/operations/navigate.rs#L32-L36

in this pr, syntax that is most easily to be implemented  and tested is chose :)

_Originally posted by @dantengsky in https://github.com/datafuselabs/databend/issues/5617#issuecomment-1138586359_

cadl:

## Summary

Support `SELECT ... FROM ... { AT TIMESTAMP => <timestamp>}` like [showflake](https://docs.snowflake.com/en/sql-reference/constructs/at-before.html):

```
mysql> select * from t12_0004 at ( TIMESTAMP => cast('2022-06-16 23:07:01.580417', timestamp) );
+------+
| c    |
+------+
|    3 |
|    1 |
|    2 |
+------+
3 rows in set (0.03 sec)

mysql> select * from t12_0004 at ( TIMESTAMP => now() );
+------+
| c    |
+------+
|    1 |
|    2 |
|    3 |
|    3 |
|    3 |
+------+
5 rows in set (0.03 sec)
Read 5 rows, 20.00 B in 0.003 sec., 1.45 thousand rows/sec., 5.67 KiB/sec.
```

Add timestamp column for `fuse_snapshot` function:

```
mysql> select timestamp from fuse_snapshot('default','t12_0004');
+----------------------------+
| timestamp                  |
+----------------------------+
| 2022-06-17 00:00:34.251824 |
| 2022-06-17 00:00:21.430472 |
| 2022-06-16 23:07:01.580417 |
| 2022-06-16 23:06:56.917064 |
+----------------------------+
4 rows in set (0.03 sec)
```

## Changelog

- New Feature

## Related Issues

Fixes #5733


tests/suites/0_stateless/12_time_travel/12_0004_time_travel_select_at.sh
```diff
@@ -22,5 +26,14 @@ echo "select count(*) from t12_0004 at (snapshot => '$SNAPSHOT_ID')" | $MYSQL_CL
 echo "planner_v2: counting the data set of first insertion, which should contains 2 rows"
 echo "set enable_planner_v2 = 1;select count(t.c) from t12_0004 at (snapshot => '$SNAPSHOT_ID') as t" | $MYSQL_CLIENT_CONNECT
 
+
+# Get a time point at/after the first insertion.
+# Cannot directly use the timestamp of first insertion, because the time precision of snapshot is higher than timestamp(6). So +1μs.
+
+TIMEPOINT=$(echo "select cast(timestamp as BIGINT) + 1 from fuse_snapshot('default', 't12_0004') where row_count=2" | $MYSQL_CLIENT_CONNECT)
```
cadl, line 33: I found the precision of snapshot timestamp is nanoseconds, and I can only get `TIMESTAMP(6)` from `fuse_snapshot()`. So add 1 microsecond to the result, is it OK?

tests/suites/0_stateless/12_time_travel/12_0004_time_travel_select_at.sh
```diff
@@ -8,6 +8,10 @@ CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 echo "create table t12_0004(c int)" | $MYSQL_CLIENT_CONNECT
 echo "two insertions"
 echo "insert into t12_0004 values(1),(2)" | $MYSQL_CLIENT_CONNECT
+
+# Ensure there is a small interval between two insertions, for time travel
+sleep 0.001
```
dantengsky, line 13: will this test case fail without sleeping? the timestamp of the table snapshot is supposed to be monotonically increased even if the local clock drifting presents

tests/suites/0_stateless/12_time_travel/12_0004_time_travel_select_at.sh
dantengsky: It seems to be a bug or at least incoherence of components, would you please open an issue for it later?

dantengsky (approved):
LGTM & thanks for contribution

BohuTANG (approved):

BohuTANG:
cc @soyeric128 for document

tests/suites/0_stateless/12_time_travel/12_0004_time_travel_select_at.sh
cadl: > will this test case fail without sleeping? the timestamp of the table snapshot is supposed to be monotonically increased even if the local clock drifting presents

Generally not, it only fails in the interval of two insertions less than 1 microsecond.

I found `snapshot.timestamp` precision on Linux is nanoseconds(on MacOS is microseconds). But in the sql statement, I can only input/output timestamp with microsecond precision. 

For example, the timestamp of two insertion snapshots is t1(`2021-08-04T13:57:08.427902838`) and t2 (`2021-08-04T13:57:12.427902838`).
I want to find a time-point between t1 and t2, to test time travel only contains snapshot1.

But the result of `select timestamp from fuse_snapshot($first_insertion)` will be `2021-08-04T13:57:08.427902`. If I pass it to `select … from … (AT TIMESTAMP)`, [FuseTable.navigate_to_time_point](https://github.com/datafuselabs/databend/blob/b6ed4596db4f6cd76617996ccc9588f083d69016/query/src/storages/fuse/operations/navigate.rs#L39) will be false. 
So I add 1 microsecond at [33L](https://github.com/datafuselabs/databend/pull/6056/files#diff-560b9c55798cd4ca36b1748bbc30eef560e9885cdabd30da058560b4b344f0baR33), to make sure time-point is larger than t1. But if t2 - t1 < 1 microsecond, time travel will contain two snapshots, and the test fails too. And so I make a sleep here…


The root cause is the time precision of the snapshot is nanosecond, but we can only use TIMESTAMP[0, 3, 6] in sql now.

Except for the test, I don’t think it’s a serious issue. In general, we won’t use the timestamp of a snapshot to make time travel, instead of snapshot id.

dantengsky: Thanks for the informative description!

dantengsky:
> cc @soyeric128 for document

hi @soyeric128, I am working on a PR the fix the timestamp precision issue (not of this PR), so that the front end user no longer needs to manually adjust the timestamp that was grabbed from `fuse_snapshot` to perform the time travel.

I'll cc you in that PR.

soyeric128:
Hi @BohuTANG and @dantengsky 

Pls review this PR: https://github.com/datafuselabs/databend/pull/6154

PR #6085, opened 2022-06-20
WIP: retry io operations of fuse table

### LINKED ISSUE #6045: bug: Transient object storage IO operation fault handling
**Summary**

Although opendal has put significant efforts to handle transient IO operation faults ([with an exponential backoff strategy](https://github.com/datafuselabs/opendal/blob/38ecf12f268d2aaca07a6739aa4c6c8f5d7ad52b/docs/rfcs/0247-retryable-error.md)), in some cases, it is hard for opendal to be sure that if IO operations could be retried without unexpected side effects.

Thus, the storage engine should take care of transient IO faults as well, to prevent queries (which may take a long time to execute) from being aborted due to momentary network failures,  or temporary unavailability of object storage...

please also see issue

- https://github.com/datafuselabs/databend/issues/6035

where transient network problem is also considered to be one of the causes.

dantengsky:

## Summary

WIP, fixes:
- issue https://github.com/datafuselabs/databend/issues/6045

## Changelog

- Bug Fix

## Related Issues

Fixes #6045

### DIFF
--- a/query/src/storages/fuse/io/retry.rs
+++ b/query/src/storages/fuse/io/retry.rs
@@ -1,0 +1,65 @@
+//  Copyright 2022 Datafuse Labs.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


dantengsky:
> Not looks good to me, opendal will convert all retryable errors (500, 502, 503, ...) to [`Interrupted`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.Interrupted).

got it

> I think we can convert errors while hyper send request to Interrupted too so that we can retry at the opendal level.

yeah, at least `read` operations are safe to retry at `opendal` level

Issue #6063, opened 2022-06-20
http handler aware panic in sql execution


youngsofun:
**Summary**

ZeaLoVe:
```
statement ok
drop database if exists db2;

statement ok
create database db2;

statement ok 
use db2;

statement ok
create table db2_t1;

statement ok
select * from db2.db2_t1;
```

Current planner_v2 is not default enabled, The last statement http handler gets ok but mysql handler get an error.

youngsofun:
```
mysql> set enable_planner_v2=1;
Query OK, 0 rows affected (0.04 sec)
Read 0 rows, 0.00 B in 0.002 sec., 0 rows/sec., 0.00 B/sec.

mysql> select * from db2.t1;
Query OK, 0 rows affected (0.04 sec)
Read 0 rows, 0.00 B in 0.004 sec., 0 rows/sec., 0.00 B/sec.

```
fixed? @ZeaLoVe

PR #6090, opened 2022-06-21
cache panic of http handlers.

youngsofun:

## Summary

HTTP handler is async, 
if the interpreter fails the state will be always `Running`,  (by now not shown as a problem when using planner v2)
it the handler itself fails, it will never return a response. 

need upgrade poem from 1.3.16 to 1.3.31 to use new middleware.

-  fix `RustlsConfig.cert/.key is deprecated`

## Changelog

- Improvement

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/6063


PsiACE (approved):

query/tests/it/servers/http/http_query_handlers.rs
```diff
@@ -684,7 +684,7 @@ async fn get_uri(ep: &EndpointType, uri: &str) -> Response {
             .finish(),
     )
     .await
-    .unwrap_or_else(|err| err.as_response())
+    .unwrap_or_else(|err| err.into_response())
```
youngsofun, line 687: @PsiACE sorry, did not thought you review so fast, rebased only to merge fix here

PsiACE: ok, but it looks like our unit tests still need some fixes.

sundy-li (approved):

PR #6117, opened 2022-06-21
chore(deps): bump deps

PsiACE:

## Summary

Bump most dependencies

Reduce the number of dependencies to be built by about 30

build: 795 -> 767
build-release: 986 -> 957
test: 883 -> 852
lint: 1118 -> 1086

## Changelog

## Related Issues

Fixes #issue


Xuanwo (approved):

Xuanwo:
Maybe we still need to update the `Cargo.lock`.

common/metrics/src/recorder.rs
```diff
@@ -87,7 +87,9 @@ fn init_prometheus_recorder() {
     let recorder = PrometheusBuilder::new().build_recorder();
     let mut h = PROMETHEUS_HANDLE.as_ref().write();
     *h = Some(recorder.handle());
-    metrics::clear_recorder();
+    unsafe {
```
BohuTANG, line 90: Why need unsafe here?

PsiACE: https://github.com/metrics-rs/metrics/pull/302

PsiACE: > `metrics::clear_recorder();` was removed from metrics (technically just marked unsafe and hidden from docs) due to being dangerous to use.

There is a potential data race here, and I think we can open an issue and try to fix it.

ygf11:
Wow, nice work!

I am curious， how you find these crates which need dump and also the target versions?

PsiACE:
> how you find these crates which need dump and also the target versions?

I have written such a document https://databend.rs/doc/faq/routine-maintenance 

The steps are simple and only require some manual tweaks.

ygf11:
> have written such a document https://databend.rs/doc/faq/routine-maintenance

Thanks, learn much :)

PR #6145, opened 2022-06-22
feat(planner): support non-equi conditions in hash join

### LINKED ISSUE #6077: Support non-equi conditions for `HashJoin`
So far, we only support using equivalent condition inside `HashJoin`, which cannot work with complicated join conditions as:
- Non-equi join condition of outer join: `select * from l left join r on l.a > r.a`
- Non-equi predicate of semi join: `select * from t where exists(select * from t1 where t.a > t1.a)`

xudong963:

## Summary

Summary about this PR

## Changelog

- New Feature

## Related Issues

Fixes #6077


query/src/common/evaluator/scalar.rs
```diff
@@ -88,8 +93,8 @@ impl ScalarEvaluator {
                     Self::build_eval_tree(&comp.right)?,
                 ];
                 let func = FunctionFactory::instance().get(comp.op.to_func_name(), &[
-                    &comp.left.data_type(),
-                    &comp.right.data_type(),
+                    &wrap_nullable(&comp.left.data_type()),
```
xudong963, line 96: Because the build side column in outer join is nullable, here I directly wrapped nullable for left and right's `data_type` to make https://github.com/datafuselabs/databend/blob/main/common/datavalues/src/columns/series.rs#L41 happy. 

Do you have some better ideas? @leiysky

query/src/common/evaluator/scalar.rs
```diff
@@ -57,9 +58,13 @@ impl ScalarEvaluator {
 
     fn build_eval_tree(scalar: &Scalar) -> Result<EvalNode> {
         match scalar {
-            Scalar::BoundColumnRef(column_ref) => Ok(EvalNode::Variable {
-                index: column_ref.column.index,
-            }),
+            Scalar::BoundColumnRef(column_ref) => {
+                let name = format_field_name(
+                    column_ref.column.column_name.as_str(),
```
leiysky, line 63: This name may be different from the name of `ColumnEntry` in `Metadata`.

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
```diff
@@ -36,19 +36,22 @@ use common_datavalues::DataValue;
 use common_datavalues::NullableColumn;
 use common_exception::Result;
 use common_planners::Expression;
+use nom::Slice;
```
leiysky, line 39: Maybe remove this?

query/src/common/evaluator/scalar.rs
leiysky: I found maybe it's better to keep using `ExpressionEvaluator` here, which will infer data type from `DataSchema`.

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
```diff
@@ -231,6 +236,66 @@ impl ChainingHashTable {
...
+        for scalar in self.other_conditions.iter() {
+            let func_ctx = self.ctx.try_get_function_context()?;
+            let eval = ScalarEvaluator::try_create(scalar)?;
+            // `predicate_vector` contains a column, which is a boolean column.
+            let predicate_vector = eval.eval(&func_ctx, &result_block)?;
+            // Here, we directly use `predicate_vector` to filter the result block.
+            // But pay attention to the fact that **reserved side** may also be filtered.
+            // **reserved side** is the left side in left join, and the right side in right join.
+            result_block = DataBlock::filter_block(&result_block, predicate_vector.vector())?;
+        }
```
leiysky, line 251: Maybe you can do this in `probe_join`?

common/datablocks/src/kernels/data_block_except.rs
```diff
@@ -0,0 +1,41 @@
...
+            first_block_rows.push(first_block.slice(idx, 1));
+        }
+
+        let mut second_block_rows = Vec::with_capacity(second_block.num_rows());
+        for idx in 0..second_block.num_rows() {
+            second_block_rows.push(second_block.slice(idx, 1));
+        }
+        first_block_rows.retain(|row| !second_block_rows.contains(row));
+        if first_block_rows.is_empty() {
+            return Ok(DataBlock::empty());
```
sundy-li, line 37: empty_with_schema

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
xudong963: Your point is that don't need to write this logic again in the `probe` function?

leiysky: https://github.com/datafuselabs/databend/blob/5e73d12f8b03b53f22a5f99fe70361515a57466a/query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs#L341-L352

I think it's fine to filter the `block` gathered from `RowSpace` with non-equi predicates, and then check if there is necessary to append `NULL`s for it.

So you don't need to do the costly `except_block` operation.

leiysky: > For condition such as `t1.a > t2.c`, it need the `merged_block` to be filtered

Yes, but the `merge_block` is invoked right after this.

https://github.com/datafuselabs/databend/blob/5e73d12f8b03b53f22a5f99fe70361515a57466a/query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs#L366

common/datablocks/src/kernels/data_block_except.rs
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+
+use common_exception::ErrorCode;
+use common_exception::Result;
+
+use crate::DataBlock;
+
+// Compute the difference of two DataBlocks.
+impl DataBlock {
+    pub fn except_blocks(first_block: &DataBlock, second_block: &DataBlock) -> Result<DataBlock> {
```
sundy-li, line 22: Can be improved by SortMerge.

xudong963: I'm replacing `Vec` with `HashSet`, then directly use https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference

xudong963: Oh, It suddenly occurred to me that this function is no longer needed and I will delete it 🤣

leiysky (approved):

PR #6191, opened 2022-06-24
feature(planner): Introduce serializable physical plan

### LINKED ISSUE #6081: Serializable physical plan for new planner
In new planner, we will transform the SQL text into a logical plan, optimize the logical plan into physical plan, and then build a `Pipeline` with the physical plan.

To leverage with distributed execution, we will split the physical plan into `QueryFragment`s, which are fragments of a whole physical plan. Each `QueryFragement` is equivalent to a subtree of physical plan, which means it shares the logic of building a `Pipeline` with physical plan.

For now, we are representing `physical plan` with `SExpr`, which is not serializable since it involves states about optimizer. We need to introduce a serializable IR to represent the `physical plan`, and refactor `PipelineBuilder` with this IR.

leiysky:

## Summary

Introduce a serializable physical plan, which is similar to `PlanNode` but without `Expression` inside.

Also support result set with duplicated column names as:
```sql
mysql> select 3 as a, 4 as a;
+------+------+
| a    | a    |
+------+------+
|    3 |    4 |
+------+------+
1 row in set (0.01 sec)
Read 1 rows, 1.00 B in 0.002 sec., 523.95 rows/sec., 523.95 B/sec.
```

## Changelog

- New Feature

## Related Issues

Close #6081


common/datablocks/src/data_block.rs
```diff
@@ -34,10 +34,16 @@ pub struct DataBlock {
...
-        }));
+        debug_assert!(
+            schema.fields().iter().zip(columns.iter()).all(|(f, c)| f
+                .data_type()
+                .data_type_id()
+                .to_physical_type()
+                == c.data_type().data_type_id().to_physical_type()),
+            "Schema: {:?}, columns: {:?}",
+            &schema,
+            &columns
```
Xuanwo, line 45:
```suggestion
            "Schema: {schema:?}, columns: {columns:?}",
```

common/datablocks/src/data_block.rs
```diff
@@ -34,10 +34,16 @@ pub struct DataBlock {
...
-            f.data_type().data_type_id().to_physical_type()
-                == c.data_type().data_type_id().to_physical_type()
-        }));
+        debug_assert!(
+            schema.fields().iter().zip(columns.iter()).all(|(f, c)| f
+                .data_type()
+                .data_type_id()
+                .to_physical_type()
+                == c.data_type().data_type_id().to_physical_type()),
+            "Schema: {:?}, columns: {:?}",
```
xudong963, line 43: `columns` will also print all `DataValue`s? If so when the amount of data is very large, it brings noise.

leiysky: It's only for debug, I think this is fine.

leiysky: Fixed.

xudong963 (approved):

leiysky:
@mergify update

PR #6219, opened 2022-06-24
bugfix(planner): Fix grouping check

### LINKED ISSUE #6209: select projection will destroy order by logical
Here, we need an error, but now it will succeed. I will create a new ISSUE.

```sql
-- SELECT DISTINCT(id) as id2 FROM a.t ORDER BY a.t.id2; --Succed.
'root'@mysqldb 17:57:14 [(none)]> SELECT DISTINCT(id) FROM a.t ORDER BY a.t.id2;
ERROR 1105 (HY000): Code: 1065, displayText = error: 
  --> SQL:1:39
  |
1 | SELECT DISTINCT(id) FROM a.t ORDER BY a.t.id2
  |                                       ^^^^^^^ for SELECT DISTINCT, ORDER BY expressions must appear in select list

```

_Originally posted by @TCeason in https://github.com/datafuselabs/databend/pull/6196#discussion_r905914520_

leiysky:

## Changelog

- Bug Fix

## Related Issues

Close #6209

### DIFF
--- a/query/src/sql/planner/binder/sort.rs
+++ b/query/src/sql/planner/binder/sort.rs
@@ -23,6 +23,7 @@ use common_exception::ErrorCode;
 use common_exception::Result;
 
 use crate::sql::binder::scalar::ScalarBinder;
+use crate::sql::binder::select::SelectList;
 use crate::sql::binder::Binder;
 use crate::sql::binder::ColumnBinding;
 use crate::sql::optimizer::SExpr;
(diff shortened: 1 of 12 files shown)


andylokandy (approved):

query/src/sql/planner/binder/sort.rs
```diff
@@ -148,14 +148,28 @@ impl<'a> Binder {
         let mut order_by_items = Vec::with_capacity(order_by.items.len());
         let mut scalars = vec![];
         for order in order_by.items {
+            if from_context.in_grouping {
```
TCeason, line 151: Maybe we need do some modify to fix this case 

```sql
select distinct(id) as id1 from t order by id1;
```

Because the in_grouping default value if false , only when the sql has aggregate it will be set true.

BohuTANG:
test_sqllogic_standalone_linux
```
2022-06-26T07:57:28.425577Z  INFO databend_query::servers::mysql::mysql_interactive_worker: Normal query: select count(number) from numbers(10) group by number+1 having number = 1;
2022-06-26T07:57:28.431370Z ERROR databend_query::servers::mysql::writers::query_result_writer: OnQuery Error: Code: 1065, displayText = error: 
  --> SQL:1:71
  |
1 | select count(number) from numbers(10) group by number+1 having number = 1;
  |                                                                       ^ column "number" must appear in the GROUP BY clause or be used in an aggregate function

```

TCeason (approved):

query/src/sql/planner/binder/sort.rs
TCeason: I think this should create a new issue? what do you think about it?@[leiysky](https://github.com/leiysky)

leiysky: @TCeason I think so. 

I have no idea about this case yet, since we haven't support `DISTINCT ON` yet, this query should be the same as `select distinct id as id1 from t order by id1`.

TCeason: Np  :) I think that’s not a regular usage. We just need record it now.

PR #6232, opened 2022-06-25
feature(planner): Decorrelate `EXISTS` subquery with non-equi condition

leiysky:

## Changelog

- New Feature

### DIFF
--- a/query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
+++ b/query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
@@ -119,7 +119,7 @@ pub struct ChainingHashTable {
     hash_table: RwLock<HashTable>,
     row_space: RowSpace,
     join_type: JoinType,
-    other_conditions: Vec<EvalNode<ColumnID>>,
+    other_predicate: Option<EvalNode<ColumnID>>,
 }
 
 impl ChainingHashTable {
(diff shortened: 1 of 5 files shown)


query/src/sql/exec/pipeline_builder.rs
```diff
@@ -487,12 +487,35 @@ impl PipelineBuilder {
         mut child_pipeline: NewPipeline,
         pipeline: &mut NewPipeline,
     ) -> Result<()> {
+        let predicate = other_conditions.iter().cloned().fold(
```
andylokandy, line 490: You can use `https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_reduce` instead.

andylokandy (approved):

query/src/sql/exec/pipeline_builder.rs
leiysky: > You can use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_reduce instead.

It's unstable feature, I don't want to enable it...

query/src/pipelines/new/processors/transforms/hash_join/chaining_hash_table.rs
```diff
@@ -469,11 +468,32 @@ impl ChainingHashTable {
                         {
                             let keys_ref = KeysRef::create(key.as_ptr() as usize, key.len());
                             let probe_result_ptr = table.hash_table.find_key(&keys_ref);
-                            if probe_result_ptr.is_none() {
-                                // No matched row for current probe row
-                                continue;
+                            match (probe_result_ptr, &self.other_predicate) {
```
sundy-li, line 471: Should we also modify `result_blocks` ?

leiysky: > Should we also modify `result_blocks` ?

We don't actually need to, this is the Semi/Anti join-specific behavior.

PR #6256, opened 2022-06-27
[WIP] Prune unused columns

### LINKED ISSUE #6017: Implement column pruning optimization
After finishing all the rewrite optimization, we can perform a column pruning to eliminate the unused columns in `LogicalGet`s and `EvalScalar`s.

For example, a query:
```sql
mysql> create table t(a int, b int, c int);

mysql> explain select a from t;
+-----------------------------+
| explain                     |
+-----------------------------+
| Project: [a]                |
|     Scan: default.default.t |
+-----------------------------+
2 rows in set (0.03 sec)
Read 0 rows, 0.00 B in 0.001 sec., 0 rows/sec., 0.00 B/sec.
```

Since we only need the column `a`, while we still need to fetch `b` and `c` from our columnar storage by default. With the optimization, we can reduce I/O cost by only fetching the needed column.

Another case:
```sql
mysql> explain select a from (select a+1 as a, b+1 as b, c+1 as c from t);
+---------------------------------------------+
| explain                                     |
+---------------------------------------------+
| Project: [a,b,c]                            |
|     EvalScalar: [+(c, 1), +(b, 1), +(a, 1)] |
|         Scan: default.default.t             |
+---------------------------------------------+
3 rows in set (0.04 sec)
Read 0 rows, 0.00 B in 0.001 sec., 0 rows/sec., 0.00 B/sec.
```

The projections can be eliminated to reduce the computation cost. This is very useful if we have redundant aggregate functions.

AngleNet:

## Summary

Prune unused columns

## Changelog

- New Feature

## Related Issues

Fixes #6017


xudong963 (commented):
Thank you @AngleNet , need to add the related test.

Nit: better to make the file name more meaningful, such as `prune_columns.rs`

Issue #5716, opened 2022-06-01
Add more metrics for databend-query


BohuTANG:
**Summary**

Such as `requests/sec`.

Already have(metrics port):
http://127.0.0.1:7070/metrics

```
# TYPE opendal_fs_stat_requests counter
opendal_fs_stat_requests 1

# TYPE session_connect_numbers counter
session_connect_numbers{tenant="admin",cluster_name=""} 1

# TYPE mysql_process_request_duration summary
mysql_process_request_duration{quantile="0"} 0.007519051
mysql_process_request_duration{quantile="0.5"} 0.05450297374362284
mysql_process_request_duration{quantile="0.9"} 0.06394704153625855
mysql_process_request_duration{quantile="0.95"} 0.06394704153625855
mysql_process_request_duration{quantile="0.99"} 0.06394704153625855
mysql_process_request_duration{quantile="0.999"} 0.06394704153625855
mysql_process_request_duration{quantile="1"} 0.0642932895900714
mysql_process_request_duration_sum 0.256558869
mysql_process_request_duration_count 5

# TYPE optimizer_optimize_usedtime summary
optimizer_optimize_usedtime{quantile="0"} 0.000685721
optimizer_optimize_usedtime{quantile="0.5"} 0.000685721
optimizer_optimize_usedtime{quantile="0.9"} 0.000685721
optimizer_optimize_usedtime{quantile="0.95"} 0.000685721
optimizer_optimize_usedtime{quantile="0.99"} 0.000685721
optimizer_optimize_usedtime{quantile="0.999"} 0.000685721
optimizer_optimize_usedtime{quantile="1"} 0.000685721
optimizer_optimize_usedtime_sum 0.001479279
optimizer_optimize_usedtime_count 2

# TYPE interpreter_usedtime summary
interpreter_usedtime{quantile="0"} 0.006074184
interpreter_usedtime{quantile="0.5"} 0.006074184
interpreter_usedtime{quantile="0.9"} 0.006074184
interpreter_usedtime{quantile="0.95"} 0.006074184
interpreter_usedtime{quantile="0.99"} 0.006074184
interpreter_usedtime{quantile="0.999"} 0.006074184
interpreter_usedtime{quantile="1"} 0.006074230150323307
interpreter_usedtime_sum 0.012961012000000001
interpreter_usedtime_count 2

# TYPE parser_parse_usedtime summary
parser_parse_usedtime{quantile="0"} 0.000190323
[...]

PR #6281, opened 2022-06-28
chore(metrics): add metrics for `databend-query/clusters`

cuichenli:

## Summary

Add some metrics for `databend-query/clusters`. Partially #5716 

Metrics added:
- HeartBeat failure count
- Error count in ClusterDiscovery
- Discovered node count in the cluster

Note: I am not sure if we need to add the labels for `cluster_id` and `tenant_id`. If so, we may need to add the corresponding fields to `ClusterDiscovery` and `ClusterHeartBeat` struct. 

## Changelog

## Related Issues

Partially of #5716

### DIFF
--- a/query/src/clusters/cluster.rs
+++ b/query/src/clusters/cluster.rs
@@ -53,6 +53,9 @@ pub struct ClusterDiscovery {
     local_id: String,
     heartbeat: Mutex<ClusterHeartbeat>,
     api_provider: Arc<dyn ClusterApi>,
+    cluster_id: String,
+    tenant_id: String,
+    flight_address: String,
 }
 
 impl ClusterDiscovery {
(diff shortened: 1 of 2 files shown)


query/src/clusters/cluster.rs
```diff
@@ -89,13 +93,38 @@ impl ClusterDiscovery {
 
     pub async fn discover(&self) -> Result<Arc<Cluster>> {
         match self.api_provider.get_nodes().await {
-            Err(cause) => Err(cause.add_message_back("(while cluster api get_nodes).")),
+            Err(cause) => {
+                label_counter_with_val_and_labels(
```
ZhiHanZ, line 97: I think it is required to add tenant id, cluster id and fligh-api-address as metrics label for both counter and gauge

zhang2014 (approved):

query/src/clusters/cluster.rs
cuichenli: updated

ZhiHanZ (approved):

query/src/clusters/cluster.rs
ZhiHanZ: thanks :)

query/src/clusters/cluster.rs
```diff
@@ -51,9 +53,14 @@ pub struct ClusterDiscovery {
     local_id: String,
     heartbeat: Mutex<ClusterHeartbeat>,
     api_provider: Arc<dyn ClusterApi>,
+    cluster_id: String,
+    tenant_id: String,
+    flight_address: String,
 }
 
 impl ClusterDiscovery {
+    const METRIC_LABEL_FUNCTION: &'static str = "function";
```
BohuTANG, line 62: Why the name is 'function'?

PR #6302, opened 2022-06-28
fixes(session): fix mysql connection close_wait or fin_wait_2

### LINKED ISSUE #6263: [bug] MySQL connection can't be closed normally
**Summary**

Remains too much closed_wait

How to produce:

1. start databend-query
2. use MySQL client to connect with `databend-query`
3. quit MySQL client
4. repeat 2-3

origin_img_v2_c30cbc66-a8be-4718-bae5-e3514cfee63g

PsiACE:

## Summary

io_shutdown was called accidentally, resulting in tcp handshake failure. Winter presumes that this type of issue is handled in the clickhouse client.

The lack of active shutdown in mysql means that it has to rely on the drop function of the tcp stream.

However, during attach, the tcp stream is referenced twice, so it cannot be dropped correctly. This should also be the reason why simply removing the codes will fix the problem.

Considering that there might be some potential problems with removing the code, Winter has done some hacking.

## Changelog

- Bug Fix

## Related Issues

Fixes #6263


sundy-li:
I did know if there is other effect, cc @BohuTANG  to review.

BohuTANG:
> I did know if there is another bad effect, cc @BohuTANG to review.

This may affect `processlist` and `shutdown`? I guess @zhang2014 knows it better since he was working on this.

query/src/servers/clickhouse/clickhouse_session.rs
```diff
@@ -31,48 +26,18 @@ pub struct ClickHouseConnection;
...
+                ClickHouseServer::run_on_stream(interactive_worker, stream).await
             });
 
             let _ = futures::executor::block_on(join_handle);
         });
 
         Ok(())
     }
-
-    fn attach_session(session: &SessionRef, blocking_stream: &std::net::TcpStream) -> Result<()> {
```
zhang2014, line 52: This is related to the `kill connection statement` in MySQL. maybe we should try to test it. but i don't known to test it.

ClSlaid:
I don't think it's a good idea, force killing of sessions seem to relies on `SessionMgr`. This PR may damage some basic functionality of the software, just to fix a connection leak problem?

query/src/servers/clickhouse/clickhouse_session.rs
zhang2014: As far as I know, we have unit test coverage this and it's using the rust MySQL driver, but I don't know why it pass. It needs to be fixed.

PsiACE: This is where I got confused, I tested the kill statement specifically but everything worked fine. Perhaps I missed something?

BohuTANG:
Would like to know the root cause of the leak, can you add it to the PR description and help us to know more about it? @PsiACE

PsiACE:
> Would like to know the root cause of the leak, can you add it to the PR description and help us to know more about it? @PsiACE

Some notes have been updated to the description.

zhang2014:
Tcp dump:
```
MySQL exit:

13:26:15.307131 IP localhost.51410 > localhost.opsession-prxy: Flags [F.], seq 107, ack 243, win 6375, options [nop,nop,TS val 2014460509 ecr 3745183794], length 0
13:26:15.307201 IP localhost.opsession-prxy > localhost.51410: Flags [.], ack 108, win 6378, options [nop,nop,TS val 3745183796 ecr 2014460509], length 0
13:26:15.308999 IP localhost.opsession-prxy > localhost.51410: Flags [F.], seq 243, ack 108, win 6378, options [nop,nop,TS val 3745183798 ecr 2014460509], length 0
13:26:15.309091 IP localhost.51410 > localhost.opsession-prxy: Flags [.], ack 244, win 6375, options [nop,nop,TS val 2014460511 ecr 3745183798], length 0

ClickHouse exit:
13:43:01.486031 IP localhost.52644 > localhost.cslistener: Flags [R.], seq 3301329816, ack 2860440633, win 6378, length 0
```

zhang2014:
In rust std [document](https://doc.rust-lang.org/std/net/struct.TcpStream.html):
```
The connection will be closed when the value is dropped. The reading and writing portions of the connection can also be shut down individually with the [shutdown](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.shutdown) method.
```

We cloned TCP stream. So it cannot be destroyed correctly.

sundy-li:
I think #6341 is a better solution. So shall we close this pr?

PR #6354, opened 2022-06-30
query(fix): fix hashmap memory leak

sundy-li:

## Summary

fix hashmap memory leak

## Changelog

- Bug Fix
 

## Related Issues

Fixes #issue

Thanks for @usamoi , can you help review this?

### DIFF
--- a/query/tests/it/common/hashtable.rs
+++ b/query/tests/it/common/hashtable.rs
@@ -12,9 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+
+use databend_query::common::HashMap;
 use databend_query::common::HashMapKind;
 use databend_query::common::HashTableGrower;
 use databend_query::common::SingleLevelGrower;
+use rand::Rng;
 
 #[test]
 fn test_hash_table_grower() {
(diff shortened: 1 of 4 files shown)


query/tests/it/common/hashtable.rs
```diff
@@ -63,3 +68,46 @@ fn test_two_level_hash_table() {
...
+    }
+
+    let item = Arc::new(AtomicUsize::new(0));
+
+    {
+        let mut table = HashMap::<u64, Vec<A>>::create();
+        for _ in 0..1200 {
+            // generate random key without external crate
+            let num1 = vec![2, 3];
+            let key = (&num1 as *const Vec<i32>) as u64 % 88;
```
usamoi, line 99: `num1` is a stacked value in loop and the alignment of `Vec<i32>` guarantees `key` is a multiple of 8. It's not random in my environment.

Xuanwo: Nice catch!

BohuTANG:
```
for _ in 0..1[200](https://github.com/datafuselabs/databend/runs/7133382335?check_suite_focus=true#step:3:209) {
             let key = rng.gen::<u64>() % 40;
-            
+
             let mut inserted = false;
             let entity = table.insert_key(&key, &mut inserted);
             if inserted {
```

zhang2014 (approved):

usamoi (approved):

PR #6360, opened 2022-06-30
improve(planner): optimize exists subquery

xudong963:

## Summary

1. optimize exists subquery
2. fix `limit` display

Before:
```sql
mysql> select  avg(a.number) from numbers(1000000) as a  where EXISTS ( select number from numbers(1000000000));
+---------------+
| avg(a.number) |
+---------------+
|      499999.5 |
+---------------+
1 row in set (5.48 sec)
Read 1001000000 rows, 7.46 GiB in 5.442 sec., 183.96 million rows/sec., 1.37 GiB/sec.
```

Now:
```sql
mysql> select  avg(a.number) from numbers(1000000) as a  where EXISTS ( select number from numbers(1000000000));
+---------------+
| avg(a.number) |
+---------------+
|      499999.5 |
+---------------+
1 row in set (3.13 sec)
Read 1100000 rows, 8.39 MiB in 3.099 sec., 355.01 thousand rows/sec., 2.71 MiB/sec.
```

## Changelog

- Improvement

## Related Issues

Fixes #issue

### DIFF
--- a/query/src/sql/optimizer/heuristic/subquery_rewriter.rs
+++ b/query/src/sql/optimizer/heuristic/subquery_rewriter.rs
@@ -35,6 +35,7 @@ use crate::sql::plans::CrossApply;
 use crate::sql::plans::EvalScalar;
 use crate::sql::plans::FunctionCall;
 use crate::sql::plans::JoinType;
+use crate::sql::plans::Limit;
 use crate::sql::plans::LogicalInnerJoin;
 use crate::sql::plans::Max1Row;
 use crate::sql::plans::OrExpr;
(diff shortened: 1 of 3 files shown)


sundy-li:
/LGTM 

But the best result of the optimization is remove the exist subquery because it's always true and there is no correlated subquery.

xudong963:
> /LGTM
> 
> But the best result of the optimization is to remove the exist subquery because it's always true and there is no correlated subquery.

Yes, will consider to some add fast paths for some specific cases later. The PR is an optimization for any exists subquery.

PR #6366, opened 2022-06-30
show query support format

### LINKED ISSUE #6298: Feature:  query statement support format substmt
**Summary**

Like this :

```SQL
show grants FORMAT TabSeparatedWithNamesAndTypes;
show tables FORMAT TabSeparatedWithNamesAndTypes;
...
-- and so on
```

TCeason:

## Summary

Actually we have three ideas:

1. For every should support format cause's query modify the planner but that will modify so many planner code.

2. In clickhouse handler token the SQL and use nom cut the FORMAT keyword and then get the format value. But it has a bad case like this: select `format` from default.`format` format xxx;

3. In this pr, all of the statement is #statment_body and the format is always at the last of the statement body. So we just need to modify the final rule.

## Changelog

- New Feature

## Related Issues

Fixes #6298

### DIFF
--- a/common/ast/tests/it/testdata/statement-error.txt
+++ b/common/ast/tests/it/testdata/statement-error.txt
@@ -44,7 +44,7 @@ error:
   --> SQL:1:15
   |
 1 | drop table if a.b
-  |               ^ expected `EXISTS`, `.`, `ALL`, or `;`
+  |               ^ expected `EXISTS`, `.`, `ALL`, `FORMAT`, or `;`
 
 
 ---------- Input ----------
@@ -54,7 +54,7 @@ error:
   --> SQL:1:21
   |
 1 | truncate table a.b.c.d
-  |                     ^ expected `PURGE` or `;`
+  |                     ^ expected `PURGE`, `FORMAT`, or `;`
 
 
 ---------- Input ----------
@@ -92,6 +92,16 @@ error:
 
 
 ---------- Input ----------
+show tables format
+---------- Output ---------
+error: 
+  --> SQL:1:19
+  |
+1 | show tables format
+  |                   ^ expected <Ident> or <QuotedString>
+
+
+---------- Input ----------
 alter database system x rename to db
 ---------- Output ---------
 error: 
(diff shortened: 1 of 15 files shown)


BohuTANG:
Need a stateless test for it, and we can easy to document it from the test case.

TCeason:
> Need a stateless test for it, to avoid someone breaking it in the future, and we can easy to document it from the test case.

Yes because the ClickHouse http handler will use old parser to parse the sql. If err it’ll directly return err.

So I only add new parser test.

And now in old parser not support format keyword in show or other query like desc.

Maybe we can set the or to draft. when the planer v2 is enable default add some test?

common/ast/tests/it/testdata/statement-error.txt
```diff
@@ -44,7 +44,7 @@ error:
   --> SQL:1:15
   |
 1 | drop table if a.b
-  |               ^ expected `EXISTS`, `.`, `ALL`, or `;`
+  |               ^ expected `EXISTS`, `.`, `ALL`, `FORMAT`, or `;`
```
BohuTANG, line 47: The hint `FORMAT` here is negative, do we need to hint it here? I have no idea how we to use `FORAMT` for drop table.

common/ast/tests/it/testdata/statement-error.txt
```diff
@@ -54,7 +54,7 @@ error:
   --> SQL:1:21
   |
 1 | truncate table a.b.c.d
-  |                     ^ expected `PURGE` or `;`
+  |                     ^ expected `PURGE`, `FORMAT`, or `;`
```
BohuTANG, line 57: ditto

common/ast/tests/it/testdata/statement-error.txt
BohuTANG: `FORMAT` option only for select(or show) query, not for DDL or others statement?

BohuTANG: `FORMAT` option only for select(or show) query, not for DDL or others statement?

TCeason: > About this issue that's the three ideas:
> 1. For every should support format cause's query modify the planner but that will modify so many planner code.
> 2. In clickhouse handler token the SQL and use nom cut the FORMAT keyword and then get the format value. But it has a bad case like this: select format from default.format format xxx;
> 3. In this pr, all of the statement is #statment_body and the format is always at the last of the statement body. So we just need to modify the final rule.

Now we choose the third. About this question, I had a talk with andylok. 

We think because FORMAT can appear here, it makes sense as an error prompt.

BohuTANG (approved):

PR #6370, opened 2022-06-30
feat(ast): add span info for `TableReference`

### LINKED ISSUE #6005: Add span info for `TableReference`
We will store the source span of query in its AST. But for now, we only store the span for `Expr`.

We should support this for `TableReference` as well, so we can provide better error message in table name resolution.

PragmaTwice:

## Summary

Add span info for every enum item in `TableReference` just like in `Expr`.

## Changelog

- Improvement

## Related Issues

Close #6005.

### DIFF
--- a/common/ast/src/ast/query.rs
+++ b/common/ast/src/ast/query.rs
@@ -132,6 +132,7 @@ pub enum TimeTravelPoint<'a> {
 pub enum TableReference<'a> {
     // Table name
     Table {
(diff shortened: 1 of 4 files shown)


PragmaTwice:
@leiysky It's ready for review now~

leiysky (approved):

BohuTANG:
Thanks for the contribution!

Issue #6353, opened 2022-06-30
streaming_load  just lost first  'N'


wubx:
**Summary**

```
$cat 11.csv
NONE
NNNN
N1

curl -H "insert_sql:insert into t1 format CSV" -H "skip_header:0" -F "upload=@./11.csv" -XPUT http://wubx:wubxwubx@192.168.1.100:8000/v1/streaming_load
{"id":"4104b51a-c1d5-42f3-b054-39faa69b64c1","state":"SUCCESS","stats":{"rows":3,"bytes":39},"error":null}

MySQL [(none)]> desc t1;
+-------+---------+------+---------+-------+
| Field | Type    | Null | Default | Extra |
+-------+---------+------+---------+-------+
| c1    | VARCHAR | YES  | NULL    |       |
+-------+---------+------+---------+-------+
1 row in set (0.006 sec)
MySQL [(none)]> select * from  t1;
+------+
| c1   |
+------+
| ONE  |
| NNN  |
| 1    |
+------+
3 rows in set (0.009 sec)

```

BohuTANG:
cc @youngsofun

xudong963:
It's ok in my local
```sql
➜  ~ sudo curl -s -u root: -XPUT "http://localhost:8000/v1/streaming_load" -H 'insert_sql: insert into t format CSV' -H 'skip_header:0' -F "upload=@./test.csv"
Password:
{"id":"d420e070-6fa4-4d6c-a7ea-903de96e8bd9","state":"SUCCESS","stats":{"rows":3,"bytes":42},"error":null}%
➜  ~ cat test.csv
NONE
NNNN
N1

mysql> select * from t;
+------+
| c    |
+------+
| NONE |
| NNNN |
| N1   |
+------+
3 rows in set (0.03 sec)
Read 3 rows, 42.00 B in 0.003 sec., 864.65 rows/sec., 11.82 KiB/sec.
```

BohuTANG:
This is a terrible bug :/

sundy-li:
ignore_insensitive_bytes("null") will eat the matched bytes , so 'n' is eaten.

We should use `CheckpointReader` to save the state.

youngsofun:
already find a way to fix it，I will continue tomorrow morning

PR #6385, opened 2022-07-01
fix(query):  Add NestedCheckpointReader for input format parser

youngsofun:

## Summary

methods in TypeDeserializer may be called recursively, and in each recursive call,  it may need to checkpoint.

add a new NestedCheckpointReader, and use it throughout the parse progress.

## Changelog

- Bug Fix

## Related Issues

Fixes https://github.com/datafuselabs/databend/issues/6353


common/io/src/buffer/nested_checkpoint_reader.rs
```diff
@@ -0,0 +1,131 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::io;
+use std::io::Result;
+
+use super::BufferRead;
+
+// this struct can avoid NestedCheckpointReader<NestedCheckpointReader<R>> in recursive functions
+pub struct NestedCheckpointReader<R: BufferRead> {
```
sundy-li, line 21: Is that`CheckpointReader`  renaming to `NestedCheckpointReader` ?

common/datavalues/src/types/deserializations/nullable.rs
```diff
@@ -84,47 +84,66 @@ impl TypeDeserializer for NullableDeserializer {
...
 
     // TODO: support null text setting
-    fn de_text<R: BufferRead>(&mut self, reader: &mut R, format: &FormatSettings) -> Result<()> {
+    fn de_text<R: BufferRead>(
+        &mut self,
+        reader: &mut NestedCheckpointReader<R>,
+        format: &FormatSettings,
+    ) -> Result<()> {
         if reader.ignore_insensitive_bytes(b"null")? {
+            reader.pop_checkpoint();
```
sundy-li, line 93: No `push` before this code, why pop here?

youngsofun: my fault, seems tests does not cover here

common/io/src/buffer/nested_checkpoint_reader.rs
youngsofun: no,  it can maintain multi checkpoints.  diff implementation

keep using the buffer instead of preadd to the buffer of the inner reader when rollback

common/datavalues/src/types/deserializations/nullable.rs
youngsofun: fixed

sundy-li (approved):

common/io/src/buffer/nested_checkpoint_reader.rs
```diff
@@ -0,0 +1,131 @@
...
+            pos: 0,
+        }
+    }
+
+    pub fn push_checkpoint(&mut self) {
+        self.checkpointing = true;
+        self.checkpoints.push(self.pos)
+    }
+
+    // todo(youngsofun): add CheckPointGuard to make it safer.
```
sundy-li, line 47: Will be great to have a Guard

youngsofun: I will do it in the next pr, with some other minor refactors.  let`s fix the bug first.

zhang2014 (approved):

common/datavalues/src/types/deserializations/nullable.rs
```diff
@@ -71,60 +71,84 @@ impl TypeDeserializer for NullableDeserializer {
 
     fn de_text_json<R: BufferRead>(
         &mut self,
-        reader: &mut R,
+        reader: &mut NestedCheckpointReader<R>,
         format: &FormatSettings,
     ) -> Result<()> {
+        reader.push_checkpoint();
```
youngsofun, line 77: not find a good way to impl a guard for checkpoint


@sundy-li  

1.  most of the time, need to pop_checkpoint early like here, guard does not help.
2.  guard need to hold a mut ref to impl Drop,  so the flowing code need to use `guard.reader.read()`,  it is not convenient.  and can not pass it to inner consumer (like de_text_json here)

```
```

```
```

PR #6389, opened 2022-07-01
[query] Support geo_to_h3 function

ariesdevil:

## Summary

Support `geo_to_h3` function.
FYI: [H3](https://h3geo.org/) is a geospatial indexing system developed by Uber that partitions the world into hexagonal cells. H3 makes geo computing faster.

Fixes #issue


ariesdevil:
@everpcpc The crate `h3ron-h3-sys` includes the C sources for libh3, so compiling it requires a C toolchain and the cmake build tool. We may need to install them.

common/functions/src/scalars/geos/geo.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use crate::scalars::geos::GeoToH3Function;
+use crate::FunctionFactory;
+
+#[derive(Clone)]
+pub struct GeoFunction;
+
+impl GeoFunction {
+    pub fn register(factory: &mut FunctionFactory) {
+        factory.register("geoToH3", GeoToH3Function::desc());
```
BohuTANG, line 23: Register name as `geo_to_h3` is better, it's not ClickHouse style :D

common/functions/src/scalars/geos/geo_to_h3.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn eval(
+        &self,
+        _func_ctx: FunctionContext,
+        columns: &ColumnsWithField,
+        input_rows: usize,
+    ) -> common_exception::Result<ColumnRef> {
+        // check datatype
+        let mut dt = columns[0].field().data_type();
+        if dt.data_type_id() != TypeID::Float64 {
+            return Err(ErrorCode::BadDataValueType(format!(
```
sundy-li, line 105: we  can accept `numeric datatypes` and use macro to dispatch.

see: [ trigonometric.rs](https://github.com/datafuselabs/databend/blob/89d004c0ee816fbf3209ffe72b1bfcda629b86fb/common/functions/src/scalars/maths/trigonometric.rs)

common/functions/src/scalars/geos/geo.rs
BohuTANG: Databend not :D

common/functions/src/scalars/geos/geo_to_h3.rs
```diff
@@ -0,0 +1,160 @@
...
+                "Invalid type {} of argument {} for function '{}'. Must be Float64",
+                dt.data_type_id(),
+                1,
+                self.name(),
+            )));
+        }
+
+        dt = columns[1].field().data_type();
+        if dt.data_type_id() != TypeID::Float64 {
```
b41sh, line 114: The data type has been checked in `try_create ` function, we don't need to repeat check it here.

common/functions/src/scalars/geos/geo_to_h3.rs
```diff
@@ -0,0 +1,141 @@
...
+    display_name: String,
+}
+
+impl GeoToH3Function {
+    pub fn try_create(
+        display_name: &str,
+        args: &[&DataTypeImpl],
+    ) -> common_exception::Result<Box<dyn Function>> {
+        let mut arg = args[0];
+        if arg.data_type_id() != TypeID::Float64 {
```
sundy-li, line 48: we can use `assert_numeric`

common/functions/src/scalars/geos/geo_to_h3.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn eval(
+        &self,
+        _func_ctx: FunctionContext,
+        columns: &ColumnsWithField,
+        input_rows: usize,
+    ) -> common_exception::Result<ColumnRef> {
+        with_match_primitive_type_id!(columns[0].data_type().data_type_id(), |$S| {
+            with_match_primitive_type_id!(columns[1].data_type().data_type_id(), |$T| {
+                with_match_primitive_type_id!(columns[2].data_type().data_type_id(), |$R| {
+                    let data_lon = Series::check_get_scalar::<f64>(columns[0].column())?;
```
sundy-li, line 106:
```suggestion
                    let data_lon = Series::check_get_scalar::<$S>(columns[0].column())?;
```

common/functions/src/scalars/geos/geo_to_h3.rs
ariesdevil: The `geo_to_h3`'s first and second args must be `Float64` and the third arg must be `UInt8`, so I don't think `assert_numeric` is a good choice here.

common/functions/src/scalars/geos/geo_to_h3.rs
ariesdevil: I try this but got:
```
error[E0308]: arguments to this function are incorrect
   --> common/functions/src/scalars/geos/geo_to_h3.rs:121:39
    |
121 |                         let h3_cell = H3Cell::from_coordinate_unchecked(&coord, res);
    |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ------  --- expected `u8`, found `i8`
    |                                                                         |
    |                                                                         expected `f64`, found `i8`
    |
    = note: expected reference `&geo_types::Coordinate`
               found reference `&geo_types::Coordinate<i8>`
note: associated function defined here
   --> /Users/ariesdevil/.cargo/registry/src/github.com-1ecc6299db9ec823/h3ron-0.14.0/src/h3_cell.rs:103:12
    |
103 |     pub fn from_coordinate_unchecked(c: &Coordinate<f64>, h3_resolution: u8) -> Self {
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the macro `__with_ty__` (in Nightly builds, run with -Z macro-backtrace for more info)
help: you can convert an `i8` to a `u8` and panic if the converted value doesn't fit
    |
121 |                         let h3_cell = H3Cell::from_coordinate_unchecked(&coord, res.try_into().unwrap());
    |                                                                                    ++++++++++++++++++++
```

Issue #6418, opened 2022-07-04
flake test in macos cluster mode


zhang2014:
**Summary**
https://github.com/datafuselabs/databend/runs/7172228410?check_suite_focus=true

PR #6434, opened 2022-07-04
feat(query): support all JsonEachRowOutputFormat variants.

youngsofun:

## Summary

- 2 basic dimensions:
   -  STRINGS:  if true, write quoted string for all type of values.
   -  COMPACT: true/false ->use array/dict for each row,
-  COMPACT formats  have extension types withNames and withNameAndFormats.

https://clickhouse.com/docs/en/interfaces/formats


BohuTANG:
This CI failure is due to #6418 , it's in fixing, please ignore it :)

sundy-li (approved):

Issue #6423, opened 2022-07-04
Flaky test for test_stateless_cluster_linux


Xuanwo:
**Summary**

Production CI failed randomly.

Run: https://github.com/datafuselabs/databend/runs/7174112610?check_suite_focus=true

```
The failure tests:
    /workspace/tests/suites/0_stateless/02_function/02_0035_function_strings_pad.sql
    /workspace/tests/suites/0_stateless/02_function/02_0041_function_strings_format.sql
    /workspace/tests/suites/0_stateless/02_function/02_0044_function_uuid.sql
    /workspace/tests/suites/0_stateless/02_function/02_0051_function_semi_structureds_get.sql
    /workspace/tests/suites/0_stateless/03_dml/03_0008_select_aggregator_combinator.sql
    /workspace/tests/suites/0_stateless/03_dml/03_0010_select_limit_offset.sql
    /workspace/tests/suites/0_stateless/03_dml/03_0016_insert_into_values.sql
    /workspace/tests/suites/0_stateless/03_dml/03_0017_insert_overwrite.sql
    /workspace/tests/suites/0_stateless/03_dml/03_0018_insert_into_variant.sql
    /workspace/tests/suites/0_stateless/03_dml/03_0023_insert_into_array.sql
    /workspace/tests/suites/0_stateless/03_dml/03_0023_insert_into_array_v2.sql
```

Can't reproducible in developing CI.

Xuanwo:
cc @zhang2014

PR #6436, opened 2022-07-04
fix(cluster): fix cannot destroy thread in cluster mode

### LINKED ISSUE #6418: flake test in macos cluster mode

zhang2014:

## Summary

fix(cluster): fix cannot destroy thread in cluster mode

Fixes #6418

### DIFF
--- a/query/src/api/rpc/exchange/exchange_manager.rs
+++ b/query/src/api/rpc/exchange/exchange_manager.rs
@@ -151,20 +151,22 @@ impl DataExchangeManager {
         }
     }
 
-    pub fn remove_query(&self, query_id: &str) {
+    pub fn on_finished_query(&self, query_id: &str, may_error: &Option<ErrorCode>) {
         let mut queries_coordinator = self.queries_coordinator.lock();
-        queries_coordinator.remove(query_id);
+        if let Some(mut query_coordinator) = queries_coordinator.remove(query_id) {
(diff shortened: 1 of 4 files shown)


Xuanwo:
The most interesting thing is this bug was only triggered in the release build on Linux. Can you explain a bit the reason to me?

zhang2014:
I'm not sure if this is the reason of https://github.com/datafuselabs/databend/issues/6423. We have a timer that has not been reset, which may wake up faster in the release version.

BohuTANG:
The CI failure is not related to this PR

BohuTANG:
@mergify update

query/src/api/rpc/exchange/exchange_manager.rs
```diff
@@ -330,13 +338,90 @@ impl QueryCoordinator {
...
+                        .send(DataPacket::ErrorCode(cause))
+                        .await
+                        .is_err()
+                    {
+                        common_tracing::tracing::warn!(
+                            "Cannot send error code, request server channel is closed."
+                        );
+                    }
+
+                    drop(request_server_tx);
```
sundy-li, line 381: I think it's useless to call drop here.

sundy-li (approved):

PR #6437, opened 2022-07-04
ci: Enable cargo sparse-registry

Xuanwo:

## Summary

This PR is a reply to [Call for testing: Cargo sparse-registry](https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html)

We will test:

- Whether sparse-registry works correctly
- If sparse-registry faster than git-based registry
- If sparse-registry works well with many git deps (we have about 15 git deps).

Let's go!


Xuanwo:
Before this PR:

image

Issue #6180, opened 2022-06-23
split databend-query to multiple crates


PsiACE:
**Summary**

Now Databend-Query is a rather large crate and we had to waste a lot of time on compiling and linking.

It's time to split it up into a few relatively small crates.

Related  https://github.com/datafuselabs/databend/pull/4399

Xuanwo:
It's very important to do this job in a progressive way. Otherwise, we can't finish it.

Xuanwo:
#4399 is a good example: It starts three months ago and ends with numerous code conflicts...

PsiACE:
> It's very important to do this job in a progressive way. Otherwise, we can't finish it.

Yes, we're having trouble finding a window to do it all at once, and I think it will probably be done in two or three rounds.

sandflee:
code one minute, cargo check ten minutes :(

PR #6456, opened 2022-07-05
refactor: split common-meta-store

PsiACE:

## Summary

After this split:

- `query/common/meta` will be available as a common crate named `common-meta-store`

Related #6180


PsiACE:
It seems that it lacks some tests and could be added if necessary.

Xuanwo:
This crate provide very litter APIs, how about merging within meta?

drmingdrmer (approved):
LGTM

PsiACE:
> This crate provide very litter APIs, how about merging within meta?

But which meta would be better for us to move it to? It looks like it's just a thin layer and the other meta components have clearer boundaries.

drmingdrmer:
> This crate provide very litter APIs, how about merging within meta?

Not very sure about it. there may be dependency issues if moving it to an existent common/meta/* crate.

PR #6467, opened 2022-07-05
refactor: split hashtable

PsiACE:

## Summary

After this split:

- `query/common/hashtable` will be available as a common crate named `common-hashtable`

Related #6180


BohuTANG:
We should document(in crate README) the hashtable is different from other hashtables, and why we used it here :)

zhang2014 (approved):

Issue #6430, opened 2022-07-04
bug: stack overflow in `TypeChecker`


xudong963:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

main

### What's Wrong?

`TypeChecker` will cause stack overflow in the case
```sql
set enable_planner_v2 = 1;
select
    o_year,
    sum(case
            when nation = 'BRAZIL' then volume
            else 0
        end) / sum(volume) as mkt_share
from
    (
        select
            extract(year from o_orderdate) as o_year,
            l_extendedprice * (1 - l_discount) as volume,
            n2.n_name as nation
        from
            part,
            supplier,
            lineitem,
            orders,
            customer,
            nation n1,
            nation n2,
            region
        where
                p_partkey = l_partkey
          and s_suppkey = l_suppkey
          and l_orderkey = o_orderkey
          and o_custkey = c_custkey
          and c_nationkey = n1.n_nationkey
          and n1.n_regionkey = r_regionkey
          and r_name = 'AMERICA'
          and s_nationkey = n2.n_nationkey
          and o_orderdate between to_date('1995-01-01') and to_date('1996-12-31')
          and p_type = 'ECONOMY ANODIZED STEEL'
        and 1 and 1 and 1 and 1 and 1
    ) as all_nations
group by
    o_year
order by
    o_year;
```
The bug was found in https://github.com/datafuselabs/databend/runs/7174312042?check_suite_focus=true tpch-q8.

Fortunately, tpch q8 does not stack overflow in main, but it does in the ticket https://github.com/datafuselabs/databend/pull/6412, which is very strange...(This ticket just adds a match branch to the `resolve` function)

So I added some more `and` in `where` to tpch q8, and as expected, the main branch also stack overflowed.

### How to Reproduce?

Run the above case or construct a SQL with many `and` in `where`

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

xudong963:
cc @leiysky

sundy-li:
So we need to rewrite the TypeChecker using visitor instead of recursive call ?

xudong963:
> So we need to rewrite the TypeChecker using visitor instead of recursive call ?

There doesn't seem to be a better way to process the issue. Can you locally reproduce it?

sundy-li:
> > So we need to rewrite the TypeChecker using visitor instead of recursive call ?
> 
[...]

PR #6480, opened 2022-07-05
chore(planner): Refine stack usage of type check

leiysky:

## Summary

Reduce stack memory usage to support deeper expression.

Before this PR, it can only support binary expression with depth 15 on my macbook pro.

After this PR, the depth is increased to 31.
image

Related to #6430


xudong963 (approved):

Xuanwo:
```shell
  --- stderr
  error occurred: Failed to find tool. Is `musl-gcc` installed?
error: failed to run custom build command for `lz4-sys v1.9.3`
Caused by:
  process didn't exit successfully: `/workspace/target/debug/build/lz4-sys-af3f644e019d95d9/build-script-build` (exit status: 1)
```

@everpcpc PTAL

leiysky:
@mergify update

BohuTANG:
The CI failures are not related to this PR, please ignore.

BohuTANG:
@mergify update

sundy-li (approved):

sundy-li:
Maybe we can use RPN to replace the recursion.

leiysky:
> Maybe we can use RPN to replace the recursion.

I was considering about that, but we have to introduce a new IR. This is not necessary for now, we can leverage with it in the future.

PR #6497, opened 2022-07-06
fix(jaeger) split package to send big package to jaeger

sandflee:

## Summary
when using jaeger to track databend, I always see "jaeger: thrift agent failed with message too long" and  set OTEL_BSP_MAX_EXPORT_BATCH_SIZE to a larger value does't take much affect.  we could use "with_auto_split_batch" to split big packages.

### DIFF
--- a/common/tracing/src/logging.rs
+++ b/common/tracing/src/logging.rs
@@ -102,6 +102,7 @@ pub fn init_global_tracing(
         let tracer = opentelemetry_jaeger::new_pipeline()
             .with_service_name(app_name)
             .with_agent_endpoint(jaeger_agent_endpoint)
+            .with_auto_split_batch(true)
             .install_batch(opentelemetry::runtime::Tokio)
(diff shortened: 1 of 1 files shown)


sandflee:
> When auto split is set to true, the exporter will try to split the batch into smaller ones so that there will be minimal data loss. It will impact the performance.
enable auto split will  impact the performance, but sending to jaeger is not enabled by default and not impact much, I think this is acceptable.

PsiACE (approved):
lgtm, but a little deeper investigation is needed

> Note that if one span is too large to export, other spans within the same batch may or may not be exported. In this case, exporter will return errors as we cannot split spans.

BohuTANG (approved):

PsiACE:
>  u want to know why "with_auto_split_batch" worked ?

I am more concerned about whether we have some specific use cases that will lead to strange errors.

PR #6508, opened 2022-07-06
feat(storage): support auto compaction

### LINKED ISSUE #6492: auto compaction
Support auto compaction.
./databend-query --cmd compaction

zhyass:

## Summary

./databend-query --cmd compaction

Fixes #6492


query/bin/databend-query.rs
```diff
@@ -220,23 +241,60 @@ async fn main(_global_tracker: Arc<RuntimeTracker>) -> common_exception::Result<
...
-            eprintln!("    Print version and the min compatible databend-meta version");
+    let executor_session = sessions.create_session(SessionType::Dummy).await?;
+    let qc = executor_session.create_query_context().await?;
+    let tenant = qc.get_tenant();
+    let cl = qc.get_catalog(CATALOG_DEFAULT)?;
+
+    let ctx = qc.clone();
+    let catalog = cl.clone();
+    let sched = JobScheduler::new().expect("Failed to new JobScheduler");
+    let cron_job = Job::new_async("@daily", move |_, _| {
```
sundy-li, line 256: `@daily` is tooooo infrequently

query/bin/databend-query.rs
```diff
@@ -34,14 +38,31 @@ use databend_query::servers::MySQLHandler;
...
+        match conf.cmd.as_str() {
+            "ver" => {
+                println!("version: {}", QUERY_SEMVER.deref());
+                println!("min-compatible-metasrv-version: {}", MIN_METASRV_SEMVER);
+            }
+            "compaction" => run_compaction_cmd(&conf).await?,
+            _ => {
+                eprintln!("Invalid cmd: {}", conf.cmd);
+                eprintln!("Available cmds:");
+                eprintln!("  --cmd ver");
```
xudong963, line 62: We also need to add `eprintln!("    --cmd compaction")`

query/src/cmd/compaction.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    let plan = OptimizeTablePlan {
+        catalog: ctx.get_current_catalog(),
+        database,
+        table: table.name().to_string(),
+        action: OptimizeTableAction::Compact,
+    };
+
+    table.compact(ctx.clone(), plan).await
```
sundy-li, line 133: This will compact all blocks. I think it's better to give a heuristic algorithm to filter some more valuable blocks.

Xuanwo:
`./databend-query --cmd compaction` looks weird and not extendable.

## Backwards

- cmd is hard to extend: we have to handle compaction-related args
- cmd (especially for query) is complex: we have to handle `./databend-query --cmd compaction --storage-type=fs --storage-s3-access-key-id=xxxx .....`
- cmd doesn't work well with other parts of the query: we can't explain it, we can't debug it, we can't show progress about it.

## Ideas

IMHO, how about building a background task scheduler so that:

- Every valid SQL could be treated as a task. We can start/stop/resume/... them
- We can scheduler timely task in background like optimize, compaction
- We can control long-running tasks (like `COPY`, `OPTIMIZE`)

query/src/cmd/compaction.rs
```diff
@@ -0,0 +1,255 @@
...
+    Ok(())
+}
+
+pub async fn do_compaction(
+    ctx: Arc<QueryContext>,
+    table: Arc<dyn Table>,
+    database: String,
+) -> common_exception::Result<()> {
+    if !table.auto_compaction() {
+        return Ok(());
```
xudong963, line 123: It's better to print some information to user that the table can't auto compaction?

query/src/cmd/compaction.rs
```diff
@@ -0,0 +1,255 @@
...
+                    new_item = queue_add_rx.recv(), if !queue_add_eof => {
+                        match new_item {
+                            Some(item) => {
+                                // insert it into the DelayQueue.
+                                task.insert(item.0, item.1);
+                            },
+                            None => {
+                                // the queue_add channel was dropped.
+                                queue_add_eof = true;
+                            },
```
dantengsky, line 209: I must have missed something,  it seems that `queue_read_tx.recv` happens just after the `queue_read_tx.send`. 

can this be collapsed into :

~~~
tokio::select! {
                      .....
                    // delayed item from the DelayQueue becomes available.
                    delayed_item = task.next() => {
                        match delayed_item {
                            Some(item) => {
                                 task.insert(item.0, item.1);
                            },
                              ...
                        }
                     }
 }                 
~~~~

query/src/cmd/compaction.rs
```diff
@@ -0,0 +1,255 @@
...
+        key,
+        key_end: Some(key_end),
+        filter_type: FilterType::Update.into(),
+    };
+    // Create a new client of metasrv.
+    let client = MetaGrpcClient::try_new(&rpc_conf)?;
+
+    // Watch the table meta update.
+    let mut client_stream = client.request(watch).await?;
+    let task = CompactionQueue::new(Duration::from_secs(3600));
```
xudong963, line 72: Maybe we can make `timeout` as a user-specified argument, not a hardcode number?

Xuanwo: As discussed in https://github.com/datafuselabs/databend/pull/6508#issuecomment-1189712979

I don't think place auto-compaction in command line args is a good UX.

sundy-li:
> `./databend-query --cmd compaction` looks weird and not extendable.

I think we just need a flag config, eg `./databend-query --enable_compaction 1`, then it'll start the compaction service as background crons. 

For commonly singleton deployed cases, we just need to start a single `databend-query`(with embedded compaction) & `databend-meta`

query/src/cmd/compaction.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    fn poll_expired(&mut self, cx: &mut Context) -> Poll<Option<TableInfo>> {
+        match ready!(self.expirations.poll_expired(cx)) {
+            Some(key) => {
+                let table_info = self.entries.remove(key.get_ref()).unwrap().0;
+                Poll::Ready(Some(table_info))
+            }
+            None => Poll::Pending,
```
dantengsky, line 168: I might be wrong, but will cpu be trapped in a busy loop of polling here if `DealyQueue` is empty?

Since `self.expirations.poll_expired(cx)` keeps returning `None`, if the DelayQueue is empty, and here we just return `Pending` to the puller.

query/src/cmd/compaction.rs
zhyass: > it's

Only for non cluster key tables. The implementation will be slow at the first time.

query/src/cmd/compaction.rs
zhyass: > Maybe we can make `timeout` as a user-specified argument, not a hardcode number?

We will consider making `timeout` and `cron time` as user-specified arguments later.

zhyass:
> > `./databend-query --cmd compaction` looks weird and not extendable.
> 
> I think we just need a flag config, eg `./databend-query --enable_compaction 1`, then it'll start the compaction service as background crons.
> 
> For commonly singleton deployed cases, we just need to start a single `databend-query`(with embedded compaction) & `databend-meta`
> 
> It can also suit the cases of cloud services.

I'll try to add a flag config.

dantengsky:
> ## Ideas
> IMHO, how about building a background task scheduler so that:
> 
> * Every valid SQL could be treated as a task. We can start/stop/resume/... them
> * We can schedule the timely task in the background like optimization, compaction
> * We can control long-running tasks (like `COPY`, `OPTIMIZE`)
>  ...


@Xuanwo something like this? https://docs.snowflake.com/en/user-guide/tasks-intro.html, that will be great! :)

personally, I also prefer to put query nodes into more passive roles. (other roles like cloud-platform, keep watching events, submit tasks to query nodes, using SQL if possible).

but, this PR, IMHO, is "proof of the concept" that the trigger of compaction task(part of it) could be implemented by watching meta kv keys.

maybe we could treat the functionalities, and the UX parts of this PR,  as preview features, for internal evaluation only, and polish the impls and UX gradually.

Xuanwo:
> maybe we could treat the functionalities, and the UX parts of this PR, as preview features, for internal evaluation only, and polish the impls and UX gradually.

Yep. I respect the existing design and implementation. We can improve them gradually. Make this feature a preview and internal evaluation only LGTM.

query/bin/databend-query.rs
BohuTANG: This should be a configuration

BohuTANG:
I think the flag way is more make sense, as @sundy-li said above. Reasons:
For databend cloud, we can not start a special databend-query with cmd for a tenant, but we can start the databend-query with the flag to enable the compaction.
Also cc @ZhiHanZ @flaneur2020

flaneur2020:
cc @ZhiHanZ @hantmac @junnplus

flaneur2020:
> For databend cloud, we can not(hard?) start a special databend-query with cmd for a tenant, but we can start the databend-query with the flag to enable the compaction in the background.

I guess it's ok to start a special databend-query with cmd for a tenant to run the compaction, however, instead of a daemon process keeps running, it'd be better to let kubernetes schedule a CronJob to run a compaction job, and exit as it finished it's job.


@ZhiHanZ @hantmac @junnplus how do you think about this workload?

sundy-li (approved):

Xuanwo:
I converted this PR into draft to prevented been merged.

Xuanwo:
Some ideas after discussion with @flaneur2020 

## Short Term

- We should remove the cron in query, and let users to do this job (K8s cronjob, Systemd timer)
- Remove the flag or cmd from the query, allow query reading from studio

So we can start a compaction job via:

`echo "<compaction_sql>" | ./databend-query`

## Long Term

- Add a task framework so that we can start/stop/show/list tasks.

zhyass:
After discussion, we decided that the scheduling of compaction should be implemented by users, and databend-query should not do this. So this PR is closed.

xudong963:
> After discussion, we decided that the scheduling of compaction should be implemented by users, and databend-query should not do this. So this PR is closed.

No code is the best code!

Xuanwo:
@mergify unqueue

PR #6512, opened 2022-07-06
remove keyword table from func is_reserved_ident

### LINKED ISSUE #6455: bug: planner v2 use keyword without `` return error
### Search before asking

### Version

```sql
'root'@mysqldb 10:59:35 [system]> set enable_planner_v2=1;

'root'@mysqldb 11:00:35 [system]> SELECT database, table, name, type, default_kind as default_type, default_expression, comment FROM system.columns  WHERE database LIKE 'default'  AND table LIKE 't2';
  --> SQL:1:18
  |
1 | SELECT database, table, name, type, default_kind as default_type, default_expression, comment FROM system.columns  WHERE database LIKE 'default'  AND table LIKE 't2'
  | ------           ^^^^^ expected `*`, `IS`, `IN`, `EXISTS`, `BETWEEN`, `+`, or 51 more ...
  | |                 
  | while parsing `SELECT ...`

.
'root'@mysqldb 11:00:37 [system]> SELECT database, `table`, name, type, default_kind as default_type, default_expression, comment FROM system.columns  WHERE database LIKE 'default'  AND `table` LIKE 't2';
+----------+-------+------+------+--------------+--------------------+---------+
| database | table | name | type | default_type | default_expression | comment |
+----------+-------+------+------+--------------+--------------------+---------+
| default  | t2    | id   | INT  |              |                    |         |
| default  | t2    | id1  | INT  |              |                    |         |
+----------+-------+------+------+--------------+--------------------+---------+
2 rows in set (0.08 sec)
Read 130 rows, 11.13 KiB in 0.018 sec., 7.13 thousand rows/sec., 610.02 KiB/sec.

```

### What's Wrong?

In some BI tools use this error sql to fetch table columns.

If we return an error, the tool will not fetch any columns.

### How to Reproduce?

_No response_

TCeason:

## Summary

Now we use the old parser to parse SQL first and then try to use planner v2.

So maybe some SQL will fail in planner v2 but success in planner v1.

In issue #6455 we find this bug. So in this pr, we fix it.

Fixes #6455

### DIFF
--- a/common/ast/src/parser/token.rs
+++ b/common/ast/src/parser/token.rs
@@ -947,7 +947,7 @@ impl TokenKind {
             // | TokenKind::SIMILAR
             | TokenKind::SOME
             // | TokenKind::SYMMETRIC
-            | TokenKind::TABLE
+            // | TokenKind::TABLE
             // | TokenKind::TABLESAMPLE
             | TokenKind::THEN
             | TokenKind::TRAILING
--- a/tests/suites/0_stateless/20+_others/20_0001_planner_v2.result
+++ b/tests/suites/0_stateless/20+_others/20_0001_planner_v2.result
@@ -426,3 +426,10 @@ NULL	NULL	NULL	1
 1
 1	2
 2	3
+====WITH_KEYWORD====
+system	settings	default	VARCHAR			
+system	settings	description	VARCHAR			
+system	settings	level	VARCHAR			
+system	settings	name	VARCHAR			
+system	settings	type	VARCHAR			
+system	settings	value	VARCHAR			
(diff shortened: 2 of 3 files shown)


xudong963 (commented):
Because in the sql standard, table is supposed to be a reserved indent.

So I'm not sure if it's appropriate to use table directly as a non-reserved indent (there could be other potential problems? But I don't think of any at the moment).

Maybe those BI tools should follow the sql standard and use `table` with `` instead of bare `table`

xudong963 (approved):

BohuTANG:
@andylokandy please take a review

BohuTANG:
table can be a reversed keyword, but we should make it works when it's in projections?

TCeason:
> 

I think not only projection but also as a expr.

TCeason:
> table can be a reversed keyword, but we should make it works when it's in projections?

Like this : table LIKE 'settings' in test sql

PR #6515, opened 2022-07-06
fix(query): fix date/timestamp deserializing error

b41sh:

## Summary

When deserializing `date` and `timestamp`, if the value is invalid, an error will be returned directly, resulting in the following quotation not being consumed from the buffer. This will cause `skip_to_next_row` can't process the remaining data correctly.

```sql
mysql> create table test(
    -> `id` int,
    -> `create_time` timestamp,
    -> `info` string
    -> );

mysql> insert into test values (1, '0000-00-00 00:00:00', 'test'),(2, '2020-01-01 00:00:00', 'test)');
ERROR 1105 (HY000): Code: 1005, displayText = sql parser error: Unterminated string literal at Line: 1, Column 57.
```

Fixes #issue


BohuTANG:
@mergify update

sundy-li (approved):

PR #6524, opened 2022-07-07
fix(query): fix input format CSV

youngsofun:

## Summary

both python and rust CSV lib support 4 QuoteStyple 
https://docs.rs/csv/latest/csv/enum.QuoteStyle.html

except the last one the 4,  CSV mainly uses quote (instead of escape) to deal with field/record delimiters in strings, and uses quote to escape it self by default.

and clickhouse input format only supports this style.

so we should support this style first.

Summary about this PR

Fixes #issue


zhang2014 (approved):

youngsofun:
flaky tests, pass on my mac

sundy-li:
results lost. cc @zhang2014

BohuTANG:
test_sqllogic_standalone_linux failure is not related to this pr, let's merge.

Issue #6517, opened 2022-07-06
Adding a test will make this PR perfect:DD


drmingdrmer:
Adding a test will make this PR perfect:DD

_Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/pull/6516#pullrequestreview-1030310947_

PR #6526, opened 2022-07-07
fix: add watch txn unit test

lichuang:

## Summary

1. fix: add watch txn unit test.
2. add function `get_start_and_end_of_prefix`.

Fixes #issue https://github.com/datafuselabs/databend/issues/6517


BohuTANG:
error: `to_string` applied to a type that implements `Display` in `format!` args
[1482](https://github.com/datafuselabs/databend/runs/7226658497?check_suite_focus=true#step:3:1492)
--> common/meta/api/src/kv_api.rs:41:31
[1483](https://github.com/datafuselabs/databend/runs/7226658497?check_suite_focus=true#step:3:1493)
|
[1484](https://github.com/datafuselabs/databend/runs/7226658497?check_suite_focus=true#step:3:1494)
41|    format!("{}{}", s[..l - 2].to_string(), (a as u8 + 1) as char)
[1485](https://github.com/datafuselabs/databend/runs/7226658497?check_suite_focus=true#step:3:1495)
| ^^^^^^^^^^^^help: remove this
[1486](https://github.com/datafuselabs/databend/runs/7226658497?check_suite_focus=true#step:3:1496)
|
[1487](https://github.com/datafuselabs/databend/runs/7226658497?check_suite_focus=true#step:3:1497)
= note: `-D clippy::to-st

common/meta/api/src/kv_api.rs
```diff
@@ -36,9 +36,21 @@ pub trait ApiBuilder<T>: Clone {
...
+    while l > 0 {
+        l -= 1;
+        let a = bytes[l];
+        if a == 255 {
+            continue;
+        }
+        unsafe {
+            ret.as_mut_str().as_bytes_mut()[l] = a + 1;
+        }
+        return ret;
```
drmingdrmer, line 51: When `l` goes back, the length of `ret` has to shrink.

There should be a unit test for this critical function.

common/meta/api/src/kv_api.rs
```diff
@@ -36,9 +36,21 @@ pub trait ApiBuilder<T>: Clone {
...
+    }
+    format!("{}{}", 255 as char, s)
```
drmingdrmer: If the input is a string of single char of ascii 255, the return value should be something else to indicate an open right bound.

ariesdevil (approved):
LGTM

common/base/src/base/string_func.rs
```diff
@@ -103,3 +103,34 @@ pub fn mask_string(s: &str, unmask_len: usize) -> String {
...
+        l -= 1;
+        if let Some(c) = s.chars().nth(l) {
+            if c == 127 as char {
+                continue;
+            }
+            return replace_nth_char(s, l, (c as u8 + 1) as char);
+        }
+    }
+    format!("{}{}", s, 127 as char)
+}
```
drmingdrmer, line 136: Why use `127` as the greatest char? Is it a concern about `utf` string?

lichuang: > Why use `127` as the greatest char? Is it a concern about `utf` string?

I think support ASCII character is enough...

drmingdrmer: > > Why use `127` as the greatest char? Is it a concern about `utf` string?
> 
> I think support ASCII character is enough...

Hmm... I agree to introduce a limit of only allowing to use ascii string.
With this assumption, it should return some error if the input string is not ascii string.

BTW, `String::chars()` returns a iterator of unicode characters:

```rust
    let a = "我".to_string();
    let x = a.chars();
    for c in x {
        println!("{:?}, {:?}", c, char::MAX);
        // '我', '\u{10ffff}'
    }

```

common/base/src/base/string_func.rs
```diff
@@ -115,22 +117,30 @@ pub fn replace_nth_char(s: &str, idx: usize, newchar: char) -> String {
...
 
-/// Return prefix of string.
+/// Return prefix of string(only support ASCII char).
 /// "a" -> "b"
 /// "1" -> "2"
 /// [96,97,127] -> [96,98,127]
 /// [127] -> [127, 127]
 /// [127,127,127, 127] -> [127,127,127, 127, 127]
-pub fn prefix_of_string(s: &str) -> String {
+pub fn prefix_of_string(s: &str) -> common_exception::Result<String> {
```
Xuanwo, line 126: The naming is very confusing to me. Can we come up with a better name here?

I can't recognize the input and output with `prefix_of_string`.

lichuang: ok, seems that `prefix_of_string` is not suitable in `string_func.rs`, i move it to `kv_api`.

common/meta/api/src/kv_api.rs
```diff
@@ -35,6 +37,39 @@ pub trait ApiBuilder<T>: Clone {
     async fn build_cluster(&self) -> Vec<T>;
 }
 
+/// Return a string that bigger than all the string prefix with input string(only support ASCII char).
+/// "a" -> "b"
+/// "1" -> "2"
+/// [96,97,127] -> [96,98,127]
+/// [127] -> [127, 127]
+/// [127,127,127, 127] -> [127,127,127, 127, 127]
+pub fn prefix_of_string(s: &str) -> common_exception::Result<String> {
```
drmingdrmer, line 46: What's going on with this...

image

PR #6534, opened 2022-07-07
feat(planner): support `ALL` and `SOME` subquery, mark join with non-equi condition, and make tpchq20 happy

xudong963:

## Summary

The PR contains four things: 

**1. Support ALL and SOME subquery**
1. `ALL` subquery will be converted to `ANY` subquery, `= All()` => `not( != Any())`; `!= All()` => `not(= Any())`; `> All()` => `not(<= Any())`, ...
2. `SOME` subquery is equal to `ANY` subquery

**2. Mark join supports non-equi join**: so we support `ANY`, `ALL`, `SOME` subquery with other comparison operators besides `=`.

**3. Fix mark join bug**

The HashTable and Mark Column should be merged to produce the final result after the probe is finished, instead of being processed in the probe.  

If processed in the probe phase, when the data volume is large, the probe input will be split into multiple parts, and each probe will generate the result once, leading to huge intermediate results and causing memory waste.  

To make matters worse, it can also lead to wrong results, consider the following case:

```sql
mysql> set enable_planner_v2 = 0;

mysql> create table t3 as select *  from numbers(10000);

mysql> insert into t3 values(1);

mysql> select count(*) from numbers(10000) as t1 where t1.number in (select t3.number from t3);
+----------+
| COUNT(*) |
+----------+
|    10001 |
+----------+
1 row in set (0.05 sec)
Read 20001 rows, 156.26 KiB in 0.023 sec., 872.2 thousand rows/sec., 6.65 MiB/sec.
```
The output was 10001, but t1 is only 10000 rows, the correct result should be 10000.

**4. make tpch q20 happy**
https://github.com/datafuselabs/databend/pull/6534/commits/27aed3d172dc71a4f635f5004a16980cc8c3863f put `mark_index` into join output columns, then it'll process nested InSubquery, such as tpch q20. 

Fixes #issue


query/src/pipelines/new/processors/transforms/hash_join/result_blocks.rs
```diff
@@ -130,65 +124,37 @@ impl JoinHashTable {
...
-                            marker[ptr.row_index as usize] = MarkerKind::True;
+                            let offset;
+                            {
+                                let chunks = self.row_space.chunks.read().unwrap();
+                                offset = chunks
+                                    .iter()
+                                    .by_ref()
+                                    .take(ptr.chunk_index as usize)
+                                    .fold(0, |acc, c| acc + c.num_rows());
+                            }
```
leiysky, line 152:
```suggestion
                            let offset = {
                                let chunks = self.row_space.chunks.read().unwrap();
                                chunks
                                    .iter()
                                    .by_ref()
                                    .take(ptr.chunk_index as usize)
                                    .fold(0, |acc, c| acc + c.num_rows());
                            }
```

xudong963:
https://github.com/datafuselabs/databend/commit/27aed3d172dc71a4f635f5004a16980cc8c3863f put `mark_index` into join output columns, then it'll process nested InSubquery, such as tpch q20.

Now we support all tpch SQL!!!

PR #6548, opened 2022-07-08
fix(query): fix load json value by csv format

b41sh:

## Summary

Implemented a custom `JSON` deserialization to process `JSON` values in `CSV` files, fix the bug that `JSON` values could not be loaded from local `CSV` file

Fixes #issue


common/datavalues/src/types/deserializations/variant.rs
```diff
@@ -113,10 +113,8 @@ impl TypeDeserializer for VariantDeserializer {
         _format: &FormatSettings,
     ) -> Result<()> {
         self.buffer.clear();
-        reader.read_quoted_text(&mut self.buffer, b'"')?;
-
-        let val = serde_json::from_slice(self.buffer.as_slice())?;
-        self.builder.append_value(val);
+        let val = reader.read_json_value()?;
```
youngsofun, line 116: csv has it own escape rule,  I suggest read this field as a csv string (unescaped) then decode with json

b41sh: In function `read_csv_string`, we use a comma as delimiter to split the string, but the comma may be part of the `JSON` value, for example `[1,2,3]`. I'm afraid the `JSON` value can't be parsed correctly from the `CSV` string.

youngsofun: Any text-based format has its own way of dealing with it using escape and/or escape.
 
the outer layer  CSV should only know the field is a string/number/date.., nothing else, no use to invent our own CSV that can not be parsed by other CSV readers.

 by default , if there is `,` in the string, the CSV need to enclose it with a quote

https://docs.python.org/3/library/csv.html
https://clickhouse.com/docs/en/interfaces/formats#csv

youngsofun: need to make sure a general CSV reader can not rely on knowledge like` the second field is a JSON`  to parse a CSV.
without this knowledge,   a CSV reader at least knows the ` the second field is a string` and can parse it accordingly, and leave to upper level to decode the string further ( e.g. decode to nested structs).

b41sh: I understand what you mean, reading as a `CSV` string format works for me but I'm not sure the `,` is enclosed with a quote in the user's `CSV` file, we need some sample test files to check it.

youngsofun: if `,` is neither enclosed or escaped, it`s invalid, we don`t need to consider it .

on the other side, normally the string must be quoted, and the the `"` is escaped as `""`(or "\"" for some formatter) 
and will leave to error whey you try to decode it as a normal JSON.

the CSV  format has to follow some rules after all.

sundy-li:
I'm not sure if it's right, I need time to validate the behavior in other DBMSs.

youngsofun (approved):

b41sh:
> I'm not sure if it's right, I need time to validate the behavior in other DBMSs.
> 
> Example in clickhouse:
> 
> ```
> -- not ok
>  echo '37,[1,2,1,4]' | clickhouse-local --structure "a UInt32, b Array(UInt32)" --input-format "CSV"  --query "select * from table"
> 
> -- ok 
>  echo '37|[1,2,1,4]' | clickhouse-local --structure "a UInt32, b Array(UInt32)" --input-format "CSV" --format_csv_delimiter "|"  --query "select * from table"
> ```

Since both `CSV` and `JSON` use `,` as element separators, it is not possible to process `JSON` directly.
The behavior of this PR is the same as that of `ClickHouse`. The correct parsing of array and object needs to meet one of the following three conditions:
1. Wrap it in single quotes, for example: `1,'[1,2,3]','{"k1":"v1","k2":"v2"}'`
2. Only one element inside, for example: `1,[1],{"k":"v"}`
3. Use different field delimiters, for example: `1|[1,2,3]|{"k1":"v1","k2":"v2"}`

sundy-li (approved):

PR #6550, opened 2022-07-09
ci: Cleanup Makefile and setup scripts

Xuanwo:

## Summary

This PR will remove not used targets in Makefile and not used functions in setup_dev.

`rustup` will install toolchains correctly, we don't need to care about them.


scripts/setup/dev_setup.sh
```diff
@@ -296,17 +296,6 @@ function install_cargo_binary {
 	fi
 }
 
-function install_toolchain {
-	version=$1
-	echo "==> Installing ${version} of rust toolchain..."
-	rustup install "$version"
```
everpcpc, line 302: this step is needed for build-tool image

Xuanwo: Addressed by running `cargo version`

Xuanwo: For more background: https://note.xuanwo.io/#/page/Rustup

everpcpc: this is needed for cross build, since it would only install x86 gnu

everpcpc: Oh my mistake, cross toolchain setup is in separate command.

everpcpc (approved):

scripts/setup/dev_setup.sh
everpcpc: so  `cargo version` is not needed.

PsiACE (approved):

PR #6561, opened 2022-07-11
refactor: split common-users

PsiACE:

## Summary

After this split:
- [x] query/users will be available as a common crate named common-users
- [x] refactor, replace query/users with common-users
  - [x] `users/auth/auth_mgr` -> `query/auth`, because of `QueryContext`

Related #6180


Xuanwo:
How about moving user-related structs out of meta-types?

PsiACE:
> How about moving user-related structs out of meta-types?

I will investigate it after replacing `query/users`.

drmingdrmer (approved):
looks great to me!

BohuTANG:
looks great to me!

PR #6565, opened 2022-07-11
fix(query): fix values parser in skip_to_next_row

sundy-li:

## Summary

Summary about this PR

 Fix values parser in skip_to_next_row, If `'` is escaped by `\`, it will continue to search a valid row.

Fixes #issue


query/src/sql/statements/value_source.rs
```diff
@@ -184,8 +184,15 @@ pub fn skip_to_next_row<R: BufferRead>(
             let c = buffer[it];
             reader.consume(it + 1);
 
+            if it == 0 && escaped {
```
b41sh, line 187: `(` and `)` can't be escaped.

sundy-li: `escaped` just works for `'`.

b41sh: if the buffer is `\(`, this loop will continue, `balance += 1` will not executed.

sundy-li:
> How can we guarantee it sync?

Oh, I did not notice there is `ValueSourceV2`, maybe we should keep updating v2 until we remove v1 in the future.

fkuner:
> 

Yes, this is awful. Maybe we better remove v1 as soon as possible.

PR #6568, opened 2022-07-11
refactor: replace infallible

PsiACE:

## Summary

After this PR:

- [x] remove useless code in `common_base::infallible`, they are just some wrappers, we can use `parking_lot` or `std::sync`
  - ~the `Mutex` in std is likely to perform better and we can use it.~

- [x] for `ReentrantMutex` & `ReentrantMutexGuard` in `parking_lot`
  - `ReentrantMutexGuard` does not give mutable references to the locked data. Use a `RefCell` if you need this.
  - but `RefCell` is `!Sync`, perhaps we could use threadsafe `AtomicRefCell`


PsiACE:
this PR broke the cluster mode and I'll try to fix it tomorrow.

query/src/sessions/query_ctx_shared.rs
```diff
@@ -33,6 +31,8 @@ use common_planners::PlanNode;
 use common_users::RoleCacheMgr;
 use common_users::UserApiProvider;
 use futures::future::AbortHandle;
+use parking_lot::Mutex;
```
Xuanwo, line 34: As std's `Mutex` has been improved a lot, we still need `parking_lot`?

PsiACE: We can use `std::sync::Mutex`, emmm, only one global replacement required

Xuanwo: Also: https://github.com/Amanieu/parking_lot/issues/338

Xuanwo:
@PsiACE Nice work! you are databend's code guard!

query/src/api/rpc/exchange/exchange_manager.rs
```diff
@@ -95,7 +97,8 @@ impl DataExchangeManager {
 
     // Execute query in background
     pub fn execute_partial_query(&self, query_id: &str) -> Result<()> {
-        let mut queries_coordinator = self.queries_coordinator.lock();
+        let queries_coordinator_lock = self.queries_coordinator.lock();
+        let queries_coordinator = unsafe { &mut *queries_coordinator_lock.as_ptr() };
```
PsiACE, line 101: This is almost equivalent to https://github.com/datafuselabs/databend/pull/6568/files#diff-b071b784b4bf25a377c65303683142fab21ad983b9ab57834a10fcc0ed3898c6

```
pub struct ReentrantMutexGuard<'a, T> {
    inner: ParkingReentrantMutexGuard<'a, UnsafeCell<T>>,
}

impl<'a, T> DerefMut for ReentrantMutexGuard<'a, T> {
    fn deref_mut(&mut self) -> &mut Self::Target {
        unsafe { &mut *self.inner.deref().get() }
    }
}
```

but I don't think we need to achieve it anymore.

Xuanwo: cc @zhang2014

query/src/sessions/query_ctx_shared.rs
PsiACE: With fewer threads, `parking_lot` still performs better, and given our current instances, the replacement to `std::sync::Mutex` doesn't feel necessary.

but, perhaps on some crates it helps us to reduce dependencies.

Xuanwo (approved):

query/src/api/rpc/exchange/exchange_manager.rs
zhang2014: Great job.

query/src/interpreters/async_insert_queue_v2.rs
```diff
@@ -294,10 +288,14 @@ impl AsyncInsertQueue {
...
+
                 let query_need_abort = ctx.query_need_abort();
-                let executor =
-                    PipelineCompleteExecutor::try_create(self.runtime.clone(), query_need_abort, pipeline).unwrap();
+                let executor = PipelineCompleteExecutor::try_create(
+                    self.runtime.clone(),
+                    query_need_abort,
+                    pipeline,
+                )
+                .unwrap();
```
zhang2014, line 298:
```suggestion
?;
```

zhang2014 (approved):
Great job.

Issue #6547, opened 2022-07-08
Tracking issue for new expression framework


andylokandy:
This is a tracking issue for the design and implementation of the new scalar expression framework.

### Summary

The new expression framework contains these improvements:

1. **Compile-time** (for SQL, aka, when planning) **type checking**, which could capture all type errors in the beginning, and provide nice-looking error reports.
2. **Untyped expression evaluation**, which means function evaluation will no longer worry about data typing.
3. **Auto vectorization, generics, auto downcasting**, and more ergonomic improvements for writing SQL function.
4. **Distributed evaluation** in mind on the first day.

### Reference

[RFC: Formal Type System](https://github.com/datafuselabs/databend/discussions/5438)
[Demo: Typed Type Exercise in Rust](https://github.com/andylokandy/typed-type-exercise-in-rust)

### Migration plan

The new expression framework will replace the legacy `common-datavalues`, `common-datablocks` and refactor all functions in `common-function`. So it's a tough task, and we are going to break it into small steps:

1. Add a new `common-expression` crate with expression definition (AST), type checking, and evaluation runtime.
2. Run benchmark on the new framework and make necessary improvements.
3. Migrate all functions to the new framework.

#### features

- [x] https://github.com/datafuselabs/databend/pull/6576
- [x] https://github.com/datafuselabs/databend/pull/6649
- [x] https://github.com/datafuselabs/databend/pull/7054
- [x] https://github.com/datafuselabs/databend/pull/6674
- [x] https://github.com/datafuselabs/databend/pull/6663
- [x] https://github.com/datafuselabs/databend/pull/6661
- [x] https://github.com/datafuselabs/databend/pull/6662
- [x] https://github.com/datafuselabs/databend/pull/6597
- [x] https://github.com/datafuselabs/databend/pull/6712
- [x] https://github.com/datafuselabs/databend/pull/6918
- [x] https://github.com/datafuselabs/databend/pull/7075
- [x] https://github.com/datafuselabs/databend/issues/7020
- [x] #6636
- [x] #6635
- [x] #6634
- [x] https://github.com/datafuselabs/databend/pull/7781
- ~~[ ] Conversion between `Boolean` and other types~~  no plan to support 
- ~~[ ] Conversion between `String` and other types~~ no plan to support 
- [ ] Literal deserialization

#### refactor

- [x] https://github.com/datafuselabs/databend/pull/6677
- [x] https://github.com/datafuselabs/databend/pull/6787
[...]

PR #6576, opened 2022-07-12
feat(expr): add new crate `common-expression`

andylokandy:

## Summary

Initial commit for the new scalar expression framework. See [tracking issue](https://github.com/datafuselabs/databend/issues/6547) for more information.

- [x] Expression definition
- [x] Expression type checker
- [x] Expression evaluator


andylokandy:
```
ERROR the following files don't have a valid license header: 
common/expression/tests/it/testdata/run-ast 
```

How can I bypass this check for goldenfiles?

PsiACE:
> How can I bypass this check for goldenfiles?

edit https://github.com/datafuselabs/databend/blob/main/.licenserc.yaml

andylokandy:
@PsiACE But why common-ast bypass the check without being included in this config?

common/expression/src/function.rs
```diff
@@ -0,0 +1,691 @@
...
+        assert!(
+            !has_nullable,
+            "Function {} has nullable argument or output, please use register_1_arg_core instead",
+            name
+        );
+
+        let property = property.preserve_not_null(true);
+
+        self.register_1_arg_core::<NullType, NullType, _>(name, property.clone(), move |_, _| {
+            Value::Scalar(())
```
ygf11, line 175: Maybe we should keep column length if it is a column?

common/expression/src/expression.rs
```diff
@@ -0,0 +1,70 @@
...
+    ColumnRef {
+        id: usize,
+        data_type: DataType,
+        property: ValueProperty,
+    },
+    // Cast {
+    //     is_try: bool,
+    //     expr: Box<Expr>,
+    //     dest_type: DataType,
+    // },
```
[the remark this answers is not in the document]
andylokandy: unless we've figured out a way to make cast a normal function, we have to keep it.

common/expression/src/function.rs
andylokandy: Is seems not necessary to keep a column length for a scalar. Is there a place we had used?

Xuanwo:
> @PsiACE But why common-ast bypass the check without being included in this config?

Nice question, I did some research:

https://github.com/apache/skywalking-eyes/blob/9b6b981a99fb9179f1c5aa58ac6474ea1dbdd552/pkg/header/config.go#L143

`skywalking-eyes` will ignore all `*.txt` files, and `common-ast`'s golden files are suffixed by `.txt`.

common/expression/src/function.rs
ygf11: Thanks, got it, NullType is just an inner data type.

andylokandy: NullType is not a inner type, but similar to `Nullable<Nothing>` in clickhouse.

sundy-li (approved):
LGTM, yet needs to cover the `Object` Column.

common/expression/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+# Workspace dependencies
+common-arrow = { path = "../arrow" }
+
+# Github dependencies
+
+# Crates.io dependencies
+chrono-tz = "0.6.1"
+enum-as-inner = "0.4"
+# https://github.com/calder/rust-goldenfile/pull/7
+goldenfile = { git = "https://github.com/datafuse-extras/rust-goldenfile", rev = "16c5783" }
```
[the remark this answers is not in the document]
Xuanwo: Also cc @PsiACE, we can remove `datafuse-extras/rust-goldenfile` after `goldenfile` upgraded to `1.4.3`.

common/expression/src/function.rs
ygf11: I think it still need column length here. It is not clear in 1 arg function.

For example, we have a 2-arg-function `plus`, the inputs are `Literal::Null` and`ColumnRef(UInt8)`.

The type check will select `plus(Null, UInt8)` as the finaly function.

The result will be a scalar, so it is not correct.

common/expression/Cargo.toml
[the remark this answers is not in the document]
andylokandy: I have updated all goldenfiles to 1.4.3.

sundy-li:
flaky sqllogic test, I did not find the reason. cc @ZeaLoVe

ZeaLoVe:
> flaky sqllogic test, I did not find the reason. cc @ZeaLoVe

9wwoCI3cuZ

Not a flaky error, this sql error code change, expect 1006 but get 1065.

BohuTANG:
> > flaky sqllogic test, I did not find the reason. cc @ZeaLoVe
> 
> 9wwoCI3cuZ
> 
> Not a flaky error, this sql error code change, expect 1006 but get 1065.

How can we fix it, in this PR or another?

youngsofun:
@BohuTANG  fixed  https://github.com/datafuselabs/databend/pull/6593

Xuanwo:
@mergify update

PR #6582, opened 2022-07-12
feat: add share metasrv ShareApi(create_share,drop_share)

lichuang:

## Summary

feat: add share metasrv ShareApi(create_share,drop_share)

Fixes https://github.com/datafuselabs/databend/issues/6581


common/protos/proto/share.proto
```diff
@@ -0,0 +1,60 @@
...
+}
+
+message ShareGrantObject {
+  uint64 ver = 100;
+  uint64 min_compatible = 101;
+
+  oneof object {
+    DatabaseNameIdent database = 1;
+    TableNameIdent table = 2;
+  }
```
drmingdrmer, line 39: Isn't there a consistency issue if a `share` identifies a DB or table by name but not by id?

common/proto-conv/src/share_from_to_protobuf_impl.rs
```diff
@@ -0,0 +1,195 @@
...
+use common_protos::pb;
+use enumflags2::BitFlags;
+
+use crate::check_ver;
+use crate::FromToProto;
+use crate::Incompatible;
+use crate::MIN_COMPATIBLE_VER;
+use crate::VER;
+
+impl FromToProto<pb::ShareNameIdent> for mt::ShareNameIdent {
```
drmingdrmer, line 33: There should be a test to ensure compatibility for this version, just like:
https://github.com/datafuselabs/databend/blob/658f1957256cbb8e0459ef175f5f24a54785830e/common/proto-conv/tests/it/proto_conv.rs#L199-L235

BTW, since new types are added, `VER` should be increased.

BohuTANG:
Looks like sqllogic test is not stable for some test cc @ZeaLoVe

common/proto-conv/src/util.rs
```diff
@@ -14,7 +14,7 @@
 
 use crate::Incompatible;
 
-pub const VER: u64 = 1;
+pub const VER: u64 = 2;
```
Xuanwo, line 17: We are going to increase the VER of proto in this PR?

common/proto-conv/src/share_from_to_protobuf_impl.rs
lichuang: > There should be a test to ensure compatibility for this version, just like:
> 
> https://github.com/datafuselabs/databend/blob/658f1957256cbb8e0459ef175f5f24a54785830e/common/proto-conv/tests/it/proto_conv.rs#L199-L235
> 
> BTW, since new types are added, `VER` should be increased.

@Xuanwo

common/proto-conv/src/util.rs
flaneur2020: what does increase the VER here means? 👀

drmingdrmer: It means it is another new version of meta-types, since it introduces `share` related types and has been different from the previous version of meta-types.

A new version databend-query that depends on `share` data can set `MIN_COMPATIBLE=2` so that it will read `share` data happily and will refuse to read metadata with `VER=1` that does not have `share` related data.

drmingdrmer: > We are going to increase the VER of proto in this PR?

Hmm... it would be nice if every time changing meta types there is a new `VER` value.
This way we'll be able to get rid of old version meta types in fine-grained progress.

drmingdrmer (approved):
looks greater to me :D

PR #6597, opened 2022-07-13
feature(expr): implement pretty print for `Chunk`

andylokandy:

## Summary

- Add checks for column length
- Pretty-print test output

Related https://github.com/datafuselabs/databend/issues/6547


common/expression/src/chunk.rs
```diff
@@ -12,8 +12,29 @@
...
 use crate::values::Column;
 
 pub struct Chunk {
-    pub columns: Vec<Column>,
+    columns: Vec<Column>,
+}
+
+impl Chunk {
+    pub fn new(columns: Vec<Column>) -> Self {
+        assert!(columns.iter().map(|col| col.len()).all_equal());
```
sundy-li, line 25: debug_assert!

common/expression/tests/it/testdata/run-ast.txt
```diff
@@ -2,119 +2,466 @@ raw expr       : and(true::Boolean, false::Boolean)
...
 type           : Array<Nothing>
 property       : {not_null}
 checked expr   : create_array<>()
-eval result    : EmptyArray
+evaluation result:
+[]
+
 
 raw expr       : create_array(NULL, true::Boolean)
 type           : Array<Nullable<Boolean>>
```
sundy-li, line 305: `Array(X)` may be better than `Array<X>`, the former follows the SQL Parser.
So does `Nullable(x)`

BohuTANG (approved):

PR #6609, opened 2022-07-13
fix: range delete panic and incorrect statistics (of in_memory_size)

### LINKED ISSUE #6570: bug: Delete using range query
### Search before asking

### Version

```
mysql> select version;
+------------------------------------------------------------------------------------------+
| version()                                                                                |
+------------------------------------------------------------------------------------------+
| 8.0.26-v0.7.121-nightly-55b9ab2-simd(rust-1.64.0-nightly-2022-07-07T16:36:21.124385335Z) |
+------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

### What's Wrong?

```
mysql> delete from t where id > 8604450183688801682;
ERROR 1105 (HY000): Code: 1068, displayText = Cannot join handle from context's runtime, cause: task 143877 panicked.
mysql> delete from t where id = 8684450183688801682;
Query OK, 0 rows affected (0.41 sec)
```

### How to Reproduce?

_No response_

dantengsky:

## Summary

- fix range deletion panic

> Dring deletion, multiple segments may become empty, while removing those segments from the snapshot, the index being used is NOT correct in some cases. this may lead to an index out of bound err, or even worse, silently deleting incorrect segments.

- incorrect statistics of block `in_memory_size`

  compressed/unpressed size of block/segment/table are not effected

- use column size(length in bytes, of column persistent in file), instead of column's `in_memory_size` for statistics of read_plan phase

Fixes #6570

### DIFF
--- a/query/src/storages/fuse/operations/delete.rs
+++ b/query/src/storages/fuse/operations/delete.rs
@@ -23,8 +23,8 @@ use common_tracing::tracing::debug;
 use crate::sessions::QueryContext;
 use crate::storages::fuse::meta::TableSnapshot;
 use crate::storages::fuse::operations::mutation::delete_from_block;
-use crate::storages::fuse::operations::mutation::mutations_collector::Deletion;
-use crate::storages::fuse::operations::mutation::mutations_collector::DeletionCollector;
+use crate::storages::fuse::operations::mutation::deletion_mutator::Deletion;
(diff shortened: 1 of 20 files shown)


BohuTANG:
@zhyass PTAL

zhyass (approved):
LGTM

PR #6612, opened 2022-07-14
feat(storage): Support read nested columns

### LINKED ISSUE #5786: Feature: Support Struct data type
**Summary**

1. implement `StructColumn`
2. support map access for `StructColumn`

b41sh:

## Summary

Support `tuple` and `named tuple` data type
```sql
MySQL root@127.0.0.1:default> create table t(id int, s tuple(a int, b int));
Query OK, 0 rows affected
Time: 0.051s
MySQL root@127.0.0.1:default> insert into t values(1, (10, 20)),(2, (30, 40));
Query OK, 0 rows affected
Time: 0.068s
MySQL root@127.0.0.1:default> select * from t;
+----+----------+
| id | s        |
+----+----------+
| 1  | (10, 20) |
| 2  | (30, 40) |
+----+----------+
2 rows in set
Time: 0.063s
```

Support read `struct` data type.

As the `struct` data type has nested child fields, a `StructColumn` may have multiple nested columns. We can't find all leaf columns meta Infos by the index of the column in `col_metas`. In order to solve this problem, we add `Vec<ColumnLeaf>`, which is generated from the `SchemaDescriptor` of `parquet`. It is a nested tree structure, `leaf_ids` is the index of the leaf column, and `children` is the optional child columns. We can get the meta Infos of the leaf column by `leaf_ids`.

```rust
/// `ColumnLeaf` contains all the leaf column ids of the column.
/// For the nested types, it may contain more than one leaf column.
#[derive(Debug, Clone)]
pub struct ColumnLeaf {
    pub name: String,
    // `leaf_ids` is the indices of all the leaf columns in DFS order,
    // through which we can find the meta information of the leaf columns.
    pub leaf_ids: Vec<usize>,
    // Optional children column for nested types.
    pub children: Option<Vec<ColumnLeaf>>,
}
```

Fixes #5786


sundy-li:
cc @andylokandy  Is it ok to have `named tuple` data type in the new type system ?

dantengsky (approved):
LGTM

dantengsky:
@sundy-li any other concerns?

BohuTANG (approved):

PR #6618, opened 2022-07-14
feat(planner): support create table as select in planner v2

### LINKED ISSUE #6427: Feature: Support `CREATE TABLE ... AS ...` in new planner

yuuch:

## Summary

Feature(planner): support create table as select in planner v2

Closes #6427


query/src/sql/planner/binder/ddl/table.rs
```diff
@@ -369,8 +372,32 @@ impl<'a> Binder {
...
+                    .columns
+                    .iter()
+                    .map(|column_binding| {
+                        DataField::new(
+                            &column_binding.column_name,
+                            *column_binding.data_type.clone(),
+                        )
+                    })
+                    .collect();
+                let  source_fields = source_schema.fields().clone();
```
leiysky, line 391: I'm not sure but does this need a reformat?

leiysky, line 391: I'm not sure but does this need a reformat?

query/src/sql/planner/plans/create_table_v2.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub catalog: String,
+    pub database: String,
+    /// The table name
+    pub table: String,
+
+    pub table_meta: TableMeta,
+
+    pub cluster_keys: Vec<String>,
+    #[serde(skip)]
+    pub as_select: Option<Box<Plan>>,
```
leiysky, line 41: Maybe you can use a optimized `PhysicalPlan` here, which is serializable and can be compiled into `Pipeline` easily.

query/src/sql/planner/binder/ddl/table.rs
yuuch: Yes. For `Create table  with specified columns`, the fields will be the Union of the `source` and `query` part.  I didn't found any method like append in `DataSchema` :(

Xuanwo: Weird, check passed...

query/src/interpreters/interpreter_table_create_v2.rs
```diff
@@ -0,0 +1,187 @@
...
+        // The casting and schema checking is in interpreter_insert.rs, function check_schema_cast.
+        let table_schema = table.schema();
+        let select_fields: Vec<DataField> = select_plan
+            .schema()
+            .fields()
+            .iter()
+            .filter_map(|f| table_schema.field_with_name(f.name()).ok())
+            .cloned()
+            .collect();
+        let schema = DataSchemaRefExt::create(select_fields);
```
sundy-li, line 156: Can we enable omitting the schema like:

```
> create table tt as select * from t1;

Elapsed: 11 ms

```

query/src/sql/planner/binder/ddl/table.rs
Xuanwo: I think it's a bug of `rustfmt`.

image

Xuanwo: Code in this block won't be formatted, for example:

image

yuuch: Sorry， I mistake   :(

Xuanwo: This format bug will be addressed by https://github.com/datafuselabs/databend/pull/6631

query/src/sql/planner/binder/ddl/table.rs
```diff
@@ -369,8 +372,32 @@ impl<'a> Binder {
...
+                let source_fields = self.concat_fields(source_fields, query_fields);
+                (
+                    DataSchemaRefExt::create(source_fields.to_vec()),
+                    source_coments,
+                )
+            }
```
leiysky, line 397: I found the semantic of `CREATE TABLE AS` with column definitions is wierd, and both `MySQL` and `Postgres` doesn't support it. 

I suggest to raise any error for such case.

yuuch: What do you mean?
 I run the create table with specified columns in mysql. It supported that. 
```sql
mysql> create table t1(i INT, j Int);

mysql> insert into t1 values (1, 2);
Query OK, 1 row affected (0.00 sec)

mysql> create table t2 (k Int, l Int) as select * from t1;
Query OK, 1 row affected (0.00 sec)
Records: 1  Duplicates: 0  Warnings: 0

mysql> desc t2;
+-------+------+------+-----+---------+-------+
| Field | Type | Null | Key | Default | Extra |
+-------+------+------+-----+---------+-------+
| k     | int  | YES  |     | NULL    |       |
| l     | int  | YES  |     | NULL    |       |
| i     | int  | YES  |     | NULL    |       |
| j     | int  | YES  |     | NULL    |       |
+-------+------+------+-----+---------+-------+
4 rows in set (0.00 sec)
mysql> select version();
+-----------+
| version() |
+-----------+
| 8.0.21    |
+-----------+
1 row in set (0.02 sec)
```

query/src/interpreters/interpreter_table_create_v2.rs
yuuch: > Can we enable omitting the schema like:
> 
> ```
> > create table tt as select * from t1;
> 
> Elapsed: 11 ms
> ```

1.
```sql
create table tt as select * from t1;
```
here, we can use the `table_schema`(line 148) and omit the `schema`(line 156).


2.
```sql
create table tt (i INT) as select * from t1
```
the `table_schema` is different from the `schema` ,  so it can't be omitted.

query/src/sql/planner/binder/ddl/table.rs
leiysky: I was wrong, MySQL does support this. 

If the data types of coulumn definition is incompatible with the select result, we may need to raise an error for that.

sundy-li: If the schema is not matched, there will be a `TransformCastSchema` between them.

This is some as:  `insert table t select * from t2`

query/src/interpreters/interpreter_table_create_v2.rs
sundy-li: Need a stateless test to cover:

```
create table tt as select * from t1;
```

query/src/sql/planner/binder/ddl/table.rs
yuuch: But I found that it is needed to check the cast type here. Otherwise, the table has been created, even though the server throws an error from the `InsertInterpreter`. :)

leiysky: > But I found that it is needed to check the cast type here. Otherwise, the table has been created, even though the server throws an error from the `InsertInterpreter`. :)

By the way, we have a plan for atomic DDL, which is tracked in #6625. You can take a took if you are interested in it.

leiysky (approved):
LGTM, nice work.

sundy-li (approved):

query/src/interpreters/interpreter_table_create_v2.rs
sundy-li: I'll add it later in #6648

Xuanwo:
image

Haha, fixed.

PR #6599, opened 2022-07-13
ci(tests): add mini dataset to stateful test

ZeaLoVe:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Add mini dataset to stateful test, include ontime and hits

Fixes #issue

PR #6623, opened 2022-07-14
feat: Allow COPY and CREATE STAGE from public read buckets without credentials

Xuanwo:

## Summary

feat: Allow COPY and CREATE STAGE from public read buckets without credentials

This PR will make the following query works:

```shell
COPY INTO ontime_mini FROM 's3://repo.databend.rs/dataset/stateful/ontime_2006_100000.csv' 
    FILE_FORMAT = ( type = 'CSV' field_delimiter = ',' record_delimiter = '\n' skip_header = 1 );
```


BohuTANG:
Can we add a stateful test for it with minio?

Xuanwo:
> Can we add a stateful test for it with minio?

We will continue those work after https://github.com/datafuselabs/databend/pull/6599

I plan to introduce a better and fine-grained stateful test case.

BohuTANG (approved):

PR #6626, opened 2022-07-14
feat(executor): Add source pipelines for `Pipeline`

leiysky:

## Summary

Add source pipelines to represent the dependency relationship between pipelines of a query.

After this PR, `Pipeline` will become a tree-like structure.

### DIFF
--- a/tests/logictest/suites/query/join/inner.test
+++ b/tests/logictest/suites/query/join/inner.test
@@ -2,10 +2,10 @@ statement ok
 set enable_planner_v2 = 1;
 
 statement ok
-create database inner;
+create database db;
 
 statement ok
-use inner;
+use db;
 
 statement ok
 create table t(a int);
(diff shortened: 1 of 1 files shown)


leiysky:
> Pipeline should not be tree structured. Maybe you want is `PipelineExecutor::from_pipelines`?

I have introduced a `flatten` method in this PR, which can leverage with `PipelineExecutor::from_pipelines`.

I do this because I found that we are passing single `Pipeline` everywhere, if we want to allow mutiple pipelines for a single query, we'd better to introduce a new struct to involve the pipelines.

Since there is dependency relationship between pipelines, so I made it a tree-like structure. As soon as we are going to execute them, we can `flatten` the tree into a `Vec`, so we can pass single `Pipeline` everywhere.

How do you think of this?

zhang2014:
Vec<Pipeline> is better. Pipeline should not be exchange between nodes. It is always build by query plan. This patch doesn't seem to bring us anything. The dependency between pipelines should be guaranteed by the executor(operation of data port). May use `from_pipelines` are better.

Issue #6452, opened 2022-07-05
Track: full subquery support and optimization


xudong963:
**Summary**
- [x] https://github.com/datafuselabs/databend/issues/6411
- [x] https://github.com/datafuselabs/databend/pull/6412 (uncorrelated InSubquery and AnySubquery)
- [x] https://github.com/datafuselabs/databend/pull/6534
- [ ] https://github.com/datafuselabs/databend/issues/6450 (correlated InSubquery and AnySubquery)
- [ ] https://github.com/datafuselabs/databend/issues/6413 (Scalar subquery will use Single Join to rewrite)
- [ ] https://github.com/datafuselabs/databend/issues/6451 (use marker join to process exists subquery with disjunction subquery)

PR #6632, opened 2022-07-14
feat(planner): optimize correlated subquery by decorrelation

xudong963:

## Summary
Optimize correlated subquery, for more details please see https://btw-2015.informatik.uni-hamburg.de/res/proceedings/Hauptband/Wiss/Neumann-Unnesting_Arbitrary_Querie.pdf
- [x] Support single join
- [x] Architecting the decorrelation framework
- [x] Decorrelate scalar subquery
- [x] Decorrelate exists subquery
- [x] Decorrelate any subquery
- [x] Remove `CrossApply` related code
- [x] Rebench tpch with scale factor 1
- [x] Add a rich test suit for correlated subquery to ensure the framework  is solid

Fixes  https://github.com/datafuselabs/databend/issues/6452


xudong963:
The ticket is ready to review. cc @leiysky @sundy-li 

Before the ticket, tpch q2 and q17 will **OOM** with 1G tpch data

After the ticket, tpch q2 needs **~2.29 sec** in my local, tpch q17 needs **~9.97s** in my local.

The reason for the difference in performance data is that in this ticket, `CrossApply` is removed, all kinds of subqueries are converted to the corresponding join, and at the execution level, there will be no more subqueries.

(I'm not sure this ticket can handle all sorts of weird subqueries, but it can already cover common use cases, and we can iterate on it later)

query/src/sql/optimizer/heuristic/decorrelate.rs
```diff
@@ -12,224 +12,725 @@
...
+    fn flatten(&mut self, plan: &SExpr, correlated_columns: &ColumnSet) -> Result<SExpr> {
+        let rel_expr = RelExpr::with_s_expr(plan);
+        let prop = rel_expr.derive_relational_prop()?;
+        if prop.outer_columns.is_empty() {
+            // Construct a LogicalGet plan by correlated columns.
+            // Finally generate a cross join, so we finish flattening the subquery.
+            let mut metadata = self.metadata.write();
+            let table_index = metadata
+                .table_index_by_column_indexes(correlated_columns)
+                .unwrap();
```
leiysky, line 412: Does this work if we have such a query:
```sql
select * from (select a + 1 as a from t) as t where exists(select * from t as t1 where t1.a < t.a)
```

leiysky: I think maybe you can copy the left child of dependent join to replace the `LogicalGet` here, since we don't have mechanism to reuse the output of a processor yet.

xudong963: Yes
```sql
mysql> create table t(a int);

mysql> select * from (select a + 1 as a from t) as t where exists(select * from t as t1 where t1.a < t.a);
+------+
| a    |
+------+
|    2 |
|    3 |
+------+
2 rows in set (0.04 sec)
Read 4 rows, 16.00 B in 0.004 sec., 922.28 rows/sec., 3.60 KiB/sec.
```

xudong963: > I think maybe you can copy the left child of dependent join to replace the `LogicalGet` here, since we don't have mechanism to reuse the output of a processor yet.

In fact, I want to add a new `rel_op` later which contains columns from left child of dependent join and will eliminate duplicate rows.

leiysky: > In fact, I want to add a new `rel_op` later which will get distinct columns from left child of dependent join.

We can do that later. In DuckDB, they have `DelimScan` and `DelimJoin` to do the work, which is heavily dependent on the runtime.

xudong963: > We can do that later. In DuckDB, they have `DelimScan` and `DelimJoin` to do the work, which is heavily dependent on the runtime.

Ok

xudong963: > Well, seems this can be converted to semi join.
> 
> What about:
> 
> ```sql
>  create table t(a int);
> insert into t values(1), (2);
> select * from (select a + 1 as a from t) as t where (select sum(a) from t as t1 where t1.a < t.a) = 1;
> ```

I get you, will give a try to copy the left child of dependent join.

xudong963: Derived an issue: https://github.com/datafuselabs/databend/issues/6688, will support the case in the next PR.

xudong963:
Need help, logic test failed https://pipelines.actions.githubusercontent.com/serviceHosts/8b8d3c09-a67f-454e-ae9f-d6ceaf4eb7aa/_apis/pipelines/1/runs/69966/signedlogcontent/60?urlExpires=2022-07-19T12%3A10%3A15.1666887Z&urlSigningMethod=HMACV1&urlSignature=j9mLqq2D0kpKsHix0flFBwuOX%2BOsvcvp%2BhUcDorFx6A%3D @ZeaLoVe

query/src/pipelines/processors/transforms/hash_join/result_blocks.rs
```diff
@@ -180,54 +193,67 @@ impl JoinHashTable {
...
+            .hash_join_desc
+            .other_predicate
+            .as_ref()
+            .unwrap()
+            .eval(&func_ctx, &merged_block)?;
+        let filter_column = type_vector.vector();
+        let mut row_ptrs = self.row_ptrs.write();
+        for (idx, build_index) in build_indexs.iter().enumerate() {
+            let self_row_ptr = row_ptrs.iter_mut().find(|p| (*p).eq(&build_index)).unwrap();
+            match filter_column.get(idx) {
```
sundy-li, line 245: We can use BooleanViewer here

leiysky (approved):

PR #6637, opened 2022-07-14
feat(query): simple cte

fkuner:

## Summary

A experiment about cte. The design needs to be disscussed. 

For example: Is it resonable to put `{cte_name: subquery}` in `BindContext`?

After some comments, I will write formal RFC.

It supports majority of syntax in test file `20_0010_statement_cte.sql`. But I think it may be prone to bug.

Hope to give some suggestions.

Fixes #issue


leiysky:
You can put the CTEs map into `Binder` itself, since it's a global state of query.

fkuner:
> You can put the CTEs map into `Binder` itself, since it's a global state of query.

Yes,  we need a map which lifetime is related to current subquery. I will think about what's better way to implement.

For example, we have a table `test1` and execute the sql:

```
SELECT * FROM (WITH test1 AS (SELECT to_int32(number) i FROM numbers(5)) SELECT * FROM test1) l INNER JOIN test1 r on l.i = r.i;
```

The right table of inner join is not same as `WITH test1 AS (SELECT to_int32(number) i FROM numbers(5)`, because right test1 not in scope of test1 defined in left.

leiysky:
> > You can put the CTEs map into `Binder` itself, since it's a global state of query.
> 
> Yes, we need a map which lifetime is related to current subquery. I will think about what's better way to implement.
> 
> For example, we have a table `test1` and execute the sql:
> 
> ```
> SELECT * FROM (WITH test1 AS (SELECT to_int32(number) i FROM numbers(5)) SELECT * FROM test1) l INNER JOIN test1 r on l.i = r.i;
> ```
> 
> The right table of inner join is not same as `WITH test1 AS (SELECT to_int32(number) i FROM numbers(5)`, because right test1 not in scope of test1 defined in left.

Oh, I see. Then it's fine to put the map inside `BindContext`.

xudong963:
Maybe we can close the stale pr? @fkuner

fkuner:
> Maybe we can close the stale pr? @fkuner

We can close it. This pr has some problems I haven't solved.

PR #6645, opened 2022-07-15
fix(query): fix variant map access filter

### LINKED ISSUE #6590: Feature: Use object data types in select where
## Create Table

```sql
CREATE TABLE test (labels OBJECT);
```

## Insert Test Data

```sql
INSERT INTO test VALUES (parse_json('{"email": "test@test.com"}'));
```

## Example

```sql
mysql> select labels:email from test where labels:email = "test@test.com";
ERROR 1105 (HY000): Code: 1010, displayText = Cast error happens in casting from Variant to String (while in processor thread 0).

mysql> select labels:email from test where labels.email = "test@test.com";
ERROR 1105 (HY000): Code: 1058, displayText = Unknown column labels.email.

mysql> select labels:email from test where get(labels, "email") = "test@test.com";
ERROR 1105 (HY000): Code: 1010, displayText = Cast error happens in casting from Variant to String (while in processor thread 0).
```

b41sh:

## Summary

Fix conflicting precedence of `MapAccess` and `And`, since the value of both of them is 20, we can't get the correct syntax tree when parsing the following SQL.
```sql
SELECT * FROM test WHERE id = 1 and labels:email = 'abc@test.com';
```

When performing data filtering, if the data types of the comparison are different, we will perform an implicit data type cast. For `JSON` values, some cast can't be successful, such as casting a string to a number or casting null to a string. Since the cast target type is not nullable, the following error will occur.
```sql
mysql> SELECT * FROM test WHERE labels:email = 'abc@test.com';
ERROR 1105 (HY000): Code: 1010, displayText = Cast error happens in casting from Variant to String (while in processor thread 0).
```

In order to solve this problem, we allow the cast target type to be nullable and mark the null value rows as deleted after the comparison operation.

Fixes #6590


sundy-li (approved):

BohuTANG:
We can add more documents about the variant map with this patch @soyeric128

Issue #6069, opened 2022-06-20
Push down filter to storage


leiysky:
Some simple predicates can be pushed down to storage to help pruning partitions, depending on the index types.

For now, databend support min/max index and bloom filter, which can utilize with range filters and equivalent filters. We can push down the predicates as follow pattern:
- `Variable RangeOp Const`: pruning with min/max values
- `Variable = Const`: pruning with bloom filter

leiysky:
Can leverage with #5919

PR #6649, opened 2022-07-15
feat(expr): Implement domain calculation

andylokandy:

## Summary

Calculate the value an expression can return by function's domain-to-range property and the domain of input columns. e.g. if all possible values of `col_a::Nullable<Boolean>`  within a datablock is set `{TRUE, FALSE}`, then within this datablock, `isnull(col_a)` must output values in set `{FALSE}`. 

See more examples in the `domain calculation` sections in the [testfile](https://github.com/datafuselabs/databend/blob/f32b1d3d1bbe590568d04f18ffa5f1aec9b9fe41/common/expression/tests/it/testdata/run-ast.txt).

Partially solve https://github.com/datafuselabs/databend/issues/6069

cc @Xuanwo because it could be used to filter datablocks from storage.


common/expression/src/property.rs
```diff
@@ -12,53 +12,204 @@
...
+
+#[derive(Debug, Clone)]
+pub struct UIntDomain {
+    pub min: u64,
+    pub max: u64,
+}
+
+#[derive(Debug, Clone)]
+pub struct BooleanDomain {
+    pub contains_false: bool,
```
Xuanwo, line 63: How about `has_false` for short?

common/expression/src/property.rs
```diff
@@ -12,53 +12,204 @@
...
+        }
     }
+}
 
-    pub fn commutative(mut self, commutative: bool) -> Self {
-        self.commutative = commutative;
-        self
+impl<T: ValueType> std::fmt::Debug for NullableDomain<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
```
Xuanwo, line 89: How about using `f.debug_struct()`?

common/expression/src/type_check.rs
```diff
@@ -19,60 +19,41 @@ use crate::expression::Literal;
...
-            property,
-        } => Some((
-            Expr::ColumnRef { id: *id },
-            data_type.clone(),
-            property.clone(),
-        )),
         RawExpr::FunctionCall { name, args, params } => {
-            let (mut args_expr, mut args_type, mut args_prop) =
-                (Vec::new(), Vec::new(), Vec::new());
+            let (mut args_expr, mut args_type) = (Vec::new(), Vec::new());
```
Xuanwo, line 35: Maybe we can pre-alloc via `with_capacity` if the size already known?

common/expression/src/display.rs
```diff
@@ -277,3 +262,81 @@ impl Debug for FunctionProperty {
         write!(f, "{self}")
     }
 }
+
+impl Display for NullableDomain<AnyType> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        if let Some(value) = &self.value {
+            if self.contains_null {
+                write!(f, "{} ∪ {{NULL}}", value)
```
Xuanwo, line 270: Is using `|` instead of `∪` a better idea?

For example:

Before: `({FALSE}, {"a"..="e"}) ∪ {NULL}`
After: `({FALSE}, {"a"..="e"}) | {NULL}`

andylokandy: I chose `∪` because it means the union of two sets. `|` seems a little be vague in sematic to me, because `{FALSE, TRUE} | {NULL}` sounds like `oh, it may be ALL BOOLEAN or ALL NULL in the datablock, but anyway you can probe it` instead of `it could be an only could be any scalar in {TRUE, FALSE, NULL}`.

Xuanwo (approved):

common/expression/src/display.rs
Xuanwo: LGTM

leiysky (approved):

Issue #6636, opened 2022-07-14
Support UInt32, UInt64, Int32, Int64, Float32, Float64 in new expression framework


andylokandy:
There are three places new data types need to be implemented:

1. `common_expression::types::DataType`, which stores all logical data types. Since new logical data types are defined, you would like to see `common_expression::type_check::{can_cast_to, common_super_type}` to add coressponding casting rules.
2. `common_expression::values::{Scalar, ScalarRef, Column}`, which stores physical data values. And then runtime casting need to be added to `common_expression::evaluator::Evaluator::run_cast()`, which should excatly following the rules defined in `can_cast_to`.
3. `common_expression::types:::number::Number` trait, for allowing to use them in static generics API.

jiaoew1991:
Hi @andylokandy I will try this issue

andylokandy:
@jiaoew1991 Thank you. I've added some instructions for the task. Feel free to ask if you find anything puzzling.

jiaoew1991:
the float32 float64 have not finished yet，should reopen it

PR #6660, opened 2022-07-16
feat(expr): Support UInt32, UInt64, Int32, Int64 in new expression

jiaoew1991:
Signed-off-by: Enwei Jiao <jiaoew2011@gmail.com>

## Summary

Support more Int family types in new expression

Fixes https://github.com/datafuselabs/databend/issues/6636


common/expression/src/evaluator.rs
```diff
@@ -92,17 +92,67 @@ impl Evaluator {
...
+                (Scalar::Int32(val), DataType::Int64) => {
+                    Some(Value::Scalar(Scalar::Int64(val as i64)))
+                }
+                (Scalar::UInt8(val), DataType::Int64) => {
+                    Some(Value::Scalar(Scalar::Int64(val as i64)))
+                }
+                (Scalar::UInt16(val), DataType::Int64) => {
+                    Some(Value::Scalar(Scalar::Int64(val as i64)))
+                }
+                (Scalar::UInt32(val), DataType::Int64) => {
```
sundy-li, line 143: Should we consider `wrapping cast` from larger size integer to a shorter size integer?

jiaoew1991: there will lose precision after downgrade cast 🤔

common/expression/src/type_check.rs
```diff
@@ -215,10 +219,15 @@ pub fn can_cast_to(src_ty: &DataType, dest_ty: &DataType) -> bool {
...
-        (DataType::UInt8, DataType::UInt16)
-        | (DataType::Int8, DataType::Int16)
-        | (DataType::UInt8, DataType::Int16) => true,
-        _ => false,
+        (src_ty, dest_ty) => {
+            let (info1, info2) = (number_type_info(src_ty), number_type_info(dest_ty));
+            match (info1, info2) {
+                (Some((size1, b1)), Some((size2, b2))) if b1 == b2 => size1 <= size2,
+                (Some((size1, true)), Some((size2, false))) if size2 > size1 => true,
+                (Some((size1, false)), Some((size2, true))) if size1 > size2 => true,
```
ygf11, line 227: Why is long size to short size allowed?

jiaoew1991: It's a bug, thanks for figure it out 😅

common/expression/src/types/number.rs
```diff
@@ -231,6 +231,82 @@ impl Number for u16 {
...
+    }
+
+    fn try_downcast_column(col: &Column) -> Option<Buffer<Self::Storage>> {
+        col.as_u_int32().cloned()
+    }
+
+    fn try_downcast_domain(domain: &Domain) -> Option<Self::Domain> {
+        domain.as_u_int().cloned()
+    }
+    fn upcast_scalar(scalar: Self::Storage) -> Scalar {
```
andylokandy, line 253:
```suggestion

    fn upcast_scalar(scalar: Self::Storage) -> Scalar {
```

common/expression/src/types/number.rs
```diff
@@ -231,6 +231,82 @@ impl Number for u16 {
...
+    }
+
+    fn upcast_column(col: Buffer<Self::Storage>) -> Column {
+        Column::UInt32(col)
+    }
+
+    fn upcast_domain(domain: Self::Domain) -> Domain {
+        Domain::UInt(domain)
+    }
+    fn full_domain() -> Self::Domain {
```
andylokandy, line 264:
```suggestion

    fn full_domain() -> Self::Domain {
```

andylokandy (approved):
LGTM

common/expression/src/evaluator.rs
andylokandy: We're still hesitant whether to throw an evaluation error or wrap the numbers on downgrade cast. Before consensus is met, I'm good to simply disable downgrade cast.

andylokandy:
https://github.com/datafuselabs/databend/pull/6661 has introduced some conflicts. There are two major changes that will affect this PR:

- `run_ast` test helper function now takes an expression string and parses it into `RawExpr`, which is good for ergonomics, but because the `common_ast` parser doesn't recognize negative numbers (`-10` is recognized as `UnaryOp::Minus(Literal::Integer(10))`), you have to add a unary overload for `minus` to construct negative number from positive number.
- `Evaluator::run_cast` now returns `Result`, which means we can throw runtime error for downgrade casts now. With this change, we can allow any casting between numbers in `can_cast_to` and then throw a runtime error when we do find a number is too large or too small.

common/expression/tests/it/main.rs
```diff
@@ -334,6 +491,30 @@ fn builtin_functions() -> FunctionRegistry {
         |lhs, rhs| lhs + rhs,
     );
 
+    registry.register_2_arg::<NumberType<i32>, NumberType<i32>, NumberType<i32>, _, _>(
+        "minus",
+        FunctionProperty::default(),
+        |lhs, rhs| {
+            Some(IntDomain {
+                min: lhs.min.checked_add(rhs.min).unwrap_or(i32::MAX as i64),
+                max: lhs.max.checked_add(rhs.max).unwrap_or(i32::MAX as i64),
```
andylokandy, line 500: The domain calculation seems not right.

common/expression/src/evaluator.rs
```diff
@@ -319,6 +433,38 @@ impl DomainCalculator {
...
+                    max: (*max).min(u64::MAX),
+                }))
+            }
+            (Domain::Int(IntDomain { min, max }), DataType::Int64) => Ok(Domain::Int(IntDomain {
+                min: (*min).max(i64::MIN).min(i64::MAX),
+                max: (*max).max(i64::MIN).min(i64::MAX),
+            })),
+            (Domain::UInt(UIntDomain { min, max }), DataType::Int64) => {
+                Ok(Domain::Int(IntDomain {
+                    min: (*min).min(i64::MAX as u64) as i64,
```
andylokandy, line 464: this might overflow

common/expression/src/type_check.rs
```diff
@@ -322,15 +330,30 @@ pub fn common_super_type(ty1: DataType, ty2: DataType) -> Option<DataType> {
         (DataType::Array(box ty1), DataType::Array(box ty2)) => {
             Some(DataType::Array(Box::new(common_super_type(ty1, ty2)?)))
         }
-        (DataType::UInt8, DataType::UInt16) | (DataType::UInt16, DataType::UInt8) => {
```
andylokandy, line 325: oooops, seems that the case for U8 and I8 was missing.

common/expression/tests/it/main.rs
jiaoew1991: it is same as the **plus** function🤔

common/expression/src/type_check.rs
jiaoew1991: what's the behavior of `(u8,i8)` 🤔，I think this pattern should return None

andylokandy: I think it should be i16. An exception is (u64, i64), and in this case we can return i64.

sundy-li: There is  [type_coercion.rs](https://github.com/datafuselabs/databend/blob/8c3730f25e5beaff68c3eee1f2228f01a0faad42/common/datavalues/src/types/type_coercion.rs)

common/expression/tests/it/main.rs
andylokandy: For example, given `a` in `10..=15` and `b` in `-5..=5`, `minus(a, b)` can be in `5..=20`, because `10-5=5` and `15-(-5) = 20`.

andylokandy: In general, `min=lhs.min-rhs.max` and `max=lhs.max-rhs.min`.

common/expression/tests/it/main.rs
```diff
@@ -362,6 +519,30 @@ fn builtin_functions() -> FunctionRegistry {
...
+    );
+
+    registry.register_2_arg::<NumberType<i64>, NumberType<i64>, NumberType<i64>, _, _>(
+        "multiply",
+        FunctionProperty::default(),
+        |lhs, rhs| {
+            Some(IntDomain {
+                min: lhs.min.checked_add(rhs.min).unwrap_or(i64::MAX),
+                max: lhs.max.checked_add(rhs.max).unwrap_or(i64::MAX),
+            })
```
andylokandy, line 541: This domain calculation for `multiply` is even more complicated. You may return `None` (which will extend the domain to the full range of `i64`) if you want to leave it for later.

common/expression/src/types/number.rs
jiaoew1991: fixed it

common/expression/src/types/number.rs
jiaoew1991: fixed it

common/expression/tests/it/main.rs
jiaoew1991: fixed it

jiaoew1991:
@andylokandy any more review suggestions?

andylokandy:
LGTM now

ygf11:
LGTM

leiysky (approved):

PR #6663, opened 2022-07-17
feat(expr): support user-defined `CAST` and `TRY_CAST`

andylokandy:

## Summary

- Support `CAST` and `TRY_CAST`
- Allow cast between any number types. Once precision downgrade is detected during evaluation, while `CAST` will abort the execution and throw an error, `TRY_CAST` will just silently replace that single scalar with NULL.
- Note that `TRY_CAST` will wrap its dest type into nullable. For example, the return type of `TRY_CAST(a AS Int8)` is `Int8 NULL`, and the return type of `TRY_CAST(a AS Array(Int8))` is `Array(Int8 NULL) NULL`.

### DIFF
--- a/common/expression/src/evaluator.rs
+++ b/common/expression/src/evaluator.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use common_arrow::arrow::bitmap;
+use common_arrow::arrow::bitmap::MutableBitmap;
 use itertools::Itertools;
+use num_traits::ToPrimitive;
 
 use crate::chunk::Chunk;
 use crate::error::Result;
@@ -34,6 +37,7 @@ use crate::values::Column;
 use crate::values::ColumnBuilder;
 use crate::values::Scalar;
 use crate::values::Value;
+use crate::with_number_type;
 
 pub struct Evaluator {
     pub input_columns: Chunk,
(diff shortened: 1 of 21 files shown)


BohuTANG:
Please take a review @leiysky @sundy-li

common/expression/src/evaluator.rs
```diff
@@ -307,6 +127,277 @@ impl Evaluator {
...
+
+            (col, dest_ty) => {
+                // number types
+                with_number_type!(SRC_TYPE, match &col {
+                    Column::SRC_TYPE(col) => {
+                        with_number_type!(DEST_TYPE, match dest_ty {
+                            DataType::DEST_TYPE => {
+                                let mut new_col = Vec::with_capacity(col.len());
+                                for &val in col.iter() {
+                                    let new_val = val.try_into().map_err(|_| {
```
sundy-li, line 264: `try_into` may cause too much branch prediction.

Can we have a faster path for `lower_bits` to `higher_bits` ?

There is `AsPrimitive` trait.

andylokandy: Are you suggesting adding a wrapping mode for the cast?

There are three situations when a cast will present: 

1. When `type_check::can_auto_cast_to` agree to accept an argument with a not totally matched type.
2. When the user uses `CAST` in SQL.
3. When the user uses `TRY_CAST` in SQL.

And also, we could provide four different ways to handle an unsuccessful cast:

1. Wrap it.
2. Make it Null.
3. Give a warning.
4. Throw an error and stop the execution.

My question is what behavior do we use for each different cast situation? And also, @leiysky is suggesting abandoning the fourth behavior (throwing error) because it will be quite unexpected to the user when an error kills a query that has been running for a whole day.

sundy-li: > Are you suggesting adding a wrapping mode for the cast?

Not yet, I was suggesting to improve the performance when casting from low-range number to high-range number.

https://togithub.com/jorgecarleitao/arrow2/issues/252

> And also, we could provide four different ways to handle an unsuccessful cast:

For number cast to number, if overflow happens, we can result in the first behavior (as clickhouse) or the fourth behavior(as duckdb,PostgreSQL) .

If users do not want errors, they can use `try_cast`.

andylokandy: addressed

PR #6665, opened 2022-07-17
refactor: decouple `Table` from `QueryContext`

dantengsky:

## Summary

part of 
- issue #6180

-----
- **Decouple `QueryContext` from `Table` by introducing trait `TableContext`**

   These seem to be one of the biggest knots in way of splitting `query` into sub-crates.
   
  `Catalog`, `Database`, `Table` and `TableContext` are extracted into **one new crate `common-catalog`** (since there are cyclic dependencies between them).
  
  only the interfaces are extracted. The implementations are kept in crate `query`, to keep this PR as small as possible, but they are likely to be able to extract from `query` now.
  
  Trait `TableContext` is a little bit coarse-grained, it could be split it into more dedicated taits later if necessary.
  
  Unit(integration) tests are not transferred yet. They will be addressed in subsequent PRs。
  
  
   
- **other on-demand extracted crates**
  - `common-config`
  - `common-settings`
     **the "semantics" of Settings has changed a little bit, with the assumption that the tenant will not be changed during the lifetime of a given session:**
     
     before this PR, the `Settings` hold a `SessionCtx`, and gives the tenant on demand by using `SessionCtx`
     https://github.com/datafuselabs/databend/pull/6665/files#diff-d2f273cc255b0aee531d4eb86b74f4f5df598bbd861389287630ca2b27f65ea9R65-R67
     
     in this PR, the tenant is acquired during the construction of `Settings`, and never changed afterward.
     
     https://github.com/datafuselabs/databend/pull/6665/files#diff-d2f273cc255b0aee531d4eb86b74f4f5df598bbd861389287630ca2b27f65ea9R219-R227
     
     this arrangement makes `Settings` no longer depend on `SessionCtx`

  - `common-fuse-meta`
      fuse meta and cache
  - `common-pipeline` (basic types)
     only the types that do not depends on `QueryContext` are extracted
  
Fixes #issue


common/catalog/src/catalog.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// Table function
+
+    // Get function by name.
+    fn get_table_function(
+        &self,
+        _func_name: &str,
+        _tbl_args: TableArgs,
+    ) -> Result<Arc<dyn TableFunction>> {
+        unimplemented!()
```
zhang2014, line 160: ErrorCode::Unimplemented is better.

zhang2014 (approved):
Dependencies are becoming complex. I'm not sure whether complex dependencies can solve slow links? But the changes is looks good to me.

common/catalog/src/catalog.rs
dantengsky: OK

query/src/interpreters/fragments/query_fragment_actions.rs
```diff
@@ -29,6 +29,8 @@ use crate::api::ExecutePartialQueryPacket;
 use crate::api::FragmentPlanPacket;
 use crate::api::InitNodesChannelPacket;
 use crate::api::QueryFragmentsPlanPacket;
+use crate::clusters::ClusterHelper;
+use crate::sessions::query_ctx::TableContext;
```
BohuTANG, line 33: `use crate::sessions::TableContext;` looks better?

dantengsky: ok

BohuTANG: Else LGTM

BohuTANG:
To block Mergify merged to the main branch by 2-approves, I changed this PR to draft. Feel free to convert it to ready anytime after the comments are fixed. @dantengsky

dantengsky:
> I'm not sure whether complex dependencies can solve slow links? 

It does not, in my local dev box, for the main crate, `databend-query`, no dramatic improvement found. 

but for the crates that have been extracted. it takes much lesser time to build. I think after this PR, more sub mods could be extracted, or easier to be extracted. and gradually, the main crate will benefit as well.

Issue #5679, opened 2022-05-30
RFC: Unify Session API


leiysky:
# Background
For now, the execution of a query is driven by query handler. There are some disadvantage of this scheme:

- The API is restricted by specific protocol(e.g. mysql, clickhouse), which makes it very hard to implement some new features
- Redundant implementation of query execution. There are multiple versions of `do_query` implemented for clickhouse handler, mysql handler and http handler, which are actually doing the same thing

To resolve the issues, and to provide better extensibility, I propose a unified `Session` API in this RFC.

As we know, database is a typical software in client/server model. Every user will connect to database server with a client(e.g. mysql shell, ODBC driver). Once the connection established, a `Session` will be assigned to user, which stores the state of current
database connection including user information, opened database, session variables and so on.

When user types a query, it will be send to database server via `Session`. As soon as the query execution is finished, the result will be send back to user via `Session` as well.

So we can have a `Session` object like:
```rust
trait Session {
    fn get_variables(&self) -> Result<Vec<Variable>>;
    fn set_variable(&mut self, name: &str, value: &str) -> Result<()>;

    fn execute(&mut self, query: &str) -> Result<DataBlockStream>;
    fn execute_async(&mut self, query: &str) -> Result<QueryID>;
    fn prepare(&mut self, query: &str) -> Result<()>;
}
```

The benifits of a unified `Session` API are:
- No duplicated implementations, all in one
- Possible to have customized features

# Design detail
**WIP**

youngsofun:
apps like snowflake worksheet may have diff requirements for `session`: not binding to connection, 
the session state (settings, vars) is not stored on databand-query,  so without loss of session we can achieve:

1. suspend in a few minutes
2. query can dispatch to diff server for HA and LB.

leiysky:
> apps like snowflake worksheet may have diff requirements for `session`: not binding to connection, the session state (settings, vars) is not stored on databand-query, so without loss of session we can achieve:
> 
> 1. suspend in a few minutes
> 2. query can dispatch to diff server for HA and LB.

The persist mechanism will be discussed in this RFC. But the storage and load scheme of session won't be involved now.

BohuTANG:
Yes.
@zhang2014 used to try to unify them (http/clickhouse/mysql).

BohuTANG:
Outdated.

PR #6672, opened 2022-07-18
fix: adhoc fix session leak

### LINKED ISSUE #6560: bug:  session leak
### Search before asking

### Version

b9e85825efcb6ef40419b466fe3ac36968b33d39

### What's Wrong?

object `Session` leaks

### How to Reproduce?

~~~
# enable memory-profiling
cargo b  -p databend-query --features memory-profiling
# fire up databend-query
MALLOC_CONF=prof:true ./target/debug/databend-query --log-level=error
~~~

keep creating new sessions and quit, and then differ the memory

~~~
curl http://localhost:8080/debug/mem/pprof/profile > base_profile
for i in {1..1000}; do echo "select 1; " | mysql -h 127.0.0.1  -P3307 -uroot -s  -N > /dev/null; done
curl http://localhost:8080/debug/mem/pprof/profile > after_inserts
jeprof ./target/debug/databend-query --base=base_profile after_inserts --pdf > session_leak.pdf
evince session_leak.pdf
~~~

and then, got something like this, which indicates memory leaks:

session_leak

[session_leak.pdf](https://github.com/datafuselabs/databend/files/9076471/session_leak.pdf)

ariesdevil:

## Summary

Fix session leak, to detect it, just wrap `Arc<Session>` to `SessionWrapper` as following, then check the new and drop are paired:
```rust
pub struct SessionWrapper {
    id: usize,
    session: Arc<Session>,
}

impl SessionWrapper {
    pub fn new(s: Arc<Session>) -> Self {
        let backtrace = Backtrace::force_capture();
        let backtrace = format!("{:?}", backtrace);
        let id = GlobalSequence::next();
        println!(
            "session wrapper {} new: {}-{}",
            id,
            Arc::strong_count(&s),
            backtrace
        );
        Self { id, session: s }
    }
}

impl Clone for SessionWrapper {
    fn clone(&self) -> Self {
        Self::new(self.session.clone())
    }
}

impl Drop for SessionWrapper {
    fn drop(&mut self) {
        let backtrace = Backtrace::force_capture();
        let backtrace = format!("{:?}", backtrace);
        println!(
            "session wrapper {} dropped: {}-{}",
            self.id,
            Arc::strong_count(&self.session),
            backtrace
        );
    }
}

impl Deref for SessionWrapper {
    type Target = Arc<Session>;

    fn deref(&self) -> &Self::Target {
        &(self.session)
    }
}
```

**NOTE**: this is just an ad-hoc fix, the refactoring of session module will be done related to [this issue](https://github.com/datafuselabs/databend/issues/5679)

Fixes #6560

### DIFF
--- a/query/src/servers/clickhouse/interactive_worker.rs
+++ b/query/src/servers/clickhouse/interactive_worker.rs
@@ -50,12 +50,10 @@ impl ClickHouseSession for InteractiveWorker {
 
         let session = self.session.clone();
         let get_query_result = InteractiveWorkerBase::do_query(ctx, session);
-        let query_ctx = self
+        let format = self
             .session
-            .get_shared_query_context()
-            .await
+            .get_format_settings()
             .map_err(to_clickhouse_err)?;
-        let format = query_ctx.get_format_settings().map_err(to_clickhouse_err)?;
         if let Err(cause) = query_writer.write(get_query_result.await, &format).await {
             let new_error = cause.add_message(&ctx.state.query);
             return Err(to_clickhouse_err(new_error));
--- a/query/src/servers/mysql/mysql_interactive_worker.rs
+++ b/query/src/servers/mysql/mysql_interactive_worker.rs
@@ -187,11 +187,7 @@ impl<W: std::io::Write + Send + Sync> AsyncMysqlShim<W> for InteractiveWorker<W>
         let instant = Instant::now();
         let blocks = self.base.do_query(query).await;
 
-        let format = self
-            .session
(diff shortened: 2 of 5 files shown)


drmingdrmer (commented):
I'm not sure about the root cause of this leak, although I partially get involved in the tracing for this problem.

After all, these modification looks good to me:
Always using `SessionRef` and making the call stack shallow will makes the codebase more clear and easier to understand.

sundy-li (approved):

Xuanwo (approved):

Issue #6554, opened 2022-07-09
'1046: Cannot parse value:Ok("5632622125792883430") to number type, cause: Overflow(5)' on loading hits data


flaneur2020:
**Version**

0.7.121-nightly

**Summary**

Schema: 

```
[...]

Issue #6669, opened 2022-07-18
clickhouse parse error with unmatched quote in format csv/tsv


youngsofun:
**Summary**

the unmatched quoted string is valid in csv/tsv if the quote it not used as enclosing.

```
"a, b
```

``` insert into t1 format TSV "a, b```

but currently the sql (together with the format part) go through tokenizer(old parser)  first, and it raise error.

```
 Expected close delimiter '"' before EOF
```

can this be solved after parse v1 is dropped?
if true we can wait for clickhouse handler, and support this kind of csv in copy first 
if not planner v2 need to deal with it.

youngsofun:
@sundy-li @leiysky

leiysky:
cc @andylokandy

andylokandy:
We don't support dialect yet. It has been tracked as a low priority task. Are we going to adjust the priority?

sundy-li:
For insert query, we should pass a partial substring into the parser.

`insert into t1 format TSV "a, b`  -> `insert into t1 format TSV`

Issue #6650, opened 2022-07-15
TSV InputFormat should not use quote for enclosing


youngsofun:
**Summary**

Issue #6553, opened 2022-07-09
bug: "found record with 100 fields, but the previous record has 105 fields" on loading hits data


flaneur2020:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.


### Version

v0.7.121-nightly

### What's Wrong?

statement:

```
COPY INTO hits FROM 's3://repo.databend.rs/hits/hits.tsv' credentials=(aws_key_id='***' aws_secret_key='***') file_format=(type='CSV' field_delimiter='\\t' record_delimiter='\\n' skip_header=1);"
```

> 2022/07/08 13:32:30 the 0 time result has error query has error: &{Code:1046 Message:Parse csv error at line 93557185, cause: CSV error: record 93557186 (line: 93557187, byte: 69916103044): found record with 100 fields, but the previous record has 105 fields (while in processor thread 3) Kind:}, stats:

but these scripts shows that the hits data is not corrupted:

```
╭─yazhou@yazhous-MBP ~/Downloads
╰─$ sed -n '93557185p' < hits.tsv | ruby -e "puts STDIN.read.split(/\t/).size"                                                 
105
╭─yazhou@yazhous-MBP ~/Downloads
╰─$ sed -n '93557184p' < hits.tsv | ruby -e "puts STDIN.read.split(/\t/).size"
105
╭─yazhou@yazhous-MBP ~/Downloads
╰─$ sed -n '93557186p' < hits.tsv | ruby -e "puts STDIN.read.split(/\t/).size"
105
╭─yazhou@yazhous-MBP ~/Downloads
╰─$ sed -n '93557187p' < hits.tsv | ruby -e "puts STDIN.read.split(/\t/).size"
105
╭─yazhou@yazhous-MBP ~/Downloads
╰─$ sed -n '93557188p' < hits.tsv | ruby -e "puts STDIN.read.split(/\t/).size"
105
```

```
╰─$ sed -n '93557185,93557189p' < hits.tsv                                                                                                                                                                                                                                
[...]

PR #6673, opened 2022-07-18
fix(query): input format TSV  should not use quote for enclosing

youngsofun:

## Summary

not work for clickhouse httphandler: https://github.com/datafuselabs/databend/issues/6669

Fixes  https://github.com/datafuselabs/databend/issues/6554
https://github.com/datafuselabs/databend/issues/6553
https://github.com/datafuselabs/databend/issues/6650
https://github.com/datafuselabs/databend/issues/6650


tests/suites/1_stateful/00_copy_v2/00_0000_copy_tsv.result
```diff
@@ -0,0 +1,10 @@
+---1
+"a	1	2
+"a"	2	3
+"a	3	2
+'a	4	2
+'a'	5	3
+'a	6	2
+'a"b'	7	5
+\\	8	1
+\\i	9	2
```
Xuanwo, line 10: Add a case that contains an empty space ` `? :

```
'a" b' 7 5
```

Xuanwo (approved):
Mostly LGTM, please make Clippy happy~

common/io/src/buffer/buffer_read_ext.rs
```diff
@@ -122,11 +122,20 @@ pub trait BufferReadExt: BufferRead {
...
-            // TODO parse complex escape sequence
-            buf.push(b'\\');
-            self.read_escaped_string_text(buf)?;
+        loop {
+            self.keep_read(buf, |f| f != b'\t' && f != b'\n' && f != b'\\')?;
+            if self.ignore_byte(b'\\')? {
+                let buffer = self.fill_buf()?;
+                let c = buffer[0];
+                match c {
+                    b'\'' | b'\"' | b'\\' => {}
```
sundy-li, line 131: Some other cases in [clickhouse](https://github.com/ClickHouse/ClickHouse/blob/aa737b7a11167fa4af3bb8f8dc4e70909d1728e2/src/IO/ReadHelpers.cpp):


1. `\x2A` to read hex codes
2. `\N`  empty string, skipped
3. ...

youngsofun:
@Xuanwo  only copy from ontime_200.csv.zst fail,  is there anything special about zstd?

common/io/src/buffer/buffer_read_ext.rs
youngsofun: ok,  these are not as important (at least not fail the loading),
I will add them after the tests passed

youngsofun:
log error like this 

tests/suites/1_stateful/00_copy_v2/00_0000_copy_from_stage2.py


```
 1 | CREATE TABLE ontime200
   | ------ while parsing `CREATE TABLE [IF NOT EXISTS] [<database>.]<table> [<source>] [<table_options>]`
   .
35 |     DepDelayMinutes                 INT,
36 |     DepDel15                        INT,
37 |     DepartureDelayGroups            VARCHAR,
38 |     DepTimeBlk                      V    DepTimeBlk                              INT,
   |     ----------                      ^
   |     |                               |
   |     |                               expected `BOOLEAN`, `BOOL`, `UINT8`, `TINYINT`, `UINT16`, `SMALLINT`, or 25 more ...
   |     |                               while parsing type name
   |     while parsing `<column name> <type> [NOT NULL | NULL] [DEFAULT <default value>] [COMMENT '<comment>']`


````

**Comment by BohuTANG** [comment]
```
Diff in /workspace/query/src/pipelines/processors/sources/file_splitter.rs at line 99:
         rows_to_skip: &mut u64,
     ) -> Result<()> {
         if data.is_empty() {
-            return Ok(())
+            return Ok(());
         }
         let mut data_slice = data;
```

common/io/src/buffer/buffer_read_ext.rs
youngsofun: done

sundy-li (approved):

Issue #6596, opened 2022-07-13
slient the logs in the sqlogictest


flaneur2020:
**Summary**

currently the sqlogic test's log is pretty big, causes hard to found the error in the github action's UI like: https://github.com/datafuselabs/databend/runs/7313929749?check_suite_focus=true

in this case, we can easily found the root cause is 'CURRENT_TIMESTAMP' in the tail of the logs output, but the log is 10mb, causing the github action UI can not easily display the output.

this is the output of a sqlogic test:

[3_Run ..githubactionstest_sqllogic_standalone_linux.txt](https://github.com/datafuselabs/databend/files/9098946/3_Run.githubactionstest_sqllogic_standalone_linux.txt)

the stateless tests only displays the test files which get executed, and SUCCESS/FAILURE about each test file. but sqlogictest displays every output of every statement, thus makes the output pretty large.

we can consider only display success/failure result of each sqlogictest file, with the output of the failed case.

flaneur2020:
cc @ZeaLoVe

flaneur2020:
this is the log output of the stateless test:

![Screen Shot 2022-07-14 at 10 57 56 AM](https://user-images.githubusercontent.com/129800/178888492-8da36496-1d86-48eb-8ab1-13a81a57d641.png)

it's an aggregated result in the file manner, I think we can consider a similiar representation.

ZeaLoVe:
As the #6680 merged,  decrease half of the total logs in logic test ci.

PR #6680, opened 2022-07-19
chore: silence logic test logs

ZeaLoVe:

## Summary

Silence logic test logs, change statements log level to debug, issue #6596 

Fixes #issue

### DIFF
--- a/tests/logictest/logictest.py
+++ b/tests/logictest/logictest.py
@@ -98,8 +98,7 @@ class LogicError(Exception):
         self.expected = expected
 
     def __str__(self):
-        return "Expected regex{}, Actual: {}".format(self.expected,
-                                                     self.message)
+        return f"Expected regex{self.expected}, Actual: {self.message}"
 
 
 class Statement:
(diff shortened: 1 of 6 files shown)


tests/logictest/logictest.py
```diff
@@ -281,28 +281,26 @@ def execute(self):
         if callable(getattr(self, "batch_execute")):
             # case batch
             for (file_path, suite_name) in self.statement_files:
-                log.info("Batch execute, suite name:{} in file {}".format(
-                    suite_name, file_path))
+                log.info("Run query with the same session every suite file, suite file path:{}".format(file_path))
```
Xuanwo, line 284: As we are using python 3.10, how about using format strings for better reading, like:

```python
log.info(f"Run query with the same session every suite file, suite file path:{file_path}")
```

tests/logictest/logictest.py
```diff
@@ -281,28 +281,26 @@ def execute(self):
...
                 statement_list = list()
                 for state in get_statements(file_path, suite_name):
                     statement_list.append(state)
                 self.batch_execute(statement_list)
         else:
             # case one by one
             for (file_path, suite_name) in self.statement_files:
-                log.info("One by one execute, suite name:{} in file {}".format(
-                    suite_name, file_path))
+                log.info("Run query without session every statements, suite file path:{}".format(file_path))
```
Xuanwo, line 292: ditto

tests/logictest/logictest.py
Xuanwo: BTW, it's better to add a space after `:`, which will make it easier for reading.

tests/logictest/logictest.py
ZeaLoVe: > ditto

PTAL

Xuanwo (approved):

BohuTANG (approved):

PR #6685, opened 2022-07-19
fix(query): remove incorrect debug_assert

### LINKED ISSUE #6659: bug: Not(expression) can't return result in new planner
### Search before asking

### Version

main

### What's Wrong?

Consider the case:

```sql
mysql> create table t3(a bool, b int);

mysql> insert into t3 values(false, 1), (true, 2);

mysql> select not(t3.a) from t3 where t3.b > 2;

mysql> select not(t3.a) from t3 where t3.b > 2;
ERROR 1105 (HY000): Code: 1104, displayText = assertion failed: !column.is_empty().
```

### How to Reproduce?

_No response_

### LINKED ISSUE #6682: Bug: `assertion failed: !column.is_empty().` when using `CASE WHEN` on empty `DataBlock`
Only can be reproduced in debug build:
```sql
select case when number > 1 then 1 when number < 1 then 2 else 1 end from numbers(2) where false;
```

Which is caused by the `debug_assertion` [here](https://github.com/datafuselabs/databend/blob/a2cac3b10af29fd25fad65b749c46abe35392f1a/common/datavalues/src/scalars/viewer.rs#L128)

sundy-li:

## Summary

Summary about this PR

 remove incorrect debug_assert

Fixes #6659
Fixes #6682


xudong963 (approved):
LGTM!

PR #6686, opened 2022-07-19
refactor: use SyncUnsafeCell to replace AtomicRefCell

PsiACE:

## Summary

`AtomicRefCell` performs an atomic memory access at runtime to validate borrowing. While this is an excellent way to validate code and ensure safety, it is an expensive operation. 

ref: https://doc.rust-lang.org/nightly/std/cell/struct.SyncUnsafeCell.html

> UnsafeCell doesn’t implement Sync, to prevent accidental mis-use. You can use SyncUnsafeCell instead of UnsafeCell to allow it to be shared between threads, if that’s intentional. Providing proper synchronization is still the task of the user, making this type just as unsafe to use.

Also, I personally speculate that this may have a slight benefit in fixing stateless-cluster-test(release).

Fixes #issue


Xuanwo:
Rust Naming: `SafeUnsafeXxxx`....

Xuanwo (approved):
Although I can't understand it fully, it's worth a try.

zhang2014 (approved):

PR #6731, opened 2022-07-21
feat: construct leaf column statistics

### LINKED ISSUE #6728: Improvement(min/max index):  construct leaf column statistics
**Summary**

Construct **leaf column** statistics  (instead of **field**, which may be nested struct)

refactoring of PR #6612 is blocked by this issue.

dantengsky:

## Summary

- Collects column statistics from the leave columns.

for nested struct like:
~~~
       schema {
          a: struct {
              b: struct {
                 c: i64,
                 d: f64,
              },
              e: f64
          }
          f: i64,
          g: f64,
       }

~~~

statistics of columns c, d, e, f, and g will be collected, instead of columns a, f, g

- id of leaf columns
   schema is traversed in a depth-first manner, and only leave are collected. index of each element map to their ids:
   - impl: 
      https://github.com/datafuselabs/databend/pull/6731/files#diff-03cded00d3322a71cc9137c8fdb05fe7c3648ab7073ab4609a4751f44d4d3cbaR84
   - test case
     https://github.com/datafuselabs/databend/pull/6731/files#diff-c52587387e9c8dd75bfb95dc1fbc1d1e848001dcacf4f1a6cbdc5b3b89c79b0eR86

since nested data structure not enabled yet, there will be no backward compatibility issues.

Fixes #6728


b41sh:
LGTM

BohuTANG:
Can we add a stateless test for this patch?

dantengsky:
> Can we add a stateless test for this patch?

sure ;) but since table with nested structures  can not be created(and inserted) yet, seems the primary case left is to check min max index still works

BohuTANG (approved):

dantengsky:
@BohuTANG  merged? the stateless test case will soon be added in another PR

Issue #6719, opened 2022-07-20
`1.903583017e+9` notation insert


BohuTANG:
### Discussed in https://github.com/datafuselabs/databend/discussions/6717

<div type='discussions-op-text'>

<sup>Originally posted by **hanyisong** July 21, 2022</sup>
1. Does databend currently have an exponential form of writing? Will it be supported later?
```bash
mysql> desc t1;
+-------+--------+------+-------------------------------------------+-------+
| Field | Type   | Null | Default                                   | Extra |
+-------+--------+------+-------------------------------------------+-------+
| c0    | DOUBLE | NO   | cast(1969-12-12 23:25:03 as Timestamp(6)) |       |
+-------+--------+------+-------------------------------------------+-------+
1 row in set (0.02 sec)
Read 0 rows, 0.00 B in 0.008 sec., 0 rows/sec., 0.00 B/sec.

mysql> INSERT INTO t1(c0) VALUES (1541870038), (1.903583017e+9);
ERROR 1105 (HY000): Code: 1005, displayText = sql parser error: Expected ), found: e.
mysql> INSERT INTO t1(c0) VALUES (1e9);
ERROR 1105 (HY000): Code: 1005, displayText = sql parser error: Expected ), found: e9.
```
2. Can't databend insert using DEFAULT
```bash
mysql> desc t1;
+-------+--------+------+-------------------------------------------+-------+
| Field | Type   | Null | Default                                   | Extra |
+-------+--------+------+-------------------------------------------+-------+
| c0    | DOUBLE | NO   | cast(1969-12-12 23:25:03 as Timestamp(6)) |       |
+-------+--------+------+-------------------------------------------+-------+
1 row in set (0.01 sec)
Read 0 rows, 0.00 B in 0.005 sec., 0 rows/sec., 0.00 B/sec.

mysql> INSERT INTO t1(c0) VALUES (DEFAULT);
ERROR 1105 (HY000): Code: 1006, displayText = Unable to get field named "DEFAULT". Valid fields: ["c0"].
```</div>

BohuTANG:
cc @sundy-li

PR #6752, opened 2022-07-22
feat(query): Enable to parse decimal notation as floats

sundy-li:

## Summary

Summary about this PR

Part of #6719


everpcpc:
maybe the upper `E` should also be supported?

BohuTANG (approved):

PR #6757, opened 2022-07-22
ci: migrate stateless test 00,01 to logic test

ZeaLoVe:

## Summary

Migration of stateless test to logic test, 00_dummy and 01_system. Can be track  in https://github.com/datafuselabs/databend/discussions/5838

Fixes #issue


tests/logictest/suites/gen/00_dummy/00_0000_dummy_select_1
```diff
@@ -16,7 +16,7 @@ a
 statement query B label(http)
 SELECT NOT(1=1);
 
----- 
+----
```
sundy-li, line 19: why do we need these empty spaces?

ZeaLoVe: It cause by gen_suites.py scripts, three handler a little diferent with two handler when parsing result label.

sundy-li: Ok, but my editor will always trim the trailing whitespace on save (most vscode users will have this settings by default).  I suggest to trim the spaces if label is empty.

sundy-li (approved):

BohuTANG (approved):

Issue #6766, opened 2022-07-22
Migrate `String` functions to new expression framework


andylokandy:
# Tasks

Legacy string functions are settle in `common/functions/src/scalars/strings`. The task is to migrate all of them to `common/functions-v2/src/scalars/string.rs`. Similarly, the legacy tests in `common/functions/tests/it/scalars/strings` should also be migrated to `common/functions-v2/tests/it/scalars/string.rs`.

- [x] #6770
- [x] #6771
- [x] `rtrim` https://github.com/datafuselabs/databend/pull/6921
- [x] `trim` https://github.com/datafuselabs/databend/pull/6921
- [x] `trim_leading` https://github.com/datafuselabs/databend/pull/6921
- [x] `trim_trailing` https://github.com/datafuselabs/databend/pull/6921
- [x] `trim_both` https://github.com/datafuselabs/databend/pull/6921
- [x] `ltrim` https://github.com/datafuselabs/databend/pull/6921
- [x] `quote` https://github.com/datafuselabs/databend/pull/6907
- [x]  `lower` https://github.com/datafuselabs/databend/pull/6775
- [x] `lcase` https://github.com/datafuselabs/databend/pull/6775
- [x] `upper` #6765
- [x] `ucase` #6765
- [x] `reverse` https://github.com/datafuselabs/databend/pull/6907
- [x] `soundex` https://github.com/datafuselabs/databend/pull/7539
- [x] `ascii` https://github.com/datafuselabs/databend/pull/6907
- [x] #6776
- [x] #6777
- [x] #6778
- [x] #6779
- [x] `ord` https://github.com/datafuselabs/databend/pull/7539
- [x] `regexp_instr` https://github.com/datafuselabs/databend/pull/7459
- [x] `regexp_like` https://github.com/datafuselabs/databend/pull/7459
- [x] `regexp_replace` https://github.com/datafuselabs/databend/pull/7459
- [x] `regexp_substr` https://github.com/datafuselabs/databend/pull/7459
- [x] `bin` https://github.com/datafuselabs/databend/pull/7219
- [x] `oct` https://github.com/datafuselabs/databend/pull/7219
- [x] `hex` https://github.com/datafuselabs/databend/pull/7219
- [x] `unhex` https://github.com/datafuselabs/databend/pull/7219
- [x] `repeat` https://github.com/datafuselabs/databend/pull/7571
- [x] `substring` https://github.com/datafuselabs/databend/pull/7583
[...]

PR #6782, opened 2022-07-23
feat(functions):  add vectorize_string_to_string in functions-v2

### LINKED ISSUE #6770: `to_base64`

### LINKED ISSUE #6771: `from_base64`

ygf11:

## Summary
`String2String` function is special, and have many implementers.
For performance reason, we use `StringColumn::try_transform` to `vectorize` functions, so I try to add ~~`register_string_2_string`~~ to do special `vectorize` in new functions.

changes:
1. ~~Add `register_string_2_string`, so others can add new string2string function easily.~~
   Add `vectorize_string_to_string` and `vectorize_string_to_string_wise` to make others to add string2string functions easily.
3. Add `from_base64` and `to_base64` in new string function level.
4. Add unit tests. 

Fixes #6770 
Fixes #6771 
Part of #6766


sundy-li:
`register_string_2_string ` can reduce extra index checks and allocations, it's deserved to have!

common/expression/src/function.rs
```diff
@@ -765,3 +815,62 @@ pub fn vectorize_with_writer_passthrough_nullable_2_arg<I1: ArgType, I2: ArgType
...
+    ValueRef<NullableType<StringType>>,
+    &GenericMap,
+) -> Result<Value<NullableType<StringType>>, String>
++ Copy {
+    move |arg1, _| match arg1 {
+        ValueRef::Scalar(None) => Ok(Value::Scalar(None)),
+        ValueRef::Scalar(Some(val)) => {
+            let offsets = vec![0, val.len() as u64];
+            let data_len = estimate_bytes_fn((val, &offsets.into()));
+            let builder =
```
sundy-li, line 859: Scalar should be mapped to `Scalar` result.

ygf11: Yes, it will return Scalar result. The offsets is only used to get `estimate_bytes`.

I do not want to match `ValueRef` in `estimate_bytes_fn`, so all `scalars` will cast to `one item` column, then pass to the `estimate_bytes_fn`. This can make `estimate_bytes_fn` simple.

The `estimate_bytes_fn` is like:
```rust
Fn((&[u8], &Buffer<u64>)) -> usize + Copy,
```

common/expression/src/types/string.rs
```diff
@@ -227,4 +227,71 @@ impl StringColumnBuilder {
...
+    {
+        let mut buf = Vec::with_capacity(estimate_bytes);
+        unsafe {
+            let bytes = std::slice::from_raw_parts_mut(buf.as_mut_ptr(), buf.capacity());
+            let len = func(val, bytes)?;
+            buf.set_len(len);
+        }
+
+        Ok(StringColumnBuilder {
+            offsets: vec![0, buf.len() as u64],
```
sundy-li, line 254: So we do not need this offsets vec for scalar.

Just return `Ok(buf)`

ygf11: You are right, the offsets is no need here.
let me remove this.

andylokandy:
I'm thinking about whether a more general solution exists. 🤔

andylokandy:
In https://github.com/datafuselabs/databend/pull/6787, auto-nullable no longer force you to use auto-vectorize, thus we're allowed to define custom `vectorize_string_to_string` in `common-functions-v2`.

BohuTANG:
After https://github.com/datafuselabs/databend/pull/6787 merged, conflicting files :)

common/functions-v2/src/scalars/string.rs
```diff
@@ -77,22 +98,31 @@ pub fn register(registry: &mut FunctionRegistry) {
     registry.register_aliases("char_length", &["character_length"]);
 }
 
+// Vectorize string to string function with default estimate_bytes, which is the original string length.
 fn vectorize_string_to_string(
```
andylokandy, line 102: This can be achieved by `common_expression::function::vectorize_with_writer_1_arg`

common/functions-v2/src/scalars/string.rs
```diff
@@ -77,22 +98,31 @@ pub fn register(registry: &mut FunctionRegistry) {
...
-    func: impl Fn(&[u8], &mut StringColumnBuilder) -> Result<(), String> + Copy,
+    func: impl Fn(&[u8], &mut [u8]) -> Result<usize, String> + Copy,
+) -> impl Fn(ValueRef<StringType>, &GenericMap) -> Result<Value<StringType>, String> + Copy {
+    vectorize_string_to_string_wise(|(data, _)| data.len(), func)
+}
+
+// Vectorize string to string function with customer estimate_bytes.
+fn vectorize_string_to_string_wise(
+    estimate_bytes: impl Fn((&[u8], &Buffer<u64>)) -> usize + Copy,
+    func: impl Fn(&[u8], &mut [u8]) -> Result<usize, String> + Copy,
```
andylokandy, line 111: Would it be better if we use `StringColumn` here?

ygf11: I want the `Builder` of `Scalar` also create with `estimate_bytes`, so I treat `scalar` as `a one item column`.

If we use `&StringColumn` as the arg of `estimate_bytes`, it has to clone the data to transfer `&[u8]` to `StringColumn`.

So I use `(&[u8], &Buffer<u64>)` to avoid data clone, which is equivalent to `&StringColumn`.

Maybe we can define it as a struct like  `StringColumnRef` for readability.

common/functions-v2/src/scalars/string.rs
andylokandy: Oh sorry, I misunderstood it. But also, it may be not necessary to have both `vectorize_string_to_string` and `vectorize_string_to_string_wise`, a single `vectorize_string_to_string_wise` is enough to handle the case of same length output.

ygf11: Let's simple use `vectorize_string_to_string_wise`, and rename it to `vectorize_string_to_string`.

common/expression/src/types/string.rs
```diff
@@ -258,4 +258,61 @@ impl StringColumnBuilder {
...
+    val: &[u8],
+    estimate_bytes: usize,
+    mut func: F,
+) -> Result<Vec<u8>, String>
+where
+    F: FnMut(&[u8], &mut [u8]) -> Result<usize, String>,
+{
+    let mut buf = Vec::with_capacity(estimate_bytes);
+    unsafe {
+        let bytes = std::slice::from_raw_parts_mut(buf.as_mut_ptr(), buf.capacity());
```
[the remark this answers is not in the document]
ygf11: As the doc of `std::slice::from_raw_parts_mut` say:
```
/// * `data` must be [valid] for both reads and writes for `len * mem::size_of::<T>()` many bytes,
```
It may result in UB here.

Do we need use `vec![0u8; estimate_bytes]` here?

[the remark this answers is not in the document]
sundy-li: How about using:

`let mut buf = vec![0; estimate_bytes] `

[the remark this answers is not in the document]
ygf11: As the doc of `std::slice::from_raw_parts_mut` says:
```
/// * `data` must be [valid] for both reads and writes for `len * mem::size_of::<T>()` many bytes,
```
It may result in UB here.

can we use `vec![0u8; estimate_bytes]`?

common/functions-v2/src/scalars/string.rs
andylokandy: In fact, `Buffer<T>` is just a smart point to the underlying data, which means the clone is nearly zero-cost.

Because functions for Scalar are usually used in constant folding, I think it doesn't worth optimizing the allocation for Scalars.

common/expression/src/types/string.rs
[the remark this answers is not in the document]
andylokandy: I'd like not to optimize allocation for String scalar. If we reach an agreement, we can remove this function totally.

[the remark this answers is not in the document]
ygf11: It is ok to me to remove this optimize for String scalar.But we still need to resolve UB for String Column.

https://github.com/ygf11/databend/blob/ea69d75871aa30624ed24d2d60bc6b9bdcc7394d/common/expression/src/types/string.rs#L270-L279

common/functions-v2/src/scalars/string.rs
ygf11: Let's remove scalar optimize and use `StringColumn` here.

ygf11: Ohh, I found the core issue here.

Since both scalar and column branches use the same func in `vectorize`, we also need to get mutable slice in `scalar` branch, then we are forced to `estimate_bytes`.

andylokandy: We could add a method to `StringColumnBuilder` like:


```rust
pub fn alloc_slice(&mut self, len: usize) -> &mut [u8] { ... }
```

andylokandy: I suggest using [base64-stream]( https://github.com/magiclen/base64-stream) instead

ygf11: What is the difference between these two?

andylokandy: `base64-stream` can output encode result to `&mut Vec<u8>`

andylokandy: Just realized thaT `base64-stream` doesn't encode with padding...

ygf11: > Just realized that base64-stream doesn't encode with padding...

Do you mean we can change the write to `&mut Vec<u8>`, then `std::slice::from_raw_parts_mut` is no needed any more?

```rust
fn vectorize_string_to_string(
    estimate_bytes: impl Fn((&[u8], Buffer<u64>)) -> usize + Copy,
    func: impl Fn(&[u8], &mut Vec<u8>) -> Result<usize, String> + Copy,
) -> impl Fn(ValueRef<StringType>, &GenericMap) -> Result<Value<StringType>, String> + Copy {
```

sundy-li: Seems it's more reasonable.

andylokandy: I [took a try at it](https://github.com/ygf11/databend/pull/1), and it seems to work.

common/functions-v2/Cargo.toml
```diff
@@ -11,9 +11,11 @@ doctest = false
 
 [dependencies] # In alphabetical order
 # Workspace dependencies
+common-arrow = { path = "../arrow" }
```
andylokandy, line 14: This seems unused

andylokandy (approved):
rest lgtm

sundy-li (approved):

common/datavalues/src/columns/string/transform.rs
```diff
@@ -27,7 +27,7 @@ impl StringColumn {
     where
         F: FnMut(&[u8], &mut [u8]) -> Result<usize>,
     {
-        let mut values: Vec<u8> = Vec::with_capacity(estimate_bytes);
+        let mut values: Vec<u8> = vec![0u8; estimate_bytes];
```
ygf11, line 30: I found many places use `std::slice::from_raw_parts_mut`.

Do I need reset it, and then other pr will fixed all togathger? @sundy-li

common/functions-v2/Cargo.toml
ygf11: Removed.

common/datavalues/src/columns/string/transform.rs
sundy-li: > Do I need reset it, and then other pr will fixed all together?

No need, we ensure the memory is initialized manually. And the old function will be removed after we have functions-v2.

PR #6785, opened 2022-07-24
docs: Add cmake deps to building from source

aseaday:

## Summary

Cmake is needed when building from source


Xuanwo (approved):
Thanks for your contribution!

PR #6787, opened 2022-07-24
refactor(expr): make auto-nullable and auto-vectorization independent

andylokandy:

## Summary

Before this PR, auto-nullable always come with auto-vectorization, which makes it difficult to define a custom vectorization strategy while benefiting from auto-nullable.

Part of https://github.com/datafuselabs/databend/pull/6782
Tracked in https://github.com/datafuselabs/databend/issues/6766


common/expression/src/function.rs
```diff
@@ -698,70 +590,29 @@ pub fn vectorize_passthrough_nullable_2_arg<I1: ArgType, I2: ArgType, O: ArgType
...
+                .unwrap();
             Ok(Value::Column((col, arg1_validity)))
         }
         (ValueRef::Column((arg1, arg1_validity)), ValueRef::Column((arg2, arg2_validity))) => {
-            let iter = I1::iter_column(&arg1)
-                .zip(I2::iter_column(&arg2))
-                .map(|(arg1, arg2)| func(arg1, arg2));
-            let col = O::column_from_iter(iter, generics);
+            let col = func(ValueRef::Column(arg1), ValueRef::Column(arg2), generics)?
+                .into_column()
```
Xuanwo, line 612: If we always unwrap the value returned by into_column, how about expect the error inside into_column instead?

sundy-li (approved):

PsiACE (approved):

common/expression/src/function.rs
andylokandy: but into_columns is derived by EnumAsInner 🤔

PR #6789, opened 2022-07-25
fix(query): fix oom when returning large results in clickhouse tcp handler

### LINKED ISSUE #6562: OOM on executing SELECT * FROM ontime in a 8c16g instance
**Summary**

I'm running databend-query in a kubernetes deployment, executing `SELECT * FROM ontime` through the HTTP API may causes the pod terminated.

in my understanding, polling the result from databend-query http api is a streaming operation, do not necessary get all the output from the query into memory.

Tasks List:

- [x] make ClickHouse TCP handler support writing streaming results #6789 
- [x] make ClickHouse HTTP handler support writing streaming results  （already have）
- [x] make HTTP handler support writing streaming results （already have）
- [-]  make MySQL handler support writing streaming results （Can't have, it's MySQL protocol limited）

sundy-li:

## Summary

Summary about this PR

make ClickHouse TCP handler support writing streaming results

Fixes #6562


query/src/servers/clickhouse/interactive_worker_base.rs
```diff
@@ -100,9 +100,9 @@ impl InteractiveWorkerBase {
         insert: InsertPlan,
         ch_ctx: &mut CHContext,
         ctx: Arc<QueryContext>,
-    ) -> Result<UnboundedReceiver<BlockItem>> {
+    ) -> Result<Receiver<BlockItem>> {
         let sample_block = DataBlock::empty_with_schema(insert.schema());
-        let (sender, rec) = channel(4);
+        let (sender, rec) = channel(32);
```
zhang2014, line 105:
```suggestion
        let (sender, rec) = channel(2);
```

query/src/servers/clickhouse/interactive_worker_base.rs
```diff
@@ -129,7 +129,8 @@ impl InteractiveWorkerBase {
             .set_source_pipe_builder(Option::from(source_pipe_builder))
             .map_err(|e| tracing::error!("interpreter.set_source_pipe_builder.error: {:?}", e));
 
-        let (mut tx, rx) = mpsc::unbounded();
+        let (mut tx, rx) = mpsc::channel(32);
```
zhang2014, line 132:
```suggestion
        let (mut tx, rx) = mpsc::channel(2);
```

TCeason (approved):

flaneur2020 (approved):
can I understand this OOM as this?

1. the data_stream() is polled in a background task and feed the data block into an channel data_tx
2. the data_tx is an unbound channel, if the speed of background task feeding into data_tx is higher than the speed of polling from `data_rx`, the unbound channel would get bigger and bigger until OOM

sundy-li:
> if the speed of background task feeding into data_tx is higher than the speed of polling from data_rx, the unbound channel would get bigger and bigger until OOM

Right!

query/src/servers/clickhouse/interactive_worker_base.rs
```diff
@@ -164,7 +165,7 @@ impl InteractiveWorkerBase {
         let cancel = Arc::new(AtomicBool::new(false));
         let cancel_clone = cancel.clone();
 
-        let (tx, rx) = mpsc::unbounded();
+        let (tx, rx) = mpsc::channel(32);
```
zhang2014, line 168:
```suggestion
        let (tx, rx) = mpsc::channel(2);
```

zhang2014 (approved):

Issue #4521, opened 2022-03-21
Cache: Query result cache


BohuTANG:
**Summary**

## Background

For a read, the main flow is:
1. Get the source plan: file name(partition file)
2. Read the files by file name  which on object storage(like AWS S3).

With the query result cache, we can do:
* Step1. Parse the query, and calculate the fingerprint: `query_id`
* Step2. Get the source plan(`read_plan`), and calculate the fingerprint: `source_plan_id`
* Step3. Check the cache
   * 3.1 If the cache is exists: `/query_id/source_plan_id/result`, get and return the result.
   * 3.2 If the cache is not exists, put the result to the cache

## Where the cache stored

Storage in the S3, path is `/<bucket>/<tenant>/result/cache/`,  and the user can download it.

## How to calculate the fingerprint

* `query_id` need based on the AST? `select * from t1 where a>1`  fingerprint is same `select * from t1 where a>1 and 1=1`
* `source_plan_id` based on the partition file name and the file offset

flaneur2020:
cc @Chasen-Zhang this may help the issues we talked yesterday

BohuTANG:
cc @drmingdrmer

drmingdrmer:
The requirement for query result cache and the requirement for data block cache is different:

- The result cache must be complete, i.e., caching only part of the result is not allowed. The entire result data is added and removed as a whole.
- The data block cache prefers partial cache: not used data should not be cached.

AFAIK, these two goals may conflict with each other:
- The result cache may be evicted due to too many block caches. 
- The result cache would be better using a least-recently-added eviction policy(a result is kept for a certain time no matter how often it is read). While block cache would be better using a least-recently-used policy.

Xuanwo:
This issue have been moved to v0.9

BohuTANG:
The result cache is finished but not used yet, let's close.

PR #6799, opened 2022-07-25
feat(RFC): Cache

Xuanwo:

## Summary

This PR will add a new RFC in databend: Cache.

[Rendered RFC](https://databend-git-fork-xuanwo-rfc-cache-databend.vercel.app/doc/contributing/rfcs/cache)


docs/doc/60-contributing/03-rfcs/20220725-cache.md
```diff
@@ -0,0 +1,151 @@
...
+endpoint = "192.168.1.2"
+password = "PASSWORD"
+```
+
+Temporary data could be shared crossing the whole cluster, so users must use distributed cache services instead of
+local fs or memory:
+
+```toml
+[temporary]
+type = "opencache"
```
BohuTANG, line 78: opencache is not for temporary data, I think it's a kind of cache server such as Redis and can be set the TTL?

drmingdrmer: > opencache is not for temporary data, I think it's a kind of cache server such as Redis and can be set the TTL?

Right... As I commented in: https://github.com/datafuselabs/databend/issues/4521#issuecomment-1193602317

- Temporary store has to be durable and persistent. The eviction policy is least-recently-added.
- Block cache is as you described in RFC, **volatile**. And the eviction policy is more like to be least-recently-used.

But this is not a big difference in implementation: 
- Temporary store requires `fsync` on write and adopts least-recently-added eviction policy.
- Block cache does not require `fsync` on write and adopts a least-recently-used eviction policy.

Another issue with the temporary store is that the hash method on the client side must be consistent on every query node. Otherwise, a query node can not see the temp data written by another query node with a different hash configuration.

Xuanwo: Thanks for the explanation, I will add it to RFC.

Xuanwo: Addressed, PTAL.

flaneur2020 (approved):

drmingdrmer:
Another issue is client-side config versioning:
If a new cache node `C` is added to the cache cluster `{A,B}`, and assuming the cache client uses a consistent hash, when the new cluster config `{A,B,C}` is published to every client, 1/3 of the data access will encounter a miss.

To upgrade cluster config smoothly, a client may need to access the cache cluster using two configs: read from the new cluster first, if there is a miss, read the old cluster, until data migration is done. During this period, there are two active config `{{A,B},ver=1}` and `{{A,B,C},ver=2}`.

And it is also possible that there are more than two active configs.

Thus for the cache client, it has to:
- Be able to upgrade cluster config on the fly,
- support a **joint** config,
- and be able to retry every config.

A cluster upgrade may looks like this:
- Initially, every cache-client has config `[{{A,B},ver=1}]`;
- A new cache node is added, and the new config is published to every client: `[{{A,B},ver=1}, {{A,B,C},ver=2}]`;
- After a while, config with ver=1 is removed, and another new config is published to every client: `[{{A,B,C},ver=2}]`.

ClSlaid (approved):

Xuanwo:
> Another issue is client-side config versioning:

OpenCache's client implementation detail will be covered by OpenDAL. I create an issue for this concern: https://github.com/datafuselabs/opendal/issues/466

drmingdrmer (approved):

BohuTANG (approved):

Xuanwo:
cc @dantengsky @zhang2014 for reviewing.

BohuTANG:
Let's merge first, the RFC review is still open along the development.

PR #6825, opened 2022-07-26
feat(query): support CHAR keyword

### LINKED ISSUE #6822: Feature: support char like varchar
**Summary**

```sql
'user1'@mysqldb 18:54:28 [(none)]> create table t(id int not null, name char(120) not null default '');
  --> SQL:1:38
  |
1 | create table t(id int not null, name char(120) not null default '')
  | ------                          ---- ^^^^
  | |                               |    |
  | |                               |    expected `BOOLEAN`, `BOOL`, `UINT8`, `TINYINT`, `UINT16`, `SMALLINT`, or 25 more ...
  | |                               |    while parsing type name
  | |                               while parsing `<column name> <type> [DEFAULT <default value>] [COMMENT
'user1'@mysqldb 18:56:16 [(none)]> 

```

```sql
'user1'@mysqldb 18:54:19 [(none)]> create table t(id int not null, name varchar(120) not null default '');

'user1'@mysqldb 18:54:25 [(none)]> show create table t;
+-------+--------------------------------------------------------------------------+
| Table | Create Table                                                             |
+-------+--------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `id` INT,
  `name` VARCHAR DEFAULT ''
) ENGINE=FUSE |
+-------+--------------------------------------------------------------------------+
1 row in set (0.05 sec)
Read 0 rows, 0.00 B in 0.007 sec., 0 rows/sec., 0.00 B/sec.


```

TCeason:

## Summary

Support keyword char.
```sql
create table t(id int not null, name char(120) not null default '');
```
Fixes #6822


sundy-li (approved):

andylokandy (approved):

common/ast/src/parser/token.rs
```diff
@@ -878,7 +880,7 @@ impl TokenKind {
             | TokenKind::WHEN => true,
             | TokenKind::ARRAY
             | TokenKind::AS
-            // | TokenKind::CHAR
+            | TokenKind::CHAR => false,
```
TCeason, line 883: we have already a function named char(), so that return false in there.@[andylokandy](https://github.com/andylokandy)

common/ast/tests/it/testdata/expr.txt
```diff
@@ -55,7 +55,7 @@ char(208, 191, 209)
 ---------- AST ------------
 FunctionCall {
     span: [
-        Ident(0..4),
+        CHAR(0..4),
```
TCeason, line 58: I think this diff is normal. @[andylokandy](https://github.com/andylokandy)

common/ast/src/parser/token.rs
andylokandy: we can simply remove this line. Or else `TokenKind::{ARRAY, AS}` will mistakenly returns `false`.

andylokandy: @TCeason this is a bug

TCeason: > @TCeason this is a bug

You mean this will cause apply and as return false?

Issue #6634, opened 2022-07-14
Support JSON in new expression framework


andylokandy:
## Tasks

1. Add `DataType::Variant`.

The parsing and display rule:

```
"JSON" => DataType::Variant
"VARIANT" => DataType::Variant

DataType::Variant => "VARIANT"
```

2. Rename generic type `AnyType` to `VariantType`, and implement `ArgType` for it:

```
ArgType::Scalar = common_expression::values::Scalar
ArgType::ScalarRef<'a> = common_expression::values::ScalarRef<'a>
ArgType::Column = Arc<[Scalar]>
ArgType::Domain = ()  // upcasts to `Domain::Undefined`
ArgType::ColumnBuilder = Vec<Scalar>
ArgType::ColumnIterator<'a> = std::vec::Iter<'a, Scalar>
```

3. Add conversion between `Scalar` and `serde_json::Value`. And then store `Variant` columns in arrow2's `BinaryArray<i64>` which stores serialized JSON strings.

4. Add `can_auto_cast_to` rules:

```rust
DataType::_ -> DataType::Variant
```

5. Add CAST rule:
```rust
DataType::_ -> DataType::Variant
```

and TRY_CAST rule:

```rust
// Must be successful
DataType::_ -> DataType::Nullable(DataType::Variant)

// Extract exact variant from JSON value, convert to NULL if type mismatches.
DataType::Variant -> DataType::Nullable(_)
```

jiaoew1991:
@andylokandy hi, I am here again. I will take this issue

jiaoew1991:
json is not like number type, its structure and expression is quit complicated, could you provide more information about this issue, like how to display it, the relationship with string type, does arrow support it , etc.

BohuTANG:
@b41sh is the original author of the JSON type, ping for more help :)

andylokandy:
@jiaoew1991 Glad to see you again! I've added some instructions to the issue, please let me know if anything is not clear. 😉

b41sh:
some advice @andylokandy @jiaoew1991 
It's better to use `DataType::Variant` and `DataType::VariantObject`, with `JSON` and `Object` as an alias, because `Variant` is more general.
We need to define a struct `VariantValue`, as a wrapper `serde_json::Value`, for two reasons:
1. `serde_json::Value` does not implement the trait `Ord` and `PartialOrd`, which is needed when sorting.
2. the performance of `serde_json::Value` is not good, we may replace it with other formats in the future.

[...]

PR #6838, opened 2022-07-26
feat(expr): add `DataType::Map(T)`

andylokandy:

## Summary

- Add a static typed `Map(T)` which is an alias of `Array(Tuple(String, T))`.


sundy-li:
`DataType::Map<T>` is different to `DataType::Variant` in the datavalues.

b41sh:
> `DataType::Map<T>` is different to `DataType::Variant` in the datavalues.

We can use `Map<T>` instead of `VariantObject`, just like `Array<T>` instead of `VariantArray`. Because both `VariantArray` and `VariantObject` are actually a special type of `Variant`, keeping too many types at the same time can lead to confusion for users.

sundy-li:
> We can use Map<T> instead of VariantObject

`Map<T>` is similar to `Map<K,V>` of ClickHouse , https://clickhouse.com/docs/en/sql-reference/data-types/map.

But its key is Fixed to String type, it can not be used instead of `VariantObject` because `T` is a fixed type but `VariantObject`  is a variant type.

andylokandy:
@sundy-li 

> Map<T> is similar to Map<K,V> of ClickHouse , https://clickhouse.com/docs/en/sql-reference/data-types/map.

We can free the `K` from the String type if we want to.

> But its key is Fixed to String type, it can not be used instead of VariantObject because T is a fixed type but VariantObject is a variant type.

`VariantMap` is semantically equal to `Map<Variant>`.

> I don't think it's ok to have Map<T> because we have to care about the behavior of duplicated keys inside the tuple.

The uniqueness of keys is invariant and should be checked on constructing a map.

sundy-li:
> We can free the K from the String type if we want to.

Yes, better to have K instead of String.

I was thinking that whether should we have a new datatype named `Map`.  Since we will have `Variant` type which could cover:  

1. `1`
2. `'1'`
3.  `{"x" : "y"}`
4. `{"x" :   {"a" : 3 }    }`
5. ..

andylokandy:
> Yes, better to have K instead of String.

I'm worried about whether the user will use other types as keys. cc @BohuTANG 

> I was thinking that whether should we have a new datatype named Map. Since we will have Variant type which could cover:

Once we have static `Map`, the `Map` will become one of the variants of `Variant`. See more in https://github.com/datafuselabs/databend/issues/6634#issuecomment-1195768310

sundy-li:
Oh, I get it now.

sundy-li (approved):

BohuTANG (approved):

PR #6841, opened 2022-07-27
fix: mysql return 'Empty Set' when result set is empty

gaoxinge:

## Summary

Fix mysql server not return 'Empty Set' when result set is empty:

```sql
mysql> create database test_db;

mysql> create table t3(a bool, b int);

mysql> insert into t3 values (false, 1), (true, 2);

mysql> set enable_planner_v2 = 0;

---------- not return empty set -----------
mysql> select * from t3 where t3.b > 2; 

mysql> select * from t3 where t3.b > 2;
Empty set (0.03 sec)
Read 2 rows, 9.00 B in 0.008 sec., 257.51 rows/sec., 1.13 KiB/sec.
```


query/src/servers/mysql/writers/query_result_writer.rs
```diff
@@ -40,12 +40,14 @@ impl<'a, W: std::io::Write> DFQueryResultWriter<'a, W> {
 
     pub fn write(
         &mut self,
-        query_result: Result<(Vec<DataBlock>, String)>,
+        query_result: Result<(Vec<DataBlock>, String, bool, DataSchemaRef)>,
```
leiysky, line 43: It's too complicated to use tuple for this, maybe we should introduce a struct type here?

query/src/servers/mysql/mysql_interactive_worker.rs
```diff
@@ -48,9 +50,34 @@ use crate::sessions::QueryContext;
 use crate::sessions::SessionRef;
 use crate::sessions::TableContext;
 use crate::sql::DfParser;
+use crate::sql::DfStatement;
 use crate::sql::PlanParser;
 use crate::sql::Planner;
 
+fn is_result_set(stmt: &DfStatement) -> bool {
```
sundy-li, line 57: We will remove old statements in the future, I think we should match new statements/planners to have `is_result_set`

sundy-li (approved):

Xuanwo:
test_stateful_standalone_linux has been failed.

BohuTANG:
There are many tests failure:
https://github.com/datafuselabs/databend/runs/7577332808?check_suite_focus=true

query/src/servers/mysql/mysql_interactive_worker.rs
```diff
@@ -339,7 +394,12 @@ impl<W: std::io::Write> InteractiveWorkerBase<W> {
                                         e.code()
                                     )));
                                 }
-                                Ok((vec![DataBlock::empty()], String::from("")))
+                                Ok(QueryResult::create(
```
Xuanwo, line 397: How about implementing `Default` for `QueryResult`?

query/src/servers/mysql/writers/query_result_writer.rs
```diff
@@ -29,6 +29,29 @@ use common_io::prelude::FormatSettings;
...
+    extra_info: String,
+    is_result_set: bool,
+    schema: DataSchemaRef,
+}
+
+impl QueryResult {
+    pub fn create(
+        blocks: Vec<DataBlock>,
+        extra_info: String,
+        is_result_set: bool,
```
Xuanwo, line 43: How about renaming it to `has_result_set`?

query/src/interpreters/interpreter_call.rs
```diff
@@ -41,6 +48,10 @@ impl Interpreter for CallInterpreter {
         "CallInterpreter"
     }
 
+    fn schema(&self) -> DataSchemaRef {
+        self.schema.read().unwrap().as_ref().unwrap().clone()
```
Xuanwo, line 52: Use `expect` instead of `unwrap` for better reading.

gaoxinge:
@Xuanwo @sundy-li @leiysky @BohuTANG 

I have fixed cases to pass test and optimize code. Please help to review. Thx.

Xuanwo (approved):

Xuanwo:
Thanks for your contributions. This PR is great!

PR #6847, opened 2022-07-27
feat: Add default role, force ALTER/CREATE USER enable planner v2

### LINKED ISSUE #6843: RBAC: add default role to user
**Summary**

an user can set up a default role with the statement:

```
ALTER USER ... WITH DEFAULT_ROLE = '<role>';
```

also the `default role` can be viewed in the `SHOW USERS` statement.

on setting a default role, the role MUST be one of the roles the user has been granted.

on an user connecting to databend-query, she'd get set the default role to the current role in the session (with a check about whether the user still has the role).

on revoking an role from the user, if the role is set as default role, the default role would be reset to None.

### LINKED ISSUE #6896: New planner doesn't support ALTER USER with TENANTSETTING IDENTIFIED xx
Query:

```sql
ALTER USER 'name'@'hostname' WITH TENANTSETTING IDENTIFIED BY 'new_password'
```

```sql
Error: Code: 1005, displayText = error:
  --> SQL:1:50
  |
1 | ALTER USER 'test'@'localhost' WITH TENANTSETTING IDENTIFIED BY 'new_password'
  | -----                                            ^^^^^^^^^^ expected `TENANTSETTING`, `NOTENANTSETTING`, `CONFIGRELOAD`, or `NOCONFIGRELOAD`
  | |
  | while parsing `ALTER USER ('<username>'@'hostname' | USER()) [IDENTIFIED [WITH <auth_type>] [BY <password>]] [WITH <role_option> ...]`
```

### LINKED ISSUE #6897: New planner doesn't support ALTER USER 'test'@'localhost' WITH NOTENANTSETTING
```sql
ALTER USER 'test'@'localhost' WITH NOTENANTSETTING
```

```shell
Error: Code: 1005, displayText = error:
  --> SQL:1:51
  |
1 | ALTER USER 'test'@'localhost' WITH NOTENANTSETTING
  | -----                                             ^ expected `TENANTSETTING`, `NOTENANTSETTING`, `CONFIGRELOAD`, or `NOCONFIGRELOAD`
  | |
  | while parsing `ALTER USER ('<username>'@'hostname' | USER()) [IDENTIFIED [WITH <auth_type>] [BY <password>]] [WITH <role_option> ...]`
```

flaneur2020:

## Summary

- [x] add default_role into pb
- [x] refactor: remove User struct, which only used in a test case file
- [x] add default_role into UserOption
- [x] syntax about `alter user set default_role = `
- [ ] ~ensure only root can `ALTER USER xxx WITH flags`~ next pr
- [x] display default_role in SHOW USERS
- [x] force ALTER/CREATE USER take planner v2
- [x] add sqlogictest about ALTER USER
- [x] add sqlogictest about CREATE USER 
- [x] test about serialization on default_role 
- [ ] ~verify before set default role~ in next pr
- [x] fix test_load_old_user
- [x] await https://github.com/datafuselabs/databend/pull/6895 merged

Fixes #6843
Fixes #6897
Fixes #6896


flaneur2020:
@Xuanwo @junnplus can you help review this pull request?

common/meta/types/src/user_info.rs
```diff
@@ -94,17 +94,49 @@ impl TryFrom<Vec<u8>> for UserInfo {
...
+    pub fn with_set_flag(mut self, flag: UserOptionFlag) -> Self {
+        self.flags.insert(flag);
+        self
     }
 
     pub fn flags(&self) -> &BitFlags<UserOptionFlag> {
         &self.flags
     }
 
+    pub fn default_role(&self) -> &Option<String> {
```
Xuanwo, line 132:
```suggestion
    pub fn default_role(&self) -> Option<&str> {
```

common/meta/types/src/user_info.rs
```diff
@@ -113,6 +145,14 @@ impl UserOption {
         self.flags.insert(flag);
     }
 
+    pub fn switch_option_flag(&mut self, flag: UserOptionFlag, on: bool) {
```
Xuanwo, line 148: I don't like this API.

How about `set_option_flag(&mut self, flag: UserOptionFlag)`?

For remove, users can call `unset_option_flag` instead.

Or, better naming:

- insert_option_flag
- remove_option_flag

query/src/sql/parsers/parser_user.rs
```diff
@@ -37,59 +35,18 @@ use crate::sql::statements::DfGrantRoleStatement;
...
-
-        let create = DfCreateUser {
-            if_not_exists,
-            user: UserIdentity { username, hostname },
-            auth_option,
-            with_options,
-        };
-        Ok(DfStatement::CreateUser(create))
+        // skip to the planner v2 syntax
+        self.skip_all(DfStatement::CreateUser(DfCreateUser::default()))
```
Xuanwo, line 44: I prefer to convert `DfStatement::CreateUser` to a unit like `DfStatement::Copy` and remove `DfCreateUser` directly (so does `parse_create_user`).

So that we don't need `Default` on `DfCreateUser` anymore.

query/src/sql/planner/binder/ddl/account.rs
```diff
@@ -150,7 +150,7 @@ impl<'a> Binder {
             if_not_exists,
             user,
             auth_option,
-            role_options,
+            user_options: role_options,
```
Xuanwo, line 153: Please unify the naming here.

query/src/storages/system/users_table.rs
```diff
@@ -52,12 +52,22 @@ impl AsyncSystemTable for UsersTable {
             .iter()
             .map(|x| x.auth_info.get_auth_string())
             .collect();
+        let default_roles: Vec<String> = users
+            .iter()
+            .map(|x| {
+                x.option
+                    .default_role()
+                    .clone()
+                    .unwrap_or_else(|| "".to_string())
```
Xuanwo, line 61: unwrap_or_default is enough?

common/protos/proto/user.proto
```diff
@@ -96,6 +96,7 @@ message UserOption {
   uint64 min_compatible = 101;
 
   uint64 flags = 1;
+  optional string default_role = 2;
```
drmingdrmer, line 99: Increase the `VER` to `VER=3`:
https://github.com/datafuselabs/databend/blob/19a1c73b46b52e2ad628aebf42edacf3345a2c05/common/proto-conv/src/util.rs#L17

Add a test that builds a protobuf message with `VER=3`(this version) and check if it can be deserialized correctly. This test will guarantee the future databend-query works correctly with this `VER=3` data. This test should be kept until `MIN_COMPATIBLE_VER` becomes greater than 3.
https://github.com/datafuselabs/databend/blob/19a1c73b46b52e2ad628aebf42edacf3345a2c05/common/proto-conv/tests/it/user_proto_conv.rs#L181

BohuTANG:
Conflicting files
tests/logictest/logictest.py

common/ast/src/ast/statements/user.rs
```diff
@@ -233,13 +227,14 @@ impl Display for AccountMgrSource {
...
 
-impl Display for RoleOption {
+impl Display for UserOptionItem {
     fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
         match self {
-            RoleOption::TenantSetting => write!(f, "TENANTSETTING"),
-            RoleOption::NoTenantSetting => write!(f, "NOTENANTSETTING"),
-            RoleOption::ConfigReload => write!(f, "CONFIGRELOAD"),
-            RoleOption::NoConfigReload => write!(f, "NOCONFIGRELOAD"),
+            UserOptionItem::TenantSetting(true) => write!(f, "TENANTSETTING"),
```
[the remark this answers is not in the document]
flaneur2020: aggree, this option is rearly used among users, we can change it as describes.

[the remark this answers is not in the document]
flaneur2020: however I suggest we can make this rename in a seperate issue.

changing a name needs some careful considerations, this pr includes a small feature and some refactors, do not intend to change existed behaviour, I'd add an issue to track this renaming.

query/src/sql/parsers/parser_user.rs
Xuanwo: We can wait for https://github.com/datafuselabs/databend/pull/6895

So that we don't need those trick.

common/meta/types/src/user_info.rs
flaneur2020: thank you, fixed in https://github.com/datafuselabs/databend/pull/6847/commits/c9a8ac951ac36bfe54358f5620d31f1c12527cde

query/src/storages/system/users_table.rs
flaneur2020: oh, I did `unwrap_or_default` at first, but clippy suggest me to change into `unwrap_or_else` instead

common/meta/types/src/user_info.rs
flaneur2020: without this api, the code in `UserOptionItem` would from:

```
impl UserOptionItem {
    pub fn apply(&self, option: &mut UserOption) {
        match self {
            Self::TenantSetting(enabled) => {
                option.switch_option_flag(UserOptionFlag::TenantSetting, *enabled);
            }
            Self::ConfigReload(enabled) => {
                option.switch_option_flag(UserOptionFlag::ConfigReload, *enabled);
            }
            Self::DefaultRole(v) => option.set_default_role(Some(v.clone())),
        }
    }
}
```

to

```
impl UserOptionItem {
    pub fn apply(&self, option: &mut UserOption) {
        match self {
            Self::TenantSetting(enabled) => {
                if enabled {
                    option.set_option_flag(UserOptionFlag::TenantSetting);
                } else {
                    option.unset_option_flag(UserOptionFlag::TenantSetting);
                }
            }
            Self::ConfigReload(enabled) => {
                if enabled {
                    option.set_option_flag(UserOptionFlag::ConfigReload);
                } else {
                    option.unset_option_flag(UserOptionFlag::ConfigReload);
                }
            }
            Self::DefaultRole(v) => option.set_default_role(Some(v.clone())),
        }
    }
}
```

in my understanding these flag equals setting a boolean value to a struct.

introducing two apis may introduce some duplications in the code.

Xuanwo: Why do we need to store a `flag` in UserOption? Code could be must simple:

```rust
if enable_tenant_setting {
   option.tenant_setting = true
}
if enable_config_reload {
    option.config_reload = true
}
...
```

flaneur2020: like this，lemme hide the flag into internal

flaneur2020: oh, I failed on this refactor due to:

1. there're code directly depends on the flag in the privilege check case 
likes in https://github.com/datafuselabs/databend/blob/59ab2530da692de17cd50864f1843d51d85f2799/query/src/procedures/procedure_factory.rs#L42
2. the flags is persisted in metasrv in binary, can not easily change the storage about flags

this refactor may need more works than I expected, I guess I need a seperate PR to consider this :(

Xuanwo (approved):

common/protos/proto/user.proto
flaneur2020: https://github.com/datafuselabs/databend/pull/6847/files#diff-05bf2ee6ffcd5df0b1398460e2ab2d4034062d2a8333946e21691dd5f410e615R118

is this change correct as your describes? q.q

drmingdrmer (approved):

PR #6848, opened 2022-07-27
fix: prevents cpu reordering, remove unnecessary fence

ariesdevil:

## Summary

Prevents potential CPU reorder.

Fixes #issue


query/src/sessions/session_ref.rs
```diff
@@ -57,14 +56,15 @@ impl Drop for SessionRef {
...
+            std::sync::atomic::fence(Ordering::Acquire);
             tracing::debug!("Destroy session {}", self.id);
             self.session_mgr.destroy_session(&self.id);
             self.quit();
         }
     }
 
     pub fn increment_ref_count(self: &Arc<Self>) {
+        std::sync::atomic::fence(Ordering::Release);
         self.ref_count.fetch_add(1, Ordering::Relaxed);
```
flaneur2020, line 68: why not fetch_add(1, Ordering::Release)? q.q

query/src/sessions/session_ref.rs
```diff
@@ -57,14 +56,15 @@ impl Drop for SessionRef {
 impl Session {
     pub fn destroy_session_ref(self: &Arc<Self>) {
         if self.ref_count.fetch_sub(1, Ordering::Release) == 1 {
-            std::sync::atomic::fence(Acquire);
+            std::sync::atomic::fence(Ordering::Acquire);
```
drmingdrmer, line 59: I think the `fence` call can be removed from both here and the following method `increment_ref_count`.

An operation to the internally referenced `Arc` is safe because it is protected by the fence inside `Arc`.

query/src/sessions/session_ref.rs
ariesdevil: As the [doc](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicIsize.html#method.fetch_add) said: Using Release makes the load part Relaxed.

query/src/sessions/session_ref.rs
```diff
@@ -57,7 +56,6 @@ impl Drop for SessionRef {
 impl Session {
     pub fn destroy_session_ref(self: &Arc<Self>) {
         if self.ref_count.fetch_sub(1, Ordering::Release) == 1 {
```
drmingdrmer, line 58: And this Ordering can be Relaxed too...

ariesdevil: Yup, you're always right.

drmingdrmer: Aye, I'm afraid you're right on this.

zhang2014 (approved):

drmingdrmer (approved):

query/src/sessions/session_ref.rs
```diff
@@ -56,8 +55,7 @@ impl Drop for SessionRef {
 
 impl Session {
     pub fn destroy_session_ref(self: &Arc<Self>) {
-        if self.ref_count.fetch_sub(1, Ordering::Release) == 1 {
-            std::sync::atomic::fence(Acquire);
+        if self.ref_count.fetch_sub(1, Ordering::Relaxed) == 1 {
```
flaneur2020, line 58: why takes Relaxed memory ordering? in my understanding this means only modify the variable locally without memory barrier at all

flaneur2020: I see, dereferencing self have a memory barrier here implicitly.

but dereferencing self may not necessarily means the varaible ref_count is synced into other cores.

thread 1:
1. deferencing self
2. read barrier
3. increment ref_count
4. (no barrier)

thread 2:
1. deferencing self
2. read barrier
3. read ref_count
4. (still the old one)

flaneur2020: IMO memory barrier is not a very high cost operation (espessically in x86), if we can not 100% sure about the semantics across the cores and it's not a critical path, selecting always the strongest ordering is an OK solution.

drmingdrmer: Because reordering an operation to the internal `Arc` and operation on this atomic will never result in an unexpected state?

flaneur2020: reordering includes the compiler reordering, and also the memory reordering.

I guess the Arc dereference operation can prevent the compiler reordering, but  deferencing Arc only produces a read barrier at most, however make a synced modification to other cores requires a write barrier 🤔

drmingdrmer: Right, but I do not think without a write barrier there will be an unexpected result.

The correctness of this piece of code does not depend on the state inside the `Arc`. Thus no matter whether a write operation on another core happened before or after updating this atomic ref_count, it always gets the right result.

PR #6853, opened 2022-07-27
feat: Forward COPY and STAGE to new planner entirely 

Xuanwo:

## Summary

This PR will forward COPY and STAGE-related queries to the new planner entirely.

Whether new planner has been enabled or not, the databend will skip the whole query and forward to the new planner instead.

All stateful tests have been updated.


BohuTANG:
Not found the forward codes, could you explain how it works?

ZhiHanZ (approved):

Xuanwo:
> Not found the forward codes, could you explain how it works?

Forward has been implemented long ago: We parse with old planner first. if  the statement is supported, we will parse with new planner again.

The problem is if old planner parsed with error, we will return directly and can't forward the query to new planner.

This PR fixed by skip the entire query and  let new planner to handle them.

PsiACE (approved):

sundy-li (approved):

PR #6866, opened 2022-07-28
fix(planner): count aggregation function in subquery

xudong963:

## Summary

For scalar subquery, we'll convert it to single join.

Single join is similar to left outer join, if there isn't matched row in the right side, we'll add **NULL** value for the right side.

But for **count** aggregation function, NULL values should be 0.

Fixes #issue

### DIFF
--- a/query/src/sql/optimizer/heuristic/subquery_rewriter.rs
+++ b/query/src/sql/optimizer/heuristic/subquery_rewriter.rs
@@ -18,6 +18,7 @@ use common_datavalues::BooleanType;
 use common_datavalues::DataTypeImpl;
 use common_datavalues::DataValue;
 use common_datavalues::NullableType;
+use common_datavalues::UInt64Type;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_functions::aggregates::AggregateFunctionFactory;
@@ -55,6 +56,10 @@ pub enum UnnestResult {
     SimpleJoin, // SemiJoin or AntiJoin
     MarkJoin { marker_index: IndexType },
     SingleJoin,
+}
+
+pub struct FlattenInfo {
+    pub from_count_func: bool,
 }
 
 /// Rewrite subquery into `Apply` operator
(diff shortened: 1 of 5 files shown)


query/src/sql/optimizer/heuristic/subquery_rewriter.rs
```diff
@@ -293,16 +306,44 @@ impl SubqueryRewriter {
...
+                        table_name: None,
+                        column_name: name,
+                        index,
+                        data_type,
+                        visible_in_unqualified_wildcard: false,
+                    },
+                });
+
+                let scalar = if flatten_info.from_count_func {
+                    // convert count aggregate function to multi_if function, if count() is null, then 0 else count()
```
leiysky, line 321: Why do we need to rewrite `COUNT(*)` into `CASE WHEN COUNT(*) IS NULL THEN 0 ELSE COUNT(*) END`?

AFAIK, `COUNT(*)` will never return a `NULL` value.

xudong963: Single join is similar to left outer join, if there isn't matched row on the right side, we'll add NULL value for the right side.  At the time, count(*) will be a NULL value.

tests/logictest/suites/query/subquery.test
```diff
@@ -583,7 +583,44 @@ ORDER BY c.c_id, o.o_id
 5  NULL  NULL
 6  90    WA
 
+onlyif mysql
+statement query II
+SELECT
```
xudong963, line 588: You can consider the case, before the PR, the result is 
```sql
----
1  3
2  3
3  NULL
4  2
5  NULL
6  1

query/src/sql/optimizer/heuristic/subquery_rewriter.rs
leiysky: Ok, I get it.

So the `CASE WHEN` is actually applied to the column produced by single join  but not the `COUNT(*)`. 

And conincidentally, different from other set functions, `COUNT(*)` won't return NULL on empty input so it cannot utilize with the NULL produced by a single join as other aggregate functions.

Is that correct?

xudong963: Yes, you totally got it!

leiysky (approved):

PR #6867, opened 2022-07-28
refactor(expr): add NullableColumn and NullableColumnBuilder

andylokandy:

## Summary

- Add NullableColumn and NullableColumnBuilder

Tracked in #6547


common/expression/src/types/nullable.rs
```diff
@@ -188,10 +150,60 @@ impl<T: ArgType> ArgType for NullableType<T> {
...
+    pub column: T::Column,
+    pub validity: Bitmap,
+}
+
+impl<T: ValueType> NullableColumn<T> {
+    pub fn len(&self) -> usize {
+        self.validity.len()
+    }
+
+    pub fn index(&self, index: usize) -> Option<Option<T::ScalarRef<'_>>> {
```
sundy-li, line 168: Just one option is ok ?

andylokandy: The first `Option` states whether the index is inbound and the second one states whether it's NULL.

leiysky (approved):

sundy-li (approved):

Issue #6876, opened 2022-07-28
Bug: `trim(null)` will panic in new planner


leiysky:
Error message:
```
mysql> set enable_planner_v2 = 1;
Query OK, 0 rows affected (0.02 sec)
mysql> select trim(null);
ERROR 1105 (HY000): Code: 1058, displayText = downcast column error, column type: "Null", expected column: "common_datavalues::columns::string::StringColumn" (while in processor thread 0).
```

TianLangStudio:
Maybe we can return null when the argument is null------------------------------------------------------------------
发件人：Yu ***@***.***>
日　期：2022年07月28日 19:08:34
***@***.***>
***@***.***>
主　题：[datafuselabs/databend] Bug: `trim(null)` will panic in new planner (Issue #6876)

Error message:
mysql> set enable_planner_v2 = 1;
Query OK, 0 rows affected (0.02 sec)
mysql> select trim(null);
ERROR 1105 (HY000): Code: 1058, displayText = downcast column error, column type: "Null", expected column: "common_datavalues::columns::string::StringColumn" (while in processor thread 0).

sundy-li:
```
 let box (trim_source, _) = self.resolve(expr, Some(StringType::new_impl())).await?;
        let args = vec![trim_source, trim_scalar];
        let func = FunctionFactory::instance()
            .get(func_name, &[&StringType::new_impl(); 2])
            .map_err(|e| ErrorCode::SemanticError(span.display_error(e.message())))?;

```

It's wrong to set `StringType::new_impl()`

PR #6883, opened 2022-07-29
fix(planner): fix trim NULL

xudong963:

## Summary

Summary about this PR

Fixes https://github.com/datafuselabs/databend/issues/6876

### DIFF
--- a/query/src/sql/planner/semantic/type_check.rs
+++ b/query/src/sql/planner/semantic/type_check.rs
@@ -1390,37 +1390,35 @@ impl<'a> TypeChecker<'a> {
         expr: &Expr<'_>,
         trim_where: &Option<(TrimWhere, Box<Expr<'_>>)>,
     ) -> Result<Box<(Scalar, DataTypeImpl)>> {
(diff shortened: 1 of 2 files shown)


query/src/sql/planner/semantic/type_check.rs
```diff
@@ -1410,17 +1410,17 @@ impl<'a> TypeChecker<'a> {
...
-            .get(func_name, &[&StringType::new_impl(); 2])
+            .get(func_name, &[&source_type, &StringType::new_impl()])
             .map_err(|e| ErrorCode::SemanticError(span.display_error(e.message())))?;
 
         Ok(Box::new((
             FunctionCall {
                 arguments: args,
                 func_name: func_name.to_string(),
-                arg_types: vec![StringType::new_impl(); 2],
+                arg_types: vec![source_type, StringType::new_impl()],
```
sundy-li, line 1423: The second data type may be null type too.

xudong963: Yes, updated

sundy-li (approved):

PR #6907, opened 2022-07-29
feat(functions): migrate quote, reverse and ascii to new functions crate

ygf11:

## Summary

Migrate `quote`, `reverse` and `ascii` to functions-v2. 

Part of #6766


common/functions-v2/tests/it/scalars/string.rs
```diff
@@ -104,14 +107,54 @@ fn test_to_base64(file: &mut impl Write) {
...
         "a",
         DataType::String,
         build_string_column(&["QWJj", "MTIz"]),
     )])
 }
 
+fn test_quote(file: &mut impl Write) {
+    run_ast(file, "quote('a\0b')", &[]);
+    // run_ast(file, "quote('a\'b')", &[]);
+    run_ast(file, "quote('a\"b')", &[]);
```
ygf11, line 121: It seems Parser can not recgonize `\'`.

common/functions-v2/tests/it/scalars/string.rs
```diff
@@ -104,14 +107,54 @@ fn test_to_base64(file: &mut impl Write) {
...
+    run_ast(file, "from_base64(a)", &[(
         "a",
         DataType::String,
         build_string_column(&["QWJj", "MTIz"]),
     )])
 }
 
+fn test_quote(file: &mut impl Write) {
+    run_ast(file, "quote('a\0b')", &[]);
+    // run_ast(file, "quote('a\'b')", &[]);
```
andylokandy, line 120:
```suggestion
    run_ast(file, r#"quote('a\'b')"#, &[]);
```

andylokandy: We may also apply it for other tests for `quote()`

common/functions-v2/tests/it/scalars/string.rs
andylokandy: please use raw strings for the following test cases for quote(), because they all unexpectedly unescpaed by rust.

ygf11: Added for other test cases.

common/functions-v2/src/scalars/string.rs
```diff
@@ -133,6 +133,56 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                Ok(())
+            },
+        ),
+    );
+
+    registry.register_1_arg::<StringType, NumberType<u8>, _, _>(
+        "ascii",
+        FunctionProperty::default(),
+        |_| None,
+        |val| val[0],
```
sundy-li, line 184: The ASCII() will return 0 If the string is empty. The old function had this bug.

common/functions-v2/tests/it/scalars/string.rs
ygf11: Addressed.

common/functions-v2/src/scalars/string.rs
ygf11: Addressed.

common/functions-v2/tests/it/scalars/string.rs
```diff
@@ -34,6 +34,9 @@ fn test_string() {
     test_char_length(file);
     test_to_base64(file);
     test_from_base64(file);
+    test_quote(file);
+    test_reverse(file);
+    test_ascii(file);
```
andylokandy, line 39: It will be better if we add more complex cases like emojis and Chinese.

common/functions-v2/src/scalars/string.rs
```diff
@@ -133,6 +133,56 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                writer.commit_row();
+                Ok(())
+            },
+        ),
+    );
+
+    registry.register_1_arg::<StringType, NumberType<u8>, _, _>(
+        "ascii",
+        FunctionProperty::default(),
+        |_| None,
```
andylokandy, line 183: Maybe we can calculate the domain?

sundy-li (approved):

common/functions-v2/tests/it/scalars/string.rs
ygf11: The `reverse` function simply reverse the byte array now, so I think `Chinese` and `emojis` tests are no meaning.

Shall we treat the input as utf8 string, then reverse?

common/functions-v2/src/scalars/string.rs
ygf11: I added the domain, but not sure it is right...

common/functions-v2/tests/it/scalars/string.rs
andylokandy: I think it is meaningful to record the behavior even if it's not useful to the user. 

Besides, what's the plan on the topic of utf8 char reserve? @sundy-li

andylokandy:
@mergify update

common/functions-v2/tests/it/scalars/string.rs
sundy-li: > Besides, what's the plan on the topic of utf8 char reserve

I  think we can have another function name for that, eg: `reverse_utf8`

Xuanwo:
To get this PR merged, please mark all conversations as resolved.

common/functions-v2/tests/it/scalars/string.rs
ygf11: > I think it is meaningful to record the behavior even if it's not useful to the user.

Added.

PsiACE (approved):

andylokandy (approved):

PR #6918, opened 2022-07-31
feat(chunk): make chunk support scalar values

sundy-li:

## Summary

Summary about this PR

1. Make chunk support scalar values
2. Add extra chunk_info into the chunk struct

Fixes #issue


common/expression/tests/it/main.rs
```diff
@@ -953,13 +953,19 @@ fn run_ast(file: &mut impl Write, text: &str, columns: &[(&str, DataType, Column
         let domain_calculator = DomainCalculator::new(input_domains.clone());
         let output_domain = domain_calculator.calculate(&expr)?;
 
+        let mut num_rows = 0;
+        if !columns.is_empty() {
+            num_rows = columns[0].2.len();
+        }
```
andylokandy, line 959:
```suggestion
        let num_rows = columns.iter(|col| col.2.len()).max().unwrap_or(0);
```

common/expression/src/chunk.rs
```diff
@@ -12,29 +12,69 @@
...
-use itertools::Itertools;
-
-use crate::values::Column;
+use crate::types::AnyType;
+use crate::Value;
 
+/// Chunk is a lightweight container for a group of values.
 pub struct Chunk {
-    columns: Vec<Column>,
+    values: Vec<Value<AnyType>>,
```
andylokandy, line 20: I think `columns` as the field name may sound more reasonable in the context of `Chunk`.

common/expression/src/chunk.rs
```diff
@@ -12,29 +12,69 @@
...
+                    Value::Column(c) => c.len() != num_rows,
+                })
+                .count()
+                == 0
+        );
+        Self {
+            values,
+            num_rows,
+            chunk_info: None,
+        }
```
andylokandy, line 44:
```suggestion
        Self::new_with_info(values, num_rows, None)
```

common/expression/src/chunk.rs
sundy-li: Because we should care about the constant column, which is memory saved representation than column.

common/expression/src/display.rs
```diff
@@ -61,18 +61,17 @@ impl Display for Chunk {
         let mut table = Table::new();
         table.load_preset("||--+-++|    ++++++");
 
-        table.set_header((0..self.num_columns()).map(|idx| format!("Column {idx}")));
+        table.set_header((0..self.num_columns()).map(|idx| format!("Value {idx}")));
```
andylokandy, line 64: Is it forgotten to be reverted?

andylokandy (approved):

leiysky (approved):

PR #6937, opened 2022-08-01
chore: add suites argument and some logs

ZeaLoVe:

## Summary

Add some log when logic test skip some test by --run-dir or --skip-dir
Add --suites arg default value is suites in current path.

Fixes #issue


tests/logictest/main.py
```diff
@@ -61,6 +61,8 @@ def run(args):
         type=int,
         help='Run each test many times (useful for e.g. flaky check)')
 
+    parser.add_argument('-q', '--suites', default="suites" , help='Path to suites dir')
```
sundy-li, line 64: `--run-dir` already cover this argument.

sundy-li (approved):

tests/logictest/main.py
youngsofun, line 64: why use `-q`?

ZeaLoVe: copy from databend-test

youngsofun (approved):

Issue #6910, opened 2022-07-30
Streaming load 60GB csv will OOM


Xuanwo:
Follow https://databend.rs/doc/learn/analyze-ontime-with-databend-on-ec2-and-s3

Start (release build) databend-query with local fs.

```shell
:( curl -H "insert_sql:insert into ontime format CSV" -H "skip_header:0" -H "field_delimiter:\t" -F "upload=@t_ontime.csv" -XPUT http://root:@127.0.0.1:8000/v1/streaming_load



curl: (56) Recv failure: Connection reset by peer
curl -H "insert_sql:insert into ontime format CSV" -H "skip_header:0" -H  -F   0.79s user 25.15s system 16% cpu 2:40.03 total
```

Databend Query

```shell
:) ./before-databend-query
Databend Query

Version: v0.7.146-nightly-94af6a2-simd(rust-1.64.0-nightly-2022-07-30T03:25:03.396932975Z)
Log:
    File: enabled level=INFO dir=./.databend/logs
    Stderr: disabled level=INFO
Meta: embedded at ./.databend/meta_embedded
Storage: fs://root=_data

MySQL
    listened at 127.0.0.1:3307
    connect via: mysql -uroot -h127.0.0.1 -P3307
Clickhouse(native)
    listened at 127.0.0.1:9000
    connect via: clickhouse-client --host 127.0.0.1 --port 9000
Clickhouse(http)
    listened at 127.0.0.1:8124
    usage:  echo 'create table test(foo string)' | curl -u root: '127.0.0.1:8124' --data-binary  @-
echo '{"foo": "bar"}' | curl -u root: '127.0.0.1:8124/?query=INSERT%20INTO%20test%20FORMAT%20JSONEachRow' --data-binary @-
Databend HTTP
    listened at 127.0.0.1:8000
    usage:  curl -u root: --request POST '127.0.0.1:8000/v1/query/' --header 'Content-Type: application/json' --data-raw '{"sql": "SELECT avg(number) FROM numbers(100000000)"}'

[1]    1085010 killed     ./before-databend-query
./before-databend-query  1672.18s user 2811.37s system 1291% cpu 5:47.12 total
```

Dmesg

```shell
[...]

PR #6943, opened 2022-08-02
chore: try support io-uring for meta

PsiACE:

## Summary

try enable io-uring for meta, so add a feature named `io-uring`

- common-meta-sled-store
- common-meta-raft-store
- common-meta-embedded
- common-meta-store
- databend-meta
- databend-query

ref: https://github.com/datafuselabs/databend/issues/6910#issuecomment-1201016123

Fixes #issue

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,5 +1,8 @@
 [advisories]
 ignore = [
+    # rio: v0.9.4 https://rustsec.org/advisories/RUSTSEC-2020-0021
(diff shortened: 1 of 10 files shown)


PsiACE:
From observation it may be good for some scenarios, but as it only supports linux and requires a higher kernel version, I have not enabled it by default.

drmingdrmer (approved):
nice

Xuanwo (approved):

lichuang (approved):

Issue #6709, opened 2022-07-20
A logic test auto complete tools


ZeaLoVe:
cc @ZeaLoVe 

It will be very useful if we can generate a **full** test file with a **prototype** test file.

For example, we can write an original **prototype** test file as:
```
statement ok
create table t(a int);

statement ok
insert into t values(1), (2), (3);

statement query I
select * from t;
----
(left empty here)
```

And then we can run the test and fill the result back to test file as:
```
python main.py --generate
```

This will generate a **full** test file:
```
statement ok
create table t(a int);

statement ok
insert into t values(1), (2), (3);

statement query I
select * from t;
----
1
2
3
```

Sometimes, developers may need to modify some test file, this mechanism will be very helpful.

_Originally posted by @leiysky in https://github.com/datafuselabs/databend/discussions/5838#discussioncomment-2946778_

ZeaLoVe:
The better one is auto generate logic test suite by xx.sql file, contains only sql statement, all others get from query response.

ZeaLoVe:
Compare result with mysql or clickhouse to make sure result is right. Take it in consideration.

PR #6950, opened 2022-08-02
fix(planner): fix case expr with case operator equal

### LINKED ISSUE #6944: bug: `case when then` return wrong results
### Search before asking

### Version

main

### What's Wrong?

databend
```sql
mysql> select
    ->        case a+1 when b then 111 when c then 222
    ->         when d then 333  when e then 444 else 555 end as col6
    -> from t1
    -> order by col6;
+------+
| col6 |
+------+
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
|  111 |
+------+
```

pg
```sql
postgres=# select
postgres-#        case a+1 when b then 111 when c then 222
postgres-#         when d then 333  when e then 444 else 555 end as col6
postgres-# from t1
postgres-# order by col6;
 col6
------
  111
  111
  111
  111
  222
  222
  222
  333
  333
  333
  333
  333
  333
  333
  333
  333
  444
  444
  444
  444
  444
  555
  555
  555
  555
  555
  555
  555
  555
  555
(30 rows)
```

### How to Reproduce?

Use table t1 in https://github.com/datafuselabs/databend/blob/main/tests/logictest/suites/ydb/select1-1.test

Then run
```sql
select
       case a+1 when b then 111 when c then 222
        when d then 333  when e then 444 else 555 end as col6
from t1
order by col6;
```

b41sh:

## Summary

If the case expression has `case_value`, compare it with each `when_value` until one of them is equal.

Fixes #6944


xudong963 (commented):
How about adding a test case in to https://github.com/datafuselabs/databend/blob/main/tests/logictest/suites/ydb/select1-1.test

```sql
select
       case a+1 when b then 111 when c then 222
        when d then 333  when e then 444 else 555 end as col6
from t1
order by col6;
```

sundy-li (approved):

tests/logictest/suites/ydb/select1-1.test
```diff
@@ -228,35 +228,35 @@ where d not between 110 and 150
 order by col1,col5,col3,col2,col4;
 
 ----
-102 2 111 1014 103
```
xudong963, line 231: Does this mean the original result was wrong? 🤣 

---

Let's find a better way to check results, not manually. @ZeaLoVe 

There are several bad points in this current approach:
- We need to compare with other databases manually, making mistakes easy to make, such as the case.
- We need to manually write answers which is inefficient, though we can copy answers from database's output, we also need to process format.

xudong963:
There are still cases with errors in the logic test that need to be handled @b41sh

tests/logictest/suites/ydb/select1-1.test
ZeaLoVe: The result is generate by query databend through http handler, it may be wrong after feature changed. maybe use another database is good idea.  Genrator of test suites now has a issue #6709 .

b41sh:
> There are still cases with errors in the logic test that need to be handled @b41sh 

I will fix them.

xudong963 (approved):

PR #6962, opened 2022-08-03
fix(planner): column with null value, but column length is 0

### LINKED ISSUE #6959: bug: column with null value, but column length is 0
### Search before asking

### Version

main

### What's Wrong?

```sql
mysql> select (select count(*) from t1 as x where x.c>t1.c and x.d<t1.d) as col3 from t1 order by col3;
ERROR 1105 (HY000): Code: 1104, displayText = index out of bounds: the len is 0 but the index is 0.
```

Btw, if there isn't `order by`, there is no error

Some debug info
```
[common/datablocks/src/kernels/data_block_take.rs:30] raw =
[
    "+-----+-----+-----+-----+-----+------+------+------+----+",
    "| 0   | 1   | 2   | 3   | 4   | 10   | 14   | 15   | 12 |",
    "+-----+-----+-----+-----+-----+------+------+------+----+",
    "| 131 | 130 | 134 | 133 | 132 | NULL | NULL | NULL | 0  |",
    "+-----+-----+-----+-----+-----+------+------+------+----+",
]
[common/datablocks/src/kernels/data_block_take.rs:36] f = DataField {
    name: "10",
    data_type: UInt64,
    nullable: true,
}
[common/datablocks/src/kernels/data_block_take.rs:37] column = NullableColumn 	 typeid: UInt64	 len: 0	 data: []
```

### How to Reproduce?

Use table t1 in https://github.com/datafuselabs/databend/blob/main/tests/logictest/suites/ydb/select1-1.test

Then run

```sql
 select (select count(*) from t1 as x where x.c>t1.c and x.d<t1.d) as col3 from t1 order by col3;
```

xudong963:

## Summary

First, see the following case, it returns **Empty Set**
```sql
mysql> create table tt(a int);

mysql> select count(*), tt.a from tt group by a;
Empty set (0.02 sec)
```

Now see the failed case in #6959 

```
select (select count(*) from t1 as x where x.c>t1.c and x.d<t1.d) as col3 from t1 order by col3;
```
image

`Filter` gets the `project`'s result in the green box as **empty**

So the build side of `Single Join` is empty.

After `Single Join`, we got empty blocks (**NullableType**) by https://github.com/datafuselabs/databend/blob/main/query/src/pipelines/processors/transforms/hash_join/row.rs#L100.

Then in the `sort` phase, we will call `block_take_by_indices` in the `sort_block` method, then will panic at `Series::take` due to empty column.

---
**Why the debug log in #6959 shows the block of `10   | 14   | 15` isn't empty?**

image

Because the `validity` of the nullable column isn't empty.

Fixes #6959


xudong963:
@sundy-li I wrote a detailed explanation about the bug in summary

sundy-li (approved):

BohuTANG (approved):

PR #6963, opened 2022-08-03
chore(query): Disable overflow-checks

sundy-li:

## Summary

Summary about this PR

Disable `overflow-checks` which may break ci tests.

SQL:	`SELECT AVG(UserID) FROM hits; `
origin_img_v2_1ae12281-7a9f-421d-a70d-cb023939970g

Fixes #issue

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,7 @@ debug = true
 
 [profile.dev]
 split-debuginfo = "unpacked"
+overflow-checks = false
 
 [profile.dev.package]
 addr2line = { opt-level = 3 }


PsiACE (approved):

Xuanwo:
> Disable `overflow-checks` which may break ci tests.

Seems dangerous to me. This change could also ignore the overflow that we don't expect.

Can we use `overflowing_add` or `saturating_add` in the logic that we expect to have overflowing?

sundy-li:
> > Disable `overflow-checks` which may break ci tests.
> 
> Seems dangerous to me. This change could also ignore the overflow that we don't expect.
> 
> Can we use `overflowing_add` or `saturating_add` in the logic that we expect to have overflowing?

That's right!

PR #7006, opened 2022-08-05
fix: hive table partition detection failure if pushdown is none

dantengsky:

## Summary

hive table should keep working even if `pushdown` passed in is `None`

Fixes #issue


dantengsky:
oh no ... incorrect commits

PR #7012, opened 2022-08-05
feat(query): deprecate clickhouse's tcp protocol support

### LINKED ISSUE #6951: bug: with clickhouse tcp mode, insert a timestamp type value with ns the ck client will hang
### Search before asking

### Version

854fc4d

### What's Wrong?

with clickhouse tcp mode, insert a timestamp type value with ns the ck client will hang

### How to Reproduce?

run databend-query

```
./target/debug/databend-query --admin-api-address=127.0.0.1:1234 --clickhouse-handler-host=0.0.0.0
```

use clickhouse-client connect the server:

```
ClickHouse client version 22.6.3.35 (official build).
Connecting to 192.168.191.66:9000 as user u1.
Connected to databend server version 2021.5.0 revision 54405.

ClickHouse client version is older than ClickHouse server. It may lack support for new features.

databend :) select version();

SELECT version()

Query id: 7c5c0128-b05b-4402-9827-26d03d07668d

Connecting to 192.168.191.66:9000 as user u1.
Connected to databend server version 2021.5.0 revision 54405.

ClickHouse client version is older than ClickHouse server. It may lack support for new features.

┌─version()───────────────────────────────────────────────────────────────────────────────────────┐
│ DatabendQuery v0.7.148-nightly-854fc4d-simd(rust-1.64.0-nightly-2022-08-02T06:44:12.549983597Z) │
└─────────────────────────────────────────────────────────────────────────────────────────────────┘

1 row in set. Elapsed: 0.054 sec. 

databend :) create table t(id int, n datetime);

CREATE TABLE t
(
    `id` int,
    `n` datetime
)

Query id: 9d7424b5-1aa9-49d2-8a61-67e3597c5d85

Ok.

0 rows in set. Elapsed: 0.046 sec. 

databend :) insert into t values(1, '2022-02-02 00:01:02');

INSERT INTO t FORMAT Values

Query id: b4887ff3-24b4-460d-a8d3-39db7022dad9

Ok.

1 row in set. Elapsed: 0.055 sec. 

databend :) insert into t values(1, '2022-02-02 00:01:02.123'); -- this sql will be hang.

INSERT INTO t FORMAT Values

Query id: 764006d6-70dc-4330-811b-593e6c66ffe0

```

* Note: with `mysql` and `http` and `clickhouse http` is ok.

So I think the hang is the ck client hang with some reason.

Because the datetime type in clickhouse just support second precision

sundy-li:

## Summary

Summary about this PR

deprecate clickhouse's tcp protocol support

See the RFC document for the reasons.

Fixes #6951


sundy-li:
@Xuanwo  @flaneur2020 


Is that ok if we remove these useless configs? Or just keep them as stale configs.

I prefer to remove them.

```
# Query ClickHouse Handler.
pub clickhouse_handler_host: String,
pub clickhouse_handler_port: u16,
```

Xuanwo:
We can remove them in the inner config, but still need to keep them in the outer config until we decide to do a break change to out config.

sundy-li:
鄙人生平有两大爱好：

1. 写代码
2. 删自己写过的代码

Xuanwo:
> 鄙人生平有两大爱好：
> 
> 1. 写代码
> 2. 删自己写过的代码

我的爱好比较简单：删 @sundy-li 写过的代码

PsiACE (commented):
LGTM :rocket: 

Now that the clickhouse tcp protocol has been removed, I think we no longer rely on clickhouse-driver.

I am also concerned if we need to keep the opensrv-clickhouse introduced in datavalues

Xuanwo (approved):

BohuTANG (approved):

PR #7013, opened 2022-08-05
fix(planner): Fix prune projection

leiysky:

## Summary
Keep at least one column for `Project` operator.

### DIFF
--- a/query/src/sql/optimizer/heuristic/prune_columns.rs
+++ b/query/src/sql/optimizer/heuristic/prune_columns.rs
@@ -116,7 +116,13 @@ impl ColumnPruner {
                 ))
             }
             RelOperator::Project(p) => {
-                let used: ColumnSet = p.columns.intersection(&required).cloned().collect();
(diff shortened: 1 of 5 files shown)


sundy-li (approved):

query/src/sql/optimizer/heuristic/prune_columns.rs
```diff
@@ -116,7 +116,13 @@ impl ColumnPruner {
                 ))
             }
             RelOperator::Project(p) => {
-                let used: ColumnSet = p.columns.intersection(&required).cloned().collect();
+                let mut used: ColumnSet = p.columns.intersection(&required).cloned().collect();
+                if used.is_empty() {
+                    // Keep at least one column for project.
+                    used.insert(*p.columns.iter().sorted().take(1).next().ok_or_else(|| {
```
andylokandy, line 122: What about using `max()` instead of `sorted().take(1).next()`

Issue #6042, opened 2022-06-17
Object name case convention


leiysky:
> > lowercase all catalog, database, and tables names
> 
> Is this behavior commonly used in other DBMS ? I thought it's always case-sensitive.

In most of the databases, the default behaviour is case-insensitive. 

And the case-sensitivity of identifiers is controled in different ways, for example:
- SQL Server has variables `SQL_IDENTIFIER_CASE` and `SQL_QUOTED_IDENTIFIER_CASE` to control this [document](https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/identifier-case?view=sql-server-ver16)
- The rule of MySQL is complicated, but in most case, the identifiers are case-insensitive https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html
- In Postgres, quoted identifiers are case-sensitive, and the rest are case-insensitive, which basically follows ANSI SQL
- Snowflake is compatible with Postgres, however, it still provide some variables to control the case-sensitivity as SQL Server does, to support different data upstream(we are going to follow this way). https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html

This is work about normalizing the convention of identifiers, which is correlated with #5717 . 

This PR only makes minor refactor of AST and unifies the behaviour of identifier case handling in new planner. It doesn't aim to resolve the issue of identifier case.

 I have made an offline discussion with @andylokandy , we decide to get rest work done later.

_Originally posted by @leiysky in https://github.com/datafuselabs/databend/issues/5909#issuecomment-1153059898_

Relate to #6041

PR #7026, opened 2022-08-08
feat(planner): Enhanced case-sensitivity of identifiers

leiysky:

## Summary

Introduce two settings `quoted_ident_case_sensitive` and `unquoted_ident_case_sensitive` to control the behavior of case-sensitivity.

By default, the **unquoted** identifiers are case-insensitive(`unquoted_ident_case_sensitive = 0`) and the **quoted** identifiers are case-sensitive(`quoted_ident_case_sensitive = 1`), which follows the SQL standard.

The case sensitivity takes effect as soon as we are creating or accessing an object.

For example, if we are running such a `CREATE TABLE` statement:
```sql
create table "T"(a int);
```

A table with the name **T** in the upper case will be created, since it's assigned with a **quoted** identifier which is case-sensitive. Therefore, if we want to access it in a query statement, we should also use a **quoted** identifier **"T"** in upper case as:
```sql
select * from "T"; -- Ok
select * from T; -- Error
```

This rule is also applied to other objects, e.g. columns, views and etc.

For the **case-insensitive** identifiers(i.e. **unquoted** identifiers by default), we will transform the identifier into **lower case** first and then do the creation or access. With this rule, it's possible to use a **case-insensitive identifier** to access an object assigned with a **case-sensitive identifier** in lower case, just as:
```sql
create table "t"("a" int);
select a from t; -- Ok
select A from T; -- Ok

create table t1(a int);
select "a" from "t1"; -- Ok
```

We can achieve compatibility with other databases by modifying the settings mentioned above.

For example, in **MySQL** all the identifiers are **case-insensitive** by default, whether it's **unquoted** or not. This behavior can be enabled by `set quoted_ident_case_sensitive = 0`.

Refer to test cases for more details.

Part of #6042


leiysky:
Blocked by https://github.com/datafuselabs/databend/issues/7027

query/src/sql/planner/semantic/name_resolution.rs
```diff
@@ -0,0 +1,75 @@
...
+        })
+    }
+}
+
+/// Normalize identifier with given `NameResolutionContext`
+pub fn normalize_identifier<'a>(
+    ident: &Identifier<'a>,
+    context: &NameResolutionContext,
+) -> Identifier<'a> {
+    if ident.quoted() {
```
andylokandy, line 52:
```suggestion
    let case_sensitive = ident.quoted() && context.quoted_ident_case_sensitive ||  !ident.quoted() && context.unquoted_ident_case_sensitive;
    if case_sensitive {
```

leiysky: Fixed.

andylokandy (approved):

Xuanwo (approved):

PR #7029, opened 2022-08-08
refactor: refactor add\remove share accounts API

lichuang:

## Summary

refactor: refactor add\remove share accounts API, support op more then one account per API.

Fixes #issue

### DIFF
--- a/common/meta/api/src/share_api_test_suite.rs
+++ b/common/meta/api/src/share_api_test_suite.rs
@@ -216,6 +216,39 @@ impl ShareApiTestSuite {
         let create_on = Utc::now();
         let if_exists = true;
 
+        info!("--- add and remove account with not exist share");
+        {
+            let req = AddShareAccountsReq {
+                share_name: share_name.clone(),
+                share_on,
+                if_exists: false,
+                accounts: vec![account.to_string()],
+            };
+
+            // get share meta and check account has been added
+            let res = mt.add_share_accounts(req).await;
+            assert!(res.is_err());
+            let err = res.unwrap_err();
+            assert_eq!(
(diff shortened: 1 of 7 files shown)


common/meta/types/src/app_error.rs
```diff
@@ -283,39 +283,43 @@ impl ShareAlreadyExists {
...
+    accounts: Vec<String>,
     context: String,
 }
 
-impl ShareAccountAlreadyExists {
+impl ShareAccountsAlreadyExists {
     pub fn new(
         share_name: impl Into<String>,
-        account: impl Into<String>,
+        accounts: impl Into<Vec<String>>,
```
drmingdrmer, line 296: Better using `Iterator<Item=String>` or `&[String]` here, IMHO.

`Into<String>` would be a good abstraction for a single account, but `Into<Vec<String>>` is not very common I think.

common/meta/api/src/share_api_test_suite.rs
```diff
@@ -286,33 +287,34 @@ impl ShareApiTestSuite {
...
 
         info!("--- add account account2");
         {
-            let req = AddShareAccountReq {
+            let req = AddShareAccountsReq {
                 share_name: share_name.clone(),
                 share_on,
-                account: account2.to_string(),
+                if_exists,
+                accounts: vec![account2.to_string()],
```
drmingdrmer, line 312: Should there be another test case for `if_exists`? Or this feature is included in other cases?

drmingdrmer (commented):
Except these two minor issues, everything else looks good to me:D

drmingdrmer (approved):

PR #7054, opened 2022-08-09
feat(expr): implement adaptive constant folding

andylokandy:

## Summary

Adaptive constant folding is more aggressive than trivial constant folding. It optimizes the expression depending on the data distribution in a chunk (aka domain), and rewrite sub-expression to constant if it is sure that all evaluation result from the specified chunk is the same.

Summarize the main differences:

1. Adaptive constant folding can take place not only in the optimizer but also in execution for each chunk (aka. datablock).
2. Adaptive constant folding can even fold expressions containing `ColumnRef`.

Let's say there is a column `a :: Int8 NULL`, and we know, in a specified chunk, `a` doesn't contain `NULL`, therefore adaptive constant folding will rewrite the expression `a IS NULL AND TRUE` into `FALSE`.


leiysky (approved):

sundy-li:
/LGTM

Conflicts need to be resolved.

Xuanwo (approved):

PR #7056, opened 2022-08-09
feat(planner): support common table expression(CTE)

### LINKED ISSUE #6246: Support CTE(Common Table Expression)
Need RFC.

xudong963:

## Summary

Support CTE in planner.

Note: we don't support recursive CTE currently.

In the future, we can optimize CTE, there is a proposal: https://github.com/datafuselabs/databend/issues/7061

Fixes #6246


query/src/sql/planner/binder/select.rs
```diff
@@ -56,7 +57,7 @@ pub struct SelectItem<'a> {
 impl<'a> Binder {
     pub(super) async fn bind_select_stmt(
         &mut self,
-        bind_context: &BindContext,
+        bind_context: &mut BindContext,
```
leiysky, line 60: Seems there is no need to make this mutable. Since we will always return an owned `BindContext`.

xudong963: `bind_context` will be used in `bind_table_reference` which requires it mutable.

query/src/sql/planner/binder/select.rs
```diff
@@ -176,11 +178,29 @@ impl<'a> Binder {
...
+                        "duplicate cte {table_name}"
+                    )));
+                }
+                let (s_expr, cte_bind_context) = self.bind_query(bind_context, &cte.query).await?;
+                let cte_info = CteInfo {
+                    columns_alias: cte.alias.columns.iter().map(|c| c.name.clone()).collect(),
+                    s_expr,
+                    bind_context: cte_bind_context.clone(),
+                };
+                bind_context.ctes_map.insert(table_name, cte_info);
```
BohuTANG, line 201: It would be better to change ctes_map to rwlock and add a function for the map operation? Make all bind_context to immutable.

xudong963: Make sense

query/src/sql/planner/binder/bind_context.rs
```diff
@@ -21,10 +24,12 @@ use common_datavalues::DataSchemaRefExt;
 use common_datavalues::DataTypeImpl;
 use common_exception::ErrorCode;
 use common_exception::Result;
+use parking_lot::RwLock;
```
BohuTANG, line 27: Why use parking_log Rwlock?

BohuTANG: Got

fkuner:
Maybe we can add more tests to verify it. I wrote some tests from clickhouse in thsi pr [simple cte](https://github.com/datafuselabs/databend/pull/6637). You can verity it.

leiysky (approved):

sundy-li (approved):

fkuner (approved):

BohuTANG:
For documentation for common table expression(CTE) , cc @soyeric128 

Reference:
snow: https://docs.snowflake.com/en/user-guide/queries-cte.html
https://dev.mysql.com/doc/refman/8.0/en/with.html#common-table-expressions

PR #7045, opened 2022-08-09
feat(query): parquet schema case in-senstive match

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

1. parquet schema case in-senstive match
2. remove useless `set enable_planner_v2 = 1` in ci

Summary about this PR

Fixes #issue

PR #7060, opened 2022-08-10
fix(planner): Fix case sensitivity of cluster by expression

leiysky:

## Summary

Introduce `Visitor` and `VisitorMut` to help traverse through AST.

Fix the issue that cannot handle `CLUSTER BY` expression with case-insensitive identifiers in the upper case.

Will fix issue encountered in #7045


query/tests/it/sql/planner/semantic/name_resolution.rs
```diff
@@ -114,3 +119,20 @@ fn test_normalize_identifier_unquoted_case_sensitive() {
         );
     }
 }
+
+#[test]
+fn test_normalize_identifiers_in_expr() {
```
andylokandy, line 124: Are you sure the unit test is turned on?

xudong963 (approved):

BohuTANG (approved):

PR #7067, opened 2022-08-10
feat(expression): add a helper trait ColumnFrom

sundy-li:

## Summary

Summary about this PR

add a helper trait ColumnFrom.

Previous way to generate a column and a nullable column:
```
Column::Int32(vec![5, 6].into()),

Column::Nullable(Box::new(NullableColumn {
                column: Column::UInt8(vec![10, 11, 12, 13, 14].into()),
                validity: vec![false, true, false, false, false].into(),
            })),

Column::Nullable(Box::new(NullableColumn {
                column: Column::String(StringColumn {
                    data: "xyzab".as_bytes().to_vec().into(),
                    offsets: vec![0, 1, 2, 3, 4, 5].into(),
                }),
                validity: vec![false, true, true, false, false].into(),
            })),

```

Now we can use an easier way:

```
Column::from_data(vec![5i32, 6]),
Column::from_data_valids(vec![10u8, 11, 12, 13, 14], vec![false, true, false, false, false]),
Column::from_data_valids(vec!["x", "y", "z", "a", "b"], vec![false, true, true, false, false]),
```

Fixes #issue


common/expression/src/values.rs
```diff
@@ -704,10 +705,14 @@ impl Column {
...
             }
-            Column::Nullable(col) => col
-                .column
-                .as_arrow()
-                .with_validity(Some(col.validity.clone())),
+            Column::Nullable(col) => {
+                let arrow_array = col.column.as_arrow();
+                match arrow_array.data_type() {
+                    ArrowType::Null => arrow_array,
+                    ArrowType::Extension(_, t, _) if **t == ArrowType::Null => arrow_array,
```
xudong963, line 712:
```suggestion
                    ArrowType::Null | ArrowType::Extension(_, t, _) if **t == ArrowType::Null => {
                        arrow_array
                    }
```

xudong963 (approved):

common/expression/src/column_from.rs
```diff
@@ -0,0 +1,128 @@
...
+use crate::types::*;
+use crate::Column;
+
+/// ColumnFrom is a helper trait to generate columns.
+pub trait ColumnFrom<D, Phantom: ?Sized> {
+    /// Initialize by name and values.
+    fn from_data(_: D) -> Column;
+
+    fn from_data_valids(d: D, valids: Vec<bool>) -> Column {
+        let column = Self::from_data(d);
```
Xuanwo, line 28: Will we have an assert here to make sure `d` has the same length with `valids`?

BTW, for better naming:

- `from_data_valids` -> `from_data_with_validity`?
- `valids` -> `validity`?

sundy-li: I considered that ,but feel`from_data_with_validity ` is too long to write.

Xuanwo: But easier for reading. I think it's a good tradeoff.

common/expression/src/values.rs
sundy-li: It does not work.

```
error[E0408]: variable `t` is not bound in all patterns
   --> common/expression/src/values.rs:711:21
    |
711 |                     ArrowType::Null | ArrowType::Extension(_, t, _) if **t == ArrowType::Null => {
    |                     ^^^^^^^^^^^^^^^ pattern doesn't bind `t`  - variable not in all patterns
```

common/expression/src/column_from.rs
sundy-li: Ok

Xuanwo (approved):

Issue #7063, opened 2022-08-10
Feature: explain AST/syntax statement


sundy-li:
**Summary**

Description for this feature.

- explain syntax:  output the formatted SQL .

Example:
```
EXPLAIN SYNTAX
SELECT number
FROM numbers(10)

┌─explain──────────┐
│ SELECT number    │
│ FROM numbers(10) │
└──────────────────┘
```	

- explain ast: output the AST struct.

Example:
```
explain ast select number from numbers(10);


┌─explain─────────────────────────────────────┐
│ SelectWithUnionQuery (children 1)           │
│  ExpressionList (children 1)                │
│   SelectQuery (children 2)                  │
│    ExpressionList (children 1)              │
│     Identifier number                       │
│    TablesInSelectQuery (children 1)         │
│     TablesInSelectQueryElement (children 1) │
│      TableExpression (children 1)           │
│       Function numbers (children 1)         │
│        ExpressionList (children 1)          │
│         Literal UInt64_10                   │
└─────────────────────────────────────────────┘
```

PR #7068, opened 2022-08-10
feat(planner): support explain raw plan

### LINKED ISSUE #7065: Feature: explain raw plan
**Summary**

Support explain raw plan (unoptimized query plan) which is useful to debug.

xudong963:

## Summary

1. Support explain unoptimized plan
2. fix `LogicalPlan` display

Fixes #7065

### DIFF
--- a/query/src/interpreters/interpreter_factory_v2.rs
+++ b/query/src/interpreters/interpreter_factory_v2.rs
@@ -14,6 +14,7 @@
 
 use std::sync::Arc;
 
+use common_ast::ast::ExplainKind;
 use common_exception::Result;
 use common_planners::EmptyPlan;
 use common_planners::PlanNode;
@@ -43,8 +44,8 @@ impl InterpreterFactoryV2 {
         matches!(stmt, DfStatement::SeeYouAgain)
     }
 
-    pub fn get(ctx: Arc<QueryContext>, plan: &Plan) -> Result<InterpreterPtr> {
-        let inner = InterpreterFactoryV2::create_interpreter(ctx.clone(), plan)?;
+    pub fn get(ctx: Arc<QueryContext>, plan: &Plan, raw_plan: &Plan) -> Result<InterpreterPtr> {
+        let inner = InterpreterFactoryV2::create_interpreter(ctx.clone(), plan, raw_plan)?;
 
         Ok(Arc::new(InterceptorInterpreter::create(
             ctx,
@@ -55,7 +56,11 @@ impl InterpreterFactoryV2 {
         )))
     }
 
-    fn create_interpreter(ctx: Arc<QueryContext>, plan: &Plan) -> Result<InterpreterPtr> {
+    fn create_interpreter(
+        ctx: Arc<QueryContext>,
+        plan: &Plan,
+        raw_plan: &Plan,
+    ) -> Result<InterpreterPtr> {
         match plan {
             Plan::Query {
                 s_expr,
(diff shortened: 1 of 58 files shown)


query/src/interpreters/interpreter_factory_v2.rs
```diff
@@ -43,8 +44,8 @@ impl InterpreterFactoryV2 {
         matches!(stmt, DfStatement::SeeYouAgain)
     }
 
-    pub fn get(ctx: Arc<QueryContext>, plan: &Plan) -> Result<InterpreterPtr> {
-        let inner = InterpreterFactoryV2::create_interpreter(ctx.clone(), plan)?;
+    pub fn get(ctx: Arc<QueryContext>, plan: &Plan, raw_plan: &Plan) -> Result<InterpreterPtr> {
```
sundy-li, line 47: Maybe we should wrap a new struct, because we need to support `explain sytnax/ast` in #7063  cc @b41sh 

```
struct PlanStmt {
 raw_plan: Plan,
 plan: Plan,
 ast:  Statement
}
```

xudong963: Make sense,  in this PR or in @b41sh  PR?

sundy-li: Better to have a new struct in this pr, thus @b41sh  can work based on this.

xudong963: done

leiysky:
It's better to return the unoptimized plan here:
https://github.com/datafuselabs/databend/blob/dc289e0619daa95dfb7795817c32d6c8985b2db2/query/src/sql/optimizer/mod.rs#L86-L89

xudong963:
> It's better to return the unoptimized plan here:
> 
> https://github.com/datafuselabs/databend/blob/dc289e0619daa95dfb7795817c32d6c8985b2db2/query/src/sql/optimizer/mod.rs#L86-L89

Nice, returning the unoptimized plan from here won't break the `plan_sql` method.

@b41sh seems you still need to return `stmt` from the `plan_sql` method. If so, if I can reserve current change? cc @sundy-li

BohuTANG:
> It's better to return the unoptimized plan here:
> 
> https://github.com/datafuselabs/databend/blob/dc289e0619daa95dfb7795817c32d6c8985b2db2/query/src/sql/optimizer/mod.rs#L86-L89

Does the `explain ...` statement return the un-optimized plan by default?

xudong963:
> Does the `explain ...` statement return the un-optimized plan by default?

No, it returns the optimized plan,  `optimize(ctx, opt_ctx, *plan)?`.

We can do the following change to make `explain raw` return the unoptimized plan

```rust
Plan::Explain { kind, plan } => {
            if &kind == ExplainKind::Raw {
                return Ok(Plan::Explain {
                    kind,
                    plan: plan.clone(),
                });
            }
            Ok(Plan::Explain {
                kind,
                plan: Box::new(optimize(ctx, opt_ctx, *plan)?),
            })
        },
```

b41sh:
> > It's better to return the unoptimized plan here:
> > https://github.com/datafuselabs/databend/blob/dc289e0619daa95dfb7795817c32d6c8985b2db2/query/src/sql/optimizer/mod.rs#L86-L89
> 
> Nice, returning the unoptimized plan from here won't break the `plan_sql` method.
> 
> @b41sh seems you still need to return `stmt` from the `plan_sql` method. If so, if I can reserve current change? cc @sundy-li

yes, I need `stmt`. Can we get `stmt` from `raw_plan`? Or do I need to add an `stmt` field to PlanKind?

leiysky:
> yes, I need `stmt`. Can we get `stmt` from `raw_plan`? Or do I need to add an `stmt` field to PlanKind?

It's a little bit tricky to store the AST in the plan, since it has dependency on source query string. I think it's possible to generate the formatted AST string in `Binder` and store it in `ExplainKind`, like:
```
enum ExplainKind {
    ...,
    AST(String),
}
```

xudong963:
close by https://github.com/datafuselabs/databend/pull/7079

Issue #7050, opened 2022-08-09
Tracking issues of `Better Project Structures`


PsiACE:
**Background**

We have a large number of crates, but they are not properly managed and categorised. Typical problems include

- Many crates re-expose crates that have already been published. (tokio, tracing)
- All split crates are put into the common directory. (user, formats)

**Basic Goals**

- [x] A clean root dir structure, all crates in `src` dir and correctly categorised
- [x] Splitting binary targets and shared libraries. 
- [x] The reclassification of unit and integration tests. and #5001
- [x] #7004
- [ ] Optimising Cargo.toml for releasing to crates.io

Xuanwo:
Will we have an RFC for the new layout?

PsiACE:
> Will we have an RFC for the new layout?

I will enrich it in this issue as it progresses. Perhaps the testing section will have a relatively formal RFC.

Xuanwo:
>  A clean root dir structure, all crates in src dir and correctly categorised

So we will remove `common` and move them into `src` with well-designed crate names?

PsiACE:
> So we will remove `common` and move them into `src` with well-designed crate names?

Most libraries will contain a databend prefix, Finally it looks like this

```
src
├── cmds // all cmds
│   ├── query.rs
│   ├── metabench.rs // or move to tools dir
│   ├── metactl.rs // or move to tools dir
│   └── meta.rs
├── meta // all meta shared libs
│   ├── databend-meta-api
│   ├── databend-meta-types
│   └── ...
├── query // all query shared libs
│   ├── databend-ast
│   ├── databend-planner
│   └── ...
├── utils // common
│   ├── databend-runtime // some common shared-libs
│   ├── rangemap // for those more generic crates, they do not require the databend prefix
│   └── ...
└── workspace-hack // hakari
```

BohuTANG:
Most is done, let's close.

PR #7074, opened 2022-08-10
refactor: reorg workspace, a basic structure

PsiACE:

## Summary

In this PR

-  Splitting binary targets and shared libraries. 
    - move fuzz to query/ast
    - move metabench, metactl to cmds
-  A clean root dir structure, all crates in src dir. 
    - move databend-query to query/service
    - move databend-meta to meta/service
-  Almost all meta-related crates are categorised
    - move common/meta, common/protos, common/proto-conv to meta.
    - add a `README.md` to meta dir
- edit makefile, codeowner and docs

Related #7050


Xuanwo:
`targets` have different meanings in rust, and it seems we don't need to add an extra layer.

How about renaming `cmds` to `bianries`?

```
binaries
  - query
  - meta
  - metabench
  - metactl
```

Xuanwo:
Rename `common` to `components`? Or we will do this in the next parts?

PsiACE:
> Rename `common` to `components`? Or we will do this in the next parts?

we will do this in the next parts

Xuanwo:
> `cargo install databend-bianries` ?

I guess we should only upload libs to `crates.io`. Binaries are so large that not suitable for users to `cargo install` them directly.

BohuTANG:
This is a monster PR!
Can we split into more than one PR? The layout change will block many other PRs, and other PRs will block this PR too.

Xuanwo (approved):
Nice work!

PsiACE:
> I guess we should only upload libs to `crates.io`. Binaries are so large that not suitable for users to `cargo install` them directly.

It has now been renamed, but I still think it can be published to crates.io, which just means that users will need to compile it themselves.

Xuanwo:
> This is a monster PR! Can we split into more than one PR? The layout change will block many other PRs, and other PRs will block this PR too.

Most works are just renaming. git is smart enough to merge them without conflicts.

PsiACE:
> > This is a monster PR! Can we split into more than one PR? The layout change will block many other PRs, and other PRs will block this PR too.
> 
> Most works are just renaming. git is smart enough to merge them without conflicts (I guess).

Yes, most of the time it is easy to avoid file conflicts by simply running `git pull` once. 

As we plan to introduce layout changes, I think it will be difficult to avoid blocking each other.

BohuTANG:
Almost LGTM.
I suggest merging this PR on the weekend or when PR numbers are low, not blocking other PRs which is working.

drmingdrmer (approved):

Xuanwo:
Ooops, there are many conflicts. Git is not smart enough :cry:

PsiACE:
> I suggest merging this PR on the weekend or when PR numbers are low, not blocking other PRs which is working.

I marked it as a draft to avoid merging at an unexpected time.

> Ooops, there are many conflicts. Git is not smart enough

emmm, It requires manual confirmation of the merge. But the fact is that hardly anything has to be modified.

Xuanwo:
This PR shows how frequently we changing our code...

Xuanwo:
cc @PsiACE, Let's try merge this PR today!

Xuanwo:
Github is more stupid than Git.

`git merge` works excellent, but GitHub marks it as conflicts.

PR #7080, opened 2022-08-11
feat(planner): support tuple map access pushdown to storage

### LINKED ISSUE #6794: Feature: Support read elements of a tuple by name and index
**Summary**

Support read elements of a `tuple` by name and index

```sql
create table t(id int, s tuple(a int, b int));
insert into t values(1, (10, 20)),(2, (30, 40));
select s.a, s.b from t;
select s.0, s.1 from t;
```

b41sh:

## Summary

Use map access syntax to select tuple inner columns and push down the query to the storage layer, only need to read the required columns.

1. `ColumnEntry` add field `path_indices` as an index to find inner columns.
2. add tuple inner columns to `Metadata`.
3. modify the `Extras` field `projection` as `enum Projection` in order to access inner columns.

for example
```sql
CREATE TABLE t(id Int, t Tuple(a Tuple(m Int64, n Int64), b Tuple(x Int64, y Int64))) Engine = Fuse;
INSERT INTO t (id, t) VALUES(1, ((10, 11), (20, 21))), (2, ((30, 31), (40, 41)));
SELECT t:a:m, t:a:n, t:b[0], t:b[1] FROM t;

+-------+-------+--------+--------+
| t:a:m | t:a:n | t:b[0] | t:b[1] |
+-------+-------+--------+--------+
| 10    | 11    | 20     | 21     |
| 30    | 31    | 40     | 41     |
+-------+-------+--------+--------+
```

Fixes #6794


common/planners/src/plan_node_extras.rs
```diff
@@ -12,13 +12,41 @@
...
 use crate::Expression;
 
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub enum Projection {
+    /// column indices of the table
+    Columns(Vec<usize>),
+    /// inner column indices for tuple data type with inner columns.
+    /// the key is the column_index of ColumnEntry.
+    /// the value is the path indices of inner columns.
+    InnerColumns(BTreeMap<usize, Vec<usize>>),
```
leiysky, line 26: Seems the nested fields of the struct are already flattened, why do we keep track of this mapping?

b41sh: For inner columns, `column_index` does not exist in the table schema. We need to use `path_indices` to read them in the storage layer.

for example
```sql
create table test(id int, t tuple(a int, b int))
```

the table has the following columns

```
| name | index | path_indices |
| ---- | ----- | ------------ |
| id   | 0     | [0]          |
| t    | 1     | [1]          |
| t:a  | 2     | [1, 0]       |
| t:b  | 3     | [1, 1]       |
```
but the table schema only has fields `id` and `t`, if we want query `t:a` and `t:b`, we need to use path `path_indices`
`[1,0]` and `[1,1]` to read them in the storage layer

sundy-li (approved):

BohuTANG:
@mergify update

BohuTANG (approved):

PR #7098, opened 2022-08-12
fix(hive): fix table column name and parquet column name may not case…

sandflee:

## Summary
fix the cases table column name is 'A', while parquet file column name is 'a'


Xuanwo:
cc @sundy-li @leiysky 

Is it possible for us to have column names like `Abc` and `abc`?

BohuTANG:
> cc @sundy-li @leiysky
> 
> Is it possible for us to have column names like `Abc` and `abc`?

No, I think the user doesn't do that.

BohuTANG (approved):

PR #7117, opened 2022-08-15
chore(query): remove udf parser in ast crate

sundy-li:

## Summary

Summary about this PR

Fixes #issue


sundy-li:
How about moving `ast` crate into `common/ast` ?

Xuanwo:
> How about moving `ast` crate into `common/ast` ?

I believe ast should become `databend-parser`.

src/common/legacy-parser/src/analyzer/analyzer_expr_sync.rs
```diff
@@ -13,27 +13,25 @@
...
-use common_ast::udfs::UDFTransformer;
 use common_datavalues::prelude::*;
 use common_datavalues::type_coercion::merge_types;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_functions::aggregates::AggregateFunctionFactory;
-use common_functions::is_builtin_function;
-use common_meta_types::UserDefinedFunction;
 use common_planners::Expression;
+use sqlparser::ast::BinaryOperator;
```
Xuanwo, line 22: cc @andylokandy, maybe we should also implement udf in the new plnner?

PsiACE:
> How about moving ast crate into common/ast ?

We will move all the parts that belong to the query into the query directory, so there is no need to move ast back to common. #7050

src/common/legacy-parser/src/analyzer/analyzer_expr_sync.rs
sundy-li: It's already done.

Xuanwo (approved):

PsiACE (approved):

PR #7126, opened 2022-08-15
feat(query): support datetime format 

### LINKED ISSUE #6947: tasks:support cast '2022-02-01T00:00:00+08:00'  into timestamp
tasks:
- [ ] support cast '2022-02-01T00:00:00+08:00'  into timestamp

_Originally posted by @sundy-li in https://github.com/datafuselabs/databend/issues/6522#issuecomment-1176983810_

TCeason:

## Summary

support datetime format:

`%Y-%m-%dT%H:%M:%S%.f%z`

and

`%Y-%m-%d`

Fixes #6947


TCeason:
```
14_0007_http_clickhouse_input_format_diagnostic:                        [ FAIL ] - result differs with:
--- /workspace/tests/suites/0_stateless/14_clickhouse_http_handler/14_0007_http_clickhouse_input_format_diagnostic.result	2022-08-15 14:27:37.998746522 +0000
+++ /workspace/tests/suites/0_stateless/14_clickhouse_http_handler/14_0007_http_clickhouse_input_format_diagnostic.stdout	2022-08-15 14:29:55.257439501 +0000
@@ -1,6 +1,6 @@
 Code: 4000, displayText = invalid data (Expected to have char '.)
 Error occurs at row: 1:
-	Column: 0, Name: a, Type: Timestamp, Parsed text: "<SINGLE QUOTE>19892-02-03 15:23:2"
+	Column: 0, Name: a, Type: Timestamp, Parsed text: "<SINGLE QUOTE>19892-02-0"
 	ERROR: DateTime must be in YYYY-MM-DD hh:mm:ss format.
 .
 
@@ -20,9 +20,9 @@
 	Error: There is no line feed. "H" found instead.
 .
 
-Code: 1046, displayText = Cannot parse value:"'2023-04-08 01:01:0" to DateTime type, cause: input contains invalid characters
+Code: 1046, displayText = Cannot parse value:'2023-04-0 to Date type, cause: input contains invalid characters
 Error occurs at row: 0:
-	Column: 0, Name: a, Type: Timestamp, Parsed text: "<SINGLE QUOTE>2023-04-08 01:01:0"
+	Column: 0, Name: a, Type: Timestamp, Parsed text: "<SINGLE QUOTE>2023-04-0"
 	ERROR: DateTime must be in YYYY-MM-DD hh:mm:ss format.
 .
 ```

I will fix this test.

**Comment by TCeason** [comment]
14_0007_http_clickhouse_input_format_diagnostic

has some err with this test. I will fix it tomorrow.

File: src/common/functions/src/scalars/expressions/cast_from_string.rs [hunk 1]
```diff
@@ -103,9 +106,26 @@ pub fn string_to_timestamp(date_str: impl AsRef<[u8]>, tz: &Tz) -> Option<DateTi
     let s = std::str::from_utf8(date_str.as_ref()).ok();
     if let Some(s) = s {
```
sundy-li, line 107: These codes are duplicated by `read_timestamp_text`.

I think we just need to keep `read_timestamp_text`  are call this function in `cast_from_string`

TCeason: > These codes are duplicated by `read_timestamp_text`.
> 
> I think we just need to keep `read_timestamp_text` are call this function in `cast_from_string`

Yes, these codes are duplicated. And I will fix the bug in read_timestamp_text and next to resolve this.

TCeason:
All tests have been fixed. Do the next thing.

TCeason:
TODO: Need add more test to cover this pr.

src/common/io/src/buffer/buffer_read_datetime_ext.rs
```diff
@@ -61,48 +70,211 @@ where R: BufferRead
...
+        west_tz: bool,
+        calc_offset: impl Fn(i64, i64, &DateTime<Tz>) -> Result<DateTime<Tz>>,
+    ) -> Result<DateTime<Tz>> {
+        if self.keep_read(buf, |f| (b'0'..=b'9').contains(&f))? != 2 {
+            return Err(ErrorCode::BadBytes(
+                "err with parse timezone, format like this:[+08:00]",
+            ));
+        }
+        let hour_offset: i32 = lexical_core::FromLexical::from_lexical(buf.as_slice()).unwrap();
+        if (0..15).contains(&hour_offset) {
```
youngsofun, line 242: ```(0..15).contains(&hour_offset)```  is slower then cmp?

src/common/io/src/buffer/buffer_read_datetime_ext.rs
```diff
@@ -61,48 +70,211 @@ where R: BufferRead
...
+            buf.clear();
+            self.ignore_byte(b':')?;
+            if self.keep_read(buf, |f| (b'0'..=b'9').contains(&f))? != 2 {
```
youngsofun, line 245: contains ->  cmp ?

src/common/io/src/buffer/buffer_read_datetime_ext.rs
TCeason: `cargo clippy` suggests using contains.

TCeason: > `(0..15).contains(&hour_offset)` is slower then cmp?

you mean this?

```
if (0..15).contains(&hour_offset) { -> if hour_offset >=0 && hour_offset < 15 {
```

youngsofun: yes

youngsofun (approved):

src/common/io/src/buffer/buffer_read_datetime_ext.rs
youngsofun: ok, maybe contains have been optimized for this

BohuTANG:
Conflicting files
tests/logictest/suites/base/03_dml/03_0016_insert_into_values_v2

src/common/io/src/buffer/buffer_read_datetime_ext.rs
```diff
@@ -61,48 +70,211 @@ where R: BufferRead
...
+                        if size > 0 && size < 3 && self.ignore_byte(b':')? {
+                            buf.clear();
+                            let size = self.keep_read(&mut buf, |f| (b'0'..=b'9').contains(&f))?;
+                            if size == 0 {
+                                return Err(ErrorCode::BadBytes(
+                                    "err with parse time part. Format like this:[03:00:00]",
+                                ));
+                            }
+                            let sec: u32 =
+                                lexical_core::FromLexical::from_lexical(buf.as_slice()).unwrap();
```
b41sh, line 132: `hour`, `min`, and `sec` have the same structure, it looks like there is some duplicate code, maybe we can handle it with a loop, or write a function.

TCeason: I think we can not use loop. Because I need get hour , min and sec. These are different variable.

And if write a closer function, can not add self.xx to the closer func.

I try to write a closer func but need more match arm to avoid some err format.

```rust
let get_time = |buf: &mut Vec<u8>, size| {
                if size > 0 && size < 3 {
                    Ok(lexical_core::FromLexical::from_lexical(buf.as_slice()).unwrap())
                } else {
                    Err(ErrorCode::BadBytes(
                        "err with parse time part. Format like this:[03:00:00]",
                    ))
                }
            };
            let mut buf = Vec::with_capacity(2);
            let size = self.keep_read(&mut buf, |f| (b'0'..=b'9').contains(&f))?;
            match get_time(&mut buf, size) {
                Ok(hour) => {
                    if self.ignore_byte(b':')? {
                        buf.clear();
                        let size = self.keep_read(&mut buf, |f| (b'0'..=b'9').contains(&f))?;
                        match get_time(&mut buf, size) {
                            Ok(min) => {
                                if self.ignore_byte(b':')? {
                                    buf.clear();
                                    let size = self.keep_read(&mut buf, |f| (b'0'..=b'9').contains(&f))?;
                                    match get_time(&mut buf, size) {
                                        Ok(sec) => dt = tz.from_local_datetime(&d.and_hms(hour, min, sec)).unwrap(),
                                        Err(e) => return Err(e)
                                    }
                                } else {
                                    return Err(ErrorCode::BadBytes(
                                        "err with parse time part. Format like this:[03:00:00]",
                                    ));
                                }
                            },
                            Err(e) => return Err(e)
                        }
                    } else {
                        return Err(ErrorCode::BadBytes(
                            "err with parse time part. Format like this:[03:00:00]",
                        ));
                    }
                },
                Err(e) => return Err(e)
            }

```

src/common/io/src/buffer/buffer_read_datetime_ext.rs
```diff
@@ -61,48 +70,199 @@ where R: BufferRead
...
+                } else {
+                    Err(ErrorCode::BadBytes(
+                        "err with parse time part. Format like this:[03:00:00]",
+                    ))
+                }
+            };
+            let mut buf = Vec::with_capacity(2);
             let size = self.keep_read(&mut buf, |f| (b'0'..=b'9').contains(&f))?;
-            let scales: i64 = lexical_core::FromLexical::from_lexical(buf.as_slice()).unwrap();
+            match get_time(&mut buf, size) {
```
sundy-li, line 116: why not use `get_time(&mut buf, size).and_then ....`

This part of code looks very ugly.

TCeason: Yes, and_then can simplify match arm. 

BTW, could we have other ideas to simplify this?

```
match res {
    Ok(d) => dt = d,
    Err(e) => return Err(e),
}
```

src/common/io/src/buffer/buffer_read_datetime_ext.rs
b41sh: how about put `hour`, `min`, and `sec` in a vector? like this
```rust
let mut times Vec::with_capacity(3);
loop {
    let size = self.keep_read(&mut buf, |f| (b'0'..=b'9').contains(&f))?;
    match get_time(&mut buf, size) {
        Ok(time) => times.push(time),
        Err(e) => return e,
    }
    if times.len() == 3 {
        break;
    }
    self.must_ignore_byte(b':')?;
}
```

TCeason: That's a pretty good idea. I will do some test.

src/common/io/src/buffer/buffer_read_datetime_ext.rs
```diff
@@ -61,48 +70,179 @@ where R: BufferRead
...
+    ) -> Result<DateTime<Tz>> {
+        if self.keep_read(buf, |f| (b'0'..=b'9').contains(&f))? != 2 {
+            return Err(ErrorCode::BadBytes(
+                "err with parse timezone, format like this:[+08:00]",
+            ));
+        }
+        let hour_offset: i32 = lexical_core::FromLexical::from_lexical(buf.as_slice()).unwrap();
+        if (0..15).contains(&hour_offset) {
+            buf.clear();
+            self.ignore_byte(b':')?;
```
b41sh, line 212: `must_ignore_byte` ?

src/common/io/src/buffer/buffer_read_datetime_ext.rs
```diff
@@ -61,48 +70,179 @@ where R: BufferRead
...
+                    "err with parse timezone, format like this:[+08:00]",
+                ));
+            }
+            let minute_offset: i32 =
+                lexical_core::FromLexical::from_lexical(buf.as_slice()).unwrap();
+            // max utc: 14:00, min utc: 00:00
+            if (hour_offset == 14 && minute_offset == 0)
+                || ((0..60).contains(&minute_offset) && hour_offset < 14)
+            {
+                if dt.year() < 1970 {
```
b41sh, line 224: Can't there be an offset before 1970? The minimum supported value for `timestamp` is 1000

TCeason: Oh, It can be. Forget to modify it.

src/common/io/src/buffer/buffer_read_datetime_ext.rs
TCeason: maybe not use must. In keep_read()

in 
```
if available.is_empty() || !f(available[0]) {
                break;
            }
```

if the byte is not [0,9) the loop will break and the buf is empty. and size != 2.

if the byte is in [0,9),

```
if self.keep_read(buf, |f| (b'0'..=b'9').contains(&f))? != 2 {
            return Err(ErrorCode::BadBytes(
                "err with parse timezone, format like this:[+08:00]",
            ));
        }
        let hour_offset: i32 = lexical_core::FromLexical::from_lexical(buf.as_slice()).unwrap();
```

It will return err in there . Because the size if not 3 or more.

b41sh:
LGTM

src/common/io/src/buffer/buffer_read_datetime_ext.rs
```diff
@@ -61,48 +70,182 @@ where R: BufferRead
...
+    fn parse_time_offset(
+        &mut self,
+        tz: &Tz,
+        buf: &mut Vec<u8>,
+        dt: &DateTime<Tz>,
+        west_tz: bool,
+        calc_offset: impl Fn(i64, i64, &DateTime<Tz>) -> Result<DateTime<Tz>>,
+    ) -> Result<DateTime<Tz>> {
+        if self.keep_read(buf, |f| (b'0'..=b'9').contains(&f))? != 2 {
+            // +0800 will err in there
```
TCeason, line 206: Comment added.@b41sh

b41sh: ok

b41sh (approved):

BohuTANG:
cc @soyeric128 , new datetime format supporting :)

PR #7146, opened 2022-08-16
fix(expr): fold more constant

andylokandy:

## Summary

Fold scalar expression regardless of domain calculation.


leiysky (approved):

src/common/expression/src/evaluator.rs
```diff
@@ -494,33 +513,51 @@ impl<'a> ConstantFolder<'a> {
...
+                            scalar,
+                        },
+                        func_domain,
+                    );
+                }
+
+                if all_args_is_scalar {
+                    let chunk = Chunk::empty();
+                    let evaluator = Evaluator::new(&chunk, self.context.clone());
+                    if let Ok(Value::Scalar(scalar)) = evaluator.run(&func_expr) {
```
sundy-li, line 551: `Value::Column(c) if c.len() == 1` should be cared.

There will be functions that accept more args than 3 like `elt`.

src/common/expression/src/evaluator.rs
```diff
@@ -108,13 +118,18 @@ impl<'a> Evaluator<'a> {
...
         #[cfg(debug_assertions)]
         if result.is_err() {
-            assert_eq!(
-                ConstantFolder::new(&self.input_columns.domains())
-                    .fold(expr)
-                    .1,
-                None,
-                "domain calculation should not return any domain for expressions that are possible to fail"
-            );
+            static RECURSING: Mutex<bool> = Mutex::new(false);
```
BohuTANG, line 121: `use parking_lot::Mutex;`
No need `unwarp` here

andylokandy: This is for debugging only, so I think it won't bother adding an extra dependency.

sundy-li (approved):

PR #7174, opened 2022-08-18
chore(query): join-key use compare_coercion instead of merge_types

### LINKED ISSUE #7164: Can't construct type from Float32(Float32) and Int64(Int64).
### Discussed in https://github.com/datafuselabs/databend/discussions/7162

<div type='discussions-op-text'>

<sup>Originally posted by **hanyisong** August 17, 2022</sup>

Is this a problem with my operation, or a feature to be improved.

```sql
CREATE DATABASE databend0;
USE databend0;
CREATE TABLE t0(c0 BIGINT NULL, c1 DOUBLE NULL DEFAULT(true));
CREATE TABLE t1(c0 FLOAT NULL);
SELECT SUM(count) FROM (SELECT ((false IS NOT NULL AND false) ::INT64)as count FROM t1 NATURAL LEFT JOIN t0) as res;

Code: 1010, displayText = Can't construct type from Float32(Float32) and Int64(Int64).
```</div>

### PR DESCRIPTION
**Pull request by sundy-li**

## Summary

Summary about this PR

Fixes #7164

### CODE REVIEW

**Review by xudong963 — APPROVED** [review]
LGTM

Btw, I think https://github.com/datafuselabs/databend/blob/main/src/query/service/src/sql/planner/binder/select.rs#L295 also needs to use `compare_coercion`.

Will you change it in this pr by hand or I will open a separate pr later?

**Comment by sundy-li** [comment]
> also needs to use compare_coercion.

Since they are not in relation of comparison op, I think we should keep it as `merge_types`.

**Comment by xudong963** [comment]
> > also needs to use compare_coercion.
> 
> Since they are not in relation of comparison op, I think we should keep it as `merge_types`.

Databend

```sql
mysql> CREATE TABLE t0(c0 BIGINT NULL, c1 DOUBLE NULL DEFAULT(true));

mysql> CREATE TABLE t1(c0 FLOAT NULL);

mysql> select c0 from t0 union select c0 from t1;
ERROR 1105 (HY000): Code: 1010, displayText = Can't construct type from Int64(Int64) and Float32(Float32).
mysql> select c0 from t0 intersect select c0 from t1;
ERROR 1105 (HY000): Code: 1010, displayText = Can't construct type from Int64(Int64) and Float32(Float32).
```

pg
```sql
postgres=# CREATE TABLE t0(c0 BIGINT NULL, c1 Float NULL);
CREATE TABLE
postgres=# CREATE TABLE t1(c0 FLOAT NULL);
CREATE TABLE
postgres=# select c0 from t0 union select c0 from t1;
 c0
----
(0 rows)
```

leiysky (approved):

PR #7175, opened 2022-08-18
feat(planner): support settings set sql_dialect

### LINKED ISSUE #7131: Feature: Add Setting `sql_dialect` to control the dialect of parser
**Summary**

Description for this feature.

Releated #7130  #7071

b41sh:

## Summary

Support settings set `sql_dialect`

Fixes #7131


src/common/settings/src/lib.rs
```diff
@@ -220,6 +239,17 @@ impl Settings {
...
+            },
+            SettingValue {
+                default_value: DataValue::String("PostgreSQL".as_bytes().to_vec()),
+                user_setting: UserSetting::create(
+                    "sql_dialect",
+                    DataValue::String("PostgreSQL".as_bytes().to_vec()),
+                ),
+                level: ScopeLevel::Session,
+                desc: "SQL dialect, support \"PostgreSQL\" and \"MySQL\", default value: \"PostgreSQL\"",
+                possible_values: Some(vec!["PostgreSQL", "MySQL"]),
```
sundy-li, line 252: 👍

sundy-li:
Need a test file in logictest.

For example:
```
set sql_dialect = 'xxxx';

select "33";
```

sundy-li (approved):

xudong963 (approved):

leiysky:
We should provide a default SQL dialect for databend.

Before persisting any of the SQL query, the SQL text must be normalized into default dialect. Otherwise we will encounter bugs if user is creating a view with MySQL dialect but query with default dialect.

For example, user is creating a view with `create view v as select "A" from t` in MySQL dialect, then the token `"A"` is ambigious, which is impossible to be fixed.

b41sh:
> @b41sh
> 
> We should provide a default SQL dialect for databend.
> 
> Before persisting any of the SQL query, the SQL text must be normalized into default dialect. Otherwise we will encounter bugs if user is creating a view with MySQL dialect but query with default dialect.
> 
> For example, user is creating a view with `create view v as select "A" from t` in MySQL dialect, then the token `"A"` is ambigious, which is impossible to be fixed.

the default SQL dialect is `PostgreSQL`, can we rewrite the view query to `PostgreSQL`?

b41sh:
@leiysky the subquery of view is formatted as `PostgreSQL` dialect. When querying, whether it is `MySQL` or `PostgreSQL` dialect, we can get the correct results.

PR #7177, opened 2022-08-18
fix: statements `desc share` and `show shares` may have resultset

dantengsky:

## Summary

- add `Plan::DescShare` and `Plan::ShowShares` to the may-have-non-empty-result-set list
- add a simple stateless test case
  since there is a non-deterministic column in the test resultset, and stateless-test supports regex filter
  
  NOTE: 

  tried adding a case :
     `create share a` and then `desc share a`, expecting descriptions of share a.
  but failed. 
  A simple investigation shows that an empty resultset is returned. Not sure if this is expected. thus this case is excluded.

Fixes #issue


Xuanwo (approved):

Xuanwo:
Is `has_result_set` the same with `plan.schema()` is not empty?

lichuang (approved):

PR #7186, opened 2022-08-18
fix(parser): fix parse float with E failed

### LINKED ISSUE #7156: bug: float exponent form does not support E
### Search before asking

### Version

8.0.26-v0.7.159-nightly-a7f2c58-simd(rust-1.64.0-nightly-2022-08-10T16:29:03.797735015Z)

### What's Wrong?

My program automatically generates a capital E
```sql
CREATE TABLE t0(c0 FLOAT NULL DEFAULT(true), c1 BOOL NULL DEFAULT(811681875));
INSERT INTO t0(c0, c1) VALUES (8.11681875E8, -568901493);

 Code: 1010, displayText = Expression size not match schema num of cols.
```

### How to Reproduce?

_No response_

b41sh:

## Summary

fix parse float with `E` failed

Fixes #7156


src/query/ast/src/parser/token.rs
```diff
@@ -125,8 +125,8 @@ pub enum TokenKind {
     #[regex(r"[0-9]+")]
     LiteralInteger,
 
-    #[regex(r"[0-9]+e[+-]?[0-9]+")]
-    #[regex(r"([0-9]*\.[0-9]+(e[+-]?[0-9]+)?)|([0-9]+\.[0-9]*(e[+-]?[0-9]+)?)")]
+    #[regex(r"[0-9]+[eE][+-]?[0-9]+")]
+    #[regex(r"([0-9]*\.[0-9]+([eE][+-]?[0-9]+)?)|([0-9]+\.[0-9]*([eE][+-]?[0-9]+)?)")]
```
Xuanwo, line 129: Is is a good idea to use https://docs.rs/nom/latest/nom/number/complete/fn.float.html directly?

b41sh: use `nom::number::complete::float` to parse `LiteralFloat` token?

sundy-li: How to use it into `derive`?

b41sh: But other Literals are implemented with regular expressions, should they be replaced by `nom`?

sundy-li: cc @andylokandy

andylokandy: tokenizer usually use regular expression instead of parser, because the formaer is faster and must terminate. nom is a parser, so we can't use it here.

BohuTANG (approved):

PR #7190, opened 2022-08-19
refactor(meta): reduce long span info

drmingdrmer:

##### refactor(meta): reduce long span info

Remove long info out of span.

Update openraft to 0.7.2-alpha.1, with normalized logging.

## Related Issues

### DIFF
--- a/src/binaries/Cargo.toml
+++ b/src/binaries/Cargo.toml
@@ -55,7 +55,7 @@ tonic = "0.7.2"
 tracing = "0.1.35"
 url = "2.2.2"
 
-openraft = { git = "https://github.com/datafuselabs/openraft", rev = "v0.7.1" }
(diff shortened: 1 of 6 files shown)


src/binaries/Cargo.toml
```diff
@@ -55,7 +55,7 @@ tonic = "0.7.2"
 tracing = "0.1.35"
 url = "2.2.2"
 
-openraft = { git = "https://github.com/datafuselabs/openraft", rev = "v0.7.1" }
+openraft = { git = "https://github.com/drmingdrmer/openraft", rev = "v0.7.2-alpha.1" }
```
Xuanwo, line 58: Is this expected?

drmingdrmer: Yes. We need to update openraft frequently to some alpha version, which should not pollute the official repo tags.

Xuanwo (approved):

PR #7199, opened 2022-08-19
feat(planner): support `default_expr` when creating table

### LINKED ISSUE #7139: bug: default value is not equal with expected type
### Search before asking

### Version

v0.7.169-nightly

### What's Wrong?

I think this statement need to get an error, it expected DEFAULT expression to have type int, but 'false' has type bool

`CREATE TABLE t (a INT DEFAULT false);`

SDwRgGFitZ

### How to Reproduce?

_No response_

xudong963:

## Summary

Summary about this PR

Fixes #7139

### DIFF
--- a/src/query/service/src/interpreters/interpreter_table_show_create.rs
+++ b/src/query/service/src/interpreters/interpreter_table_show_create.rs
@@ -25,8 +25,8 @@ use tracing::debug;
 use crate::interpreters::Interpreter;
 use crate::sessions::QueryContext;
 use crate::sessions::TableContext;
+use crate::sql::executor::PhysicalScalar;
 use crate::sql::is_internal_opt_key;
-use crate::sql::PlanParser;
 
 pub struct ShowCreateTableInterpreter {
     ctx: Arc<QueryContext>,
@@ -71,8 +71,8 @@ impl Interpreter for ShowCreateTableInterpreter {
             for (idx, field) in schema.fields().iter().enumerate() {
                 let default_expr = match field.default_expr() {
                     Some(expr) => {
-                        let expression = PlanParser::parse_expr(expr)?;
-                        format!(" DEFAULT {}", expression.column_name())
+                        let expression: PhysicalScalar = serde_json::from_str(expr)?;
+                        format!(" DEFAULT {expression}")
                     }
                     None => "".to_string(),
                 };
--- a/src/query/service/src/sql/planner/binder/ddl/table.rs
+++ b/src/query/service/src/sql/planner/binder/ddl/table.rs
@@ -24,6 +24,7 @@ use common_ast::parser::tokenize_sql;
 use common_ast::walk_expr_mut;
 use common_ast::Backtrace;
 use common_ast::Dialect;
+use common_datavalues::type_coercion::compare_coercion;
 use common_datavalues::DataField;
 use common_datavalues::DataSchemaRef;
 use common_datavalues::DataSchemaRefExt;
@@ -48,8 +49,10 @@ use crate::sql::optimizer::OptimizerContext;
 use crate::sql::planner::semantic::normalize_identifier;
 use crate::sql::planner::semantic::IdentifierNormalizer;
 use crate::sql::plans::create_table_v2::CreateTablePlanV2;
+use crate::sql::plans::CastExpr;
 use crate::sql::plans::Plan;
 use crate::sql::plans::RewriteKind;
+use crate::sql::plans::Scalar;
 use crate::sql::BindContext;
 use crate::sql::ColumnBinding;
 use crate::sql::ScalarExpr;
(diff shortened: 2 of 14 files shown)


tests/logictest/suites/crdb/default
```diff
@@ -1,97 +1,81 @@
...
+CREATE TABLE t ( a INT DEFAULT 42,  b TIMESTAMP DEFAULT now(),  c FLOAT DEFAULT 1.0,  d DATE DEFAULT now());
 
--- statement ok
--- CREATE TABLE null_default (ts TIMESTAMP NULL DEFAULT NULL);
+statement ok
+INSERT INTO t(a,b,c,d) VALUES ();
 
--- statement ok
--- drop table if exists t;
+statement query
```
ZeaLoVe, line 10: need fix here, query can not without query options

drmingdrmer:
- `main` branch is force pushed, update your working branch please: #7261

tests/logictest/suites/base/06_show/06_0009_show_shares
```diff
@@ -1,64 +1,63 @@
...
 
-statement query RTTTTTT
-show shares;
+-- statement query RTTTTTT
+-- show shares;
 
-----
-$DATE_IN_SHARE OUTBOUND test_share show_shares test_tenant x comment
+-- ----
+-- $DATE_IN_SHARE OUTBOUND test_share show_shares test_tenant x comment
```
xudong963, line 35: I think `$DATE_IN_SHARE` is meaningless, ideally, it should be

```sql
mysql> show shares;
+--------------------------------+----------+------------+---------------+-------+------+---------+
| Created_on                     | Kind     | Name       | Database_name | From  | To   | Comment |
+--------------------------------+----------+------------+---------------+-------+------+---------+
| 2022-08-24 01:46:04.191485 UTC | OUTBOUND | test_share | show_shares   | admin | x    | comment |
+--------------------------------+----------+------------+---------------+-------+------+---------+
```

sundy-li: It's regexp match because the result may be different.

xudong963: got it

ZeaLoVe: > I think `$DATE_IN_SHARE` is meaningless, ideally, it should be

There are two kinds of Date regex used in stateless test, It is really meaningless. You can change in dir logictest/regex_type.py, all ref test suites need changing at the same time.

src/query/service/src/sql/planner/binder/ddl/table.rs
```diff
@@ -757,10 +761,22 @@ impl<'a> Binder {
...
+                            if !expr_type.eq(&data_type) {
+                                expr = Scalar::CastExpr(CastExpr {
+                                    argument: Box::new(expr),
+                                    from_type: Box::new(expr_type),
+                                    target_type: Box::new(data_type),
+                                })
+                            }
+                            let mut builder = PhysicalScalarBuilder;
+                            let serializable_expr = builder.build(&expr)?;
+                            Some(serde_json::to_string(&serializable_expr)?)
```
leiysky, line 779: We should serialize the default expression into SQL string to prevent compatibility issues.

tests/logictest/suites/base/06_show/06_0001_show_create_table
```diff
@@ -13,25 +13,24 @@ SHOW CREATE TABLE `test`.`a`;
 ----
 a CREATE TABLE `a` (
   `a` BIGINT,
-  `b` INT DEFAULT 3,
-  `c` VARCHAR DEFAULT 'x',
+  `b` INT DEFAULT CAST(3 AS INT),
```
leiysky, line 16: Shall we ignore the explicit cast?

src/query/service/src/interpreters/interpreter_table_show_create.rs
```diff
@@ -71,8 +71,8 @@ impl Interpreter for ShowCreateTableInterpreter {
             for (idx, field) in schema.fields().iter().enumerate() {
                 let default_expr = match field.default_expr() {
                     Some(expr) => {
-                        let expression = PlanParser::parse_expr(expr)?;
-                        format!(" DEFAULT {}", expression.column_name())
+                        let expression: PhysicalScalar = serde_json::from_str(expr)?;
```
leiysky, line 74: Is it possible to use `PhysicalScalar` as `DataFiled::default_expr`

xudong963: I have tried it but seems to need a nontrivial change

tests/logictest/suites/base/06_show/06_0001_show_create_table
xudong963: Maybe the explicit cast is meaningful and will give user a hint (for the following case, there is precision truncation)

```sql
mysql> create table t(b int, a int default 0.1)
    -> ;

mysql> show create table t;
+-------+--------------------------------------------------------------------------------+
| Table | Create Table                                                                   |
+-------+--------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `b` INT,
  `a` INT DEFAULT CAST(0.1 AS INT)
) ENGINE=FUSE |
+-------+--------------------------------------------------------------------------------+
1 row in set (0.02 sec)

mysql> select * from t;
+------+------+
| b    | a    |
+------+------+
|    3 |    0 |
+------+------+
1 row in set (0.04 sec)
Read 1 rows, 8.00 B in 0.007 sec., 135.24 rows/sec., 1.06 KiB/sec.

```

leiysky (approved):

PR #7215, opened 2022-08-20
feat(planner): support explain ast

### LINKED ISSUE #7063: Feature: explain AST/syntax statement

b41sh:

## Summary

Support `explain ast`.
It is not easy to implement `CockroachDB` style explain format with `pretty`, so I still use `FormatTreeNode`.

Fixes #7063


Xuanwo:
It seems like we have to maintian three kinds of `Display` for `stmt`. Do we have better choices?

b41sh:
> It seems like we have to maintian three kinds of `Display` for `stmt`. Do we have better choices?

It is indeed a problem. If a `statement` is changed, we need to modify three places.

drmingdrmer:
- `main` branch is force pushed, update your working branch please: #7261

drmingdrmer:
`main` branch is force pushed(**again** `:(`, to cleanup big object in history), update your working branch please: [To everybody: main branch is force pushed. #7263](https://github.com/datafuselabs/databend/issues/7263)

Xuanwo:
> It is indeed a problem. If a `statement` is changed, we need to modify three places.

Is it possible to automatically convert the statement with the same rules instead of writing similar code three times?

sundy-li:
I did not find better solution for this.  cc @leiysky  do you have any suggestions?

b41sh:
> > It is indeed a problem. If a `statement` is changed, we need to modify three places.
> 
> Is it possible to automatically convert the statement with the same rules instead of writing similar code three times?

It is not easy to write a transfer rule, because every statement has different fields.
`Clickhouse` does this by adding a `formatImpl` function to each statement.
https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/IAST.h#L231
Although there are many statements, they are relatively few changes, we can check and modify them regularly.

src/query/ast/src/ast/format/ast/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use crate::ast::ExplainKind;
+use crate::ast::FormatTreeNode;
+use crate::ast::Statement;
+
+pub fn format_statement(stmt: Statement) -> Result<String> {
+    format_stmt(stmt).format_pretty()
+}
+
+fn format_stmt(stmt: Statement) -> FormatTreeNode<String> {
```
sundy-li, line 37: can we use a common visitor to apply the `format` ?

b41sh: ok

sundy-li (approved):

b41sh:
I implemented `trait Visitor`, if there are new `Statement` in the future, we only need to add the corresponding visit method. PTAL @Xuanwo

Xuanwo (approved):
Great!

BohuTANG:
cc @soyeric128 for documentation :)

```
explain ast
```

PR #7219, opened 2022-08-21
feat(functions-v2): migrate bin, oct, hex, and unhex.

RinChanNOWWW:

## Summary

Migrate functions `bin`, `oct`, `hex` and `unhex`.

Related #6766


src/query/functions-v2/src/scalars/string.rs
```diff
@@ -311,6 +342,144 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    ];
+
+    for dtype in all_numerics_types {
+        with_unsigned_number_mapped_type!(T, match dtype {
+            DataType::T => {
+                registry.register_1_arg::<NumberType<T>, StringType, _, _>(
+                    "bin",
+                    FunctionProperty::default(),
+                    |_| None,
+                    |val| format!("{:b}", val as u64).as_bytes().to_vec(),
```
sundy-li, line 367: maybe we just need to implement `u64` version. 

For other unsigned numbers, the evaluator will cast them into u64 in `type_checker`.

RinChanNOWWW: How about signed integer type and float type?

sundy-li: signed integer types  --> i64
float types -> f64

We have a special case for  `arithmetics` because they are commonly used.

For other numeric arguments, it's ok just to declare the largest type.

RinChanNOWWW: Understood.

src/query/functions-v2/src/scalars/string.rs
```diff
@@ -26,6 +27,32 @@ use common_expression::NumberDomain;
 use common_expression::Value;
 use common_expression::ValueRef;
 
+trait OctString {
```
andylokandy, line 30: maybe defining a trait here is unnescessay because it's used only once.

src/query/functions-v2/src/scalars/string.rs
```diff
@@ -381,3 +442,90 @@ fn vectorize_string_to_string_2_arg(
         }
     }
 }
+
+fn register_bin(registry: &mut FunctionRegistry) {
+    registry.register_1_arg::<NumberType<u64>, StringType, _, _>(
+        "bin",
+        FunctionProperty::default(),
+        |_| None,
+        |val| format!("{:b}", val).as_bytes().to_vec(),
```
andylokandy, line 451: you can directly `write!()` into writer.data, using register_passthrough_nullable_1_arg along with vectorize_with_writer_1_arg.

src/query/functions-v2/src/scalars/string.rs
```diff
@@ -381,3 +442,90 @@ fn vectorize_string_to_string_2_arg(
...
+    registry.register_passthrough_nullable_1_arg::<StringType, StringType, _, _>(
+        "hex",
+        FunctionProperty::default(),
+        |_| None,
+        vectorize_string_to_string(
+            |col| col.data.len() * 2,
+            |val, writer| {
+                let mut buffer = vec![0u8; val.len() * 2];
+                let buff = &mut buffer[0..val.len() * 2];
+                let _ = hex::encode_to_slice(val, buff);
```
andylokandy, line 524: you can encode to writer.data.

andylokandy (approved):

src/query/functions-v2/src/scalars/string.rs
```diff
@@ -443,74 +433,94 @@ fn vectorize_string_to_string_2_arg(
...
+        }),
     );
-    registry.register_1_arg::<NumberType<i64>, StringType, _, _>(
+    registry.register_passthrough_nullable_1_arg::<NumberType<i64>, StringType, _, _>(
         "oct",
         FunctionProperty::default(),
         |_| None,
-        |val| val.oct_string().as_bytes().to_vec(),
+        vectorize_with_writer_1_arg::<NumberType<i64>, StringType>(|val, writer| {
+            write_and_commit!(writer, "{}", oct_i64(val))
```
sundy-li, line 495: Can we remove `oct_i64` as `bin` function did ? It creates temporary string memory.

src/query/functions-v2/src/scalars/string.rs
```diff
@@ -443,74 +433,94 @@ fn vectorize_string_to_string_2_arg(
...
         |_| None,
-        |val| {
-            let val = val as f64;
-            let ret = if val.ge(&0f64) {
-                format!(
-                    "{:b}",
-                    val.max(i64::MIN as f64).min(i64::MAX as f64).round() as i64
-                )
+        vectorize_with_writer_1_arg::<NumberType<f64>, StringType>(|val, writer| {
+            if (val as f64).ge(&0f64) {
```
sundy-li, line 470: val is already f64.

src/query/functions-v2/src/scalars/string.rs
RinChanNOWWW: Sure.

src/query/functions-v2/src/scalars/string.rs
RinChanNOWWW: Oh, I didn't note this.

sundy-li (approved):

sundy-li:
BTW `oct` should not output negative results by default.

We should add [`conv` function ](https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_conv)

- OCT is aliased to [CONV(N,10,8)](https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_conv).
- bin is aliased to [CONV(N,10,2)](https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_conv)
- hex is aliased to [CONV(N,10,16)](https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_conv)

[example results](https://www.db-fiddle.com/f/eiMLou2QfijZtw9iiQtYt4/1)

MySQL-related codes:

https://github.com/mysql/mysql-server/blob/8d8c986e5716e38cb776b627a8eee9e92241b4ce/sql/item_strfunc.cc#L2860-L2925

PR #7248, opened 2022-08-23
chore: fix `13_0000_prepare.sh` script

xudong963:

## Summary
https://github.com/datafuselabs/databend/runs/7969999341?check_suite_focus=true

Fixes #issue


sundy-li (approved):

sundy-li:
@ZeaLoVe  Do you think we should move tpch into stateful tests like mini-hits?

BohuTANG (approved):

ZeaLoVe:
> @ZeaLoVe Do you think we should move tpch into stateful tests like mini-hits?

LGTM, but I think it better move into sqllogictest, but I do not know how to load data in this scenario. Maybe change to copy into?

PR #7260, opened 2022-08-24
ci: Add file size watcher in CI

Xuanwo:

## Summary

ci: Add file szie watcher in CI

### DIFF
--- a/.github/actions/check/action.yml
+++ b/.github/actions/check/action.yml
@@ -12,6 +12,9 @@ runs:
       with:
         bypass_env_vars: RUSTFLAGS,RUST_LOG,GITHUB_TOKEN
 
+    - uses: actions/setup-node@v3
+      with:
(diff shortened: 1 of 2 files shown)


BohuTANG (approved):
:100:

Xuanwo:
Node.js sucks...

Xuanwo:
Cc @everpcpc, `npx` will install binaries to `tmp` and than link back to `$PATH` which can't run based on current setup. Any ideas to fix this?

everpcpc:
> Cc @everpcpc, `npx` will install binaries to `tmp` and than link back to `$PATH` which can't work based on current setup. Any ideas to fix this?

Maybe It's related to self-hosted runner, maybe we can separate this check in a new workflow?

Xuanwo:
This PR doesn't work as expected.

PR #7291, opened 2022-08-25
fix(planner): fix union data type match

xudong963:

## Summary

Unify union data type for schema and block

Fixes https://github.com/datafuselabs/databend/issues/7292


src/query/service/src/sql/planner/binder/select.rs
```diff
@@ -335,7 +335,11 @@ impl<'a> Binder {
         right_expr: SExpr,
         distinct: bool,
     ) -> Result<SExpr> {
-        let union_plan = UnionAll {};
+        let mut column2type = HashMap::with_capacity(bind_context.columns.len());
+        for column_binding in bind_context.columns.iter() {
+            column2type.insert(column_binding.index, column_binding.data_type.clone());
+        }
+        let union_plan = UnionAll { column2type };
```
leiysky, line 342: It's better to add new column entries to metadata instead of modifying the data type here.

xudong963: Here I just store <column_idx, column_data_type>, then used in building physical plan, not modifying the data type.

leiysky: Then the data type can be inferred from input schema?

xudong963: Then we'll generate new schema as union's final schema from union's left child's output schema, the new schema's fields' data type will use data type of `column2type`.

xudong963: Make sense, will try using `EvalScalar` to cast type.

xudong963:
```
failures:

---- grpc::metasrv_grpc_kv_api::test_metasrv_kv_api stdout ----
thread 'grpc::metasrv_grpc_kv_api::test_metasrv_kv_api' panicked at 'got unexpired', /workspace/src/meta/api/src/kv_api_test_suite.rs:331:13
stack backtrace:
```
https://github.com/datafuselabs/databend/runs/8053732601?check_suite_focus=true 

cc @drmingdrmer

xudong963:
@mergify update

tests/logictest/suites/base/15_query/union.test
```diff
@@ -99,3 +99,10 @@ DROP TABLE data2015;
 
 statement ok
 DROP TABLE data2016;
+
+statement query TTTTI
+SELECT 'Кирилл' as a, 'Müller' as b, '我是谁' as c, 'ASCII' as d, 2 as id UNION SELECT NULL as a, NULL as b, NULL as c, NULL as d, 1 as id order by id;
```
sundy-li, line 104: What's the result of :

```
SELECT 'Кирилл' as a, 'Müller' as b, '我是谁' as c, 'ASCII' as d, 2 as id UNION SELECT NULL, NULL, NULL as c, NULL as d, 1 order by id;
```

xudong963: same as `
SELECT 'Кирилл' as a, 'Müller' as b, '我是谁' as c, 'ASCII' as d, 2 as id UNION SELECT NULL as a, NULL as b, NULL as c, NULL as d, 1 as id order by id;`

drmingdrmer:
> ```
> failures:
> 
> ---- grpc::metasrv_grpc_kv_api::test_metasrv_kv_api stdout ----
> thread 'grpc::metasrv_grpc_kv_api::test_metasrv_kv_api' panicked at 'got unexpired', /workspace/src/meta/api/src/kv_api_test_suite.rs:331:13
> stack backtrace:
> ```
> 
> https://github.com/datafuselabs/databend/runs/8053732601?check_suite_focus=true
> 
> cc @drmingdrmer

It looks like the expiration time of `1 second` is too small in a CI, which makes this test flaky.
I'll extend the expiration time for the test in the next PR!

Thanks man!

sundy-li (approved):

leiysky (approved):

PR #7300, opened 2022-08-26
refactor(fuse): split DataBlock into multiple pages if necessary during writing

dantengsky:

## Summary

- During writing `DataBlock` to object storage, split `DataBlock` into multiple pages if row numbers of the data block exceed the given threshold.
  so that
  - data could be deserialized into smaller chunks and passed to computing components early
  - peak memory usage is likely to be lesser

A simple test with local fs backend verified the above assumptions:

 TODO: desc of test scenario

<del>
- multi pages (oversize_threshold:0, max row per page, 10_000 rows)

  ~~~
  Time (mean ± σ):     849.3 ms ±  26.0 ms    [User: 2.2 ms, System: 1.3 ms]
  Range (min … max):   795.2 ms … 927.4 ms    100 runs
  ~~~

- single page(oversize_threshold:0)

  ~~~
   Time (mean ± σ):      1.160 s ±  0.131 s    [User: 0.003 s, System: 0.001 s]
   Range (min … max):    0.908 s …  1.342 s    100 runs
  ~~~
</del>

NOTE: 
- Reading of column pages is still in one-go （to better utilize the high bandwidth of object storage service)
- need further perf (with s3 backend and other test data set)

Fixes #issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,7 +150,6 @@ dependencies = [
 [[package]]
 name = "arrow2"
 version = "0.12.0"
-source = "git+https://github.com/datafuse-extras/arrow2?rev=2b427cf#2b427cf4f49cde6623db391d39f32363bfe441d0"
 dependencies = [
  "ahash",
  "arrow-format",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,6 +150,7 @@ dependencies = [
 [[package]]
 name = "arrow2"
 version = "0.12.0"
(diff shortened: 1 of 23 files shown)


Xuanwo:
Notes: OpenDAL is going to release multipart support https://github.com/datafuselabs/opendal/issues/439

We can upload different parts of the same parquet at the same time.

dantengsky:
Performance does not meet expectations, withdraw this PR. Will be explored later in another PR

PR #7273, opened 2022-08-24
feat(planner): Push limit down further

TCeason:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

cherry pick from  [7127](https://github.com/datafuselabs/databend/pull/7127).

And will fix test in this pr.

Fixes https://github.com/datafuselabs/databend/issues/6865

PR #7302, opened 2022-08-26
feat(storage): add topn pruner

sundy-li:

## Summary

Summary about this PR

Add partition based topn pruner in fuse engine

Continue #7273


src/query/storages/fuse/src/pruning/pruning_executor.rs
```diff
@@ -150,15 +151,33 @@ impl BlockPruner {
...
+                let metas = joint
+                    .into_iter()
+                    .collect::<Result<Vec<_>>>()?
+                    .into_iter()
+                    .flatten();
+                Ok(metas.collect())
+            });
+        let metas = metas?;
+
+        // if there are ordering + limit clause, use topn pruner
```
TCeason, line 166: I push down the order_by and limit in 
[src/query/service/src/sql/optimizer/rule/rewrite/rule_push_down_sort_scan.rs](https://github.com/datafuselabs/databend/pull/7273/files#diff-8b117d9593a34e61fd76a5946e339731c0e5ac90530d380a6c664bb3c8292064)
```rust
    fn apply(&self, s_expr: &SExpr, state: &mut TransformState) -> Result<()> {
        let sort: Sort = s_expr.plan().clone().try_into()?;
        let child = s_expr.child(0)?;
        let mut get: LogicalGet = child.plan().clone().try_into()?;
        if get.order_by.is_none() {
            get.order_by = Some(sort.items);
        }
        if let Some(limit) = sort.limit {
            get.limit = Some(get.limit.map_or(limit, |c| cmp::max(c, limit)));
        }
        let get = SExpr::create_leaf(RelOperator::LogicalGet(get));
        state.add_result(s_expr.replace_children(vec![get]));
        Ok(())
    }
```

if only valid when ordering + limit clause, maybe I could modify the apply?

Only the in order by + limit case apply the partten push_down_sort_scan?

sundy-li: It's ok now.

1. There is only a limit clause, it'll use LimitPruner (this is an accurate pruner).
2. There is only order clause, it'll not use LimitPruner and TopN Pruner.
3. There is `limit + order` clause, it'll use TopN Pruner.

dantengsky (approved):
LGTM

src/query/service/tests/it/storages/fuse/pruning.rs
```diff
@@ -145,49 +146,56 @@ async fn test_block_pruner() -> Result<()> {
...
+    e3.order_by = vec![Expression::Sort {
+        expr: Box::new(col("b")),
+        asc: true,
+        nulls_first: false,
+        origin_expr: Box::new(col("b")),
+    }];
+    e3.limit = Some(3);
+
+    // Sort desc Limit
+    let mut e3 = Extras::default();
```
dantengsky, line 170: oh, maybe here you mean `e4`

BohuTANG (approved):

TCeason (approved):

Xuanwo:
This PR merged with failed unit tests. Should be addressed.

PR #7311, opened 2022-08-26
feat(hive):  split hive file to small partition and support read more than one rowgroups

sandflee:

## Summary
* split hive file to small partition
* support read more than one rowgroups


sandflee:
many test failues,  seems not related

TCeason:
@mergify update

dantengsky:
@sandflee hi, does this PR also solve the high system cpu problem?

src/query/storages/hive/src/hive_blocks.rs
```diff
@@ -0,0 +1,70 @@
...
+            let start = row_group.columns()[0].byte_range().0;
+            let mid = start + row_group.compressed_size() as u64 / 2;
+            if !self.part.range.contains(&mid) {
+                continue;
+            }
+            // todo predict pushdown
+            self.valid_rowgroups.push(idx);
+        }
+        self.has_blocks()
+    }
```
dantengsky, line 49: It would be much better if we have some docs here

dantengsky:
@sandflee would you please add some descriptions about the motivation of  "split hive file to small partition"? I failed to get to point :(

sandflee:
> @sandflee hi, does this PR also solve the high system cpu problem?

no， this PR read one rowgroup by another， high sys cpu problem maybe resolved by read page level datas in one rowgroup

sandflee:
> @sandflee would you please add some descriptions about the motivation of "split hive file to small partition"? I failed to get to point :(

supose there only one hive file with 100 rowgroups,  we could only have one thread to read the data (only one thread), now we could have at most 100 threads.

src/query/storages/hive/src/hive_blocks.rs
```diff
@@ -0,0 +1,73 @@
...
+            part,
+            valid_rowgroups: vec![],
+            current_index: 0,
+        }
+    }
+
+    // there are some conditions to filter invalid row_groups:
+    // 1. the rowgroup doesn't belong to the partition
+    // 2. filtered by predict pushdown (todo)
+    pub fn prune(&mut self) -> bool {
```
sundy-li, line 41: Can we read the `file_meta` in `read_partition` method?

Thus we can split the row-groups there without introducing `hive_file_splitter`.

sandflee: it'll be too slow to read file meta in read_partition if there're too many files.

sundy-li: So we can parallel this in `read_partition` and cache the meta in part info?

Now seems we may read meta per file more than once.

dantengsky:
> > @sandflee would you please add some descriptions about the motivation of "split hive file to small partition"? I failed to get to point :(
> 
> supose there only one hive file with 100 rowgroups, we could only have one thread to read the data (only one partition), now we could have at most 100 threads.

yeah, reading data concurrently makes sense.  but if we just want concurrency at the level of rowgroup,  shall we just split the task at the boundary of row groups?

Hope I understand this PR right :

**the planning phase**

- Gather all the files to be scanned
- For each file, split it into logical parts `HivePartInfo`
  at the size boundary of 128MB. (one hdfs block?)

 **the scanning phase**
for each hive part
- found the row groups been covered  

    By checking if the part contains a specific point of a given row group, in our case, the middle point of the first column.
    
    - A part may cover zero or more (than one) row groups
      
- read the relevant columns from these row groups   

Thus,  to some extent, data are accessed in the unit of one row-group. (let's ignored the concurrent reading of cols, to simplify the discussion).

My suggestion is ：

- let's just split the parts at the boundary of row groups.
  
  maybe something like this?
~~~
pub struct HivePartInfo {  
    pub filename: String,
    pub partitions: Option<String>,
    pub row_group_index: uszie,
}
~~~
      
- > it'll be too slow to read file meta in read_partition if there're too many files.

  Yeah, I got your concern. we could read them concurrently, as `HiveParquetBlockReader::read_columns_data` or `HiveTable::list_files_from_dirs` did.

sandflee:
@dantengsky  @sundy-li ,    in fact my v1 implement was read meta concurrently (50) from all files cache it and split files at the level of rowgroup,  but it takes serval seconds to split  thousands of files (I forgot the exact nums) in idle hdfs env.  in real env with much more files, this's not acceptable.  so I split files on size not rowgroup likes what hadoop does.

sandflee:
>  By checking if the part contains a specific point of a given row group, in our case, the middle point of the first column.

at the middle point of rowgroup not first column,  it's a good way to spread rowgroups between partitions.

dantengsky:
> at the middle point of rowgroup not first column, it's a good way to spread rowgroups between partitions.

yep. I misinterpreted the code.

dantengsky:
> @dantengsky @sundy-li , in fact my v1 implement was read meta concurrently (50) from all files cache it and split files at the level of rowgroup, but it takes serval seconds to split thousands of files (I forgot the exact nums) in idle hdfs env. in real env with much more files, this's not acceptable. so I split files on size not rowgroup likes what hadoop does.

got it. thanks for the explanations : -)

dantengsky (approved):
LGTM

src/query/storages/hive/src/hive_table_source.rs
```diff
@@ -140,9 +161,28 @@ impl Processor for HiveTableSource {
 
     async fn async_process(&mut self) -> Result<()> {
         match std::mem::replace(&mut self.state, State::Finish) {
-            State::ReadData(part) => {
-                let (meta, chunks) = self.block_reader.read_columns_data(part.clone()).await?;
-                self.state = State::Deserialize(meta, chunks, part);
+            State::ReadMeta(part) => {
+                let part = HivePartInfo::from_part(&part)?;
+                let file_meta = self.block_reader.read_meta_data(&part.filename).await?;
```
sundy-li, line 166: Can we cache this? If the file is splitted into multi partitions, we just need to read once for this file.

dantengsky: we have a BloomIndexMetaCache, which caches `FileMetaData`, but is dedicated to bloom filter index files now.

https://github.com/datafuselabs/databend/blob/1b576fbe78552ca5d024205ee356af70b5b9edf1/src/query/storages/fuse-meta/src/caches/cache.rs#L41.

https://github.com/datafuselabs/databend/blob/26c74149b6b97b7152288b41ea0564114d617aaf/src/query/storages/fuse-meta/src/caches/memory_cache.rs#L50

maybe we could do the refactor in another dedicated PR?

sundy-li: ok

sundy-li (approved):

src/query/storages/hive/src/hive_table_source.rs
sandflee: > Can we cache this? If the file is splitted into multi partitions, we just need to read once for this file.

yes, we need to do more works,  like try allocate partitions belong to one file to one worker, let's do it in another PR?

PR #7340, opened 2022-08-27
feat(query): A naive implementation of PREWHERE optimization.

### LINKED ISSUE #7166: Feature: add prewhere optimization in fuse engine
**Summary**

Description for this feature.

With prewhere optimization, we only read the filter columns and apply the filter operator. If there will be at least one row after the filter operator, then the other columns could be filled from the storage, otherwise, we can skip this block.

This may help in query like:

```
select * from table where url like '%google%';
```

RinChanNOWWW:

## Summary

The implementation can be divide into two parts: the planner part and the fuse engine part.

### Planner Part

After all logical plan optimizations are done, I introduced a new optimizer `WhereOptimizer` to push down prewhere information to logical get / physical scan plan.

Currently, prewhere optimization only focuses on such pattern:

```
Filter
   \
  LogicalGet
```

And by some heuristic rules, viable filter expressions will be pushed down to logical get / physical scan plan.

Current naive rule to pick viable expressions:

- If a experssion doesn't involve all the columns to scan (the projection pushed down), we can move it to prewhere.

The pushed down `PrewhereInfo` mainly contains three fields:

- The viable filter expressions.
- Columns involved in viable filter expressions.
- Columns not involved in the filter expressions but are needed to scan.

### Fuse Engine Part

In this part, we should modify the data source state machine in Fuse engine.

I changed the state machine to be like:

```
                        (In fact this branch is unreachable)
                        No column remains
                       ┌────────────────────────────────┐
                       │                                │
                       │                                ▼
ReadDataPrewhere────►Filter────►ReadDataRemain────►Deserialize────►Generate
       │                                                ▲
       │No filter                                       │
       └────────────────────────────────────────────────┘
```

## TODO

- [x] Add some unit tests for where optimizer.
- [x] Fix failures in tests.

## Next steps

- Prune scan operator's  push down projection.
- Introduce more heuristic rules (can be with some stastics of columns) to `WhereOptimizer`. Like [what's ClickHouse dose ](https://github.com/ClickHouse/ClickHouse/blob/abfddf17e29199be3345b61be782f099e1a09330/src/Storages/MergeTree/MergeTreeWhereOptimizer.cpp#L269-L297)

Fixes #7166


src/query/datablocks/src/data_block.rs
```diff
@@ -188,6 +188,22 @@ impl DataBlock {
         Ok(Self { columns, schema })
     }
 
+    #[inline]
+    pub fn append(self, other: Self) -> Result<Self> {
+        let mut columns = self.columns.clone();
+        let mut fields = self.schema.fields().clone();
+
+        columns.extend(other.columns.into_iter());
+        fields.extend(other.schema.fields().clone().into_iter());
```
sundy-li, line 197: extend_slice ?

RinChanNOWWW:
Currently, due to the storage layout, PRWHERE cannot outperform non-PREWHERE if the amount of partitions is small.

ClSlaid:
You're my hero.

ClSlaid (approved):
You are my hero!

BohuTANG:
Great.
But `setting` related tests failure :)

RinChanNOWWW:
> But setting related tests failure.

It seems that cluster test cannot get the new setting option. Is there anything more do I need to modify?

src/query/storages/fuse/src/operations/read.rs
```diff
@@ -195,15 +278,79 @@ impl Processor for FuseTableSource {
...
+                                true => State::Generated(None, DataBlock::empty()),
+                                false => {
+                                    State::Generated(Some(partitions.remove(0)), DataBlock::empty())
+                                }
+                            };
+                            return Ok(());
+                        }
+                    }
+                    // generate the boolean column for fiter
+                    let boolean_col: &BooleanColumn = Series::check_get(&predicates)?;
```
sundy-li, line 307: const_bool is not covered.

tests/logictest/suites/base/06_show/06_0003_show_settings
```diff
@@ -36,5 +37,6 @@ SHOW SETTINGS LIKE 'enable%';
 enable_async_insert 0 0 SESSION Whether the client open async insert mode, default value: 0 UInt64
 enable_new_processor_framework 1 1 SESSION Enable new processor framework if value != 0, default value: 1 UInt64
 enable_planner_v2 1 1 SESSION Enable planner v2 by setting this variable to 1, default value: 1 UInt64
+enable_prewhere  1  1  SESSION  Enable prewhere optimization by setting this variable to 1, default value: 1  UInt64
```
sundy-li, line 40: Maybe we should not introduce a new setting.

1. we can control it inside the code without exporting a new setting to users.
2. we will have an optimizer rule setting to control whether optimizers could be enabled or not in the future.

src/query/storages/fuse/src/operations/read.rs
```diff
@@ -195,15 +278,79 @@ impl Processor for FuseTableSource {
...
+                    // deserialize prewhere data block first
+                    let block = prewhere
+                        .need_columns_block_reader
+                        .deserialize(part.clone(), chunks.clone())?;
+
+                    // do filter
+                    let res = prewhere.filter.execute(&block)?;
+                    let predicates = DataBlock::cast_to_nonull_boolean(res.column(0))?;
+                    // shortcut, if predicates is const boolean (or can be cast to boolean)
+                    if let Some(const_bool) = DataBlock::try_as_const_bool(&predicates)? {
```
sundy-li, line 292: use `if ! DataBlock::filter_exists(..)` is enough.

src/query/storages/fuse/src/operations/read.rs
```diff
@@ -91,26 +138,40 @@ impl FuseTable {
...
+/// PrewhereExtraData is used for operations after prewhere filter
+type PrewhereExtraData = (Vec<(usize, Vec<u8>)>, BooleanColumn);
+
 enum State {
-    ReadData(PartInfoPtr),
-    Deserialize(PartInfoPtr, Vec<(usize, Vec<u8>)>),
+    PrewhereReadData(PartInfoPtr),
+    PrewhereFilter(PartInfoPtr, Vec<(usize, Vec<u8>)>),
+    ReadData(PartInfoPtr, Option<PrewhereExtraData>),
+    Deserialize(PartInfoPtr, Vec<(usize, Vec<u8>)>, Option<BooleanColumn>),
```
sundy-li, line 148: use `Option<ColumnRef>` so we could hold constant column case.

src/query/storages/fuse/src/operations/read.rs
```diff
@@ -171,16 +245,25 @@ impl Processor for FuseTableSource {
...
 
     fn process(&mut self) -> Result<()> {
         match std::mem::replace(&mut self.state, State::Finish) {
-            State::Deserialize(part, chunks) => {
-                let data_block = self.block_reader.deserialize(part, chunks)?;
+            State::Deserialize(part, chunks, filter) => {
+                let data_block = if let Some(filter) = filter {
+                    let block = self.block_reader.deserialize(part, chunks)?;
+                    // the last step of prewhere
+                    DataBlock::filter_block_with_bool_column(block, &filter)?
```
sundy-li, line 262: DataBlock::filter_block

src/query/storages/fuse/src/operations/read.rs
sundy-li, line 148: `Option<ColumnRef>` to have the constant column case.

src/query/storages/fuse/src/operations/read.rs
sundy-li, line 262: `DataBlock::filter_block`

tests/logictest/suites/base/06_show/06_0003_show_settings
RinChanNOWWW: I see.

RinChanNOWWW:
There is a precision problem in TPCH Q5 test in [test_stateless_standalone_linux](https://github.com/datafuselabs/databend/runs/8061389483?check_suite_focus=true#logs)

```
--- /workspace/tests/suites/0_stateless/13_tpch/13_0005_q5.result	2022-08-28 23:44:33.037390764 +0000
+++ /workspace/tests/suites/0_stateless/13_tpch/13_0005_q5.stdout	2022-08-28 23:46:54.632711861 +0000
@@ -1,4 +1,4 @@
-CHINA	7822103.0
+CHINA	7822102.999
 INDIA	6376121.508
 JAPAN	6000077.218
 INDONESIA	5580475.402
```

I have no idea about it. It will not occur on my own machine.

xudong963:
> There is a precision problem in TPCH Q5 test in [test_stateless_standalone_linux](https://github.com/datafuselabs/databend/runs/8061389483?check_suite_focus=true#logs)
> 
> ```
> --- /workspace/tests/suites/0_stateless/13_tpch/13_0005_q5.result	2022-08-28 23:44:33.037390764 +0000
> +++ /workspace/tests/suites/0_stateless/13_tpch/13_0005_q5.stdout	2022-08-28 23:46:54.632711861 +0000
> @@ -1,4 +1,4 @@
> -CHINA	7822103.0
> +CHINA	7822102.999
>  INDIA	6376121.508
>  JAPAN	6000077.218
>  INDONESIA	5580475.402
> ```
> 
> I have no idea about it. It will not occur on my own machine.

https://github.com/datafuselabs/databend/pull/7344#pullrequestreview-1087976292

src/query/service/src/sql/planner/plans/prewhere.rs
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::Scalar;
+use crate::sql::optimizer::ColumnSet;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Prewhere {
```
xudong963, line 19: `Prewhere` isn't a real plan, I think it's confusing to put it in the plans dir.

RinChanNOWWW: How about put it in planner/logical_get.rs.

RinChanNOWWW: OR in src/query/service/src/sql/common/mod.rs

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
```diff
@@ -0,0 +1,154 @@
...
+                    PatternPlan {
+                        plan_type: RelOp::LogicalGet,
+                    }
+                    .into(),
+                ),
+            ),
+        }
+    }
+
+    fn collect_columns_no_subqueries(expr: &Scalar, columns: &mut ColumnSet) {
```
xudong963, line 51: Maybe `collect_columns` is ok?  We just need some annotations to explain subqueries aren't considered.

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
```diff
@@ -0,0 +1,46 @@
+statement ok
```
xudong963, line 1: Need more test cases, such as the following
```sql
mysql> create table if not exists t_where_optimizer (a int, b int);
ERROR 2013 (HY000): Lost connection to MySQL server during query
Connection id:    9

mysql> explain select a from t_where_optimizer where sum(a) > 1;
+-----------------------------------------------------------------------------------------------------------+
| explain                                                                                                   |
+-----------------------------------------------------------------------------------------------------------+
| Project: [a (#0)]                                                                                         |
| └── Filter: [sum(a) > 1]                                                                                  |
|     └── Scan: default.default.t_where_optimizer, filters: [sum(a) > 1], Sort: [none], limit: [none]       |
+-----------------------------------------------------------------------------------------------------------+
3 rows in set (0.04 sec)

mysql> select a from t_where_optimizer where sum(a) > 1;
ERROR 1105 (HY000): Code: 1015, displayText = Unsupported physical scalar: AggregateFunction(AggregateFunction { display_name: "sum(a)", func_name: "sum", distinct: false, params: [], args: [BoundColumnRef(BoundColumnRef { column: ColumnBinding { database_name: Some("default"), table_name: Some("t_where_optimizer"), column_name: "a", index: 0, data_type: Int32(Int32), visible_in_unqualified_wildcard: true } })], return_type: Int64(Int64) }).
mysql>
```

src/query/service/src/sql/planner/plans/prewhere.rs
xudong963: > How about put it in planner/logical_get.rs.

It's ok

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
xudong963: another case
```
mysql> explain select a from t_where_optimizer where abs(a) > 1;
+----------------------------------------------------------------------------------------------------------------------------------+
| explain                                                                                                                          |
+----------------------------------------------------------------------------------------------------------------------------------+
| Project: [a (#0)]                                                                                                                |
| └── Filter: [abs(t_where_optimizer.a (#0)) > 1]                                                                                  |
|     └── Scan: default.default.t_where_optimizer, filters: [abs(t_where_optimizer.a (#0)) > 1], Sort: [none], limit: [none]       |
+----------------------------------------------------------------------------------------------------------------------------------+
3 rows in set (0.03 sec)
Read 0 rows, 0.00 B in 0.002 sec., 0 rows/sec., 0.00 B/sec.
```

src/query/service/src/sql/optimizer/heuristic/mod.rs
```diff
@@ -94,7 +95,10 @@ impl HeuristicOptimizer {
         let pruner = prune_columns::ColumnPruner::new(self.metadata.clone());
         let require_columns: ColumnSet =
             self.bind_context.columns.iter().map(|c| c.index).collect();
-        pruner.prune_columns(&s_expr, require_columns)
+        let s_expr = pruner.prune_columns(&s_expr, require_columns)?;
+
+        let where_opt = where_optimizer::WhereOptimizer::new(self.metadata.clone());
```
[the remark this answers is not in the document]
RinChanNOWWW: I think we should do where optimization after all other rules are applied.

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
RinChanNOWWW: Ok. This naming is just following ClickHouse.

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
RinChanNOWWW: > select a from t_where_optimizer where sum(a) > 1;`

This sql is invalid. You mean `select a from t_where_optimizer group by a having sum(a) > 1;` ? This kind of sql will not be optimized by `WhereOptimizer`.

RinChanNOWWW: I think almost every case is covered by other tests in stateless-test and sqllogictest.

src/query/service/src/sql/optimizer/heuristic/mod.rs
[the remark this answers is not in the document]
RinChanNOWWW: Or we can merge it into `rule_push_down_filter_scan`?

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
RinChanNOWWW: > another case
> explain select a from t_where_optimizer where abs(a) > 1;

In the current prewhere optimizaiton rule, we will not move the expression involved all columns to prewhere. In this case, we only need to scan column a, and where clause only involves column a, so we it will not be moved to prewhere.

Such case is already covered, like: `explain select a from t_where_optimizer where a = 1;`.

src/query/service/src/sql/optimizer/heuristic/mod.rs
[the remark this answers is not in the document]
xudong963: > Or we can merge it into `rule_push_down_filter_scan`?

SGTM

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
xudong963: > > select a from t_where_optimizer where sum(a) > 1;
> 
> This sql is invalid. You mean `select a from t_where_optimizer group by a having sum(a) > 1;` ? This kind of sql will not be optimized by `WhereOptimizer`.

Yes, it's invalid, so we don't need to consider aggregation function in `collect_columns_no_subqueries`?

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
```diff
@@ -0,0 +1,154 @@
...
+pub struct WhereOptimizer {
+    metadata: MetadataRef,
+    pattern: SExpr,
+}
+
+impl WhereOptimizer {
+    pub fn new(metadata: MetadataRef) -> Self {
+        Self {
+            metadata,
+            pattern: SExpr::create_unary(
```
sundy-li, line 36: can we merge this optimizer into `filter_pushdown`? The pattern is the same.

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
xudong963: > In the current prewhere optimizaiton rule, we will not move the expression involved all columns to prewhere

Why we won't move the expression involved all columns to prewhere?

src/query/storages/fuse/src/operations/read.rs
```diff
@@ -62,27 +67,75 @@ impl FuseTable {
...
         ctx: Arc<dyn TableContext>,
         plan: &ReadDataSourcePlan,
         pipeline: &mut Pipeline,
     ) -> Result<()> {
-        let block_reader =
-            self.create_block_reader(&ctx, self.projection_of_push_downs(&plan.push_downs))?;
+        let projection = self.projection_of_push_downs(&plan.push_downs);
+        let prewhere_info = if let Some(prewhere) = self.prewhere_of_push_downs(&plan.push_downs) {
+            let schema = self.table_info.schema();
+            let expr_field = prewhere.filter.to_data_field(&schema)?;
```
b41sh, line 88: If the filter column is a inner column of tuple, we can't get `expr_field` from `schema`

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
```diff
@@ -0,0 +1,46 @@
...
+Scan: default.default.t_where_optimizer, filters: [t_where_optimizer.b (#1) = 1], Sort: [none], limit: [none]
+
+statement query T
+explain select a from t_where_optimizer where b = 1;
+
+----
+Project: [a (#0)]
+└── Scan: default.default.t_where_optimizer, filters: [t_where_optimizer.b (#1) = 1], Sort: [none], limit: [none]
+
+statement ok
```
b41sh, line 45: please add some test cases of the filter by tuple inner column. for example:
```sql
create table t(id int, s tuple(a int, b int));
explain select * from t where s:a > 0;
```

b41sh (changes requested):

tests/logictest/suites/mode/standalone/explain/where_optimizer.test
RinChanNOWWW: Because it's meaningless. For such case, prewhere will degenerate to normal `Filter - Scan`. So we can just follow the flow without prewhere.

```
   Prewhere
     ┌──►Pre Read Data────►Filter───►Read Remaining Columns and Merge Chunks
     │                                               │
     │                                               │
Start│                                               │
     │                                               │
     │                                               ▼
     └──►Read All Columns─────────────────────►Deserialize ─────►Generate
 No Prewhere
```

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
RinChanNOWWW: LGTM

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        // columns in subqueries are not considered
+        Self::collect_columns_no_subqueries(expr, &mut columns);
+
+        // viable conditions:
+        // 1. Condition depend on some column. Constant expressions are not moved.
+        // 2. Do not move conditions involving all queried columns.
+        // 3. Only current table columns are considered. (This condition is always true in current Pattern (Filter -> LogicalGet)).
+        (
+            !columns.is_empty() && columns.len() < columns_to_scan.len(),
```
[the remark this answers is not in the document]
RinChanNOWWW: > And if they are equal, we can also push prewhere down to the storage layer.

I think not pushing down is better, and it's what exactly ClickHouse does.

For my current design in Databend, this will reduce one step in the data source state machine.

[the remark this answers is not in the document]
b41sh: ok

src/query/storages/fuse/src/operations/read.rs
RinChanNOWWW: Something off topic but related to this: `delete from t where s:a = 1` is also not supported?

RinChanNOWWW: > If the filter column is a inner column of tuple, we can't get expr_field from schema

So the solution is to generate a one-boolean-column schema manually?

src/query/service/src/sql/optimizer/heuristic/mod.rs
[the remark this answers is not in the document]
RinChanNOWWW: We should not merge it to `filter_pushdown`, because we need do this optimization after `prune_columns`. Or `prewhere.need_columns` will be wrong.

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
xudong963: > We should not merge it to `filter_pushdown`, because we need do this optimization after `prune_columns`. Or `prewhere.need_columns` will be wrong.

Could you please give an example?

src/query/service/src/sql/optimizer/heuristic/mod.rs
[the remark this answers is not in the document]
xudong963: Let move the thread to @sundy-li  thread.

src/query/storages/fuse/src/operations/read.rs
b41sh: `delete from t where s:a = 1` is not supported as it has not migrated to new planner.

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
RinChanNOWWW: For example:

With table t2:

```
mysql> desc t2;
+-------+------+------+---------+-------+
| Field | Type | Null | Default | Extra |
+-------+------+------+---------+-------+
| a     | INT  | NO   | 0       |       |
| b     | INT  | NO   | 0       |       |
+-------+------+------+---------+-------+
```

`select a from t2 where a = 1;`

When matching the pattern for `rule_filter_scan_pushdown`, `LogicalGet.columns` is vec![0 ,1]:

image

After `prune_columns`, `LogicalGet.columns` will be changed to vec![0]:

image


If we do prewhere optimization within `rule_filter_scan_pushdown`, we pushed down prewhere information will be wrong.

src/query/storages/fuse/src/operations/read.rs
b41sh: We can generate a schema of inner columns by `path_indices`, like this:
https://github.com/datafuselabs/databend/blob/main/src/query/service/src/storages/storage_table_read_plan.rs#L75

src/query/service/src/sql/optimizer/heuristic/where_optimizer.rs
sundy-li: Ok. I get it.

src/query/storages/fuse/src/operations/read.rs
RinChanNOWWW: I think we can do just like this:

```rust
let expr_schema = DataSchemaRefExt::create(vec![DataField::new(
    prewhere.filter.column_name().as_str(),
    BooleanType::new_impl(),
)]);
```

BohuTANG:
@mergify update

b41sh (approved):
LGTM

BohuTANG:
👍 
Thanks for your contribution!

PR #7346, opened 2022-08-28
fix(query): fix JSON value incorrect memory size

### LINKED ISSUE #7329: `memory_size`  function is incorrect in object column
**Summary**

```
fn memory_size(&self) -> usize {
        self.values.len() * std::mem::size_of::<T>()
}
```

This is incorrect because `T` is an enum and it may have reference to other large objects.

We need to find another lightweight function to have the memory size in rust rather than serde the object into binary.

b41sh:

## Summary

fix `JSON` value incorrect memory size
`serde_json::Value` takes a very high memory size. It will be greatly improved after adopting the `JSONB` format.

Fixes #7329


sundy-li (approved):

wubx:
LGTM

BohuTANG (approved):

PR #7368, opened 2022-08-29
fix(cluster): fix panic if exchange key is nullable

zhang2014:

## Summary

fix(cluster): fix panic if exchange key is nullable

Fixes #issue


BohuTANG (approved):

BohuTANG:
Some tests still need order by clause.

BohuTANG:
It seems that the cluster test still failure caused by the order.

zhang2014:
> It seems that the cluster test still failure caused by the order.

When we improve performance, it will happen more easily, and I will try to fix it.

PR #6668, opened 2022-07-18
fix(parquet): support read i96 timestamp from parquet file

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR

It's compatible change:

```
create table t( a timestamp(0), b timestamp(3), d timestamp(6));
insert into t select now(), now(), now() from numbers(10);

create table t2( a timestamp(0), b timestamp(3), d timestamp(6));
insert into t2 select now(), now(), now() from numbers(10);
```

`databend-query` with Old/New version can read table t and t2.

Fixes #6627

PR #7369, opened 2022-08-29
feat(hive): add timestamp type for hive

sandflee:

## Summary

after https://github.com/datafuselabs/databend/pull/6668,  timestamp datas could be parsed successfully.


sandflee:
arrow2 generates physical type  int64 for timestamp, while the timestamp the parquet file may be int96,  so we use the real parquet meta to decode the data.

BohuTANG:
@mergify update

sundy-li (approved):

BohuTANG:
@mergify update

BohuTANG (approved):

PR #7370, opened 2022-08-29
feat: add new interval kind quarter

ariesdevil:

## Summary

Add new interval kind `quarter`

Fixes #issue


src/meta/proto-conv/src/data_from_to_protobuf_impl.rs
```diff
@@ -491,6 +491,7 @@ impl FromToProto for dv::IntervalKind {
     where Self: Sized {
         let dv_kind = match p {
             pb::IntervalKind::Year => dv::IntervalKind::Year,
+            pb::IntervalKind::Quarter => dv::IntervalKind::Quarter,
```
drmingdrmer, line 494: IMHO, adding another enum item makes the message incompatible:
`Quoter` is 8, when an old program loads a new message, it will fail to load it and return an error:

https://github.com/datafuselabs/databend/blob/2dfb0a2b3236cc861e81633cf67dcb94e3d25eec/src/meta/proto-conv/src/data_from_to_protobuf_impl.rs#L520-L529

Enum item is not like struct message field. An unknown struct field will be just ignored.

tests/logictest/suites/base/02_function/02_0012_function_datetimes
```diff
@@ -800,7 +800,23 @@ select to_year(to_datetime(1646404329000000)) = 2022;
...
+2018-04-02
+
+statement query T
+select date_sub(QUARTER, 1, to_date('2018-01-02'));
 
+----
+2017-10-02
+
+statement query T
+select date_trunc(month, to_date('2022-07-07'));
```
b41sh, line 816: `QUARTER` ?

ariesdevil: This statement is just for `date_trunc`，so `MONTH` is fine.

src/meta/proto-conv/src/data_from_to_protobuf_impl.rs
ariesdevil: God_please_no

TCeason (approved):

src/meta/proto-conv/src/data_from_to_protobuf_impl.rs
drmingdrmer: This is an incompatible feature. It looks like there is no other option.

The choice is to leave it as is, which would be a silent incompatible change, or to increase `MIN_COMPATIBLE_VER` to `VER` to mark it as incompatible explicitly.

drmingdrmer: When upgrading a query cluster, some new query nodes may start feeding new version data into the store, and the old version query nodes have a chance to consume the new version data. 

Thus compatibility is maintained in bi-direction, between query-query or query-meta or meta-meta.

The changes in this PR kind of break the contract of `VER` and `MIN_COMPATIBLE_VER`: compatible versions of programs should be running together forever without any trouble.

But if such a short-term incompatible state is not a thing to worry about while upgrading a cluster, I'd be cool with it.

drmingdrmer (approved):

src/meta/proto-conv/src/data_from_to_protobuf_impl.rs
sundy-li: Got it, I'll be happy to ignore this state.  Let's have this `short-term incompatible` now.

PR #7412, opened 2022-08-31
feat(query): make aggregate function return null on empty set

### LINKED ISSUE #6216: Result of aggregation on empty set
According to SQL specification:

> Case:
> 
>               i) If TXA is empty, then the result is the null value.
> 
>              ii) If AVG is specified, then the result is the average of the
>                  values in TXA.
> 
>             iii) If MAX or MIN is specified, then the result is respec-
>                  tively the maximum or minimum value in TXA. These results
>                  are determined using the comparison rules specified in
>                  Subclause 8.2, "<comparison predicate>".
> 
>              iv) If SUM is specified, then the result is the sum of the
>                  values in TXA. If the sum is not within the range of the
>                  data type of the result, then an exception condition is
>                  raised: data exception-numeric value out of range.

The aggregate functions should return `NULL` except `COUNT` if the input is an empty set. Both MySQL and Postgres follows this rule.

But Databend will return the default value of input type:
```sql
mysql> select sum(number) from numbers(0);
+-------------+
| sum(number) |
+-------------+
|           0 |
+-------------+
1 row in set (0.04 sec)

mysql> select sum(number::float64) from numbers(0);
+---------------------+
| sum(number::DOUBLE) |
+---------------------+
|                 0.0 |
+---------------------+
1 row in set (0.03 sec)
```

Maybe we should follow the SQL standard? It's more reasonable to return `NULL` for empty set. For example, the semantic of `sum([0])` is different from `sum([])`, we should return different result to distinguish between them.

### LINKED ISSUE #7413: bug: AggregateMinMaxFunction' result is wrong when there is a NULL value.
### Search before asking

### Version

v0.7.66-nightly-e6c565e-simd(rust-1.64.0-nightly-2022-08-31T06:25:42.547873538Z)

### What's Wrong?

```
mysql> select * from t;
+------+------+
| a    | b    |
+------+------+
| NULL |    1 |
|    1 |    1 |
|    2 |    2 |
|    3 |    2 |
|    3 |    3 |
|    4 |    3 |
| NULL |    4 |
| NULL |    4 |
+------+------+
mysql> select min(a) from t;
+--------+
| min(a) |
+--------+
|      0 |
+--------+
mysql> select min(a), b from t group by b;
+--------+------+
| min(a) | b    |
+--------+------+
|      0 |    1 |
|      0 |    4 |
|      3 |    3 |
|      2 |    2 |
+--------+------+
```

- [x] Yes I am willing to submit a PR!

sundy-li:

## Summary

Summary about this PR

Make aggregate function return null on empty set.

Now we will make most aggr functions return Nullable State to be consistent with other DBMS.

Example:

Before:

```
MySQL [(none)]> select  sum(number), min(number), avg(number) from numbers_mt (10) where 1 = 2;
+-------------+-------------+-------------+
| sum(number) | min(number) | avg(number) |
+-------------+-------------+-------------+
|           0 |           0 |         NaN |
+-------------+-------------+-------------+
1 row in set (0.004 sec)
```

This will make the aggregation min/max index return 0 on all nulls data which is incorrect.

Now:
```
MySQL [(none)]> select  sum(number), min(number), avg(number) from numbers_mt (10) where 1 = 2;
+-------------+-------------+-------------+
| sum(number) | min(number) | avg(number) |
+-------------+-------------+-------------+
|        NULL |        NULL |        NULL |
+-------------+-------------+-------------+

-- aggregate function: count, uniq will keep non-nullable via using default value 0

MySQL [(none)]> select  count(number), uniq(number), count_distinct(number) from numbers_mt (10) where 1 = 2;
+---------------+--------------+------------------------+
| count(number) | uniq(number) | count_distinct(number) |
+---------------+--------------+------------------------+
|             0 |            0 |                      0 |
+---------------+--------------+------------------------+
```

So the performance will be decreased by 10%~20% on the numbers dataset.

But it's trivial in s3 engine.

Fixes #6216
Fixes #7413


zhang2014 (approved):

BohuTANG (approved):

zhyass (approved):
LGTM

PR #7414, opened 2022-08-31
feat(planner): Fold simple count aggregation

### LINKED ISSUE #7267: optimizer_statistics_exact is missing in new planner
**Summary**

`optimizer_statistics_exact` works in old planner, but it's missing in new planner

leiysky:

## Summary

Fold simple `COUNT()` aggregation if the precise statistics information is available.

This will accelerate the queries like:
```sql
select count(*) from t;
```

Close #7267


src/query/service/src/sql/executor/format.rs
```diff
@@ -62,6 +63,9 @@ fn table_scan_to_format_tree(
     plan: &TableScan,
     metadata: &MetadataRef,
 ) -> Result<FormatTreeNode<String>> {
+    if plan.table_index == DUMMY_TABLE_INDEX {
```
andylokandy, line 66: why specially handle it?

src/query/service/src/sql/optimizer/rule/rewrite/rule_fold_count_aggregate.rs
```diff
@@ -0,0 +1,107 @@
...
+            return Ok(());
+        }
+
+        let rel_expr = RelExpr::with_s_expr(s_expr);
+        let input_prop = rel_expr.derive_relational_prop_child(0)?;
+
+        let is_simple_count = agg.group_items.is_empty()
+            && agg.aggregate_functions.iter().all(|agg| match &agg.scalar {
+                Scalar::AggregateFunction(agg_func) => {
+                    agg_func.func_name == "count" && agg_func.args.is_empty() && !agg_func.distinct
```
andylokandy, line 80: why do all agg_func have to be count()?

src/query/service/src/sql/executor/format.rs
leiysky: For convenient, we can add a new `PhysicalPlan` for `DummyTableScan` instead of this.

src/query/service/src/sql/optimizer/rule/rewrite/rule_fold_count_aggregate.rs
leiysky: Only `count` can be folded now.

BohuTANG (approved):

xudong963 (approved):

src/query/service/src/sql/optimizer/rule/rewrite/rule_fold_count_aggregate.rs
andylokandy: I mean, why count(*) can be folded when other aggr functions exists.

sundy-li (approved):

leiysky:
@mergify update

src/query/service/src/sql/executor/physical_plan_builder.rs
```diff
@@ -232,6 +235,21 @@ impl PhysicalPlanBuilder {
                     table_index: scan.table_index,
                 }))
             }
+            RelOperator::DummyTableScan(_) => {
+                let catalogs = self.ctx.get_catalog_manager()?;
+                let table = catalogs
+                    .get_catalog(CATALOG_DEFAULT)?
```
sandflee, line 241: @leiysky , using 'CATALOG_DEFAULT',   does it infect hive query?

PR #7423, opened 2022-09-01
chore(base): sync runtime dropper

zhang2014:

## Summary

chore(base): sync runtime dropper

part of memory tracker


Xuanwo (approved):

zhang2014:
20689b81cf54711fd584b42be8941495173eabb6 is a temporary fixes. The root cause is that we can't drop runtime because some tasks are blocked during the drop. CC: @youngsofun

BohuTANG (approved):

PR #7425, opened 2022-09-01
feat(query): migrate func to func-v2

TCeason:

## Summary

```
lpad
rpad
replace
strcmp
locate， POSITION(substr IN str) is a synonym for LOCATE(substr,str).
instr alias locate
```

Fixes part of #6766


src/query/functions-v2/src/scalars/string.rs
```diff
@@ -106,6 +120,346 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            };
+            match res {
+                Ordering::Equal => 0,
+                Ordering::Greater => 1,
+                Ordering::Less => -1,
+            }
+        },
+    );
+
+    // registry.register_4_arg::<StringType, NumberType<i64>, NumberType<i64>, StringType, StringType, _, _>(
```
TCeason, line 228: Add 4_arg function. I'm not sure about the number of functions which has 4 args. If not too much I think maybe we just need to use register_function_factory impl the function?@andylokandy

src/query/expression/src/function.rs
```diff
@@ -509,6 +509,182 @@ impl FunctionRegistry {
             }));
     }
 
+    pub fn register_4_arg<I1: ArgType, I2: ArgType, I3: ArgType, I4: ArgType, O: ArgType, F, G>(
```
TCeason, line 512: That's so crazy.

andylokandy: OMG

sundy-li:
We should add macro to represent `register_x_arg `

andylokandy:
@sundy-li i'll take the task. so, @TCeason, would you like to revert to register_x_args version?

TCeason:
> @sundy-li i'll take the task. so, @TCeason, would you like to revert to register_x_args version?

Ok, I will wait your pr

sundy-li (approved):

BohuTANG (approved):
+1

PR #7430, opened 2022-09-01
feat: save share config whenever share meta has been changed

lichuang:

## Summary

feat: save share config whenever share meta has been changed

Fixes #issue


src/query/storages/share/src/share.rs
```diff
@@ -0,0 +1,238 @@
...
+        accounts: Vec<String>,
+    },
+    RemoveShareAccounts {
+        accounts: Vec<String>,
+    },
+}
+
+impl ModShareSpec {
+    pub async fn save(&self, tenant: String, share_id: u64, operator: Operator) -> Result<()> {
+        let location = format!("{}/{}/{}_share.json", SHARE_CONFIG_PREFIX, tenant, share_id);
```
flaneur2020, line 143: we can save all the shares' info into one file? like `tenant/shares.json`.

the config would not become too big, as we can introduce some quotas on limit of shares on the higher level.

src/meta/app/src/share/share.rs
```diff
@@ -611,3 +628,38 @@ pub struct ShareInfo {
     pub name_ident: ShareNameIdent,
     pub meta: ShareMeta,
 }
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Default, Eq, PartialEq)]
+pub struct ShareTableSpec {
+    pub name: String,
+    pub id: u64,
+    pub location: String,
+    pub version: u64,
```
flaneur2020, line 637: the version field can be moved to the top of share config file?

IMO we can keep the config evolution at the top level.

src/query/storages/share/src/share.rs
```diff
@@ -0,0 +1,49 @@
...
+    share_specs: BTreeMap<u64, ShareSpec>,
+}
+
+pub async fn save_share_spec(
+    tenant: String,
+    share_id: u64,
+    operator: Operator,
+    spec: Option<ShareSpec>,
+) -> Result<()> {
+    let location = format!("{}/{}/share_specs.json", SHARE_CONFIG_PREFIX, tenant);
```
flaneur2020, line 34: we do not a tenant prefix here, as all the storages in this databend-query instance is prefixed with a tenant id

src/query/storages/share/src/share.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    if let Some(spec) = spec {
+        let data = operator.object(&location).read().await?;
+        let mut spec_vec: ShareSpecVec = serde_json::from_slice(&data)?;
+        spec_vec.share_specs.remove(&share_id);
+        operator
+            .object(&location)
+            .write(serde_json::to_vec(&spec)?)
```
flaneur2020, line 42: If I understands correctly, this function reads the share_specs.json file, remove the updated ShareSpec in the data, but write backs the new ShareSpec, is this behaviour expected?

src/meta/api/src/share_api_impl.rs
```diff
@@ -1289,3 +1355,68 @@ async fn drop_all_database_from_share(
...
+    let mut tables = vec![];
+    for (_, entry) in share_meta.entries.iter() {
+        if let ShareGrantObject::Table(table_id) = entry.object {
+            let tbid = TableId { table_id };
+
+            let (_tb_meta_seq, tb_meta): (_, Option<TableMeta>) =
+                get_struct_value(kv_api, &tbid).await?;
+            if let Some(tb_meta) = tb_meta {
+                let options = &tb_meta.options;
+                let location = options
```
flaneur2020, line 1394: is this the location the last snapshot of table or the prefix of table?

src/meta/api/src/share_api_impl.rs
```diff
@@ -1289,3 +1355,68 @@ async fn drop_all_database_from_share(
...
+    share_name: &str,
+    share_meta: ShareMeta,
+) -> MetaResult<ShareSpec> {
+    let database = if let Some(database) = share_meta.database {
+        if let ShareGrantObject::Database(db_id) = database.object {
+            let id_key = DatabaseIdToName { db_id };
+
+            let (_db_meta_seq, db_name): (_, Option<DatabaseNameIdent>) =
+                get_struct_value(kv_api, &id_key).await?;
+            if let Some(db_name) = db_name {
```
flaneur2020, line 1370: this if else can take a map in the Option struct:

```
db_name.map(|db_name| ShareDatabaseSpec{ name: db_name.db_name, id: db_id })
```

src/meta/app/src/share/share.rs
```diff
@@ -611,3 +628,38 @@ pub struct ShareInfo {
...
+    pub name: String,
+    pub id: u64,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Default, Eq, PartialEq)]
+pub struct ShareSpec {
+    pub name: String,
+    pub database: Option<ShareDatabaseSpec>,
+    pub tables: Vec<ShareTableSpec>,
+    pub tenants: Vec<String>,
```
flaneur2020, line 664: besides the tenants list, a share may also be granted to ALL, which allows everyone to access this share.

how about adds a `pub global: bool` field?

src/meta/api/src/share_api_impl.rs
flaneur2020: we'd prefer a path in a directory manner, thus we can check whether a path is inside the prefix,  no matter what type of this file

dantengsky:
@lichuang  For locations of shares:

The following commit impls the location generation logic that we've today, it could be picked into this PR without conflicts.

https://github.com/dantengsky/fuse-query/commit/4a597bab8beca60ab795cdb0e4cf95489908fd80


The JSON will be spit looks like this:

https://github.com/dantengsky/fuse-query/commit/4a597bab8beca60ab795cdb0e4cf95489908fd80#diff-76a76b939c63895100f1f27f24fa314f58886c74314909b808defdffe765fa9eR169-R189

src/query/storages/share/src/share.rs
```diff
@@ -0,0 +1,193 @@
...
+                tenants: spec.tenants,
+            }
+        }
+    }
+
+    /// Returns prefix path which covers all the data of give table.
+    /// something like "query-storage-bd5efc6/tnc7yee14/501248/501263/", where
+    ///   - "/query-storage-bd5efc6/tnc7yee14/" is the storage prefix
+    ///   - "501248/" is the database id suffixed with '/'
+    ///   - "501263/" is the table id  suffixed with '/'
```
flaneur2020, line 108: nice comment!

src/query/storages/share/src/share.rs
```diff
@@ -0,0 +1,193 @@
...
+    share_specs: BTreeMap<u64, ext::ShareSpecExt>,
+}
+
+pub async fn save_share_spec(
+    share_id: u64,
+    operator: Operator,
+    spec: Option<ShareSpec>,
+) -> Result<()> {
+    let location = format!("{}/share_specs.json", SHARE_CONFIG_PREFIX);
+    let data = operator.object(&location).read().await?;
```
flaneur2020, line 36: will this `await?` returns error when the file not exists at the first time?

src/query/storages/share/src/share.rs
```diff
@@ -0,0 +1,193 @@
...
+use opendal::Operator;
+
+pub const SHARE_CONFIG_PREFIX: &str = "_share_config";
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Default, Eq, PartialEq)]
+pub struct ShareSpecVec {
+    share_specs: BTreeMap<u64, ext::ShareSpecExt>,
+}
+
+pub async fn save_share_spec(
```
flaneur2020, line 30: there might have one risk on the write-after-read approach from a s3 file:

not all object storages guarantee read-your-write consistency, if two GRANT SHARE operations on one instance closely, it might risks lost-update.

```
Unlike other cloud providers, Amazon S3 delivers strong read-after-write consistency for any storage request, without changes to performance or availability, without sacrificing regional isolation for applications, and at no additional cost.
```

I guess there're ways to work around this:

1. dump all the tenant's shares data from meta on each save
2. dump each shares data with a standalone file name with version and save the latest version in metastore

flaneur2020 (approved):

ZhiHanZ (approved):

drmingdrmer:
@lichuang you added me as a reviewer, but it's still a draft.

Is it ready? 🤔

lichuang:
> 

gogogo

drmingdrmer (commented):
There is a consistency risk with meta-and-storage: When granting a share, meta is updated successfully but update to storage fails.

And there is a chance these updates are applied out of order:
- update A to meta Done
- update B to meta Done
- update B to storage Done
- update A to storage Done

Then the app will be confused.

In short, it's a typical transactional update in distributed env.

The app that consumes these data must be able to tolerate such inconsistent states.

src/meta/api/src/share_api_impl.rs
```diff
@@ -1289,3 +1349,92 @@ async fn drop_all_database_from_share(
...
+            &share_name,
+            format!("get_tenant_share_spec_vec: {}", &share_name),
+        )
+        .await;
+
+        let (_share_id_seq, share_id, _share_meta_seq, share_meta) = match res {
+            Ok(x) => x,
+            Err(_e) => {
+                continue;
+            }
```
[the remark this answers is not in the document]
lichuang: fix it.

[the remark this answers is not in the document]
lichuang: refresh ... may be github has bugs:)

drmingdrmer:
> There is a consistency risk with meta-and-storage: When granting a share, meta is updated successfully but update to storage fails.
> 
> And there is a chance these updates are applied out of order:
> 
> * update A to meta Done
> * update B to meta Done
> * update B to storage Done
> * update A to storage Done
> 
> Then the app will be confused.
> 
> In short, it's a typical transactional update in distributed env.
> 
> The app that consumes these data must be able to tolerate such inconsistent states.
> 
> Everything else is fine to me.:D

Is the consistency issue a concern?
@flaneur2020 @everpcpc @ZhiHanZ

flaneur2020:
> 



> > There is a consistency risk with meta-and-storage: When granting a share, meta is updated successfully but update to storage fails.
> > And there is a chance these updates are applied out of order:
> > 
> > * update A to meta Done
> > * update B to meta Done
> > * update B to storage Done
> > * update A to storage Done
> > 
> > Then the app will be confused.
> > In short, it's a typical transactional update in distributed env.
> > The app that consumes these data must be able to tolerate such inconsistent states.
> > Everything else is fine to me.:D
> 
> Is the consistency issue a concern? @flaneur2020 @everpcpc @ZhiHanZ

I guess eventual consistency is ok, but it'd better to avoid the eventual inconsistency problems 🤔

PR #7436, opened 2022-09-01
feat(query): support using `DEFAULT` to fill default value in `INSERT` statement

### LINKED ISSUE #7312: Feature: Support using `DEFAULT` to fill default value in `INSERT` statement
Sometimes we want to fill the value of some of the fields with its default value, in `PostgreSQL` we can achieve that with:

```sql
create table temp (a int default 2, b int);
insert into temp values(default, 1);
```

RinChanNOWWW:

## Summary

Summary about this PR

Fixes #7312


docs/doc/30-reference/30-sql/10-dml/dml-insert.md
```diff
@@ -105,3 +105,47 @@ SELECT * FROM aggregate_table ORDER BY b;
 |    9 |
 +------+
 ```
+
+## Insert with `DEFAULT` to fill default value
+
+### Syntax
+
+```sql
+INSERT INTO [db.]table [(c1, c2, c3)] VALUES (v1|DEFAULT, v2|DEFAULT, v2|DEFAULT) ...
```
xudong963, line 114:
```suggestion
INSERT INTO [db.]table [(c1, c2, c3)] VALUES (v1|DEFAULT, v2|DEFAULT, v3|DEFAULT) ...
```

src/query/service/src/sql/planner/binder/insert.rs
```diff
@@ -442,6 +443,36 @@ pub fn skip_to_next_row<R: BufferRead>(
...
+    } else {
+        // If field data type is nullable, then we'll fill it with null.
+        if field.data_type().is_nullable() {
+            let scalar = Scalar::ConstantExpr(ConstantExpr {
+                value: DataValue::Null,
+                data_type: Box::new(field.data_type().clone()),
+            });
+            expressions.push((Evaluator::eval_scalar(&scalar)?, field.name().to_string()));
+        } else {
+            expressions.push((
```
xudong963, line 464: 👍

RinChanNOWWW:
In #7139, it is not allowed to fill default value if the column is defined as nullable.

Like:

```sql
create table t (a int, b int null);
insert into t values (1); -- this is not allowed.
```

In this PR, both `insert into t values (1, default)` and `insert into t values (1)` are valid.

Which one should we use?

sundy-li (approved):

sundy-li:
> insert into t values (1); -- this is not allowed.

I think it's not allowed.

If we did not indicate the schema, it's full schema by default in databend.

RinChanNOWWW:
> > insert into t values (1); -- this is not allowed.
> 
> I think it's not allowed.
> 
> If we did not indicate the schema, it's full schema by default in databend.

After #7139, such sql is allowed:

```sql
create table t (a int, b int, c int);
insert into t values (); -- insert (0, 0, 0)
insert into t values(1); -- insert (1, 0, 0)
insert into t values(1,2) -- insert (1, 2, 0)
```

should we keep this feature?

sundy-li:
I think we should remove this "feature", it's confusing.

xudong963 (approved):

PR #7441, opened 2022-09-01
fix(query): fix insert format size

sundy-li:

## Summary

Summary about this PR

Fixes #issue


sundy-li:
These codes seem incorrect in `sequential_format_source.rs`.  cc @youngsofun 
```
 while !data_slice.is_empty() {
                    let (read_size, is_full) = self
                        .input_format
                        .read_buf(data_slice, &mut self.input_state)?;

                    data_slice = &data_slice[read_size..];

                    if is_full {
                        let state = &mut self.input_state;
                        let mut blocks = self.input_format.deserialize_data(state)?;

                        self.data_block.reserve(blocks.len());
                        while let Some(block) = blocks.pop() {
                            progress_values.rows += block.num_rows();
                            self.data_block.push(block);
                        }
                    }
                }
```

src/query/service/src/sql/planner/binder/insert.rs
```diff
@@ -213,12 +213,24 @@ impl<'a> Binder {
...
                 let mut input_state = input_format.create_state();
                 let skip_size = input_format.skip_header(data_slice, &mut input_state, 0)?;
+                data_slice = &data_slice[skip_size..];
+
+                let mut blocks = Vec::new();
+                while !data_slice.is_empty() {
+                    input_state = input_format.create_state();
+                    let (read_size, _) = input_format.read_buf(data_slice, &mut input_state)?;
+
+                    if read_size <= data_slice.len() {
```
zhang2014, line 227:
```suggestion
                    if read_size < data_slice.len() {
```

sundy-li: Use `deserialize_complete_split` instead.

youngsofun (approved):

BohuTANG (approved):

PR #7442, opened 2022-09-01
feat: parser for tabular function

### LINKED ISSUE #5731: Tabular SQL UDFs (UDTFs)
**Summary**

A user-defined table function can apply to a SQL query and returns 0, 1, or multi-rows, each result contains 1 or more columns.
A snow style UDTFs:
```
create function sum_range_func(l int, r int)
  returns table (x integer)
  as 'select sum(a) from test where a > l and a < r'
  ;
  
select * from table(sum_range_func(1,9));
```

Reference:
Databend scalar UDF: https://databend.rs/doc/reference/sql/ddl/udf/ddl-create-function
Snowflake tabular UDF: https://docs.snowflake.com/en/developer-guide/udf/sql/udf-sql-tabular-functions.html

GrapeBaBa:

## Summary

support tabular function

Fixes #5731


src/query/ast/tests/it/testdata/statement.txt
```diff
@@ -6265,3 +6265,299 @@ ShowGrantsOfShare(
...
+            format: None,
+        },
+    },
+)
+
+
+---------- Input ----------
+CREATE FUNCTION t (a, b) RETURNS TABLE (a INT) AS select a from t where b=a and c=b;
+---------- Output ---------
+CREATE FUNCTION t (a, b) RETURNS TABLE (a Int32 NOT NULL) AS SELECT a FROM t WHERE b = a AND c = b
```
BohuTANG, line 6365: `a, b` should have a data type?

src/query/ast/src/ast/statements/tabular_function.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::ast::CreateTableSource;
+use crate::ast::Expr;
+use crate::ast::Identifier;
+use crate::ast::Query;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct CreateTabularFunctionStmt<'a> {
+    pub if_not_exists: bool,
+    pub name: Identifier<'a>,
+    pub args: Vec<Expr<'a>>,
```
GrapeBaBa, line 28: @BohuTANG Following snowflake doc the function arguments should be expression(may be column name or literal). I have some confusion here, I feel in definition these should be parameters which are just placeholder, why snowflake using arguments in definition? If here is arguments, what does the passing arguments do when calling this tabular function? Also if here is arguments, should I evaluate the expression in function definition plan or in function invocation plan?

GrapeBaBa: I noticed in createUDF struct, the parameters were parsed as Identifier, I also don't understand.

BohuTANG: > why snowflake using arguments in definition?
I guess it is for the data type check. Bigquery also has type:
https://cloud.google.com/bigquery/docs/reference/standard-sql/table-functions

BohuTANG: For your question, I would ping @sundy-li

src/query/ast/tests/it/testdata/statement.txt
GrapeBaBa: yes, will add. I already found this problem.

PR #7467, opened 2022-09-04
docs: sum only supports numeric expressions

hanyisong:

## Summary

```sql
MySQL [databend4]> select sum('sdf') from t1;
ERROR 1105 (HY000): Code: 1065, displayText = error:
  --> SQL:1:8
  |
1 | select sum('sdf') from t1
  |        ^^^^^^^^^^ AggregateSumFunction does not support type 'String(String)'

.
```


BohuTANG (approved):
👍

Issue #7318, opened 2022-08-26
Documents about sql behavior in databend


sundy-li:
**Summary**

We should add some documents to help users understand the SQL behavior in databend.

- [ ] DataType  DataType cast/coercion rules   @sundy-li 
- [x] SQL identifiers syntax   @TCeason 
- [x] DataType Nulable and null @b41sh

TCeason:
> DataType DataType cast/coercion rules

I can help to finish it.

rahul3002:
Hey there!

So, you wanna know if these rules will work for coercion, huh?

Check them out:

- Data Compatibility
- Explicit Coercion
- Implicit Coercion

Now, let's talk Decimal type casting:

- Implicit Decimal Type Casting
- Explicit Decimal Type Casting
- Avoiding Data Loss and Precision Issues
- Considerations for Artistic Expression

rahul3002:
Hey there!

I've put together some seriously kick-ass documentation about DataType cast/coercion rules. It's got all the juicy details you need to know. Now, here's the deal: I'd be thrilled if you could take a look at it and let me know what changes you reckon should be made. Your feedback is pure gold to me, so don't hold back!

https://doc.clickup.com/9002042687/p/h/8c90h9z-684/7a7a00ce89a3f0d

sundy-li:
Great, you can create a pr to improve this [doc](https://github.com/sundy-li/fuse-query/blob/write-progress/docs/doc/13-sql-reference/10-data-types/70-data-type-conversion.md)

rahul3002:
Hey there! Have you taken a peek at the documentation I crafted for you? 

If any tweaks or adjustments are needed, I'm all ears and ready to make it perfect!

sundy-li:
> Hey there! Have you taken a peek at the documentation I crafted for you?

I can't open that document, you can create a draft pr and we will comment on that pr.

PR #7472, opened 2022-09-04
docs: add doc for nullable data type

b41sh:

## Summary

add doc for nullable data type

part of #7318


docs/doc/30-reference/10-data-types/60-data-type-nullable-types.md
```diff
@@ -0,0 +1,69 @@
...
+    age Int32 NOT NULL
+);
+```
+
+`NULL` indicates that the data type of this field is `Nullable` and can store `NULL` values.
+`NOT NULL` indicates that this field cannot store `NULL` values.
+If not explicitly set to be `Nullable`, it defaults to `NOT NULL`.
+
+:::note
+Using `Nullable` will almost always have a negative impact on performance. Setting it to `NOT NULL` is a better option.
```
BohuTANG, line 29: The user also wants to know: what's the Default behavior for the databend column?

sundy-li (approved):

PR #7475, opened 2022-09-05
chore: switch to upstream parquet2

### LINKED ISSUE #6064: Improvement: abandon internal patches of parquet2
**Summary**

We have two internal patches of parqeut2, which mainly address the requirement

- acquire the parquet file meta, right after the parquet file has been written, without re-read the file

It works, but awkwardly: each time we sync with upstream(official parquet2), there are some extra works to do (rebase, resolve potential conflicts...)

Among the new features that parquet2 has introduced recently,  the following two seem to be able to resolve the above requirement.

- https://github.com/jorgecarleitao/parquet2/pull/147
- https://github.com/jorgecarleitao/parquet2/pull/148

Thus, 
- we should replace our own internal patches using the new APIs that parquet2 exposes.
- and pin the `parquet2` cargo dependency to the rev of the official parquet2 commit

dantengsky:

## Summary

- deprecate legacy compression algo `lz4`
  use `Lz4Raw` instead.
- use upstream parquet2 (instead patched version of our own)

**About backward compatibility:**

After this PR is merged, users who have legacy data that used legacy `Lz4`, will encounter runtime exceptions like this:

            Deprecated compression algorithm [Lz4] detected.
    
            The legacy compression algorithm [Lz4] is no longer supported.
            To migrate data from the old format, please consider re-create the table, 
            by using an old compatible version [v0.8.25-nightly … v0.7.12-nightly].
            
            - Bring up the compatiable version of databend-query
            - re-create the table
               Suppose the name of the table is T
                ~~~
                create table tmp_t as select * from T;
                drop table T all;
                alter table tmp_t rename to T;
                ~~~
            Please note that the history of table T WILL BE LOST.
                                 
To the best of my knowledge, there are no serious datasets that uses legacy `Lz4`

Fixes #6064

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5337,8 +5337,9 @@ dependencies = [
 
 [[package]]
 name = "parquet2"
-version = "0.14.1"
-source = "git+https://github.com/datafuse-extras/parquet2?rev=3a468fc3c4#3a468fc3c43154b11c22800aff0b83cd22fd4d3c"
+version = "0.14.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f3eb21c37048a08693214d1f0d34eea97a8d90322e15c4a5e2017d56420e1195"
 dependencies = [
  "async-stream",
  "bitpacking",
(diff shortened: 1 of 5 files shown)


Xuanwo:
Great step!

Xuanwo (approved):

PsiACE (approved):

BohuTANG (approved):
:+1

PR #7477, opened 2022-09-05
feat(query): add Users in config file.

youngsofun:

## Summary

allow config user using config file

some design choices:

- wrap `users` in IDMConfig for extentions
-  as flatten as possible to be friendly to toml(and maybe env/clap later, currently skipped)
   -  IDMConfig as top config instead of part of QueryConfig
   -  UserAuthConfig flattened.
- auth_type, auth_string is the same as that in SQL
- auth_string for password hash is in hex, which is the same as system.users

the first pr only aim to let it work:

- [x] add inner and outer config
- [x] give all privileges for configured user (temp)
- [x] can not create a user with the same name as in the config
- [x] unit test

not included in this pr (need ):

- config grants
- config roles
- mv `jwt_key_file` into IDMConfig?
- rm hard code root

Fixes #issue


src/query/config/src/outer_v0.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
```
Xuanwo, line 1: typo

src/query/service/tests/it/configs.rs
```diff
@@ -142,6 +142,9 @@ root = ""
 [catalog]
 meta_store_address = "127.0.0.1:9083"
 protocol = "binary"
+
+[iam]
```
Xuanwo, line 146: `iam` in config seems confusing.

What happened if users want to deploy databend on azure?

src/query/config/src/outer_v0.rs
youngsofun: which word？

Xuanwo: This line is above license header.

src/query/service/tests/it/configs.rs
youngsofun: you are right, although it is in fact a general abbr. https://en.wikipedia.org/wiki/Identity_management

Do you have any suggestions?

- users: it may include other things.
- security: should include TLS configs too

youngsofun: > How about `idm`? (Although it may confuse with `Internet Download Manager`)
> 
> `identities` also LGTM.

Ok, I prefer `identities`

Xuanwo (approved):

src/query/config/src/outer_v0.rs
```diff
@@ -1217,3 +1228,100 @@ impl Debug for MetaConfig {
...
+}
+
+impl From<AuthInfo> for UserAuthConfig {
+    fn from(inner: AuthInfo) -> Self {
+        let auth_type = inner.get_type().to_str().to_owned();
+        let auth_string = inner.get_auth_string();
+        let auth_string = if auth_string.is_empty() {
+            None
+        } else {
+            Some(hex::encode(auth_string))
```
sundy-li, line 1320: why do we need `hex::encode` ?  If `AuthType`  is plain, it should not be encoded.

sundy-li (approved):

PR #7478, opened 2022-09-05
feat(hive): try to use the file size returned by list()

sandflee:

## Summary
after https://github.com/datafuselabs/opendal/pull/577,  we could use the file size from list() , avoid to fetch the meta from seperated files.  cc @Xuanwo

### DIFF
--- a/src/query/storages/hive/src/hive_table.rs
+++ b/src/query/storages/hive/src/hive_table.rs
@@ -549,7 +549,7 @@ async fn do_list_files_from_dir(
                 let filename = path.to_string();
                 let length = match de.content_length() {
                     Some(len) => len,
-                    None => get_file_length(operator.clone(), &path).await?,
+                    None => get_file_length(operator.clone(), path).await?,
                 };
                 all_files.push(HiveFileInfo::create(filename, length));
             }
--- a/src/query/storages/hive/src/hive_table.rs
+++ b/src/query/storages/hive/src/hive_table.rs
@@ -546,9 +546,11 @@ async fn do_list_files_from_dir(
         }
         match de.mode() {
             ObjectMode::FILE => {
-                // todo, support in opendal#list
                 let filename = path.to_string();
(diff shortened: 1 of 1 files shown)


src/query/storages/hive/src/hive_table.rs
```diff
@@ -546,9 +540,8 @@ async fn do_list_files_from_dir(
         }
         match de.mode() {
             ObjectMode::FILE => {
-                // todo, support in opendal#list
                 let filename = path.to_string();
-                let length = get_file_length(operator.clone(), path).await?;
+                let length = de.content_length().unwrap();
```
Xuanwo, line 544: The content_length in `DirEntry` is `best-effort`. Users should get via `metadata` again if the content length is not exist.

sandflee: yes,  get metadata again is more friendly with other storages. and now which storages supports or not supports list with metadata?   for it will effect the performance.

Xuanwo: > now which storages support or not supports list with metadata

First, It's the internal implementation details that users should not depend on. In the future, OpenDAL MAY provide functions to finish this work (get from de, if not found, get from metadata).

Until now, all object storage services and hdfs will return `content_length` during the list. Other services like `fs` don't.

Xuanwo (approved):

BohuTANG:
```
error: this expression creates a reference which is immediately dereferenced by the compiler
   --> src/query/storages/hive/src/hive_table.rs:552:63
    |
552 |                     None => get_file_length(operator.clone(), &path).await?,
    |                                                               ^^^^^ help: change this to: `path`
    |
    = note: `-D clippy::needless-borrow` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
```

sandflee:
Wow,  update to fix cargo clippy

BohuTANG (approved):
👍

PR #7481, opened 2022-09-05
chore(scripts): print logging to provide more detailed information.

PsiACE:

## Summary

This may not be a good practice, but it can provide more detailed information.

logging

Fixes #issue

### DIFF
--- a/.github/actions/publish_binary/databend-scripts-for-release/start.sh
+++ b/.github/actions/publish_binary/databend-scripts-for-release/start.sh
@@ -5,8 +5,8 @@ echo "Deploy new Databend(standalone)"
 ulimit  -n 65535
 nohup bin/databend-meta --config-file=configs/databend-meta.toml 2>&1 >meta.log &
 sleep 3
+cat meta.log
 # export STORAGE_S3_ENABLE_VIRTUAL_HOST_STYLE=true
 nohup bin/databend-query --config-file=configs/databend-query.toml 2>&1 >query.log &
 sleep 3
-echo "Usage: mysql -h127.0.0.1 -P3307 -uroot"
-echo "or: curl -u root: --request POST '127.0.0.1:8001/v1/query/' --data-raw '{\"sql\": \"<your-query>\"}' -H 'Content-Type: application/json'"
+cat query.log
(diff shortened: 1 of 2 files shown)


PsiACE:
cc @Xuanwo Any suggestions for better ways to do this?

Xuanwo:
> cc @Xuanwo Any suggestions for better ways to do this?

The root cause is we can't manage our services.

In standard production level services, we may use `systemd`: 

- Register the `databend-query` service and `databend-meta` service.
- Start `databend-query` and `databend-meta` in the same scope
- Use `journalctl` to fetch or stream the log output: `journalctl -f -u databend`

Also, we can use `docker-compose` to start a service:

- `docker-compose up`

No matter which way, we need to introduce the **real** service idea instead of hacking them at script level.

---

Last words: I hate scripts.

PsiACE:
> No matter which way, we need to introduce the real service idea instead of hacking them at script level.

You are right. I will open an issue to track it. I think we can try to introduce `systemd` first. 

---

But back to the issue, we need to fix the wrong hint.

.github/actions/publish_binary/databend-scripts-for-release/start.sh
```diff
@@ -5,8 +5,8 @@ echo "Deploy new Databend(standalone)"
 ulimit  -n 65535
 nohup bin/databend-meta --config-file=configs/databend-meta.toml 2>&1 >meta.log &
 sleep 3
+cat meta.log
 # export STORAGE_S3_ENABLE_VIRTUAL_HOST_STYLE=true
 nohup bin/databend-query --config-file=configs/databend-query.toml 2>&1 >query.log &
 sleep 3
-echo "Usage: mysql -h127.0.0.1 -P3307 -uroot"
-echo "or: curl -u root: --request POST '127.0.0.1:8001/v1/query/' --data-raw '{\"sql\": \"<your-query>\"}' -H 'Content-Type: application/json'"
+cat query.log
```
Xuanwo, line 12: We can use `tail -f meta.log query.log`

BohuTANG (approved):
💯

Xuanwo (approved):

PR #7512, opened 2022-09-07
refactor(meta/error): improve error usage

drmingdrmer:

##### refactor(meta/error): improve error usage

- When shutting down, an error returned from sub-task should not impede
  the shutting down process. Just leaves an error log and goes on.

- Use a more specific error:
  `InitializeError` for `MetaNode::initialize()`.
  `SledError` for `MetaNode.sto.db::size_on_disk()`.

##### feat(meta): add error MetaManagementError for join/leave operation

Cluster management does not need to return `MetaError`, which is too big
for such tasks.

Currently MetaManagementError has only two variants: `Join()` and
`Leave()`.

In the future low level errors may be added into MetaManagementError to
expose more info about what's going wrong, such as raft-related errors.

## Related Issues


src/meta/service/tests/it/grpc/metasrv_grpc_api.rs
```diff
@@ -157,10 +158,10 @@ async fn test_retry_join() -> anyhow::Result<()> {
 
         tc1.config.raft_config.join = vec![bad_addr.clone()];
         let ret = start_metasrv_with_context(&mut tc1).await;
-        let expect = MetaError::MetaServiceError(format!(
-            "join cluster via {:?} fail",
-            tc1.config.raft_config.join
-        ));
+        let expect = MetaManagementError::Join(AnyError::error(format!(
+            "fail to join {} cluster via {:?}",
+            1, tc1.config.raft_config.join
```
Xuanwo, line 163: Is this a typo, a normal `1`?

drmingdrmer: In this test it tries to join node-1 to the cluster thus here it is just a `1`.

Xuanwo (approved):

PsiACE (approved):

PR #7515, opened 2022-09-07
fix(processor): cancel task when pipeline is finished

### LINKED ISSUE #7509: Feature:  fast finish pipeline if leaf processor finished.
**Summary**
 we found simple select query with `max_thread = 60`  is  more slower than the same query with `max_thread =1 `.  besides the factor of network band, the pipeline could't be finished when limit node (leaf node) finished, for now it waits the finish of all node, especially for waiting for the HiveEngineSource to read data.  I think we could fast finish the pipeline exeutor if the leaf node finished. 

```
+------------------------------------------------------------+
| explain                                                    |
+------------------------------------------------------------+
| LimitTransform × 1 processor                               |
|   Merge (Rename × 60 processors) to (LimitTransform × 1)   |
|     Rename × 60 processors                                 |
|       Project × 60 processors                              |
|         HiveEngineSource × 60 processors                   |
+------------------------------------------------------------+

```

zhang2014:

## Summary

fix(processor): cancel task when pipeline is finished

Fixes #7509

### DIFF
--- a/src/query/service/src/pipelines/executor/pipeline_executor.rs
+++ b/src/query/service/src/pipelines/executor/pipeline_executor.rs
@@ -178,11 +178,11 @@ impl PipelineExecutor {
             }
         }
 
-        if let Err(error_code) = self.graph.check_finished() {
-            let may_error = Some(error_code);
-            (self.on_finished_callback)(&may_error)?;
-            return Err(may_error.unwrap());
-        }
+        // if let Err(error_code) = self.graph.check_finished() {
+        //     let may_error = Some(error_code);
+        //     (self.on_finished_callback)(&may_error)?;
+        //     return Err(may_error.unwrap());
+        // }
 
         (self.on_finished_callback)(&None)?;
         Ok(())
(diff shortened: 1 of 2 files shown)


src/query/service/src/pipelines/executor/pipeline_executor.rs
```diff
@@ -178,11 +178,11 @@ impl PipelineExecutor {
             }
         }
 
-        if let Err(error_code) = self.graph.check_finished() {
-            let may_error = Some(error_code);
-            (self.on_finished_callback)(&may_error)?;
-            return Err(may_error.unwrap());
-        }
+        // if let Err(error_code) = self.graph.check_finished() {
```
Xuanwo, line 181: Remove this or leave comments here?

zhang2014: I will uncomment it in other PR(we need it, but it will cause an error when canceling the task).

BohuTANG (approved):

PR #7518, opened 2022-09-07
feat(query): support purge option in copy into table

### LINKED ISSUE #7443: feat: add PURGE option to COPY command
**Summary**

https://databend.rs/doc/reference/sql/dml/dml-copy-into-table#copyoptions

Add new option `PURGE`
```
copyOptions ::=
  [ SIZE_LIMIT = <num> ]
  PURGE = TRUE | FALSE
```

Remove the data files from the stage/locations automatically after the data is loaded successfully.

sundy-li:

## Summary

Summary about this PR

Fixes #7443


src/query/ast/src/ast/statements/copy.rs
```diff
@@ -38,6 +38,7 @@ pub struct CopyStmt<'a> {
     /// TODO(xuanwo): parse into validation_mode directly.
     pub validation_mode: String,
     pub size_limit: usize,
+    pub purge: bool,
```
Xuanwo, line 41: Display for CopyStmt should also be updated.

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -113,6 +113,26 @@ impl CopyInterpreterV2 {
...
     }
 
+    async fn purge_files(&self, from: &ReadDataSourcePlan, files: &Vec<String>) -> Result<()> {
+        match &from.source_info {
+            SourceInfo::StageSource(table_info) => {
+                if table_info.stage_info.copy_options.purge {
+                    let rename_me: Arc<dyn TableContext> = self.ctx.clone();
+                    let op = StageSourceHelper::get_op(&rename_me, &table_info.stage_info).await?;
+                    for path in files {
+                        op.object(path).delete().await?;
```
BohuTANG, line 123: Suggest not returning an error if the `delete` opeartion failed. We should skip it and write a log.

drmingdrmer (approved):
The `VER` should be increased to reflect the what's changed:
https://github.com/datafuselabs/databend/blob/f26bbf31498f6e6a3bcf7095737e8f52bbe76925/src/meta/proto-conv/src/util.rs#L17

This can just be merged. Let me fire another to do this.
The `VER` should be more semantic to users.

BohuTANG (approved):
💯

PR #7523, opened 2022-09-08
chore: enlarge SLEEP() limit to 300s

flaneur2020:

## Summary

I'm running databend-query behind a proxy, however on getting long running queries the proxy raises 504 timeout sometimes. I tried to enlarge the timeout of the proxy, however it's hard to test if I made the timeout corretly, because SLEEP() function have a 30s upper limit.

Is there any risk on increasing the limit of SLEEP() function? A higher limit may help diagnose the timeout issues across proxies.


BohuTANG:
We can make the sleep with a integer second argument.

zhang2014 (approved):

flaneur2020:
> > I'm running databend-query behind a proxy, however on getting long running queries the proxy raises 504 timeout sometimes. I tried to enlarge the timeout of the proxy, however it's hard to test if I made the timeout corretly, because SLEEP() function have a 30s upper limit.
> 
> Increasing the sleep limit at the databend is ok, but do we have a better way to test Cloudflare's 504 timeout error?

it's easy to test 504 standalone, however a end-to-end test about the long running query also have its benefits

Xuanwo (approved):

PR #7524, opened 2022-09-08
chore(query): take_while tokenizer is ok

### LINKED ISSUE #6669: clickhouse parse error with unmatched quote in format csv/tsv

sundy-li:

## Summary

Summary about this PR

Fixes #6669

### DIFF
--- a/src/query/service/src/sql/planner/mod.rs
+++ b/src/query/service/src/sql/planner/mod.rs
@@ -71,7 +71,7 @@ impl Planner {
         if let Some(Ok(t)) = tokenizer.next() {
             if t.kind == TokenKind::INSERT {
                 let mut token_limit = MIN_TOKEN_FOR_INSERT;
-                while token_limit <= 8 * MAX_TOKEN_FOR_INSERT {
+                while token_limit <= MAX_TOKEN_FOR_INSERT {
                     let tokenizer = Tokenizer::new(sql);
                     match tokenizer
                     .take(token_limit)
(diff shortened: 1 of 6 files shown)


src/query/service/src/sql/planner/mod.rs
```diff
@@ -69,15 +70,24 @@ impl Planner {
...
-                    match self.plan_sql_inner(sql, Some(last_token.span.start)).await {
-                        Ok(v) => return Ok(v),
-                        Err(err) => {
-                            tracing::warn!(
-                                "Faster parser path for insert failed: {}, start to fallback",
-                                err
-                            );
+                let mut token_limit = MIN_TOKEN_FOR_INSERT;
+                while token_limit <= MAX_TOKEN_FOR_INSERT {
+                    let tokenizer = Tokenizer::new(sql);
```
b41sh, line 75: Do we need to create a new `Tokenizer` every time here?

sundy-li: Yes, because it's consumed in the loop.

b41sh (approved):

sundy-li:
a better pr in #7530

PR #7304, opened 2022-08-26
fix(planner): Fix case sensitivity of `USING` clause

leiysky:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Fixes #7277

PR #7525, opened 2022-09-08
fix: change generated alias name for scalar expression to lowercase

xudong963:

## Summary

In https://github.com/datafuselabs/databend/pull/7304, `using_columns` are changed to lowercase.

So for case `select A from (select A from t) t1 join (select A from t) t2 using(A);` in the pr test, column `a` specified in USING clause doesn't exist in the left table(which is `A`).

Fixes https://github.com/datafuselabs/databend-perf/issues/72 Q7


leiysky (approved):

PsiACE (approved):

xudong963:
@mergify update

xudong963:
CI is weird, it keeps the following state 🤣 
image

Xuanwo:
> CI is weird, it keeps the following state

Please ping me if this happen again~

PR #7526, opened 2022-09-08
ci(proto): add compat test for CopyOptions::purge

drmingdrmer:

##### test(proto): add compat test for CopyOptions::purge

- Move the UserStageInfo test into separate files.

  A backward compatibility test should not compare the decoded data with a
  dedicated sample, but not with the data of the latest version.

- Doc: explain how to maintain `VER` and add a simple change-log for
  `VER`, to make it clear when to remove a no longer supported feature
  in the future.

- This is a supplementary PR for #7518, to complete metadata version management and provide test for the new feature introduced.

## Related Issues

### DIFF
--- a/src/meta/proto-conv/tests/it/proto_conv.rs
+++ b/src/meta/proto-conv/tests/it/proto_conv.rs
@@ -28,6 +28,8 @@ use common_proto_conv::VER;
 use common_protos::pb;
 use maplit::btreemap;
 
+use crate::common::print_err;
+
 fn s(ss: impl ToString) -> String {
     ss.to_string()
 }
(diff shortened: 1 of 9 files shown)


src/meta/proto-conv/tests/it/proto_conv.rs
```diff
@@ -233,13 +234,13 @@ fn test_pb_from_to() -> anyhow::Result<()> {
...
-    p.ver = 6;
-    p.min_compatible = 6;
+    p.ver = VER + 1;
+    p.min_compatible = VER + 1;
 
     let res = mt::DatabaseMeta::from_pb(p);
     assert_eq!(
         Incompatible {
-            reason: s("executable ver=5 is smaller than the message min compatible ver: 6")
+            reason: s("executable ver=6 is smaller than the message min compatible ver: 7")
```
sundy-li, line 243: can we use `format!("xxx", VER + 1)` here ?

lichuang (approved):

PR #7530, opened 2022-09-08
refactor(query): replace recursion for fast-path insert with loop

andylokandy:

## Summary

Summary about this PR

Related to https://github.com/datafuselabs/databend/pull/7485

Fixes https://github.com/datafuselabs/databend/issues/6669


src/query/ast/src/parser/statement.rs
```diff
@@ -999,10 +998,7 @@ pub fn insert_source(i: Input) -> IResult<InsertSource> {
             let rest_str = &rest_tokens[0].source
```
sundy-li, line 998: There is a special case that rest_tokens may be empty, so it's not safe to call `rest_tokens[0]`  

See tests in :  https://github.com/datafuselabs/databend/pull/7524/files#diff-777d8f965ebb763eb535886aba615867aab230a5c9caea59851822777fa5d523R10-R14

src/query/service/src/sql/planner/mod.rs
```diff
@@ -66,86 +65,62 @@ impl Planner {
...
+        let mut tokenizer = Tokenizer::new(sql).peekable();
+        let is_insert_stmt = tokenizer
+            .peek()
+            .and_then(|token| Some(token.as_ref().ok()?.kind))
+            == Some(TokenKind::INSERT);
+        // Only tokenize the beginning tokens for `INSERT INTO` statement because it's unnecessary to tokenize tokens for values.
+        let mut tokens: Vec<Token> = if is_insert_stmt {
+            (&mut tokenizer)
+                .take(PROBE_INSERT_INITIAL_TOKENS)
+                .collect::<Result<_>>()?
```
sundy-li, line 81: I think we should use `take_while(|token| token.is_ok())`.

sundy-li (approved):

BohuTANG (approved):

Issue #6338, opened 2022-06-29
Stream COPY from a directory


BohuTANG:
**Summary**

Like:
https://docs.snowflake.com/en/user-guide/data-load-considerations-load.html#loading-older-files

Draft:
Generate a digest from metadata, including:
1. database id
2. table id
3. File etag(md5?)
4. File name

* [x] #7548
* [x] #7549
* [x] #7550
* [x] #7561

BohuTANG:
@junnplus As seen you are working on https://github.com/datafuselabs/databend/pull/6344, are you also interested in this task?

Xuanwo:
@junnplus May be busy on other works, let me take this.

junnplus:
@Xuanwo I haven't started this task yet. You are free to assign.

BohuTANG:
This is related to metasrv, such as the unique identify ID can be stored in the metasrv.
cc @lichuang

BohuTANG:
Because databend can COPY from remote files, so the file may not be an object from s3.

Xuanwo:
Cc @ClSlaid, can you help add the `ETag` field in StageFile?

PR #7531, opened 2022-09-08
feat: save copy into table stage file meta, avoid duplicate copy file

lichuang:

## Summary

1. introduce  `TableStageFileInfo` in meta to save copy into table stage file info, it will be expired after 64 days.
2. avoid duplicate copy file operation with compare `TableStageFileInfo.md5`.
3. add proc compatible test

Fixes https://github.com/datafuselabs/databend/issues/6338


src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -52,6 +60,49 @@ impl CopyInterpreterV2 {
...
+            let stage_file = list_files(&self.ctx, &table_info.stage_info, file, "").await?;
+            if stage_file.is_empty() {
+                continue;
+            }
+
+            let stage_file = &stage_file[0];
+
+            if let Some(file_info) = resp.file_info.get(file) {
+                // No need to copy the file again only if md5 match.
+                if file_info.md5 == stage_file.md5 {
```
Xuanwo, line 93: We need to check the following things:

- name
- length
- last modified
- etag (use etag instead of md5)

Xuanwo (commented):
This PR is a bit complex that mixed with meta, catalog and query.

Can you give an RFC first so we can understand how and why we address the issue in this way?

drmingdrmer (commented):
I did not finish reading all of the code.
The first of my advices is to keep `UpsertKVReq::udpate()` as it is.
And use `UpsertKVReq::udpate().with(KVMeta {expire_at: Some(now + 2)}` to update the expire time if needed.

This way it does not have to modify all the usages of `update()`.

Ref: https://github.com/datafuselabs/databend/blob/18f90a3ac665ff0e8b9f960731fbcfe6ed11c47f/src/meta/api/src/kv_api_test_suite.rs#L278-L285

src/meta/api/src/schema_api_test_suite.rs
```diff
@@ -3115,6 +3119,160 @@ impl SchemaApiTestSuite {
...
+                let ident = TableIdent::new(tb_id, seq);
+
+                let want = TableInfo {
+                    ident: ident.clone(),
+                    desc: format!("'{}'.'{}'.'{}'", tenant, db_name, tbl_name),
+                    name: tbl_name.into(),
+                    meta: table_meta(created_on),
+                };
+                assert_meta_eq_without_updated!(want, got.as_ref().clone(), "get created table");
+                ident
```
drmingdrmer, line 3199: It does not have to test create-table. Creating a table is already guaranteed to work as expected in other cases.

src/query/service/src/interpreters/interpreter_copy_v2.rs
Xuanwo: > if the location is https(not s3), the etag is?

ETag in OpenDAL is the same `ETag` defined in HTTP Standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

lichuang: the type `StageFile`, which save stage file meta info, has no `etag` field:

```
#[derive(Default, Clone)]
pub struct StageFile {
    pub path: String,
    pub size: u64,
    pub md5: Option<String>,
    pub last_modified: DateTime<Utc>,
    pub creator: Option<UserIdentity>,
}
```

lichuang:
@drmingdrmer @BohuTANG @Xuanwo 

I split the original issue into sub-issues, cause I think this pr is too large, including meta and query changes, and need some help of @Xuanwo  to add the etag field of `StageFile`, so I close this issue, but your review advice will bring into the sub-issue pr, thank you.

PR #7539, opened 2022-09-09
feat(query): add char/ord/soundex to function-v2

### LINKED ISSUE #7543: bug: expression explain error
### Search before asking

### Version

8.0.26-v0.8.29-nightly-e849dde-simd(rust-1.64.0-nightly-2022-09-09T03:25:49.093997816Z)

### What's Wrong?

```sql
MySQL [databend7]> select (true IN (false, (NULL NOT BETWEEN NULL AND NULL)));

ERROR 1105 (HY000): Code: 1010, displayText = Can't cast column from nullable data into non-nullable type (while in processor thread 0).
```

### How to Reproduce?

_No response_

sundy-li:

## Summary

Summary about this PR

Related #6766

Fixes #7543


src/query/functions-v2/src/scalars/string_multi_args.rs
```diff
@@ -309,4 +240,106 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    let len = args.iter().find_map(|arg| match arg {
+        ValueRef::Column(col) => Some(col.len()),
+        _ => None,
+    });
+    let input_rows = len.unwrap_or(1);
+
+    let mut values: Vec<u8> = vec![0; input_rows * args.len()];
+    let values_ptr = values.as_mut_ptr();
+
+    for (i, arg) in args.iter().enumerate() {
```
RinChanNOWWW, line 318: What if `v` is not in the range of `u8`? For example: `select char(77, 111111, 77);`. MySQL will return `NULL` if there is any invalid value(arg). And MySQL will ignore `NULL` in values(args).

sundy-li: I think we can just accept u8 as the signature. For other behaviors in MySQL, we don't need to follow strictly.

RinChanNOWWW:
LGTM.

RinChanNOWWW (approved):

src/query/expression/src/function.rs
```diff
@@ -1042,3 +1047,52 @@ pub fn passthrough_nullable_3_arg<I1: ArgType, I2: ArgType, I3: ArgType, O: ArgT
...
     }
 }
+
+pub fn wrap_nullable<F>(
+    f: F,
+) -> impl Fn(&[ValueRef<AnyType>], &GenericMap) -> Result<Value<AnyType>, String> + Copy
+where F: Fn(&[ValueRef<AnyType>], &GenericMap) -> Result<Value<AnyType>, String> + Copy {
+    move |args, generics| {
+        type T = NullableType<AnyType>;
+        type Result = AnyType;
```
TCeason, line 1057: type Result = AnyType;
type T = NullableType<Result>;

Maybe can use Result replace all AnyTYpe in this function?

src/query/expression/src/function.rs
```diff
@@ -1042,3 +1047,52 @@ pub fn passthrough_nullable_3_arg<I1: ArgType, I2: ArgType, I3: ArgType, O: ArgT
...
+        let mut bitmap: Option<MutableBitmap> = None;
+        let mut nonull_args: Vec<ValueRef<AnyType>> = Vec::with_capacity(args.len());
```
TCeason, line 1059: let mut nonull_args: Vec<ValueRef<Result>> = Vec::with_capacity(args.len());

TCeason (approved):
LGTM

src/query/functions-v2/src/scalars/string.rs
```diff
@@ -604,3 +673,34 @@ fn vectorize_string_to_string_2_arg(
         }
     }
 }
+
+struct Soundex;
+
+impl Soundex {
```
andylokandy, line 679: we can make it a mod

PR #7541, opened 2022-09-09
docs: add avoid-duplicate-when-copy-into-table rfc

lichuang:

## Summary

docs: add avoid-duplicate-when-copy-into-table rfc

Fixes #issue


docs/doc/60-contributing/03-rfcs/20220909-avoid-duplicate-when-copy-into-table.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Avoid duplicate when copy stage files into a table
```
Xuanwo, line 2: RFC's title should be short and clear.

How about `Idempotent Copy`?

docs/doc/60-contributing/03-rfcs/20220909-avoid-duplicate-when-copy-into-table.md
```diff
@@ -0,0 +1,65 @@
...
+description: Avoid duplicating when copy stage files into a table
+---
+
+- Tracking Issue: <https://github.com/datafuselabs/databend/issues/6338>
```
Xuanwo, line 6:
```suggestion
- Tracking Issue: https://github.com/datafuselabs/databend/issues/6338
```

drmingdrmer (approved):
Crystal clear explanation.

One of my question is, in this sentence:
`copy the stage file and up-insert into the table stage file meta`
These are two RPC operations to databend-meta, right?

BTW, the words in this font is not very easy to read, I'm afraid...

image

BohuTANG:
Clear 👍 
One question:
does the copy file meta in metaservice with the SQL digest?
```
COPY into t1 select c1 from @stage;
```

is not same as:
```
COPY into t1 select parse_json(c2) from @stage;
```

BohuTANG:
Another question:
If we `truncate table` and COPY from a stage/location again, all the files will be skipped.

docs/doc/60-contributing/03-rfcs/20220909-avoid-duplicate-when-copy-into-table.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Summary
+
+When streaming copy stage files from a directory into a table, there is a chance that some files have already been copied, So it needs some ways to avoid duplicate copying files.
+
+## Save copy into table stage files meta information in meta service
+
+Whenever copy stage files into a table, save the stage file meta information into the meta service:
```
flaneur2020, line 14: is the stage files metadata updated in a transaction with the data commit?

lichuang:
> Crystal clear explanation.
> 
> One of my question is, in this sentence: `copy the stage file and up-insert into the table stage file meta` These are two RPC operations to databend-meta, right?
> 
> BTW, the words in this font is not very easy to read, I'm afraid...
> 
> image

`copy the stage file` means copy the stage file into table, it is not an operation on meta service, but like inserting data into a table, so it happen on object store such as s3 in databend.

 `up-insert into the table stage file meta` means up-insert the stage file meta into meta service, so the next time can use these meta data to  compare with original stage file meta, and ignore the stage files which has not been modified since last copied.

lichuang:
> 



> Another question:
> If we `truncate table` and COPY from a stage/location again, all the files will be skipped.

emm, not only `truncate table`, but also `drop table` MUST consider the table file stage meta data...

docs/doc/60-contributing/03-rfcs/20220909-avoid-duplicate-when-copy-into-table.md
lichuang: > is the stage files metadata updated in a transaction with the data commit?

what do you mean about `data commit`, i guess it means `copy the stage files into table`, if this is what you means, the answer is no, cause `copy the stage files into table` happen in object store, but `table stage files metadata updated` happen in meta service, they cannot be in a transaction.

lichuang:
> Clear 👍 One question: does the copy file meta in metaservice with the SQL digest?
> 
> ```
> COPY into t1 select c1 from @stage;
> ```
> 
> is not same as:
> 
> ```
> COPY into t1 select parse_json(c2) from @stage;
> ```

no, the key of `table stage copy stage file meta` combined with (tenant,database,table,file_name), which has been analyzed by the SQL parser.

Xuanwo:
`StageFile` is an internal data type we don't want to persist.

Can we just construct the key the specific way instead of storing `StageFile` in the meta?

For example:

```
{prefix}/{tenant}/{db_name}/{table_name}/{file_name}/{content_length}/{etag}/{last_modified}
```

We just need to check whether this key exists or not without reading its content.

lichuang:
> `StageFile` is an internal data type we don't want to persist.
> 
> Can we just construct the key the specific way instead of storing `StageFile` in the meta?
> 
> For example:
> 
> ```
> {prefix}/{tenant}/{db_name}/{table_name}/{file_name}/{content_length}/{etag}/{last_modified}
> ```
> 
> We just need to check whether this key exists or not without reading its content.

since meta now has no  mechanism to recycle the expire data, so if the key is `{prefix}/{tenant}/{db_name}/{table_name}/{file_name}/{content_length}/{etag}/{last_modified}` will make more than one key of a file, i prefer `{prefix}/{tenant}/{db_name}/{table_name}/{file_name} -> {content_length}/{etag}/{last_modified}` instead.

docs/doc/60-contributing/03-rfcs/20220909-Idempotent-Copy.md
```diff
@@ -0,0 +1,56 @@
...
+## Save copy into table stage files meta information in meta service
+
+Whenever copy stage files into a table, save the stage file meta information into the meta service:
+
+- key: combined with `(tenant, database, table, file name)`.
+- value: value MUST includes all the meta of a stage file, such as `content-length`,`etag`,`last modified`.
+
+
+
+![](/img/rfc/20220909-Idempotent-Copy/stage-file-meta.png)
```
Xuanwo, line 21: Please use `/img/rfc/20220909-idempotent-copy/stage-file-meta.png`

So that we don't have a upcase folder.

Xuanwo:
In an edge case, what will happen if users copy from different stages:

```sql
copy from @gcs_stage into mytable;
copy from @aws_stage into mytable;
```

docs/doc/60-contributing/03-rfcs/20220909-Idempotent-Copy.md
```diff
@@ -0,0 +1,56 @@
...
+
+Take the image above as an example:
+
+* Client make a request to copy thress files (file1, file2, file3) into table.
+
+* Get the table stage file meta of (file1, file2, file3).
+
+* In the meta service, only found (file1,file3) stage file information.
+
+* Compare the table stage file information with stage file information, and found that file1 has not been changed, so file1 will be ignored in this copy operation, and (file2,file3) will be copied.
```
b41sh, line 51: In the example.png
`5. file1 matched, will be ignored. Only copy (file1, file3).`
The second `file1` seemed is a typo.

Xuanwo:
Vercel built failed:

image

PR #7545, opened 2022-09-09
feat(metrics-count): abstract active instance counting

drmingdrmer:

##### feat(metrics-count): abstract active instance counting

- Add `WithCount` as a wrapper of every type `T` that needs to count the
  active instances of it.
  So that the mechanism of metrics reporting such as of live connection can be reused.

- Re-implement `RaftClient` with `WithCount`.

## Related Issues

### DIFF
--- a/src/common/metrics/src/counter.rs
+++ b/src/common/metrics/src/counter.rs
@@ -1,0 +1,132 @@
+// Copyright 2022 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This mod provides mechanism to track the count of active instances of some type `T`.
+//! The count is maintained by a `Count` implementation and will be increased or decreased when a wrapper of `T` `WithCounter` is created or dropped.
+//!
+//! Example:
+//!
+//! ```ignore
+//! struct Connection{}
+//! impl Connection {
+//!     fn ping() {}
+//! }
+//!
+//! struct MyCounter{ identifier: String, }
+//! impl Count for MyCounter {/*...*/}
+//!
+//! {
+//!     let conn = WithCounter::new(Connection{}, MyCounter{}); // increase count with `MyCounter`
+//!     conn.ping();
+//! } // decrease count with `MyCounter`
+//! ```
+
+use std::ops::Deref;
+use std::ops::DerefMut;
+
+/// Defines how to report counter metrics.
+pub trait Count {
+    fn incr_count(&mut self, n: i64);
+}
+
+/// Binds a counter to a `T`.
+///
+/// It counts the number of instances of `T` with the provided counter `Count`.
+pub struct WithCount<C, T>
+where C: Count
+{
+    counter: C,
+    inner: T,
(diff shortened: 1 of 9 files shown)


src/common/metrics/src/counter.rs
```diff
@@ -0,0 +1,132 @@
...
+impl<C, T> DerefMut for WithCount<C, T>
+where C: Count
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
+
+#[cfg(test)]
+mod t {
```
PsiACE, line 97: `tests`

PsiACE (commented):
lgtm, t -> tests plz

lichuang (approved):

Xuanwo:
Maybe you will be interested in how OpenDAL implements the same feature:

OpenDAL exposes a trait `Layer`

```rust
pub trait Layer {
    /// Intercept the operations on the underlying storage.
    fn layer(&self, inner: Arc<dyn Accessor>) -> Arc<dyn Accessor>;
}
```

So we can implement metrics on every public API of underlying `Accessor`:

https://github.com/datafuselabs/opendal/blob/main/src/layers/metrics.rs

```rust
async fn stat(&self, args: &OpStat) -> Result<ObjectMetadata> {
    increment_counter!(
        METRIC_REQUESTS_TOTAL,
        LABEL_SERVICE => self.meta.scheme().into_static(),
        LABEL_OPERATION => Operation::Stat.into_static(),
    );

    let start = Instant::now();
    let result = self.inner.stat(args).await;
    let dur = start.elapsed().as_secs_f64();

    histogram!(
        METRIC_REQUESTS_DURATION_SECONDS, dur,
        LABEL_SERVICE => self.meta.scheme().into_static(),
        LABEL_OPERATION => Operation::Stat.into_static(),
    );

    result
}
```

End users just need to:

```rust
use anyhow::Result;
use opendal::layers::MetricsLayer;
use opendal::Operator;
use opendal::Scheme;

let _ = Operator::from_env(Scheme::Fs)
    .expect("must init")
    .layer(MetricsLayer);
```

drmingdrmer:
> Maybe you will be interested in how OpenDAL implements the same feature:
> 
> OpenDAL exposes a trait `Layer`
> 
> ```rust
> pub trait Layer {
>     /// Intercept the operations on the underlying storage.
>     fn layer(&self, inner: Arc<dyn Accessor>) -> Arc<dyn Accessor>;
> }
> ```
> 
> So we can implement metrics on every public API of underlying `Accessor`:
> 
> https://github.com/datafuselabs/opendal/blob/main/src/layers/metrics.rs
> 
> ```rust
> async fn stat(&self, args: &OpStat) -> Result<ObjectMetadata> {
>     increment_counter!(
>         METRIC_REQUESTS_TOTAL,
>         LABEL_SERVICE => self.meta.scheme().into_static(),
>         LABEL_OPERATION => Operation::Stat.into_static(),
>     );
> 
>     let start = Instant::now();
>     let result = self.inner.stat(args).await;
>     let dur = start.elapsed().as_secs_f64();
> 
>     histogram!(
>         METRIC_REQUESTS_DURATION_SECONDS, dur,
>         LABEL_SERVICE => self.meta.scheme().into_static(),
>         LABEL_OPERATION => Operation::Stat.into_static(),
>     );
> 
>     result
> }
> ```
> 
> End users just need to:
> 
> ```rust
> use anyhow::Result;
> use opendal::layers::MetricsLayer;
> use opendal::Operator;
> use opendal::Scheme;
> 
> let _ = Operator::from_env(Scheme::Fs)
>     .expect("must init")
>     .layer(MetricsLayer);
> ```

It's a bit different from the per method call metrics:
- The count is increased when a client is created and is decreased when dropped.
- The inner type `T` does not have a fixed `trait` for the upper layer to use.

Thus the layered pattern may not apply in this case. But I still love its cleanness and neatness:D

PsiACE (approved):

PR #7558, opened 2022-09-09
feat: save table stage file info into meta, remove these data when truncate table

lichuang:

## Summary

1. add `get_table_stage_file_info` and `upsert_table_stage_file_info` API to get and save table stage file info into meta.
2. add `truncate_table` API to remove table stage file data.
3. test suite and proc compatible test.
4. update proc version.

Fixes https://github.com/datafuselabs/databend/issues/7549


src/meta/api/src/schema_api_keys.rs
```diff
@@ -294,3 +296,48 @@ impl KVApiKey for CountTablesKey {
...
+// __fd_table_stage_file/tenant/db_name/table_name/file_name -> TableStageFileInfo
+impl KVApiKey for TableStageFileNameIdent {
+    const PREFIX: &'static str = PREFIX_TABLE_STAGE_FILE;
+
+    fn to_key(&self) -> String {
+        format!(
+            "{}/{}/{}/{}/{}",
+            Self::PREFIX,
+            self.tenant,
+            self.db_name,
```
BohuTANG, line 309: Is it possible to use db_id and table_id for the prefix?
If the user `drop table t1 and create table t1`, `drop table t1` failed to drop the stage meta files in metasrv, I think there may be some issues.

src/meta/api/src/schema_api_test_suite.rs
```diff
@@ -2484,17 +2492,51 @@ impl SchemaApiTestSuite {
...
+            file_info.insert("file".to_string(), stage_info.clone());
+
+            let req = UpsertTableStageFileReq {
+                table: TableNameIdent {
+                    tenant: tenant1.to_string(),
+                    db_name: db1_name.to_string(),
+                    table_name: tb1_name.to_string(),
+                },
+                file_info: file_info.clone(),
+                expire_at: Some((Utc::now().timestamp() + 86400) as u64),
```
BohuTANG, line 2525: Can we remove the expire_at?
I am not sure why snowflake has this limit?
For example:
A has a public dataset on s3, and B wants to load the increment files from A's bucket by `COPY into t1 FROM s3://B/bucket/`. When the expire reached after 86400, B will reload all the data, this is a big problem.

lichuang: > Can we remove the expire_at? I am not sure why snowflake has this limit? For example: A has a public dataset on s3, and B wants to load the increment files from A's bucket by `COPY into t1 FROM s3://B/bucket/`. When the expire reached after 86400, B will reload all the data, this is a big problem.

`UpsertTableStageFileReq.expire_at` is an option field, which can be set by the `schema_api` user, if copy into table stage file meta needs not to be expired, this field can set to be none.

BohuTANG: OK, let's set it to `None` make it never expired.
Thanks.

src/meta/api/src/schema_api_keys.rs
```diff
@@ -294,3 +296,48 @@ impl KVApiKey for CountTablesKey {
         Ok(CountTablesKey { tenant })
     }
 }
+
+// __fd_table_stage_file/tenant/db_name/table_name/file_name -> TableStageFileInfo
```
BohuTANG, line 300: This comments should modify: `/db_id/table_id`

BohuTANG (approved):
Lgtm

src/meta/api/src/schema_api_keys.rs
```diff
@@ -43,6 +44,7 @@ const PREFIX_TABLE_ID_LIST: &str = "__fd_table_id_list";
 const PREFIX_TABLE_COUNT: &str = "__fd_table_count";
 const PREFIX_DATABASE_ID_TO_NAME: &str = "__fd_database_id_to_name";
 const PREFIX_TABLE_ID_TO_NAME: &str = "__fd_table_id_to_name";
+const PREFIX_TABLE_STAGE_FILE: &str = "__fd_table_stage_file";
```
Xuanwo, line 47: This is not a table for `stage_file`. How about `copied_files`?

src/meta/api/src/schema_api_keys.rs
```diff
@@ -294,3 +296,48 @@ impl KVApiKey for CountTablesKey {
...
+            Self::PREFIX,
+            self.tenant,
+            self.db_id,
+            self.table_id,
+            self.file,
+        )
+    }
+
+    fn from_key(s: &str) -> Result<Self, KVApiKeyError> {
+        let mut elts = s.split('/');
```
Xuanwo, line 316: Maybe buggy. Filename could be `path/to/file`.

src/meta/protos/proto/table.proto
```diff
@@ -162,4 +162,13 @@ message DatabaseIdTableName {
...
   string table_name = 2;
+}
+
+message TableStageFileInfo {
+  uint64 ver = 100;
+  uint64 min_compatible = 101;
+
+  optional string etag = 1;
+  uint64 content_length = 2;
+  string last_modified = 3;
```
Xuanwo, line 173: last_modified is also `optional`

src/meta/api/src/schema_api_keys.rs
lichuang: ok

src/meta/api/src/schema_api_keys.rs
lichuang: add `KVApiKeyError::AtleastSegments` to handle these case, and add some test in `schema_api_keys.rs`.

src/meta/api/src/schema_api_keys.rs
```diff
@@ -341,3 +354,76 @@ impl KVApiKey for TableStageFileNameIdent {
         })
     }
 }
+
+#[cfg(test)]
```
BohuTANG, line 358: Can we move the test block to a test file?

drmingdrmer: A unit test to test private fields/methods has to be a nested `tests` mod.

A unit test that tests only one mod being a nested `tests` mod is also practical and efficient, IMHO.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1715,6 +1725,267 @@ impl<KV: KVApi> SchemaApi for KV {
...
+
+            let (_file_seq, file_stage_info): (_, Option<TableStageFileInfo>) =
+                get_struct_value(self, &key).await?;
+
+            if let Some(file_stage_info) = file_stage_info {
+                file_info.insert(file, file_stage_info);
+            }
+        }
+
+        Ok(GetTableStageFileReply { file_info })
```
drmingdrmer, line 1793: This is a read operation without a consistency guarantee:

The table may be moved to another db with `rename_table`.
The files may be modified while reading every one of them.

I can not tell if this logic will work correctly, it depends on whether its caller can tolerate these inconsistencies.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1715,6 +1725,267 @@ impl<KV: KVApi> SchemaApi for KV {
...
+                name = display(&tenant_dbname_tbname),
+                table_meta = debug(&tb_meta),
+                "upsert_table_stage_file_info"
+            );
+
+            let mut condition = vec![
+                txn_cond_seq(&tenant_dbname, Eq, db_id_seq),
+                txn_cond_seq(&dbid_tbname, Eq, tb_id_seq),
+                txn_cond_seq(&tbid, Eq, tb_meta_seq),
+            ];
```
drmingdrmer, line 1852: Does `upsert_table_stage_file_info` have to assert about the tenant and `dbid-name` when applying a modification?

IMHO, renaming a table or else should not prohibit any operation that works only with table data.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1929,14 +2200,44 @@ impl<KV: KVApi> SchemaApi for KV {
...
+        let (file_seq, _opt): (_, Option<TableStageFileInfo>) =
+            get_struct_value(kv_api, &file).await?;
+        if file_seq != 0 {
+            condition.push(txn_cond_seq(&file, Eq, file_seq));
+            if_then.push(txn_op_del(&file));
+        }
+    }
+
+    Ok(())
+}
```
drmingdrmer, line 2231: Inconsistent write:

The file list may change after `list_keys` at line-2220.

To assert consistency, every time copying-files change, the seq of table-meta must increase. So that when deleting these files, by checking table-meta-seq it guarantees there are not changes to the file list.

This is similar to the db-table pattern: a **parent** node is responsible for the consistency of the **set** of its child nodes.

drmingdrmer (commented):
There is some consistency concern.

src/meta/api/src/schema_api_impl.rs
lichuang: yes, i think `copy into` can tolerate these inconsistencies.

src/meta/types/src/proto_display.rs
```diff
@@ -144,11 +144,22 @@ impl Display for TxnGetRequest {
...
+                )
+            }
+            Some(expire_at) => {
+                write!(
+                    f,
+                    "Put key={}, need prev_value: {} expire at: {}",
+                    self.key, self.prev_value, expire_at,
+                )
+            }
+        }
```
drmingdrmer, line 162:
```suggestion
        write!(f, "Put key={}, need prev_value: {}", self.key, self.prev_value)?;
        if let Some(expire_at) = self.expire_at {
                   write!(f, " expire at: {}", expire_at)?;
        }
        Ok(())
```

src/meta/proto-conv/tests/it/proto_conv.rs
```diff
@@ -491,5 +509,19 @@ fn test_load_old() -> anyhow::Result<()> {
         assert_eq!(want, got);
     }
 
+    // TableCopiedFileInfo is loadable
+    {
+        let stage_file: Vec<u8> = vec![
+            10, 4, 101, 116, 97, 103, 16, 128, 8, 26, 23, 50, 48, 49, 52, 45, 49, 49, 45, 50, 57,
+            32, 49, 50, 58, 48, 48, 58, 48, 57, 32, 85, 84, 67, 160, 6, 6, 168, 6, 1,
+        ];
+        let p: pb::TableCopiedFileInfo =
```
drmingdrmer, line 518: Add a version suffix to let the future you know what the time it is to remove a test case when a feature is no longer supported.

src/meta/app/src/schema/table.rs
```diff
@@ -616,3 +616,47 @@ pub struct CountTablesReply {
...
     pub table_id: u64,
 }
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq, Default)]
+pub struct TableCopiedFileNameIdent {
+    pub tenant: String,
+    pub db_id: u64,
+    pub table_id: u64,
+    pub file: String,
+}
```
drmingdrmer, line 626: Does a file name identity need `tenant` or `db_id`?

In my opinion, it is a table-specific concept.

src/meta/app/src/schema/table.rs
```diff
@@ -616,3 +616,45 @@ pub struct CountTablesReply {
...
+    pub expire_at: Option<u64>,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct UpsertTableCopiedFileReply {}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct TruncateTableReq {
+    pub table: TableNameIdent,
+}
```
drmingdrmer, line 657: IMHO, truncate-table should use table-id as the identifier.

According to the current implementation of truncate-table, it uses table-id plus a seq to update the table-meta(The same for other XXXReq structs: using table-name as identifier may restrict the calling layer from accessing a specific table by id+ver):
https://github.com/datafuselabs/databend/blob/dae90d856e380ea29716e87148cc69d07ccff8ff/src/query/storages/fuse/src/operations/truncate.rs#L31-L83

src/meta/app/src/schema/table.rs
```diff
@@ -616,3 +616,45 @@ pub struct CountTablesReply {
...
+pub struct GetTableCopiedFileReply {
+    pub file_info: BTreeMap<String, TableCopiedFileInfo>,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct UpsertTableCopiedFileReq {
+    pub table: TableNameIdent,
+    pub file_info: BTreeMap<String, TableCopiedFileInfo>,
+    pub expire_at: Option<u64>,
+}
```
drmingdrmer, line 649: GetTableCopiedFileReq and UpsertTableCopiedFileReq will be convert to using table-id, too?

src/meta/app/src/schema/table.rs
```diff
@@ -643,7 +643,7 @@ pub struct GetTableCopiedFileReply {
 
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct UpsertTableCopiedFileReq {
-    pub table: TableNameIdent,
+    pub table_id: u64,
```
drmingdrmer, line 646: For GetTableCopiedFileReq, UpsertTableCopiedFileReq and TruncateTableReq, to use table id, it still needs a `seq` to identify a specific version of table, i.e., use `TableIdent`.

lichuang: @BohuTANG does the table copied files meta need to precise to table version? Or just to table id is enough?

If it needs to precise to table version, it means that query layer need some transaction too:

1. get the table id and seq.
2. operate with table id and seq.
3. if get table seq in meta is not match to the seq passed, goto step 1 and try again...

lichuang: add `TableCopiedFileLock` to make sure consistent.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2022,23 +2267,28 @@ async fn gc_dropped_table(
...
             for key in remove_table_keys.iter() {
+                // remove table id keys not changed
+                condition.push(txn_cond_seq(&key.0, Eq, key.1));
+
+                // remove out of time table meta
                 if_then.push(txn_op_del(&key.0));
+
+                // remove stage file info of the table
+                remove_table_copied_files(kv_api, key.0.table_id, &mut condition, &mut if_then)
+                    .await?;
```
drmingdrmer, line 2284: It looks like it does not assert the seq of the TableCopiedFileLock.

BTW consider moving asserting seq of TableCopiedFileLock into `remove_table_copied_files()`?

drmingdrmer (commented):
Everything else looks good to me.

drmingdrmer (approved):

PR #7566, opened 2022-09-11
fix(query): `scan_progress` should be incr before prewhere filter.

### LINKED ISSUE #7565: How many rows a query really read after prewhere optimization is wrong.
**Summary**
Currently,  the count of rows read in a query is calculated as how many rows dose Fuse Engine produce.

Like:

```
mysql> select * from t where a > 5000 and b < 5003 ;
+------+------+
| a    | b    |
+------+------+
| 5001 | 5002 |
| 5001 | 5002 |
+------+------+
2 rows in set (0.16 sec)
Read 2 rows, 16.00 B in 0.102 sec., 19.62 rows/sec., 156.98 B/sec.
```

In fact, this query read 17500 rows, but most of the rows are filtered by Fuse Engine. So, it should be like:

```
mysql> select * from t where a > 5000 and b < 5003 ;
+------+------+
| a    | b    |
+------+------+
| 5001 | 5002 |
| 5001 | 5002 |
+------+------+
2 rows in set (0.16 sec)
Read 17500 rows, 136.72 KiB in 0.102 sec., 172.57 thousand rows/sec., 1.31 MiB/sec.
```

RinChanNOWWW:

## Summary

Summary about this PR

Fixes #7565


BohuTANG:
@mergify update

sundy-li (approved):

BohuTANG:
@mergify update

PR #7571, opened 2022-09-12
feat(query): add function v2 repeat

sundy-li:

## Summary

Summary about this PR

Related #6766

### DIFF
--- a/src/query/functions-v2/tests/it/scalars/string.rs
+++ b/src/query/functions-v2/tests/it/scalars/string.rs
@@ -55,6 +55,7 @@ fn test_string() {
     test_char(file);
     test_soundex(file);
     test_ord(file);
+    test_repeat(file);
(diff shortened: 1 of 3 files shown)


src/query/functions-v2/tests/it/scalars/string.rs
```diff
@@ -688,6 +689,13 @@ fn test_ord(file: &mut impl Write) {
     run_ast(file, "ord('💖')", &[]);
 }
 
+fn test_repeat(file: &mut impl Write) {
+    run_ast(file, "repeat('3', NULL)", &[]);
```
andylokandy, line 693: What about adding a case for `too many times`

andylokandy (approved):
rest LGTM

BohuTANG (approved):

PR #7522, opened 2022-09-08
feat(query): support jsonb format

b41sh:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Support `jsonb` format.

Part of #6994

PR #7572, opened 2022-09-12
feat(expr): support variant type

andylokandy:

## Summary

Value of variant type is stored in `bson` byte slice. When evaluating the variant value, it should be parsed just in time.

We're going to move to jsonb encode eventually, but since it's unimplemented currently, bson is used as a placeholder. cc @b41sh 

Related to https://github.com/datafuselabs/databend/pull/7522
Fixes https://github.com/datafuselabs/databend/issues/6634


src/query/functions-v2/tests/it/scalars/testdata/variant.txt
```diff
@@ -0,0 +1,64 @@
...
+optimized expr : NULL
+output type    : NULL
+output domain  : {NULL}
+output         : NULL
+
+
+error: 
+  --> SQL:1:1
+  |
+1 | parse_json('nuLL')
```
andylokandy, line 13: @sundy-li Is this expected?

src/query/functions-v2/tests/it/scalars/testdata/variant.txt
```diff
@@ -0,0 +1,64 @@
...
+  | ^^^^^^^^^^^^^^^^^^ unable to parse 'nuLL': expected ident at line 1 column 3
```
andylokandy, line 14: @sundy-li Is this expected?

sundy-li: Yes

src/query/expression/src/types/variant.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// BSON bytes representation of `{v: null}`.
+pub const DEFAULT_BSON: &[u8] = &[0x08, 0x00, 0x00, 0x00, 0x0A, 0x76, 0x00, 0x00];
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct VariantType;
+
+impl ValueType for VariantType {
+    type Scalar = Vec<u8>;
+    type ScalarRef<'a> = &'a [u8];
```
b41sh, line 37: Can we use `RawDocumentBuf` and `RawDocument` here? They are wrappers of `Vec<u8>` and `[u8]`.
```rust
pub struct RawDocumentBuf {
    data: Vec<u8>,
}
pub struct RawDocument {
    data: [u8],
}
```

leiysky (approved):

src/query/expression/src/types/variant.rs
andylokandy: There are two reasons `RawDocument` is not used for `Scalar` here:

1.  I reused `StringColumn` for variant columns, it becomes difficult to wrap the result from `StringColumn` into `RawDocument`.
2. 2, the scalar function can easily wrap inputs into `RawDocument` on the fly.

b41sh: ok

b41sh (approved):

PR #7593, opened 2022-09-13
refactor: Old Planner Never See Again (Part X)

Xuanwo:

## Summary

This PR is trying to remove the `ExprParser`.

After this PR, our storage will don't depend on `sqlparser` anymore.

This PR introduces a dangerous function, as I said in comment:

```rust
/// unchecked_expressions_analyze will analyzer given expr str into `Expression`.
///
/// `unchecked` means there is no correction checks for expr:
///
/// - We don't if the expr is **safe**.
/// - We don't if the expr is **existed**.
/// - We don't if the expr is **valid**.
///
/// We just convert the given expr into Expression, and hoping everything works.
///
/// This function should only be used by very limited use-cases like fuse engines
/// `cluster_keys` which already checked by our parsers.
///
/// **FBI WARNING: DON'T USE THIS FUNCTION UNLESS YOU KNOW WHAT I MEAN**
///
/// Any usage to this function should be reviewed carefully.
///
/// # Future Plan
///
/// This function should be removed once our new expression framework is ready.
pub fn unchecked_expressions_analyze(expr: &str) -> Result<Vec<Expression>> { ... }
```


leiysky (approved):

src/query/storages/fuse/src/table_functions/clustering_informations/table_args.rs
```diff
@@ -42,7 +41,7 @@ pub fn parse_func_table_args(table_args: &TableArgs) -> Result<(String, String)>
 pub fn get_cluster_keys(table: &FuseTable, definition: &str) -> Result<Vec<Expression>> {
     let cluster_keys = if !definition.is_empty() {
         let schema = table.schema();
-        let exprs = ExpressionParser::parse_exprs(definition)?;
+        let exprs = unchecked_expressions_analyze(definition)?;
```
sundy-li, line 44: what if `cluster by (to_date(ts),  -a, b )`

Xuanwo: `unchecked_expressions_analyze` only supports limited `Expr` types. We can't support complex cluster keys like `(to_date(ts), -a, b)` so far.

`TypeChecker` and `ExpressionBuilder` can parse those expr. But they need `QueryContext` which is not provided in `storage`. We need a refactor here.

Xuanwo: The test cases of `hits` also failed. Maybe this PR is going the wrong way...

Xuanwo:
`unchecked_expressions_analyze` can't parse complex cluster key, I'm seeking new ways to address this problem.

Xuanwo:
@mergify update

Xuanwo:
This PR is blocked by `Planner`.

Now, planner is implemented inside `query/service/src/sql/planner`, and It depends on types like `crate::evaluator::Evaluator` to parse `DataBlock` at bind time (aka, `ValueSourceV2`). Thus, we can't move `Planner` out to be a standalone crate `common-planner`.

However, `common-storages-fuse` needs `Planner` to parse exprs correctly (and safely).

Cc @leiysky @sundy-li

sundy-li:
> Ideally, we should not parse DataBlock during bind. Can we move that logic to interpreter-insert?

Yes,  I think it'll be  done in #7613

Xuanwo:
> Yes, I think it'll be done in #7613

Cc @youngsofun for confirmation

sundy-li:
I'll try to apply a fast hot pr to improve this.

Xuanwo:
This PR is out-of-date, I will start a new to replace it.

PR #7600, opened 2022-09-14
refactor: Make planner depends on TableContext trait

Xuanwo:

## Summary

This PR depends on https://github.com/datafuselabs/databend/pull/7597. Please review after #7597 is merged.

---

This PR makes the planner depends on the `TableContext` trait instead of the `QueryContext` struct.

Make space for the future refactor.

### DIFF
--- a/src/query/catalog/src/table_context.rs
+++ b/src/query/catalog/src/table_context.rs
@@ -111,5 +111,7 @@ pub trait TableContext: Send + Sync {
     // Get user manager api.
     fn get_user_manager(&self) -> Arc<UserApiProvider>;
     fn get_cluster(&self) -> Arc<Cluster>;
+    async fn get_table(&self, catalog: &str, database: &str, table: &str)
+    -> Result<Arc<dyn Table>>;
     async fn get_processes_info(&self) -> Vec<ProcessInfo>;
 }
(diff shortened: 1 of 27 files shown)


PsiACE (approved):
lgtm

src/query/catalog/src/table_context.rs
```diff
@@ -111,5 +111,7 @@ pub trait TableContext: Send + Sync {
     // Get user manager api.
     fn get_user_manager(&self) -> Arc<UserApiProvider>;
     fn get_cluster(&self) -> Arc<Cluster>;
+    async fn get_table(&self, catalog: &str, database: &str, table: &str)
```
leiysky, line 114: Maybe we don't need this method here? Since we usually get table from `dyn Catalog`, which can be retrieved by `get_catalog`.

Xuanwo: > Maybe we don't need this method here? Since we usually get table from `dyn Catalog`, which can be retrieved by `get_catalog`.

`QueryContext` has some internal logic to keep the returning `Table` consistent throughout the query.

https://github.com/datafuselabs/databend/blob/f2a8f0da2e966376b954bf7e357ff0c4ce86704f/src/query/service/src/sessions/query_ctx_shared.rs#L190-L210

It's OK to remove it if we have other ways to make sure that.

Xuanwo: We do need some refactor for our `Catalog` and `TableContext` to make the behavior consistent.

Cc @dantengsky @zhang2014

zhang2014 (approved):

leiysky (approved):

PR #7644, opened 2022-09-15
fix: Update env config of S3

xychu:

## Summary

Fix k8s deploy env config of S3

Fixes #issue


Xuanwo:
The `STORAGE_S3_XXX` is the correct style. Maybe we should update all `S3_XXX` to `STORAGE_S3_XXX` instead.

xychu:
`databend-query` try to use configs of `s3.storage.xxxx` under `StorageConfig`.

https://github.com/datafuselabs/databend/blob/main/src/query/config/src/outer_v0.rs#L176-L211

xychu:
> The `STORAGE_S3_XXX` is the correct style. Maybe we should update all `S3_XXX` to `STORAGE_S3_XXX` instead.

Got it. 

I failed to get query start while following this doc:
https://databend.rs/doc/deploy/deploying-databend-on-kubernetes#step-3-deploy-a-databend-query-cluster

After some dig, I found that `S3_STORAGE_XXX` instead of `STORAGE_S3_XXX` make `databend-query` happy and running.

Xuanwo:
> I failed to get query start while following this doc:
> https://databend.rs/doc/deploy/deploying-databend-on-kubernetes#step-3-deploy-a-databend-query-cluster

This docs `query-cluster.yaml` refers to a very old version of the databend. Thus the config is incorrect.

Please use our latest image: `datafuselabs/databend-query:v0.6.100-nightly`

Xuanwo:
Cc @everpcpc, how about release a docker image with tag `datafuselabs/databend-query:latest-nightly` so that we can refer our k8s scripts?

Xuanwo:
@xychu Thanks for reporting. This PR will be closed as it's not the real case. 

But maybe you will be interested in contributing a patch to change

https://github.com/datafuselabs/databend/blob/8337678ad9b5473c3a8428fadbff0b967d506724/scripts/kubernetes/query-cluster.yaml#L167

into `datafuselabs/databend-query:latest` (added in https://github.com/datafuselabs/databend/pull/7648)

PR #7645, opened 2022-09-15
refactor(query): improve in function

sundy-li:

## Summary

Summary about this PR

```

select count() from numbers(1000000000) where number::String not in ('33', '555', '6666');


databend: 2.570 sec 
databend-in-improve: 1.192 sec

```

Fixes #issue


src/query/functions/src/scalars/conditionals/in_basic.rs
```diff
@@ -79,16 +81,66 @@ macro_rules! scalar_contains {
...
+            builder.append(valid && ((contains && !NEGATED) || (!contains && NEGATED)));
+        }
+        return Ok(builder.build($ROWS));
+    }};
+}
+
+macro_rules! string_contains {
+    ($T: ident, $INPUT_COL: expr, $ROWS: expr, $COLUMNS: expr, $CAST_TYPE: ident, $FUNC_CTX: expr) => {{
+        let mut builder: ColumnBuilder<bool> = ColumnBuilder::with_capacity($ROWS);
+        let mut vals_set: HashSetWithStackMemory<64, KeysRef> = HashSetWithStackMemory::create();
```
sandflee, line 109: using stack-allocated hashset,  if user input a sql with too many args , this may cause stack overflow?  
sql likes 
```
select * from xxx where id in (xx,xx, ...)
```

src/query/functions/src/scalars/conditionals/in_basic.rs
```diff
@@ -79,16 +81,66 @@ macro_rules! scalar_contains {
...
+            let col = cast_column_field(col, col.data_type(), &$CAST_TYPE, &$FUNC_CTX)?;
+            let col_viewer = $T::try_create_viewer(&col)?;
+            if col_viewer.valid_at(0) {
+                let val = col_viewer.value_at(0);
+                let key = KeysRef::create(val.as_ptr() as usize, val.len());
+                vals_set.insert_key(&key, &mut inserted);
+            }
+        }
+        let input_viewer = $T::try_create_viewer(&$INPUT_COL)?;
+        for (row, val) in input_viewer.iter().enumerate() {
```
sandflee, line 121: if `INPUT_COL ` is a const-column,  we could erase the iter() ?

sandflee: this may happens in hive storage,  the partition column is virtual and const

src/query/functions/src/scalars/conditionals/in_basic.rs
sundy-li: HashSetWithStackMemory will fallback into heap allocation if it overflows.

src/query/functions/src/scalars/conditionals/in_basic.rs
sundy-li: `input_viewer` already handles the constant cases.

See:

```
 #[inline]
    fn value_at(&self, index: usize) -> bool {
        self.values.get_bit(index & self.non_const_mask)
    }
```

sandflee: cool !

sandflee: correct me if I mis-understanding,  if INPUT_COL is const-column, we could rewrite code like this, seems more effective.
```
    let val = input_viewer.value_at(0);
    let contains = vals_set.contains(val);
    let valid = input_viewer.valid_at(0);
  // construct a bool const collumn
    return constColumn::new(contains,  xx)
```

sandflee:
hash table is a generic and effective way to justice the equality of values.  I wondered if there are more simple and hack way to do this if there are  one or two args for infunction.
some sqls are like:
```
select xx from xx 
where p_date between '20210216' and '20210217'
first_industry_name in ('xx',
                'xx')
   and delivery_system in ('FANSTOP',
                           'ECOM_NIU') --分产品
 and resource_type in('INCENTIVE_LIVE')
group by xx
```

src/query/functions/src/scalars/conditionals/in_basic.rs
sundy-li: Yes, so we set `self.non_const_mask` to zero for const column, the index will always be zero after `index & 0`.

sundy-li:
> hash table is a generic and effective way to justice the equality of values. I wondered if there are more simple and hack way to do this if there are one or two args for infunction. 

It should evaluate in function multiple times and combine the boolean filters into one for now.

But we can have a optimization that build the Set once and probe it in each block.

src/common/hashtable/src/keys_ref.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use ahash::AHasher;
+
+use super::HashTableKeyable;
+
+#[derive(Clone, Copy)]
+pub struct KeysRef {
+    pub length: usize,
+    pub address: usize,
+}
```
RinChanNOWWW, line 25: Can src/query/functions-v2/src/aggregates/aggregate_distinct_state.rs also use this struct? It seems there is a duplication.

sundy-li: Yes, they are moved there

RinChanNOWWW: I didn't note this, sorry.

BohuTANG:
@mergify update

RinChanNOWWW:
I have a question. Why we `wrap_nullable` and `remove_nullable` here? Can we `remove_null` directly?

https://github.com/datafuselabs/databend/blob/e4c371cf49830765e4abcca775248bf030e9e7ce/src/query/functions/src/scalars/conditionals/in_basic.rs#L145-L151

Xuanwo:
Dev Linux / test_sqllogic_base_standalone_linux (pull_request) failed for timeout.

Seems `02_0045_conditional_function_in` is wrong.

sundy-li:
> I have a question. Why we wrap_nullable and remove_nullable here? Can we remove_null directly

`remove_nullable` is to get a datatype to match, `least_super_dt` is used as a common super type for both sides. We can't remove null directly because `remove null ` will result to a column with default value.

src/query/functions/src/scalars/conditionals/in_basic.rs
```diff
@@ -280,7 +384,8 @@ impl<const NEGATED: bool> Function for InFunction<NEGATED> {
                     input_rows,
                     columns,
                     least_super_dt,
-                    func_ctx
+                    func_ctx,
+                    return_nullable
                 )
             }
             _ => Result::Err(ErrorCode::BadDataValueType(format!(
```
RinChanNOWWW, line 391: How about the other types? Will they be supported in the future?

Xuanwo (approved):
Nice!

RinChanNOWWW (approved):

PR #7678, opened 2022-09-17
refactor(meta): remove MetaError::Fatal.

drmingdrmer:

##### refactor(meta): remove MetaError::Fatal.

Crate meta/types: remove error `MetaError::Fatal`. Usages in test are
replaced with panic. Usages in meta-service startup is replaced with
`AnyError`, since no one really cares about the type of this error.

Crate meta/api:

- Refactor: meta/api: rename mod `kv_api_utils` to `util`.

- Refactor: move `get_kv_data()` to mod `testing`, it is only used for
  tests, and will panic if the expected key does not exist.

## Related Issues


Xuanwo:
@mergify update

drmingdrmer:
> @mergify update

Glad to see you back mergify bot!

PR #7685, opened 2022-09-17
refactor(meta): extract KVAppError from MetaError

drmingdrmer:

##### refactor(meta): extract KVAppError from MetaError

`MetaError` is too big: it includes every every that relates to
meta-store. Actually there are two levels of errors:
- The errors raised by a mod that relies on meta-store, such as
  `SchemaApi`,
- and the errors raised by meta-store itself.

Thus to make errors more specific `KVAppError` is extracted out of
`MetaError`, which is an error raised by applications that rely on
KVApi, such as SchemaApi.

Changes:

- Replace `MetaError` with `KVAppError` for `KVApi` implemenations and the
  traits that rely on `KVApi`.

- Remove `AppError` from `MetaError`: meta-store should not be aware of application level errors.

## Related Issues


Xuanwo (approved):

Xuanwo:
I'm a bit worried about different error types in meta. Will we have an overview docs about relation between them>

BohuTANG (approved):

PR #7686, opened 2022-09-17
fix(query): fix cast deterministic error

### LINKED ISSUE #7682: bug: `cast(rand() as string)` returns the same result
**Summary**

```
mysql> select rand() from numbers(10);
+----------------------+
| rand()               |
+----------------------+
| 0.016333804973599486 |
|  0.49383331056312574 |
|   0.9841902151336861 |
| 0.043982804663117325 |
|   0.2259206402471785 |
|  0.00997215327383083 |
|   0.7977313791817102 |
|   0.7871726221864119 |
|   0.9605240485760037 |
|  0.30119168210376523 |
+----------------------+
10 rows in set (0.01 sec)
Read 10 rows, 80.00 B in 0.002 sec., 4.35 thousand rows/sec., 34.02 KiB/sec.

mysql> select cast(rand() as string) from numbers(10);
+------------------------+
| cast(rand() as string) |
+------------------------+
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
| 0.9355074606222086     |
+------------------------+
```

b41sh:

## Summary

fix cast deterministic error

Fixes #7682


src/query/service/src/sql/planner/plans/scalar.rs
```diff
@@ -511,7 +511,7 @@ impl ScalarExpr for CastExpr {
     }
 
     fn is_deterministic(&self) -> bool {
-        true
+        false
```
sundy-li, line 514: `Cast` should be deterministic.

`random` is not deterministic, so `cast(random())` is not deterministic.  We should find the root reason.

sundy-li: should be `self.argument.is_deterministic()`

b41sh: ok

sundy-li (approved):

Xuanwo (approved):

src/query/service/src/sql/planner/plans/scalar.rs
```diff
@@ -511,7 +511,7 @@ impl ScalarExpr for CastExpr {
     }
 
     fn is_deterministic(&self) -> bool {
-        true
+        self.argument.is_deterministic()
```
Xuanwo, line 514: Please add a sqllogic test for this.

Xuanwo: We can compare the result of two `cast(rand() as string)`

BohuTANG (approved):

PR #7687, opened 2022-09-18
chore(query): unify the accessor check

BohuTANG:

## Summary

* Unify the accessor check: move all checks to the `AccessorChecker` trait for the interpreter
* Move up the `cargo clippy` in Makefile, to make the cargo work if the `yapf` is not work.

Fixes #issue

### DIFF
--- a/src/query/service/src/interpreters/interpreter_factory_interceptor.rs
+++ b/src/query/service/src/interpreters/interpreter_factory_interceptor.rs
@@ -23,7 +23,7 @@ use common_streams::ProgressStream;
 use common_streams::SendableDataBlockStream;
 use parking_lot::Mutex;
 
-use crate::interpreters::access::ManagementModeAccess;
+use crate::interpreters::access::Accessor;
 use crate::interpreters::Interpreter;
 use crate::interpreters::InterpreterPtr;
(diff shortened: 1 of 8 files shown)


src/query/service/src/interpreters/interpreter_factory_interceptor.rs
```diff
@@ -75,11 +75,8 @@ impl Interpreter for InterceptorInterpreter {
     }
 
     async fn execute(&self, ctx: Arc<QueryContext>) -> Result<SendableDataBlockStream> {
-        // Management mode access check.
-        match &self.new_plan {
-            Some(p) => self.management_mode_access.check_new(p)?,
-            _ => self.management_mode_access.check(&self.plan)?,
-        }
+        // Access check.
+        self.accessor_checker.check(&self.new_plan, &self.plan)?;
```
zhang2014, line 79: Maybe move to InterpreterFactory is better?

BohuTANG: > Maybe move to InterpreterFactory is better? 
Looks better to add them to `create_interpreter`.

>  We can ignore queries without permission in the query_log.

Does it mean we don't log the error query in the query_log?

Issue #7516, opened 2022-09-07
Feature: Versioning Databend's Documents


PsiACE:
**Summary**

After the release of 0.8, we plan to version the documentation.

We now plan to use a similar approach to docusaurus:
- keeping only the latest version
- providing permanent links to older versions

ref: https://docusaurus.io/

Xuanwo:
Let's defer this issue until we release v0.9

https://github.com/datafuselabs/databend/pull/7688#pullrequestreview-1111485359

PsiACE:
> Let's defer this issue until we release v0.9

Perhaps it should be tagged with `v0.9` to hint that we need to finish the job eventually.

Xuanwo:
> Perhaps it should be tagged with `v0.9` to hint that we need to finish the job eventually.

Added into v0.9 milestone.

PR #7688, opened 2022-09-18
docs: Add versioned dropdown

Xuanwo:

## Summary

Fix https://github.com/datafuselabs/databend/issues/7516

image

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -40,12 +40,6 @@ const config = {
                     routeBasePath: 'doc',
                     sidebarPath: require.resolve('../docs/doc/sidebars.js'),
                     editUrl: 'https://github.com/datafuselabs/databend/edit/main/databend',
-                    lastVersion: 'current',
-                    versions: {
-                        current: {
-                            label: `v0.8.x 🚧`,
-                        },
-                    },
                 },
                 blog: {
                     showReadingTime: true,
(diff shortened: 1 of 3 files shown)


website/docusaurus.config.js
```diff
@@ -90,16 +90,30 @@ const config = {
                     src: 'img/logo/logo-no-text.svg',
                 },
                 items: [
+                    {
+                        type: 'docsVersionDropdown',
+                        position: 'right',
+                        dropdownItemsAfter: [
+                            {
+                                href: 'https://databend-2pvgtjnrv-databend.vercel.app',
+                                label: 'v0.7.172-nightly',
```
BohuTANG, line 99: Why v0.7.172-nightmy?

BohuTANG (approved):

website/docusaurus.config.js
Xuanwo: The latest tag for `v0.7` track.

BohuTANG: I think we should not add 0.7 or 0.6 here, because we don't want user to use them, suggestions keep one 0.8 and next version

Xuanwo: So you wanna to have `v0.8.0-nightly` and `Next` here?

BohuTANG: What about the latest version: v0.8.38-nightly and the next?

Xuanwo: They are nearly the same (only one day's commits)

Xuanwo: Or we just defer this issue until we released `v1.0` and established a release policy?

BohuTANG: OK, let's consider enable the version after 0.9 released, many users are using v0.8 :)

PR #7694, opened 2022-09-18
chore(query): move insert format deser from binder into interpreter

sundy-li:

## Summary

Summary about this PR

move insert format deser from binder into interpreter

Related #7593


src/query/service/src/sql/planner/binder/insert.rs
```diff
@@ -12,54 +12,25 @@
...
-use common_datavalues::TypeDeserializerImpl;
-use common_exception::ErrorCode;
 use common_exception::Result;
-use common_formats::FormatFactory;
-use common_io::prelude::BufferReader;
-use common_io::prelude::*;
-use common_pipeline_transforms::processors::transforms::Transform;
 use tracing::debug;
 
 use crate::clusters::ClusterHelper;
```
Xuanwo, line 24: Do we need `ClusterHelper`?

youngsofun (approved):

src/query/service/src/sql/planner/binder/insert.rs
sundy-li: enable_distributed_optimization: !self.ctx.get_cluster().is_empty(),
   |                                                                              ^^^^^^^^ method not found in `std::sync::Arc<Cluster>`

BohuTANG (approved):

PR #7718, opened 2022-09-19
feat(query): fix setting value as_string

sundy-li:

## Summary

Summary about this PR

Fixes #issue


Xuanwo (approved):

Xuanwo:
Maybe we should support auto convert for `as_u64`.

sundy-li:
> Maybe we should support auto convert for `as_u64`.

Now `as_string` is used even if it's u64.  We can support it if we need in the future.

BohuTANG (approved):

Issue #7561, opened 2022-09-10
feat: add FORCE to copyOptions


BohuTANG:
**Summary**

```
copyOptions ::=
  [ SIZE_LIMIT = <num> ]
  [ PURGE = <bool> ]
  [ FORCE = TRUE | FALSE ]
```

`FORCE = TRUE | FALSE`

TRUE to load all files, FALSE not, duplicating data in a table.
Default: FALSE

Use case:
When users want to force load all files to a table, even the stage file meta is available in the metasrv.

PR #7719, opened 2022-09-19
feat: idempotent-copy file

lichuang:

## Summary

1. idempotent-copy file
2. add `force` option in `copy into` SQL.
3. add `truncate_table` in catalog to truncate all the copied files info in meta service.
4. add test for idempotent-copy and force copy.

Fixes https://github.com/datafuselabs/databend/issues/7550, https://github.com/datafuselabs/databend/issues/7561


src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -49,6 +54,95 @@ impl CopyInterpreterV2 {
...
+            files: files.to_owned(),
+        };
+        let mut file_map = BTreeMap::new();
+
+        if !force {
+            let resp = catalog.get_table_copied_file_info(req).await?;
+            for file in files.iter() {
+                let stage_file = list_files(&self.ctx, &table_info.stage_info, file, "").await?;
+                if stage_file.is_empty() {
+                    continue;
```
BohuTANG, line 83: Should add a `warn!` log to remind the file is skipped.

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -49,6 +54,95 @@ impl CopyInterpreterV2 {
...
+                }
+
+                let stage_file = &stage_file[0];
+
+                if let Some(file_info) = resp.file_info.get(file) {
+                    // No need to copy the file again if etag is_some and match.
+                    if stage_file.etag.is_some() && stage_file.etag == file_info.etag {
+                        continue;
```
BohuTANG, line 91: Should add a warn! log to remind the file&meta is skipped.

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -49,6 +54,95 @@ impl CopyInterpreterV2 {
...
+                    }
+
+                    if file_info.content_length == stage_file.size
+                        && file_info.last_modified == Some(stage_file.last_modified)
+                    {
```
BohuTANG, line 96: When do we fall in this case?

BohuTANG: If the ETag is the same, does the check is not need?

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -320,14 +415,59 @@ impl Interpreter for CopyInterpreterV2 {
...
+                                &files,
+                            )
+                            .await?;
+
+                        tracing::info!(
+                            "matched copy unduplicate files: {:?}",
+                            &copy_stage_files.keys(),
+                        );
+
+                        if copy_stage_files.is_empty() {
```
BohuTANG, line 437: It looks better to check the empty in `copy_files_to_table` function.

BohuTANG:
@mergify update

src/query/service/src/interpreters/interpreter_copy_v2.rs
lichuang: > If the ETag is the same, does the check is not needed?

yes, now the logic is:

1. compare etag, if etag is_some and matched, ignore copy this file.
2. compare content-length and last_modified, if matched, ignore copy this file.

BohuTANG: Do not get the point:
If the etag is the same why need to check content_length&size again?

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -49,6 +54,108 @@ impl CopyInterpreterV2 {
...
+                file_map.insert(file.clone(), TableCopiedFileInfo {
+                    etag: stage_file.etag.clone(),
+                    content_length: stage_file.size,
+                    last_modified: Some(stage_file.last_modified),
+                });
+            }
+        } else {
+            // if force is true, copy all the file.
+            for file in files.iter() {
+                let stage_file = list_files(&self.ctx, &table_info.stage_info, file, "").await?;
```
Xuanwo, line 123: Don't use `list_files` to get the file's metadata. We can implement a function like the following:

```rust
async fn stat_file(ctx: &Arc<QueryContext>, stage: &UserStageInfo, path: &str) -> Result<StageFile>`
```

In which we can use `op.object(path).metadata().await`.

BohuTANG (approved):

BohuTANG:
@soyeric128 
Please update the copy command, we introduced a new option: FORCE

PR #7722, opened 2022-09-19
refactor(processor): save pipeline executor into query context

zhang2014:

## Summary

refactor(processor): save pipeline executor into query context

Fixes #issue

### DIFF
--- a/src/query/service/src/storages/storage_table_read_wrap.rs
+++ b/src/query/service/src/storages/storage_table_read_wrap.rs
@@ -46,13 +46,11 @@ impl<T: Table> TableStreamReadWrap for T {
         self.read2(ctx.clone(), plan, &mut pipeline)?;
 
         let settings = ctx.get_settings();
-        let query_need_abort = ctx.query_need_abort();
         pipeline.set_max_threads(settings.get_max_threads()? as usize);
         let executor_settings = ExecutorSettings::try_create(&settings)?;
 
-        let executor =
-            PipelinePullingExecutor::try_create(query_need_abort, pipeline, executor_settings)?;
-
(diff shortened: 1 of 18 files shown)


src/query/service/src/storages/storage_table_read_wrap.rs
```diff
@@ -46,13 +46,11 @@ impl<T: Table> TableStreamReadWrap for T {
...
         let settings = ctx.get_settings();
-        let query_need_abort = ctx.query_need_abort();
         pipeline.set_max_threads(settings.get_max_threads()? as usize);
         let executor_settings = ExecutorSettings::try_create(&settings)?;
 
-        let executor =
-            PipelinePullingExecutor::try_create(query_need_abort, pipeline, executor_settings)?;
-
+        let executor = PipelinePullingExecutor::try_create(pipeline, executor_settings)?;
+        ctx.set_executor(Arc::downgrade(&executor.get_inner()));
```
BohuTANG, line 53: It's hard to make sure someone to set the executor or what will happen if not set?
It seems we can add a new ctx function to create and set executor?

zhang2014: > It's hard to make sure someone to set the executor or what will happen if not set_executor? It seems we can add a new ctx function to create and set_executor if it needs set?

We will only create executor (pulling executor, pushing executor, complete executor) in `interpreter::execute`. so, it's fine?

`recluster interpreter` and `optimize interpreter` will not be create executor after the next pr(other will only be used in the test and will not have any effect).

zhang2014: > or what will happen if not set_executor? 

Cannot kill query if not set_executor

BohuTANG (approved):

PR #7725, opened 2022-09-19
feat(query): enable range filter to pre-filter `is_null` and `is_not_null` ops.

### LINKED ISSUE #7336: Feature: nullable pruner in fuse engine
**Summary**

Description for this feature.

For example

```
select a, b from t where a is null and b is not null
```

We can use the `null_count` in  [`ColumnStatistics`](https://github.com/datafuselabs/databend/blob/fc868936df12e06f5c907c335ac813e950c3eb8b/src/query/storages/fuse-meta/src/meta/common.rs)  to prune the partition

```
#[derive(serde::Serialize, serde::Deserialize, Debug, Clone, Eq, PartialEq)]
pub struct ColumnStatistics {
    pub min: DataValue,
    pub max: DataValue,

    pub null_count: u64,
    pub in_memory_size: u64,
}

```

RinChanNOWWW:

## Summary

Summary about this PR

Fixes #7336


dantengsky (approved):
LGTM

sundy-li (approved):

PR #7739, opened 2022-09-19
refactor(function): make unit test goldenfile only display the used columns

andylokandy:


src/query/functions-v2/tests/it/scalars/mod.rs
```diff
@@ -117,6 +117,21 @@ pub fn run_ast(file: &mut impl Write, text: &str, columns: &[(&str, DataType, Co
                 Value::Column(output_col) => {
                     test_arrow_conversion(&output_col);
 
+                    // Only display the used input columns
+                    let used_columns = expr.column_refs();
+                    let input_domains = input_domains
```
sundy-li, line 122: some functions should generate result column with row_size even without args.

For example:

`today()`, `random()`

So we should keep at least one column.


And we should refactor in  collect_columns function in https://github.com/datafuselabs/databend/blob/b3c2cb5b8a15e5f5c9c6f10f51edaa0ce162b145/src/query/functions-v2/tests/it/aggregates/mod.rs

andylokandy: > some functions should generate result column with row_size even without args.

ok, let's do it in a separate pr. BTW, why `today()` needs the row size either?

b41sh (approved):

src/query/functions-v2/tests/it/scalars/mod.rs
sundy-li: sorry, `today` does not need.

sundy-li (approved):

src/query/functions-v2/tests/it/aggregates/mod.rs
```diff
@@ -103,15 +104,15 @@ pub fn run_agg_ast(file: &mut impl Write, text: &str, columns: &[(&str, DataType
...
                         } else {
-                            vec![0]
+                            HashSet::from_iter([0])
                         }
                     }
-                    false => column_ids,
+                    false => used_columns,
                 };
 
                 for id in ids.iter() {
```
sundy-li, line 118: the iterator is  in arbitrary order, which makes the result unstable.

sundy-li (approved):

b41sh:
`regexp.txt` also need to modify

PR #7751, opened 2022-09-20
refactor: Move plans from old planner to new

Xuanwo:

## Summary

This PR will move plans from the old planner to the new one.

Since this PR, no code change should be added in `common-legacy-planners` anymore.


Xuanwo:
@dantengsky In this PR, I did some refactoring on `truncate`, PTAL.

And I think we should not depend on `Plan` (like TruncateTablePlan) directly in our `Table` API. What do you think?

leiysky (approved):

sundy-li (approved):

BohuTANG (approved):

PR #7767, opened 2022-09-21
refactor: Old Planner Never See Again (Part 2)

Xuanwo:

## Summary

This PR removes `PlanNode` and `PlanParser`, which makes a big step towards removing `sqlparser-rs`.

Let's open the champagne to celebrate this moment!

Thanks @leiysky @andylokandy @sundy-li @zhang2014 

## Notes for async insert

Async insert is implemented on old planner which is not used after this PR.

The new async insert is implemented in https://github.com/datafuselabs/databend/blob/main/src/query/service/src/interpreters/async_insert_queue_v2.rs


BohuTANG:
The PR desc missing: why we removed the async insert part.

Xuanwo:
> The PR desc missing: why we removed the async insert part.

Added.

leiysky (approved):

sundy-li (approved):

PR #7768, opened 2022-09-21
[NM] support external location for tables

ariesdevil:

## Summary

Summary about this PR

Fixes #issue


src/query/service/src/sessions/query_ctx.rs
```diff
@@ -359,8 +362,14 @@ impl TableContext for QueryContext {
         self.shared.get_query_str()
     }
     // Get the storage data accessor operator from the session manager.
-    fn get_storage_operator(&self) -> Result<Operator> {
-        let operator = self.shared.storage_operator.clone();
+    fn get_storage_operator(&self, sp: Option<StorageParams>) -> Result<Operator> {
```
Xuanwo, line 365: I prefer the following API:

```rust
fn get_internal_storage_operator(&self) -> Operator
fn get_external_storage_operator(&self, sp: StorageParams) -> Result<Operator>
```

src/query/storages/fuse/src/io/read/meta_readers.rs
```diff
@@ -96,17 +115,28 @@ where T: BufReaderProvider + Sync + Send
...
 impl BufReaderProvider for &dyn TableContext {
-    async fn buf_reader(&self, path: &str, len: Option<u64>) -> Result<BufReader<BytesReader>> {
-        let operator = self.get_storage_operator()?;
+    async fn buf_reader(
+        &self,
+        path: &str,
+        len: Option<u64>,
+        sp: Option<StorageParams>,
+    ) -> Result<BufReader<BytesReader>> {
+        let operator = self.get_storage_operator(sp)?;
```
Xuanwo, line 139: Cloning an operator is cheap, but building an operator is far more expensive, especially for services like s3.

Try our best to reuse the same operator.

src/query/storages/fuse/src/operations/commit.rs
```diff
@@ -123,37 +124,42 @@ impl FuseTable {
...
-                        break Err(ErrorCode::OCCRetryFailure(format!(
-                            "can not fulfill the tx after retries({} times, {} ms), aborted. table name {}, identity {}",
-                            retry_times,
-                            Instant::now()
-                                .duration_since(backoff.start_time)
-                                .as_millis(),
-                            tbl.table_info.name.as_str(),
-                            tbl.table_info.ident,
-                        )));
+                Err(e) if self::utils::is_error_recoverable(&e, transient) => {
```
Xuanwo, line 127: OpenDAL has its logic for backoff. And I have a plan to refactor here. So we can return directly here to make things simpler.

src/meta/protos/proto/table.proto
```diff
@@ -124,6 +124,8 @@ message TableMeta {
 
   // Comments of each field.
   repeated string field_comments = 25;
+
+  optional string storage_params = 26;
```
Xuanwo, line 128: Store our config in `string` is very easy to break.

Maybe we can use `StageStorage`?

https://github.com/datafuselabs/databend/blob/3b78959ca483cbd62abc322faadd1a0c59b2fead/src/meta/protos/proto/user.proto#L131

Visit @drmingdrmer for advice.

drmingdrmer: Right, using a structured type would be better.

src/query/storages/fuse/src/pruning/bloom_pruner.rs
```diff
@@ -158,10 +165,17 @@ mod util {
         bloom_index_col_names: &[String],
         index_location: &Location,
         index_length: u64,
+        storage_params: Option<StorageParams>,
```
drmingdrmer, line 168: There is already a `dal` arg at line 162, why does it need `storage_params` at all?

I assume the `dal` can get any data it needs, without depending on any additional parameters.

This weirdness may be rooted in the previous design: the `dal` is not used but instead another **reader** is created.

PR #7789, opened 2022-09-21
feat: impl externalLocation for create table

### LINKED ISSUE #7297: feat: external location for fuse engine
**Summary**

If we `create table ... engine=fuse`and write some data, the progress like:
1. Put the table schema to meta service
2. Read config from `storage.s3` and create a storage operator
3. Write parquet file to storage
4. Commit the snapshot to meta service

This is how databend fuse engine work now: fuse engine data will write to the storage configured in the databend-query.toml.

If we support(The syntax only for proposal):
```
create table ... engine=fuse <externalLocation>

externalLocation ::=
  's3://<bucket>[<path>]'
  CONNECTION = (
        ENDPOINT_URL = 's3[gcp]://<endpoint-URL>'
        ACCESS_KEY_ID = '<your-access-key-ID>'
        SECRET_ACCESS_KEY = '<your-secret-access-key>'
        REGION = '<region-name>'
        ENABLE_VIRTUAL_HOST_STYLE = true|false
  )
```

Things are changed: we can write fuse engine data to an external location that is controlled by the user, and the user can read the parquet file created by the fuse engine since they are all parquet format!

ariesdevil:

## Summary

Impl externalLocation for create table

Fixes #7297


src/query/service/src/sql/planner/binder/ddl/table.rs
```diff
@@ -363,6 +366,24 @@ impl<'a> Binder {
...
+                    name: uri.name.clone(),
+                    path: uri.path.clone(),
+                    connection: uri.connection.clone(),
+                };
+                parse_uri_location(&uri)?
+            }
+            // todo(ariesdevil): get default config storage params and path.
+            None => {
+                let sp = self.ctx.get_storage_params();
+                let p = sp.get_root_path();
```
Xuanwo, line 382: How about using `op.metadata().root()`?

- https://opendal.databend.rs/opendal/struct.Operator.html#method.metadata
- https://opendal.databend.rs/opendal/struct.AccessorMetadata.html#method.root

BohuTANG:
Please add a stateful test (like minio external location).

src/common/storage/src/config.rs
```diff
@@ -116,6 +117,24 @@ impl StorageParams {
...
+            StorageParams::Obs(v) => v.root.clone(),
+            StorageParams::S3(v) => v.root.clone(),
+            StorageParams::Ipfs(v) => v.root.clone(),
+
+            StorageParams::Http(_) | StorageParams::Ftp(_) | StorageParams::Memory => {
+                "".to_string()
+            }
+
+            #[cfg(feature = "storage-hdfs")]
+            StorageParams::Hdfs(v) => v.root.clone(),
```
drmingdrmer, line 135: Not a issue for this PR:

`StorageParams` becomes two different types, if `Hdfs` can be toggled on and off by a feature. A binary without feature `storage-hdfs` can not read data written by a binary with feature `storage-hdfs`.

src/meta/app/src/schema/table.rs
```diff
@@ -165,6 +166,12 @@ pub struct TableStatistics {
     pub index_data_bytes: u64,
 }
 
+#[derive(serde::Serialize, serde::Deserialize, Default, Clone, Debug, Eq, PartialEq)]
+#[serde(default)]
+pub struct TableStorageParams {
+    pub storage: StorageParams,
+}
+
```
drmingdrmer, line 174: Why is it necessary to create a wrapper type for `StorageParam`?

Is it ok to use `StorageParam` directly?

src/meta/protos/proto/table.proto
```diff
@@ -57,6 +58,16 @@ message TableIdent {
...
 }
 
+//message TableStorage {
+//  oneof storage {
+//    S3StorageConfig s3 = 1;
+//    FsStorageConfig fs = 2;
+//    GcsStorageConfig gcs = 3;
+//  }
+//}
+
```
drmingdrmer, line 68: You can just remove these lines unless you'd like to have a monument for the code you've written. 🤔

src/common/storage/src/config.rs
Xuanwo: Tracked at https://github.com/datafuselabs/databend/issues/7936

src/meta/app/src/schema/table.rs
ariesdevil: Ok, but I prefer to add a wrapper since `StorageParam` is not only used by table meta.

src/query/service/src/sql/planner/binder/ddl/table.rs
```diff
@@ -363,6 +366,28 @@ impl<'a> Binder {
...
+                let uri = UriLocation {
+                    protocol: uri.protocol.clone(),
+                    name: uri.name.clone(),
+                    path: uri.path.clone(),
+                    connection: uri.connection.clone(),
+                };
+                parse_uri_location(&uri)?
+            }
+            None => {
+                let sp = self.ctx.get_storage_params();
```
sundy-li, line 380: If it's none, we should store it inside the meta?

What if we already migrate the storage data into another storage data?

We must modify the meta of all the fuse tables.

src/meta/proto-conv/src/table_from_to_protobuf_impl.rs
```diff
@@ -175,6 +177,12 @@ impl FromToProto for mt::TableMeta {
             catalog,
             engine: p.engine,
             engine_options: p.engine_options,
+            storage_params: TableStorageParams::from_pb(p.storage_params.ok_or_else(|| {
```
sundy-li, line 180: It's none for old metas.

src/query/service/tests/it/storages/fuse/pruning.rs
```diff
@@ -78,8 +78,8 @@ async fn test_block_pruner() -> Result<()> {
         table: test_tbl_name.to_string(),
         schema: test_schema.clone(),
         engine: Engine::Fuse,
-        storage_params: ctx.get_storage_params(),
-        path: ctx.get_storage_operator()?.metadata().root().to_string(),
+        storage_params: None,
```
sundy-li, line 81: `storage_params` and `path` should be some or none together. So why not make them inside a new struct ?

ariesdevil: The `path` can be gotten from the operator, so I think we no need to store it.

sundy-li (approved):

src/query/ast/tests/it/parser.rs
```diff
@@ -296,6 +297,10 @@ fn test_statement() {
         r#"SHOW GRANTS OF SHARE t;"#,
     ];
 
+    // let cases = &[
+    //     r#"create table if not exists a.b (a int) 's3://testbucket/admin/data/' connection=(aws_key_id='minioadmin' aws_secret_key='minioadmin' endpoint_url='http://127.0.0.1:9900');"#,
+    // ];
+
```
drmingdrmer, line 303: Can be removed?

drmingdrmer (commented):
Remember to increment the version:
https://github.com/datafuselabs/databend/blob/336d64fbe9ead46398730a7c754bcaf3f1110283/src/meta/proto-conv/src/util.rs#L50

And add test that decodes a message of this version:
https://github.com/datafuselabs/databend/blob/c5c12c35e2f47af361a2f6a618b68728fb2ea4b1/src/meta/proto-conv/tests/it/proto_conv.rs#L538

src/query/ast/tests/it/parser.rs
ariesdevil: Yup

BohuTANG:
Please add documentation:
 Option for `Create a table with external location` cc @soyeric128

PR #7786, opened 2022-09-21
feat(query): migrate comp

TCeason:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

migrate func v1 comp to function v2

Fixes #7636

PR #7781, opened 2022-09-21
feat(function-v2): support constructing array and CAST(... AS VARIANT)

andylokandy:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

- add variadic function `array<T0>(T0, T0, T0, ...) :: Array(T0)`
- support CAST(<Anything> AS VARIANT)

## TODO:

- [x] Build `Variant::Array` from Array(Variant). Awaiting jsonb encoding support from @b41sh
- [x] Build `Varaint::Object` from Tuple.
- [x] Add unit tests for `CAST(... AS VARIANT)`

Blocked by https://github.com/datafuselabs/databend/pull/7802
Tracked in https://github.com/datafuselabs/databend/issues/6547

PR #7802, opened 2022-09-22
feat(query): add jsonb builtin functions build_array and compare

b41sh:

## Summary

add `JSONB` builtin functions `build_array` and `compare`

used by #7781 and #7786


andylokandy:
We may also need a function to convert `impl IntoIterator<Item=(&str, &[u8])>` to bytes of `Value::Object`

src/common/jsonb/src/functions.rs
```diff
@@ -0,0 +1,313 @@
...
+            buf[jentry_index + i] = *b;
+        }
+        jentry_index += 4;
+    }
+    Ok(())
+}
+
+// `JSONB` values supports partial decode for comparison,
+// if the values are found to be unequal,
+// the result will be returned immediately
```
TCeason, line 63: Maybe the comment `// Null > Container(Array > Object) > String > Number > Boolean` should on here? Not on `compare_scalar`?

b41sh: those are `JEntry` types, I will add some comments for `fn compare`

andylokandy (approved):

BohuTANG (approved):

Issue #6908, opened 2022-07-30
Feature: databend-meta: add backoff for send_append_entries, send_install_snapshot, send_vote RPC if there is error occurs


drmingdrmer:
If a node encounters `StorageError`, the raft node stays in a state where it receives a request but can not handle it. In this case, the leader will keep sending replication data(logs or snapshot), and consumes all bandwidth.

To avoid this, the leader should sleep for a while if a previous RPC fails.
The sleep time should increase if more errors occur. And there should be a max sleep interval.

PR #7809, opened 2022-09-22
feat(meta): back off in meta's rpc

### LINKED ISSUE #7713: retry backoff for meta replication
**Summary**

When some node in cluster restart or in wrong state, log file would be filled with same high-frequency error messages.

ClSlaid:

## Summary

This PR enables exponential backoff for meta's RPCs.

- Fix: #7713 
- Fix: #6908


src/meta/service/src/network.rs
```diff
@@ -156,51 +174,85 @@ impl RaftNetwork<LogEntry> for Network {
...
         info!("send_vote: target: {} rpc: {}", target, rpc.summary());
 
         let mut client = self.make_client(&target).await?;
-        let req = common_tracing::inject_span_to_tonic_request(rpc);
 
-        self.incr_meta_metrics_sent_bytes_to_peer(&target, req.get_ref());
+        let mut mes = Default::default();
+        // all back-offed requests are failed
+        let mut last_err = None;
+        for backoff in ExponentialBackoff::default() {
```
drmingdrmer, line 228: It'd better make a method building a backoff instance: explicitly specify the ratio and the max sleep time.

src/meta/service/src/network.rs
```diff
@@ -125,21 +127,37 @@ impl RaftNetwork<LogEntry> for Network {
...
+                last_err = Some(e);
+
+                // backoff and retry sending
+                sleep(backoff).await;
+                continue;
+            }
+            let resp = resp.unwrap();
+            mes = resp.into_inner();
+            // clean up
+            last_err = None;
```
drmingdrmer, line 152: If you were gonna use both the Err and Ok, use a `match` would be better than a `unwrap()`

src/meta/service/src/network.rs
```diff
@@ -81,9 +129,24 @@ impl Network {
...
     }
 
+    pub fn with_backoff_policy(mut self, policy: BackOffPolicy) -> Self {
+        self.back_off_policy = policy;
+        self
+    }
+
+    pub(crate) fn backoff(
+        &self,
+    ) -> std::iter::Chain<ExponentialBackoff, std::vec::IntoIter<Duration>> {
```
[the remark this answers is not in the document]
ClSlaid: Dynamic dispatch may be needed.
```suggestion
    ) -> Box<dyn Iterator<Item = Duration>> {
```

src/meta/service/src/network.rs
```diff
@@ -133,18 +87,42 @@ impl Network {
...
-    pub(crate) fn backoff(
-        &self,
-    ) -> std::iter::Chain<ExponentialBackoff, std::vec::IntoIter<Duration>> {
+    pub(crate) fn back_off(&self) -> impl Iterator<Item = Duration> {
         // the last period of back off should be zero
-        // so the longest backoff will not be wasted
+        // so the longest back off will not be wasted
         let zero = vec![Duration::default()].into_iter();
-        self.back_off_policy.backoff().chain(zero)
+        self.back_off_policy.clone().chain(zero)
```
Xuanwo, line 125: I didn't get this. Why?

This function changes the semantic meaning of `backon::Backoff`:

https://docs.rs/backon/0.2.0/backon/trait.Backoff.html

> - `Some(Duration)` means caller need to sleep(Duration) and retry the same request
> - `None` means we have reaching the limits, caller needs to return current error instead.

We will have a `sleep(0)` for every backoff loop.

src/meta/service/src/network.rs
```diff
@@ -133,18 +87,42 @@ impl Network {
...
+    ///   should be not smaller than 1.0
+    ///
+    /// - `min_delay`: minimum back off duration, where the backoff duration vary starts from
+    ///
+    ///   random jitter between [0, min_back) will be set in each backoff, to minimize conflicts.
+    ///
+    /// - `max_delay`: maximum back off duration, if the backoff duration is larger than this, no backoff will be raised
+    ///
+    /// - `chances`: maximum back off times, chances off backoff
+    pub fn with_back_off_policy(
```
Xuanwo, line 103: Why do we need to store `ExponentialBackoff` in the `Network` struct? It's not a reuseable struct.

src/meta/service/src/network.rs
```diff
@@ -184,6 +184,7 @@ impl RaftNetwork<LogEntry> for Network {
                     mes = resp.into_inner();
                     // clean up
                     last_err = None;
+                    break;
```
Xuanwo, line 187: How about `return xxxx` directly?

src/meta/service/src/network.rs
ClSlaid: Sorry for forgetting to `break` after a success request.

We are using ExponentialBackOff directly so things get different with just using a `backon::retry()` method or `backon::Retry` structure.

When the iteration just restarts from the longest back off, the next item in iterator will be `None`, then the `for` loop will be quited and the longest back off will be wasted.

To avoid this, to retry the same request after the longest back off, a zero duration sleep will be added. If the very last request failed, `sleep(0).await` cost little, the loop will reach the end and the error will be returned in time.

src/meta/service/src/network.rs
ClSlaid: Good, we don't need to care about `last_err` then.

src/meta/service/src/network.rs
Xuanwo: > To avoid this, to retry the same request after the longest back off, a zero duration sleep will be added. If the very last request failed, `sleep(0).await` cost little, the loop will reach the end and the error will be returned in time.

I got it. Maybe we need to do something in `backon` to make using `Backoff` easier.

Xuanwo: Tracked at https://github.com/Xuanwo/backon/issues/23

drmingdrmer: > Sorry for forgetting to `break` after a success request.

Oh, thank god that every consensus protocol tolerates duplicated messages. :D

zhang2014:
@mergify update

Xuanwo (approved):

src/meta/service/src/network.rs
```diff
@@ -186,21 +276,38 @@ impl RaftNetwork<LogEntry> for Network {
...
+        if let Some(e) = last_err {
+            // all back-offed requests failed
+            return Err(anyhow::Error::from(e));
         }
 
-        let resp = resp?;
-        let mes = resp.into_inner();
         let resp = serde_json::from_str(&mes.data)?;
-
         Ok(resp)
```
drmingdrmer, line 311: Why is there a chance the loop of backoff finished, but there isn't an error?

ClSlaid: If you're asking why there is always at least one chance being consumed, that's because we are using the `ExponentialBackoff` structure directly and for each iteration, a chance will be consumed. So this is unavoidable since at least a loop should run to get a response.

If you are suggesting that the parsing error should also trigger a backoff, the newly pushed commits should fill your satisfaction.

drmingdrmer (approved):

Xuanwo:
@ClSlaid please the conflict.

drmingdrmer (approved):

Issue #7816, opened 2022-09-23
Tracing issues of Catalog Refactor


Xuanwo:
## Summary

- [x] https://github.com/datafuselabs/databend/pull/7855
- [x] https://github.com/datafuselabs/databend/pull/7827
- [x] https://github.com/datafuselabs/databend/pull/7835
- [ ] https://github.com/datafuselabs/databend/pull/7869
- [ ] Rename `read2` to `read`
- [ ] Rename `append2` to `append`
- [ ] Catalog / Table should don't depend on any `Plan`
  - Migrate `Extras` in `read_partitions`
  - Migrate `ReadDataSourcePlan` in `read2`
  - Migrate `DeletePlan` in `delete`

### For Migrate `ReadDataSourcePlan` in `read2`

We will call `table.read2()` during `fn build_table_scan(&mut self, scan: &TableScan) -> Result<()>`:

```rust
fn build_table_scan(&mut self, scan: &TableScan) -> Result<()> {
    let table = self.ctx.build_table_from_source_plan(&scan.source)?;
    self.ctx.try_set_partitions(scan.source.parts.clone())?;
    table.read2(self.ctx.clone(), &scan.source, &mut self.main_pipeline)?;
    let schema = scan.source.schema();
    let projections = scan
        .name_mapping
        .iter()
        .map(|(name, _)| schema.index_of(name.as_str()))
        .collect::<Result<Vec<usize>>>()?;

    self.main_pipeline.add_transform(|input, output| {
        Ok(TransformProject::create(input, output, projections.clone()))
    })?;

    self.main_pipeline.add_transform(|input, output| {
        Ok(TransformRename::create(
            input,
            output,
            scan.output_schema()?,
        ))
    })
}
```

We should build a new args for `table.read()` so that we don't need to use logical plan `ReadDataSourcePlan`.

PR #7835, opened 2022-09-23
feat(meta): Add catalog in TableMeta

Xuanwo:

## Summary

feat(meta): Add catalog in TableMeta

Part of https://github.com/datafuselabs/databend/issues/7816


zhang2014:
@mergify update

src/meta/proto-conv/tests/it/proto_conv.rs
```diff
@@ -412,57 +413,118 @@ fn test_load_old() -> anyhow::Result<()> {
...
+                202, 1, 1, 99, 202, 1, 1, 99, 202, 1, 1, 99, 202, 1, 1, 99, 202, 1, 1, 99, 202, 1,
+                1, 99, 202, 1, 1, 99, 202, 1, 1, 99, 202, 1, 1, 99, 202, 1, 1, 99, 202, 1, 1, 99,
+                160, 6, 10, 168, 6, 1,
+            ];
+            let p: pb::TableMeta = common_protos::prost::Message::decode(tbl_meta_v10.as_slice())
+                .map_err(print_err)?;
+
+            let got = mt::TableMeta::from_pb(p).map_err(print_err)?;
+
+            let want = new_table_meta();
```
drmingdrmer, line 525: There should be a dedicated instance, but not the latest version of table-meta,  for testing this `v10` table-meta.

And using some value other than the default value `"default"` for the compatibility test would be better, to assert that the field `catalog` is loaded from the message but not just filled with a default value.

Xuanwo: Addressed in https://github.com/datafuselabs/databend/pull/7835/commits/c5c12c35e2f47af361a2f6a618b68728fb2ea4b1

Xuanwo:
`tbl_meta_v1` is changed by `rustfmt`, no real content change.

drmingdrmer (approved):
nice...

lichuang (approved):

Issue #7740, opened 2022-09-20
Migrate aggregation functions to function-v2


RinChanNOWWW:
**Summary**

- [x] combinator distinct. #7615
- [x] combinator if. #7615
- [x] count. #7615
- [x] sum. #7615
- [x] avg. #7615
- [x] uniq. #7615
- [x] min, max, any. #7787
- [x] arg_min, arg_max. #7901
- [x] std, stddev, stddev_pop #8034
- [x] covar_pop, covar_samp. #7926
- [x] window_funnel #8034
- [x] retention. #7952

**something may be helpful**

- [x] add `try_downcast_builder` for `ValueType`. #7838
- [x] unit tests for group by. #7845
- [x] ~add Default for ValueType::Scalar or add default_scalar for ValueType.~

RinChanNOWWW:
> - add Default for ValueType::Scalar or add default_scalar for ValueType.
> - add try_downcast_builder for ValueType.

Are these necessary? cc @sundy-li

RinChanNOWWW:
The `Default` bound seems not necessary, and not suitable for `AnyType`. Let's not consider this.

TCeason:
Window funnel assign to @TCeason

PR #7838, opened 2022-09-23
feat(query): add `try_downcast_builder` for `ValueType`.

RinChanNOWWW:

## Summary

A fast path for simple types to get their specific builders. 

`try_downcast_builder` for complex types like: `ArrayType<T: ValueType>`, `NullableType<T: ValueType>`, and `KvPair<K: ValueType, V: ValueType>`(Map) always return `None`. There isn't a simple way to downcast their inner builders.

Related with #7740


src/query/expression/src/types/nullable.rs
```diff
@@ -80,6 +81,12 @@ impl<T: ValueType> ValueType for NullableType<T> {
         }
     }
 
+    fn try_downcast_builder<'a>(
```
sundy-li, line 84: I used to add this function, but nullable will be none.

sundy-li (approved):

BohuTANG (approved):

Issue #7819, opened 2022-09-23
`Database` trait add table API.


lichuang:

BohuTANG:
@lichuang Could you give more background about this issue?

lichuang:
> @lichuang Could you give more background about this issue?

In order to support `share database` query tables.

In API such as `catalog.get_table(tenant, db_name, table_name)`, catalog just pass `(tenant,db_name,table_name)` to meta, if it is a normal table, it will works fine.

But if `(tenant,db_name,table_name)` is a shared table, it may happen that this meta service has no share table meta.

So we need to add table API into Database trait, move all Table API in catalog route to Database trait. 

Take the `catalog.get_table(tenant, db_name, table_name)` API for example:

1. first get Database trait impl(now has only default database impl, but there will be share database impl for Database trait).
2. And if found Database is a share database, the share database impl for Database trait will route the table request to the meta which the provider located.

PR #7841, opened 2022-09-23
feat: add into database trait Table API

lichuang:

## Summary

feat: add into database trait Table API, each table API of catalog `MUST` first get a `Database` trait.

Fixes https://github.com/datafuselabs/databend/issues/7819


src/query/catalog/src/database.rs
```diff
@@ -41,4 +63,115 @@ pub trait Database: DynClone + Sync + Send {
     async fn init_database(&self, _tenant: &str) -> Result<()> {
         Ok(())
     }
+
+    // Build a `Arc<dyn Table>` from `TableInfo`.
+    fn get_table_by_info(&self, _table_info: &TableInfo) -> Result<Arc<dyn Table>> {
```
Xuanwo, line 68:
```suggestion
    fn get_table_by_info(&self, table_info: &TableInfo) -> Result<Arc<dyn Table>> {
    let _ = table_info;
```

For better reading in IDE.

src/query/catalog/src/database.rs
```diff
@@ -41,4 +63,115 @@ pub trait Database: DynClone + Sync + Send {
...
+        Err(ErrorCode::UnImplement(format!(
+            "UnImplement get_table_by_info in {} Database",
+            self.name()
+        )))
+    }
+
+    // Get one table by db and table name.
+    async fn get_table(
```
Xuanwo, line 76: ditto

src/query/catalog/src/database.rs
```diff
@@ -41,4 +63,115 @@ pub trait Database: DynClone + Sync + Send {
...
+        &self,
+        _tenant: &str,
```
Xuanwo, line 78: Can we emit the `tenant` and `db_name` args? The current API indicate that we can get table from other database.

src/query/catalog/src/database.rs
```diff
@@ -41,4 +63,115 @@ pub trait Database: DynClone + Sync + Send {
...
+        _db_name: &str,
+        _table_name: &str,
+    ) -> Result<Arc<dyn Table>> {
+        Err(ErrorCode::UnImplement(format!(
+            "UnImplement get_table in {} Database",
+            self.name()
+        )))
+    }
+
+    async fn list_tables(&self, _tenant: &str, _db_name: &str) -> Result<Vec<Arc<dyn Table>>> {
```
Xuanwo, line 88: The same question.

drmingdrmer (commented):
Can you give some explanation about why getting a `Table` has to be done via a `Database`?

src/query/catalog/src/database.rs
```diff
@@ -41,4 +63,115 @@ pub trait Database: DynClone + Sync + Send {
...
+        )))
+    }
+
+    // Get one table by db and table name.
+    async fn get_table(
+        &self,
+        _tenant: &str,
+        _db_name: &str,
+        _table_name: &str,
+    ) -> Result<Arc<dyn Table>> {
```
drmingdrmer, line 81: Such an API looks strange: `self` is already a `Database`, but getting a `Table` requires another `db_name: str`.

This API suggests a `Database` implementation does not know what itself is. And for developers who implement a `Database`, it is also confusing: what should `get_table()` do if the arg `db_name` is different from the name of `self`?

lichuang:
> Can you give some explanation about why getting a `Table` has to be done via a `Database`?

> @lichuang Could you give more background about this issue?

In order to support `share database` query tables.

In API such as `catalog.get_table(tenant, db_name, table_name)`, catalog just pass `(tenant,db_name,table_name)` to meta, if it is a normal table, it will works fine.

But if `(tenant,db_name,table_name)` is a shared table, it may happen that this meta service has no share table meta.

So we need to add table API into Database trait, move all Table API in catalog route to Database trait. 

Take the `catalog.get_table(tenant, db_name, table_name)` API for example:

1. first get Database trait impl(now has only default database impl, but there will be share database impl for Database trait).
2. And if found Database is a share database, the share database impl for Database trait will route the table request to the meta which the provider located.

src/query/catalog/src/database.rs
BohuTANG: Agree.

drmingdrmer:
> Take the `catalog.get_table(tenant, db_name, table_name)` API for example:
> 
> 1. first get Database trait impl(now has only default database impl, but there will be share database impl for Database trait).
> 2. And if found Database is a share database, the share database impl for Database trait will route the table request to the meta which the provider located.

There are two types related to a database: one is about the data a DB has, i.e., the `DatabaseMeta`, and the other is about the behavior a DB provides, i.e., `dyn Database`.

To get a table, the `catalog` only needs the data part of a DB, not the behavior part.

Binding get-table functions to `dyn Database` introduces unnecessary coupling. I am not quite sure it will work.

src/query/catalog/src/catalog.rs
```diff
@@ -151,24 +151,40 @@ pub trait Catalog: DynClone + Send + Sync {
...
+        db_name: &str,
         req: UpsertTableOptionReq,
     ) -> Result<UpsertTableOptionReply>;
 
-    async fn update_table_meta(&self, req: UpdateTableMetaReq) -> Result<UpdateTableMetaReply>;
+    async fn update_table_meta(
+        &self,
+        tenant: &str,
+        db_name: &str,
+        req: UpdateTableMetaReq,
```
Xuanwo, line 163: If we already passed `tenant` and `db` args, how about passing `TableId` directly? So that FUSE engine can build `UpdateTableMetaReq` byself.

And catalog like `dive` doesn't need to know about `UpdateTableMetaReq` anymore.

Xuanwo (approved):
Mostly LGTM.

We can improve the APIs in the future. 

Some quick ideas: Catalog/Database/Table trait should provide their our input args instead of using meta's types directly like `UpsertTableCopiedFileReq`. If underlying catalog/database/table's metadata relays on `metasrv`, they can build the request themselves.

This change will make space for integrating with `hive` and [delta sharing](https://github.com/datafuselabs/databend/issues/7830)

PR #7844, opened 2022-09-23
feat: copy from FTP and secured FTP source

### LINKED ISSUE #7660: Feature: Allow COPY FROM FTP
**Summary**

FTP is an ancient protocol, but still been used everywhere. Allow copying FTP makes databend a more open data platform.

As OpenDAL supported FTP already, this issue should be simple.

ClSlaid:

## Summary

Summary about this PR

Fixes #7660

### Guide

- Build databend with `storage-ftp` feature enabled:

```bash
cargo build --bin databend-query --features=storage-ftp
```

- connect to query with your favorite SQL client

```bash
mycli -p3307 -uroot
```

- copy from FTP, need to be secure

```mysql
COPY INTO books FROM 'ftp://my_ftp.example/books.csv' CONNECTION=(SECURE='true' USERNAME='me' PASSWORD='pwd example') FILE_FORMAT=(TYPE='CSV');
```

in the future, copy with `FTPS` scheme should be supported. waiting for a new version of [opendal](https://docs.rs/opendal).

```mysql
COPY INTO books FROM 'ftps://127.0.0.1/books.csv' CONNECTION=(USERNAME='me' PASSWORD='pwd example') FILE_FORMAT=(TYPE='CSV');
```

### DIFF
--- a/src/common/storage/Cargo.toml
+++ b/src/common/storage/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2021"
 
 [features]
 storage-hdfs = ["opendal/services-hdfs"]
+storage-ftp = ["opendal/services-ftp"]
 
 [dependencies]
 common-base = { path = "../base" }
--- a/src/common/storage/src/config.rs
+++ b/src/common/storage/src/config.rs
@@ -35,6 +35,8 @@ pub struct StorageConfig {
 pub enum StorageParams {
     Azblob(StorageAzblobConfig),
     Fs(StorageFsConfig),
+    #[cfg(feature = "storage-ftp")]
+    Ftp(StorageFtpConfig),
     Gcs(StorageGcsConfig),
     #[cfg(feature = "storage-hdfs")]
     Hdfs(StorageHdfsConfig),
@@ -61,6 +63,10 @@ impl Display for StorageParams {
                 v.container, v.root, v.endpoint_url
             ),
             StorageParams::Fs(v) => write!(f, "fs://root={}", v.root),
+            #[cfg(feature = "storage-ftp")]
+            StorageParams::Ftp(v) => {
+                write!(f, "ftp://root={},endpoint={}", v.root, v.endpoint)
+            }
             StorageParams::Gcs(v) => write!(
                 f,
                 "gcs://bucket={},root={},endpoint={}",
@@ -101,6 +107,8 @@ impl StorageParams {
         match self {
             StorageParams::Azblob(v) => v.endpoint_url.starts_with("https://"),
             StorageParams::Fs(_) => false,
+            #[cfg(feature = "storage-ftp")]
+            StorageParams::Ftp(v) => v.secure,
             #[cfg(feature = "storage-hdfs")]
             StorageParams::Hdfs(_) => false,
             StorageParams::Http(v) => v.endpoint_url.starts_with("https://"),
@@ -146,6 +154,41 @@ impl Default for StorageFsConfig {
         Self {
             root: "_data".to_string(),
         }
+    }
+}
+
+pub const STORAGE_FTP_DEFAULT_ENDPOINT: &str = "ftps://127.0.0.1";
+/// Config for FTP and FTPS data source
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StorageFtpConfig {
+    pub endpoint: String,
+    pub root: String,
+    pub username: String,
+    pub password: String,
+    pub secure: bool,
+}
+
+impl Default for StorageFtpConfig {
+    fn default() -> Self {
+        Self {
+            endpoint: STORAGE_FTP_DEFAULT_ENDPOINT.to_string(),
+            username: "".to_string(),
+            password: "".to_string(),
+            root: "/".to_string(),
+            secure: false,
+        }
+    }
+}
+
+impl Debug for StorageFtpConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("StorageFtpConfig")
+            .field("endpoint", &self.endpoint)
+            .field("root", &self.root)
+            .field("username", &self.username)
+            .field("password", &mask_string(self.password.as_str(), 3))
+            .field("secure", &self.secure.to_string())
+            .finish()
     }
 }
 
--- a/src/common/storage/src/location.rs
+++ b/src/common/storage/src/location.rs
@@ -73,6 +73,15 @@ pub fn parse_uri_location(l: &UriLocation) -> Result<(StorageParams, String)> {
             account_key: l.connection.get("account_key").cloned().unwrap_or_default(),
             root: root.to_string(),
         }),
+        #[cfg(feature = "storage-ftp")]
+        Scheme::Ftp => StorageParams::Ftp(crate::StorageFtpConfig {
+            endpoint: l.name.clone(),
+            root: root.to_string(),
+            username: l.connection.get("username").cloned().unwrap_or_default(),
+            password: l.connection.get("password").cloned().unwrap_or_default(),
+            secure: l.protocol.starts_with("ftps")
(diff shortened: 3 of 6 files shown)


src/common/storage/src/operator.rs
```diff
@@ -93,6 +97,23 @@ pub fn init_fs_operator(cfg: &StorageFsConfig) -> Result<Operator> {
...
+pub fn init_ftp_operator(cfg: &super::StorageFtpConfig) -> Result<Operator> {
+    let mut builder = ftp::Builder::default();
+
+    let bd = builder
+        .endpoint(cfg.endpoint.as_str())
+        .user(cfg.username.as_str())
+        .password(cfg.password.as_str())
+        .root(cfg.root.as_str());
+    if cfg.secure {
+        bd.enable_secure();
```
Xuanwo, line 111: Service FTP's API is weird. Either we provide `ftps://xxxx` or `enable_secure` instead of both of them. What happens if user inputs a `ftps://127.0.0.1` without `enable_secure`? Or a `ftp://127.0.0.1` with `enable_secure`?

We need to polish it here.

Xuanwo: Well, I got it. There is no `ftps` protocol provided by `service-ftp`, only `enable_secure`. I think we should:

- Remove `ftps` in opendal and databend
- Expose `enable_secure` to users directly

ClSlaid: I think we should let users use FTP like HTTP. Since the relationship between FTP and FTPS is really like that between HTTP and HTTPS.

Xuanwo: They are different.

`service-http` understand `http` and `https`, we pass them in the endpoint to HTTP client directly. But `service-ftp` doesn't. The FTP client accepts `192.168.1.1` and requires the user to set `enable_secure`.

Thus, we can't use `ftp` or `ftps` here. Instead, we use `ftp://exmaple.com` and always require the user to set `enable_secure`.

Xuanwo: Please revert PR https://github.com/datafuselabs/opendal/pull/704 and only use `ftp://`.

ClSlaid: > Well, I got it. There is no `ftps` protocol provided by `service-ftp`, only `enable_secure`.

The `enable_secure` means using `FTPS` actually. `FTPS` is TLS-enhanced `FTP`.

Xuanwo: > The `enable_secure` means using `FTPS` actually. `FTPS` is TLS-enhanced `FTP`.

Hey, we are confusing two layers of APIs.

First of all, opendal.

OpenDAL Service FTP provide only one API:

- `endpoint` like `127.0.0.1:9000` or `ftp.example.com`
- `enable_secure` to use tls or not.

Then, databend.

That's OK if you prefer to use `ftp` & `ftps`. In this hey, you must hide `enable_secure` from users.

- If user inputs `ftp://`, disable secure.
- If user inputs `ftps://`, enable secure.

Don't provide the same API across different layers.

ClSlaid: > Service FTP's API is weird. Either we provide ftps://xxxx or enable_secure instead of both of them. 

Got it, I'd rather remove the `enable_secure` option.

src/common/storage/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 
 [features]
 storage-hdfs = ["opendal/services-hdfs"]
+storage-ftp = ["opendal/services-ftp"]
```
Xuanwo, line 11: Let's enable this by default. It's not very big and doesn't require external setup like `JAVA_RUNTIME`..

src/common/storage/src/operator.rs
Xuanwo: > Got it, but I'd rather remove the `SECURE` option in `CONNECTION`.

OK, you are the boss of this feature. Go ahead!

ClSlaid:
wait for #7897 to merge first.

Xuanwo:
> wait for #7897 to merge first.

Done, go ahead!

src/common/storage/src/config.rs
```diff
@@ -149,6 +154,41 @@ impl Default for StorageFsConfig {
...
 
+pub const STORAGE_FTP_DEFAULT_ENDPOINT: &str = "ftps://127.0.0.1";
+/// Config for FTP and FTPS data source
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StorageFtpConfig {
+    pub endpoint: String,
+    pub root: String,
+    pub username: String,
+    pub password: String,
+    pub secure: bool,
```
Xuanwo, line 165: Do we need to remove this?

src/common/storage/src/location.rs
```diff
@@ -73,6 +73,17 @@ pub fn parse_uri_location(l: &UriLocation) -> Result<(StorageParams, String)> {
...
+        Scheme::Ftp => StorageParams::Ftp(crate::StorageFtpConfig {
+            endpoint: if !l.protocol.is_empty() {
+                format!("{}://{}", l.protocol, l.name)
+            } else {
+                l.name.to_string()
+            },
+            root: root.to_string(),
+            username: l.connection.get("username").cloned().unwrap_or_default(),
+            password: l.connection.get("password").cloned().unwrap_or_default(),
+            secure: l.protocol.starts_with("ftps://"),
```
Xuanwo, line 85: How could protocol starts with `ftps://`?

src/common/storage/src/config.rs
ClSlaid: I prefer keeping it so that no extra calculation is needed in `StorageParams::is_secure`.

Xuanwo: > I prefer keeping it so that no extra calculation is needed in `StorageParams::is_secure`.

But this will leak `secure` config to users.

ClSlaid: How about making it private and add an `fn is_secure(&self) -> bool` to StorageFtpConfig?

ClSlaid: > How about making it private and add an `fn is_secure(&self) -> bool` method to StorageFtpConfig?

Ok this is a bad idea, I'll just remove it.

src/common/storage/src/location.rs
```diff
@@ -73,6 +73,17 @@ pub fn parse_uri_location(l: &UriLocation) -> Result<(StorageParams, String)> {
             account_key: l.connection.get("account_key").cloned().unwrap_or_default(),
             root: root.to_string(),
         }),
+        Scheme::Ftp => StorageParams::Ftp(crate::StorageFtpConfig {
+            endpoint: if !l.protocol.is_empty() {
+                format!("{}://{}", l.protocol, l.name)
```
Xuanwo, line 78: We are missing port.

Xuanwo: Please add a test case for this.

src/common/storage/src/operator.rs
```diff
@@ -93,6 +95,19 @@ pub fn init_fs_operator(cfg: &StorageFsConfig) -> Result<Operator> {
     Ok(Operator::new(builder.build()?).layer(LoggingLayer))
 }
 
+/// init_ftp_operator will init a opendal ftp operator.
+pub fn init_ftp_operator(cfg: &super::StorageFtpConfig) -> Result<Operator> {
+    let mut builder = ftp::Builder::default();
+
+    let bd = builder
+        .endpoint(cfg.endpoint.as_str())
```
Xuanwo, line 103: Can we use `&cfg.endpoint`?

src/common/storage/src/location.rs
ClSlaid: ok

Xuanwo (approved):

Xuanwo:
Nice PR, let's merge!

PR #7855, opened 2022-09-24
fix: Plan should never implement Serialize and Deserialize

Xuanwo:

## Summary

fix: Plan should never implement Serialize and Deserialize

---

`ReadDataSourcePlan` is referred to by `TableScan`, a `PhysicalPlan`. We will address this in later PRs.


PsiACE (approved):

BohuTANG:
Can you give more on we remove the trait`Serialize` and `Deserialize` of the plan?
If we want to add a new RPC: execute(plan:Plan) to make databend-query as physical executor, it need the trait.

Xuanwo:
> If we want to add a new RPC: execute(plan:Plan) to make databend-query as physical executor, it needs the trait.

We should send `PhysicalPlan` between nodes.

```rust
#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
pub enum PhysicalPlan { .. }
```

BohuTANG:
> We should send `PhysicalPlan` between nodes.
> 
> ```rust
> #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
> pub enum PhysicalPlan { .. }
> ```

That's great!

BohuTANG (approved):

PR #7858, opened 2022-09-24
ci: Add auto suggest change for taplo

Xuanwo:

## Summary

ci: Add auto suggest change for taplo


Xuanwo:
We need to use review API.

Issue #7748, opened 2022-09-20
Tracking issue of bloom filter refactoring


drmingdrmer:
## Problems

- Bloom filter index takes too much storage space, e.g., about 50% of the original data size.
- With a bloom filter index, an exact-match query from 10 billion rows still takes minutes.

The goals are:
- to try to accelerate queries, at least for explicitly simple queries.
- to reduce index size if possible.

Issues:

- #7379
- #7674

Let me know if there are any other issues related bloom filter.:)

## Status

- 2022-09-20: It's still under investigation and I can not tell for now what will be done:(

drmingdrmer:
@BohuTANG @dantengsky @sundy-li @youngsofun

BohuTANG:
@drmingdrmer 

We will first try the XOR filter instead of the old raw bloom filter, then @dantengsky will help me to adapt the raw bloom filter to the new XOR filter.
The new XOR filter is here:
https://github.com/datafuselabs/databend/pull/7860

From my test, it looks great, let's try :)

BohuTANG:
@drmingdrmer 

I have moved the issues related to the index(not only the bloom filter) to https://github.com/datafuselabs/databend/issues/7823. Let's close this :)

PR #7860, opened 2022-09-24
feat: add XOR bloom filter

BohuTANG:

## Summary

* Introduce a new bloom trait
* Use XOR instead of bloom filter

Test result(bitmap encode size / raw data size):
```
u64: 
bitmap enc:1230069, raw:8000000, ratio:0.15375863

bool:
bitmap enc:61, raw:1000000, ratio:0.000061

string:
string enc:123067, raw:3000000, ratio:0.041022334

100000 records of the same key:
string enc:61, raw:3000000, ratio:0.000020333333
```

Part of #7748

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,6 +96,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arbitrary"
+version = "1.1.6"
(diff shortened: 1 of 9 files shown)


dantengsky (approved):
let's roll

dantengsky (approved):

PR #7865, opened 2022-09-24
fix: disable bloom filter index in write path

### LINKED ISSUE #7780: feat: disable bloomfilter index
**Summary**

Some customers told us it was too slow for the point query.

cc @dantengsky @drmingdrmer

dantengsky:

## Summary

hot fix, should have disabled bloom index in write path as well.

Fixes #7780


dantengsky:
closed. replace  with xor bloom filter  instead

PR #7866, opened 2022-09-24
feat: parser and binder for tabular function

### LINKED ISSUE #5731: Tabular SQL UDFs (UDTFs)

GrapeBaBa:

## Summary

feat: parser and binder for tabular function

Fixes #5731


src/query/service/src/sql/planner/plans/create_tabular_function.rs
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+use std::sync::Arc;
+
+use common_datavalues::DataSchema;
+use common_datavalues::DataSchemaRef;
+
+#[derive(Clone, Debug)]
+pub struct CreateTabularFunctionPlan {
```
GrapeBaBa, line 20: @BohuTANG @sundy-li  Need help. I am not sure how to handle this plan in the interpreter. I use `DataSchemaRef` for args and source so that it can't be used in meta crate like stage or udf. This plan seems more like view or table which has different interpreter logic with stage/udf.

BohuTANG: Need a new TabularFunctionInterpreter?

Xuanwo:
Feel free to open it again :heart:

Issue #7808, opened 2022-09-22
cannot pass `make lint` after `cargo update`


ClSlaid:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

[d747a6c](https://github.com/datafuselabs/databend/commit/d747a6c5fd1e359fd0c1a345415c07f23c95db08)

### What's Wrong?

After `cargo update` all dependency cache, `make lint` will fail on checking `common-hive-meta-store`:

```
error[E0432]: unresolved import `thrift::protocol::TSerializable`
  --> /databend/target/debug/build/common-hive-meta-store-a3775ca046a58940/out/hms_patched.rs:16:142
   |
16 | ...ssageType, TInputProtocol, TOutputProtocol, TSerializable, TSetIdentifier, TStructIdentifier, TType};
   |                                                ^^^^^^^^^^^^^ no `TSerializable` in `protocol`
```

### How to Reproduce?

check out the latest main branch like [d747a6c](https://github.com/datafuselabs/databend/commit/d747a6c5fd1e359fd0c1a345415c07f23c95db08), and run `cargo update`, then `make lint`.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Xuanwo:
Thrift released v0.17 serval days ago, but its rust lib is not updated. If you are using Archlinux, please rollback to v0.16 until the upstream fixed this problem.

Xuanwo:
Already fixed.

PR #7868, opened 2022-09-24
fix: Fix build for thrift v0.17

Xuanwo:

## Summary

fix: Fix build for thrift v0.17

Archlinux already upgrades `thrift` package to v0.17, but its corresponding rust lib not been published: https://issues.apache.org/jira/projects/THRIFT/issues/THRIFT-5640?filter=allopenissues

```rust
error[E0432]: unresolved import `thrift::protocol::TSerializable`
  --> /databend/target/debug/build/common-hive-meta-store-a3775ca046a58940/out/hms_patched.rs:16:142
   |
16 | ...ssageType, TInputProtocol, TOutputProtocol, TSerializable, TSetIdentifier, TStructIdentifier, TType};
   |                                                ^^^^^^^^^^^^^ no `TSerializable` in `protocol`
```

This PR uses the git version of thrift to address that.

Fix https://github.com/datafuselabs/databend/issues/7808


Cargo.lock
```diff
@@ -1088,7 +1088,7 @@ dependencies = [
  "semver",
  "serde",
  "serfig",
- "thrift",
+ "thrift 0.18.0",
```
PsiACE, line 1091: It looks like this means that there are two versions of thrift at the same time, and I think that thrift 0.18 works, but what about the other older version?

Xuanwo: thrift specify the version to `v0.18` after releasing `v0.17`.

Xuanwo: > but what about the other older version?

`opentelemetry-jaeger` depends on `v0.15`, we don't know if we can patch this too.

Xuanwo:
Cc @dantengsky @sandflee Can you give this PR a test?

PR #7867, opened 2022-09-24
feat(cluster): experimental distributed eval index

zhang2014:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

feat(cluster): experimental distributed eval index

Fixes #issue

PR #7857, opened 2022-09-24
chore(cache): use parking lock to replace tokio async lock

zhang2014:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

chore(cache): use parking lock to replace tokio async lock

Fixes #issue

PR #7870, opened 2022-09-24
fix: re-enable (xor) bloom filter index

dantengsky:

## Summary

as xor bloom has been merged in PR https://github.com/datafuselabs/databend/pull/7860, replace the vanilla bloom filter with it.

- for legacy index data of v1 bloom filter, the index will be ignored
  but will be removed if table data is dropped
- xor index is created under path "_i_b_v2"
  so that, other types of indexes could be put in separated "dir"s
- bloom filter index is enabled for string type column only

Fixes #issue


BohuTANG:
Cool.
The bloom filter should have a type so that we can try fuse binary bloom in the future:
[Binary Fuse Filters: Fast and Smaller Than Xor Filters](https://arxiv.org/pdf/2201.01174.pdf)
image

BohuTANG:
From the test, `XOR8` is better than `FUSE8` for the applications, `FUSE8` should check the duplicate keys(like hyperloglog)
Test Result:
```
fuse8: u64 bitmap encoding:1130544 bytes, raw:8000000 bytes, ratio:0.141318
fuse8: bool bitmap encoding:1130544 bytes, raw:1000000 bytes, ratio:1.130544
fuse8: string encoding:118832 bytes, raw:3000000 bytes, ratio:0.039610665
fuse8: same string encoding:118832 bytes, raw:3000000 bytes, ratio:0.039610665
xor8: u64 bitmap encoding:1230069 bytes, raw:8000000 bytes, ratio:0.15375863
xor8: bool bitmap encoding:61 bytes, raw:1000000 bytes, ratio:0.000061
xor8: string encoding:123067 bytes, raw:3000000 bytes, ratio:0.041022334
xor8: same string encoding:61 bytes, raw:3000000 bytes, ratio:0.000020333333
```

https://github.com/BohuTANG/databend/commit/b45793c044c17cc6f8706c42bbe21201590f359f

src/query/storages/index/tests/it/bloom/bloom_filter.rs
```diff
@@ -0,0 +1,92 @@
...
+    let block = DataBlock::create(schema.clone(), vec![
+        Series::from_data(numbers),
+        Series::from_data(bools),
+        Series::from_data(strs),
+    ]);
+
+    let index = BloomFilterIndexer::try_create(&[&block])?;
+
+    // only one index column should be generated
+    assert_eq!(2, index.bloom_block.columns().len());
```
dantengsky, line 53:
```suggestion
    assert_eq!(1, index.bloom_block.columns().len());
```

BohuTANG:
I will bench the xor bloom index loading after https://github.com/datafuselabs/databend/pull/7867 and https://github.com/datafuselabs/databend/pull/7857

BohuTANG:
Test result:

Gen data(unique):
```
create table t10b as select number as c1, cast(rand() as string) as c2 from numbers(1000000000);
```

main branch:
```
mysql> select * from t10b where c2='0.4091273956277217';
Connection id:    9
Current database: bloom

+-----------+--------------------+
| c1        | c2                 |
+-----------+--------------------+
| 125000007 | 0.4091273956277217 |
+-----------+--------------------+
1 row in set (1 min 13.72 sec)
Read 1000000000 rows, 24.47 GiB in 73.697 sec., 13.57 million rows/sec., 340.05 MiB/sec.
```

This branch:
```
mysql> select * from t10b where c2='0.4091273956277217';
Connection id:    9
Current database: bloom

+-----------+--------------------+
| c1        | c2                 |
+-----------+--------------------+
| 125000007 | 0.4091273956277217 |
+-----------+--------------------+
1 row in set (41.11 sec)
Read 6000000 rows, 157.95 MiB in 41.090 sec., 146.02 thousand rows/sec., 3.84 MiB/sec.
```

XOR filter is ~2X faster and scans less data(only need to scan xor filter index):
```
mysql> explain select * from t10b where c2='0.4091273956277217';
+-----------------------------------------------------------------------------+
| explain                                                                     |
+-----------------------------------------------------------------------------+
| TableScan                                                                   |
| ├── table: default.bloom.t10b                                               |
| ├── read rows: 6000000                                                      |
| ├── read bytes: 126715443                                                   |
| ├── partitions total: 1000                                                  |
| ├── partitions scanned: 6                                                   |
| └── push downs: [filters: [(c2 = '0.4091273956277217')], limit: NONE]       |
+-----------------------------------------------------------------------------+
```
With https://github.com/datafuselabs/databend/pull/7893 would be great for the point query.

BohuTANG (approved):
👍

Issue #6926, opened 2022-08-01
Tracking issues of `upstream first` violation


Xuanwo:
Databend will adopt `Upstream First` rules across the whole project, which means:

- We will try our best to keep our dependences in the latest stable version.
- We will try our best to report issues we met upstream.
- We will try our best to contribute back our own changes.

In this tracking issue, we will record the temporary forks that we are using and our plan for them:

- `sqlparser = { git = "https://github.com/datafuse-extras/sqlparser-rs", rev = "7f246e3" }`
- `sled = { git = "https://github.com/datafuse-extras/sled", tag = "v0.34.7-datafuse.1", default-features = false }`
- `parquet2 = { version = "0.14.1", optional = true, git = "https://github.com/datafuse-extras/parquet2", rev = "3a468fc3c4" }`
  - [x] https://github.com/datafuselabs/databend/issues/6064

## Already cleaned

- `bincode`
- `chrono`: https://github.com/datafuselabs/databend/issues/6928
- `clickhouse-driver`: https://github.com/datafuselabs/databend/issues/6967
- `cargo-license`: https://github.com/datafuselabs/databend/issues/6969
- `async-trait`: https://github.com/datafuselabs/databend/issues/6968

Xuanwo:
cc @andylokandy @leiysky, maybe we can remove `sqlparser` after the new planner is ready?

Xuanwo:
cc @drmingdrmer @lichuang, will we maintain sled as a new upstream? Or we will use `openkv` to replace `sled` finally?

Xuanwo:
cc @PsiACE, will we need to use a fork of `cargo-license`?

Xuanwo:
cc @dantengsky, can we migrate `parquet2` to upstream instead of our own fork?

drmingdrmer:
> cc @drmingdrmer @lichuang, will we maintain sled as a new upstream? Or we will use `openkv` to replace `sled` finally?

To me. I'm not sure about this. `sled` has not been actively maintained for a while. Its author recently switched to another LSM storage project.
The extra/sled is mainly for easily adding a tag for databend use.

Xuanwo:
> To me. I'm not sure about this. `sled` has not been actively maintained for a while. Its author recently switched to another LSM storage project.

We need a decision here. Possible option are as follows:

- Fork `sled` (like openraft from async-raft) and actor as a new upstream.
- Switch to other action k-v storage projects
[...]

PR #7883, opened 2022-09-26
chore: bump deps

### LINKED ISSUE #7852: Many of databend dependences are out of date
**Summary**

After `cargo upgrade`:

image

Build will fail.

PsiACE:

## Summary

tonic v0.8
prost v0.11
opentelemetry-jaeger v0.17
h3ron v0.15

Fixes #7852

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,4 +115,5 @@ rustc-demangle = { opt-level = 3 }
 [patch.crates-io]
 # If there are dependencies that need patching, they can be listed below.
 # For example:
-# http-types = { git = "https://github.com/Xuanwo/http-types", rev = "106fc6d" }
+# arrow-format = { git = "https://github.com/datafuse-extras/arrow-format", rev = "78dacc1" }
+arrow-format = { git = "https://github.com/datafuse-extras/arrow-format", rev = "78dacc1" }
--- a/src/binaries/Cargo.toml
+++ b/src/binaries/Cargo.toml
@@ -51,10 +51,10 @@ anyhow = "1.0.65"
 clap = { version = "3.2.22", features = ["derive", "env"] }
 openraft = { git = "https://github.com/datafuselabs/openraft", tag = "v0.7.3" }
 sentry = "0.27.0"
-serde = { version = "1.0.144", features = ["derive"] }
+serde = { version = "1.0.145", features = ["derive"] }
 serde_json = "1.0.85"
 tokio-stream = "0.1.10"
-tonic = "0.7.2"
+tonic = "0.8.1"
 tracing = "0.1.36"
 url = "2.3.1"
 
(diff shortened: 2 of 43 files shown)


src/query/functions/src/scalars/geos/geo_to_h3.rs
```diff
@@ -114,7 +114,7 @@ impl Function for GeoToH3Function {
             // x must be Longitude and y must be Latitude
             // `h3ron` will transform `Coordinate{x, y}` to `GeoCoord{lat:y, lon:x}` internally.
             let coord = Coordinate { x: lon, y: lat };
-            let h3_cell = H3Cell::from_coordinate_unchecked(&coord, res);
+            let h3_cell = H3Cell::from_coordinate(coord, res).unwrap();
```
PsiACE, line 117: > Most `*_unchecked` methods are gone now as H3 returns natively error codes.

cc @ariesdevil

ariesdevil (approved):

Cargo.toml
```diff
@@ -115,4 +115,5 @@ rustc-demangle = { opt-level = 3 }
 [patch.crates-io]
 # If there are dependencies that need patching, they can be listed below.
 # For example:
-# http-types = { git = "https://github.com/Xuanwo/http-types", rev = "106fc6d" }
+# arrow-format = { git = "https://github.com/datafuse-extras/arrow-format", rev = "78dacc1" }
+arrow-format = { git = "https://github.com/datafuse-extras/arrow-format", rev = "78dacc1" }
```
Xuanwo, line 119: Please update https://github.com/datafuselabs/databend/issues/6926

src/binaries/Cargo.toml
```diff
@@ -51,10 +51,10 @@ anyhow = "1.0.65"
 clap = { version = "3.2.22", features = ["derive", "env"] }
 openraft = { git = "https://github.com/datafuselabs/openraft", tag = "v0.7.3" }
 sentry = "0.27.0"
-serde = { version = "1.0.144", features = ["derive"] }
+serde = { version = "1.0.145", features = ["derive"] }
```
Xuanwo, line 54: How about always using `1` instead? So that we only need to maintain `Cargo.lock`.

src/binaries/Cargo.toml
```diff
@@ -51,10 +51,10 @@ anyhow = "1.0.65"
...
 serde_json = "1.0.85"
 tokio-stream = "0.1.10"
-tonic = "0.7.2"
+tonic = "0.8.1"
```
Xuanwo, line 57: Use `0.8` instead?

PsiACE: I would like the version of tonic to be deterministic, there may be compilation issues if the settings are too flexible (the same applies to prost).

Xuanwo: `0.8.1` is not deterministic. In fact, `Cargo.toml` can't determine anything. We should maintain `Cargo.lock` always.

src/binaries/Cargo.toml
PsiACE: This will be addressed in another PR, where I will try to introduce dependencies inheritance.

src/binaries/Cargo.toml
PsiACE: OK, will address in another PR.

src/binaries/Cargo.toml
Xuanwo: > where I will try to introduce dependencies inheritance.

Not a good idea to me, as we didn't specify deps at the workspace level.

Xuanwo (approved):

Issue #7231, opened 2022-08-22
bug: copy from parquet file OOM


BohuTANG:
**Summary**

Table:
```
CREATE TABLE transactions (
    hash STRING,
    nonce STRING,
    transaction_index BIGINT UNSIGNED,
    from_address STRING,
    to_address STRING,
    value DOUBLE,
    gas BIGINT UNSIGNED,
    gas_price BIGINT UNSIGNED,
    method_id STRING,
    input STRING,
    max_fee_per_gas BIGINT UNSIGNED,
    max_priority_fee_per_gas BIGINT UNSIGNED,
    transaction_type BIGINT UNSIGNED,
    block_hash STRING,
    block_number BIGINT UNSIGNED,
    block_timestamp BIGINT UNSIGNED
);
```

Copy SQL:
```
COPY INTO transactions FROM 's3://etharchive.org/pub/transactions/' pattern = 'transactions*.*parquet' FILE_FORMAT = (type = 'PARQUET');
```
Each parquet file is about 600MB(raw size will be 1.6G, and if we have ), there are four files.
![image](https://user-images.githubusercontent.com/172204/185868190-fe33a859-f54b-46f5-9b71-46607053bada.png)

This will work:
```
set max_threads=1;
COPY INTO transactions FROM 's3://etharchive.org/pub/transactions/' pattern = 'transactions*.*parquet' FILE_FORMAT = (type = 'PARQUET');
```

Issue #7732, opened 2022-09-19
tracing new InputFormat


youngsofun:
**Summary**

- [ ] new framework  https://github.com/datafuselabs/databend/pull/7613/files
    - [ ] replace copy 
- [ ] replace all
   - [ ] streaming load
   - [ ] clickhouse 

 speed up
-  [ ] use cursor instead of BufferRead for most cases
-  [ ]Parallel read
    -  [ ] TSV/ndjson (read beyond split boundary)
    -  [ ] parquet

features

-  [ ] per file progress/affect
-  [ ] refactor format settings/options

Issue #7889, opened 2022-09-26
Streaming COPY


Xuanwo:
## Background

Databend will list all files before copying data:

https://github.com/datafuselabs/databend/blob/46da962d5637b35555e8f3f49885d1b58211ab41/src/query/service/src/interpreters/interpreter_copy_v2.rs#L399-L460

This will lead to:

- https://github.com/datafuselabs/databend/issues/7862

We need to introduce streaming copy which will copy simultaneously during the listing.

## Problems

- GetTableCopiedFileReply will return all files
- UpsertTableCopiedFileReq requires update all copied files at once

We need to change them to streaming-friendly. For example, only check a batch of files.

## Tasks

- [x] https://github.com/datafuselabs/databend/issues/7892
- [ ] Copy file streamingly

Xuanwo:
Cc @lichuang

BohuTANG:
Before this work, I think we should finished https://github.com/datafuselabs/databend/issues/7755

ClSlaid:
further update on this?

Xuanwo:
Wait for https://github.com/datafuselabs/databend/issues/7892

lichuang:
> Wait for #7892

#7892 will be processed after finishing share table impl

lichuang:
https://github.com/datafuselabs/databend/issues/7892 has been closed.

Xuanwo:
There are other works to finish on copy side. Let's reopen it.

BohuTANG:
I will take this, we need to consider more situations:
If `COPY` will be distributed, streaming reading files in `StageTable:read_partitions` is hard.
For now, it looks better to follow the current mechanism for copying. If a bucket has many files, now *slow*(list files) is by design.

cc @Xuanwo

Xuanwo:
> streaming reading files in `StageTable:read_partitions` is hard.

Partitions scanning and data reading are different kinds of workloads that can be executed at the same time. Adding this feature will also improve the performance of `FuseTable`.

BohuTANG:
New COPY design will be use StageTable as table engine and re-use the planner& pipeline distribution:
https://github.com/datafuselabs/databend/issues/8635

PR #7903, opened 2022-09-26
feat(parquet): read in parallel.

youngsofun:

## Summary

Summary about this PR

part of https://github.com/datafuselabs/databend/issues/7732

Fixes https://github.com/datafuselabs/databend/issues/7231  (big parquet oom)


youngsofun:
@dantengsky  I often see  errors like this when I  test `copy into ontime200 .. ` manually , even on a new table (with table name never used before)

```
 2022-09-26T12:53:10.803063Z ERROR common_meta_api::schema_api_impl: error=TableVersionMismatched: 160 expect `== 163` but `168`  while `update_table_meta
```

src/query/pipeline/sources/src/processors/sources/input_formats/input_format_text.rs
```diff
@@ -135,6 +110,58 @@ impl<T: InputFormatTextBase> InputFormat for InputFormatText<T> {
     fn exec_stream(&self, ctx: Arc<InputContext>, pipeline: &mut Pipeline) -> Result<()> {
         InputFormatTextPipe::<T>::execute_stream(ctx, pipeline)
     }
+
+    async fn get_splits(
+        &self,
+        plan: &CopyIntoPlan,
```
Xuanwo, line 116: It's better not to rely on logical plan directly. How about passing `stage_info` and `files`?

BTW, what will happen if `files` is empty?

youngsofun: this CopyIntoPlan is not logical plan. I may rename it later.

---

> BTW, what will happen if files is empty?

```
2022-09-26T14:04:00.726888Z  INFO databend_query::interpreters::interpreter_copy_v2: matched copy unduplicate files: []
2022-09-26T14:04:00.727168Z  INFO databend_query::interpreters::interpreter_query_log: {"log_type":2,"handler_type":"... }

```

dantengsky:
> @dantengsky I often see errors like this when I test `copy into ontime200 .. ` manually , even on a new table just created (with table name never used before)
> 
> ```
>  2022-09-26T12:53:10.803063Z ERROR common_meta_api::schema_api_impl: error=TableVersionMismatched: 160 expect `== 163` but `168`  while `update_table_meta
> ```
> 
> I deleted local `./databend` and `testbucket/admin/1` on minio

yes ,I'have saw the same errors occasionally, in some unexpected situations(where I think there should be no TableVersionMismatched exception).  unfortunately, I forget the details : (

but in my case, it is not related to the deletion of data from storage.

is there a way to reproduce this error?

src/query/pipeline/sources/src/processors/sources/input_formats/input_format_text.rs
Xuanwo: Ok, I got it. There is a plan to rewrite copy in a streaming way: https://github.com/datafuselabs/databend/issues/7889

Can you retry this function to accept a `impl Stream<Result<StageFile>>` instead of files? (The `files` could contain 200W entries)

It's OK to keep this PR AS-IS and do this change in the future.

youngsofun: got it.

youngsofun:
interpreters::interpreter_table_recluster::test_alter_recluster_interpreter fill again @zhyass

Xuanwo (approved):

dantengsky (approved):

sundy-li (approved):

PR #7908, opened 2022-09-27
chore(format):  clean up some unused code.

youngsofun:

## Summary

Summary about this PR

part of #7732


Xuanwo (approved):

fkuner:
Is the format diagnosis no longer used?

BohuTANG (approved):

youngsofun:
> Is the format diagnosis no longer used?

yes, it was used mainly for csv/tsv

new impls cut file to bytes of rows and fields first,  then interpreter bytes of each field according to datatype，
the error type and err context is already clear enough, and more easier to get the error data (no need checkpoint).

src/query/service/src/pipelines/executor/executor_graph.rs
```diff
@@ -241,8 +241,15 @@ impl ExecutingGraph {
                 if state_guard_cache.is_none() {
                     state_guard_cache = Some(node.state.lock().unwrap());
                 }
-
-                let processor_state = match node.processor.event()? {
+                let event = node.processor.event()?;
+                tracing::debug!(
```
sundy-li, line 245: we should use trace level here.

youngsofun: fixed in https://github.com/datafuselabs/databend/pull/7927

PR #7910, opened 2022-09-27
feat(query): support unload multi files into stage

### LINKED ISSUE #7618: Feature: Copy unload table support multi dest files
**Summary**
https://databend.rs/doc/reference/sql/dml/dml-copy-into-location
Unload to one file has many issues:
1.  Easy OOM
2. Too large to be handled well by downstream

sundy-li:

## Summary

Summary about this PR

Fixes #7618

Changes:
- https://github.com/datafuselabs/databend/issues/7618
- Remove useless tests codes 

cc @soyeric128 about the document changes
Add two copy options:

- SINGLE:  default to false,  unload dest in a single file or not.
- MAX_FILE_SIZE: default to 0,  Number (> 0) that specifies the upper size limit (in bytes) of each file to be generated in parallel per thread. 

More: https://docs.snowflake.com/en/sql-reference/sql/copy-into-location.html


Xuanwo:
`MAX_FILE_SIZE=0` seems to overlap with `SINGLE`.

sundy-li:
> MAX_FILE_SIZE = 0

It'll default to limit 5G if MAX_FILE_SIZE = 0.

Xuanwo:
What will happen if the user set `MAX_FILE_SIZE=100` and `SINGLE=TRUE`?

sundy-li:
> What will happen if the user set MAX_FILE_SIZE=100 and SINGLE=TRUE?

If SINGLE = TRUE,  MAX_FILE_SIZE is useless.

src/query/ast/src/ast/statements/copy.rs
```diff
@@ -74,6 +76,11 @@ impl Display for CopyStmt<'_> {
             write!(f, " SIZE_LIMIT = {}", self.size_limit)?;
         }
 
+        if self.max_file_size != 0 {
+            write!(f, " MAX_FILE_SIZE = {}", self.max_file_size)?;
+        }
+
+        write!(f, " single = {}", self.single)?;
```
Xuanwo, line 83:
```suggestion
        write!(f, " SINGLE = {}", self.single)?;
```

src/query/service/src/storages/stage/stage_table.rs
```diff
@@ -141,75 +140,57 @@ impl Table for StageTable {
...
-                ContextSink::create(input_port, ctx.clone()),
-            );
+        let single = self.table_info.stage_info.copy_options.single;
+        let op = StageTable::get_op(&ctx, &self.table_info.stage_info)?;
+
+        let uuid = uuid::Uuid::new_v4().to_string();
+        let group_id = AtomicUsize::new(0);
+
+        // parallel compact unload, the partial block will flush into next operator
+        if !single && pipeline.output_len() > 1 {
```
BohuTANG, line 150: Where do we do the parallel compact?

sundy-li: if single is not set, it's on parallel.

src/meta/protos/proto/user.proto
```diff
@@ -194,6 +194,8 @@ message UserStageInfo {
     OnErrorMode on_error = 1;
     uint64 size_limit = 2;
     bool purge = 3;
+    bool single = 4;
```
drmingdrmer, line 197: By default `single` is false if it is absent. Does it matter?

drmingdrmer (commented):
There should be a test case that loads this new version message like:
https://github.com/datafuselabs/databend/blob/49d40185e2d06b13d4fc41a38e4fcb7d2ef4e805/src/meta/proto-conv/tests/it/user_stage.rs#L47

src/meta/protos/proto/user.proto
Xuanwo: Yes, single default to false.

src/query/ast/src/ast/statements/copy.rs
Xuanwo: @sundy-li golden file needs a regenerate.

src/query/service/src/storages/stage/stage_table.rs
BohuTANG: The pipeline only adds one transform with:
```
pipeline.add_transform(|input, output| {
```

So, the parallel is in?

src/meta/protos/proto/user.proto
sundy-li: Yes, it's  false by default

src/query/service/src/storages/stage/stage_table.rs
sundy-li: If the pipeline's output size is larger than 1, `pipeline.add_transform` will add a transformer for each output  (like TransformLimit).

The idealized parallel can be achieved after #7939 to avoid extra Exchange in cluster queries.

src/meta/protos/proto/user.proto
drmingdrmer: I mean, does being `false` have any unexpected side effect? E.g. all uploaded will be multiparted, is that OK?

drmingdrmer (approved):
> I mean, does being `false` have any unexpected side effect? E.g. all uploads will be multiparted, is that OK?


Except this(it is marked as resolved, thus I suppose it's all right with it.), everything else is great to me.

Xuanwo (approved):

sundy-li:
> > I mean, does being `false` have any unexpected side effect? E.g. all uploads will be multiparted, is that OK?
> 
> Except this(it is marked as resolved, thus I suppose it's all right with it.), everything else is great to me.

Yes， because the default `MAX_FILE_SIZE` is 5G if not set, there will be no multipart small files by default.

PR #7912, opened 2022-09-27
fix: concurrency control of segment loading

dantengsky:

## Summary

Loading of segment data is too aggressive during pruning. this patch constrains the degree of parallelism according to the setting of `max_threads`.

Fixes #issue


BohuTANG (approved):
I have test this PR and the memory is ok now.

PR #7918, opened 2022-09-27
refactor(meta/metrics): simplify metrics

drmingdrmer:

##### refactor(meta/metrics): simplify metrics

Use crate `metrics` to replace `prometheus`.

Use mod name space to classify metrics.

Add test to assert metrics will be built.

## Related Issues


Xuanwo (approved):
Nice!

src/meta/service/Cargo.toml
```diff
@@ -46,6 +46,7 @@ anyhow = "1.0.65"
 async-trait = "0.1.57"
 clap = { version = "3.2.22", features = ["derive", "env"] }
 futures = "0.3.24"
+metrics = "0.20.1"
```
Xuanwo, line 49: Maybe we need to remove `prometheus`?

Xuanwo:
Check failed for

```rust
unused dependencies:
`databend-meta v0.1.0 (/workspace/src/meta/service)`
└─── dependencies
     └─── "prometheus"
Note: These dependencies might be used by other targets.
      To find dependencies that are not used by any target, enable `--all-targets`.
Note: They might be false-positive.
      For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
      To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
```

PR #7920, opened 2022-09-27
feat: add sync_prune method for BlockPruner

### LINKED ISSUE #7913: Feature:  sync pruning api
**Summary**

https://github.com/datafuselabs/databend/pull/7867 needs a sync style pruning api.

dantengsky:

## Summary

add `sync_prune` method for `BlockPruner`

Fixes #7913

### DIFF
--- a/src/query/storages/fuse/src/pruning/pruning_executor.rs
+++ b/src/query/storages/fuse/src/pruning/pruning_executor.rs
@@ -46,6 +46,20 @@ impl BlockPruner {
         Self { table_snapshot }
     }
 
+    // Sync version of method `prune`
+    //
+    // Please note that it will take a significant period of time to prune a large table, and
+    // thread that calls this method will be blocked.
+    #[tracing::instrument(level = "debug", skip(self, schema, ctx), fields(ctx.id = ctx.get_id().as_str()))]
+    pub fn sync_prune(
+        &self,
+        ctx: &Arc<dyn TableContext>,
+        schema: DataSchemaRef,
+        push_down: &Option<Extras>,
+    ) -> Result<Vec<(usize, BlockMeta)>> {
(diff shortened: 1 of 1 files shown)


src/query/storages/fuse/src/pruning/pruning_executor.rs
```diff
@@ -46,6 +46,20 @@ impl BlockPruner {
...
+    // Please note that it will take a significant period of time to prune a large table, and
+    // thread that calls this method will be blocked.
+    #[tracing::instrument(level = "debug", skip(self, schema, ctx), fields(ctx.id = ctx.get_id().as_str()))]
+    pub fn sync_prune(
+        &self,
+        ctx: &Arc<dyn TableContext>,
+        schema: DataSchemaRef,
+        push_down: &Option<Extras>,
+    ) -> Result<Vec<(usize, BlockMeta)>> {
+        futures::executor::block_on(self.prune(ctx, schema, push_down))
```
BohuTANG, line 60: Does the `block_on` need a special Runtime, like GlobalRuntime or others?

dantengsky: Not sure about if this sync version method will NEVER be called in an async context.

It seems that GlobalRuntime is a tokio runtime, and the `block_on` method it provides is a wrapper of tokio runtime/handle's `block_on`. thus, if we use GlobalRuntime::instance().block_on(...), in a async context, it will panic.

including the following scenario:

~~~
  async fn x {
       sync fn y {
          // y assumes itself in a sync context but not
          let rt : tokio::runtime::Runtime = ...;
          rt.block_on( some_async_func())   // <-- this will panic
      }
  }
~~~

zhang2014 (approved):

PR #7926, opened 2022-09-27
feat(query): migrate aggregation covariance functions to v2.

RinChanNOWWW:

## Summary

The computation logic is almost copied from v1.

Part of #7740


src/query/expression/src/types/number.rs
```diff
@@ -877,6 +877,55 @@ macro_rules! with_number_mapped_type {
     }
 }
 
+#[macro_export]
+macro_rules! with_number_data_types {
```
sundy-li, line 881: why not use `ALL_NUMERICS_TYPES` with `with_number_mapped_type`

RinChanNOWWW: This macro deals with two types simultaneously. This macro is from the old codes.

If use `with_number_mapped_type`, we need to write two code blocks.

RinChanNOWWW:
@mergify update

sundy-li (approved):

PR #7937, opened 2022-09-28
refactor(query/index): split index Filter trait to two trait: FilterBuilder and Filter

drmingdrmer:

##### refactor(query/index): split index Filter trait to two trait: FilterBuilder and Filter

For a index filter, when it's build, no more keys should be added.
Hence there should be two set of APIs: API to build a filter and
API to query a filter.

Other changes:

- Normalize type/method names.

- Add associated error type to trait FilterBuilder and Filter.

## Related Issues


BohuTANG (approved):
👍

PR #7948, opened 2022-09-28
feat(format): track memory size of VariantDeserializer.

youngsofun:

## Summary

Summary about this PR
in ndjson, a row is now decode to json::Value as a whole, so no chance to know the size of variant fields before decode.

this pr calculate_memory_size is called for each field only once,  I think the cost is acceptable.


src/query/service/src/pipelines/executor/executor_graph.rs
```diff
@@ -241,7 +241,16 @@ impl ExecutingGraph {
                 if state_guard_cache.is_none() {
                     state_guard_cache = Some(node.state.lock().unwrap());
                 }
-                let processor_state = match node.processor.event()? {
+                let event = node.processor.event()?;
+                if tracing::enabled!(tracing::Level::TRACE) {
+                    tracing::trace!(
+                        "node id:{:?}, name:{:?}, event: {:?}",
```
Xuanwo, line 247:
```suggestion
                        "node id: {:?}, name: {:?}, event: {:?}",
```

Simple change, but more beautiful.

Xuanwo (approved):

zhang2014 (approved):

PR #7955, opened 2022-09-29
fix: find smallest column for pruning unused columns

### LINKED ISSUE #7949: bug: ERROR 2013 (HY000): Lost connection to server during query
### Search before asking

### Version

main

### What's Wrong?

ERROR 2013 (HY000): Lost connection to server during query

### How to Reproduce?
case1:
```sql
DROP DATABASE IF EXISTS databend64;
CREATE DATABASE databend64;
USE databend64;
CREATE TABLE t0(c0VARCHAR VARCHAR NULL DEFAULT('3'), c1FLOAT DOUBLE NULL);
INSERT INTO t0(c0varchar) VALUES (''), ('992');

SELECT 1 FROM t0 where null NOT IN (null ) LIMIT 1;
ERROR 2013 (HY000): Lost connection to server during query

select * from t0;
ERROR 2006 (HY000): Server has gone away
Connection id:    363
Current database: databend64

+-----------+--------------------+
| c0varchar | c1float            |
+-----------+--------------------+
|           |               NULL |
| 992       |               NULL |
| 3         | 0.6593031287193298 |
| 3         | 0.6593031287193298 |
| 3         | 0.6593031287193298 |
+-----------+--------------------+
5 rows in set (0.012 sec)
```
case2:
```sql
DROP DATABASE IF EXISTS databend8;
CREATE DATABASE databend8;
USE databend8;
CREATE TABLE t0(c0BOOLEAN BOOL NULL DEFAULT(false), c1VARCHAR VARCHAR NULL DEFAULT('219'));
CREATE TABLE t1(c0INT INT NULL, c1BOOLEAN BOOL NULL);
INSERT INTO t0(c0boolean, c1varchar) VALUES (true, '87'), (false, '923462'), (false, '');

SELECT t0.c1varchar FROM t1 INNER  JOIN t0 ON (NOT (((-1973069172)-(1970171640)) NOT BETWEEN ((-510940986)/(-1895633557)) AND 1228230968)) WHERE (0.9649899005889893 BETWEEN 0.9755672812461853 AND 0.5659580826759338) GROUP BY t0.c1varchar;
ERROR 2013 (HY000): Lost connection to server during query
```

xudong963:

## Summary

Summary about this PR

Fixes #7949


RinChanNOWWW (approved):

src/query/service/src/sql/planner/optimizer/heuristic/prune_unused_columns.rs
```diff
@@ -60,9 +60,8 @@ impl UnusedColumnPruner {
                         "prewhere columns should be a subset of scan columns"
                     );
                     // `used` is the columns which prewhere scan needs to output for its upper operator.
-                    if used.is_empty() {
-                        let prewhere_columns =
-                            pw.prewhere_columns.iter().copied().collect::<Vec<_>>();
+                    let prewhere_columns = pw.prewhere_columns.iter().copied().collect::<Vec<_>>();
+                    if used.is_empty() && !prewhere_columns.is_empty() {
```
RinChanNOWWW, line 64: I think `!prehwere_column.is_empty()` should always be `true`.

xudong963: For cases in the issue https://github.com/datafuselabs/databend/issues/7949, it's empty

RinChanNOWWW: Oh, I know. It;s because the predicate in the where clause is a constant, so prehwere optimizer didn't collect it.

xudong963: Yes, for the case, we can fallback to `find_smallest_column_by_table_index`

Xuanwo (approved):

PR #7958, opened 2022-09-29
fix: shrink min max index

### LINKED ISSUE #7829: Feature: Trim Strings during the construction of min/max statistics
**Summary**

While collecting the min/max values of columns, we kept the exact values of them. For columns of type string(alike), the min/max values may be large(say, a column of type CHAR(4096)) and that makes the meta files that contain the statistics large.

It would be better if we can trim the strings to some moderate length, say 8 chars, in a way that preserves the property of min/max statistics: the trimmed max should be larger than the non-trimmed one, and the trimmed min should be lesser than the non-trimmed one.

Thus, with some loss of accuracy (slightly more likely to be false-positive, which IMO we can afford), the size of fuse table meta files could be reduced.

where the min/max vals are gathered:

(https://github.com/datafuselabs/databend/blob/dae90d856e380ea29716e87148cc69d07ccff8ff/src/query/storages/fuse/src/statistics/column_statistic.rs#L45-L53)

dantengsky:

## Summary

- [x] trim min/max values of string type columns
- [x] *sidecar task*: enable xor bloom filter index for integral types

update : 2022-09-30
move task "customize ser/de of `DataValue::String` (in ser/de of segment info)" to dedicated pr

Fixes #7829

### DIFF
--- a/src/query/storages/fuse/src/statistics/column_statistic.rs
+++ b/src/query/storages/fuse/src/statistics/column_statistic.rs
@@ -22,6 +22,7 @@ use common_fuse_meta::meta::ColumnStatistics;
 use common_fuse_meta::meta::StatisticsOfColumns;
 use common_storages_index::MinMaxIndex;
 use common_storages_index::SupportedType;
+use tracing::debug;
 
 pub fn gen_columns_statistics(data_block: &DataBlock) -> Result<StatisticsOfColumns> {
     let mut statistics = StatisticsOfColumns::new();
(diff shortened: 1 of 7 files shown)


src/query/storages/fuse/src/statistics/column_statistic.rs
```diff
@@ -104,3 +115,112 @@ pub mod traverse {
...
+                            }
+                        }
+
+                        // check the position to be replaced
+                        let replacement_point = if let Some(v) = idx {
+                            v
+                        } else {
+                            // if all the character is larger than the REPLACEMENT_CHAR
+                            // just ignore it
+                            debug!(
```
BohuTANG, line 199: Remove unused debug?

src/query/storages/fuse/src/statistics/column_statistic.rs
```diff
@@ -104,3 +115,112 @@ pub mod traverse {
...
+                            } else {
+                                r.push(STATS_REPLACEMENT_CHAR);
+                            }
+                        }
+
+                        Some(DataValue::String(r.into_bytes()))
+                    }
+                }
+                Err(_) => {
+                    debug!("Invalid utf8 string detected while collecting max statistics");
```
BohuTANG, line 219: Ditto

dantengsky: ok

BohuTANG (approved):

Issue #7856, opened 2022-09-24
CopiedFileTable could be updated even copy failed


Xuanwo:
**Summary**

`CopiedFileTable` could be updated even when we met an error during the copy. And users can't copy the same file again because the ETag matches and been filtered.

Cc @lichuang

zhang2014:
https://github.com/datafuselabs/databend/actions/runs/3149439345/jobs/5121266930

Xuanwo:
Has been fixed.

PR #7959, opened 2022-09-29
fix(processor): try fix cannot kill optimize table

### LINKED ISSUE #7956: bug: optimize table hang
**Summary**

If a table with large data set, like `optimize table xx compact` will never abort.
We need some logs to debug.

zhang2014:

## Summary

fix(processor): try fix cannot kill optimize table

Fixes #7956


zhang2014:
rerun flake stateful test failure and comment to #7856

BohuTANG:
A test:

Case1: if the compact deal with block, we can abort:
```
mysql> optimize table t7861 compact;
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
ERROR 1152 (08S01): Code: 1043, displayText = Aborted query, because the server is shutting down or the query was killed.
```

Case2: drop table xx all;
It seems we can't deal with block, so this case can not abort.

BohuTANG (approved):

PR #7963, opened 2022-09-29
fix(processor): try fix cannot kill drop table

### LINKED ISSUE #7961: bug: cannot kill drop table xx all
A test:

Case1: if the compact deal with block, we can abort: -- can be aborted
```
mysql> optimize table t7861 compact;
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
^C^C -- query aborted
ERROR 1152 (08S01): Code: 1043, displayText = Aborted query, because the server is shutting down or the query was killed.
```

Case2: drop table xx all; -- can not be aborted
It seems we can't deal with block, so this case can not abort.

_Originally posted by @BohuTANG in https://github.com/datafuselabs/databend/issues/7959#issuecomment-1261944878_

zhang2014:

## Summary

fix(processor): try fix cannot kill drop table

Fixes #7961

### DIFF
--- a/src/query/storages/fuse/src/operations/gc.rs
+++ b/src/query/storages/fuse/src/operations/gc.rs
@@ -210,12 +210,6 @@ impl FuseTable {
                             cache.pop(path);
                         }
 
-                        if aborting.load(Ordering::Relaxed) {
-                            return Err(ErrorCode::AbortedQuery(
-                                "Aborted query, because the server is shutting down or the query was killed.",
-                            ));
-                        }
-
                         self.remove_location(accessor, bloom_index_location.0.as_str())
                             .await?;
                     }
--- a/src/query/storages/fuse/src/operations/gc.rs
+++ b/src/query/storages/fuse/src/operations/gc.rs
@@ -13,6 +13,7 @@
 //  limitations under the License.
 
 use std::collections::HashSet;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
 
 use common_cache::Cache;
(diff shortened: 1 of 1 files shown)


src/query/storages/fuse/src/operations/gc.rs
```diff
@@ -228,29 +266,58 @@ impl FuseTable {
...
+                    "Aborted query, because the server is shutting down or the query was killed.",
+                ));
+            }
+
             if let Some(c) = CacheManager::instance().get_table_segment_cache() {
                 let cache = &mut *c.write();
                 cache.pop(x.as_str());
             }
+
+            if aborting.load(Ordering::Relaxed) {
```
BohuTANG, line 291: No need to check here?

BohuTANG: Let's delete the extra block if aborted before.

src/query/storages/fuse/src/operations/gc.rs
```diff
@@ -228,29 +266,58 @@ impl FuseTable {
...
+                ));
+            }
+
             let loc = locs.snapshot_location_from_uuid(id, *ver)?;
             if let Some(c) = CacheManager::instance().get_table_snapshot_cache() {
                 let cache = &mut *c.write();
                 cache.pop(loc.as_str());
             }
+
+            if aborting.load(Ordering::Relaxed) {
```
BohuTANG, line 315: Same suggestion as above :)

Xuanwo (approved):

PR #7967, opened 2022-09-29
feat(optimizer): convert outer join to inner join

### LINKED ISSUE #7957: Feature: add optimization rule to convert outer join to inner join
**Summary**

Databend supports **left/right/full outer join**.

For some cases, we can transfer them to inner join. Such as the following case:

```sql
select * from t0 left outer join t1 on ... where t0.a > 100. 
```
If t0.a is null, then t0.a > 100 is false, so we'll filter the null rows for t0.

Finally, there are no null rows in the results, it's correct to transfer outer join to inner join.

xudong963:

## Summary

Speed up outer join by converting to inner join 🚀 

Fixes #7957


src/query/service/src/sql/planner/optimizer/heuristic/heuristic.rs
```diff
@@ -47,6 +47,7 @@ pub static DEFAULT_REWRITE_RULES: Lazy<Vec<RuleID>> = Lazy::new(|| {
         RuleID::FoldCountAggregate,
         RuleID::SplitAggregate,
         RuleID::PushDownFilterScan,
+        RuleID::ReduceOuterJoin,
```
leiysky, line 50: `EliminateOuterJoin` is better.

leiysky:
I think it's better to encapsulate this optimization in `PushDownFilterJoin`.

xudong963:
> I think it's better to encapsulate this optimization in `PushDownFilterJoin`.

Why it's better to encapsulate this optimization in `PushDownFilterJoin`?

I think they are two independent optimization rules
1. two separated rules will be clearer and easy to debug.
2. `PushDownFilterJoin` can leverage the result of `EliminateOuterJoin` to push down predicates after left/right outer join is converted to inner join.

leiysky:
> 1. two separated rules will be clearer and easy to debug.
> 2\. `PushDownFilterJoin` can leverage the result of `EliminateOuterJoin` to push down predicates after left/right outer join is converted to inner join.

They are actually doing the same thing, we will decide whether convert cross join to inner join or convert outer join to inner join at the same time.

And besides, the rule name `EliminateOuterJoin` can be reserved for the optimization that exploits the functional dependency of join keys to convert outer joins into inner joins.

xudong963:
https://github.com/datafuselabs/databend/actions/runs/3157114050/jobs/5137590522 expression-related test failed @andylokandy @sundy-li

andylokandy:
A [patch](https://github.com/datafuselabs/databend/pull/7991) is going to fix the test

xudong963:
@mergify update

leiysky (approved):

Issue #7717, opened 2022-09-19
Feature: Migrate variant functions to new expression framework


b41sh:
**Summary**

Migrate variant functions to new expression framework
- [ ] array_get
- [ ] array_length
- [ ] as_type
- [ ] check_json
- [ ] get
- [ ] get_path
- [ ] json_extract_path_text
- [ ] length
- [ ] object_keys
- [ ] try_parse_json

PR #7972, opened 2022-09-30
feat(query): migrate variant func to func-v2

b41sh:

## Summary

migrate variant functions to functions-v2

```
parse_json
try_parse_json
check_json
length
object_keys
get
```

part of #7717

### DIFF
--- a/src/common/jsonb/src/functions.rs
+++ b/src/common/jsonb/src/functions.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use core::convert::TryInto;
+use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::collections::VecDeque;
 
@@ -20,11 +21,12 @@ use super::constants::*;
 use super::error::*;
 use super::jentry::JEntry;
 use super::number::Number;
+use super::value::JsonPath;
 
 // builtin functions for `JSONB` bytes without decode all Values
 
-// build `JSONB` array from items
-// Assuming that the input values is valid JSONB data
+/// Build `JSONB` array from items.
+/// Assuming that the input values is valid JSONB data.
 pub fn build_array<'a>(
     items: impl IntoIterator<Item = &'a [u8]>,
     buf: &mut Vec<u8>,
@@ -60,8 +62,8 @@ pub fn build_array<'a>(
     Ok(())
 }
 
-// build `JSONB` object from items
-// Assuming that the input values is valid JSONB data
+/// Build `JSONB` object from items.
+/// Assuming that the input values is valid JSONB data.
 pub fn build_object<'a, K: AsRef<str>>(
     items: impl IntoIterator<Item = (K, &'a [u8])>,
     buf: &mut Vec<u8>,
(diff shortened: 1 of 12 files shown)


src/common/jsonb/src/functions.rs
```diff
@@ -110,6 +110,156 @@ pub fn build_object<'a>(
...
+                jentry_offset += 4;
+                val_offset += val_jentry.length as usize;
+            }
+            None
+        }
+        _ => None,
+    }
+}
+
+// get the keys of a `JSONB` object
```
andylokandy, line 228: We can make them doc comments

```suggestion
/// Get the keys of a `JSONB` object
```

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
             Ok(())
         }),
     );
+
+    registry.register_passthrough_nullable_1_arg::<VariantType, VariantType, _, _>(
```
andylokandy, line 79: `VaraintType` should be registered as the last resort, otherwise, numbers will be cast to variant at the call of `parse_json`.

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    registry.register_1_arg_core::<NullableType<StringType>, NullableType<StringType>, _, _>(
+        "check_json",
+        FunctionProperty::default(),
+        |_| None,
+        |arg, _| match &arg {
+            ValueRef::Column(NullableColumn { validity, column }) => {
+                let size = column.len();
+                let mut bitmap = MutableBitmap::with_capacity(size);
+                let mut builder = StringColumnBuilder::with_capacity(size, 0);
+                for idx in 0..size {
```
andylokandy, line 344:
```suggestion
                for (valid, s) in validity.iter().zip(column.iter()) {
```

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    registry.register_1_arg_core::<VariantType, NullableType<VariantType>, _, _>(
+        "object_keys",
+        FunctionProperty::default(),
+        |_| None,
+        |arg, _| match &arg {
+            ValueRef::Column(column) => {
+                let size = column.len();
+                let mut bitmap = MutableBitmap::with_capacity(size);
+                let mut builder = StringColumnBuilder::with_capacity(size, 0);
+                for idx in 0..size {
```
andylokandy, line 556:
```suggestion
                for v in column.iter() {
```

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    registry.register_1_arg_core::<NullableType<VariantType>, NullableType<VariantType>, _, _>(
+        "object_keys",
+        FunctionProperty::default(),
+        |_| None,
+        |arg, _| match &arg {
+            ValueRef::Column(NullableColumn { validity, column }) => {
+                let size = column.len();
+                let mut bitmap = MutableBitmap::with_capacity(size);
+                let mut builder = StringColumnBuilder::with_capacity(size, 0);
+                for idx in 0..size {
```
andylokandy, line 599: ditto

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                };
+                Ok(val)
+            }
+        },
+    );
+
+    registry.register_function_factory("get", |_, args_type| {
+        if args_type.len() != 2 {
+            return None;
+        }
```
andylokandy, line 643: Argument constraint will be checked by type checker following the `signature.args_type`.
```suggestion
```

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            },
+            calc_domain: Box::new(|_| None),
+            eval: Box::new(get_by_index_fn),
+        }))
+    });
+
+    registry.register_function_factory("get", |_, args_type| {
+        if args_type.len() != 2 {
+            return None;
+        }
```
andylokandy, line 659: ditto

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            },
+            calc_domain: Box::new(|_| None),
+            eval: Box::new(wrap_nullable(get_by_index_fn)),
+        }))
+    });
+
+    registry.register_function_factory("get", |_, args_type| {
+        if args_type.len() != 2 {
+            return None;
+        }
```
andylokandy, line 678: ditto

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            },
+            calc_domain: Box::new(|_| None),
+            eval: Box::new(get_by_name_fn),
+        }))
+    });
+
+    registry.register_function_factory("get", |_, args_type| {
+        if args_type.len() != 2 {
+            return None;
+        }
```
andylokandy, line 694: ditto

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        ValueRef::Column(col) => Some(col.len()),
+        _ => None,
+    });
+
+    let source_arg = args[0].try_downcast::<VariantType>().unwrap();
+    let index_arg = args[1].try_downcast::<UInt64Type>().unwrap();
+    let size = len.unwrap_or(1);
+    let mut bitmap = MutableBitmap::with_capacity(size);
+    let mut builder = StringColumnBuilder::with_capacity(size, 0);
+    for idx in 0..size {
```
andylokandy, line 725:
```suggestion
    for (s, index) in source_arg.iter().zip(index_arg.iter()) {
```

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        ValueRef::Column(col) => Some(col.len()),
+        _ => None,
+    });
+
+    let source_arg = args[0].try_downcast::<VariantType>().unwrap();
+    let name_arg = args[1].try_downcast::<StringType>().unwrap();
+    let size = len.unwrap_or(1);
+    let mut bitmap = MutableBitmap::with_capacity(size);
+    let mut builder = StringColumnBuilder::with_capacity(size, 0);
+    for idx in 0..size {
```
andylokandy, line 777: ditto

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        let s = unsafe { source_arg.index_unchecked(idx) };
+        if s.is_empty() {
+            bitmap.push(false);
+            builder.commit_row();
+            continue;
+        }
+        let name = unsafe { name_arg.index_unchecked(idx) };
+        let name = String::from_utf8_lossy(name);
```
andylokandy, line 785: Should we throw an error on this case?

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                }
+            }
+            None => Ok(Value::Scalar(Scalar::Null)),
+        },
+    }
+}
+
+// boolean, variant, number types are valid json,
+// return NULL directly without check
+fn check_json_fn(size: usize) -> Result<Value<NullableType<StringType>>, String> {
```
andylokandy, line 818:
```suggestion
fn constant_empty_string_column(len: usize) -> Value<NullableType<StringType>>, String {
```

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -43,4 +75,756 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    let mut builder = StringColumnBuilder::with_capacity(size, 0);
+    let mut bitmap = MutableBitmap::with_capacity(size);
+    for _ in 0..size {
+        bitmap.push(true);
+        builder.commit_row();
+    }
+    let col = NullableColumn {
+        validity: bitmap.into(),
+        column: builder.build(),
+    };
```
andylokandy, line 828:
```suggestion
    let col = NullableColumnBuilder {
        validity: constant_bitmap(true, len),
        builder: StringColumnBuilder::repeat("", len),
    }.build();
```

src/query/functions-v2/src/scalars/variant.rs
b41sh: `source_arg` and `index_arg` are `ValueRef`, don't have `iter()` method.

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -40,4 +75,731 @@ pub fn register(registry: &mut FunctionRegistry) {
             Ok(())
         }),
     );
+
+    registry.register_passthrough_nullable_1_arg::<BooleanType, VariantType, _, _>(
+        "parse_json",
```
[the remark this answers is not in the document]
andylokandy: I was confused before but yet has no chance to make a question that 

When clickhouse adopts `CAST(... AS VARIANT)`, `TRY_CAST(... AS VARIANT)` style conversions, and snowflake adopts `parse_json(...)` and `try_parse_json(...)` style conversions, which style should we pick?

AFAICS, the `CAST` style works better with strong typed `Array` because it allows to `CAST(Array(Type) AS Array(VARIANT))`, while the `parse_json(...)` style allows users to give additional parameters (`parse_json` in snowflake doesn't  have extra parameters so I'll take `to_timestamp` as an example: [`to_timestamp(<string>, <format>)`](https://docs.snowflake.com/en/sql-reference/functions/to_timestamp.html#:~:text=timestampFunction%20(%20%3Cstring_expr%3E%20%5B%20%2C%20%3Cformat%3E%20%5D%20)) allows to specify the string format while it will be hard for `CAST(<string> AS TIMESTAMP)`)

TBO, it is sort of a mass in databend. Maybe we could discuss it before pushing forward.

cc @sundy-li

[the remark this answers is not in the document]
b41sh: > Do they duplicate with `CAST(... AS VARIANT)`, which will call t[his conversion function](https://github.com/datafuselabs/databend/blob/df03f06e55e6ea2c7d05fd2d745e9e6b692726c7/src/query/expression/src/types/variant.rs#L160)?

yes, we can remove `Boolean` and `Number`.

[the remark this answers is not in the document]
b41sh: I think we can keep them both, `CAST(... AS VARIANT)` is used to convert basic data types to `Variant`, and `parse_json(...)` is used to convert a `JSON` string to `Variant`.

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -76,42 +69,6 @@ pub fn register(registry: &mut FunctionRegistry) {
...
-                                Ok(())
-                            },
-                        ),
-                    );
-            }
-        })
-    }
-
     registry.register_passthrough_nullable_1_arg::<VariantType, VariantType, _, _>(
         "parse_json",
```
andylokandy, line 73: Can we also remove this one?

b41sh: It depends on whether we need to compatible with `Snowflake`, maybe we need some discussion.

src/query/functions-v2/src/scalars/variant.rs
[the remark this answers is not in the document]
sundy-li: I think `parse_json` can be treated as another function that converts a `JSON` string to `Variant`. It's different from standard cast rules.

src/query/functions-v2/src/scalars/variant.rs
BohuTANG: We should.

b41sh:
add other variant functions:
```
get_ignore_case
get_path
json_extract_path_text
as_boolean
as_integer
as_float
as_string
as_array
as_object
```

BohuTANG (approved):

PR #7979, opened 2022-09-30
chore(query): add convert from datavalues to new expression values

sundy-li:

## Summary

Summary about this PR

Fixes #issue


sundy-li:
Some todos should be fixed in other prs:

- `Value::Scalar` can't convert to full column correctly if it's `Scalar::NULL`, maybe we should refactor the struct `Value` in new expression.
- new expression did not cover `DataType::Date` and `DataType::Interval`.

andylokandy (approved):

leiysky (approved):

RinChanNOWWW (approved):

PR #7984, opened 2022-09-30
fix: fix the bug in logic or and add test cases.

### LINKED ISSUE #7966: bug: filter genereates wrong answer
### Search before asking

### Version

main

### What's Wrong?

```sql
mysql> select * from onecolumn;
+------+
| x    |
+------+
|   44 |
| NULL |
|   42 |
|   42 |
+------+
4 rows in set (0.03 sec)
Read 4 rows, 18.00 B in 0.006 sec., 684.59 rows/sec., 3.01 KiB/sec.

mysql> select * from onecolumn where x > 42 or x < 43;
+------+
| x    |
+------+
|   44 |
| NULL |
|   42 |
|   42 |
+------+
4 rows in set (0.04 sec)
Read 4 rows, 18.00 B in 0.007 sec., 548.09 rows/sec., 2.41 KiB/sec.

```

`NULL` should be filtered

### How to Reproduce?

_No response_

RinChanNOWWW:

## Summary

Summary about this PR

Fixes #7966


src/query/functions-v2/src/scalars/boolean.rs
```diff
@@ -145,9 +145,11 @@ pub fn register(registry: &mut FunctionRegistry) {
             }
             None
         },
-        // value = lhs | rhs,  valid = (lhs_v & rhs_v) | (lhs | rhs))
+        // value = lhs | rhs,  valid = (lhs_v & rhs_v) | (lhs_v & lhs) | (rhs_v & rhs)
         vectorize_2_arg::<NullableType<BooleanType>, NullableType<BooleanType>, NullableType<BooleanType>>(|lhs, rhs, _| {
-            let valid = (lhs.is_some() & rhs.is_some()) | (lhs.unwrap_or_default() | rhs.unwrap_or_default());
+            let lhs_v = lhs.is_some();
```
sundy-li, line 150:
```
match (lhs, rhs) {
  
   Some(true), _ => Some(true),
   _, Some(true)  => Some(true),

  Some(false), Some(false) => Some(false),

  _ => None

}
```

sundy-li: image

RinChanNOWWW: I think my codes are the same with this?

Xuanwo: I'm guessing @sundy-li is talking about readability.

RinChanNOWWW: I think my codes are the same with this?

sundy-li (approved):

PR #8005, opened 2022-10-01
feat(meta/store): add snapshot_id codec support

drmingdrmer:

## Related Issues

### DIFF
--- a/src/meta/raft-store/src/state_machine/snapshot.rs
+++ b/src/meta/raft-store/src/state_machine/snapshot.rs
@@ -13,6 +13,10 @@
 // limitations under the License.
 
 use common_meta_sled_store::openraft;
+use common_meta_types::anyerror::AnyError;
+use common_meta_types::LogId;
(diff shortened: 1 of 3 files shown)


src/meta/raft-store/src/state_machine/snapshot.rs
```diff
@@ -25,3 +29,73 @@ pub struct Snapshot {
     /// The data of the state machine at the time of this snapshot.
     pub data: Vec<u8>,
 }
+
+impl Snapshot {
+    pub fn encode_snapshot_id(last_applied: Option<LogId>, snapshot_index: u64) -> String {
```
Xuanwo, line 34: I prefer using `format_snapshot_id` and `parse_snapshot_id` to indicate that we are building/consuming a `String` instead of a binary.

Xuanwo (approved):

PsiACE (approved):

PR #7893, opened 2022-09-26
feat: parallel  bloom index pruning

dantengsky:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Besides the ongoing work of distributed index pruning  #7867.  Local parallel execution of pruning, especially the bloom index pruning,  will also boost the performance (since currently, in single node deployment, CPU and network utilization is rather low, during the planning phase).

- [X] POC, verify that it works
- [x] concurrency control
   Introduces a new setting `max_concurrent_prune`, with default value 1000.
   The max concurrent task of pruning (per query) is limited by it. 
- [x] disable buffered reader

Fixes #issue

PR #8012, opened 2022-10-01
feat(runtime): add try_spawn_batch

BohuTANG:
https://github.com/datafuselabs/databend/pull/7893

## Summary

Introduce a new function for runtime: `try_spawn_batch`, making the fixed numbers concurrency for many futures easier to use, like https://github.com/datafuselabs/databend/pull/7893.

`try_spawn_batch` is not same as: `futures::stream::iter(futures).buffer_unordered(max_concurrent)`:
* `buffer_unordered` should wait until the whole batch is finished and start the new batch.
* `try_spawn_batch` should wait for one future finish and start a new one.
* For a comparison, please see https://github.com/BohuTANG/joint

Fixes #issue


BohuTANG:
Would it be simpler for #7893 to use `try_spawn_batch`? cc @dantengsky

src/common/base/tests/it/runtime.rs
```diff
@@ -77,3 +81,43 @@ async fn test_shutdown_long_run_runtime() -> Result<()> {
 
     Ok(())
 }
+
+lazy_static! {
```
PsiACE, line 85: Perhaps we can rewrite it using once_cell.

So that in the future we can migrate painlessly from once_cell to std

src/common/base/src/base/runtime.rs
```diff
@@ -186,6 +187,33 @@ impl Runtime {
...
+    where
+        F: Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        let semaphore = Arc::new(Semaphore::new(max_concurrent));
+        let mut join_handlers = Vec::with_capacity(futures.len());
+        for future in futures {
+            let semaphore = semaphore.clone();
+            let h = self.try_spawn(async move {
+                let _segment_prune_permit = semaphore.acquire().await;
```
BohuTANG, line 210: Need rename

dantengsky:
> Would it be simpler for #7893 to use `try_spawn_batch`? cc @dantengsky

trying the apis in 
- https://github.com/datafuselabs/databend/pull/7893
, the way of passing semaphore / permits may need some tweaks, let me try to figure it out

dantengsky (approved):

BohuTANG:
This PR will be picked in https://github.com/datafuselabs/databend/pull/7893/files#diff-1cc3b24765de856d881427c893f520ad6042883fb4606a89b521a3d32fe42494R191-R216, let's close.

Issue #8029, opened 2022-10-04
bug: join order is incorrect


zhang2014:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

main branch

### What's Wrong?

mysql> explain select * from numbers(10000) where number in (select number from numbers(1000));
```
+------------------------------------------------------------+
| explain                                                    |
+------------------------------------------------------------+
| Filter                                                     |
| ├── filters: [2 (#2)]                                      |
| └── HashJoin                                               |
|     ├── join type: MARK                                    |
|     ├── build keys: [numbers.number (#0)]                  |
|     ├── probe keys: [subquery_1 (#1)]                      |
|     ├── filters: []                                        |
|     ├── TableScan(Build)                                   |
|     │   ├── table: default.system.numbers                  |
|     │   ├── read rows: 10000                               |
|     │   ├── read bytes: 80000                              |
|     │   ├── partitions total: 2                            |
|     │   ├── partitions scanned: 2                          |
|     │   └── push downs: [filters: [], limit: NONE]         |
|     └── TableScan(Probe)                                   |
|         ├── table: default.system.numbers                  |
|         ├── read rows: 1000                                |
|         ├── read bytes: 8000                               |
|         ├── partitions total: 1                            |
|         ├── partitions scanned: 1                          |
|         └── push downs: [filters: [], limit: NONE]         |
+------------------------------------------------------------+
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Issue #7837, opened 2022-09-23
Optimize subquery: tranform mark join to semi/anti join under certain conditions


xudong963:
**Summary**

Description for this feature.

PR #8030, opened 2022-10-04
feat(query): enable RuleCommuteJoin for mark join.

youngsofun:

## Summary

- [x] 1.  NumbersTable impl `  fn statistics(&self, ctx: Arc<dyn TableContext>) -> Result<Option<TableStatistics>> ` 
     - after this, cbo works with inner join on numbers
- [x]  MarkJoin can apply RuleCommuteJoin 
- [x]  logistic test 
- [x] JoinHashTable::mark_join2:  subquery as build side  ( a benefit: known hash_null early,  can output the probe table in stream)
    -  [x] set `has_null` when building build_side

Fixes  https://github.com/datafuselabs/databend/issues/8029


zhang2014:
What's happen `explain select * from numbers(1000) where number in (select number from numbers(10000));`?

youngsofun:
> explain select * from numbers(1000) where number in (select number from numbers(10000));

```
mysql> explain select * from numbers(1000) where number in (select number from numbers(10000));
+------------------------------------------------------------+
| explain                                                    |
+------------------------------------------------------------+
| Filter                                                     |
| ├── filters: [2 (#2)]                                      |
| └── HashJoin                                               |
|     ├── join type: MARK                                    |
|     ├── build keys: [numbers.number (#0)]                  |
|     ├── probe keys: [subquery_1 (#1)]                      |
|     ├── filters: []                                        |
|     ├── TableScan(Build)                                   |
|     │   ├── table: default.system.numbers                  |
|     │   ├── read rows: 1000                                |
|     │   ├── read bytes: 8000                               |
|     │   ├── partitions total: 1                            |
|     │   ├── partitions scanned: 1                          |
|     │   └── push downs: [filters: [], limit: NONE]         |
|     └── TableScan(Probe)                                   |
|         ├── table: default.system.numbers                  |
|         ├── read rows: 10000                               |
|         ├── read bytes: 80000                              |
|         ├── partitions total: 2                            |
|         ├── partitions scanned: 2                          |
|         └── push downs: [filters: [], limit: NONE]         |
+------------------------------------------------------------+

```

tests/logictest/suites/mode/standalone/explain/prune_column.test
```diff
@@ -83,6 +83,7 @@ HashJoin
             ├── partitions scanned: 1
             └── push downs: [filters: [], limit: NONE]
 
+skipif http
```
youngsofun, line 86: this test is flaky with http handler only.
seems to be a problem of http handler, not related to this pr.

Xuanwo:
```shell
error: the loop variable `idx` is used to index `probe_indexes`
   --> src/query/service/src/pipelines/processors/transforms/hash_join/result_blocks.rs:379:24
    |
379 |             for idx in 0..boolean_viewer.len() {
    |                        ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::needless-range-loop` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_range_loop
help: consider using an iterator
    |
379 |             for (idx, <item>) in probe_indexes.iter().enumerate().take(boolean_viewer.len()) {
    |                 ~~~~~~~~~~~~~    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error: could not compile `databend-query` due to previous error
```

src/query/service/src/pipelines/processors/transforms/hash_join/join_hash_table.rs
```diff
@@ -542,10 +543,21 @@ impl HashJoinState for JoinHashTable {
         }
 
         let mut chunks = self.row_space.chunks.write().unwrap();
+        let mut has_null = false;
         for chunk_index in 0..chunks.len() {
             let chunk = &mut chunks[chunk_index];
             let mut columns = Vec::with_capacity(chunk.cols.len());
             let markers = if self.hash_join_desc.join_type == Mark {
+                if self.hash_join_desc.marker_join_desc.subquery_as_build_side && !has_null {
+                    if let Some(validity) = chunk.cols[0].validity().1 {
```
xudong963, line 552: Here, `cols` in `chunk` are `build_keys`(join attributes for build side), if any column in `cols` has null, `has_null` will be set to true.

src/query/service/src/pipelines/processors/transforms/hash_join/result_blocks.rs
```diff
@@ -270,6 +281,118 @@ impl JoinHashTable {
...
+    where
+        Key: HashTableKeyable + Clone + 'static,
+        IT: Iterator<Item = Key> + TrustedLen,
+    {
+        let has_null = {
+            let has_null_ref = self.hash_join_desc.marker_join_desc.has_null.read();
+            *has_null_ref
+        };
+
+        let mut markers = vec![MarkerKind::False; input.num_rows()];
```
xudong963, line 337: For markers's initialization, we need to consider `NULL`

```
for each r in R
  if r.a is NULL
    mark r as NULL
  else
    mark r as FALSE
```

src/query/service/src/pipelines/processors/transforms/hash_join/result_blocks.rs
```diff
@@ -270,6 +281,118 @@ impl JoinHashTable {
...
+        let valids = &probe_state.valids;
+        if self.hash_join_desc.other_predicate.is_none() {
+            // todo(youngsofun): can be optimized as semi-join?
```
xudong963, line 340: I'll do an optimization later: convert mark join to semi join, https://github.com/datafuselabs/databend/issues/7837

src/query/service/src/pipelines/processors/transforms/hash_join/result_blocks.rs
```diff
@@ -270,6 +281,118 @@ impl JoinHashTable {
         Ok(())
     }
 
+    fn create_marker_block(&self, has_null: bool, markers: Vec<MarkerKind>) -> Result<DataBlock> {
```
xudong963, line 284: It seems the method is duplicated with `mark_join_blocks`, better to extract the same logic to one method.

xudong963 (approved):
Generally, look great to me. Thanks @youngsofun !

I still have two suggestions:
- Before the PR, I implemented mark join, strictly to speak, it's **left mark join**, probe side is subquery, build side is outer table. I noticed you introduce `subquery_as_build_side` to distinguish your mark join with mine. I think it better to introduce a **right mark join** join type.
- Maybe it's better to add some result tests to check if results are correct after left mark join to right mark join?

src/query/service/src/pipelines/processors/transforms/hash_join/result_blocks.rs
youngsofun: ok

xudong963:
> > Before the PR, I implemented mark join, strictly to speak, it's left mark join, probe side is subquery, build side is outer table. I noticed you introduce subquery_as_build_side to distinguish your mark join with mine. I think it better to introduce a right mark join join type.
> 
> do I need to rename the origin mark join `left mark join`?

couldn't be better！

youngsofun:
> For markers's initialization, we need to consider NULL


this seems like a bug:

https://github.com/datafuselabs/databend/blob/main/src/query/service/src/pipelines/processors/transforms/hash_join/join_hash_table.rs#L550

according to the comment ```Only all columns' values are NULL, we set the marker to Null.``` and usage of combine_validities_3.
should not break if one column is_all_null?

src/query/service/src/pipelines/processors/transforms/hash_join/result_blocks.rs
youngsofun: refactor done

src/query/service/src/pipelines/processors/transforms/hash_join/result_blocks.rs
```diff
@@ -270,6 +279,153 @@ impl JoinHashTable {
...
+    pub(crate) fn init_markers(cols: &[ColumnRef]) -> Vec<MarkerKind> {
+        let num_rows = cols[0].len();
+        let mut markers = vec![MarkerKind::False; cols[0].len()];
+        if cols.iter().any(|c| c.is_nullable() || c.is_null()) {
+            let mut valids = None;
+            for col in cols.iter() {
+                let (is_all_null, tmp_valids_option) = col.validity();
+                if !is_all_null {
+                    if let Some(tmp_valids) = tmp_valids_option.as_ref() {
+                        if tmp_valids.unset_bits() == 0 {
```
youngsofun, line 332: @xudong963 the fix is here

youngsofun:
failed tests.

```
ErrorType: statement query get result not equal to expected
Message: 
 Expected:
42 53
44 NULL
NULL NULL                                                         
 Actual:
42  53
NULL  51
NULL  52
NULL  45                                               
 Statement:
Parsed Statement
    at_line: 335,
    s_type: Statement: query, type: IT, query_type: IT, retry: False,
    suite_name: crdb/join,
    text:
        SELECT o.x, t.y FROM onecolumn o LEFT OUTER JOIN twocolumn t ON (o.x=t.x AND t.y=53) order by o.x;
    results: [(<re.Match object; span=(0, 4), match='----'>, 341, '42 53\n44 NULL\nNULL NULL')],
    runs_on: {'http', 'clickhouse', 'mysql'},
 Start Line: 341, Result Label: 
```


https://github.com/datafuselabs/databend/actions/runs/3194269740/jobs/5213747393

```
The failure tests:
    /workspace/tests/suites/0_stateless/13_tpch/13_0004_q4.sql
    /workspace/tests/suites/0_stateless/13_tpch/13_0013_q13.sql
    /workspace/tests/suites/0_stateless/13_tpch/13_0022_q22.sql
```

youngsofun:
bugs fixed

PR #8031, opened 2022-10-05
feat: enable join commute for more join types

xudong963:

## Summary

We have supported right join, right semi join, right anti join, so it's time to enable join commute rule for those join types

Fixes #issue


xudong963:
@mergify update

BohuTANG (approved):

PR #8051, opened 2022-10-08
feat(expression): numeric/string can  cast to boolean when it's in filter

sundy-li:

## Summary

Summary about this PR

1. feat(expression): numeric/string can  cast to boolean when it's in filter
2. fix nullable with div by zero corner case 
3. improve bitmap collect from trusted iterator

Fixes #issue


sundy-li:
Looks ambiguous after this auto-cast feature.
```
select  not 33;  -- false
select not 'abc'; -- false
select 3 and 'abc'; -- false
select false or 'ab'; -- true
```

I would like to keep the old way.

sundy-li:
enable it in the filter kernel.

src/query/expression/src/kernels/filter.rs
```diff
@@ -84,21 +87,42 @@ impl Chunk {
     // Must be nullable boolean or boolean value
```
andylokandy, line 87: should update the comment

src/query/expression/src/kernels/filter.rs
```diff
@@ -84,7 +84,7 @@ impl Chunk {
         }
     }
 
-    // Must be nullable boolean or boolean value
+    // Must be numeric/boolean or string value type
```
andylokandy, line 87:
```suggestion
    // Must be numeric, boolean, or string value type
```

andylokandy (approved):

sundy-li:
@mergify update

PR #8058, opened 2022-10-09
feat(query): support `Date` and `Interval` in new expression framework

### LINKED ISSUE #6635: Support Timestamp, Date, Interval in new expression framework
`Date` shares the same value type as `i32`, when `Interval` and `Timestamp` shares the one as `i64`. Therefore the tasks are simpler because we don't need to add value varients for `Scalar` and `Column`:

1. Add logical type `DataType::{Timestamp, Date, Interval}`.

2.  
   a. Add generic type `common_expression::types::number::{TimestampType, DateType, IntervalType}`, and implement `Number` trait for them. This is the current strategy using in `common_functions`.

   b. An alternative is to add the generic types `common_expression::types::{TimestampType, DateType, IntervalType}`, and implement `ValueType` and `ArgType` trait for them. In this way, we would introduce more duplicated code, but also unlocked the possibility to decorate the scalar to be more informative types like `std::duration::Duration` or something alike in `chrono`. cc @sundy-li 

3. Add `can_cast_to` rule and `Evaluator::cast` rule between temporal types and native number types.

RinChanNOWWW:

## Summary

- [x] Basic type implementation.
- [x] Display.
- [x] Cast with number types.

Fixes #6635


sundy-li (approved):

src/query/expression/src/kernels/scatter.rs
```diff
@@ -121,6 +123,18 @@ impl Column {
                 indices,
                 scatter_size,
             ),
+            Column::Date(column) => Self::scatter_scalars::<Int32Type, _>(
```
sundy-li, line 126: If we map DateType into Int32Type, how could we register functions for DateType ?

BohuTANG (approved):
+1

PR #8080, opened 2022-10-09
feat(planner): Top-down optimization

### LINKED ISSUE #7508: Feature: Implement top-down optimization for cascades optimizer
Currently, the cascades optimizer of databend is implemented in a recursive fashion instead of top-down, which is easy to maintain but can not really enumerate all the search space.

Besides, we can hardly implement lower bound optimization without top-down optimization.

leiysky:

## Summary

Introduce a top-down approach to perform query optimization, which will explore groups on demand instead of exhausting all the search space all the time.

With this algorithm, we can achieve more fine-grained control of query optimization and also guide to more complete search space.

Close #7508


src/query/service/src/sql/planner/optimizer/cascades/scheduler.rs
```diff
@@ -0,0 +1,49 @@
...
+                continue;
+            }
+            task.execute(optimizer, self)?;
+        }
+
+        Ok(())
+    }
+
+    pub fn add_task(&mut self, task: Task) {
+        self.task_queue.push_back(task);
```
youngsofun, line 47: Cascades and Columbia use stack?
why use a queue here?
I thought the search progress is depth-first with a stack, so we get `final` cost of some part of the tree earlier, which can be used as Upperbound later .

youngsofun: also no need ref count if using a stack?

leiysky: > why use a queue here? Cascades and Columbia use stack. I thought the search progress is depth-first with a stack, so we get `final` cost of some part of the tree earlier, which can be used as Upperbound later .

It's not actually a depth-first fashion, the task on a subtree can be terminated by another task on another subtree. 

I think there isn't much difference between queue and stack here. I would rather implement this with a recursion approach than using a state machine with the stack.

youngsofun: >  I think there isn't much difference between queue and stack here
so what`s the reason to prefer queue here?

I guess stack make pruning happen earlier,  especially coupled with rule promise.

youngsofun: > Task objects offer significantly more flexibility than procedure invocations, in particu- lar with respect to search algorithm and search control. A task object exists for each task that has yet to be done; all such task objects are collected in a task structure. The task structure is currently realized as a last-in-first-out stack; however, other structures can easily be envisioned. In particular, task objects can be reordered very easily at any point, enabling very flexible mechanisms for heuristic guidance. Moreover, we plan on representing the task structure by a graph that captures dependencies or the topological ordering among tasks and permit efficient parallel search (using shared memory). However, in order to obtain a working system fast, the current implemen- tation is restricted to a LIFO stack, and scheduling a task is very similar to invoking a function, with the exception that any work to be done after a sub-task completes must be scheduled as a separate task

youngsofun (approved):

BohuTANG (approved):

xudong963 (approved):

xudong963:
A flakey test in `test_stateless_cluster_linux`

```
13_0004_q4:                                                             [ FAIL ] - result differs with:
--- /workspace/tests/suites/0_stateless/13_tpch/13_0004_q4.result	2022-10-10 13:26:45.999800048 +0000
+++ /workspace/tests/suites/0_stateless/13_tpch/13_0004_q4.stdout	2022-10-10 13:29:41.519447115 +0000
@@ -1,5 +1,5 @@
-1-URGENT	999
-2-HIGH	997
-3-MEDIUM	1031
-4-NOT SPECIFIED	989
-5-LOW	1077
+1-URGENT	1001
+2-HIGH	995
+3-MEDIUM	1029
+4-NOT SPECIFIED	993
+5-LOW	1075
```
I'll take a look.

PR #8093, opened 2022-10-10
fix(function-v2): fix concat_ws

andylokandy:

## Summary

concat_ws was having not output rows on nullable column inputs.


sundy-li (approved):

src/query/expression/src/types/timestamp.rs
```diff
@@ -277,7 +288,7 @@ impl TimestampColumnBuilder {
     }
 
     pub fn push_default(&mut self) {
-        self.ts.push(0);
+        self.ts.push(MIN_TIMESTAMP.ts);
```
b41sh, line 291: the default value of the timestamp should be `1970-01-01 00:00:00`

b41sh (approved):

PR #8095, opened 2022-10-10
chore(sql): refactor debug assert in parse_stage_location_v2

### LINKED ISSUE #7772: Refactor `debug_assert` in parse_stage_location_v2
Why use `debug_assert` here?

_Originally posted by @BohuTANG in https://github.com/datafuselabs/databend/pull/7771#discussion_r976050519_

ClSlaid:

## Summary

Fixes #7772


src/query/service/src/sql/planner/binder/copy.rs
```diff
@@ -562,7 +562,10 @@ pub async fn parse_stage_location_v2(
     name: &str,
     path: &str,
 ) -> Result<(UserStageInfo, String)> {
-    debug_assert!(path.starts_with('/'), "path should starts with '/'");
+    if !path.starts_with('/') {
+        let e = ErrorCode::SyntaxException("path should starts with '/'");
```
Xuanwo, line 566: Don't expose internal logic to users. I use `debug_assert` here because I expect input path must be `starts_with('/')`. If not, something must be wrong.

Please check the caller of `parse_stage_location_v2` and add unit tests for them to make it sure.

ClSlaid:
There are only 2 sources of path: `CopyUnit::StageLocation` and `PresignStmt::StageLocation`. The `path` field of those structures is passed into the function, unedited.

It's able to ensure that the outcome of those AST will always come out with to be starting with '/'. But if someday we added another statement, this PR might not able to cover it.

src/query/ast/tests/it/parser.rs
```diff
@@ -253,6 +253,15 @@ fn test_statement() {
                     skip_header = 1
                 )
                 size_limit=10;"#,
+        r#"COPY INTO mytable
+                FROM @external_stage/path/to/dir/
```
Xuanwo, line 257: Why will happen if user input `@external_stage/path/to/dir`?

Xuanwo:
After adding new tests for the parser, I think we can just remove this check and add a comment for this function.

Xuanwo (approved):

BohuTANG (approved):

PR #8097, opened 2022-10-10
feat(query): ignore deprecated setting apply

sundy-li:

## Summary

Summary about this PR

Fixes #issue


BohuTANG (approved):

src/query/settings/src/lib.rs
```diff
@@ -90,7 +90,12 @@ impl Settings {
             for global_setting in global_settings {
                 let name = global_setting.name;
                 let val = global_setting.value.as_string()?;
-                settings.set_settings(name, val, false)?;
+
+                // the settings may be deprecated
+                if !settings.has_setting(&name) {
+                    continue;
+                }
+                tracing::info!("Apply global setting {} = {}", name, val);
```
zhang2014, line 98: Maybe warn if not has setting is better? because we create settings for each session(too many log in http).

Issue #8067, opened 2022-10-09
performance: fuse_snapshot very slow


BohuTANG:
**Summary**

```
mysql> select count() from fuse_snapshot('db7861', 't7861');

+---------+
| count() |
+---------+
|    3179 |
+---------+
1 row in set (12 min 2.99 sec)
```

Task:
- [x] Get all the ss files from the root path with `list` https://github.com/datafuselabs/databend/pull/8098
- [x] Change `optimize table xx purge` to parallel reads

BohuTANG:
Now the snapshot read is using opendal `list`and read in parallel, it's fast than before, let's close.

PR #8098, opened 2022-10-10
chore: try to fast read_snapshot

BohuTANG:

## Summary

* Add `read_snapshots_by_root_file` to use `list` fast the snapshot reads. This is the v1 to fast snapshot, the final plan waits for @dantengsky  to add some hist to the snapshot file, and we can avoid using the `list` operation.
* Change `fuse_snapshot_table` from stream to `AsyncSource`, and align it with other functions as fuse_segment did.

Parts #8067


src/query/storages/fuse/src/fuse_snapshot.rs
```diff
@@ -12,36 +12,97 @@
...
+    ctx: Arc<dyn TableContext>,
+    root_snapshot_file: String,
+    format_version: u64,
+    data_accessor: &Operator,
+) -> Result<Vec<Result<Arc<TableSnapshot>>>> {
+    let mut snapshot_files = vec![];
+    if let Some(path) = Path::new(&root_snapshot_file).parent() {
+        let snapshot_prefix = path.to_str().unwrap_or("");
+        if snapshot_prefix.contains('/') {
+            let mut ds = data_accessor.object(snapshot_prefix).list().await?;
```
Xuanwo: We don't have a limit here. What will happen if we have many snapshots?

dantengsky:
TableSnapshotLite :+1:

dantengsky (approved):
LGTM

PR #8106, opened 2022-10-10
feat(rbac): Add builtin PUBLIC role

### LINKED ISSUE #6827: RBAC: Add public role
**Summary**

- add a PUBLIC role as the initial roles, currently includes account_admin
- PUBLIC role is granted to ALL other roles are implicitly
- PUBLIC role is granted to ALL users implicitly

flaneur2020:

## Summary

This PR adds another builtin role PUBLIC, which plays as the builtin role with the lowest privileges.

It's also a special role, which is the child by every other role by default. Thus every other role can access the data objects owned by PUBLIC role, any role can switch to PUBLIC role.

The opposite side is the ACCOUNT_ADMIN role, which plays as the highest builtin role. ACCOUNT_ADMIN can access every data objects owned by any role, and can be switched to any role.

Fixes #6827

### DIFF
--- a/src/query/users/src/role_mgr.rs
+++ b/src/query/users/src/role_mgr.rs
@@ -23,6 +23,9 @@ use common_meta_types::UserPrivilegeSet;
 use crate::role_util::find_all_related_roles;
 use crate::UserApiProvider;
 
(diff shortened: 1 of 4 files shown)


BohuTANG (approved):
👍

PR #8126, opened 2022-10-11
fix(parser): fix version()

sandflee:

## Summary
`select version_xx from table`  generates databend version info.


src/query/service/src/servers/mysql/mysql_federated.rs
```diff
@@ -177,7 +177,7 @@ impl MySQLFederated {
     fn federated_mixed_check(&self, query: &str) -> Option<DataBlock> {
         let rules: Vec<(&str, Option<DataBlock>)> = vec![
             (
-                "(?i)^(SELECT VERSION())",
+                r"(?i)^(SELECT VERSION\(\s*\))",
```
BohuTANG, line 180: Can you give a raw query sent from MySQL?

sandflee:
```
SELECT version_name,\n       os,\n       p_date, file_type, video_succ_cnt\nFROM xxx\nWHERE p_date >= '20190101'\n  AND p_date <= '20191010'\nORDER BY version_name,\n         os,\n         p_date, file_type, video_succ_cnt\nLIMIT 100
```

Xuanwo: The old regex `(?i)^(SELECT VERSION())` is wrong and will match all query that start with `select version`

image

BohuTANG: 👍 
Could you add a test for this query in https://github.com/datafuselabs/databend/blob/main/src/query/service/tests/it/servers/mysql/mysql_federated.rs

sandflee: updated

Xuanwo (approved):

b41sh (approved):

PR #8148, opened 2022-10-12
chore(ci): get pre release

PsiACE:

## Summary

In the previous fix, I tried to provide the github token but could not find the release and was prompted with a 404.

Given that we have always released Pre-releases, a new action has been introduced to try and resolve this issue.

Fixes #issue


PsiACE:
The downside of GitHub workflow or any other CI is that we lack an effective way to simulate it in a real environment. I'm not sure what I'm missing. :cry:

everpcpc (approved):
😢

ariesdevil (approved):

PR #8150, opened 2022-10-12
feat: add share table meta info

lichuang:

## Summary

1. add share name and tenant into `Tableinfo`, if it is a share table, can use these informations to get where this table came from.
2. if it is a table share from other tenant, set engine to `SHARE_FUSE`, reuse `FuseTable` to handle share table, next pr will add `SHARE_FUSE` table logic in `FuseTable`.

Fixes #issue


drmingdrmer (approved):

BohuTANG (approved):
+1

src/query/service/src/storages/storage_factory.rs
```diff
@@ -86,12 +86,17 @@ impl StorageFactory {
...
         });
 
-        // Register FUSE table engine.
+        // Register FUSE and SHARE_FUSE table engine.
         creators.insert("FUSE".to_string(), Storage {
             creator: Arc::new(FuseTable::try_create),
             descriptor: Arc::new(FuseTable::description),
         });
 
+        creators.insert("SHARE_FUSE".to_string(), Storage {
```
BohuTANG, line 95: This will cause the test failed, because we register twice FUSE engine.

BohuTANG:
Is share table will be a new table engine not in fuse table?

drmingdrmer (approved):

PR #8154, opened 2022-10-12
refactor(query): remove precision in timestamp type

sundy-li:

## Summary

Summary about this PR

remove precision in timestamp type

discussion:  https://github.com/datafuselabs/databend/discussions/8149

Fixes #issue


b41sh (approved):

drmingdrmer (commented):
Everything is good except there should be a test case that loads the `TimestampType` message of the version `15`, similar to this:
https://github.com/datafuselabs/databend/blob/fb9c11c8c5cf5b7e9c18831dc66532896dd7d1b7/src/meta/proto-conv/tests/it/user_stage.rs#L305

src/query/ast/src/parser/expr.rs
```diff
@@ -1126,9 +1126,7 @@ pub fn type_name(i: Input) -> IResult<TypeName> {
     let ty_date = value(TypeName::Date, rule! { DATE });
     let ty_datetime = map(
         rule! { (DATETIME | TIMESTAMP) ~ ( "(" ~ #literal_u64 ~ ")" )? },
```
RinChanNOWWW, line 1128: Also remove `#literal_u64` ?

sundy-li: Some SQL dialects may need this, I prefer just keep it now.

drmingdrmer (approved):

RinChanNOWWW (approved):
LGTM

PR #8191, opened 2022-10-13
refactor(expr): improve combine_nullable

andylokandy:

## Summary

Summary about this PR

Fixes #issue

### DIFF
--- a/src/query/functions-v2/src/scalars/variant.rs
+++ b/src/query/functions-v2/src/scalars/variant.rs
@@ -69,7 +69,7 @@ pub fn register(registry: &mut FunctionRegistry) {
         |_| None,
         vectorize_with_builder_1_arg::<StringType, NullableType<VariantType>>(|s, output, _| {
             if s.trim().is_empty() {
-                output.push(Some(JSONB_NULL));
+                output.push(JSONB_NULL);
             } else {
                 match parse_value(s) {
                     Ok(value) => {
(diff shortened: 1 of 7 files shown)


sundy-li (approved):

b41sh (approved):

src/query/functions-v2/src/scalars/variant.rs
```diff
@@ -127,15 +124,11 @@ pub fn register(registry: &mut FunctionRegistry) {
...
             if v.is_empty() {
-                output.validity.push(false);
-                output.builder.commit_row();
+                output.push_null()
             } else {
                 match object_keys(v) {
-                    Some(keys) => output.push(Some(&keys)),
-                    None => {
-                        output.validity.push(false);
-                        output.builder.commit_row();
```
Xuanwo, line 137: We don't need to `commit_row` in this case?

src/query/functions-v2/tests/it/scalars/testdata/array.txt
```diff
@@ -48,7 +48,7 @@ raw expr       : length(array())
 checked expr   : length<Array(Nothing)>(array<>())
 optimized expr : 0_u8
 output type    : UInt8
-output domain  : Unknown
```
Xuanwo, line 51: Can you add a case for `a = 888`?

src/query/functions-v2/src/scalars/variant.rs
b41sh: `push_null` call `Variant::push_default`, which will call `commit_row`.

src/query/functions-v2/tests/it/scalars/testdata/array.txt
b41sh: This case is just for an empty array, it will always return zero.
We already have some cases for no empty array.

PR #8198, opened 2022-10-13
feat: add share table query

lichuang:

## Summary

1. feat: add share table query
2. add `share_endpoint_address` config, if it is empty, create a dummy operator.

Fixes https://github.com/datafuselabs/databend/issues/7820


src/common/storage/src/operator.rs
```diff
@@ -255,6 +255,7 @@ fn init_obs_operator(cfg: &StorageObsConfig) -> Result<Operator> {
 #[derive(Clone, Debug)]
 pub struct StorageOperator {
     operator: Operator,
+    share_endpoint_address: Option<String>,
```
Xuanwo, line 258: We should not touch this struct.

src/common/storage/src/operator.rs
```diff
@@ -271,20 +272,25 @@ static STORAGE_OPERATOR: OnceCell<Singleton<StorageOperator>> = OnceCell::new();
 impl StorageOperator {
     pub async fn init(
         conf: &StorageConfig,
+        share_endpoint_address: &str,
         v: Singleton<StorageOperator>,
     ) -> common_exception::Result<()> {
-        v.init(Self::try_create(conf).await?)?;
+        v.init(Self::try_create(conf, share_endpoint_address).await?)?;
```
Xuanwo, line 278: same here.

src/query/sharing/src/layer.rs
```diff
@@ -109,4 +142,23 @@ impl Accessor for SharedAccessor {
         let resp = self.client.send_async(req).await?;
         Ok(resp.into_body().reader())
     }
+
+    async fn stat(&self, path: &str, args: OpStat) -> Result<ObjectMetadata> {
+        let (_, _) = (path, args);
+        Ok(ObjectMetadata::new(ObjectMode::FILE))
```
Xuanwo, line 148: We don't need metadata? What happens if this file does not exist?

src/common/storage/src/operator.rs
lichuang: we need some way to get the share endpoint address when creating FuseTable, where will be more suitable?

Xuanwo: How about placing it in table options?

lichuang: `TableInfo` is return by meta server, and meta is unaware of share endpoint.

Xuanwo: I think we need to refactor the `get_table_info` API and insert `endpoint_address` there.

src/query/storages/fuse/src/fuse_table.rs
```diff
@@ -157,13 +172,28 @@ impl FuseTable {
...
     }
 
-    pub fn snapshot_loc(&self) -> Option<String> {
-        let options = self.table_info.options();
-        options
-            .get(OPT_KEY_SNAPSHOT_LOCATION)
-            // for backward compatibility, we check the legacy table option
-            .or_else(|| options.get(OPT_KEY_LEGACY_SNAPSHOT_LOC))
-            .cloned()
+    pub async fn snapshot_loc(&self) -> Option<String> {
```
BohuTANG, line 175: It's better to return `Result<Option<String>>`

BohuTANG:
LGTM, cc @dantengsky for final review :)

src/common/exception/src/exception_code.rs
```diff
@@ -145,6 +145,7 @@ build_exceptions! {
     InvalidTimestamp(1080),
     InvalidClusterKeys(1081),
     UnknownFragmentExchange(1082),
+    InvalidUtf8(1083),
```
Xuanwo, line 148: I don't think it's a meaning error code that should be returned to users. How about use `InternalError`?

dantengsky:
@lichuang sorry for the late response, being trapped in the bug-fixing task.

dantengsky (approved):
LGTM

BohuTANG (approved):

PR #8242, opened 2022-10-16
fix(sessions): fix dead lock when force kill sessions

zhang2014:

## Summary

fix(sessions): fix dead lock when force kill sessions

Fixes #issue


drmingdrmer (approved):
💯

BohuTANG (approved):

PR #8254, opened 2022-10-17
feat(meta/schema_api): add API list_all_tables

drmingdrmer:

##### feat(meta/schema_api): add API list_all_tables

`list_all_tables()` returns a list of every table that stored in
the meta-store.

It is meant for interanl usages such cleaning up garbage data(e.g.
cleaning up unreferenced object in s3).

## Related Issues

### DIFF
--- a/src/meta/api/src/schema_api_impl.rs
+++ b/src/meta/api/src/schema_api_impl.rs
@@ -101,9 +101,12 @@ use common_meta_types::errors::app_error::WrongShareObject;
 use common_meta_types::ConditionResult;
 use common_meta_types::GCDroppedDataReply;
 use common_meta_types::GCDroppedDataReq;
+use common_meta_types::InvalidReply;
 use common_meta_types::KVAppError;
 use common_meta_types::MatchSeqExt;
+use common_meta_types::MetaError;
 use common_meta_types::MetaId;
+use common_meta_types::MetaNetworkError;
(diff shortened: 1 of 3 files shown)


src/meta/api/src/schema_api_impl.rs
```diff
@@ -1017,6 +1020,33 @@ impl<KV: KVApi> SchemaApi for KV {
...
+    async fn list_all_tables(&self) -> Result<Vec<(TableId, u64, TableMeta)>, KVAppError> {
+        debug!("SchemaApi: {}", func_name!());
+
+        let reply = self
+            .prefix_list_kv(&vec![TableId::PREFIX, ""].join("/"))
+            .await?;
+
+        let mut res = vec![];
+
+        for (kk, vv) in reply.into_iter() {
```
lichuang, line 1036: If the data is too big, may be limit the output?

drmingdrmer: Fix it in another issue

BohuTANG: I think it's ok now, but maybe the meta server needs to throttle.

BohuTANG (approved):

BohuTANG:
We can do the gc for the s3 directory with this API now, cc @everpcpc @flaneur2020

Issue #8208, opened 2022-10-14
Feature(http handler): check status after query is detached (by call final_url)


youngsofun:
**Summary**

Description for this feature.

clarify `detach` as :

1.  query will continue run in background, writing data to result_table,  so client can download the compelect result later.
2.  server no longer maintains pages in memory.
3.  client can still poll the status and progress.

todo 

- [ ] client can still poll the status and progress. after final/detach is called.
- [ ] when call `download`, check if the query is running first.

youngsofun:
~ curl  --header 'Content-Type: application/json'  --request POST '127.0.0.1:8000/v1/query/'  --data-raw '{"sql": "select sleep(300)"}' -u root:
{"id":"b83c5555-c1ee-44e8-9d51-c511434da0dd","session_id":"e5569b31-853d-4a2c-979f-192fdba5f6a0","session":null,"schema":{"fields":[],"metadata":{}},"data":[],"state":"Running","error":null,"stats":{"scan_progress":{"rows":1,"bytes":1},"write_progress":{"rows":0,"bytes":0},"result_progress":{"rows":0,"bytes":0},"running_time_ms":1003.7266250000001},"affect":null,"stats_uri":"/v1/query/b83c5555-c1ee-44e8-9d51-c511434da0dd","final_uri":"/v1/query/b83c5555-c1ee-44e8-9d51-c511434da0dd/final","next_uri":"/v1/query/b83c5555-c1ee-44e8-9d51-c511434da0dd/page/0","kill_uri":"/v1/query/b83c5555-c1ee-44e8-9d51-c511434da0dd/kill"}%
➜  ~ curl  --header 'Content-Type: application/json'  --request POST '127.0.0.1:8000/v1/query/'  --data-raw '{"sql": "select 1"}' -u root:
➜  ~ curl -u root: 127.0.0.1:8000/v1/query/b83c5555-c1ee-44e8-9d51-c511434da0dd/final
➜  ~ curl -u root: 127.0.0.1:8000/v1/query/b83c5555-c1ee-44e8-9d51-c511434da0dd/download
Code: 1072, displayText = result for query_id b83c5555-c1ee-44e8-9d51-c511434da0dd not exists.%
```

after 300s

```
➜  ~ curl -u root: 127.0.0.1:8000/v1/query/b83c5555-c1ee-44e8-9d51-c511434da0dd/download
0
```

youngsofun:
[...]

PR #8258, opened 2022-10-17
feat(http handler): not remove query when detached.

youngsofun:

## Summary

Summary about this PR

Fixes https://github.com/datafuselabs/databend/issues/8208


flaneur2020:
what's detached means?

src/query/service/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -168,7 +168,7 @@ async fn query_detach_handler(
     Path(query_id): Path<String>,
 ) -> impl IntoResponse {
     let http_query_manager = HttpQueryManager::instance();
-    match http_query_manager.remove_query(&query_id).await {
```
flaneur2020, line 171: not remove_query ?

PR #8205, opened 2022-10-13
fix(storage): support mutation during insertion

zhyass:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Support compact、recluster and delete during insert.

In order to quickly fix bug, the code will be refactored sooner or later.

Fixes #8108

PR #8261, opened 2022-10-17
Feat: compact segment

### LINKED ISSUE #8247: Feature: optimize table t compact segment only
**Summary**

new option for optimizing table

~~~
 optimize table t compact [segment]
~~~

to compact table meta data only (mainly the segments). This operation is expected 
- to be much lighter than `optimize table t compact`,
- and can be invoked concurrently while the same table is being appended.

There are the cases that a table is kept being appended will large data chunks(but not large enough to construct a compact segment -- at least 1000 blocks, for the default table options), and at a relatively high frequency(say several commits per sec); the blocks generated are mostly non-fragmented, but the segments are not compact (usually only contain 1 ~10  meta data of blocks).

the un-compact segments harm both the reading and the data ingestion of the table
- for the former, reading lots of fragmented segments harms
- for the latter, the size  of snapshot metadata, and the time to write it down,  increases,

dantengsky:

## Summary

WIP

Fixes #8247

### DIFF
--- a/src/query/storages/fuse/src/operations/commit.rs
+++ b/src/query/storages/fuse/src/operations/commit.rs
@@ -351,6 +351,62 @@ impl FuseTable {
         Ok((seg_locs, s))
     }
 
+    pub fn merge_segments(
+        segments: &[(&Location, &SegmentInfo)],
+    ) -> Result<(Vec<String>, Statistics)> {
+        let (s, seg_locs) = segments.iter().try_fold(
+            (Statistics::default(), Vec::with_capacity(segments.len())),
+            |(mut acc, mut seg_acc), (location, segment_info)| {
+                let loc = &location.0;
+                let stats = &segment_info.summary;
+                acc.row_count += stats.row_count;
+                acc.block_count += stats.block_count;
+                acc.uncompressed_byte_size += stats.uncompressed_byte_size;
+                acc.compressed_byte_size += stats.compressed_byte_size;
+                acc.index_size = stats.index_size;
+                acc.col_stats = if acc.col_stats.is_empty() {
+                    stats.col_stats.clone()
+                } else {
+                    statistics::reduce_block_statistics(&[&acc.col_stats, &stats.col_stats])?
+                };
+                seg_acc.push(loc.clone());
+                Ok::<_, ErrorCode>((acc, seg_acc))
+            },
+        )?;
+
+        Ok((seg_locs, s))
+    }
+
+    pub fn merge_segments_new<'a>(
+        mut segments: impl Iterator<Item = (&'a Location, &'a SegmentInfo)>,
+    ) -> Result<(Vec<String>, Statistics)> {
+        let len_hint = segments
+            .size_hint()
+            .1
+            .unwrap_or_else(|| segments.size_hint().0);
+        let (s, seg_locs) = segments.try_fold(
+            (Statistics::default(), Vec::with_capacity(len_hint)),
+            |(mut acc, mut seg_acc), (location, segment_info)| {
+                let loc = &location.0;
+                let stats = &segment_info.summary;
+                acc.row_count += stats.row_count;
+                acc.block_count += stats.block_count;
+                acc.uncompressed_byte_size += stats.uncompressed_byte_size;
+                acc.compressed_byte_size += stats.compressed_byte_size;
+                acc.index_size = stats.index_size;
+                acc.col_stats = if acc.col_stats.is_empty() {
+                    stats.col_stats.clone()
+                } else {
+                    statistics::reduce_block_statistics(&[&acc.col_stats, &stats.col_stats])?
+                };
+                seg_acc.push(loc.clone());
+                Ok::<_, ErrorCode>((acc, seg_acc))
+            },
+        )?;
+
+        Ok((seg_locs, s))
+    }
+
     // Left a hint file which indicates the location of the latest snapshot
     async fn write_last_snapshot_hint(
         operator: &Operator,
--- a/src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
+++ b/src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
@@ -69,7 +69,7 @@ impl CompactSegmentMutator {
 
 #[async_trait::async_trait]
 impl TableMutator for CompactSegmentMutator {
-    async fn blocks_select(&mut self) -> Result<bool> {
+    async fn target_select(&mut self) -> Result<bool> {
         let fuse_segment_io = FuseSegmentIO::create(self.ctx.clone(), self.data_accessor.clone());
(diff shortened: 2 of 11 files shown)


src/query/storages/fuse/src/fuse_table.rs
```diff
@@ -454,9 +454,10 @@ impl Table for FuseTable {
     async fn compact(
         &self,
         ctx: Arc<dyn TableContext>,
+        segments_only: bool,
```
BohuTANG, line 457: Use `enum`?

dantengsky: yeah, should use enum here. 
will be addressed while merging with PR #8205

tests/logictest/suites/base/09_fuse_engine/09_0008_fuse_optimize_table
```diff
@@ -160,12 +160,37 @@ optimize table m purge;
...
+
+statement ok
+insert into t1 values (5);
+
+statement ok
+insert into t1 values (6);
+
+statement ok
+insert into t1 values (7);
+
```
BohuTANG, line 175: statement query II
select segment_count, block_count from fuse_snapshot('db_09_0008', 't1') limit 1;

dantengsky: right, thanks ;)

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
```diff
@@ -0,0 +1,384 @@
...
+
+fn detect_conflicts(base: &TableSnapshot, latest: &TableSnapshot) -> Conflict {
+    let base_segments = &base.segments;
+    let latest_segments = &latest.segments;
+
+    let base_segments_len = base_segments.len();
+    let latest_segments_len = latest_segments.len();
+
+    if latest_segments_len >= base_segments_len
+        && base_segments[base_segments_len - 1] == latest_segments[latest_segments_len - 1]
```
BohuTANG, line 364: Is`base_segments_len` will 0?

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+fn detect_conflicts(base: &TableSnapshot, latest: &TableSnapshot) -> Conflict {
+    let base_segments = &base.segments;
+    let latest_segments = &latest.segments;
+
+    let base_segments_len = base_segments.len();
+    let latest_segments_len = latest_segments.len();
+
+    if latest_segments_len >= base_segments_len
```
zhyass, line 363: Is it correct to compare only the head and tail?  If delete was executed during this process?

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
```diff
@@ -0,0 +1,384 @@
...
+    {
+        Conflict::ReconcilableAppend(0..(latest_segments_len - base_segments_len))
+    } else {
+        Conflict::Irreconcilable
+    }
+}
+
+async fn abort_segment_compaction(data_accessor: &Operator, locations: &[Location]) {
+    // parallel delete?
+    for (path, _) in locations {
```
BohuTANG, line 375: Use `remove_file_in_batch `? https://github.com/datafuselabs/databend/blob/main/src/query/storages/fuse/src/io/files.rs#L39

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
dantengsky: yeah, we should not rely on that assumption, thanks

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
dantengsky: might be :+1:

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
dantengsky: > Use `remove_file_in_batch `? https://github.com/datafuselabs/databend/blob/main/src/query/storages/fuse/src/io/files.rs#L39

that is handy

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
dantengsky: the logic of conflicts detection of mutators has been unified, in commit.rs, please take another look

src/query/storages/fuse/src/operations/mutation/compact_segment_mutator.rs
dantengsky: the logic of conflicts detection of mutators has been unified, in commit.rs, please take another look

src/query/storages/fuse/src/operations/commit.rs
```diff
@@ -452,15 +463,22 @@ impl FuseTable {
...
-                    let suffix = new_segments
-                        .split_off(latest_snapshot.segments.len() - base_snapshot.segments.len());
-                    if suffix.ne(&base_snapshot.segments) {
-                        abort_operation
-                            .abort(ctx.clone(), self.operator.clone())
-                            .await?;
-                        return Err(ErrorCode::StorageOther(
-                            "mutation conflicts, concurrent mutation detected while committing segment compaction operation",
-                        ));
+                    match MutatorConflictDetector::detect_conflicts(
```
zhyass, line 466: https://github.com/datafuselabs/databend/blob/0f4ebb46f151ae627b6b292591027534ad112f96/src/query/storages/fuse/src/operations/commit.rs#L455-L462 can be removed.

src/query/storages/fuse/src/operations/commit.rs
```diff
@@ -491,6 +509,44 @@ impl FuseTable {
...
+    ResolvableAppend(Range<usize>),
+}
+
+// wraps a namespace, to clarify the who is detecting conflict
+pub struct MutatorConflictDetector;
+
+impl MutatorConflictDetector {
+    // detects conflicts, as a mutator, working on the base snapshot, with latest snapshot
+    pub fn detect_conflicts(base: &TableSnapshot, latest: &TableSnapshot) -> Conflict {
+        if base.snapshot_id == latest.snapshot_id {
```
zhyass, line 525: There is no condition that two snapshots have the same snapshot_id, maybe it can be deleted?

dantengsky: right, it is too defensive

zhyass (approved):
LGTM

BohuTANG (approved):

PR #8280, opened 2022-10-18
feat(query): modify function-v2 comparison like

### LINKED ISSUE #8278: feat: Need modify function-v2 like
Need modify function-v2 like

_Originally posted by @TCeason in https://github.com/datafuselabs/databend/issues/8220#issuecomment-1281805762_

TCeason:

## Summary

The find cost 840.270221ms is faster than the contains cost 1.487372717s. Use find.

```rust
➜  ~/database/source-code/hello👆 git:(master) ⚡ cat src/main.rs
use std::time::Instant;
fn main() {
    let needle = (0..100).map(|_| "b").collect::<String>();
    let heystack = (0..100_000).map(|_| "a").collect::<String>();

    println!("Data ready.");

    let now = Instant::now();
    for _ in 0..1_000_000 {
        if heystack.find(&needle).is_some() { }
    }
    let ned = Instant::now();
    println!("cost:{:?}", ned-now);
}
➜  ~/database/source-code/hello👆 git:(master) ⚡ cargo run -r          
    Finished release [optimized] target(s) in 0.00s
     Running `target/release/hello`
Data ready.
cost:840.270221ms


➜  ~/database/source-code/hello👆 git:(master) ⚡ cat src/main.rs
use std::time::Instant;
fn main() {
    let needle = (0..100).map(|_| "b").collect::<String>();
    let heystack = (0..100_000).map(|_| "a").collect::<String>();

    println!("Data ready.");

    let now = Instant::now();
    for _ in 0..1_000_000 {
        if heystack.contains(&needle) { }
    }
    let ned = Instant::now();
    println!("cost:{:?}", ned-now);
}
➜  ~/database/source-code/hello👆 git:(master) ⚡ cargo run -r  
   Compiling hello v0.1.0 (/home/taichong/database/source-code/hello)
    Finished release [optimized] target(s) in 0.34s
     Running `target/release/hello`
Data ready.
cost:1.487372717s
```

Fixed #8278

### DIFF
--- a/src/query/functions-v2/src/scalars/comparison.rs
+++ b/src/query/functions-v2/src/scalars/comparison.rs
@@ -235,18 +235,22 @@ fn register_like(registry: &mut FunctionRegistry) {
         "like",
         FunctionProperty::default(),
         |_, _| None,
-        vectorize_regexp(|str, pat, map, _| {
-            let mut sub_strings: Vec<&str> = vec![];
+        vectorize_regexp(|str, pat, map, string_map, _| {
             let pattern = if let Some(pattern) = map.get(pat) {
                 pattern
             } else {
                 let pattern_str = simdutf8::basic::from_utf8(pat).map_err(|err| {
                     format!("unable to convert the LIKE pattern to string: {err}")
                 })?;
-                sub_strings = pattern_str
+
+                let mut sub_strings: Vec<&str> = pattern_str
                     .split(|c: char| c == '%' || c == '_' || c == '\\')
                     .collect();
                 sub_strings.retain(|&substring| !substring.is_empty());
+                if !sub_strings.is_empty() {
+                    string_map.insert(pat.to_vec(), sub_strings[0].to_string());
+                }
+
                 let re_pattern = like_pattern_to_regex(pattern_str);
                 let re = Regex::new(&re_pattern)
                     .map_err(|err| format!("unable to build the LIKE pattern: {err}"))?;
@@ -254,11 +258,10 @@ fn register_like(registry: &mut FunctionRegistry) {
                 map.get(pat).unwrap()
             };
 
-            if !sub_strings.is_empty() {
+            if string_map.get(pat).is_some() {
                 let lhs_str =
                     std::str::from_utf8(str).expect("Unable to convert lhs value to string: {}");
-                let required_string = sub_strings[0];
-                let contain = lhs_str.find(required_string);
+                let contain = lhs_str.find(string_map.get(pat).unwrap().as_str());
                 if contain.is_none() {
                     Ok(false)
                 } else {
@@ -274,7 +277,7 @@ fn register_like(registry: &mut FunctionRegistry) {
         "regexp",
         FunctionProperty::default(),
         |_, _| None,
-        vectorize_regexp(|str, pat, map, _| {
+        vectorize_regexp(|str, pat, map, _, _| {
             let pattern = if let Some(pattern) = map.get(pat) {
                 pattern
             } else {
@@ -288,7 +291,13 @@ fn register_like(registry: &mut FunctionRegistry) {
 }
 
 fn vectorize_regexp(
-    func: impl Fn(&[u8], &[u8], &mut HashMap<Vec<u8>, Regex>, FunctionContext) -> Result<bool, String>
+    func: impl Fn(
+        &[u8],
+        &[u8],
+        &mut HashMap<Vec<u8>, Regex>,
+        &mut HashMap<Vec<u8>, String>,
+        FunctionContext,
+    ) -> Result<bool, String>
     + Copy,
 ) -> impl Fn(
     ValueRef<StringType>,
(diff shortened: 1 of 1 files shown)


TCeason:
In local pc, release build. 10G tpch data result:

```sql
Database changed
'root'@mysqldb 15:59:40 [tpch]>  select count(*) from orders where o_comment not like '%pending%deposits%';
+----------+
| count(*) |
+----------+
| 14838130 |
+----------+
1 row in set (1.17 sec)
Read 15000000 rows, 920.67 MiB in 1.163 sec., 12.9 million rows/sec., 791.68 MiB/sec.

```

TCeason:
I do a test with rust regex 1.6.0 and cpp re2.

rust regex cost 2.5ms, cpp re2 cost 17ms.

```rust
➜  ~/database/source-code/hello👆 git:(master) ⚡ cat src/main.rs                  
use std::time::Instant;
use regex::Regex;
fn main() {
    let now = Instant::now();
    let re = Regex::new(".*hello.*").unwrap();
    for _ in 0..100000 {
      if re.is_match("123 hello:1234") {
      }
    }
    let end = Instant::now();
    println!("cost: {:?}", end - now);
}
➜  ~/database/source-code/hello👆 git:(master) ⚡ cargo run -r   
    Finished release [optimized] target(s) in 0.00s
     Running `target/release/hello`
cost: 2.536025ms

```


```cpp
➜  ~/database/source-code/a👆 cat a.cpp 
#include <iostream>
#include <string>
#include <re2/re2.h>
#include<sys/timeb.h>

using namespace std;

long long systemtime()
{
}

int main(int argc, char **argv){
  RE2 regex(".*hello.*");
  string text = "123 hello:1234";
  re2::StringPiece sp(text);
  re2::StringPiece submatch;
  timeb t;
  ftime(&t);

  long long start = t.time*1000+t.millitm;
  for (int i=0; i < 100000; i++) {
    if (regex.Match(sp, 0, text.size(), RE2::UNANCHORED, &submatch, 1)) {
    }
  } 
  timeb e;
  ftime(&e);
  long long end = e.time*1000+e.millitm;

  cout << "cost :" << end - start<< "ms" << endl;
  return 0;
}

➜  ~/database/source-code/a👆 g++  a.cpp -lre2 && ./a.out 
cost :17ms


```

src/query/functions-v2/src/scalars/comparison.rs
```diff
@@ -236,19 +236,37 @@ fn register_like(registry: &mut FunctionRegistry) {
         FunctionProperty::default(),
         |_, _| None,
         vectorize_regexp(|str, pat, map, _| {
+            let mut sub_strings: Vec<&str> = vec![];
             let pattern = if let Some(pattern) = map.get(pat) {
                 pattern
             } else {
                 let pattern_str = simdutf8::basic::from_utf8(pat).map_err(|err| {
```
sundy-li, line 243: These codes are inside the whole loop, need to have faster path.

src/query/functions-v2/src/scalars/comparison.rs
```diff
@@ -236,19 +236,37 @@ fn register_like(registry: &mut FunctionRegistry) {
...
                     format!("unable to convert the LIKE pattern to string: {err}")
                 })?;
+                sub_strings = pattern_str
+                    .split(|c: char| c == '%' || c == '_' || c == '\\')
+                    .collect();
+                sub_strings.retain(|&substring| !substring.is_empty());
                 let re_pattern = like_pattern_to_regex(pattern_str);
                 let re = Regex::new(&re_pattern)
```
Xuanwo, line 251: regex's time complex is `O(n)`; every char in the str will only be checked once. 

The real problem here is that we need to move `Regex::new(&re_pattern)` out of the loop.

Xuanwo: Oh, we already cached the pattern. Please ignore.

sundy-li: but  the  regexp is cached, the sub_strings will be always empty in the iteration.

src/query/functions-v2/src/scalars/comparison.rs
TCeason: > These codes are inside the whole loop, need to have faster path.

Got

sundy-li (approved):

src/query/functions-v2/src/scalars/comparison.rs
```diff
@@ -270,7 +291,13 @@ fn register_like(registry: &mut FunctionRegistry) {
 }
 
 fn vectorize_regexp(
-    func: impl Fn(&[u8], &[u8], &mut HashMap<Vec<u8>, Regex>, FunctionContext) -> Result<bool, String>
+    func: impl Fn(
+        &[u8],
+        &[u8],
+        &mut HashMap<Vec<u8>, Regex>,
+        &mut HashMap<Vec<u8>, String>,
+        FunctionContext,
```
BohuTANG, line 299: In general, the `FunctionContext` parameter is put first. We can. refine them all later in another PR.

BohuTANG (approved):

src/query/functions-v2/src/scalars/comparison.rs
TCeason: ok

src/query/functions-v2/src/scalars/comparison.rs
```diff
@@ -235,28 +235,49 @@ fn register_like(registry: &mut FunctionRegistry) {
...
+
                 let re_pattern = like_pattern_to_regex(pattern_str);
                 let re = Regex::new(&re_pattern)
                     .map_err(|err| format!("unable to build the LIKE pattern: {err}"))?;
                 map.insert(pat.to_vec(), re);
                 map.get(pat).unwrap()
             };
-            Ok(pattern.is_match(str))
+
+            if string_map.get(pat).is_some() {
```
andylokandy, line 261: using if let Some(...) instead of unwrap eliminates the unwrap.

PR #8294, opened 2022-10-18
fix: failed to release the permits on time in the block pruner

dantengsky:

## Summary

failed to release the permits on time in the block pruner, which may lead to OOM, even if `max_storage_io_requests` is set to a small number,  if the number of blocks is large

Fixes #issue


BohuTANG (approved):

dantengsky:
the issue is actually resolved in commit 

https://github.com/datafuselabs/databend/pull/8261/commits/83d0a7bf9d5d3ff11cbf019eaa3c1bc03614f997

PR #8298, opened 2022-10-19
refactor(meta/service): try best to leave a cluster

drmingdrmer:

##### refactor(meta/service): try best to leave a cluster

When leaving a node from a meta-service cluster, no matter the node is a
voter and non-voter(learner), it should try the best removing it with
the following step:

- Remove it from membership list.
- Stop replication to it.
- Remove its node info record from meta store.

Before this commit, if the node is not in membership list it won't do
anything, hense there is no way to clean up a dirty state, e.g., to
remove a non-voter(learner).

## Related Issues

### DIFF
--- a/src/meta/service/src/meta_service/meta_leader.rs
+++ b/src/meta/service/src/meta_service/meta_leader.rs
@@ -170,21 +170,18 @@ impl<'a> MetaLeader<'a> {
         // 2. Stop replication
         let res = self.meta_node.raft.remove_learner(node_id).await;
         if let Err(e) = res {
-            let res = match e {
+            match e {
                 RemoveLearnerError::ForwardToLeader(e) => {
-                    Err(RaftChangeMembershipError::ForwardToLeader(e))
+                    return Err(RaftChangeMembershipError::ForwardToLeader(e));
                 }
                 RemoveLearnerError::NotLearner(_e) => {
                     error!("Node to leave the cluster is not a learner: {}", node_id);
-                    Ok(())
                 }
                 RemoveLearnerError::NotExists(_e) => {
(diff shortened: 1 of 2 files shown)


everpcpc (approved):
❤️

src/meta/service/src/meta_service/meta_leader.rs
```diff
@@ -153,33 +154,23 @@ impl<'a> MetaLeader<'a> {
...
 
-        // When write(RemoveNode{}) returns, the replication to that node should be dropped.
-        // But there is chance `add_configured_non_voters()` adds it back because it runs in another task.
-        // Thus we need to ensure the replication is removed here.
+        // 2. Stop replication
         let res = self.meta_node.raft.remove_learner(node_id).await;
-
         if let Err(e) = res {
-            return match e {
+            let res = match e {
```
Xuanwo, line 173: Confusing...

How about return error directly?

Xuanwo:
```rust
        if let Err(e) = res {
            match e {
                RemoveLearnerError::ForwardToLeader(e) => {
                    return Err(RaftChangeMembershipError::ForwardToLeader(e))
                }
                RemoveLearnerError::NotLearner(_e) => {
                    error!("Node to leave the cluster is not a learner: {}", node_id);
                }
                RemoveLearnerError::NotExists(_e) => {
                    info!("Node to leave the cluster does not exists: {}", node_id);
                }
                RemoveLearnerError::Fatal(e) => return Err(RaftChangeMembershipError::Fatal(e)),
            };
        }

Xuanwo (approved):

PR #8308, opened 2022-10-19
refactor(meta/service): merge two to-meta-server rpc into one

drmingdrmer:

##### refactor(meta/service): merge two to-meta-server rpc into one

`MetaClient::do_read()` and `MetaClient::do_write()` does almost the
same things, thus merge them into one rpc `kv_api()`.

This is a breaking change: With this commit databend-query requires at least(inclusive) the 0.8.35 meta-server:
https://github.com/datafuselabs/databend/releases/tag/v0.8.35-nightly

## Related Issues


src/meta/client/src/grpc_client.rs
```diff
@@ -780,81 +779,31 @@ impl MetaGrpcClient {
...
-        let raft_reply = result?;
-
-        let res: Result<R, KVAppError> = raft_reply.into();
-
-        res
-    }
-
-    #[tracing::instrument(level = "debug", skip(self, v))]
-    pub(crate) async fn do_read<T, R>(&self, v: T) -> Result<R, KVAppError>
+    pub(crate) async fn kv_api<T, R>(&self, v: T) -> Result<R, KVAppError>
```
Xuanwo, line 782: `kv_api` is more like a trait or struct, not a function name. How about `send` or `request` (like in HTTP client)?

BohuTANG (approved):

src/meta/client/src/grpc_client.rs
drmingdrmer: Well... there are several RPC in this struct: 
- KV oriented: get, list, upsert, mget
- transactional API: `txn(condition1, condition2..., action1, action2...)`
- admin API, such as that returning a member list.
- ...

`send()` may not be specific enough to tell what it does, IMHO.

Xuanwo (approved):

PR #8317, opened 2022-10-19
feat(query): improve performance of count distinct.

RinChanNOWWW:

## Summary

Try to improve performance of count distinct. (Mainly for string type).

Modifications:

1. Fix the name of count function to avoid `build_column`.
2. Add a specific distinct state for string type, whose hash set's hash key is uint128 (instead of string).

For the 2nd one:

Use siphash128 to hash a string into a uint128 number, and use the uint128 value as a hash key of the hash set.

Collision of siphash128 of strings is almost impossible. (Some discussion: https://togithub.com/ClickHouse/ClickHouse/issues/9506)

### Experiments under release build

Pre:

```sql
create table t1 (a int);
insert into t select * from numbers(10000000);
create table t2 (a int); -- fuse engine
insert into t2 select to_string(number) from numbers(10000000);
```

Old version (commit: 1c506fda072e3418c0eef77f0f270c8ec25ab9f9):

| SQL | Time |
| -- | -- |
| `select count(distinct number) from numbers(10000000);` | about 1.5s |
| `select count(distinct to_string(number)) from numbers(10000000);` | about 3s |
| `select count(distinct a) from t1` | about 1.6s |
| `select count(distinct a) from t2` | about 3s |

New version (this PR):

| SQL | Time |
| -- | -- |
| `select count(distinct number) from numbers(10000000);` | about 1.3s |
| `select count(distinct to_string(number)) from numbers(10000000);` | about 1.8s |
| `select count(distinct a) from t1` | about 1.5s |
| `select count(distinct a) from t2` | about 2s |

- result of the 1st modification: about 10% time reduction for all types.
- result of the 2nd modification: about 30%~40% time reduction for string type.


sundy-li (approved):

Xuanwo:
@mergify update

PR #8322, opened 2022-10-19
fix: add a threshold for compact block

### LINKED ISSUE #8241: feat: optimize table xx compact limit 1000
**Summary**

If a table is larger, such as 100TB, `optimize table xx compact` will take a long time, and if an error happens, the full compaction will be failed. We can add a `limit` to batch compaction.
```
optimize table xx compact limit 1000;
```

Random to pick 1000 segments which need to compact.

### LINKED ISSUE #8316: bug:  optimize table tb_name compact let the data grow biger
### Search before asking

### Version

latest version

### What's Wrong?

There is nothing written for this table.
Every time execute: optimize table tb_name compact;  make data increases.

image

### How to Reproduce?

_No response_

zhyass:

## Summary

Add a threshold 2 for compact blocks, if the selected blocks are less than 2*max_threads, compact is skipped.

 Add a limit to batch compaction.
```
optimize table [database.]table_name compact [segment] [limit segment_count];
```

Fixes #8316 
Fixes #8241

### DIFF
--- a/src/query/storages/fuse/src/operations/compact.rs
+++ b/src/query/storages/fuse/src/operations/compact.rs
@@ -115,11 +115,6 @@ impl FuseTable {
             return Ok(None);
         }
 
-        let max_threads = ctx.get_settings().get_max_threads()? as usize;
-        if mutator.selected_blocks().len() < max_threads * 2 {
-            return Ok(None);
-        }
-
         let partitions_total = mutator.partitions_total();
         let (statistics, parts) = Self::read_partitions_with_metas(
             ctx.clone(),
--- a/src/query/storages/fuse/src/operations/compact.rs
+++ b/src/query/storages/fuse/src/operations/compact.rs
@@ -33,9 +33,10 @@ use crate::TableMutator;
 use crate::DEFAULT_BLOCK_PER_SEGMENT;
 use crate::FUSE_OPT_KEY_BLOCK_PER_SEGMENT;
 
-struct CompactOptions {
-    base_snapshot: Arc<TableSnapshot>,
-    block_per_seg: usize,
+pub struct CompactOptions {
+    pub base_snapshot: Arc<TableSnapshot>,
+    pub block_per_seg: usize,
+    pub limit: Option<usize>,
 }
 
 impl FuseTable {
@@ -43,6 +44,7 @@ impl FuseTable {
         &self,
         ctx: Arc<dyn TableContext>,
         target: CompactTarget,
+        limit: Option<usize>,
         pipeline: &mut Pipeline,
     ) -> Result<Option<Box<dyn TableMutator>>> {
         let snapshot_opt = self.read_table_snapshot(ctx.clone()).await?;
@@ -63,6 +65,7 @@ impl FuseTable {
         let compact_params = CompactOptions {
             base_snapshot,
             block_per_seg,
+            limit,
         };
 
         match target {
@@ -103,10 +106,9 @@ impl FuseTable {
         let block_per_seg = options.block_per_seg;
         let mut mutator = FullCompactMutator::try_create(
             ctx.clone(),
-            options.base_snapshot,
+            options,
             block_compactor.clone(),
             self.meta_location_generator().clone(),
-            block_per_seg,
             self.cluster_key_meta.is_some(),
             self.operator.clone(),
         )?;
--- a/src/query/storages/fuse/src/operations/compact.rs
+++ b/src/query/storages/fuse/src/operations/compact.rs
@@ -115,6 +115,11 @@ impl FuseTable {
             return Ok(None);
         }
 
+        let max_threads = ctx.get_settings().get_max_threads()? as usize;
+        if mutator.selected_blocks().len() < max_threads * 2 {
+            return Ok(None);
+        }
+
         let partitions_total = mutator.partitions_total();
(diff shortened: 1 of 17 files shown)


wubx (approved):

src/query/storages/fuse/src/operations/compact.rs
```diff
@@ -117,6 +117,11 @@ impl FuseTable {
             return Ok(None);
         }
 
+        let max_threads = ctx.get_settings().get_max_threads()? as usize;
```
BohuTANG, line 120: Please add some comments.

BohuTANG:
Some sqllogic tests related to compact still need to modify.

tests/logictest/suites/base/09_fuse_engine/09_0008_fuse_optimize_table
```diff
@@ -27,6 +27,9 @@ select * from t order by a;
 6
 7
 
+statement ok
+set max_threads=1;
```
BohuTANG, line 31: To check this PR works, we should add a new test for it with a big max_threads.

BohuTANG: We can add a test with:
set max_threads=100;

src/query/storages/fuse/src/operations/mutation/compact_mutator/full_compact_mutator.rs
```diff
@@ -96,17 +93,32 @@ impl FullCompactMutator {
...
-        for (idx, segment) in segments.iter().enumerate() {
+        let segments = segments_io
+            .read_segments(segment_locations)
+            .await?
+            .into_iter()
+            .collect::<Result<Vec<_>>>()?;
+
+        let limit = self.compact_params.limit.unwrap_or(segments.len());
+        if limit < segments.len() {
+            for i in limit..segments.len() {
```
BohuTANG, line 111: We should choose the limit segment from the oldest.
It seems `SegmentInfo` should add timestamp and order by the time desc.
Reference:
https://github.com/datafuselabs/databend/blob/dc5e25186f1479dc7560e1eac7d5dfcd78670424/src/query/storages/fuse/src/io/snapshots.rs#L221-L225

zhyass: I don't think we need to add timestamp.
The current processing logic is that the segments generated by insert are placed in front and the segments generated by compact are placed after.
For example
Table t has 3 segments.
```
s1 s2 s3
```
Now execute `optimize table t compact segment limit 2 `. Compact will select s1 and s2, and assume that compact produces a new segment s4. Then s4 will be appended to the back of the segments: 
```
s3 s4
```
If the insert is executed afterwards and a new segment (s5) is generated. The s5 will be appended to the front of the segments:
```
s5 s3 s4
```
If we then execute `optimize table t compact segment limit 2`, s5 and s3 will be selected. 
But if we select by timestamp, s3 and s4 will be selected.

src/query/storages/fuse/src/operations/mutation/compact_mutator/full_compact_mutator.rs
```diff
@@ -96,17 +93,32 @@ impl FullCompactMutator {
...
-        let mut summarys = Vec::new();
 
+        // Blocks that need to be reorganized into new segments.
+        let mut remain_blocks = Vec::new();
         // Read all segments information in parallel.
         let segments_io = SegmentsIO::create(self.ctx.clone(), self.data_accessor.clone());
-        let segments = segments_io.read_segments(segment_locations).await?;
-        for (idx, segment) in segments.iter().enumerate() {
+        let segments = segments_io
+            .read_segments(segment_locations)
```
BohuTANG, line 104: If we have the limit, is there no need to read all segments?

src/query/storages/fuse/src/operations/mutation/compact_mutator/full_compact_mutator.rs
zhyass: > We should choose the limit segment from the oldest. It seems `SegmentInfo` should add timestamp and order by the time desc. Reference:
> 


I don't think we need to add timestamp.
The current processing logic is that the segments generated by insert are placed in front and the segments generated by compact are placed after.
For example
Table t has 3 segments.
```
s1 s2 s3
```
Now execute `optimize table t compact segment limit 2 `. Compact will select s1 and s2, and assume that compact produces a new segment s4. Then s4 will be appended to the back of the segments: 
```
s3 s4
```
If the insert is executed afterwards and a new segment (s5) is generated. The s5 will be appended to the front of the segments:
```
s5 s3 s4
```
If we then execute `optimize table t compact segment limit 2`, s5 and s3 will be selected. 
But if we select by timestamp, s3 and s4 will be selected.

src/query/storages/fuse/src/operations/mutation/compact_mutator/full_compact_mutator.rs
zhyass: > If we have the limit, is there no need to read all segments?

No, because it needs to calculate snapshot.summary.

BohuTANG:
@soyeric128 for the new option:
optimize table <table> compact [segment] [limit <segment count>];

Issue #7931, opened 2022-09-28
Feature: support Array functions


sundy-li:
**Summary**

Description for this feature.

## common functions for array   @TCeason 

- [x] #9803
- [x] length
- [x] array_contains(arr, elem)  alias `array_has`
- [x] #9839
- [x] array_slice(array, offset[, length])
- [x] #9940
- [x] array_reverse_sort(array) // array_sort now contains this feature.

## aggr combinators  @b41sh 

- [x] array_unique(array)   Counts the unique elements of a list.
- [x] array_distinct(array) Removes all duplicates and NULLs from a list.
- [x] array_aggr(array, 'agg'), aka `array_<aggr>(array)`,  eg: `array_aggr([1,2,3], 'sum')`, `array_aggr([1,2,3], 'avg')`, `array_sum([1,2,3]`
- [x] array_prepend(element, array)
- [x] array_append(array, element)
- [x] array_any_value(array)

## low-priority tasks:

- [x] array_transform(array, lambda)    (low-priority), eg: `array_transform([1,2,3], x -> x + 1)`
- [x] array_filter(array, lambda)    (low-priority), eg: `array_filter([1,2,3], x -> x > 1)`
- [ ] python style array comprehension (low-priority), eg: `select [x+3 for x in [1,2,3]]`
- [ ] array_join(array)
- ...

BohuTANG:
Comments by @sundy-li :
1. https://duckdb.org/docs/sql/functions/nested
2. After the new expression is done.(Middle of November)

BohuTANG:
@sundy-li 

There are two array-related aggregate functions are required from the user:
- [ ] array_reduce https://clickhouse.com/docs/en/sql-reference/functions/array-functions/#arrayreduce
- [ ] group_array_moving_avg https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/grouparraymovingavg/

ClickHouse SQL like:
```
select UserID,  c, arrayReduce('count',RequestNum1) from (  
  select UserID, count(RequestNum) as c, groupArrayMovingAvg(50)(RequestNum) as RequestNum1 from
    (select UserID, RequestNum from hits_v1 order by RequestNum DESC)  
  group by UserID order by c DESC limit 1000)
```

Or are these two functions easy to use window functions?

sundy-li:
we have `array_<aggr>` which covert `arrayReduce` function.

`group_array_moving_avg ` can be implemented after we have window aggr function.

PR #8326, opened 2022-10-19
feat(query): implement array function `slice`, `remove_fist`, `remove_last`

b41sh:

## Summary

implement array function `array_slice`, `array_pop_front`, `array_pop_back`
fix `StringColumnBuilder` append another column with the first offset is not zero.

part of #7931


TCeason:
So quick 🥳

src/query/functions-v2/src/scalars/array.rs
```diff
@@ -124,4 +127,96 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                    let arr_slice = arr.slice(range);
+                    output.push(arr_slice);
+                }
+                Ok(())
+            }
+        ),
+    );
+
+    registry.register_passthrough_nullable_1_arg::<EmptyArrayType, EmptyArrayType, _, _>(
+        "array_pop_front",
```
andylokandy, line 168: the new expression framework allows it be named as pop_front (without the array prefix). This could make it more concise but also different from snowflake.

Xuanwo: > the new expression framework allows it be named as pop_front (without the array prefix).

Does this means we can apply `pop_front`, `pop_back` operations on different data types?

andylokandy: besides, its behavior seems more properly be remove_first.

TCeason: I think we are the same as duckdb.

TCeason (approved):

src/query/functions-v2/src/scalars/array.rs
andylokandy: the 'pop' brings the confusion of whether the item or the array is returned.

b41sh: rename as `remove_first` and `remove_last`

src/query/expression/src/types/string.rs
```diff
@@ -284,10 +284,24 @@ impl StringColumnBuilder {
     }
 
     pub fn append(&mut self, other: &Self) {
-        self.data.extend_from_slice(&other.data);
+        // the first offset of other column may not be zero
```
andylokandy, line 287:
```suggestion
    pub fn append(&mut self, other: &Self) {
        // the first offset of other column may not be zero
        let start = self.offsets.last().cloned().unwrap() - *other.offsets.first().unwrap();
        self.data.extend_from_slice(&other.data);
        self.offsets
            .extend(other.offsets.iter().skip(1).map(|offset| start + offset));
    }
```

andylokandy (approved):

src/query/expression/src/types/string.rs
b41sh: `other.data` needs to slice by other start.

b41sh: `start - other_start` may overflow

PR #8328, opened 2022-10-19
fix: fix copy Version mismatch bug, optimize upsert table copied file info

lichuang:

## Summary

fix: fix copy Version mismatch bug, optimize upsert table copied file info

Fixes #issue


lichuang:
There is a bug issue (https://github.com/datafuselabs/databend/issues/8356 ) of this pr, so now close this pr.

PR #8334, opened 2022-10-20
feat: try to get opeartor metrics with DalContext

### LINKED ISSUE #8331: bug: fuse table lost DalContext layer bind
**Summary**

After refactoring a few days ago, now the fuse table storage operator init:
https://github.com/datafuselabs/databend/blob/c0030eb977b4e2fee50d464c749f240ed044ad0f/src/query/storages/fuse/src/fuse_table.rs#L88-L113

We can not get the metrics from DalContext.

By design, it should be:
https://github.com/datafuselabs/databend/blob/c0030eb977b4e2fee50d464c749f240ed044ad0f/src/query/service/src/sessions/query_ctx.rs#L311-L319

After refactoring, putting dal_ctx  in the query context is not right, it looks every operator(DataOperator/CacheOpeartor) should have his metrics and get them from the Query Context.

BohuTANG:

## Summary

* Add DalContext to catalog

Fixes #8331


BohuTANG:
Too complex and tedious, let's close.

PR #8361, opened 2022-10-20
fix(processor): fix throw error when abort and limit

### LINKED ISSUE #8352: Bug: full outer join abort
**Summary**

Run sql in https://github.com/datafuselabs/databend/pull/8351/files#diff-84732f671e927431feb54f246b093cd139f5dfc4e3aec92bbc592dbb9b535b85

Got `ERROR 1105 (HY000): Code: 1043, displayText = Aborted query, because the server is shutting down or the query was killed..`

Expected:
https://github.com/datafuselabs/databend/pull/8351/files#diff-9ea25e7f2ca64bd84f2e6bbca384b5d4a0dd63f22b08082db607791758851f16

zhang2014:

## Summary

fix(processor): fix throw error when abort and limit

fixes #8352

### DIFF
--- a/src/query/service/src/pipelines/executor/pipeline_pulling_executor.rs
+++ b/src/query/service/src/pipelines/executor/pipeline_pulling_executor.rs
@@ -24,6 +24,8 @@ use common_base::base::Thread;
 use common_datablocks::DataBlock;
 use common_exception::ErrorCode;
 use common_exception::Result;
+use common_exception::ABORT_QUERY;
+use common_exception::ABORT_SESSION;
 use parking_lot::Condvar;
 use parking_lot::Mutex;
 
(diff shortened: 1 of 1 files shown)


src/query/service/src/pipelines/executor/pipeline_pulling_executor.rs
```diff
@@ -184,14 +186,25 @@ impl PipelinePullingExecutor {
...
                         self.executor.finish(None);
                     }
 
                     self.state.wait_finish();
 
                     if self.state.is_catch_error() {
-                        return Err(self.state.get_catch_error());
+                        let error_code = self.state.get_catch_error();
+
+                        // abort errors are caused by interrupts and should be ignored.
```
xudong963, line 201: I don't understand the comment.

1. By interrupting and ? should be ignored.
2. Why was the SQL in issue aborted?  (Server wasn't shot down and the query wasn't killed)

zhang2014: for 'select * from xx limit 20', when 20 rows of data are returned, the query should have been finished. since the processor cannot timely known that, we try to kill the query in here.

zhang2014: Updated comment.

xudong963 (approved):

Issue #8345, opened 2022-10-20
Feature: Support multiple table progress


Xuanwo:
Address https://github.com/datafuselabs/databend/issues/8331

## Summary

We are storing `dal_metrics_read_bytes`, `dal_metrics_write_bytes` in `dal_metrics_write_bytes` which is shared across the whole query. This doesn't make sense now because we could have multiple tables in the same query. Users should be able to know the metrics about different tables.

Thus, we should support multiple table progress reports.

## Plan

- rename `dal_metrics_read_bytes` to `table_data_read_bytes`
- rename `dal_metrics_write_bytes` to `table_data_write_bytes`
- Add support for `table_cache_{read/write}_bytes`.
- Init the table data metrics while creating table instances instead of query start.

Other progress like `scan_progress_read_rows` and `scan_progress_read_bytes` should also be updated.

## Tasks

- [ ] Implement single table support
- [ ] Implement multiple table support

Xuanwo:
> Can we get all-in-one metrics for a query? Like:

We can get all tables attached to this query and then do a simple calculation.

Xuanwo:
We will start from reusing current logic first.

PR #8363, opened 2022-10-20
feat: Introduce data metrics for table

Xuanwo:

## Summary

- Fixes https://github.com/datafuselabs/databend/issues/8345
- Fixes https://github.com/datafuselabs/databend/issues/8331

This PR introduces a new idea called `StorageMetrics`. The same as our operator, we could have:

- data metrics: Read/Write metrics for data
- cache metrics: Read/Write metrics for cache

As in this PR, only data metrics is added.

Also, this PR decouples metrics from our query context. Since this PR, `Table` trait will provide a new API called `get_data_metrics`. By accessing this API, users can read the data metrics about this table.

And in `show processlist`, we will fetch all attached table in this query, fetch their metrics and calculate to get the overall metrics.


BohuTANG (approved):
Looks great to me 🚀

PR #8372, opened 2022-10-20
feat(function-v2): add tuple() and get() for tuple

andylokandy:

## Summary

newly supported syntaxes:

```sql
SELECT (1, 2, col_a, 'str');
SELECT (1, (2, 3));
SELECT ('a', 'b').1;  -- 'a'
```

related to https://github.com/datafuselabs/databend/issues/6547


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1805,10 +1817,14 @@ impl<'a> TypeChecker<'a> {
...
             let accessor = accessores.pop().unwrap();
             let accessor_lit = match accessor {
-                MapAccessor::Bracket { key } => key.clone(),
+                MapAccessor::Bracket {
+                    key: box Expr::Literal { lit, .. },
+                } => lit,
                 MapAccessor::Period { key } | MapAccessor::Colon { key } => {
                     Literal::String(key.name.clone())
                 }
+                MapAccessor::PeriodNumber { key } => Literal::Integer(key - 1),
```
andylokandy, line 1826: The inner name of the tuple field is starting from 0, while the index of `<tuple>.<index>` in the surface SQL is starting from 1. Should we change the inner name for consistency?

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1823,21 +1823,26 @@ impl<'a> TypeChecker<'a> {
...
+                    if idx as usize > inner_types.len() {
                         return Err(ErrorCode::SemanticError(format!(
                             "tuple index {} is out of bounds for length {}",
                             idx,
                             inner_types.len()
                         )));
                     }
-                    let inner_name = inner_names.get(idx as usize).unwrap();
-                    let inner_type = inner_types.get(idx as usize).unwrap();
+                    let inner_name = inner_names.get(idx as usize - 1).unwrap();
```
andylokandy, line 1844: @b41sh The inner name of the tuple field is starting from 0, while the index of `<tuple>.<index>` in the surface SQL is starting from 1. Should we change the inner name for consistency?

b41sh: yes, we can change `inner_name` start from 1 to keep consistency.

src/query/functions-v2/src/scalars/tuple.rs
```diff
@@ -0,0 +1,158 @@
...
+use common_expression::FunctionProperty;
+use common_expression::FunctionRegistry;
+use common_expression::FunctionSignature;
+use common_expression::Scalar;
+use common_expression::ScalarRef;
+use common_expression::Value;
+use common_expression::ValueRef;
+
+pub fn register(registry: &mut FunctionRegistry) {
+    registry.register_function_factory("tuple", |_, args_type| {
```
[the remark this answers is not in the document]
andylokandy: have discussed with @sundy-li before, we'd like support named tuple only in the schema level, which means, only the tuple defined in the schema can be named.

src/query/ast/src/ast/expr.rs
```diff
@@ -233,9 +233,11 @@ pub enum Literal {
 #[derive(Debug, Clone, PartialEq)]
 pub enum MapAccessor<'a> {
     /// `[0][1]`
-    Bracket { key: Literal },
+    Bracket { key: Box<Expr<'a>> },
     /// `.a.b`
     Period { key: Identifier<'a> },
+    /// `.1`
+    PeriodNumber { key: u64 },
```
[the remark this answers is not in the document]
andylokandy: this is only used for tuple so far... because it has different sematic to Braket.

.<number> must be a const, while [<expr>] can be any arbitrary expression including column ref.

src/query/sql/src/planner/semantic/type_check.rs
andylokandy: Will it break any storage data?

b41sh: We have not introduced the `tuple` structure yet, so we can consider that no one has used it in production.

src/query/ast/src/ast/expr.rs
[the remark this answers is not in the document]
b41sh: ok

sundy-li (approved):

b41sh (approved):

andylokandy:
@mergify update

PR #8370, opened 2022-10-20
fix(binder): lose non-equi condition for join

xudong963:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Before
```
mysql> explain select   c_custkey  from   customer  left outer join  orders  on c_custkey + o_custkey = o_custkey;
+--------------------------------------------------------+
| explain                                                |
+--------------------------------------------------------+
| HashJoin                                               |
| ├── join type: RIGHT OUTER                             |
| ├── build keys: []                                     |
| ├── probe keys: []                                     |
| ├── filters: []                                        |
| ├── TableScan(Build)                                   |
| │   ├── table: default.tpch.customer                   |
| │   ├── read rows: 1500000                             |
| │   ├── read bytes: 135769198                          |
| │   ├── partitions total: 10                           |
| │   ├── partitions scanned: 10                         |
| │   └── push downs: [filters: [], limit: NONE]         |
| └── TableScan(Probe)                                   |
|     ├── table: default.tpch.orders                     |
|     ├── read rows: 15000000                            |
|     ├── read bytes: 711675022                          |
|     ├── partitions total: 10                           |
|     ├── partitions scanned: 10                         |
|     └── push downs: [filters: [], limit: NONE]         |
+--------------------------------------------------------+
```
After
```
mysql> explain select   c_custkey  from   customer  left outer join  orders  on c_custkey + o_custkey = o_custkey;
+--------------------------------------------------------------------------------------------------+
| explain                                                                                          |
+--------------------------------------------------------------------------------------------------+
| HashJoin                                                                                         |
| ├── join type: RIGHT OUTER                                                                       |
| ├── build keys: []                                                                               |
| ├── probe keys: []                                                                               |
| ├── filters: [=(+(customer.c_custkey (#0), orders.o_custkey (#9)), orders.o_custkey (#9))]       |
| ├── TableScan(Build)                                                                             |
| │   ├── table: default.tpch.customer                                                             |
| │   ├── read rows: 1500000                                                                       |
| │   ├── read bytes: 135769198                                                                    |
| │   ├── partitions total: 10                                                                     |
| │   ├── partitions scanned: 10                                                                   |
| │   └── push downs: [filters: [], limit: NONE]                                                   |
| └── TableScan(Probe)                                                                             |
|     ├── table: default.tpch.orders                                                               |
|     ├── read rows: 15000000                                                                      |
|     ├── read bytes: 711675022                                                                    |
[...]

PR #8377, opened 2022-10-21
feat(planner): optimize join plan to make filter push down

### LINKED ISSUE #8073: TPCH: q13 oom even if gets resonable join order
**Summary**

For tpch q13

```sql

select
    c_count,
    count(*) as custdist
from
    (
        select
            c_custkey,
            count(o_orderkey) as c_count
        from
            customer
                left outer join
            orders
            on c_custkey = o_custkey
                and o_comment not like '%pending%deposits%'
        group by
            c_custkey
    )
        c_orders
group by
    c_count
order by
    custdist desc,
    c_count desc;
```

It's explain info
```
+-------------------------------------------------------------------------------------------------------+
| explain                                                                                               |
+-------------------------------------------------------------------------------------------------------+
| Sort                                                                                                  |
| ├── sort keys: [custdist DESC NULLS LAST, c_count DESC NULLS LAST]                                    |
| └── EvalScalar                                                                                        |
|     ├── expressions: [COUNT(*) (#22)]                                                                 |
|     └── AggregateFinal                                                                                |
|         ├── group by: [c_count]                                                                       |
|         ├── aggregate functions: [count()]                                                            |
|         └── AggregatePartial                                                                          |
|             ├── group by: [c_count]                                                                   |
|             ├── aggregate functions: [count()]                                                        |
|             └── EvalScalar                                                                            |
|                 ├── expressions: [count(o_orderkey) (#19)]                                            |
|                 └── AggregateFinal                                                                    |
|                     ├── group by: [c_custkey]                                                         |
|                     ├── aggregate functions: [count(o_orderkey)]                                      |
|                     └── AggregatePartial                                                              |
|                         ├── group by: [c_custkey]                                                     |
|                         ├── aggregate functions: [count(o_orderkey)]                                  |
|                         └── HashJoin                                                                  |
|                             ├── join type: RIGHT OUTER                                                |
|                             ├── build keys: [CAST(customer.c_custkey (#0) AS BIGINT NULL)]            |
|                             ├── probe keys: [CAST(orders.o_custkey (#9) AS BIGINT NULL)]              |
|                             ├── filters: [NOT LIKE(orders.o_comment (#16), %pending%deposits%)]       |
|                             ├── TableScan(Build)                                                      |
|                             │   ├── table: default.tpch.customer                                      |
|                             │   ├── read rows: 15000000                                               |
|                             │   ├── read bytes: 1357000371                                            |
|                             │   ├── partitions total: 32                                              |
|                             │   ├── partitions scanned: 32                                            |
|                             │   └── push downs: [filters: [], limit: NONE]                            |
|                             └── TableScan(Probe)                                                      |
|                                 ├── table: default.tpch.orders                                        |
|                                 ├── read rows: 150000000                                              |
|                                 ├── read bytes: 7304745883                                            |
|                                 ├── partitions total: 32                                              |
|                                 ├── partitions scanned: 32                                            |
|                                 └── push downs: [filters: [], limit: NONE]                            |
+-------------------------------------------------------------------------------------------------------+
```

Maybe we need to improve right outer join performance.

xudong963:

## Summary

Summary about this PR

Fixes #8073 https://github.com/datafuselabs/databend/pull/8370


xudong963:
After the ticket, tpch q13 won't oom in 16C, 32G standalone mode!

sundy-li (approved):

src/query/sql/src/planner/binder/join.rs
```diff
@@ -43,6 +43,13 @@ use crate::plans::Scalar;
 use crate::plans::ScalarExpr;
 use crate::BindContext;
 
+pub struct JoinConditions {
+    pub(crate) left_conditions: Vec<Scalar>,
+    pub(crate) right_conditions: Vec<Scalar>,
+    pub(crate) non_equi_conditions: Vec<Scalar>,
+    pub(crate) other_conditions: Vec<Scalar>,
```
leiysky, line 50: What's the difference between `other_conditions` and `non_equi_conditions`?

xudong963: `other_conditions` can be pushed down. 
such as `select * from t left join t1 on t1.a > 1`, `t1.a > 1` is a `other_condition` and it can be pushed down.

`other_conditions` only exist in logical join, they'll be eliminated in the optimizer.

`non_equi_conditions` are neither `equi_conditions` nor conditions that can be pushed down. They should keep in `PhysicalJoin` and execute in `HashJoin` or we'll get the wrong results.

leiysky:
I think pushing down the join conditions like `x = 3` in `bind_join` is possible, so we don't need to add a new optimization rule.

And besides, it seems we don't actually need both `other_condtions` and `non_equi_conditions` at the same time.

tests/logictest/suites/mode/cluster/04_0002_explain_v2
```diff
@@ -116,7 +116,7 @@ EvalScalar
 ├── scalars: [t1.a (#0), t1.b (#1), t2.a (#2), t2.b (#3)]
 └── LogicalJoin: INNER
     ├── equi conditions: [(t1.a (#0) = t2.a (#2)) AND (t1.b (#1) = t2.b (#3))]
-    ├── other conditions: [t1.a (#0) > 2]
```
leiysky, line 119: Should this predicate be pushed down to `t1`? It's disappeared in the explain result.

tests/logictest/suites/mode/standalone/explain/explain.test
```diff
@@ -102,7 +102,7 @@ EvalScalar
 ├── scalars: [t1.a (#0), t1.b (#1), t2.a (#2), t2.b (#3)]
 └── LogicalJoin: INNER
     ├── equi conditions: [(t1.a (#0) = t2.a (#2)) AND (t1.b (#1) = t2.b (#3))]
-    ├── other conditions: [t1.a (#0) > 2]
```
leiysky, line 105: ditto.

tests/logictest/suites/crdb/join
```diff
@@ -306,21 +306,9 @@ SELECT * FROM onecolumn NATURAL JOIN twocolumn;
 42 53
 
 statement query IIII
-SELECT * FROM twocolumn AS a JOIN twocolumn AS b ON a.x = a.y;
```
leiysky, line 309: Is this correct? The result set changes.

xudong963: The original result is wrong, the ticket caught the error. Now the result is correct.

xudong963:
> so we don't need to add a new optimization rule.

Nice, fixed

leiysky (approved):

xudong963:
@mergify update

xudong963:
> refusing to allow a GitHub App to create or update workflow `.github/workflows/build-tool.yml` without `workflows` permission err-code: ABBA4

strange...

PR #8388, opened 2022-10-21
feat(query): get the system open file limits and set

TszKitLo40:

## Summary

If the open file limit is less than 65535, set it to 65535.

Fixes #8387

### DIFF
--- a/src/binaries/query/main.rs
+++ b/src/binaries/query/main.rs
@@ -76,8 +76,7 @@ async fn main(_global_tracker: Arc<RuntimeTracker>) -> common_exception::Result<
     set_panic_hook();
 
     #[cfg(not(target_os = "macos"))]
-    // Try to set max open files.
-    try_set_max_open_files();
+    check_max_open_files();
 
     GlobalServices::init(conf.clone()).await?;
     let mut shutdown_handle = ShutdownHandle::create()?;
@@ -259,12 +258,20 @@ fn run_cmd(conf: &Config) -> bool {
 }
 
 #[cfg(not(target_os = "macos"))]
-fn try_set_max_open_files() {
-    let limits = get_own_limits().unwrap();
+fn check_max_open_files() {
+    let limits = match get_own_limits() {
+        Ok(limits) => limits,
+        Err(err) => {
+            warn!("get system limit of databend-query failed: {:?}", err);
+            return;
+        }
+    };
     let max_open_files_limit = limits.max_open_files.soft;
     if let Some(max_open_files) = max_open_files_limit {
         if max_open_files < 65535 {
-            warn!("The open file limit is too low for the databend-query. Please consider increase it by running `ulimit -n 65535`");
+            warn!(
+                "The open file limit is too low for the databend-query. Please consider increase it by running `ulimit -n 65535`"
+            );
         }
     }
 }
--- a/src/binaries/query/main.rs
+++ b/src/binaries/query/main.rs
@@ -34,7 +34,9 @@ use databend_query::servers::MySQLHandler;
 use databend_query::servers::Server;
 use databend_query::servers::ShutdownHandle;
 use databend_query::GlobalServices;
+use limits_rs::get_own_limits;
 use tracing::info;
+use tracing::warn;
 
 #[databend_main]
 async fn main(_global_tracker: Arc<RuntimeTracker>) -> common_exception::Result<()> {
@@ -72,6 +74,10 @@ async fn main(_global_tracker: Arc<RuntimeTracker>) -> common_exception::Result<
 
     init_default_metrics_recorder();
     set_panic_hook();
+
+    #[cfg(not(target_os = "macos"))]
+    // Try to set max open files.
+    try_set_max_open_files();
 
     GlobalServices::init(conf.clone()).await?;
     let mut shutdown_handle = ShutdownHandle::create()?;
@@ -250,4 +256,23 @@ fn run_cmd(conf: &Config) -> bool {
     }
 
     true
+}
+
+#[cfg(not(target_os = "macos"))]
+fn try_set_max_open_files() {
+    let limits = get_own_limits().unwrap();
+    let max_open_files_limit = limits.max_open_files.soft;
+    if let Some(max_open_files) = max_open_files_limit {
+        if max_open_files < 65535 {
+            let set = sysinfo::set_open_files_limit(max_open_files.try_into().unwrap());
+            match set {
+                true => {
+                    warn!("Open files limit has been set to {}", max_open_files);
+                }
+                false => {
+                    warn!("Open files limit set to {} failed", max_open_files);
+                }
+            }
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


src/binaries/query/main.rs
```diff
@@ -73,6 +76,17 @@ async fn main(_global_tracker: Arc<RuntimeTracker>) -> common_exception::Result<
     init_default_metrics_recorder();
     set_panic_hook();
 
+    let limits = get_own_limits().unwrap();
```
BohuTANG, line 79: Does this require root permissions??

BohuTANG, line 79: We prefer to move this logic to a method(Such as set_open_file_limits()). If there are internal errors we can ignore them and allow the server to start.

BohuTANG:
@mergify update

src/binaries/query/main.rs
TszKitLo40: I refer to https://crates.io/crates/limits-rs, but it does not any description of root permission.

TszKitLo40: OK. what do you think `try_set_open_files_limit()` should be put? Can I just put it in main.rs in query package?

BohuTANG:
I have pushed a demo FYI:
https://github.com/BohuTANG/databend/commit/c6fb93436755e01f4c6c45923b548ac956d6ee32

TszKitLo40:
> I have pushed a demo FYI: [BohuTANG@c6fb934](https://github.com/BohuTANG/databend/commit/c6fb93436755e01f4c6c45923b548ac956d6ee32)

updated

TszKitLo40:
@BohuTANG It seems that limits-rs can't work on MacOS.

BohuTANG:
@mergify update

BohuTANG:
> @BohuTANG It seems that limits-rs can't work on MacOS.

We can add:
```
#[cfg(not(target_os = "macos"))]
```

to check.

Xuanwo:
I don't like this PR.

It's not a good idea to set open file limits in code. How about printing a warning message with the action needs to take?

For example:

```shell
The open file limit is too low for the databend-query. Please consider increase it by running `ulimit -n 65536`
```

sundy-li:
> How about printing a warning message with the action needs to take?

That's right!

TszKitLo40:
> > How about printing a warning message with the action needs to take?
> 
> That's right!

OK，I will print a warning message instead of set open file limits in this PR.

src/binaries/query/main.rs
```diff
@@ -251,3 +257,14 @@ fn run_cmd(conf: &Config) -> bool {
 
     true
 }
+
+#[cfg(not(target_os = "macos"))]
+fn try_set_max_open_files() {
```
Xuanwo, line 262: As a naming convention in rust, functions named like `try_xxx` usually return `Result<yyy>` to indicate that this function could return an error.

Since we just get and print a warning message here, how about renaming this function to `check_max_open_files()`?

src/binaries/query/main.rs
```diff
@@ -251,3 +257,14 @@ fn run_cmd(conf: &Config) -> bool {
...
+    let limits = get_own_limits().unwrap();
```
Xuanwo, line 263: `get_own_limits` could return errors. If unwrap here, databend-query will fail to start. How about printing a message instead?

For example:

```suggestion
    let limits = match get_own_limits() {
        Ok(limits) => limits,
        Err(err) => {
            warn!("get system limit of databend-query failed: {e:?}";
            return
        }
    }
```

Xuanwo:
@TszKitLo40 Thanks for your work! This PR is almost LGTM now. I left some review here, PTAL.

Xuanwo (approved):
Thanks for your contribution!

PsiACE (approved):

Xuanwo:
CI failure need to be addressed:

```rust
error[E0425]: cannot find value `e` in this scope
   --> src/binaries/query/main.rs:265:64
    |
265 |             warn!("get system limit of databend-query failed: {e:?}");
    |                                                                ^ not found in this scope
```

TszKitLo40:
> CI failure need to be addressed:
> 
> ```rust
> error[E0425]: cannot find value `e` in this scope
>    --> src/binaries/query/main.rs:265:64
>     |
> 265 |             warn!("get system limit of databend-query failed: {e:?}");
>     |                                                                ^ not found in this scope
> ```

I think I have fixed it.

Xuanwo:
Please merge with the `main` branch to make sure our PR are updated~

src/binaries/query/main.rs
```diff
@@ -34,7 +34,9 @@ use databend_query::servers::MySQLHandler;
 use databend_query::servers::Server;
 use databend_query::servers::ShutdownHandle;
 use databend_query::GlobalServices;
+use limits_rs::get_own_limits;
```
BohuTANG, line 37: #[cfg(not(target_os = "macos"))]

src/binaries/query/main.rs
```diff
@@ -34,7 +34,9 @@ use databend_query::servers::MySQLHandler;
...
 use tracing::info;
+use tracing::warn;
```
BohuTANG, line 39: Remove the use.

src/binaries/query/main.rs
```diff
@@ -251,3 +256,22 @@ fn run_cmd(conf: &Config) -> bool {
...
+        Ok(limits) => limits,
+        Err(err) => {
+            warn!("get system limit of databend-query failed: {:?}", err);
+            return;
+        }
+    };
+    let max_open_files_limit = limits.max_open_files.soft;
+    if let Some(max_open_files) = max_open_files_limit {
+        if max_open_files < 65535 {
+            warn!(
```
BohuTANG, line 272: warn! --> tracing::warn!

BohuTANG:
This PR breaks the MacOS:
```
error: unused import: `limits_rs::get_own_limits`
  --> src/binaries/query/main.rs:37:5
   |
37 | use limits_rs::get_own_limits;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`

error: unused import: `tracing::warn`
  --> src/binaries/query/main.rs:39:5
   |
39 | use tracing::warn;
   |     ^^^^^^^^^^^^^
```

src/binaries/query/main.rs
```diff
@@ -251,3 +256,22 @@ fn run_cmd(conf: &Config) -> bool {
 
     true
 }
+
+#[cfg(not(target_os = "macos"))]
+fn check_max_open_files() {
+    let limits = match get_own_limits() {
```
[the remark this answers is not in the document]
TszKitLo40: Do I need to file another PR to fix this problems?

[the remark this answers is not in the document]
BohuTANG: image
This works on my MacOS.

PR #8390, opened 2022-10-21
feat(query): support SELECT output nothing.

### LINKED ISSUE #8379: Feature: a SQL can output nothing.
**Summary**

Sometime, we just want to test the performance of a SQL and don't care about the output table. 
Like `do_something > /dev/null`.

We can introduce such a syntax:

```sql
SELECT * FROM t FORMAT NULL;
```

The result will be like:

```
mysql> select 1;
+------+
| 1    |
+------+
|    1 |
+------+
1 row in set (0.02 sec)
Read 1 rows, 1.00 B in 0.002 sec., 592.23 rows/sec., 592.23 B/sec.

mysql> select 1 FORMAT NULL;
1 row in set (0.02 sec)
Read 1 rows, 1.00 B in 0.002 sec., 592.23 rows/sec., 592.23 B/sec.
```

In the future, we can also add other output format .

RinChanNOWWW:

## Summary

```
mysql> select number from numbers(2);
+--------+
| number |
+--------+
|      0 |
|      1 |
+--------+
2 rows in set (0.04 sec)
Read 2 rows, 16.00 B in 0.004 sec., 445.89 rows/sec., 3.48 KiB/sec.

mysql> select number from numbers(2) ignore_result;
Empty set (0.03 sec)
Read 2 rows, 16.00 B in 0.004 sec., 462.98 rows/sec., 3.62 KiB/sec.
```

Fixes #8379

### DIFF
--- a/docs/doc/30-reference/30-sql/20-query-syntax/dml-select.md
+++ b/docs/doc/30-reference/30-sql/20-query-syntax/dml-select.md
@@ -20,6 +20,7 @@ SELECT
     [ORDER BY {col_name | expr | col_alias | col_position} [ASC | DESC], [ NULLS { FIRST | LAST }]
     [LIMIT row_count]
     [OFFSET row_count]
+    [FORMAT format]
     ]
 ```
 
@@ -43,7 +44,7 @@ SELECT number FROM numbers(3);
 ## FROM Clause
 
 ```sql
-SELECT number FROM numbers(3) AS a; 
+SELECT number FROM numbers(3) AS a;
 +--------+
 | number |
 +--------+
@@ -180,7 +181,7 @@ INSERT INTO t_null VALUES (3);
 INSERT INTO t_null VALUES (NULL);
 INSERT INTO t_null VALUES (NULL);
 
---Databend considers NULL values larger than any non-NULL values. 
+--Databend considers NULL values larger than any non-NULL values.
 --The NULL values appear last in the following example that sorts the results in ascending order:
 
 SELECT number FROM t_null order by number ASC;
@@ -251,6 +252,27 @@ SELECT number FROM numbers(5) ORDER BY number OFFSET 2;
 |      3 |
 |      4 |
 +--------+
+```
+
+## FORMAT Clause
+
+Format the output of the query result.
+
+Current supported format:
+
+- NULL: output nothing.
+
+```sql
+SELECT number FROM numbers(2);
++--------+
+| number |
++--------+
+|      0 |
+|      1 |
++--------+
+
+mysql> SELECT number FROM numbers(2) FORMAT NULL;
+-- Empty set
 ```
 
 ## Nested Sub-Selects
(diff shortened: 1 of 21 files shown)


docs/doc/30-reference/30-sql/20-query-syntax/dml-select.md
```diff
@@ -20,6 +20,7 @@ SELECT
     [ORDER BY {col_name | expr | col_alias | col_position} [ASC | DESC], [ NULLS { FIRST | LAST }]
     [LIMIT row_count]
     [OFFSET row_count]
+    [FORMAT format]
```
youngsofun, line 23: I prefer `IGNORE_RESULT` or `NO_OUTPUT` instead of reuse `FORMAT`

b41sh: +1, `FORMAT NULL` may be confused.

sundy-li: What's the problem of reusing the Null as a `format` name?

RinChanNOWWW: > What's the problem of reusing the Null as a format name?

It's a little confused.

sundy-li:
We should allow one of `FORMAT XXX | IGNORE_RESULT`

`select * from t format CSV IGNORE_RESULT` is not allowed.

RinChanNOWWW:
> We should allow one of `FORMAT XXX | IGNORE_RESULT`
> 
> `select * from t format CSV IGNORE_RESULT` is not allowed.

`FORMAT` is only used in clickhouse handler and `IGNORE_RESULT` is only used in mysql handler.

RinChanNOWWW:
@mergify update

sundy-li (approved):

Xuanwo:
`select * from abc ignore_result` is not a good extension to me.

How about introducing a new statement like `PIPE INTO NULL select * from abc`?

In the future, we can extend it to `pipe into 's3://result' format csv select * from abc`.

Xuanwo:
> How about introducing a new statement like `PIPE INTO NULL select * from abc`?

Result cache seems a particular case like `PIPE INTO @internal_stage/query-id.parquet select * from abc`. Maybe we can share some code in the future?

Cc @youngsofun for comments.

RinChanNOWWW:
> PIPE INTO NULL select * from abc

It seems a nice idea. Is there any progress about this feature now?

Xuanwo:
> It seems a nice idea. Is there any progress about this feature now?

No, it's a brand-new idea that came up one hour ago.

BohuTANG (approved):

Xuanwo:
> If this feature looks good, we can start from `pipe into null` first.

most interesting idea from @BohuTANG: `COPY INTO NULL FROM (select * from abc)`

PR #8393, opened 2022-10-21
feat(storage): add blocking api for fuse source

sundy-li:

## Summary

Summary about this PR

Fixes #issue


src/query/storages/fuse/src/io/read/block_reader.rs
```diff
@@ -344,6 +344,39 @@ impl BlockReader {
...
+        for index in indices {
+            let column_meta = &part.columns_meta[&index];
+
+            let op = self.operator.clone();
+
+            let location = part.location.clone();
+            let offset = column_meta.offset;
+            let length = column_meta.length;
+
+            threads.push(std::thread::spawn(move || {
```
zhang2014, line 369: std::thread -> common_base::base::Thread

Xuanwo:
The layer's issue will be resolved by https://github.com/datafuselabs/databend/pull/8421

src/query/storages/fuse/src/io/read/block_reader.rs
Xuanwo: Maybe we can use https://github.com/rayon-rs/rayon?

zhang2014: common_base::base::Thread is used for resource tracking.

Xuanwo (approved):

BohuTANG (approved):

PR #8404, opened 2022-10-22
feat: support copy from xml.

eliasyaoyc:
Signed-off-by: Elias.Yao <siran0611@gmail.com>

## Summary
my implementation references mysql, and test complele stream load and copy stage
format1:
```
<row column1="value1" column2="value2" .../>
```
format2: 
```
<row>
  <column1>value1</column1>
  <column2>value2</column2>
</row>
```
format3:
```
<row>
  <field name='column1'>value1</field>
  <field name='column2'>value2</field>
</row>
```
finally,  line 119 in input_context.rs, would it be better to have then as arguments to the input_context.


drmingdrmer:
If proto files change, add a line to increase the version and describe what changes are introduced:

https://github.com/datafuselabs/databend/blob/421c0fb60c26c7a3ce551bcec1a579f9c4fee753/src/meta/proto-conv/src/util.rs#L65

And add test to load this version of the message, reference:

drmingdrmer:
If proto files change, add a line to increase the version and describe what changes are introduced:

https://github.com/datafuselabs/databend/blob/421c0fb60c26c7a3ce551bcec1a579f9c4fee753/src/meta/proto-conv/src/util.rs#L65

And add test to load this version of the message, reference: 

https://github.com/datafuselabs/fuse-query/blob/ce4e7a40ae890eb4f5ea3b943c93a50811d13ea8/src/meta/proto-conv/tests/it/user_proto_conv.rs#L171

src/meta/types/src/user_stage.rs
```diff
@@ -138,7 +138,8 @@ impl FromStr for StageFileFormatType {
             "TSV" | "TABSEPARATED" => Ok(StageFileFormatType::Tsv),
             "NDJSON" | "JSONEACHROW" => Ok(StageFileFormatType::NdJson),
             "PARQUET" => Ok(StageFileFormatType::Parquet),
-            "XML" | "ORC" | "AVRO" | "JSON" => Err(format!(
+            "XML" => Ok(StageFileFormatType::Xml),
+            "ORC" | "AVRO" | "JSON" => Err(format!(
                 "File format type '{s}' not implemented yet', must be one of ( CSV | TSV | NDJSON | PARQUET)"
```
sundy-li, line 143: XML

drmingdrmer (approved):

src/query/settings/src/lib.rs
```diff
@@ -269,6 +269,26 @@ impl Settings {
...
+                possible_values: None,
+            },
+            SettingValue {
+                default_value: UserSettingValue::String("data".to_owned()),
+                user_setting: UserSetting::create(
+                    "rowset_tag",
+                    UserSettingValue::String("data".to_owned()),
+                ),
+                level: ScopeLevel::Session,
+                desc: "In xml format, this fields is represented as a row set tag, e.g.  <list> <row>...</row>, <row>...</row> </list>.",
```
youngsofun, line 289: example with default value `data`?

youngsofun: not clear where `rowset_tag` is used from the `desc`

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
```diff
@@ -0,0 +1,254 @@
...
+            data: vec![],
+            row_ends: vec![],
+            field_ends: vec![],
+            path: state.path.to_string(),
+            batch_id: state.batch_id,
+            offset: 0,
+            start_row: Some(state.rows),
+        };
+
+        let mut cols = Vec::<Vec<u8>>::with_capacity(state.num_fields);
```
youngsofun, line 144: should use a map here?

src/query/settings/src/lib.rs
eliasyaoyc:
```
<?xml version="1.0"?>

<!-- rowset tag -->
<data>
  <!-- row tag -->
  <row title="Transaction Processing" author="Jim Gray" date="1992"/>
</data>

<data>
  <row>
  <title>Readings in Database Systems</title>
  <author>Michael Stonebraker</author>
  <date>2004</date>
  </row>
</data>

<data>
  <row>
  <field name="title">aaaa</field>
  <field name="author">bbbb</field>
  <field name="date">2022</field>
  </row>
</data>
```

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
eliasyaoyc: there's no need to add a key here, so i'm not using a map.

youngsofun: I mean order column names in XML may be different with that in table definination?

src/query/settings/src/lib.rs
youngsofun: seems mysql does not use `rowset tag`, only `row tag`
 https://dev.mysql.com/doc/refman/8.0/en/load-xml.html.

can you explain some explanation why we need it?
if not highly necessary, I prefer fewer options.

eliasyaoyc: yes, mysql is use <list> tag to wrap the row.  this forces the user to use <list>,which is unreadable if the user uses <data> tag to wrap the row. 
```
<list>
<row_tag>
..
</row_tag>
</list>
```

another problem is that the user's xml might be, then have to specify the rowset tag.
```
<list>
<row_tag>
..
</row_tag>
</list>

<list>
<row_tag>
..
</row_tag>
</list>
```
🤔 but if we need to simplify options, we can specify it explicity in the dosc.

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
eliasyaoyc: i think there is a problem, herer we should not force the user to be defined in the same order as the corresponding table. 
😊 i'll fix it later.

src/query/settings/src/lib.rs
eliasyaoyc: i do this to make the user more flexible, maybe im just overthinking it.

youngsofun: `rowset tag` is `list`?

youngsofun: do you mean `row_tag` should be under `rowset tag`?

Mysql seems to only match the tag no matter which level is in.

e.g.  `address` is under `person` in the example (search  ` ROWS IDENTIFIED BY '<address>'` in https://dev.mysql.com/doc/refman/8.0/en/load-xml.html)

eliasyaoyc: i re-checked my code an recalled why i used rowset_tag eariler, its stupid. 😭
i just want to fit the following format:
```
<data>
  <row>
  <title>Readings in Database Systems</title>
  <author>Michael Stonebraker</author>
  <date>2004</date>
  </row>
</data>
```
when the attrs of tag is empty, only has three possibilities. <rowset_tag>,<row_tag>,<column_tag>.
i did a really stupid thing where i added a <rowset_tag> options to partition and get the <column_tag>.

🧐 i can use a temp variable such as has_row_start to record the start of the row and remove the <rowset_tag>

eliasyaoyc:
```
<?xml version="1.0"?>

<!-- rowset tag -->
<data>
  <!-- row tag -->
  <row title="Transaction Processing" author="Jim Gray" date="1992"/>
</data>

<data>
  <row>
  <title>Readings in Database Systems</title>
  <author>Michael Stonebraker</author>
  <date>2004</date>
  </row>
</data>

<data>
  <row>
  <field name="title">aaaa</field>
  <field name="author">bbbb</field>
  <field name="date">2022</field>
  </row>
</data>
```

result:
image

src/meta/types/src/user_stage.rs
Xuanwo: Please fix this.

Xuanwo:
@mergify update

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
```diff
@@ -0,0 +1,257 @@
+use std::collections::HashMap;
```
Xuanwo, line 1: This use is out of order

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
```diff
@@ -0,0 +1,257 @@
...
+use common_exception::Result;
+use common_io::prelude::FormatSettings;
+use common_io::prelude::NestedCheckpointReader;
+use common_meta_types::StageFileFormatType;
+use common_settings::Settings;
+use xml::reader::XmlEvent;
+use xml::ParserConfig;
+
+use crate::processors::sources::input_formats::impls::input_format_tsv::format_column_error;
+use crate::processors::sources::input_formats::input_format_text::get_time_zone;
```
Xuanwo, line 30: Use `super::xxx` for better reading.

src/query/pipeline/sources/Cargo.toml
```diff
@@ -35,3 +35,4 @@ parking_lot = "0.12.1"
 serde_json = { workspace = true }
 similar-asserts = "1.4.2"
 tracing = "0.1.36"
+xml-rs = "0.8.4"
```
Xuanwo, line 38: Add xml-rs in `[package.metadata.cargo-machete]` to make `cargo-machete` happy.

Like

https://github.com/datafuselabs/databend/blob/1dd316bd4d22c1b38ab2cec04f5a23a7e8f20bd5/src/query/config/Cargo.toml#L16-L17

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
```diff
@@ -0,0 +1,257 @@
...
+impl InputFormatXML {
+    fn read_row(
+        buf: &[u8],
+        deserializers: &mut [TypeDeserializerImpl],
+        schema: &DataSchemaRef,
+        format_settings: &FormatSettings,
+        path: &str,
+        row_index: usize,
+    ) -> Result<()> {
+        let mut row_json: serde_json::Value = serde_json::from_reader(buf)?;
```
Xuanwo, line 48: What's this for? We need to read XML as json first?

Xuanwo:
Please add a stateful test case to make sure we can load data from XML correctly~

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
youngsofun: I guess this pr is not ready for review  @eliasyaoyc? maybe you can convert this pr a draft?

eliasyaoyc:
```
let cols_bytes = serde_json::to_vec(&cols)?;

output.data.extend_from_slice(&cols_bytes);
```
since i converted each row of data to json, i originally used Vec<u8>, but in order to solve the problem of inconsistent column order  in xml and table order , i used map to solve this problem, so i conveted it to json for read_row to read easy

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
eliasyaoyc: This looks weird 👀
```
use super::super::input_format_text::get_time_zone;
```

youngsofun:
not familiar with the XML lib you use, if it does not support streaming (I hope not), you can choose to read the whole file before parsing (which limits the max XML size we can load).

by the way, do you know any real life cases of loading from XML?

src/query/pipeline/sources/src/processors/sources/input_formats/impls/input_format_xml.rs
youngsofun: not familiar with the XML lib you use, 
if it does not support streaming in(bytes) and out(XMLEvent),  I suggest to use Aligner only to accumulate and concat all ReadBatch, and sent all content of the File a RowBatch and call XML reader there.

Xuanwo:
> by the way, do you know any real life cases of loading from XML?

We can test against https://www.treasurydirect.gov/xml/

Docs here: https://community.snowflake.com/s/article/How-to-Easily-Load-XML-with-SQL-Pt-1#:~:text=LOADING%20XML%20DATA%20INTO%20SNOWFLAKE&text=Create%20a%20new%20table%20with,column%20of%20the%20new%20table.

youngsofun:
is it possible to support both style of snowflake XML(load only) and MySQL XML loading.
also considering clickhouse xml (unload only).

given not require the exact same option names.

youngsofun:
Personally, I don`t approve to use ontime for this tests.  (more generally,  real-life data vs man-made data)

except for the original (not transformed) and commonly used data.

1.  big in size
2. longer ci time.
3. [VIP] not knowing how the file is generated, there may be errror when generating it ( our ontime.csv seems buggy too).
4. not showing much value despite the cost; can not cover all cases (e.g. definitely not for  data types)
5. no help for others to demo more features.

Xuanwo:
> is it possible to support both style of snowflake XML(load only) and MySQL XML loading.

How about creating an issue for this and improving them in the future?

> Personally, I don`t approve to use ontime to test XML.

Any ideas about that? How about generating a simple XML with 200 lines for testing (like other formats)?

youngsofun:
Wecan start with demo，using a simple schema with all basic tpyes，sevrel line is enough，add some 
basic corner cases

1.variant of special xml structure
2.empty null bool ……

youngsofun:
ontime.xml can be removed now.
and need resolve the conflicts.

src/meta/proto-conv/src/util.rs
```diff
@@ -63,9 +63,13 @@ const META_CHANGE_LOG: &[(u64, &str)] = &[
     ),
     (15, "2022-10-12: Remove: precision in TimestampType"),
     (16, "2022-09-29: Add: CopyOptions::split_size"),
-    (17, "2022-10-28: Add: StageType::LegacyInternal"),
-    (18, "2022-10-28: Add: FILEFormatOptions::escape"),
-    (19, "2022-10-31: Add: StageType::UserStage"),
+    (
+        17,
+        "2022-10-22: Add: users.proto/FileFormatOptions::row_tag",
```
youngsofun, line 68: should not change the orders, for this pr, you can rebase locally if need

tests/suites/1_stateful/05_input_formats/05_0003_xml_v1.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
```
youngsofun, line 2: better if add a test to use custom row_tag?

tests/suites/1_stateful/05_input_formats/05_0003_xml_v1.sh
```diff
@@ -0,0 +1,43 @@
...
+        <DATA>\"猛\"男</DATA>
+        <CREATE_TIME>2022-11-01 10:51:14</CREATE_TIME>
+        <EMPTY>123</EMPTY>
+    </row>
+</data>
+EOF
+
+curl -sH "insert_sql:insert into test_xml format XML" -F "upload=@/tmp/simple_v1.xml" -u root: -XPUT "http://localhost:${QUERY_HTTP_HANDLER_PORT}/v1/streaming_load" | grep -c "SUCCESS"
+echo "select * from test_xml" | $MYSQL_CLIENT_CONNECT
+echo "truncate table test_xml" | $MYSQL_CLIENT_CONNECT
```
youngsofun, line 43: add newline at file end.

src/common/io/src/format_settings.rs
```diff
@@ -68,6 +68,8 @@ pub struct FormatSettings {
     pub json_quote_denormals: bool,
     pub json_escape_forward_slashes: bool,
     pub ident_case_sensitive: bool,
+
+    pub row_tag: Vec<u8>,
```
BohuTANG, line 72: `xml_row_tag` is better?

youngsofun: agree, but we still using `row_tag` as a format option for the end user, since we are flowing snowflake style to not prefix format_name, https://docs.snowflake.com/en/sql-reference/sql/create-file-format.html

src/meta/proto-conv/tests/it/user_stage.rs
```diff
@@ -53,6 +53,50 @@ fn test_user_stage_oss_latest() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_user_stage_fs_v20() -> anyhow::Result<()> {
+    // Encoded data of version 17 of user_stage_fs:
+    // It is generated with common::test_pb_from_to.
+    let user_stage_fs_v17 = vec![
```
youngsofun, line 60: v20

src/meta/protos/proto/user.proto
```diff
@@ -181,8 +181,10 @@ message UserStageInfo {
     string record_delimiter = 4;
 
     StageFileCompression compression = 5;
-
-    string escape = 6;
+    
+    string row_tag = 6;
```
youngsofun, line 185: should not change this order

youngsofun:
others LGTM

sundy-li (approved):
LGTM, thank you for your persistence and contribution.

Xuanwo:
cc @soyeric128, we supports copy from xml format now!

soyeric128:
> cc @soyeric128, we supports copy from xml format now!

OK. Nice to see COPY INTO grows stronger, BTW.

soyeric128:
Some quick questions about the file formats:

- Does the streaming load API also support for xml ?
- Does COPY INTO support for TSV files?

Xuanwo:
> * Does the streaming load API also support for XML ?

Y.

> * Does COPY INTO support for TSV files?

Y.

soyeric128:
Does a single data file have a size limit when working with the streaming load API or COPY INTO command?

Xuanwo:
> Does a single data file have a size limit when working with the streaming load API or COPY INTO command?

No such limit so far.

---

It's better to start a new discussion at https://github.com/datafuselabs/databend/discussions so that we are not disturbing subscribers of this PR :love_letter:

PR #8406, opened 2022-10-22
docs(meta/management): explain when to join a cluster

drmingdrmer:

##### refactor(meta/management): when joining a node, detect membership instead of log

A databend-meta node may always be started with `--join` arg.
With this arg, meta node should decide whether to execute the
**join-to-a-cluster** process.

Before this commit it detect if there are logs. This way is not very
relliable: it's possible the leader has setup a replication to this new
node but not yet added it as a **voter**. In such a case, this node will
never be added into the cluster automatically.

In stead, in this commit it detect if there is a **membership** config
that includes this node to decide if to join a cluster. Thus only when a
node has already joined to a cluster(leader committed the membership and
has replicated it to this node), it skips the join process.

##### chore(meta): add log of handshake

- Part of #8383

## Related Issues

### DIFF
--- a/src/meta/service/src/api/grpc/grpc_service.rs
+++ b/src/meta/service/src/api/grpc/grpc_service.rs
@@ -47,6 +47,7 @@ use tonic::Request;
 use tonic::Response;
 use tonic::Status;
 use tonic::Streaming;
+use tracing::debug;
 use tracing::info;
 
 use crate::meta_service::meta_service_impl::GrpcStream;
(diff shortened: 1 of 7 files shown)


drmingdrmer:
Looks like the new policy exposes another issue with `metactl --import`. I'll fix it tomorrow.

src/meta/service/src/api/grpc/grpc_service.rs
```diff
@@ -149,6 +151,8 @@ impl MetaService for MetaServiceImpl {
                 payload: token.into_bytes(),
             };
             let output = futures::stream::once(async { Ok(resp) });
+
+            info!("handshake OK");
```
BohuTANG, line 155: Change to debug level?

PR #8417, opened 2022-10-24
feat: Add metainfo in datablock

zhyass:

## Summary

Summary about this PR

Fixes #issue

### DIFF
--- a/src/query/datablocks/src/meta_info.rs
+++ b/src/query/datablocks/src/meta_info.rs
@@ -1,0 +1,40 @@
+// Copyright 2022 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


src/query/datablocks/src/meta_info.rs
```diff
@@ -0,0 +1,40 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::any::Any;
+use std::sync::Arc;
+
+#[typetag::serde(tag = "type")]
+pub trait MetaInfo: Send + Sync {
```
Xuanwo, line 19: Why we need to introduce a trait for this?

zhyass: > Why we need to introduce a trait for this?

The output and input of the pipeline rely on the datablock. by adding metainfo to the datablock, some additional information can be passed downstream. This is very useful in delete, update, etc.

zhang2014 (approved):

src/query/datablocks/src/meta_info.rs
BohuTANG: It would be better to add a unit test to show how to use it.

sundy-li (approved):

BohuTANG:
cc @youngsofun

PR #8421, opened 2022-10-24
fix: OpenDAL layers doesn't pass not implemented methods correctly

Xuanwo:

## Summary

fix: OpenDAL layers doesn't pass not implemented methods correctly


Xuanwo:
Simple change, let's merge

PR #8427, opened 2022-10-24
refactor(planner): use right mark join as subquery's default join type

xudong963:

## Summary

The ticket contains two things:

- Right mark join is executed by streaming, there is no blocking pipeline operator. So use right mark join as default correlated subquery join type.
- Fix some bugs in right mark join

Fixes #issue

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/hash_join/common.rs
+++ b/src/query/service/src/pipelines/processors/transforms/hash_join/common.rs
@@ -143,7 +143,6 @@ impl JoinHashTable {
                 }
             }
         }
-        dbg!(markers.clone());
         markers
     }
 
--- a/src/query/service/src/pipelines/processors/transforms/hash_join/common.rs
+++ b/src/query/service/src/pipelines/processors/transforms/hash_join/common.rs
@@ -143,6 +143,7 @@ impl JoinHashTable {
                 }
             }
         }
+        dbg!(markers.clone());
         markers
     }
 
--- a/src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_state_impl.rs
+++ b/src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_state_impl.rs
@@ -137,7 +137,7 @@ impl HashJoinState for JoinHashTable {
                     .map(|x| Some(*x))
                     .collect(),
                 JoinType::RightMark => {
-                    if !has_null {
+                    if !has_null && !chunk.cols.is_empty() {
                         if let Some(validity) = chunk.cols[0].validity().1 {
                             if validity.unset_bits() > 0 {
                                 has_null = true;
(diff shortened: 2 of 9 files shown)


src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_state_impl.rs
```diff
@@ -137,7 +137,7 @@ impl HashJoinState for JoinHashTable {
                     .map(|x| Some(*x))
                     .collect(),
                 JoinType::RightMark => {
-                    if !has_null {
+                    if !has_null && !chunk.cols.is_empty() {
```
xudong963, line 140: Bug in right mark join: `cols` is possibly empty; Such as the  case: ` select 1 = all(select 1);` cc @youngsofun

src/query/service/src/pipelines/processors/transforms/hash_join/probe_join/right_mark.rs
```diff
@@ -44,8 +44,7 @@ impl JoinHashTable {
     {
         let valids = &probe_state.valids;
         let has_null = *self.hash_join_desc.marker_join_desc.has_null.read();
-        let mut markers = Self::init_markers(input.columns(), input.num_rows());
```
xudong963, line 47: Bug in right mark join: we should use `probe_keys` to init markers, not probe input columns. cc @youngsofun

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_state_impl.rs
youngsofun: ok

src/query/service/src/pipelines/processors/transforms/hash_join/common.rs
```diff
@@ -143,6 +143,7 @@ impl JoinHashTable {
                 }
             }
         }
+        dbg!(markers.clone());
```
BohuTANG, line 146: The debug codes are no need?

src/query/sql/src/planner/optimizer/heuristic/decorrelate.rs
```diff
@@ -342,16 +342,16 @@ impl SubqueryRewriter {
                     )
                 };
                 let mark_join = LogicalInnerJoin {
-                    left_conditions,
-                    right_conditions,
+                    left_conditions: right_conditions,
```
youngsofun, line 345: Maybe we can give a little different additional cost for different Join?

youngsofun (approved):

src/query/sql/src/planner/optimizer/heuristic/decorrelate.rs
xudong963: Do you mean join cardinality in join order?

youngsofun: for example, let left mark join cost a little more than right mark join.

PR #8449, opened 2022-10-25
feat: reduce the memory usage when compact

BohuTANG:

## Summary

Reduce the memory usage for `optimize table xx compact limit 100`, trim the col_stats:

Before: `fuse_compact_segments_memory_usage 1735155007 bytes`
This PR:`fuse_compact_segments_memory_usage 317131997 bytes`

Fixes #issue


BohuTANG:
But compact block is the memory killer again :\

PR #8451, opened 2022-10-25
fix: avoid oom when compact&recluster

BohuTANG:

## Summary

Set the max io requests to max threads when compact&recluster to avoid OOM.

Fixes #issue


dantengsky (approved):
LGTM

PR #8464, opened 2022-10-26
fix(query): div zero return err

### LINKED ISSUE #8463: bug: div zero should return err.
**Summary**

Now `select 1/0` will return inf.

Actually, it should return err.

TCeason:

## Summary

div 0 will return err in `functions` and `function-v2`

Fixes #8463


sundy-li (approved):

tests/logictest/suites/crdb/and_or
```diff
@@ -7,36 +7,32 @@ CREATE TABLE t (k INT null, a INT null, b INT null);
 statement ok
 INSERT INTO t VALUES (1, NULL, NULL), (2, NULL, 1), (3, 1, NULL), (4, 2, 0), (5, 3, 3);
 
-statement query T
```
TCeason, line 10: The err maybe should not return err. I will have a look

TCeason:
After this pr modify, now in `functions-v1` I think there have some othere bugs.

```sql
mysql> CREATE TABLE t (k INT null, a INT null, b INT null);

mysql> INSERT INTO t VALUES (1, NULL, NULL), (2, NULL, 1), (3, 1, NULL), (5, 3, 3);

mysql> select * from t order by k;
+------+------+------+
| k    | a    | b    |
+------+------+------+
|    1 | NULL | NULL |
|    2 | NULL |    1 |
|    3 |    1 | NULL |
|    5 |    3 |    3 |
+------+------+------+
4 rows in set (0.09 sec)
Read 4 rows, 51.00 B in 0.075 sec., 53.51 rows/sec., 682.30 B/sec.
```


If `select 3/b` with all table, the result is abnormal.

```sql
mysql> select 3/b from t;
ERROR 1105 (HY000): Code: 1006, displayText = / by zero.

```

But if I `select 3/b` line by line, the result is normal.

```
mysql> select 3/b from t where k=1;
+-------+
| 3 / b |
+-------+
|  NULL |
+-------+
1 row in set (0.10 sec)
Read 4 rows, 34.00 B in 0.083 sec., 48.13 rows/sec., 409.10 B/sec.

mysql> select 3/b from t where k=2;
+-------+
| 3 / b |
+-------+
|   3.0 |
+-------+
1 row in set (0.08 sec)
Read 4 rows, 34.00 B in 0.069 sec., 57.74 rows/sec., 490.75 B/sec.

mysql> select 3/b from t where k=3;
+-------+
| 3 / b |
+-------+
|  NULL |
+-------+
1 row in set (0.12 sec)
Read 4 rows, 34.00 B in 0.100 sec., 40.2 rows/sec., 341.70 B/sec.

mysql> select 3/b from t where k=5;
+-------+
| 3 / b |
+-------+
|   1.0 |
+-------+
1 row in set (0.12 sec)
Read 4 rows, 34.00 B in 0.106 sec., 37.62 rows/sec., 319.75 B/sec.

```

PR #8469, opened 2022-10-26
feat(hive): hive predict pushdown

sandflee:

## Summary
1. get column stats from parquet file
2. use range filter to do the predict

original committed by lizhenwei

### DIFF
--- a/src/query/settings/src/lib.rs
+++ b/src/query/settings/src/lib.rs
@@ -354,6 +354,7 @@ impl Settings {
                 desc: "The maximum query execution time. it means no limit if the value is zero. default value: 0.",
                 possible_values: None,
             },
+            #[cfg(feature = "hive")]
             SettingValue {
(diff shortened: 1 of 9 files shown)


sandflee:
@mergify update

src/query/settings/src/lib.rs
```diff
@@ -354,6 +354,17 @@ impl Settings {
                 desc: "The maximum query execution time. it means no limit if the value is zero. default value: 0.",
                 possible_values: None,
             },
+            #[cfg(feature = "hive")]
+            SettingValue {
```
BohuTANG, line 358: Why need a setting to control push down or not?

sandflee: if pushdown is not work as expected, we may lost data.  disable pushdown is one way to analyse the problem.

BohuTANG (approved):

PR #8473, opened 2022-10-26
fix(storage): compact limit

### LINKED ISSUE #8474: bug: compact limit cannot be optimized
**Summary**

For example
```
optimize table t compact limit 100;
```
When the first 100 segments do not need to be optimized, all subsequent segments cannot be optimized even if the optimize statement is executed again.

zhyass:

## Summary

For example
```
optimize table t compact limit 100;
```
When the first 100 segments do not need to be optimized, continue to the next 100 segments. This allows all segments to be optimized.

Fixes #8474


dantengsky (approved):
LGTM

Issue #7986, opened 2022-09-30
Flaky test: tpch test on cluster failed for reading a body from connection: broken pipe


Xuanwo:
**Summary**

```shell
--- /workspace/tests/suites/0_stateless/13_tpch/13_0007_q7.result        2022-09-30 05:37:31.856084133 +0000
+++ /workspace/tests/suites/0_stateless/13_tpch/13_0007_q7.stdout        2022-09-30 05:38:54.725091643 +0000
@@ -1,4 +1 @@
-FRANCE        GERMANY        1995        4637235.15
-FRANCE        GERMANY        1996        5224779.573
-GERMANY        FRANCE        1995        6232818.703
-GERMANY        FRANCE        1996        5557312.112
+ERROR 1105 (HY000) at line 2: Code: 1067, displayText = error reading a body from connection: broken pipe (while in processor thread 0).
```

https://github.com/datafuselabs/databend/actions/runs/3156329493/jobs/5136357403

Xuanwo:
@zhang2014 This query seems stable in recent CI, maybe it has been resolved?

xudong963:
> @zhang2014 This query seems stable in recent CI, maybe it has been resolved?

A recent case from @BohuTANG https://github.com/datafuselabs/databend/issues/8398#issuecomment-1287763686

Xuanwo:
```shell
--- /workspace/tests/suites/0_stateless/13_tpch/13_0003_q3.result	2022-10-25 01:10:26.021571701 +0000
+++ /workspace/tests/suites/0_stateless/13_tpch/13_0003_q3.stdout	2022-10-25 01:12:40.323415686 +0000
@@ -1,5 +1 @@
-223140	355369.0698	1995-03-14	0
-584291	354494.7318	1995-02-21	0
-405063	353125.4577	1995-03-03	0
-573861	351238.27699999994	1995-03-09	0
-554757	349181.7426	1995-03-14	0
+ERROR 1105 (HY000) at line 2: Code: 1067, displayText = error reading a body from connection: broken pipe, source: Some(hyper::Error(Body, Error { kind: Io(Kind(BrokenPipe)) })).
```

This error is from `tonic::Status`, the connection between the query and meta's grpc service.

Maybe the conn for grpc has been closed unexpectedly?

Cc @zhang2014

zhang2014:
> ```shell
[...]

PR #8511, opened 2022-10-28
fix: Move flight service client init to global io runtime

Xuanwo:

## Summary

fix: Move flight service client init to global io runtime

Related to https://github.com/datafuselabs/databend/issues/7986

Hyper client's connection is created upon runtime. If we start a HTTP client in current runtime (like query context runtime), this client could be dropped too early.

This PR is just for a non-harmful try. The root cause of #7986 is not identified.


BohuTANG:
Could you give more background about this PR?

Xuanwo:
> Could you give more background about this PR?

Updated the PR's description.

BohuTANG (approved):

PR #8519, opened 2022-10-28
rfc: User Stage

Xuanwo:

## Summary

```sql
COPY INTO my_table FROM @~;
LIST @~;
PRESIGN @~/data.csv;
REMOVE @~ PATTERN = 'ontime.*';
```


docs/doc/60-contributing/03-rfcs/20221028-user-stage.md
```diff
@@ -0,0 +1,133 @@
...
+    User,
+    External,
+}
+```
+
+`StageType::Named` will deprecate `StageType::Internal`. Since this RFC, we will not create a new stage with `StageType::Internal` anymore.
+
+The stage prefix rule will be:
+
+- `Internal` => `stage/{stage_name}`
```
BohuTANG, line 76: It looks more complex to the user.
I think we should remove the `create stage` from the document and let the user use `@~` by default.
So the user can:
1. Create an external stage
2. Use the default internal stage `@~`

Xuanwo: Users won't know about these implementation details.

For user:

```sql
CREATE STAGE abc; => named stage
LIST @~ => user stage (can't be created)
CREATE STAGE abc url='s3://load/files/'; => external stage
```

We can keep the named stage support but replace all our examples with the user stage.

BohuTANG: Ok.
From my understanding, the document still: `internal stage` and `external stage`
And we introduce a default internal stage: `@~` for the user?

BohuTANG (approved):
🚀

docs/doc/60-contributing/03-rfcs/20221028-user-stage.md
flaneur2020, line 76: will this prefix affect the existed stages? q.q

after prefix applied, the storage path might changed

flaneur2020: oh i see, Internal is for backwards compatibility, how about calls it LegacyInternal ?

Xuanwo: Nice idea!

docs/doc/60-contributing/03-rfcs/20221028-user-stage.md
```diff
@@ -0,0 +1,133 @@
...
+- `Named` => `stage/named/{stage_name}`
```
flaneur2020, line 78: the external stage also got "Named", I guess in the past, both "Internal" and "External" both have a "Named" semantic implicitly

docs/doc/60-contributing/03-rfcs/20221028-user-stage.md
```diff
@@ -0,0 +1,133 @@
...
+
+We can introduce the table stage like snowflake does:
+
+```sql
+COPY INTO my_table FROM @#my_table;
+```
+
+### Cleanup while drop users
+
+Users' stage should be purged while dropping user.
```
flaneur2020, line 133: I guess cleaning up obsoleted user stages in a garbage collection manner is also feasible, as it do not nesscarily need to keep the strong consistency about dropping user & cleaning up the user's stage.

also it might be a time consuming work.

docs/doc/60-contributing/03-rfcs/20221028-user-stage.md
Xuanwo: Ahhh, naming!

Snowflake will have the following stages:

- Internal
- External

And internal could have three types: `named`, `user`, and `table`. The named stage is only used as the internal stage. So it does for the databend.

We don't have an un-named external stage, so I choose to use `Named` instead of `NamedInteral` here.

docs/doc/60-contributing/03-rfcs/20221028-user-stage.md
Xuanwo: I will add this in `Future possibilities`

docs/doc/60-contributing/03-rfcs/20221028-user-stage.md
flaneur2020: how about calls it "NamedInternal" or just keep it "Internal" ? q.q

Xuanwo: LGTM, let's do this.

flaneur2020 (approved):

sundy-li (approved):

Issue #8465, opened 2022-10-26
add distinct value counting to table meta


lichuang:
- [x] #8766
- [x] #8776

Xuanwo:
What's HLL means?

lichuang:
> What's HLL means?

HyperLogLog  https://en.wikipedia.org/wiki/HyperLogLog

Xuanwo:
Thanks for the quick response. So my new question is, what does adding HLL (an algorithm) in table meta mean?

Xuanwo:
I updated the issue title to reflect the real task. Please feel free to update it again @lichuang

PR #8521, opened 2022-10-28
feat: add column ndv

lichuang:

## Summary

feat: add column ndv

Estimate of hll error rate and hll size:

```
    // The error rate of hll, the less error rate, the more hll size
    // Error rate  hll size(bytes)
    // 0.04         1057
    // 0.09         289
    // 0.11         161
```
I choose 0.11 as error rate for error rate and size balance.

Fixes https://github.com/datafuselabs/databend/issues/8465


lichuang:
Still a draft, since the hll crate used does not support `Eq`, i promote a pr to the upstream: https://github.com/alecmocatta/streaming_algorithms/pull/15

after that pr has been merged, i will change hll crate to upstream.

@BohuTANG  and @dantengsky  can have a look at this draft pr.

dantengsky:
> Still a draft, since the hll crate used does not support `Eq`, i promote a pr to the upstream: [alecmocatta/streaming_algorithms#15](https://github.com/alecmocatta/streaming_algorithms/pull/15)
> 
> after that pr has been merged, i will change hll crate to upstream.
> 
> @BohuTANG and @dantengsky can have a look at this draft pr.

does this help?
https://docs.rs/streaming_algorithms/latest/streaming_algorithms/struct.HyperLogLogMagnitude.html

> Like [HyperLogLog](https://docs.rs/streaming_algorithms/latest/streaming_algorithms/struct.HyperLogLog.html) but implements Ord and Eq by using the estimate of the cardinality.

src/query/storages/fuse-meta/src/meta/common.rs
```diff
@@ -37,6 +39,53 @@ pub struct ColumnStatistics {
...
+            hll: Some(HyperLogLog::new(0.00408)),
+        }
+    }
+
+    pub fn add(&mut self, value: &DataValue) {
+        self.hll.as_mut().unwrap().push(value);
+    }
+
+    pub fn hll(&self) -> &HyperLogLog<DataValue> {
+        self.hll.as_ref().unwrap()
```
dantengsky, line 71: ditto

src/query/storages/fuse-meta/src/meta/common.rs
```diff
@@ -37,6 +39,53 @@ pub struct ColumnStatistics {
...
+    }
+
+    pub fn merge(&mut self, other: &ColumnStatistics) {
```
dantengsky, line 74: not sure if `merge` should be a method of ColumnStatistics:
-  this method merges hll only
-  `ColumnStatistics` is more likely to be a value object

src/query/storages/fuse-meta/src/meta/common.rs
```diff
@@ -37,6 +39,53 @@ pub struct ColumnStatistics {
...
+        if let Some(ref hll) = other.hll {
+            self.hll.as_mut().unwrap().union(hll);
+        }
+    }
+
+    pub fn number_of_distinct_values(&self) -> u64 {
+        self.hll.as_ref().unwrap().len() as u64
+    }
+
+    pub fn calc_number_of_distinct_values(&mut self, col: &Arc<dyn Column>) {
```
dantengsky, line 84: ditto. maybe we should put this method in the "column_statistic.rs"

Xuanwo:
@mergify update

Xuanwo:
@lichuang I'm using this PR to test the mergify problem. Sorry for the disturbing and extra work.

src/query/storages/fuse-meta/src/meta/common.rs
lichuang: rename `merge` to `merge_number_of_distinct_values` to be more clearly.

src/query/storages/fuse-meta/src/meta/common.rs
lichuang: `calc_number_of_distinct_values` will change some state in `ColumnStatistics`, so i think as a `ColumnStatistics ` member method is fine.

dantengsky:
> Estimate of hll error rate and hll size:
> 
>     // The error rate of hll, the less error rate, the more hll size
>     // Error rate  hll size(bytes)
>     // 0.04         1057
>     // 0.09         289
>     // 0.11         161
> I choose 0.11 as error rate for error rate and size balance.

I am afraid that we can not afford to keep HLLs this way, even at an error rate of 11%:

For each column of a single block, it takes 161 bytes, 
A block may have thousands of columns,
A compacted segment may have hll state of 1000 blocks.

src/query/storages/fuse-meta/src/meta/common.rs
```diff
@@ -37,6 +42,64 @@ pub struct ColumnStatistics {
...
+        ColumnStatistics {
+            min: DataValue::Null,
+            max: DataValue::Null,
+            null_count: 0,
+            in_memory_size: 0,
+            hll: Some(HyperLogLogMagnitude::new(ColumnStatistics::HLL_ERROR_RATE)),
+        }
+    }
+
+    pub fn add(&mut self, value: &DataValue) {
```
leiysky, line 76: Can we ensure that the value has the correct data type here?

src/query/storages/fuse-meta/src/meta/common.rs
```diff
@@ -37,6 +42,64 @@ pub struct ColumnStatistics {
...
     pub in_memory_size: u64,
+    pub hll: Option<HyperLogLogMagnitude<DataValue>>,
+}
+
+impl ColumnStatistics {
+    // The error rate of hll, the less error rate, the more hll size
+    // Error rate  hll size(bytes)
+    // 0.04         1057
+    // 0.09         289
+    // 0.11         161
```
BohuTANG, line 53: 161 bytes is too big here :)

BohuTANG:
Can you give more background about the ndv design? 
* What is the approximate size used for each column? This is related to the performance(We have done a lot of optimization on the bloom filter index).
* How much performance drop after we add ndv?

dantengsky:
> @lichuang
[fix: move column ndv to TableSnapshot](https://github.com/datafuselabs/databend/pull/8521/commits/05163fe13405a7a793027be5cfac38101a412c0c)

shall we keep the HLL state separately from the snapshot meta file? to another file for example. just keep the **number** of distinct values in the snapshot meta file, since only the numbers are needed during query execution.

BohuTANG:
Please rebase with the main branch, the directory of sotrages has changed.

src/query/storages/fuse/fuse/src/fuse_table.rs
```diff
@@ -388,12 +424,14 @@ impl Table for FuseTable {
         overwrite: bool,
     ) -> Result<()> {
         self.check_mutable()?;
+        let ndvs = calc_column_ndvs_of_data_blocks(&operations)?;
```
dantengsky, line 427: the data blocks referred to by `operations` are `AppendOperationLogEntry`s, not the data blocks appended to table

src/query/storages/fuse/fuse/src/io/locations.rs
```diff
@@ -128,3 +144,21 @@ impl SnapshotLocationCreator for SnapshotVersion {
         }
     }
 }
+
+impl SnapshotLocationCreator for TableSnapshotStatisticsVersion {
+    fn create(&self, id: &Uuid, prefix: impl AsRef<str>) -> String {
+        format!(
+            "{}/{}/{}{}",
+            prefix.as_ref(),
```
dantengsky, line 152: how about adding a dedicated sud dir for the state of statistics? 
currently,  SnapshotsIO assumes that there are only snapshot files under `_ss` dir.

dantengsky:
@lichuang besides the comments in the review, also wondering if it is possible to enable NDV for a non-empty existing table

BohuTANG:
> it is possible to enable NDV for a non-empty existing table

we can let the user do: `create table new_table as select * from old_table` to re-created the ndv index.

src/query/storages/fuse/fuse/src/operations/mutation/block_filter.rs
```diff
@@ -98,6 +99,7 @@ pub async fn delete_from_block(
         whole_block_reader.read_with_block_meta(block_meta).await?
     };
 
+    println!("in delete 8");
```
BohuTANG, line 102: need removed

src/query/storages/fuse/fuse/src/operations/fuse_sink.rs
```diff
@@ -145,6 +150,7 @@ impl Processor for FuseTableSink {
                 output.finish();
             }
             self.state = State::Finished;
+            self.ctx.push_precommit_column_ndvs(self.ndvs.clone());
```
BohuTANG, line 153: If a delete/insert is distributed(for example: insert into t1 select * from t2), how do we calculate the ndv?

lichuang: emm, that is a question, i will check what `push_precommit_block` function has been done and the whole insert\delete process, to see if `push_precommit_column_ndvs ` is collect, wait please.

src/query/storages/fuse/fuse/src/operations/mutation/block_filter.rs
```diff
@@ -34,7 +36,9 @@ pub async fn delete_from_block(
     ctx: &Arc<dyn TableContext>,
     filter_column_proj: Projection,
     filter_expr: &Expression,
+    statistics: &mut Option<TableSnapshotStatistics>,
 ) -> Result<Deletion> {
+    println!("in delete 7");
```
sundy-li, line 41: extra prints.

src/query/storages/table-meta/src/meta/common.rs
```diff
@@ -89,3 +91,62 @@ impl Compression {
...
+            match self.column_hll.get_mut(idx) {
+                Some(hll) => hll.union(other_hll),
+                None => {
+                    self.column_hll.insert(*idx, other_hll.clone());
+                }
+            }
+        }
+    }
+
+    pub fn add_column(&mut self, idx: u32, col: &Arc<dyn Column>) {
```
sundy-li, line 119: Do you have any performance numbers about making all column with HLL indexes ?

I suggest to load large datasets into databend and have the comparisions.

https://databend.rs/doc/learn/analyze-hits-dataset-with-databend

lichuang:
Benchmark result:

I download the hits.tsv for test, and make a 20m test data:
```
for((i=1;i<20;i++));
do
cat hits_1m.tsv >> hits_20m.tsv
done
```

and use `time curl -H "insert_sql:insert into hits format TSV"  -F  "upload=@./hits_20m.tsv"  -XPUT http://user1:abc123@127.0.0.1:8000/v1/streaming_load` for benchmark:

Latest version(v0.8.114): 0.31s user 3.71s system 5% cpu 1:13.46 total
hll version: 0.34s user 3.86s system 5% cpu 1:23.56 total

about 12% performance recession.

And add some counts and statistics file location in `TableSnapshot`:
``
pub struct TableSnapshot {
// ...
    pub column_counts: Option<HashMap<ColumnId, u64>>,

    // table statistics file location
    pub statistics_location: Option<String>,
}
``
make the _ss/snapshot file size from 7.9KB to 19KB.

statistics need to save to a new `_ts` dir, and its size is related to the column number, i have estimate that each column need 7.5KB file size . In hits table, there is 107 column, and the `_ts` file is 777KB size.

@dantengsky @BohuTANG @sundy-li

sundy-li:
Well, to be honest, I'm not sure if this solution is feasible, or it will be refactored immediately later, introducing a compatibility burden.

 1. We will definitely have a histogram index in the future. Based on the histogram, the estimated number of NDV can be generated. The introduction of NDV is currently a temporary solution.

2. So at present we are introducing a lightweight solution before the histogram, and the current implementation based on HLL + error ratio can meet the needs, but the cost performance is relatively low (NDV, which stores HLL files, is not good for insert performance, large index read there is some loss in performance)

3. Based on the `Vec<DataValue> `obtained by column.values(), there is a lot of memory overhead, and the hash of the enum DataValue does not take advantage of the vectorization capability. A better way is to hash the column into Vec<u64> and then calculate the HLL

The current mainstream solution is to store a count value directly, and calculate a density based on `count/rows`. Taking an average of this density can calculate the global average density, and then estimate the global NDV. I prefer this solution. In the case of normal data distribution, the error rate of this scheme will not be lower than that of HLL.


So If let me have the NDV index, I'll 

1. Implement a new aggregate function named `approx_count_distinct(x)` which gives the approximate count of distintinct elements using HyperLogLog.


2. During the insert of fuse data, I will use `eval_aggr(approx_count_distinct)` to generate the ndv value of this column and store it into statistics. This is the same logic like `min`, `max`:
 https://github.com/datafuselabs/databend/blob/3dc4d2235bd4518327b54efc824a26ebb3ccbe40/src/query/storages/fuse/fuse/src/statistics/column_statistic.rs#L45-L46


3. Store the sum of density `ndv/row_count`in segment and snapshot level and we can get the approximate global ndv by `density / block_count`

4.   Ignore the deletion and mutation cause we don't need accurate NDV value.

BohuTANG:
I prefer the `approx_count_distinct ` solution(DuckDB has it), but is it ready to implement in our expression framework now?

sundy-li:
> I prefer the `approx_count_distinct ` solution(DuckDB has it), but is it ready to implement in our expression framework now?

Yes, it's ready like `min`, `max`

Issue #8514, opened 2022-10-28
Feature: `COPY INTO` Idempotent key need elimination strategy


yufan022:
**Summary**
Databend already supports idempotent import. Maybe we should consider the idempotent key elimination strategy to prevent the large-scale import from causing too many keys

BohuTANG:
The copies file key in metasrv have a expired time  cc @lichuang

drmingdrmer:
- #8425

lichuang:
https://github.com/datafuselabs/databend/pull/8524, set the default expire time as 7 day.

PR #8524, opened 2022-10-28
feat: add table copied file key expire time

lichuang:

## Summary

feat: add table copied file key expire time

Fixes https://github.com/datafuselabs/databend/issues/8514

### DIFF
--- a/src/query/service/src/interpreters/interpreter_copy_v2.rs
+++ b/src/query/service/src/interpreters/interpreter_copy_v2.rs
@@ -42,7 +42,7 @@ use crate::sql::plans::Plan;
 use crate::storages::stage::StageTable;
 
 const MAX_QUERY_COPIED_FILES_NUM: usize = 50;
-
+const TABLE_COPIED_FILE_KEY_EXPIRE_AT: Option<u64> = Some(7);
 pub struct CopyInterpreterV2 {
     ctx: Arc<QueryContext>,
     plan: CopyPlanV2,
@@ -87,7 +87,7 @@ impl CopyInterpreterV2 {
         let req = UpsertTableCopiedFileReq {
             table_id,
             file_info: copy_stage_files.clone(),
-            expire_at: None,
+            expire_at: TABLE_COPIED_FILE_KEY_EXPIRE_AT,
         };
         catalog
             .upsert_table_copied_file_info(&tenant, &database_name, req)
--- a/src/query/service/src/interpreters/interpreter_copy_v2.rs
+++ b/src/query/service/src/interpreters/interpreter_copy_v2.rs
@@ -16,6 +16,7 @@ use std::path::Path;
 use std::sync::Arc;
 
 use common_base::base::GlobalIORuntime;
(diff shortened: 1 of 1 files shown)


drmingdrmer (approved):

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -42,7 +42,7 @@ use crate::sql::plans::Plan;
 use crate::storages::stage::StageTable;
 
 const MAX_QUERY_COPIED_FILES_NUM: usize = 50;
-
+const TABLE_COPIED_FILE_KEY_EXPIRE_AT: Option<u64> = Some(7);
```
Xuanwo, line 45: Is there a list of all the magic numbers we are using? How do we prevent the use of duplicated numbers here?

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -87,7 +87,7 @@ impl CopyInterpreterV2 {
         let req = UpsertTableCopiedFileReq {
             table_id,
             file_info: copy_stage_files.clone(),
-            expire_at: None,
+            expire_at: TABLE_COPIED_FILE_KEY_EXPIRE_AT,
```
drmingdrmer, line 90: `expire_at` is the unix timestamp since 1970 in seconds.

drmingdrmer (changes requested):

drmingdrmer:
https://github.com/datafuselabs/databend/blob/06466d607157637035b8848eb52641bbf8737d7e/src/meta/types/src/seq_value.rs#L29

BohuTANG:
Is this PR ready to merge?

src/query/service/src/interpreters/interpreter_copy_v2.rs
lichuang: This magic number limit in the file only, does not exported outside as public.

BohuTANG:
@mergify update

Xuanwo:
@Mergifyio rebase

Xuanwo:
Strange. Why @PsiACE showed up here...

src/query/service/src/interpreters/interpreter_copy_v2.rs
Xuanwo: > This magic number limit in the file only, does not exported outside as public.

This is why it's easier to be broken. We could declare the duplicated numbers.

PsiACE:
> Strange. Why @PsiACE showed up here...

I'm not sure what happened, I didn't even perform any action.

Xuanwo:
@mergify update

src/query/service/src/interpreters/interpreter_copy_v2.rs
lichuang: For now, this magic number is used in this file only, and there is no file to save these const vars.

BohuTANG (approved):

PR #8526, opened 2022-10-28
feat: Add internal and legacy internal stage support

Xuanwo:

## Summary

feat: Add internal and legacy internal stage support


drmingdrmer (approved):

src/meta/proto-conv/tests/it/user_stage.rs
```diff
@@ -745,3 +746,9 @@ fn test_user_stage_s3_v1() -> anyhow::Result<()> {
     common::test_load_old(func_name!(), user_stage_s3_v1.as_slice(), want)?;
     Ok(())
 }
+
+#[test]
+fn test_internal_stage_v17() -> anyhow::Result<()> {
+    common::test_pb_from_to("internal_stage_v17", test_internal_stage_info_v17())?;
+    Ok(())
+}
```
drmingdrmer, line 754: The compatibility test should decode a message from a raw bytes slice, like `test_user_stage_s3_v1()` does

BohuTANG (approved):

flaneur2020 (approved):

drmingdrmer (approved):

Issue #8490, opened 2022-10-27
meta: Cleaning up keys should not affect the state in other key spaces


drmingdrmer:
Clean up several expired keys(e.g. at most hundreds) every time a new log is applied to state-machine.

With openraft-0.7, expired keys can be deleted only when a new log is applied because the meta-store must use the time embedded in a log entry to expire keys, in order to keep consistency across raft nodes.

In a new version of openraft, the leader periodically appends a blank log, so that expired keys will be cleaned at a relatively fixed interval.

Cleaning up keys should not affect the state in other key spaces.I.e., a meta-node that expires keys and a meta-node that does not expire keys should have the same key-value states(the generic-kv key space).

Xuanwo:
Updated issue's title for better reading, feel free to update it again cc @drmingdrmer

PR #8539, opened 2022-10-31
feat(meta): auto-clean expired keys

drmingdrmer:

##### feat(meta): auto-clean expired keys

- Clean up to 32 expired keys before applying a raft-log.

- When modifying(insert/update/delete) a kv record, it emits at most two events:
  - If the existent key expires, emit a `delete` event because an
    expired key is seen and removed.
  - Then, the event of the update follows.

- Fix: #8489
- Fix: #8490
- Fix: #8540

## Related Issues

### DIFF
--- a/src/meta/raft-store/src/state_machine/sm.rs
+++ b/src/meta/raft-store/src/state_machine/sm.rs
@@ -73,11 +73,14 @@ use tracing::info;
 
 use crate::config::RaftConfig;
 use crate::sled_key_spaces::ClientLastResps;
+use crate::sled_key_spaces::Expire;
 use crate::sled_key_spaces::GenericKV;
 use crate::sled_key_spaces::Nodes;
 use crate::sled_key_spaces::Sequences;
 use crate::sled_key_spaces::StateMachineMeta;
 use crate::state_machine::ClientLastRespValue;
+use crate::state_machine::ExpireKey;
+use crate::state_machine::ExpireValue;
 use crate::state_machine::MetaSnapshotId;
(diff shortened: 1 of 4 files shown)


src/meta/raft-store/src/state_machine/sm.rs
```diff
@@ -809,43 +900,109 @@ impl StateMachine {
...
+        }
+
+        // No change, no need to update expiration index
+        if prev == res {
+            return Ok((expired, prev, res));
+        }
+
+        // Remove previous expiration index, add a new one.
+
+        if let Some(sv) = &prev {
```
Xuanwo, line 942: Let's play with `#![feature(let_chains)]`!

```
if let Some(sv) = &prev && let Some(m) = &sv.meta && let Some(exp) = m.expire_at {
    expires.remove(&ExpireKey::new(exp * 1000, sv.seq))?;
}
```

drmingdrmer: Smart guy!

drmingdrmer: Hmm... Clion does not recognize such syntax and refuses to format codes for me.

I'd have to leave it as is.

Xuanwo: Tracked at https://github.com/intellij-rust/intellij-rust/issues/8625

drmingdrmer: image

Xuanwo (approved):

PR #8545, opened 2022-10-31
fix(storage): Only update metrics while read finished

Xuanwo:

## Summary

fix(storage): Only update metrics while read finished

Release note of OpenDAL: https://github.com/datafuselabs/opendal/discussions/898


Xuanwo:
Simple change, let's merge!

PR #8548, opened 2022-10-31
refactor: make block meta easier to be cloned

dantengsky:

## Summary

After reading `SegmentInfo` from storage,  block meta is owned by an `Arc<SegmentInfo>` (which may be cached). 
For situations that need block metas only, like pruning, we have no choice but to clone the block meta, which hurts the performance.

In this PR, changes SegmentInfo from

~~~
pub struct SegmentInfo {
   ....
    pub blocks: Vec<BlockMeta>,
   ....
}
~~~

to

~~~
pub struct SegmentInfo {
   ....
    pub blocks: Vec<Arc<BlockMeta>>,
   ....
}
~~~

and refactor related components accordingly.

Fixes #issue

### DIFF
--- a/src/query/service/tests/it/storages/fuse/operations/mutation/deletion_mutator.rs
+++ b/src/query/service/tests/it/storages/fuse/operations/mutation/deletion_mutator.rs
@@ -58,7 +58,7 @@ async fn test_deletion_mutator_multiple_empty_segments() -> Result<()> {
         // structures are filled with arbitrary values, no effects for this test case
         let block_id = Uuid::new_v4().simple().to_string();
         let location = (block_id, DataBlock::VERSION);
-        let test_block_meta = BlockMeta::new(
+        let test_block_meta = Arc::new(BlockMeta::new(
             1,
             1,
             1,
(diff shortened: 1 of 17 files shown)


Xuanwo:
`BlockMeta` itself is cheap to be cloned. How about move `col_stats` & `col_metas` to `Arc<HashMap>` instead?

dantengsky:
> `BlockMeta` itself is cheap to be cloned. 

according to @sundy-li 's profiling,  cloning of `BlockMeta` do hurt performance and highly likely hurts badly for the large tables.

> How about move `col_stats` & `col_metas` to `Arc<HashMap>` instead?

sounds reasonable to me. let's do the refactor if profiling identifies them to be bottlenecks.

Xuanwo:
> cloning of `BlockMeta` do hurt performance and highly likely hurts badly for the large tables.

Yep. Most cost happened at the clone for hasmap.

sundy-li (approved):

sundy-li:
It works

BohuTANG:
Because we clone BlockMeta so the hashmap is cloned, and with this PR it will avoid.

BohuTANG (approved):

PR #8553, opened 2022-10-31
feat(rbac): only display available roles in SHOW ROLES statement 

### LINKED ISSUE #8420: RBAC: make SHOW ROLES only display the current available roles
**Summary**

Currently the SHOW ROLES statement returns ALL the roles existed in one tenant.

We can limit it only returns the ROLES which granted to the account. There're some scenarios where we need list an user's roles.

related links:

- https://docs.snowflake.com/en/sql-reference/sql/show-roles.html

flaneur2020:

## Summary

- [x] scaffold about SHOW ROLES interpreters/plans etc 
- [x] logic about only display available roles
- [x] add is_current & is_default marker
- [x] tests

Fixes #8420


src/query/service/src/interpreters/interpreter_role_show.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021 Datafuse Labs.
```
Xuanwo, line 1: Happy new year!

src/query/sql/src/planner/binder/binder.rs
```diff
@@ -172,7 +173,7 @@ impl<'a> Binder {
             Statement::AlterUser(stmt) => self.bind_alter_user(stmt).await?,
 
             // Roles
-            Statement::ShowRoles => self.bind_rewrite_to_query(bind_context, "SELECT name, inherited_roles FROM system.roles ORDER BY name", RewriteKind::ShowRoles).await?,
+            Statement::ShowRoles => Plan::ShowRoles(Box::new(ShowRolesPlan {})),
```
Xuanwo, line 176: Can we add more where in the rewrite logic?

flaneur2020: the new SHOW ROLES statement displays some session context aware logic:

1. only displays the available roles for the current session
2. add is_default, is_current field like https://docs.snowflake.com/en/sql-reference/sql/show-roles.html

the session context related infomation may hard to refactor in the system table view :( as the system table would always display the same data without any regard about the session context.

BohuTANG (approved):

Issue #8455, opened 2022-10-25
Add config  meta = "embedded|remote" to specify what meta-store to use


drmingdrmer:

drmingdrmer:
@wubx

Xuanwo:
Maybe `type = "embedded"` and `type = "remote"`?

PR #8571, opened 2022-11-01
feat: add meta_type config

lichuang:

## Summary

feat: add meta_type config

`meta_type` has these types: `remote`, `Fallback` and `embedded`, use `Fallback` as default.
Fallback is used for forward compatbility, that is:
First check embedded config, then endpoints, finally address.

Fixes https://github.com/datafuselabs/databend/issues/8455


src/query/config/src/inner.rs
```diff
@@ -289,11 +291,22 @@ impl Default for MetaConfig {
             auto_sync_interval: 10,
             rpc_tls_meta_server_root_ca_cert: "".to_string(),
             rpc_tls_meta_service_domain_name: "localhost".to_string(),
+            meta_type: MetaType::Remote.to_str().to_string(),
```
Xuanwo, line 294: default to remote will break the existing usage of embedded meta usage.

src/query/config/src/inner.rs
```diff
@@ -289,11 +291,22 @@ impl Default for MetaConfig {
...
         }
     }
 }
 
 impl MetaConfig {
+    pub fn is_embedded_meta(&self) -> Result<bool> {
```
Xuanwo, line 300: Check embedded_dir while converting from `OuterConfig` to `InnerConfig`.

src/query/config/src/outer_v0.rs
```diff
@@ -1306,6 +1336,11 @@ pub struct MetaConfig {
 
     #[clap(long = "meta-rpc-tls-meta-service-domain-name", default_value_t)]
     pub rpc_tls_meta_service_domain_name: String,
+
+    /// MetaStore type: remote or embedded
+    #[clap(long = "meta-type", default_value = MetaType::Remote.to_str())]
+    #[serde(alias = "meta_type")]
```
Xuanwo, line 1342: I prefer `type`.

src/query/config/src/outer_v0.rs
```diff
@@ -1366,6 +1403,7 @@ impl Debug for MetaConfig {
                 "rpc_tls_meta_service_domain_name",
                 &self.rpc_tls_meta_service_domain_name,
             )
+            .field("meta_type", &self.meta_type)
```
Xuanwo, line 1406: Move type as the first field.

src/query/config/src/outer_v0.rs
```diff
@@ -1306,6 +1336,11 @@ pub struct MetaConfig {
 
     #[clap(long = "meta-rpc-tls-meta-service-domain-name", default_value_t)]
     pub rpc_tls_meta_service_domain_name: String,
+
+    /// MetaStore type: remote or embedded
+    #[clap(long = "meta-type", default_value = MetaType::Remote.to_str())]
+    #[serde(alias = "meta_type")]
+    pub meta_type: String,
```
Xuanwo, line 1343: Move this field to the very first of meta config.

drmingdrmer (commented):
It's able to serialize an enum directly(looks good in our case?):

```rust
#[derive(serde::Serialize, serde::Deserialize)]
enum Foo {
    #[serde(rename = "a")]
    A,
    #[serde(rename = "b")]
    B,
}

    let a = Foo::A;
    let s = serde_json::to_string(&a).unwrap();
    println!("{}", s); // "a"

```

src/query/config/src/outer_v0.rs
lichuang: all field in MetaConfig named prefix by `meta-`

src/query/config/src/outer_v0.rs
```diff
@@ -1258,41 +1258,25 @@ impl From<InnerHiveCatalogConfig> for HiveCatalogConfig {
...
 }
 
 /// Meta config group.
 /// TODO(xuanwo): All meta_xxx should be rename to xxx.
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Args)]
 #[serde(default)]
 pub struct MetaConfig {
+    /// MetaStore type: remote or embedded
+    #[clap(long = "meta-type", default_value = "embedded")]
+    #[serde(alias = "meta_type")]
```
Xuanwo, line 1277: This should be `type`. Other wise, user need to input `meta_type="embedded"`

src/query/config/src/outer_v0.rs
```diff
@@ -1353,6 +1332,12 @@ impl TryInto<InnerMetaConfig> for MetaConfig {
     type Error = ErrorCode;
 
     fn try_into(self) -> Result<InnerMetaConfig> {
+        let t: MetaType = serde_json::from_str(&self.meta_type)?;
+        if t == MetaType::Embedded && self.embedded_dir.is_empty() {
```
Xuanwo, line 1336: We can also do the `endpoints` and `address` check here.

src/query/config/src/outer_v0.rs
```diff
@@ -1258,11 +1258,25 @@ impl From<InnerHiveCatalogConfig> for HiveCatalogConfig {
...
+
 /// Meta config group.
 /// TODO(xuanwo): All meta_xxx should be rename to xxx.
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Args)]
 #[serde(default)]
 pub struct MetaConfig {
+    /// MetaStore type: remote or embedded
+    #[clap(long = "meta-type", default_value = "embedded")]
+    #[serde(alias = "meta_type")]
+    pub meta_type: String,
```
drmingdrmer, line 1278: Is it possible to use `pub meta_type: MetaType` to avoid manual conversion?

src/query/config/src/inner.rs
```diff
@@ -289,11 +291,17 @@ impl Default for MetaConfig {
             auto_sync_interval: 10,
             rpc_tls_meta_server_root_ca_cert: "".to_string(),
             rpc_tls_meta_service_domain_name: "localhost".to_string(),
+            meta_type: serde_json::to_string(&MetaType::Embedded).unwrap(),
```
Xuanwo, line 294: Don't use `serde_json::to_string`.

image

drmingdrmer: > Don't use `serde_json::to_string`.

Any suggestions about what to use? 🤔

src/query/config/src/outer_v0.rs
Xuanwo: > Is it possible to use `pub meta_type: MetaType` to avoid manual conversion?

This manual conversion is by design. We will maintain the forward compatibility by implementing `try_into` from the outer config to the inner config. This way, we can do anything to `MetaType` without worrying about breaking our users' config.

drmingdrmer: make sense

src/query/config/src/inner.rs
Xuanwo: Use `MetaType` directly in `InnerMetaConfig` and `String` in `OuterMetaConfig`.

lichuang: I use `strum_macros` in `MetaType` to convert to\from string.

drmingdrmer (approved):

src/query/config/src/inner.rs
```diff
@@ -29,6 +29,7 @@ use common_tracing::Config as LogConfig;
 use common_users::idm_config::IDMConfig;
 
 use super::outer_v0::Config as OuterV0Config;
+use super::outer_v0::MetaType;
```
Xuanwo, line 32: We should not depend on any types added in `outer_v0`. `MetaType` should be declared in `inner`.

src/query/config/src/inner.rs
```diff
@@ -275,6 +276,7 @@ pub struct MetaConfig {
     /// Certificate for client to identify meta rpc serve
     pub rpc_tls_meta_server_root_ca_cert: String,
     pub rpc_tls_meta_service_domain_name: String,
+    pub meta_type: String,
```
Xuanwo, line 279: I prefer to use `MetaType` here.

Xuanwo (approved):

PR #8590, opened 2022-11-01
refactor: compact segments

dantengsky:

## Summary

new implementation of `compact segments`

- which strictly preserves the order of ingestion

  to avoid the `ripple effects` which may lead to a large number of consecutive compacted segments being split and merged again and again, for even a single fragmented segment, during compaction, segments of size [threshold, 2 * threshold) are allowed to be generated.
  
- the "semantic" of the statement optimize table t compact segment `limit n` adjusted to the max number of fragmented segments to be compacted
  note that at the end of compaction, if there are fragmented segments left (the combined size of them should be lesser than the threshold), they will be compacted into a segment, those segments are not counted for the `limit`.

Fixes #issue


src/query/storages/fuse/fuse/src/operations/mutation/compact_mutator/segment_compact_mutator.rs
```diff
@@ -237,10 +163,133 @@ impl TableMutator for SegmentCompactMutator {
...
+            threshold,
+            accumulated_num_blocks: 0,
+            fragmented_segments: vec![],
+            segment_writer,
+            compacted_state: Default::default(),
+        }
+    }
+
+    // accumulate one segment
+    pub async fn add(&mut self, segment_info: &'a SegmentInfo, location: Location) -> Result<()> {
```
zhyass, line 210: So we cannot ensure the exact number of blocks in a segment is   threshold.

dantengsky: u r right! the reason behind this:

>to avoid the ripple effects which may lead to a large number of consecutive compacted segments being split and merged again and again, for even a single fragmented segment, during compaction, segments of size [threshold, 2 * threshold) are allowed to be generated.

zhyass (approved):
LGTM

PR #8596, opened 2022-11-02
fix(planner): Fix offset statement error handling

### LINKED ISSUE #8588: `SELECT * FROM xyzw OFFSET 1 + y` should be checked during planner bind
**Summary**

Our logic test has a strange case (which expects a logic error):

```rust
SELECT * FROM xyzw OFFSET 1 + y;
```

We will got error at:

https://github.com/datafuselabs/databend/blob/faeeff76a37d371bc6a8246f18b740338167b3e0/src/query/service/src/pipelines/processors/transforms/transform_limit.rs#L30-L45

Maybe we should check it during planner bind stage?

---

Also:

```shell
SELECT * FROM xyzw OFFSET -100;
SELECT * FROM xyzw ORDER BY x OFFSET 1 + 0.0;
```

leiysky:

## Summary

Close #8588


Xuanwo (approved):

Xuanwo:
Please also update the logic test

https://github.com/datafuselabs/databend/blob/main/tests/logictest/suites/crdb/select#L163-L165

Xuanwo:
```shell
-- Wait for https://github.com/datafuselabs/databend/issues/8588
-- statement error 1015
-- SELECT * FROM xyzw OFFSET -100;
```

THis case is not fixed?

Xuanwo (approved):

sundy-li (approved):

Issue #8573, opened 2022-11-01
idea: Adopt async-backtrace for efficiently track and view the state of asynchronous tasks


Xuanwo:
**Summary**

Reference: https://tokio.rs/blog/2022-10-announcing-async-backtrace

We can start with applying them for our pipelines

jswrenn:
I'm the maintainer of `async-backtrace`. If anyone picks up this issue, I'd love to hear how it goes. The `async-backtrace` is early in its development, and input from early-adopters is essential for shaping the future of the crate.

Xuanwo:
Let me build the simplest example so others can explore `async-backtrace` in the databend~

Xuanwo:
Hi, @jswrenn, I have added the basic integration with async-backtrace in https://github.com/datafuselabs/databend/issues/8573.

In that PR, I add `#[async_backtrace::framed]` for our async IO functions, and add a new HTTP handler `/debug/async_backtrace` to get the output.

However, I can't get the expected async_backtrace like in the example.

- No changes in `tokio-console` (maybe I missed something?)
- HTTP GET `/debug/async_backtrace` always returns empty.

Questions:

- Is this usage matches your expectation?
- Can we have a `level` idea on async backtrace? (so we can reduce some cost in critical path)

Xuanwo:
Implemented by @zhang2014

PR #8597, opened 2022-11-02
feat: Basic integration with async_backtrace

Xuanwo:

## Summary

feat: Basic integration with async_backtrace

Related to https://github.com/datafuselabs/databend/issues/8573


src/common/http/src/debug/async_backtrace.rs
```diff
@@ -0,0 +1,25 @@
...
+use async_backtrace;
+use poem::IntoResponse;
+
+// run async-backtrace
+// example: /debug/async_backtrace
+// req query contains async_backtrace information
+
+#[poem::handler]
+pub async fn debug_async_backtrace_handler() -> impl IntoResponse {
+    async_backtrace::taskdump_tree(true)
```
jswrenn, line 24: To reduce the cost of taskdumping, use:
```suggestion
    async_backtrace::taskdump_tree(false)
```

By doing so, `taskdump_tree` won't wait for currently-running tasks to suspend so it can print their state (it'll just skip them).

jswrenn (commented):
In https://github.com/datafuselabs/databend/issues/8573#issuecomment-1299586592 you write:
> HTTP GET /debug/async_backtrace always returns empty.

This will occur if no other `#[async_backtrace::framed]` tasks are suspended when `taskdump_tree` is invoked. Double-check that the application is under load when `/debug/async_backtrace` is invoked.

Also consider applying `#[async_backtrace::framed]` much more widely. It only imposes a meaningful (but still quite small) overhead at the root of the futures tree it's applied to, so annotating sub-invocated async fns doesn't do any harm.

For instance, here:
```rust
    async fn read(&self, path: &str, args: OpRead) -> Result<BytesReader> {
        let op = self.inner.clone();
        let path = path.to_string();
        self.runtime
            .spawn(async move { op.read(&path, args).await })
            .await
            .expect("join must success")
    }
```

You should ensure the `async` block is included in stack traces:
```rust
    async fn read(&self, path: &str, args: OpRead) -> Result<BytesReader> {
        let op = self.inner.clone();
        let path = path.to_string();
        self.runtime
            .spawn(async_backtrace::frame!(async move { op.read(&path, args).await }))
            .await
            .expect("join must success")
    }
```

...and that inner async fns like `Accessor::read` are also included in async backtraces.

Xuanwo:
Thanks for the tips, I got an async-backtrace now!

image

Let me add more.

Xuanwo:
@mergify update

PR #8606, opened 2022-11-02
feat(TSV): allow nested field in TSV.

youngsofun:

### 1. nested field in TSV in stead of escape the whole field as a TSV string, to be more readable and still parsable.

like in clickhouse:

```
$ curl  -XPOST "http://localhost:8123" -d "SELECT array('b\tb', 'a'), 'b\tb' format TSV"
['b\tb','a']    b\tb
```

### 2. add an important test to check that databend can load the file it unloaded.
--
1. for main types (extend more type later)
2. for common corner cases in values. (add more later)
3. for default config of formats. (format with options should pass this test too later)

also [test: add stateful test load_unload.sh](https://github.com/datafuselabs/databend/commit/6765a79ec6fedffbb8960703dd8ac32dc348bd39)

ndjson still need some work for variant and nested types


tests/suites/1_stateful/05_input_formats/05_0003_load_unload_all.result
```diff
@@ -0,0 +1,8 @@
+---CSV
+"a""b",1.0,"['a""b']","{""k"":""v""}","2044-05-06 10:25:02.868894"
```
b41sh, line 2: Why `\"` is shown as `""`. It looks a little strange.

youngsofun: most CSV  use `""` instead of `\"` by default.

b41sh (approved):

youngsofun:
rebased last commit just to fix a typo.

PR #8503, opened 2022-10-27
feat: create catalog by sql

ClSlaid:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Try to implement `CREATE CATALOG` SQL in databend-query

PR #8620, opened 2022-11-03
feat: multiple catalog create (planner and catalog manager)

ClSlaid:

## Summary

This is the planner implementation and the catalog manager refactor part of the Multiple Catalog

implemented statements:
- `show catalogs`

  图片

- `create catalog`

```
MySQL root@127.0.0.1:(none)> CREATE CATALOG ctl TYPE=DEFAULT CONNECTION=(cnn='cnn');
(1105, 'Code: 2318, displayText = Creating default catalog is not allowed!.')
MySQL root@127.0.0.1:(none)> CREATE CATALOG ctl TYPE=HIVE CONNECTION=(cnn='cnn');
(1105, 'Code: 2004, displayText = Expected HMS_ADDRESS.')
MySQL root@127.0.0.1:(none)> CREATE CATALOG ctl TYPE=HIVE CONNECTION=(HMS_ADDRESS='cnn');
Query OK, 0 rows affected
Time: 0.022s
MySQL root@127.0.0.1:(none)> show catalogs;
+---------+
| catalog |
+---------+
| ctl     |
| default |
| hive    |
+---------+
3 rows in set
Time: 0.030s
MySQL root@127.0.0.1:(none)> CREATE CATALOG ctl TYPE=HIVE CONNECTION=(HMS_ADDRESS='cnn');
(1105, 'Code: 2319, displayText = Catalog ctl already exists.')
```

Part of #8503


src/common/exception/src/exception_code.rs
```diff
@@ -192,6 +192,8 @@ build_exceptions! {
     DropDbWithDropTime(2315),
     UndropDbWithNoDropTime(2316),
     TxnRetryMaxTimes(2317),
+    CreateUnsupportedCatalog(2318),
```
Xuanwo, line 195: Keep the same error naming style.

I prefer to use `CataglogNotSupported`.

Also, please add a doc on what this error means and when to use this error.

src/meta/app/src/schema/catalog.rs
```diff
@@ -28,3 +32,49 @@ impl Display for CatalogType {
...
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CatalogMeta {
+    pub catalog_type: CatalogType,
+    pub options: BTreeMap<String, String>,
+    pub created_on: DateTime<Utc>,
+}
+
+impl Display for CatalogMeta {
```
Xuanwo, line 43: Why do we need to implement `Display`? And it's content seems the same with `Debug`.

src/query/service/src/catalogs/catalog_manager.rs
```diff
@@ -80,4 +87,47 @@ impl CatalogManagerHelper for CatalogManager {
...
+                #[cfg(feature = "hive")]
+                {
+                    let catalog_options = req.meta.options;
+                    let hms_address = catalog_options
+                        .get("hms_address")
+                        .ok_or_else(|| ErrorCode::InvalidArgument("Expected HMS_ADDRESS"))?;
+                    let catalog: Arc<dyn Catalog> = Arc::new(HiveCatalog::try_create(hms_address)?);
+                    let ctl_name = req.name_ident.catalog_name.clone();
+                    match self.catalogs.entry(ctl_name.clone()) {
+                        dashmap::mapref::entry::Entry::Vacant(v) => {
```
Xuanwo, line 115: Too deep in dashmap. Why not `get_mut` instead?

src/query/service/src/catalogs/catalog_manager.rs
```diff
@@ -80,4 +87,47 @@ impl CatalogManagerHelper for CatalogManager {
...
+                {
+                    Err(ErrorCode::CreateUnsupportedCatalog(
+                        "Hive catalog support is not enabled in your databend-query distribution.",
+                    ))
+                }
+                #[cfg(feature = "hive")]
+                {
+                    let catalog_options = req.meta.options;
+                    let hms_address = catalog_options
+                        .get("hms_address")
```
Xuanwo, line 110: I prefer to use `address`.

src/query/service/src/interpreters/interpreter_catalog_create.rs
```diff
@@ -0,0 +1,77 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+#[cfg(feature = "hive")]
```
Xuanwo, line 17: Maybe we can always use `CatalogManager` here to avoid check hive feature every time.

src/query/sql/src/planner/binder/ddl/catalog.rs
```diff
@@ -0,0 +1,130 @@
...
+
+impl<'a> Binder {
+    pub(in crate::planner::binder) async fn bind_show_catalogs(
+        &mut self,
+        bind_context: &BindContext,
+        stmt: &ShowCatalogsStmt<'a>,
+    ) -> Result<Plan> {
+        let ShowCatalogsStmt { limit } = stmt;
+        let mut query = String::new();
+        write!(query, "SELECT name AS Catalog FROM system.catalogs").unwrap();
```
Xuanwo, line 49:
```suggestion
        write!(query, "SELECT name AS Catalogs FROM system.catalogs").unwrap();
```

src/query/service/src/catalogs/catalog_manager.rs
ClSlaid: Concurrent writings will occur here, so this needs to be carefully dealt with.

DashMap organizes data with different `shard`s, and all reading and writings happen on those `shards`. Like a vector of `RwLock`s.

Both `DashMap::get_mut` and `DashMap::entry` will acquire a write lock on the shard, but if the entry happened to be vacant:

- `DashMap::get_mut` will return `None` and then the shard is `unlocked`, see [this](https://docs.rs/dashmap/latest/src/dashmap/lib.rs.html#986-1006) for detail;
- `DashMap::entry` will return a `VacantEntry`, which occupies the write lock on the shard, see [this](https://docs.rs/dashmap/latest/src/dashmap/lib.rs.html#1105-1121) and [this](https://docs.rs/dashmap/latest/src/dashmap/mapref/entry.rs.html#87-90) for detail;

Surely you can see, using `DashMap::get_mut` while ensuring concurrent safety is more complex than `DashMap::entry`.

src/meta/app/src/schema/catalog.rs
ClSlaid: To align with other structs in `meta`

Xuanwo: cc @drmingdrmer for comments.

I prefer to remove not needed functions instead of `align`.

src/query/service/src/catalogs/catalog_manager.rs
Xuanwo: Lesson learned. Please wrap this function in `CatalogManager` with the detailed explanation so that we don't need to learn this every time.

src/meta/app/src/schema/catalog.rs
drmingdrmer: Agree. Add only what you need.

BohuTANG (commented):
👍

BohuTANG (approved):

src/meta/app/src/schema/catalog.rs
```diff
@@ -12,8 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::BTreeMap;
 use std::fmt::Display;
 
+use common_datavalues::chrono::DateTime;
+use common_datavalues::chrono::Utc;
```
Xuanwo, line 19: I prefer to introduce `chrono` directly.

Xuanwo (commented):
Mostly LGTM!

Xuanwo (approved):

PR #8626, opened 2022-11-03
refactor(planner): use const `MAX_BLOCK_SIZE` in join probe

xudong963:

## Summary

Based on our testing and user scenarios, change `max_block_size` in join to 65535 by using const `MAX_BLOCK_SIZE`.

Fixes #issue


src/query/service/src/pipelines/processors/transforms/hash_join/desc.rs
```diff
@@ -30,6 +30,8 @@ use crate::sql::evaluator::Evaluator;
 use crate::sql::executor::HashJoin;
 use crate::sql::plans::JoinType;
 
+pub const MAX_BLOCK_SIZE: usize = 65535;
```
BohuTANG, line 33: why not set the `get_max_block_size` default to 65535?

BohuTANG: https://github.com/datafuselabs/databend/blob/ac98a1b608c4b23f30ca7e14a90fad652395d059/src/query/settings/src/lib.rs#L124-L133

I think you should test the number table with 65535~

xudong963: `get_max_block_size` is also used by `insert`. If we change it to 65535 as default, may influent query performance.

For join scenarios, we've proven that `65535` has performance gains.

sundy-li: we should rename it to `JOIN_BATCH_SIZE` to avoid confusion.

BohuTANG (approved):

src/query/service/src/pipelines/processors/transforms/hash_join/desc.rs
```diff
@@ -24,12 +21,13 @@ use common_sql::IndexType;
 use parking_lot::RwLock;
 
 use crate::pipelines::processors::transforms::hash_join::row::RowPtr;
-use crate::sessions::QueryContext;
 use crate::sql::evaluator::EvalNode;
 use crate::sql::evaluator::Evaluator;
 use crate::sql::executor::HashJoin;
 use crate::sql::plans::JoinType;
 
+pub const MAX_BLOCK_SIZE: usize = 65535;
```
BohuTANG, line 29: Better to `JOIN_MAX_BLOCK_SIZE`?

xudong963: OK

sundy-li (approved):

PR #8628, opened 2022-11-03
feat: copy files order by last modified time asc

BohuTANG:

## Summary

Copy files in asc order of file last modification time.

Closes #issue

### DIFF
--- a/src/query/service/src/interpreters/interpreter_copy_v2.rs
+++ b/src/query/service/src/interpreters/interpreter_copy_v2.rs
@@ -281,11 +281,17 @@ impl CopyInterpreterV2 {
         from: &ReadDataSourcePlan,
         copy_stage_files: BTreeMap<String, TableCopiedFileInfo>,
     ) -> Result<PipelineBuildResult> {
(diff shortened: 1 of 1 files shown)


Xuanwo:
Why we need this change? After stream copy implemented, we can't do sort like this anymore.

BohuTANG:
> Why we need this change? After stream copy implemented, we can't do sort like this anymore.

If we have many files need to copy, let the oldest files fisrt, we can keep the insert order same as the file modified order.

PsiACE:
It feels possible to add a switch for it if necessary.

PsiACE:
> We can't sort the files if there are 10W files, we will copy them in stream which will in the order returned by list (in the near future).

So I suggest leaving it up to the user to decide whether to use this change

```sql
set order_by_last_modified_time = 1
```

BohuTANG:
1. Each file size is larger, for example, 100GB
2. We have a lot of these files like 1), and it's being produced all the time 
3. The order is important, let's the oldest copy first.

Issue #8311, opened 2022-10-19
further compact blocks when insert/load data with many threads


youngsofun:
**Summary**

already compacted in each threads.

youngsofun:
```
compactor----------->SinkBigBlock-------------------+
                                                    |
                                                    |
                                                    v
                                                 compactor--------->Sink
                                                    ^
                                                    |
                                                    |
compactor----------->SinkBigBlock-------------------+
```

Sink to encode block to parquet and write to s3.

this is already used in unloading to file.
we can extract this as a shared code (`Sink` as a type parameter)


@sundy-li  @zhyass @dantengsky

youngsofun:
related

https://github.com/datafuselabs/databend/issues/7760

https://github.com/datafuselabs/databend/issues/8353

youngsofun:
because a processor has only one output,  have to choose 1 of the 2:

1. merge compactor into Sink ( call compactor inside)
2. add a flag to the DataBlock indicating whether to SinkBigBlock should process it or pass it to downstream

sundy-li:
> because a processor has only one output

why not use resize the processor into one ?

youngsofun:
@sundy-li  I mean SinkBigBlock do not have information for branching: sink at once or pass it to downstream (compactor and final sink, resized to 1)

PR #8644, opened 2022-11-05
feat(compact): optimize compact for data load.

youngsofun:

## Summary

an improved and simpler version of https://github.com/datafuselabs/databend/issues/8311

simply resize to 1 and resize back to max_threads is enough: new BlockCompactorNoSplit make sure big block pass through it to downstream(Sink) fast.

The premise is that  DeserializerProcessor in multi threads already try to accumulate big Block, before output to the compactor.

other optimize:
-   call block.memory_size() only once.
-   avoid call concat_blocks for each new block

refactor:

- get real BlockCompactThresholds  from dest table

Closes https://github.com/datafuselabs/databend/issues/8311

### DIFF
--- a/src/query/catalog/src/table.rs
+++ b/src/query/catalog/src/table.rs
@@ -18,6 +18,7 @@ use std::sync::Arc;
 
 use chrono::DateTime;
 use chrono::Utc;
+use common_datablocks::BlockCompactThresholds;
 use common_datablocks::DataBlock;
 use common_datavalues::chrono;
 use common_datavalues::DataSchemaRef;
@@ -220,6 +221,14 @@ pub trait Table: Sync + Send {
             self.name(),
             self.get_table_info().engine(),
         )))
+    }
+
+    fn get_block_compact_thresholds(&self) -> BlockCompactThresholds {
+        BlockCompactThresholds {
+            max_rows_per_block: 1000 * 1000,
+            min_rows_per_block: 800 * 1000,
+            max_bytes_per_block: 100 * 1024 * 1024,
+        }
     }
 
     async fn compact(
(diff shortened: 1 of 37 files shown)


youngsofun:
this also makes it easier if we want to copy to commit by file later.
because we can maintain blocks for all reading files in one map.

BohuTANG:
Some issues with this PR need to be addressed:
```
Following failed statements:
---------------------------------------------
Runner: mysql
ErrorType: statement ok execute with exception
Message: Failed to execute. Collected info: 1105 (HY000): Code: 1068, displayText = Cannot join handle from context's runtime, cause: task 14675 panicked.

Parsed Statement
    at_line: 22,
    s_type: Statement: ok, type: None,
    suite_name: base/03_dml/03_0028_copy_into_stage,
    text:
        copy into test_table from @test;
    results: [],
    runs_on: {'mysql', 'clickhouse', 'http'},
---------------------------------------------
Runner: mysql
ErrorType: statement ok execute with exception
Message: Failed to execute. Collected info: 1105 (HY000): Code: 1068, displayText = Cannot join handle from context's runtime, cause: task 15186 panicked.
```

https://github.com/datafuselabs/databend/actions/runs/3398563740/jobs/5651782219#step:4:2092

src/query/catalog/src/table.rs
```diff
@@ -162,6 +163,7 @@ pub trait Table: Sync + Send {
         ctx: Arc<dyn TableContext>,
         pipeline: &mut Pipeline,
         need_output: bool,
+        _is_ingest: bool,
```
BohuTANG, line 166: Prefer to add an argument instead the `is_ingest`:
```
pub enum AppendMode {
    // From INSERT operation
    Normal,
    // From COPY operation
    Copy,
}
```

youngsofun: done

youngsofun:
the effect is demonstrated in the test. https://github.com/datafuselabs/databend/pull/8644/commits/e0dee7f3113230144c3036f1fdd415330caea861

BohuTANG:
From my `COPY`  test, this PR works well 👍 :
image

BohuTANG (approved):

PR #8662, opened 2022-11-06
refactor(catalog): remove delete

BohuTANG:

## Summary

Remove the DeletePlan from catalog, and decouple the catalog from the plan:
```
    async fn delete(
        &self,
        ctx: Arc<dyn TableContext>,
        projection: &Projection,
        selection: &Option<String>,
    ) -> Result<()> {
```

Closes #issue


BohuTANG:
```
/// # TODO
///
/// From @xuanwo
///
/// Ideally, we need to use `Scalar` in DeletePlan.selection. But we met a
/// cycle deps here. So we have to change `selection` in String first, and
/// change into `Scalar` when our `Planner` has been moved out.
///
/// At this stage, DeletePlan's selection expr will be parsed twice:
///
/// - Parsed during `bind` to get column index and projection index.
/// - Parsed during `execution` to get the correct columns
///
/// It's an ugly but necessary price to pay. Without this, we would sink in
/// hell forever.
```

I think this TODO can TO-DO after this PR is merged, cc @Xuanwo

BohuTANG:
@Xuanwo ,
I finally know the situation on the `DeletePlan.selection`, it's one step from the `old-expression` to the `new-expression`, now it's hard to do it, so lets the query&planner team do it :)

cc @leiysky @sundy-li

PR #8673, opened 2022-11-07
refactor(copy): try move list files to read_partitions

BohuTANG:

## Summary

*  move list files from interpreter to `read_partitions`, first part of the distributed COPY

Closes #issue

### DIFF
--- a/src/query/catalog/src/plan/pushdown.rs
+++ b/src/query/catalog/src/plan/pushdown.rs
@@ -14,6 +14,8 @@
 
 use std::fmt::Debug;
 
+use common_meta_types::UserStageInfo;
+
 use crate::plan::Expression;
 use crate::plan::Projection;
 
@@ -27,6 +29,13 @@ pub struct PrewhereInfo {
     pub remain_columns: Projection,
     /// filter for prewhere
     pub filter: Expression,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct CopyInfo {
+    pub files: Vec<String>,
+    pub pattern: String,
+    pub stage_info: UserStageInfo,
 }
 
 /// Extras is a wrapper for push down items.
@@ -44,4 +53,6 @@ pub struct PushDownInfo {
     pub limit: Option<usize>,
     /// Optional order_by expression plan, asc, null_first
     pub order_by: Vec<(Expression, bool, bool)>,
+    /// Optional copy info, used for COPY into <table> from stage
+    pub copy: Option<CopyInfo>,
 }
--- a/src/query/service/tests/it/storages/fuse/operations/read_plan.rs
+++ b/src/query/service/tests/it/storages/fuse/operations/read_plan.rs
@@ -131,6 +131,7 @@ fn test_to_partitions() -> Result<()> {
         limit: None,
         order_by: vec![],
         prewhere: None,
+        copy: None,
     });
 
     let (stats, parts) = FuseTable::to_partitions(&blocks_metas, &column_leafs, push_down);
@@ -171,6 +172,7 @@ async fn test_fuse_table_exact_statistic() -> Result<()> {
             prewhere: None,
             limit: None,
             order_by: vec![],
+            copy: None,
         };
         let (stats, parts) = table.read_partitions(ctx.clone(), Some(push_downs)).await?;
         assert_eq!(stats.read_rows, num_blocks * rows_per_block);
--- a/src/query/service/tests/it/storages/memory.rs
+++ b/src/query/service/tests/it/storages/memory.rs
@@ -81,6 +81,7 @@ async fn test_memorytable() -> Result<()> {
                             limit: None,
                             order_by: vec![],
                             prewhere: None,
+                            copy: None,
                         }
                     })
                 })
--- a/src/query/sql/src/executor/physical_plan_builder.rs
+++ b/src/query/sql/src/executor/physical_plan_builder.rs
@@ -598,6 +598,7 @@ impl PhysicalPlanBuilder {
             prewhere: prewhere_info,
             limit: scan.limit,
             order_by: order_by.unwrap_or_default(),
+            copy: None,
         })
     }
 }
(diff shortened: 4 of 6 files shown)


BohuTANG:
@youngsofun 

I need your help here :)
This PR reads the files from `read_partitions` and want to read data in `read_data`.
`read_data` is a sync method. How can I build `InputContext` in the `read_data` function? It's an async method.
https://github.com/datafuselabs/databend/blob/a0c388c47c3d6b9cf142bbfd9d9ca8dbf2c71b18/src/query/storages/stage/src/stage_table.rs#L102-L112

youngsofun:
@BohuTANG why have to mv the building of InputContext out of `read_partitions`?

BohuTANG:
> @BohuTANG why have to mv the building of InputContext out of `read_partitions`?

This PR is preparing for distributed COPY, the coordinate databend-query node will get all the files needed to be copied from `read_partitions`, and then do the distributed plan, the `read_partitions` is not for this case.

youngsofun:
try_create_from_copy need to be async because it read infos about files,  and split large files.

```
        let splits = format
            .get_splits(&plan, &operator, &settings, &schema)
            .await?;
```

we can extract this step and remain it in read_partition.
read_partition should return a list of `split` instead of paths as items in `partitions`, 
so large files can be distributed uniformly.
`split` was designed with this in mind from the beginning.

otherwise, you may need even more refactoring with InputContext.

BohuTANG:
Distribution COPY mainly resolve the case is that many not large files(such as 100MB raw file or small compressed file), not for one big large file(this is not a usually case for the user).

youngsofun:
no direct function for it.

you can refer to the body of `try_create_from_copy`

splits is just a field of InputContext

BohuTANG:
> no direct function for it.
> 
> you can refer to the body of `try_create_from_copy`
> 
> splits is just a field of InputContext

Can I use `GlobalIORuntime::instance().spawn` in `read_data` to build `InputContext`?

BohuTANG:
Some tests failure, I will address them this afternoon.

zhang2014 (approved):

tests/suites/1_stateful/00_copy/00_0000_copy_from_stage.sh
```diff
@@ -63,7 +63,7 @@ copy_from_named_external_stage_cases=(
   # copy bz2 csv
   "copy into ontime200 from @named_external_stage FILES = ('ontime_200.csv.bz2') FILE_FORMAT = (type = 'CSV' field_delimiter = ',' compression = 'bz2'  record_delimiter = '\n' skip_header = 1);"
   # copy auto csv
-  "copy into ontime200 from @named_external_stage FILES = ('ontime_200.csv.gz','ontime_200.csv.bz2','ontime_200.csv.zst') FILE_FORMAT = (type = 'CSV' field_delimiter = ',' compression = 'auto'  record_delimiter = '\n' skip_header = 1);"
+  "copy into ontime200 from @named_external_stage FILES = ('ontime_200.csv.gz','ontime_200.csv.bz2','ontime_200.csv.zst') FILE_FORMAT = (type = 'CSV' field_delimiter = ',' compression = 'auto'  record_delimiter = '\n' skip_header = 1) FORCE = true;"
```
BohuTANG, line 66: This test had no effects before because all files have been copied, they will skip by the copied file check. cc @Xuanwo

Xuanwo: Nice catch.

src/query/catalog/src/plan/pushdown.rs
```diff
@@ -29,6 +31,18 @@ pub struct PrewhereInfo {
     pub filter: Expression,
 }
 
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct StageInfo {
```
Xuanwo, line 35: About `StageContext`? And I prefer to rename `UserStageInfo` into `StageInfo` instead.

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
```
Xuanwo, line 1: this use is on the top of file.

src/query/service/src/interpreters/interpreter_copy_v2.rs
```diff
@@ -57,36 +62,71 @@ impl CopyInterpreterV2 {
...
+        let select_interpreter = SelectInterpreterV2::try_create(
+            self.ctx.clone(),
+            *(bind_context.clone()),
+            *s_expr.clone(),
+            metadata.clone(),
+        )?;
 
-        Ok(())
+        // Building data schema from bind_context columns
+        // TODO(leiyskey): Extract the following logic as new API of BindContext.
```
Xuanwo, line 90: cc @leiysky, how about make this an issue?

src/query/service/src/interpreters/interpreter_list.rs
```diff
@@ -64,11 +65,8 @@ impl Interpreter for ListInterpreter {
                     &plan.pattern, e
                 ))
             })?;
-
+            files.retain(|v| regex.is_match(&v.path));
```
Xuanwo, line 68: Nice catch!

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
```diff
@@ -182,7 +182,7 @@ pub trait InputFormatPipe: Sized + Send + 'static {
                     break;
                 };
             }
-            tracing::info!("end copy splits feeder");
+            tracing::debug!("end copy splits feeder");
```
Xuanwo, line 185: Use `debug` instead of `tracing::debug`.

src/query/catalog/src/plan/pushdown.rs
BohuTANG: This is used for pushdown, I changed the name to `StagePushDownInfo` but not `StageContext`.
BTW, `UserStageInfo` is better than `StageInfo` here, because in distributed execute system, there is a `Stage`, `StageInfo` may be confused for them.

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
BohuTANG: Let's keep the style for input_formants.

Xuanwo (approved):

PR #8676, opened 2022-11-07
feat: add inlist expr in new expression

sundy-li:

## Summary

Summary about this PR

Closes #issue


src/query/functions-v2/tests/it/scalars/testdata/control.txt
```diff
@@ -275,3 +275,78 @@ evaluation (internal):
...
+checked expr   : or<Boolean, Boolean>(eq<Boolean, Boolean>(false, false), eq<Boolean, Boolean>(false, true))
+optimized expr : true
+output type    : Boolean
+output domain  : Unknown
+output         : true
+
+
+ast            : int8_col not in (1, 2, 3, 4, 5, null)
+raw expr       : not(in(ColumnRef(0)::Int8, array(1_u8, 2_u8, 3_u8, 4_u8, 5_u8)))
+checked expr   : not<Boolean>(in<Int16, Array(Int16)>(CAST(ColumnRef(0) AS Int16), CAST(array<T0=UInt8><T0, T0, T0, T0, T0>(1_u8, 2_u8, 3_u8, 4_u8, 5_u8) AS Array(Int16))))
```
sundy-li, line 298: why did it  turne into `Int16`? @andylokandy

src/query/functions-v2/src/scalars/control.rs
```diff
@@ -173,4 +186,108 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    fn cal_in<T: ArgType>(
+        lhs: ValueRef<T>,
+        rhs: ValueRef<ArrayType<T>>,
+    ) -> Result<Value<BooleanType>, String>
+    where
+        T::Scalar: HashtableKeyable,
+    {
+        // must be scalar in rhs, if rhs contains column, it'll be rewrited to `a = b1 or a = b2 or a = b3` in type_checker
+        let array = rhs.as_scalar().unwrap();
+        let mut set = StackHashSet::<_, 128>::with_capacity(T::column_len(array));
```
RinChanNOWWW, line 199: We use a global hashset in v1. Can we achieve this in v2?

sundy-li: Yes, it's not in high-priority, we plan to do it after new expression is done

src/query/functions-v2/src/scalars/control.rs
```diff
@@ -173,4 +186,108 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            "in",
+            FunctionProperty::default(),
+            |_, _| None,
+            |lhs, rhs, _| cal_in(lhs, rhs)
+        );
+
+    registry.register_passthrough_nullable_2_arg::<BooleanType, ArrayType<BooleanType>, BooleanType, _, _>(
+        "in",
+        FunctionProperty::default(),
+        |_, _| None,
```
andylokandy, line 246: The domain calculation should be simple for this case.

src/query/functions-v2/src/scalars/control.rs
```diff
@@ -173,4 +186,108 @@ pub fn register(registry: &mut FunctionRegistry) {
...
     );
+
+    fn cal_in<T: ArgType>(
+        lhs: ValueRef<T>,
+        rhs: ValueRef<ArrayType<T>>,
+    ) -> Result<Value<BooleanType>, String>
+    where
+        T::Scalar: HashtableKeyable,
+    {
+        // must be scalar in rhs, if rhs contains column, it'll be rewrited to `a = b1 or a = b2 or a = b3` in type_checker
```
andylokandy, line 197: I think the assumption doesn't apply when the function is called explicitly by the user.

src/query/functions-v2/src/scalars/control.rs
```diff
@@ -173,4 +186,108 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                    FunctionProperty::default(),
+                    |_, _| None,
+                    |lhs, rhs, _| cal_in(lhs, rhs)
+                );
+            }
+        });
+    }
+
+    registry
+        .register_passthrough_nullable_2_arg::<DateType, ArrayType<DateType>, BooleanType, _, _>(
```
andylokandy, line 229: do you think that we need to implement `in(GenericType<0>, ArrayType<GenericType<0>>) :: BooleanType` and `in(VariantType, ArrayType<VariantType>) :: BooleanType`?

src/query/functions-v2/src/scalars/control.rs
```diff
@@ -173,4 +186,108 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                Ok(Value::Column(result))
+            }
+        }
+    }
+
+    for left in ALL_NUMERICS_TYPES {
+        with_number_mapped_type!(|NUM_TYPE| match left {
+            NumberDataType::NUM_TYPE => {
+                registry.register_passthrough_nullable_2_arg::<NumberType<NUM_TYPE>, ArrayType<NumberType<NUM_TYPE>>, BooleanType, _, _>(
+                    "in",
```
andylokandy, line 219: `contains` may sounds more reasonable to me.

src/query/functions-v2/tests/it/scalars/testdata/control.txt
andylokandy: because `i16` is the common super type of `u8` and `i8`

src/query/functions-v2/src/scalars/control.rs
RinChanNOWWW: Currently, all `col in (col1, col2, ... )` will be converted to `col = col1 or col = col2 ...` in the parse phase.

src/query/functions-v2/src/scalars/control.rs
sundy-li: We can add them if we really need them. For now, they are rarely used.

src/query/functions-v2/src/scalars/control.rs
sundy-li: we will add `contains(Array, item)` function later, `contains` can accept Array as non-scalar values. In function is only used for `Expr::InList`

src/query/functions-v2/src/scalars/control.rs
sundy-li: Ok, let's throw errors as clickhouse did and let users use  `Expr::InList` or `contains` function instead (TBD).

RinChanNOWWW (approved):

andylokandy (approved):

PR #8693, opened 2022-11-08
chore(cluster): exchange error backtrace

zhang2014:

## Summary

chore(cluster): exchange error backtrace

Closes #issue

### DIFF
--- a/src/common/exception/src/exception_flight.rs
+++ b/src/common/exception/src/exception_flight.rs
@@ -12,17 +12,32 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::Arc;
+
 use common_arrow::arrow_format::flight::data::FlightData;
 
+use crate::exception::ErrorCodeBacktrace;
 use crate::ErrorCode;
 use crate::Result;
 
 impl From<ErrorCode> for FlightData {
     fn from(error: ErrorCode) -> Self {
-        // TODO: has stack trace
+        let error_message = error.message();
+        let error_backtrace = error.backtrace_str();
+
+        let message = error_message.as_bytes();
+        let backtrace = error_backtrace.as_bytes();
+
+        let mut data_body = Vec::with_capacity(16 + message.len() + backtrace.len());
+
+        data_body.extend((message.len() as u64).to_be_bytes());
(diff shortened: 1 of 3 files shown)


src/common/exception/src/exception_flight.rs
```diff
@@ -12,17 +12,32 @@
...
 use common_arrow::arrow_format::flight::data::FlightData;
 
+use crate::exception::ErrorCodeBacktrace;
 use crate::ErrorCode;
 use crate::Result;
 
 impl From<ErrorCode> for FlightData {
     fn from(error: ErrorCode) -> Self {
-        // TODO: has stack trace
+        let error_message = error.message();
```
Xuanwo, line 25: How about encode as `bincode`? Maybe we also need to add some message to indicate that this is a backtrace from other nodes.

Xuanwo: We also need to be compatible with flight data from older versions of the query.

zhang2014: Flight protocol does not need compatibility.

zhang2014: > How about encode as `bincode`? Maybe we also need to add some message to indicate that this is a backtrace from other nodes.

This needs to be obtained in the exchange processor. Here we only need to serialize and deserialize errors.

src/common/exception/src/exception_flight.rs
```diff
@@ -37,8 +52,28 @@ impl TryFrom<FlightData> for ErrorCode {
             Err(_) => Err(ErrorCode::BadBytes("Cannot parse inf usize.")),
             Ok(slice) => {
                 let code = u16::from_be_bytes(slice);
-                let message = String::from_utf8(flight_data.data_body)?;
-                Ok(ErrorCode::create(code, message, None, None))
+                let data_body = flight_data.data_body;
+
+                let message_len = u64::from_be_bytes(data_body[0..8].try_into().unwrap()) as usize;
```
BohuTANG, line 57: It's  more safe change to `data_body[0..sizeof(u64)]` instead of `8`

BohuTANG (approved):

PR #8701, opened 2022-11-08
feat(query): continue to improve sort and fix bug.

RinChanNOWWW:

## Summary

**This PR will stay in draft until the new expression migration is done.**

Introduce a new struct `SimpleCursor` for single order key with primitive type.

Fix bug: should return to `Consume` if one block is drained if the input port is not finished. Thus the performance test in #8677 is incorrect.

Closes #issue


RinChanNOWWW:
Need to reimplement this after migrating new expression.

Issue #8615, opened 2022-11-03
bug: col1 like col2 will be oom kill


TCeason:
**Summary**

table 15000000 rows

```sql
databend> desc orders;
+-----------------+---------+------+---------+-------+
| Field           | Type    | Null | Default | Extra |
+-----------------+---------+------+---------+-------+
| o_orderkey      | BIGINT  | NO   | 0       |       |
| o_custkey       | BIGINT  | NO   | 0       |       |
| o_orderstatus   | VARCHAR | NO   |         |       |
| o_totalprice    | DOUBLE  | NO   | 0       |       |
| o_orderdate     | DATE    | NO   | 0       |       |
| o_orderpriority | VARCHAR | NO   |         |       |
| o_clerk         | VARCHAR | NO   |         |       |
| o_shippriority  | INT     | NO   | 0       |       |
| o_comment       | VARCHAR | NO   |         |       |
+-----------------+---------+------+---------+-------+

```

execute query on databend-query release:

```sql
select count() from orders where o_clerk like o_comment;
```

in /var/log/syslog

```
26811 Nov  3 10:26:26  kernel: [50984.173571] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=c7d8e9fa439db04e33d5cd92d1330ed38d965c345e2d396d7f3ba920c5b65bfa,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/user@1000.service,task=databen      d-query,pid=170805,uid=1000
26812 Nov  3 10:26:26  kernel: [50984.173640] Out of memory: Killed process 170805 (databend-query) total-vm:23283508kB, anon-rss:11226736kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:33288kB oom_score_adj:0
```

execute query on ck :

```sql
ck :)  select count() from orders where o_clerk like o_comment;

SELECT count()
FROM orders
WHERE o_clerk LIKE o_comment

Query id: ee2c5d3e-b4c4-44ed-9332-9eb04c274e0c

┌─count()─┐
│       0 │
└─────────┘

1 row in set. Elapsed: 14.368 sec. Processed 15.00 million rows, 1.22 GB (1.04 million rows/s., 84.80 MB/s.)

```

TCeason:
> We can try the main branch to check this issue is oom or not :)

In my local it also will be oom. 

But I see it use swap memory.

```sql
[...]

PR #8726, opened 2022-11-10
feat: adjust max io requests when read block data to avoid oom

### LINKED ISSUE #8721: feat: adjust max_storage_io_requests with block size and memory size
**Summary**

`max_storage_io_requests` default is 1000. If we run 
```
insert into log select * from log;
```
it's oom because we fetch `1000 * block size` > memory size, adjust `max_storage_io_requests` as:
`memory size / block size`

BohuTANG:

## Summary

Change the read block data max_io_requests to:
```
// Assume 160MB one block file.
let block_file_size = 160 * 1024 * 1024_usize;
let table_column_len = self.table_info.schema().fields().len();
let per_column_bytes = block_file_size / table_column_len;
let column_memory_usage = per_column_bytes * projection.len();
let max_io_requests = max_memory_usage / column_memory_usage
```

* `max_storage_max_io_requests` only controls the non-block data read IO requests, such as segment/index files read.
* `set max_memory_usage= xx` to control the read block IO requests.

`insert into log select * from log;` 25GB memory usage avg.
image

`select * from log ignore_result`:
image

Closes #8721


dantengsky (approved):
LGTM

TCeason:
Maybe this pr can also fix this oom kill? https://github.com/datafuselabs/databend/issues/8615

Xuanwo:
stateful tests failed for timeout.

src/query/settings/src/lib.rs
```diff
@@ -105,13 +108,26 @@ impl Settings {
...
-                    num_cpus::get() as u64
-                } else {
-                    conf.query.num_cpus
-                };
-                ret.set_max_threads(cpus)?;
+            {
+                if ret.get_max_threads()? == 0 {
+                    let cpus = if conf.query.num_cpus == 0 {
+                        sysinfo::System::new_with_specifics(
+                            RefreshKind::new().with_cpu(CpuRefreshKind::everything()),
```
sundy-li, line 115: num_cpus has a physical cpu num method.

src/query/storages/fuse/fuse/src/operations/read_data.rs
```diff
@@ -114,13 +124,36 @@ impl FuseTable {
...
+            }
+            ReadDataKind::BlockDataAdjustIORequests => {
+                // Assume 160MB one block file.
+                let block_file_size = 160 * 1024 * 1024_usize;
+
+                let table_column_len = self.table_info.schema().fields().len();
+                let per_column_bytes = block_file_size / table_column_len;
+                let column_memory_usage = per_column_bytes * projection.len();
+                let max_memory_usage = ctx.get_settings().get_max_memory_usage()? as usize;
+                max_memory_usage / column_memory_usage
```
sundy-li, line 145: should check whether it's in a proper range, eg:  `[1, 1000]`

zhang2014 (approved):

src/query/settings/src/lib.rs
BohuTANG: I am not sure the physical CPU is right for some containers.

BohuTANG: We used `num_cpus::get()` as available CPUs, not physical cores.

PR #8740, opened 2022-11-10
refactor(out format): move output_format_xxx.rs to a sub-folder.

youngsofun:

## Summary

Summary about this PR

Closes #issue


BohuTANG (approved):

src/query/formats/src/file_format_type.rs
```diff
@@ -27,15 +27,15 @@ use common_meta_types::StageFileFormatType;
 use common_settings::Settings;
 
 use super::clickhouse::ClickhouseSuffix;
+use crate::output_format::output_format_csv::CSVOutputFormat;
```
Xuanwo, line 30: A bit verbose to me, how about `output_format::csv::CSVOutputFormat`?

Since `CSVOutputFormat` is self-contained, we can export as `crate::CSVOutputFormat`.

sundy-li: Let's do this later until it's mature. We have lots of crates to publish.

sundy-li (approved):

src/query/formats/src/file_format_type.rs
youngsofun: thank you, refactor done, please review again.

PsiACE (approved):

PR #8743, opened 2022-11-11
feat: Multiple catalog config

### LINKED ISSUE #8708: Feature: implement create multiple catalogs from configuration
**Summary**

Support creating multiple extern catalogs from the configuration.

related: #8436

ClSlaid:

## Summary

This PR implements the Configuration part of multiple catalog

Closes #8708

### DIFF
--- a/src/common/storage/src/config.rs
+++ b/src/common/storage/src/config.rs
@@ -480,19 +480,23 @@ impl ShareTableConfig {
 }
 
 #[derive(Clone, Default, PartialEq, Eq, Debug, Serialize, Deserialize)]
+#[serde(default)]
 pub struct CatalogConfig {
+    #[serde(flatten)]
     pub catalogs: HashMap<String, CatalogDescription>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(tag = "type")]
 pub enum CatalogDescription {
+    #[serde(rename = "hive")]
     Hive(HiveCatalogConfig),
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(tag = "protocol")]
 pub enum ThriftProtocol {
+    #[serde(rename = "binary")]
     Binary,
     // Compact,
 }
--- a/src/common/storage/src/config.rs
+++ b/src/common/storage/src/config.rs
@@ -12,16 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::fmt::Debug;
 use std::fmt::Display;
 use std::fmt::Formatter;
 use std::path::Path;
+use std::str::FromStr;
 use std::sync::Arc;
 
 use common_auth::RefreshableToken;
 use common_auth::TokenFile;
 use common_base::base::tokio::sync::RwLock;
 use common_base::base::Singleton;
+use common_exception::ErrorCode;
+use common_exception::Result;
 use once_cell::sync::OnceCell;
 use serde::Deserialize;
 use serde::Serialize;
@@ -471,6 +475,59 @@ impl ShareTableConfig {
         match SHARE_TABLE_CONFIG.get() {
             None => panic!("ShareTableConfig is not init"),
             Some(config) => config.get(),
+        }
+    }
+}
+
+#[derive(Clone, Default, PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub struct CatalogConfig {
+    pub catalogs: HashMap<String, CatalogDescription>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum CatalogDescription {
+    Hive(HiveCatalogConfig),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "protocol")]
+pub enum ThriftProtocol {
+    Binary,
+    // Compact,
+}
+
+impl FromStr for ThriftProtocol {
+    type Err = ErrorCode;
+
+    fn from_str(s: &str) -> Result<ThriftProtocol> {
+        let s = s.to_lowercase();
+        match s.as_str() {
+            "binary" => Ok(ThriftProtocol::Binary),
+            _ => Err(ErrorCode::StorageOther("invalid thrift protocol spec")),
+        }
+    }
+}
+
+impl Display for ThriftProtocol {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Binary => write!(f, "binary"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct HiveCatalogConfig {
+    pub meta_store_address: String,
+    pub protocol: ThriftProtocol,
+}
+
+impl Default for HiveCatalogConfig {
+    fn default() -> Self {
+        Self {
+            meta_store_address: "127.0.0.1:9083".to_string(),
+            protocol: ThriftProtocol::Binary,
         }
     }
 }
--- a/src/query/catalog/src/catalog/manager.rs
+++ b/src/query/catalog/src/catalog/manager.rs
@@ -1,0 +1,87 @@
+// Copyright 2021 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use common_base::base::Singleton;
+use common_exception::ErrorCode;
+use common_exception::Result;
+use dashmap::mapref::entry::Entry;
+use dashmap::DashMap;
+use once_cell::sync::OnceCell;
+
+use super::Catalog;
+
+pub const CATALOG_DEFAULT: &str = "default";
+
+static CATALOG_MANAGER: OnceCell<Singleton<Arc<CatalogManager>>> = OnceCell::new();
+
+pub struct CatalogManager {
+    pub catalogs: DashMap<String, Arc<dyn Catalog>>,
+}
+
+impl CatalogManager {
+    pub fn get_catalog(&self, catalog_name: &str) -> Result<Arc<dyn Catalog>> {
+        self.catalogs
+            .get(catalog_name)
+            .as_deref()
+            .cloned()
+            .ok_or_else(|| ErrorCode::BadArguments(format!("no such catalog {}", catalog_name)))
+    }
+
+    pub fn instance() -> Arc<CatalogManager> {
+        match CATALOG_MANAGER.get() {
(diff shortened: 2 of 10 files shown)


src/common/storage/src/config.rs
```diff
@@ -474,3 +478,60 @@ impl ShareTableConfig {
         }
     }
 }
+
+#[derive(Clone, Default, PartialEq, Eq, Debug, Serialize, Deserialize)]
+#[serde(default)]
+pub struct CatalogConfig {
+    #[serde(flatten)]
+    pub catalogs: HashMap<String, CatalogDescription>,
```
Xuanwo, line 486: Catalogs should be placed outside. `common-storage` only need to hold something like `StorageParams`.

src/common/storage/src/config.rs
```diff
@@ -474,3 +478,60 @@ impl ShareTableConfig {
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum CatalogDescription {
+    #[serde(rename = "hive")]
```
Xuanwo, line 492: We don't need this rename as we will not expose this to users.

src/common/storage/src/config.rs
```diff
@@ -474,3 +478,60 @@ impl ShareTableConfig {
...
+    Hive(HiveCatalogConfig),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "protocol")]
+pub enum ThriftProtocol {
+    #[serde(rename = "binary")]
```
Xuanwo, line 499: The same here.

src/common/storage/src/config.rs
```diff
@@ -474,3 +478,60 @@ impl ShareTableConfig {
...
+    Binary,
+    // Compact,
+}
+
+impl FromStr for ThriftProtocol {
```
Xuanwo, line 504: Not needed. `ThriftProtocol` in `common-storage` should always construct by hand or `serde`.

src/common/storage/src/config.rs
```diff
@@ -474,3 +478,60 @@ impl ShareTableConfig {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Binary => write!(f, "binary"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct HiveCatalogConfig {
+    pub meta_store_address: String,
```
Xuanwo, line 526: How about `address`?

src/common/storage/src/config.rs
```diff
@@ -474,3 +478,60 @@ impl ShareTableConfig {
...
+impl Display for ThriftProtocol {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Binary => write!(f, "binary"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct HiveCatalogConfig {
```
Xuanwo, line 525: Keep the same naming style with storage: `CatalogHiveConfig`.

src/query/config/src/inner.rs
```diff
@@ -56,7 +54,7 @@ pub struct Config {
     // external catalog config.
     // - Later, catalog information SHOULD be kept in KV Service
     // - currently only supports HIVE (via hive meta store)
-    pub catalog: HiveCatalogConfig,
+    pub catalogs: CatalogConfig,
```
Xuanwo, line 57: We must be compatible here.

We should keep `catalog` as-is. And a new `catalogs` config.

We could print a warning message if `catalog` is not empty.

src/query/config/src/outer_v0.rs
```diff
@@ -98,8 +100,8 @@ pub struct Config {
     // external catalog config.
     // - Later, catalog information SHOULD be kept in KV Service
     // - currently only supports HIVE (via hive meta store)
-    #[clap(flatten)]
-    pub catalog: HiveCatalogConfig,
+    #[clap(skip)]
+    pub catalogs: CatalogConfig,
```
Xuanwo, line 104: The same. Make sure the old config work.

src/query/config/src/inner.rs
ClSlaid: Ok, let deprecate `catalog` later

src/common/storage/src/config.rs
```diff
@@ -474,3 +478,60 @@ impl ShareTableConfig {
         }
     }
 }
+
+#[derive(Clone, Default, PartialEq, Eq, Debug, Serialize, Deserialize)]
+#[serde(default)]
```
Xuanwo, line 483: Remove this. Make `Deserialize` fail if `catalogs` is missing.

src/common/storage/src/config.rs
ClSlaid: Then a new `common-extern-catalog` module in `src/common` may be needed.🤔

Cannot use `common-catalog` because it conflicts with `src/query/catalog`.

Xuanwo: Maybe we can place it in `src/query/catalog` directly.

Xuanwo: > Catalogs should be placed outside. `common-storage` only need to hold something like `StorageParams`.

Oh, here is some misunderstanding.

I mean, we should not provide:

```rust
pub struct CatalogConfig {
    #[serde(flatten)]
    pub catalogs: HashMap<String, CatalogDescription>,
}
```

Instead, we need to provide `CatalogDescription` directly without a `hashmap`. Users need to build their own `HashMap` while needed.

After this change, maybe we can rename `CatalogDescription` to `CatalogConfig`?

ClSlaid: > Maybe we can place it in `src/query/catalog` directly.

`common-config` cannot depend on `common-config` (`src/query/catalog`), or there will be a circled dependency.

Xuanwo: > `common-config` cannot depend on `common-catalog` (`src/query/catalog`), or there will be a circled dependency.

We can place it at `query/config/src/inner`. `common-storage` is not a good place.

src/common/storage/src/config.rs
ClSlaid: Let's keep this, I found `serde` or `clap` will fail on conflicted namings.

src/query/catalog/src/catalog/manager.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright 2021 Datafuse Labs.
```
Xuanwo, line 1: Nice change!

src/query/catalog/src/catalog/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// catalog_interface is the module defining `Catalog` trait
+mod interface;
+/// the catalog manager implementation
+mod manager;
+
+pub use interface::*;
+pub use manager::*;
```
Xuanwo, line 21: I prefer to export specified structs like `pub use manager::CatalogManager` so we can avoid exporting unintended types.

src/query/config/src/inner.rs
```diff
@@ -382,3 +342,49 @@ impl Debug for MetaConfig {
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum CatalogConfig {
+    Hive(CatalogHiveConfig),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ThriftProtocol {
+    Binary,
+    // Compact,
```
Xuanwo, line 354: Remove this line if we are not supporting this. We can add a doc comments for `ThriftProtocol` like:

```
/// TODO: add compact protocol support if needed
pub enum ThriftProtocol {
```

src/query/config/src/inner.rs
```diff
@@ -382,3 +342,49 @@ impl Debug for MetaConfig {
...
+}
+
+impl FromStr for ThriftProtocol {
```
Xuanwo, line 357: I thought we would never use this.

src/query/config/src/inner.rs
```diff
@@ -382,3 +342,49 @@ impl Debug for MetaConfig {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Binary => write!(f, "binary"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CatalogHiveConfig {
+    pub meta_store_address: String,
```
Xuanwo, line 379: How about using `address`?

src/query/config/src/outer_v0.rs
```diff
@@ -307,6 +344,91 @@ impl TryInto<InnerStorageConfig> for StorageConfig {
...
+        Self {
+            ty: "hive".to_string(),
+            hive: CatalogHiveConfig::default(),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Args)]
+#[serde(default = "CatalogHiveConfig::empty")]
+pub struct CatalogHiveConfig {
```
Xuanwo, line 366: Maybe we need different config types for compatible and new catalogs?

I prefer to rename `meta_store_address` into `address` and remove clap support for new catalogs.

src/query/service/tests/it/configs.rs
```diff
@@ -771,6 +777,11 @@ root = ""
 [catalog]
 meta_store_address = "127.0.0.1:9083"
 protocol = "binary"
+
+[catalogs.my_hive]
+type = "hive"
+meta_store_address = "127.0.0.1:9083"
```
Xuanwo, line 783: Make this field to be `address`.

Xuanwo (commented):
Almost LGTM! Some minor code style changes.

src/query/config/src/inner.rs
ClSlaid: > I thought we would never use this.

It's used here in line 412:
https://github.com/ClSlaid/databend/blob/e17bfcb78f6887a699601ca6c1311417a6833fcf/src/query/config/src/outer_v0.rs#L407-L415

Xuanwo: You are right! :heart:

src/query/config/src/outer_v0.rs
ClSlaid: rename `meta_store_address` in outer config will annoy Clap:
```
Argument names must be unique, but 'address' is in use by more than one argument or group'
```
I'll keep the field's name as `meta_store_address`, but make SerDe use `address` by default and `meta_store_address` as an alias.

Xuanwo: How about providing two different structs for the old hive config and new catalog configs?

We will not support using clap to configure the new catalog configs, so we don't need to bother `clap::Args` here.

Xuanwo: In the future, we will remove all clap Args support for complex configs like storage/cache/catalog.

src/query/config/src/outer_v0.rs
```diff
@@ -307,6 +344,125 @@ impl TryInto<InnerStorageConfig> for StorageConfig {
...
+    #[serde(alias = "meta_store_address")]
+    pub address: String,
+    pub protocol: String,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Args)]
+#[serde(default = "LegacyCatalogHiveConfig::empty")]
+pub struct LegacyCatalogHiveConfig {
+    #[clap(long = "hive-meta-store-address", default_value_t)]
+    #[serde(rename = "address", alias = "meta_store_address")]
```
Xuanwo, line 376: We don't need this rename. Just leave them as-is.

Xuanwo (approved):
Great work!

PR #8747, opened 2022-11-11
chore(log): remove logs in union processor

sandflee:

## Summary

Summary about this PR

Closes #issue


sandflee:
cc @leiysky

xudong963 (approved):
I guess you want to remove it because when `block` is vast, it'll influence all logs output.

Xuanwo (approved):

sandflee:
yes,  this seems for debug, no need to print the whole block,  and'll influence the performance of union

PR #8777, opened 2022-11-13
feat: add for_each in Column trait

lichuang:

## Summary

feat: add for_each in Column trait

Closes https://github.com/datafuselabs/databend/issues/8776


sundy-li (approved):
LGTM.

BTW, we are going to make `datavalues` crate deprecated by the new expression.  So it's not recommended to add some trivial functions into it.

PR #8789, opened 2022-11-14
refactor(expr): allow sparse column id in chunk

andylokandy:

## Summary

Columns in a `Chunk` has an individual column id that is not necessarily continuous.

When converting between `Chunk` and `DataBlock`, we assume the ids are continuous.

### DIFF
--- a/src/query/expression/src/chunk.rs
+++ b/src/query/expression/src/chunk.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::ops::Range;
 
 use crate::types::AnyType;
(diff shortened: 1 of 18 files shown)


leiysky (approved):

sundy-li (approved):

RinChanNOWWW:
The memory consumption of `HashMap` may be larger than `Vec` when the chunk is almost full.

sundy-li:
The index of the hashmap seems useless to represent a chunk.

Maybe we should have this map out of the chunk.

andylokandy:
The idea behind this change is to use global column id all the way through the process of planning, block pruning, and execution. It helps reduce the complexity of reusing the same Expr in different phrases in which the column refernces used in the Expr are different(some of the column refs are eliminated by constant folder), especially when considering to give the chunk a unique Expr with unique column ref set.

RinChanNOWWW:
> The idea behind this change is to use the global column id all the way through the process of planning, block pruning, and execution.

The columns order is the same as the fields order in the schema. When we wanted to operate one column in the chunk, we would write such codes:

```rust
let index = schema.index_of(&col_name)?; // which is to find an index  in  Vec<DataField>
let col = chunk.columns()[[index];
```

How do we achieve this after refactoring `Chunk`?

b41sh:
> The idea behind this change is to use the global column id all the way through the process of planning, block pruning, and execution. It helps reduce the complexity of reusing the same column id in different phrases in which the column refernces used in the Expr are different(some of the column refs are eliminated by constant folder), especially when considering to give every chunk a unique Expr with unique column ref set in the future.

Can we put `ColumnId` into `Schema`? When using it, the index of `Column` in `Chunk` can be obtained from the `Schema` according to the `ColumnId`.
This makes the structure of `Chunk` simpler. At the same time, `Schema` and `Chunk` also need to be used together, and will not bring too many additional operations.

PR #8793, opened 2022-11-14
feat: shuffle segments during distributed pruning

dantengsky:

## Summary

shuffle segment before distributed pruning, to mitigate potential data skew

Closes #issue


src/query/storages/fuse/fuse/src/operations/read_partitions.rs
```diff
@@ -60,8 +63,30 @@ impl FuseTable {
...
+                    };
+                    segments_locations.sort_by(|a, b| {
+                        let hl = hash(&a.0);
+                        let hr = hash(&b.0);
+                        hl.cmp(&hr)
+                    });
+
+                    // build the PartInfos
+                    for segment_location in segments_locations {
+                        segments.push(FuseLazyPartInfo::create(segment_location))
```
sundy-li, line 89: Sorting the segments may not be enough.

We should modify `self.ctx.try_get_part();` to make the segment cache more local affinity with the nodes.

sundy-li:
If the segment number is too slow (less than query nodes), the `redistribute_source_fragment` may be inaccurate.

FuseLazyPartInfo seems too lazy to schedule.

BohuTANG (approved):

PR #8805, opened 2022-11-15
feat(query): optimize count(Nullable(col))

### LINKED ISSUE #8804: Feature: optimize count(Nullable(col))
**Summary**

The column stat has record `null_count` and table statics has record `row_count`, if we execute `select count(nullable(col)) from table`

the count can be folded.

TCeason:

## Summary

If simple count(nullable(col)), fold it.

|now|before optimize|
|:--:|:--:|
| 0.002s | 0.091s |
 
```sql
CREATE TABLE orders
(
    o_orderkey       BIGINT  null,
    o_custkey        BIGINT  null,
    o_orderstatus    STRING  null,
    o_totalprice     DOUBLE  null,
    o_orderdate      DATE  null,
    o_orderpriority  STRING  null,  
    o_clerk          STRING  null, 
    o_shippriority   INTEGER  null,
    o_comment        STRING  null
);

mysql> select count(o_clerk) from orders;
+----------------+
| count(o_clerk) |
+----------------+
|       15000000 |
+----------------+
1 row in set (0.02 sec)
Read 1 rows, 1.00 B in 0.002 sec., 469.46 rows/sec., 469.46 B/sec.

mysql> select count(o_clerk) from orders;
+----------------+
| count(o_clerk) |
+----------------+
|       15000000 |
+----------------+
1 row in set (0.10 sec)
Read 15000000 rows, 330.81 MiB in 0.091 sec., 164.49 million rows/sec., 3.54 GiB/sec.

```

Closes #8804

### DIFF
--- a/src/query/sql/src/planner/binder/table.rs
+++ b/src/query/sql/src/planner/binder/table.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use common_ast::ast::Indirection;
@@ -50,6 +51,7 @@ use crate::plans::ConstantExpr;
 use crate::plans::LogicalGet;
 use crate::plans::Scalar;
 use crate::BindContext;
+use crate::ColumnSet;
 use crate::IndexType;
 
 impl<'a> Binder {
@@ -302,6 +304,9 @@ impl<'a> Binder {
         let mut bind_context = BindContext::with_parent(Box::new(bind_context.clone()));
         let columns = self.metadata.read().columns_by_table_index(table_index);
         let table = self.metadata.read().table(table_index).clone();
+
+        let mut column_indexs: ColumnSet = HashSet::new();
+        let mut has_path_indices = false;
         let mut col_stats: HashMap<IndexType, Option<ColumnStatistics>> = HashMap::new();
         for column in columns.iter() {
             let column_binding = ColumnBinding {
@@ -317,16 +322,22 @@ impl<'a> Binder {
                 },
             };
             bind_context.add_column_binding(column_binding);
-            let column_id = column.index();
+            column_indexs.insert(column.index());
+            if !has_path_indices {
+                // create table t (c0 tuple(int, int), c1 string null) in planner col index is : (0: c0, 1: c1, 2: c0:2 3: c0:2)
+                // in storage column (0: PrimitiveColumn, 1: PrimitiveColumn, 2: NullableColumn)
+                has_path_indices = column.has_path_indices();
+            }
             let col_stat = table
                 .table()
                 .column_statistics_provider()
                 .await?
-                .column_statistics(column_id as ColumnId);
-            col_stats.insert(column_id, col_stat);
+                .column_statistics(column.index() as ColumnId);
+            col_stats.insert(column.index(), col_stat);
         }
-        let stat = table.table().table_statistics()?;
 
+        let is_accurate = !has_path_indices && table.table().engine().to_lowercase() == "fuse";
+        let stat = table.table().table_statistics()?;
         Ok((
             SExpr::create_leaf(
                 LogicalGet {
@@ -338,6 +349,7 @@ impl<'a> Binder {
                     statistics: stat,
                     prewhere: None,
                     col_stats,
+                    is_accurate,
                 }
                 .into(),
             ),
--- a/src/query/sql/src/planner/binder/table.rs
+++ b/src/query/sql/src/planner/binder/table.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use common_ast::ast::Indirection;
@@ -27,6 +28,8 @@ use common_ast::Backtrace;
 use common_ast::Dialect;
 use common_ast::DisplayError;
 use common_catalog::catalog_kind::CATALOG_DEFAULT;
+use common_catalog::table::ColumnId;
+use common_catalog::table::ColumnStatistics;
 use common_catalog::table::NavigationPoint;
 use common_catalog::table::Table;
 use common_catalog::table_function::TableFunction;
@@ -79,6 +82,7 @@ impl<'a> Binder {
         );
 
         self.bind_base_table(bind_context, database, table_index)
+            .await
     }
 
     pub(super) async fn bind_table_reference(
@@ -168,8 +172,9 @@ impl<'a> Binder {
                                 .write()
                                 .add_table(catalog, database.clone(), table_meta);
 
-                        let (s_expr, mut bind_context) =
-                            self.bind_base_table(bind_context, database.as_str(), table_index)?;
+                        let (s_expr, mut bind_context) = self
+                            .bind_base_table(bind_context, database.as_str(), table_index)
+                            .await?;
                         if let Some(alias) = alias {
                             bind_context.apply_table_alias(alias, &self.name_resolution_ctx)?;
                         }
@@ -224,8 +229,9 @@ impl<'a> Binder {
                     table.clone(),
                 );
 
-                let (s_expr, mut bind_context) =
-                    self.bind_base_table(bind_context, "system", table_index)?;
+                let (s_expr, mut bind_context) = self
+                    .bind_base_table(bind_context, "system", table_index)
+                    .await?;
                 if let Some(alias) = alias {
                     bind_context.apply_table_alias(alias, &self.name_resolution_ctx)?;
                 }
@@ -287,7 +293,7 @@ impl<'a> Binder {
         Ok((cte_info.s_expr.clone(), new_bind_context))
     }
 
-    fn bind_base_table(
+    async fn bind_base_table(
         &mut self,
         bind_context: &BindContext,
         database_name: &str,
@@ -296,6 +302,7 @@ impl<'a> Binder {
         let mut bind_context = BindContext::with_parent(Box::new(bind_context.clone()));
         let columns = self.metadata.read().columns_by_table_index(table_index);
         let table = self.metadata.read().table(table_index).clone();
+        let mut col_stats: HashMap<IndexType, Option<ColumnStatistics>> = HashMap::new();
         for column in columns.iter() {
             let column_binding = ColumnBinding {
                 database_name: Some(database_name.to_string()),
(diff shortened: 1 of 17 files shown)


src/query/sql/src/planner/binder/table.rs
```diff
@@ -312,6 +317,18 @@ impl<'a> Binder {
...
         }
         let stat = table.table().table_statistics()?;
+        let mut col_stats: HashMap<u32, Option<ColumnStatistics>> = HashMap::new();
+        for column in columns.iter() {
+            let column_id = column.index() as u32;
+            let col_stat = table
+                .table()
+                .column_statistics_provider()
+                .await
+                .unwrap()
```
BohuTANG, line 327: unwrap is not safe, please use `?`

xudong963 (commented):
Unlike other optimizations (such as join reorder), it does not cause false results even if the statistics are inaccurate, or at worst, poor performance.

The optimization depends on `column_statistics`, if the stat is inaccurate, the result will be wrong. 

Ping @dantengsky -- Can the statistical information used in this PR be guaranteed to be 100% accurate?

dantengsky:
> Ping @dantengsky -- Can the statistical information used in this PR be guaranteed to be 100% accurate?

the `number_of_distinct_values` provided currently, is FAKED, just the number of rows; @lichuang is working on a real impl of ndv: https://github.com/datafuselabs/databend/pull/8521, but it will only provide approximated values.

the `null_count` shall be accurate, though

src/query/sql/src/planner/optimizer/rule/rewrite/rule_fold_count_aggregate.rs
```diff
@@ -87,6 +87,17 @@ impl Rule for RuleFoldCountAggregate {
...
+                Scalar::AggregateFunction(agg_func) => {
+                    agg_func.func_name == "count"
+                        && (agg_func.args.is_empty()
+                            || matches!(agg_func.args[0].data_type(), Nullable(_)))
+                        && !agg_func.distinct
+                }
+                _ => false,
+            });
+
         if let (true, Some(card)) = (is_simple_count, input_prop.precise_cardinality) {
```
sundy-li, line 101: why not introduce another field like `    pub precise_cardinality: Option<u64>,`

src/query/sql/src/planner/optimizer/rule/rewrite/rule_fold_count_aggregate.rs
```diff
@@ -101,8 +112,36 @@ impl Rule for RuleFoldCountAggregate {
...
+            input_prop.column_stats,
+            input_prop.precise_cardinality,
+        ) {
+            let mut scalars = agg.aggregate_functions;
+            // Now count function must only have one arg.
+            for item in scalars.iter_mut() {
+                if let Scalar::AggregateFunction(agg_func) = item.scalar.clone() {
+                    let col_set = agg_func.args[0].used_columns();
+                    for index in col_set {
+                        let col_stat = column_stats.get(&index);
```
TCeason, line 126: The index maybe not true column_stats needs index.

```rust
CREATE TABLE `t` (
  `id` INT,
  `b` TUPLE(INT, INT),
  `c` VARCHAR NULL
)

mysql> insert into t(id) values(0),(1),(2),(3);

mysql> select * from t;
+------+-------+------+
| id   | b     | c    |
+------+-------+------+
|    0 | (0,0) | NULL |
|    1 | (0,0) | NULL |
|    2 | (0,0) | NULL |
|    3 | (0,0) | NULL |
+------+-------+------+
4 rows in set (0.04 sec)
Read 4 rows, 89.00 B in 0.020 sec., 198.4 rows/sec., 4.31 KiB/sec.

mysql> select count(c) from t;
+----------+
| count(c) |
+----------+
|        4 |
+----------+
1 row in set (0.02 sec)
Read 1 rows, 1.00 B in 0.002 sec., 461.1 rows/sec., 461.10 B/sec.
```

In BoundColumn the index of col b is 1, index of col c is 2.

But in RelationalProperty the index of col c is 3.

So, I convert this pr state to a draft.

TCeason:
> the column_stats of hive in planning stat is not correct. may leading wrong result in hive sql?

Could you please give me a demo, I can test it.

sandflee:
> Could you please give me a demo, I can test it.

hive doesn't provide column_statistics_provider now,  even provided most hive tables on-product does't have columns stats, and columns stats is not 100% trustable.

sundy-li:
> > Could you please give me a demo, I can test it.
> 
> hive doesn't provide column_statistics_provider now, even provided most hive tables on-product does't have columns stats, and columns stats is not 100% trustable.

That's ok. 

HiveTable return None TableStatistics
```
 fn table_statistics(&self) -> Result<Option<TableStatistics>> {
        Ok(None)
    }
```

sandflee:
> > > Could you please give me a demo, I can test it.
> > 
> > 
> > hive doesn't provide column_statistics_provider now, even provided most hive tables on-product does't have columns stats, and columns stats is not 100% trustable.
> 
> That's ok.
> 
> HiveTable return None TableStatistics
> 
> ```
>  fn table_statistics(&self) -> Result<Option<TableStatistics>> {
>         Ok(None)
>     }
> ```
this optimize should't depend on the TableStatistics of hive, we planed to implement it next month :(

sundy-li:
That' ok, we can introduce another field named `is_accruate` to indicate whether this `TableStatistics ` is fully trusted.

TCeason:
> That' ok, we can introduce another field named `is_accruate` to indicate whether this `TableStatistics ` is fully trusted.

I add a function in src/query/catalog/src/table.rs, now only fuse table is accurate.

```rust
fn is_accurate(&self) -> bool {
    false
}
```

src/query/catalog/src/table.rs
```diff
@@ -200,6 +200,10 @@ pub trait Table: Sync + Send {
         Ok(None)
     }
 
+    fn is_accurate(&self) -> bool {
```
sundy-li, line 203: no...

src/query/sql/src/planner/binder/table.rs
```diff
@@ -317,32 +321,58 @@ impl<'a> Binder {
...
-                .column_statistics(column_id as ColumnId);
-            col_stats.insert(column_id, col_stat);
+            column_indexs.insert(column.index());
+            if !has_tuple_col {
+                // create table t (c0 tuple(int, int), c1 string null) in planner col index is : (0: c0, 1: c1, 2: c0:2 3: c0:2)
+                // in storage column (0: PrimitiveColumn, 1: PrimitiveColumn, 2: NullableColumn)
+                has_tuple_col = matches!(column.data_type(), DataTypeImpl::Struct(_))
+            }
+        }
+        if table.table().is_accurate() && !has_tuple_col {
```
sundy-li, line 331: if `has_tuple_col`, the statistics will be none, that's unacceptable.

sundy-li: 因小失大

TCeason: Ok, I get your idea. we should have statistics data.

TCeason: Now only the table columns do not have tuple col and table engine = fuse, 

we judge the is_accurate as true.

src/query/sql/src/planner/binder/table.rs
```diff
@@ -322,6 +348,8 @@ impl<'a> Binder {
                     order_by: None,
                     statistics: stat,
                     prewhere: None,
+                    col_stats,
```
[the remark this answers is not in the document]
TCeason: Good idea. It's much more concise.

src/query/sql/src/planner/binder/table.rs
```diff
@@ -310,7 +322,21 @@ impl<'a> Binder {
...
+                // create table t (c0 tuple(int, int), c1 string null) in planner col index is : (0: c0, 1: c1, 2: c0:2 3: c0:2)
+                // in storage column (0: PrimitiveColumn, 1: PrimitiveColumn, 2: NullableColumn)
+                has_path_indices = column.has_path_indices();
+            }
+            let col_stat = table
+                .table()
+                .column_statistics_provider()
+                .await?
+                .column_statistics(column.index() as ColumnId);
+            col_stats.insert(column.index(), col_stat);
```
b41sh, line 337: `ColumnIndex` in `Metadata` is a global id, it is different from `ColumnId` in parquet. We need a transformation function to do the mapping.

src/query/sql/src/planner/binder/table.rs
```diff
@@ -320,7 +347,11 @@ impl<'a> Binder {
                     push_down_predicates: None,
                     limit: None,
                     order_by: None,
-                    statistics: stat,
+                    statistics: Statistics {
+                        statistics: stat,
+                        col_stats,
```
b41sh, line 352: `col_stats` is only used by `count(col)` aggregate function, it seems that there is no need for a binder to every table, and lazy loading seems to be a better choice.

sundy-li (approved):

src/query/sql/src/planner/binder/table.rs
xudong963: Good point

TCeason: I try to modify the code. And add it in post_optimize, but we should use the function 

```rust
table
    .column_statistics_provider()
    .await?
    .column_statistics(col_id as ColumnId);
col_stats.insert(column.index(), col_stat);

// because of in pub async fn snapshot_loc(&self) -> Result<Option<String>> : self.operator.object(url).read().await?;
```

If set post_optimize to async, It's too invasive.

So, I'm not sure, do we have other way to get the col_stats? cc @dantengsky  @lichuang

xudong963 (approved):

src/query/sql/src/planner/binder/table.rs
TCeason: > `col_stats` is only used by `count(col)` aggregate function, it seems that there is no need for a binder to every table, and lazy loading seems to be a better choice.

Actually we try to lazy loading this, but there has a question:

the statistics is saved in logicalget, now we can get a column_stat provider(thx @dantengsky ) but in post_optimize, the count() pattern type is :

```sql
pattern: SExpr::create_unary(
                PatternPlan {
                    plan_type: RelOp::Aggregate,
                }
                .into(),
                SExpr::create_leaf(
                    PatternPlan {
                        plan_type: RelOp::Pattern,
                    }
                    .into(),
                ),
            ),
```

So we can't downcast the relop::Pattern to logicalget. 

Because of this, I set this comment to resolved

PR #8818, opened 2022-11-16
feat(query): support setting default max_memory_usage in config

### LINKED ISSUE #8816: Feature: support setting default max_memory_usage in query config
**Summary**

`max_memory_usage` is automatically determined with sysinfo and may not working with cgroup.

everpcpc:

## Summary

Summary about this PR

Closes #8816


BohuTANG (approved):

Xuanwo (approved):

src/query/settings/src/lib.rs
```diff
@@ -125,7 +125,11 @@ impl Settings {
             // Set max memory usage.
             {
                 if ret.get_max_memory_usage()? == 0 {
-                    let max_usage = sysinfo::System::new_all().available_memory();
+                    let max_usage = if conf.query.max_memory_usage == 0 {
+                        sysinfo::System::new_all().available_memory()
```
sundy-li, line 129: did it work in virtual machine ?

BohuTANG: No, it will get the host machine's Memory, so we add `max_memory_usage` to the config to throttle.

everpcpc: > did it work in virtual machine ?

Virtual machines are OK,  but it would not work with container resources limited by cgroup.

PR #8824, opened 2022-11-17
refactor(storage): deletion

### LINKED ISSUE #7964: improvement: move delete to pipeline builder
## Summary

Now the delete processor delete block serially one by one. It would be great to move delete processor to the pipeline_builder. 
https://github.com/datafuselabs/databend/blob/5ae594effc7b0f4766e5e680f1b709f5a179af2d/src/query/storages/fuse/src/operations/delete.rs#L102-L125

## Discussion

The delete process is to select a group of blocks through `BlockPruner`, and then judge whether the block needs to be rewritten according to the selection. If it needs to be rewritten, rewrite the block and replace the old `BlockMeta` with the new `BlockMeta`. This process will be split into multiple transforms, and some information needs to be passed between pipes, such as the old and new BlockMeta, etc.
There are the following options:
1. Add [`MetaInfo`](https://github.com/datafuselabs/databend/commit/c3538c8efbaa8b8e549c90e85867dff6484c9383) to the `DataBlock` to store the required information.
```
#[typetag::serde(tag = "type")]
pub trait MetaInfo: Send + Sync + DynClone {
    fn as_any(&self) -> &dyn Any;

    #[allow(clippy::borrowed_box)]
    fn equals(&self, info: &Box<dyn MetaInfo>) -> bool;
}

pub type MetaInfoPtr = Box<dyn MetaInfo>;

pub struct DeleteMeta {
     ... ...
}

impl MetaInfo for DeleteMeta {
    ... ...
}

pub struct DataBlock {
    schema: DataSchemaRef,
    columns: Vec<ColumnRef>,
    meta: Option<MetaInfoPtr>,
}
```
2. Write the required information into the `DataBlock` and pass it to the downstream.

### LINKED ISSUE #9213: bug: delete panic
For the clustering table, if cluster key is expr, it will panic when run delete.
```
mysql> create table t(a int) cluster by (a+1);

mysql> insert into t values(3),(4);

mysql> delete from t where a=2;
ERROR 1105 (HY000): Code: 1068, displayText = Cannot join handle from context's runtime, cause: task 1057 panicked.
```

Solution: When generating cluster statistics for deletion, we need to do expression execute based on cluster key.

zhyass:

## Summary

1.  Strip the `CompactSink` to a more generalized `MutationSink`.

     Add `MergeSegmentsTransform`, used to collect and organize newly generated segments for compact.

2. Refactor Deletion

    The flow of Pipeline is as follows:
    ```
    +---------------+
    |DeletionSource1| ------
    +---------------+       |      +-----------------+      +------------+
    |     ...       | ...   | ---> |DeletionTransform| ---> |MutationSink|
    +---------------+       |      +-----------------+      +------------+
    |DeletionSourceN| ------
    +---------------+
    ```

    `DeletionSource`: filter out the rows that meet the filter conditions to generate a new block and write the object. The process is similar to `FuseSource`.
    ```
    Read block -> Filter -> Read Remain -> Generate new block -> Serialize block -> Generate Deletion -> Output
    
    pub enum Deletion {
      // none of the rows should be removed.
      DoNothing,
      // some of the row should be removed.
      Replaced(Arc<BlockMeta>),
      // all the rows should be removed.
      Deleted,
    }
    ```
    
    `DeletionTransform`: collect and organize the `Deletion`s to generate new segments.

3. Optimization
    ```
    delete from t where 1 = 1;
    delete from t where now();
    ```
    if the `filter_expr` is of "constant" nullary : for the whole block, whether all of the rows should be kept or dropped. we can just return without accessing the block data.

4. Fix bug

    For the clustering table, if cluster key is expr, it will panic when run delete.
    Solution: When generating cluster statistics for deletion, we need to do expression execute based on cluster key.

Closes #7964 
Closes #9213


zhyass:
Following failed statements:
---------------------------------------------
Runner: mysql
ErrorType: statement query execute with exception
Message: Failed to execute. Collected info: 1105 (HY000): result row write failed: Code: 1001, displayText = invalid type: integer `34`, expected a sequence at line 1 column 7(while in query pipeline init).

Parsed Statement
    at_line: 91,
    s_type: Statement: query, type: B, query_type: B, retry: False,
    suite_name: base/03_dml/03_0025_delete_from,
    text:
        select count(*) = 1 from t where c IS NULL;
    results: [(<re.Match object; span=(0, 4), match='----'>, 95, '1')],
    runs_on: {'http', 'mysql', 'clickhouse'},

src/query/storages/fuse/fuse/src/operations/delete.rs
```diff
@@ -55,105 +75,188 @@ impl FuseTable {
...
-        self.commit_mutation(
-            &ctx,
-            del_holder.base_snapshot(),
-            segments,
-            summary,
-            abort_operation,
-        )
-        .await
+        match pipeline.pipes.last() {
+            None => Err(ErrorCode::Internal("Cannot resize empty pipe.")),
```
BohuTANG, line 229: The error msg with `resize` is confusing.

src/query/sql/src/planner/plans/delete.rs
```diff
@@ -14,34 +14,21 @@
 
 use std::sync::Arc;
 
-use common_catalog::plan::Projection;
 use common_datavalues::DataSchema;
 use common_datavalues::DataSchemaRef;
 use common_meta_app::schema::TableIdent;
 
-/// # TODO
```
Xuanwo, line 22: Thanks!

src/query/storages/fuse/fuse/src/operations/delete.rs
```diff
@@ -15,29 +15,49 @@
...
+use crate::pipelines::processors::port::InputPort;
+use crate::pipelines::processors::port::OutputPort;
+use crate::pipelines::Pipe;
+use crate::pipelines::Pipeline;
 use crate::pruning::BlockPruner;
 use crate::statistics::ClusterStatsGenerator;
 use crate::FuseTable;
 
 impl FuseTable {
+    // The flow of Pipeline is as follows:
```
Xuanwo, line 47: Please use `///` for comments.

src/query/storages/fuse/fuse/src/operations/mutation/compact/merge_segments_transform.rs
```diff
@@ -0,0 +1,186 @@
...
+            abort_operation: AbortOperation::default(),
+        })))
+    }
+
+    fn get_current_input(&mut self) -> Option<Arc<InputPort>> {
+        let mut finished = true;
+        let mut index = self.cur_input_index;
+
+        loop {
+            let input = &self.inputs[index];
```
Xuanwo, line 84: Do we need to check the inputs' length?

tests/logictest/suites/base/03_dml/03_0025_delete_from
```diff
@@ -140,6 +140,29 @@ select count(*) = 0 from t;
 
 
 
+statement ok
```
Xuanwo, line 143: How about adding a stateful test based on `ontime` and `hits` in which we will delete data on different condition and check the remaining count.

BohuTANG: `ontime/hits` are stateful tests, the delete should be a stateless test first, but I think we need more test cases here, such as generating a table from `select ... from numbers(100)` and doing some delete operations on this table.

src/query/storages/fuse/fuse/src/operations/mutation/compact/merge_segments_transform.rs
zhyass: https://github.com/datafuselabs/databend/pull/8824/files/7700c95cbb534196d9743b80b314653afc4a682d#diff-9ea44ca0145e52317af2fa0e3f0c5e8a0018712dfd449a4ed10e82c79b6dd406R96-R98

src/query/catalog/src/table.rs
```diff
@@ -227,10 +226,11 @@ pub trait Table: Sync + Send {
     async fn delete(
         &self,
         ctx: Arc<dyn TableContext>,
-        projection: &Projection,
-        selection: &Option<String>,
+        filter: Option<Expression>,
```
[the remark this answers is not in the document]
zhyass: Using scalar in query/catalog directly will get an error "error: cyclic package dependency".

src/query/storages/fuse/fuse/src/operations/delete.rs
```diff
@@ -55,105 +75,188 @@ impl FuseTable {
...
-        let schema = self.table_info.schema();
-        // TODO refine pruner
-        let extras = PushDownInfo {
+        let func_ctx = FunctionContext::default();
+
+        let dummy_field = DataField::new("dummy", NullType::new_impl());
+        let dummy_schema = Arc::new(DataSchema::new(vec![dummy_field.clone()]));
+        let dummy_column = DataValue::Null.as_const_column(&NullType::new_impl(), 1)?;
+        let dummy_data_block = DataBlock::create(dummy_schema.clone(), vec![dummy_column]);
+
```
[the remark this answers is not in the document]
dantengsky: DataBlock:empty() not work…

tests/logictest/suites/base/03_dml/03_0025_delete_from
zhyass: Added test in [03_0025_delete_from](https://github.com/datafuselabs/databend/pull/8824/files/7700c95cbb534196d9743b80b314653afc4a682d#diff-e06a0a882879de1a4f36a1f8e8e949f101e8a71fc6dc3eab407b256e80682704). We had a number of test cases  to cover various scenarios previously.

src/query/catalog/src/table.rs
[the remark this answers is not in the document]
BohuTANG: @sundy-li should know better. Any suggestions?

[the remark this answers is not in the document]
sundy-li: Keep it as you want, we will refactor it later.

dantengsky (approved):

BohuTANG:
Let's merge, on the way to update...

PR #8826, opened 2022-11-17
chore(query): add deserializer new expression

sundy-li:

## Summary

Summary about this PR

Closes #issue


src/query/datavalues/src/types/deserializations/timestamp.rs
```diff
@@ -53,7 +53,6 @@ impl TypeDeserializer for TimestampDeserializer {
         for row in 0..rows {
             let mut reader = &reader[step * row..];
             let value: i64 = reader.read_scalar()?;
-            check_timestamp(value)?;
```
b41sh, line 56: don't need to check the timestamp?

src/query/expression/src/deserializations/array.rs
```diff
@@ -0,0 +1,124 @@
...
+            return Err(ErrorCode::BadDataValueType("Array is empty".to_string()));
+        }
+        let total = self.offsets.pop().unwrap();
+        Ok((total - *self.offsets.last().unwrap()) as usize)
+    }
+}
+
+impl TypeDeserializer for ArrayDeserializer {
+    fn memory_size(&self) -> usize {
+        self.inner.memory_size()
```
b41sh, line 52: the memory of offsets also needs to add.

src/query/expression/src/deserializations/timestamp.rs
```diff
@@ -36,23 +42,63 @@ impl TypeDeserializer for TimestampDeserializer {
...
+                let v = v.clone();
+                let mut reader = Cursor::new(v.as_bytes());
+                let ts = reader.read_timestamp_text(&format.timezone)?;
+
+                let micros = ts.timestamp_micros();
+                check_timestamp(micros)?;
+                self.builder.push(micros);
+                Ok(())
+            }
+            _ => Err(ErrorCode::from("Incorrect boolean value")),
```
b41sh, line 68: Incorrect timestamp value

src/query/expression/src/deserializations/tuple.rs
```diff
@@ -0,0 +1,119 @@
...
+                        obj.len()
+                    )));
+                }
+                for (inner, item) in self.inners.iter_mut().zip(obj.iter()) {
+                    let (_, val) = item;
+                    inner.de_json(val, format)?;
+                }
+                Ok(())
+            }
+            _ => Err(ErrorCode::from("Incorrect boolean value")),
```
b41sh, line 72: Incorrect tuple value

src/query/expression/src/deserializations/tuple.rs
```diff
@@ -0,0 +1,119 @@
...
+                inner.de_binary(&mut reader, format)?;
+            }
+        }
+        Ok(())
+    }
+
+    fn append_data_value(&mut self, value: Scalar, format: &FormatSettings) -> Result<()> {
+        let v = value
+            .as_tuple()
+            .ok_or_else(|| ErrorCode::from("Unable to get timestamp value"))?;
```
b41sh, line 95: Unable to get tuple value

src/query/expression/src/deserializations/tuple.rs
```diff
@@ -0,0 +1,119 @@
...
+        match value {
+            serde_json::Value::Object(obj) => {
+                if self.inners.len() != obj.len() {
+                    return Err(ErrorCode::from_string(format!(
+                        "Incorrect json value, expect {} values, but get {} values",
+                        self.inners.len(),
+                        obj.len()
+                    )));
+                }
+                for (inner, item) in self.inners.iter_mut().zip(obj.iter()) {
```
b41sh, line 66: The JSON Object is iterated in lexicographical order, which may be different from the inner deserializer order of the tuple

src/query/datavalues/src/types/deserializations/timestamp.rs
sundy-li: It's de from binary mode (we ensure it's right we serialize) , we can skip this check to improve the performance.

src/query/expression/src/deserializations/tuple.rs
sundy-li: Ok ,so we can only accept array value now.

src/query/datavalues/src/types/deserializations/timestamp.rs
b41sh: ok

b41sh (approved):

PR #8832, opened 2022-11-17
fix(storage): segment duplication during fuse table compaction

### LINKED ISSUE #8831: bug: segment duplication during fuse table compaction
**Summary**

A simple case (provided by @zhyass ):
~~~
mysql> create table t(a int);

mysql> optimize table t compact limit 5;

mysql> select * from t;
+------+
| a    |
+------+
|    1 |
|    1 |
|    2 |
|    2 |
+------+
4 rows in set (0.03 sec)
~~~

NOTE that this bug could be triggered WITHOUT using `limit`:
 
given a snapshot, with "perfect" segment followed by segments that have fragmented blocks, compaction of it, will highly likely trigger this bug:

(provided by @zhyass )
~~~
mysql> create table t(c int) block_per_segment=2 row_per_block=2;

mysql>
mysql> insert into t values(1),(2),(3),(4);

mysql>
mysql> optimize table t compact;

mysql>
mysql> insert into t values(5),(6),(7);

mysql>
mysql> optimize table t compact;

mysql>
mysql> select * from t order by c;
+------+
| c    |
+------+
|    1 |
|    1 |
|    2 |
|    2 |
|    3 |
|    3 |
|    4 |
|    4 |
|    5 |
|    5 |
|    6 |
|    6 |
|    7 |
|    7 |
+------+
14 rows in set (0.05 sec)
Read 14 rows, 56.00 B in 0.025 sec., 559.66 rows/sec., 2.19 KiB/sec.
~~~

@zhyass pls correct me if the above description is wrong.

zhyass:

## Summary

The end maybe set to 0, cause the segment duplication bug.

Closes #8831


dantengsky:
@zhyass  we need a test case for it

BohuTANG:
Can we add a test for it?

tests/logictest/suites/base/09_fuse_engine/09_0019_issue_8831
```diff
@@ -0,0 +1,55 @@
+statement ok
+DROP DATABASE IF EXISTS issue_8130;
+
+statement ok
+CREATE DATABASE issue_8130;
+
+statement ok
+USE issues_8130;
```
dantengsky, line 8:
```suggestion
USE issue_8130;
```

BohuTANG (approved):

Issue #8854, opened 2022-11-19
Feature: extract or clause to push down potential predicates for join


xudong963:
**Summary**

For example: `select * from t1, t2 where (t1.a=1 and t2.b=2) or (t1.a=2 and t2.b=1)`

The predicate will be rewritten to `((t1.a=1 and t2.b=2) or (t1.a=2 and t2.b=1)) and (t1.a=1 or t1.a=2) and (t2.b=2 or t2.b=1)`

So `(t1.a=1 or t1.a=1), (t2.b=2 or t2.b=1)` may be pushed down join and reduce rows between join.

PR #8855, opened 2022-11-19
feat: extract or clause  to push down potential predicates for join

xudong963:

## Summary

Extract or predicates from Filter to push down them to join.

For example: `select * from t1, t2 where (t1.a=1 and t2.b=2) or (t1.a=2 and t2.b=1)`

The predicate will be rewritten to `((t1.a=1 and t2.b=2) or (t1.a=2 and t2.b=1)) and (t1.a=1 or t1.a=2) and (t2.b=2 or t2.b=1)`

So `(t1.a=1 or t1.a=1), (t2.b=2 or t2.b=1)` may be pushed down join and reduce rows between join

The optimization will speed up tpch q7, benchmark data:**before 3.398 sec, after0.769 sec**

Closes https://github.com/datafuselabs/databend/issues/8854


src/query/sql/src/planner/optimizer/util.rs
```diff
@@ -12,9 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use common_datavalues::type_coercion::compare_coercion;
```
BohuTANG, line 15: Prefer split(rename) this `util.rs` into separate files, otherwise, it will be bigger and bigger.

xudong963:
@mergify update

sundy-li (approved):

BohuTANG:
@mergify update

PR #8883, opened 2022-11-21
feat(query): unset max_threads/memory_usage should recover to start value

### LINKED ISSUE #8881: feat: unset max_threads should set detect val
@TCeason For max_threads or max_memory_usage, I think we should auto detect again when unset and set the detect value.

_Originally posted by @BohuTANG in https://github.com/datafuselabs/databend/issues/8870#issuecomment-1322106640_

TCeason:

## Summary

1. unset max_threads/memory_usage should recover to start value
2. max_memory_usage modify to : total_memory * 0.8

Closes #8881

### DIFF
--- a/src/query/service/src/interpreters/interpreter_unsetting.rs
+++ b/src/query/service/src/interpreters/interpreter_unsetting.rs
@@ -17,9 +17,6 @@ use std::sync::Arc;
 use common_exception::Result;
 use common_settings::ScopeLevel::Global;
 use common_sql::plans::UnSettingPlan;
-use sysinfo::CpuRefreshKind;
-use sysinfo::RefreshKind;
-use sysinfo::SystemExt;
 
 use crate::interpreters::Interpreter;
 use crate::pipelines::PipelineBuildResult;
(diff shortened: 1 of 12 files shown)


src/query/service/src/interpreters/interpreter_unsetting.rs
```diff
@@ -46,17 +49,44 @@ impl Interpreter for UnSettingInterpreter {
         let mut keys: Vec<String> = vec![];
         let mut values: Vec<String> = vec![];
         let mut is_globals: Vec<bool> = vec![];
+        let settings = self.ctx.get_settings();
         for var in plan.vars {
             let (ok, value) = match var.to_lowercase().as_str() {
                 // To be compatible with some drivers
                 "sql_mode" | "autocommit" => (false, String::from("")),
+                "max_threads" => {
```
sundy-li, line 57: There are duplicate codes to handle the default value for specific setting, it makes the codes hard to maintain.

Why not use `let default_val = settings.check_and_get_default_value(setting)?.to_string();` instead ?

src/query/service/tests/it/storages/testdata/system-tables.txt
```diff
@@ -311,41 +311,41 @@ DB.Table: 'system'.'roles', Table: roles-table_id:1, ver:0, Engine: SystemRoles
...
+| max_block_size                  | 65536      | 65536      | SESSION | Maximum block size for reading, default value: 65536.                                                                        | UInt64 |
+| max_execute_time                | 0          | 0          | SESSION | The maximum query execution time. it means no limit if the value is zero. default value: 0.                                  | UInt64 |
+| max_memory_usage                | 1073741824 | 0          | SESSION | The maximum memory usage for processing single query, in bytes. By default the value is 0 it means determined automatically. | UInt64 |
+| max_storage_io_requests         | 1000       | 1000       | SESSION | The maximum number of concurrent IO requests. By default, it is 1000.                                                        | UInt64 |
+| max_threads                     | 2          | 0          | SESSION | The maximum number of threads to execute the request. By default the value is 0 it means determined automatically.           | UInt64 |
```
sundy-li, line 338: max_thread should not defaults to 0.

src/query/service/src/interpreters/interpreter_unsetting.rs
TCeason: If so, I think the setting `max_memory_usage` can also handle as same now.

src/query/service/tests/it/storages/testdata/system-tables.txt
BohuTANG: Same as `max_memory_usage`.

src/query/service/src/interpreters/interpreter_unsetting.rs
```diff
@@ -46,19 +46,34 @@ impl Interpreter for UnSettingInterpreter {
...
-                        }
-                        ScopeLevel::Session => {}
+                    if matches!(settings.get_setting_level(setting)?, Global) {
+                        self.ctx.get_settings().try_drop_setting(setting)?;
                     }
-                    let default_val = settings.check_and_get_default_value(setting)?.to_string();
+                    let default_val = {
+                        if setting == "max_memory_usage" {
+                            let conf = self.ctx.get_config();
+                            if conf.query.max_memory_usage == 0 {
```
BohuTANG, line 61: No need to check `conf.query.max_memory_usage == 0`, just set the max_memory_usage to check_and_get_default_value?

BohuTANG: I see.

BohuTANG (approved):

Issue #7888, opened 2022-09-26
bug: http handler blocks for 3 second when creating an existing db.


drmingdrmer:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

0.8.47-nightly

### What's Wrong?

When creating an existing db with http-handler, databend-query blocks for 3 second before returning.

## Re-produce

```
time curl -u root: --request POST 127.0.0.1:8000/v1/query/ --header 'Content-Type: application/json' --data-raw '{"sql": "create database ddd"}' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   645  100   615  100    30    201      9  0:00:03  0:00:03 --:--:--   211
{
  "id": "4f67f638-dcf2-4d93-abde-bb94afd900aa",
  "session_id": "e56290dd-e159-48c6-85d2-4ff5f1a7ad06",
  "session": null,
  "schema": {
    "fields": [],
    "metadata": {}
  },
  "data": [],
  "state": "Failed",
  "error": {
    "code": 2301,
    "message": "Database 'ddd' already exists"
  },
  "stats": {
    "scan_progress": {
      "rows": 0,
      "bytes": 0
    },
    "write_progress": {
      "rows": 0,
      "bytes": 0
    },
    "result_progress": {
      "rows": 0,
      "bytes": 0
    },
    "running_time_ms": 16.962708
  },
  "affect": null,
  "stats_uri": "/v1/query/4f67f638-dcf2-4d93-abde-bb94afd900aa",
  "final_uri": "/v1/query/4f67f638-dcf2-4d93-abde-bb94afd900aa/final",
  "next_uri": null,
  "kill_uri": "/v1/query/4f67f638-dcf2-4d93-abde-bb94afd900aa/kill"
}

real    0m3.056s
user    0m0.024s
sys     0m0.007s
```

--- 

## query logs:

```
query日志:
2022-09-26T04:07:41.893374Z DEBUG worker_loop: common_meta_client::grpc_client: MetaGrpcClient send response to the handle resp=Get(Ok(Some(SeqV { seq: 1057, meta: None, data: "[binary]" })))
[...]

PR #8894, opened 2022-11-22
fix(http handler): avoid dropping runtime when task on it not finished.

youngsofun:

## Summary

the original release session/query_context right after the computation is finished,  because only some result/status is useful, but a task is still running.
we can fix it first if this pr does not lead to other problems, and refactor the code later.

Closes  https://github.com/datafuselabs/databend/issues/7888


flaneur2020 (approved):

BohuTANG:
Can we add a test for the http handler in this case?

zhang2014 (approved):

PR #8904, opened 2022-11-23
refactor(mem-tracker): universal tracker restore for normal quit and panic

drmingdrmer:

## Related Issues

### DIFF
--- a/src/common/base/src/base/runtime_tracker.rs
+++ b/src/common/base/src/base/runtime_tracker.rs
@@ -52,6 +52,17 @@ pub struct ThreadTracker {
     buffer: StatBuffer,
 }
 
+/// A guard that restores the thread local tracker to `old` when being dropped.
+pub struct TrackerGuard {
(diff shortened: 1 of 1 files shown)


zhang2014 (approved):

src/common/base/src/base/runtime_tracker.rs
```diff
@@ -245,10 +265,12 @@ impl MemoryTracker {
     }
 }
 
+/// A [`Future`] that enters its thread tracker when being polled.
+///
+/// [`Future`]: std::future::Future
```
Xuanwo, line 270: `rustdoc` is smart enough that We don't need this line if we have imported `Future`.

Xuanwo: For example: https://opendal.databend.rs/src/opendal/io.rs.html#29

drmingdrmer: Roger that. Next PR.

PR #8910, opened 2022-11-23
feat(format): Output float NaN instead of nan

### LINKED ISSUE #8901: Feature: Output float NaN instead of nan
**Summary**

Description for this feature.

NaN is more standardized and common used than nan.

image

ariesdevil:

## Summary

`NaN` is more standardized and commonly used than `nan`.

Closes #8901


ariesdevil:
FYI: normally, `inf` and `nan` are the uniform format(eg: [Inf, NaN], [inf, nan]), but in Rust they are [inf, NaN] :(

src/common/io/src/consts.rs
```diff
@@ -19,5 +19,5 @@ pub const FALSE_BYTES_NUM: &str = "0";
 pub const NULL_BYTES_UPPER: &str = "NULL";
 pub const NULL_BYTES_LOWER: &str = "null";
 pub const NULL_BYTES_ESCAPE: &str = "\\N";
-pub const NAN_BYTES_LOWER: &str = "nan";
+pub const NAN_BYTES_LOWER: &str = "NaN";
```
sundy-li, line 22: it's not lower anymore

youngsofun:
>>>  in Rust they are [inf, NaN] 

1. nan repr for each format will be able to set by users later,  maybe after the refactor of settings/ options. 
1. for the default value, what is import the standard in  Rust,  but the format 
2. `NaN` should have its own name (NAN_BYTES_SNAKE)  instead of  NAN_BYTES_LOWER

ariesdevil:
> > > > in Rust they are [inf, NaN]
> 
> 1. nan repr for each format will be able to set by users later,  maybe after the refactor of settings/ options.
> 2. for the default value, what is import the standard in  Rust,  but the format
> 3. `NaN` should have its own name (NAN_BYTES_SNAKE)  instead of  NAN_BYTES_LOWER

Yup, Change NAN_BYTES_LOWER and INF_BYTES_LOWER to NAN_BYTES_SNAKE, INF_BYTES_SNAKE are better.

youngsofun:
‘-inf’, and ‘nan' is used for TSV in clickhouse https://clickhouse.com/docs/en/interfaces/formats, which is used for data injection and driver.

youngsofun:
some of the current default value may not be the best choice.
but it should be change for one format each time with sufficient reasons.
because we need to keep what`s unloaded earlier  loadeable

ariesdevil:
So we need to make the format style for each file format? cc @sundy-li

youngsofun:
> And we need to keep both `NaN` and `nan` existing?

yes

youngsofun:
https://github.com/datafuselabs/databend/issues/8914 @ariesdevil we can discuss for each format

sundy-li:
So we should introduce a new settings?

youngsofun:
introduce one option until necessary, most DB (except clickhouse) seems works well with only a few options,
https://github.com/datafuselabs/databend/issues/8914

ariesdevil:
As discussed with @youngsofun I will change format one by one, so close this PR for now. cc @sundy-li

PR #8924, opened 2022-11-23
fix: remove meta_type config

lichuang:

## Summary

fix: remove meta_type config, Cannot set both embedded dir and [address|endpoints] config.

Closes #issue


docker/databend-query-docker.toml
```diff
@@ -37,7 +37,6 @@ dir = "./.databend/logs"
 
 [meta]
 # To enable embedded meta-store, set address to ""
-embedded_dir = "_meta_embedded_1"
 address = "0.0.0.0:9191"
```
BohuTANG, line 40: Change all `address = "0.0.0.0:9191"` to `endpoints = ["0.0.0.0:9191"]`?

src/binaries/query/main.rs
```diff
@@ -39,6 +39,7 @@ use tracing::info;
 #[databend_main]
 async fn main(_global_tracker: Arc<MemoryTracker>) -> common_exception::Result<()> {
     let conf: Config = Config::load()?;
+    println!("config: {:?}", conf);
```
BohuTANG, line 42: Move this print info to the segment with `Print information to users.`?

src/query/config/src/outer_v0.rs
```diff
@@ -1542,43 +1533,17 @@ pub struct MetaConfig {
...
-
-        // Fallback is used for forward compatbility, that is:
-        // First check embedded config, then endpoints, finally address.
-        if t == MetaType::Fallback {
-            if !self.embedded_dir.is_empty() && self.endpoints.is_empty() {
-                self.meta_type = MetaType::Embedded.to_string();
-            } else {
-                self.meta_type = MetaType::Remote.to_string();
-            }
+        println!(
```
BohuTANG, line 1536: This `println` is for debug?

src/query/config/src/inner.rs
```diff
@@ -264,13 +243,12 @@ pub struct MetaConfig {
     /// Certificate for client to identify meta rpc serve
     pub rpc_tls_meta_server_root_ca_cert: String,
     pub rpc_tls_meta_service_domain_name: String,
-    pub meta_type: MetaType,
 }
 
 impl Default for MetaConfig {
     fn default() -> Self {
         Self {
-            embedded_dir: "./.databend/meta_embedded".to_string(),
```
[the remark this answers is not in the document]
lichuang: `MetaType ` now has been removed, and cannot set bot embedded_dir and [address|endpoints ], if want to use embedded meta, user MUST set `embedded_dir` ifself.

BohuTANG:
```
cargo-machete found the following unused dependencies in /workspace:
common-config -- /workspace/src/query/config/Cargo.toml:
	strum_macros
```

BohuTANG:
Some test still failure, need to addressed

drmingdrmer:
Shall we remove `address` too? It's still confusing to users with two config entries for service endpoints.

Xuanwo:
> Shall we remove `address` too? It's still confusing to users with two config entries for service endpoints.

I'm second with this.

src/query/config/src/inner.rs
```diff
@@ -264,29 +242,44 @@ pub struct MetaConfig {
...
+        let has_remote = !self.endpoints.is_empty();
+        if has_embedded_dir && has_remote {
+            return Err(ErrorCode::InvalidConfig(
+                "Cannot set both embedded dir and endpoints in meta config".to_string(),
+            ));
+        }
+
+        if !has_embedded_dir && !has_remote {
+            return Err(ErrorCode::InvalidConfig(
+                "Set embedded_dir or endpoints config in meta config".to_string(),
```
BohuTANG, line 278:
```suggestion
                "Please set your meta endpoints config: endpoints = ["<your-meta-service-endpoints>"]".to_string(),
```

src/query/config/src/inner.rs
```diff
@@ -264,29 +242,44 @@ pub struct MetaConfig {
...
-        Ok(self.meta_type == MetaType::Embedded)
+        Ok(!self.embedded_dir.is_empty())
+    }
+
+    pub fn check_valid(&self) -> Result<()> {
+        let has_embedded_dir = !self.embedded_dir.is_empty();
+        let has_remote = !self.endpoints.is_empty();
+        if has_embedded_dir && has_remote {
+            return Err(ErrorCode::InvalidConfig(
+                "Cannot set both embedded dir and endpoints in meta config".to_string(),
```
BohuTANG, line 272:
```suggestion
                "Can not set embedded_dir and endpoints at the same time, embedded_dir is only for testing, please remove this config".to_string(),
```

src/binaries/query/main.rs
```diff
@@ -44,6 +44,7 @@ async fn main(_global_tracker: Arc<MemoryTracker>) -> common_exception::Result<(
         return Ok(());
     }
 
+    conf.meta.check_valid()?;
```
BohuTANG, line 47: Can we move this check to the config load?

BohuTANG: Or add an new function `check_valid` for `Config`. Inside the `check_valid` we can do our all valid checking.

drmingdrmer (approved):

src/binaries/query/main.rs
lichuang: > Can we move this check to the config load?

If check when config loaded, some test may fail, cause `endpoints` and `embedded_dir` now has no default values.

BohuTANG:
```
    let conf: Config = Config::load()?; -- Why test may fail if we check it in the end of load?

    if run_cmd(&conf) {
        return Ok(());
    }

    conf.meta.check_valid()?;
```

BohuTANG: Like this:
https://github.com/BohuTANG/databend/commit/04cda338e866c42cc50b3365bc8102ddee0b5875

I think it's OK.

BohuTANG: Or we can add a check_valid for Config.

src/query/config/src/inner.rs
```diff
@@ -65,11 +65,19 @@ impl Config {
...
+        let cfg: Self = OuterV0Config::load()?.try_into()?;
 
+        // Only check meta config when cmd is empty.
+        if cfg.cmd.is_empty() {
+            cfg.meta.check_valid()?;
+        }
         Ok(cfg)
     }
 
+    pub fn load_with_configs(kvs: Vec<(String, Option<String>)>) -> Result<Self> {
```
BohuTANG, line 77: This is used for test ,can we move it out from inner.rs to test?

src/query/service/tests/it/sessions/session_context.rs
```diff
@@ -23,7 +23,10 @@ use databend_query::sessions::SessionContext;
 
 #[tokio::test]
 async fn test_session_context() -> Result<()> {
-    let conf = Config::load()?;
+    let conf = Config::load_with_configs(vec![(
```
BohuTANG, line 26: If the load only affects this test, I think we can not call `conf.load` for it, build the Config as:
```
let conf = Config {....}
```

BohuTANG (approved):

BohuTANG:
cc @ZhiHanZ @flaneur2020  for checking whether the cloud is using:
```
embedded_dir = "_meta_embedded_1"
or
address = "0.0.0.0:9191"
```
Otherwise, the query will stop when it starts.

PR #8927, opened 2022-11-23
feat(mem-limit): limit global memory usage

drmingdrmer:

## Related Issues


src/query/config/src/outer_v0.rs
```diff
@@ -1218,6 +1222,7 @@ impl TryInto<InnerQueryConfig> for QueryConfig {
             tenant_id: self.tenant_id,
             cluster_id: self.cluster_id,
             num_cpus: self.num_cpus,
+            max_total_memory: self.max_total_memory,
```
BohuTANG, line 1225: Already have a max_memory_usage, it seems no need add a new one.

drmingdrmer: It looks like a per-tenant setting.

drmingdrmer: I am not quite sure about the purpose of `conf.query.max_memory_usage` but it is used to override a user setting as a default value: 
https://github.com/datafuselabs/databend/blob/c1fff835e7fddec52de7cf5aa2940387be7b9005/src/query/settings/src/lib.rs#L123-L130

Is it meant to be the max memory a process can use?

BohuTANG: Yes, it's same as your total_memory here

drmingdrmer:
@zhang2014 suggested putting the memory limiter into `MemoryTracker`, which will be a totally different design. Close this PR and re-impl it in:

- #8946

Issue #8914, opened 2022-11-23
tracking the options for each format


youngsofun:
**Summary**

provide
- the options for each format
- the default value and limits to each option
- the reasons and the refers

formats:

- CSV https://github.com/datafuselabs/databend/issues/8913

PR #8929, opened 2022-11-23
feat(csv_nan): Change nan for csv to NaN

### LINKED ISSUE #8913: CSV format options
**Summary**

Description for this feature.

# nan_string = `nan` (not settable for now) 

python csv lib use `nan` （I tried）

# inf

- python lib https://[docs.python.org](https://docs.python.org/3/library/csv.html)/3/library/csv.html
- pandas
- execl
- rust lib https://github.com/BurntSushi/rust-csv
- databases， see https://github.com/datafuselabs/databend/issues/8914

ariesdevil:

## Summary

Make `NaN` format in settings and change CSV from `nan` to `NaN`.

Closes #8913


src/query/pipeline/sources/src/processors/sources/input_formats/input_format.rs
```diff
@@ -30,6 +30,8 @@ pub trait InputFormat: Send + Sync {
 
     fn default_field_delimiter(&self) -> u8;
 
+    fn default_nan_display(&self) -> String;
```
youngsofun, line 33: we do not need this in the trait InputFormat, nan is used only in FieldDecoder.
for csv/tsv, InputFormat is used for cutting rows and fields with delimiter/quote/escape. these bytes of fields is feed to FieldDecoder for interpretation.

youngsofun:
there are 3 things to discuss

1.  the name of the option， `xxx_display`  https://github.com/datafuselabs/databend/discussions/8931
2.  default value for csv,  need more survey https://github.com/datafuselabs/databend/issues/8914

src/query/formats/src/file_format_type.rs
```diff
@@ -324,6 +325,14 @@ fn check_options(options: &mut FileFormatOptions) -> Result<()> {
         options.record_delimiter = '\n'.to_string();
     }
 
+    if options.nan_display.is_empty() {
+        options.nan_display = "NaN".to_string();
```
youngsofun, line 329: this apply to all format?

src/query/formats/src/file_format_type.rs
```diff
@@ -324,6 +325,14 @@ fn check_options(options: &mut FileFormatOptions) -> Result<()> {
...
+    } else if options.nan_display.len() != 3 || options.nan_display.to_lowercase() != "nan" {
+        return Err(ErrorCode::InvalidArgument(
+            "nan_display must be case-sensitive literal `nan`",
```
youngsofun, line 332: case-insensitive?

src/query/service/tests/it/storages/testdata/system-tables.txt
```diff
@@ -325,6 +325,7 @@ DB.Table: 'system'.'settings', Table: settings-table_id:1, ver:0, Engine: System
 | format_empty_as_default         | 1          | 1           | SESSION | Format empty_as_default, default value: 1.                                                                        | UInt64 |
 | format_escape                   |            |             | SESSION | format escape char, default value: "", which means the format`s default setting.                                  | String |
 | format_field_delimiter          |            |             | SESSION | Format field delimiter, default value is "": use default of the format.                                           | String |
+| format_nan_display              | nan        | nan         | SESSION | Format field delimiter, default value is "nan": use default of the format.                                        | String |
```
youngsofun, line 328: "nan" can not stands for `use default of the format`

src/query/pipeline/sources/src/processors/sources/input_formats/input_format.rs
youngsofun: default_field_delimiter is also temporary,  need  refactor later:
 maybe add a enum to hold used options in XXXInputFormat.

BohuTANG: Maybe we can add a display dictionary to do it instead of dealing with 'default_nan_display' for `nan`.

BohuTANG: To avoid the logic and codes fat and fat, we should do the common(For example, are there other displays that also have such a need?) instead add more special cases.

sandflee:
BTW,  could u add a setting to change the default value of `inf` ?  mysql jdbc could't regnize `inf` values.

src/query/pipeline/sources/src/processors/sources/input_formats/input_format.rs
youngsofun: yes, as I commented above

1.  default_nan_display is useless here
2. default_field_delimiter will be removed from this trait later.

youngsofun:
> BTW, could u add a setting to change the default value of `inf` ? mysql jdbc could't regnize `inf` values.

thank you for your advise, we really need user side  feedback for these options?

what repr for Inf do you need? "Inf"? (with uppercase "I") and "NaN"

handler does have not interface to let user choose options.
mysql handler use ValueFormat internal,   we can set the option (fixed) in mysql writer  when creating ValuesOutputFormat object

src/query/formats/src/file_format_type.rs
ariesdevil: case-sensitive is right here, eg: nan, NaN, Nan.

src/query/formats/src/file_format_type.rs
ariesdevil: Yup, as default value

src/query/formats/src/file_format_type.rs
```diff
@@ -324,6 +325,14 @@ fn check_options(options: &mut FileFormatOptions) -> Result<()> {
         options.record_delimiter = '\n'.to_string();
     }
 
+    if options.nan_display.is_empty() {
+        options.nan_display = "nan".to_string();
+    } else if options.nan_display.len() != 3 || options.nan_display.to_lowercase() != "nan" {
```
ariesdevil, line 330: @youngsofun should we consider `null` here?

youngsofun: yes

sundy-li (approved):

src/meta/proto-conv/src/user_from_to_protobuf_impl.rs
```diff
@@ -494,11 +494,18 @@ impl FromToProto for mt::FileFormatOptions {
             })?,
         )?;
 
+        let nan_display = if p.nan_display.is_empty() {
```
youngsofun, line 497: why set to "nan"?

src/query/formats/src/field_decoder/csv.rs
```diff
@@ -48,7 +48,7 @@ impl FieldDecoderCSV {
                 true_bytes: TRUE_BYTES_LOWER.as_bytes().to_vec(),
                 false_bytes: FALSE_BYTES_LOWER.as_bytes().to_vec(),
                 null_bytes: NULL_BYTES_ESCAPE.as_bytes().to_vec(),
-                nan_bytes: NAN_BYTES_LOWER.as_bytes().to_vec(),
+                nan_bytes: NAN_BYTES_SNAKE.as_bytes().to_vec(),
```
youngsofun, line 51: should get from FileFormatOptionsExt

src/query/formats/src/file_format_type.rs
```diff
@@ -324,6 +325,14 @@ fn check_options(options: &mut FileFormatOptions) -> Result<()> {
         options.record_delimiter = '\n'.to_string();
     }
 
+    if options.nan_display.is_empty() {
```
youngsofun, line 328: diff value for diff format  here

src/query/settings/src/lib.rs
```diff
@@ -254,6 +254,16 @@ impl Settings {
...
                 possible_values: None,
             },
+            SettingValue {
+                default_value: UserSettingValue::String("nan".to_owned()),
+                user_setting: UserSetting::create(
+                    "format_nan_display",
+                    UserSettingValue::String("nan".to_owned()),
+                ),
+                level: ScopeLevel::Session,
+                desc: "must be literal `nan` (case-sensitive), default value is \"nan\".",
```
youngsofun, line 264: default value is not  "nan"?  you want CSV use NaN in the pr?

youngsofun: depends on format

src/meta/proto-conv/src/user_from_to_protobuf_impl.rs
youngsofun: if need this,  should be done at a higher level and format by format. the check and set default logic is mainly in file_format_type.rs now,  which may need  refactor later too.

src/query/formats/src/field_decoder/csv.rs
youngsofun: and for all formats (may be later)

src/meta/proto-conv/tests/it/user_stage.rs
```diff
@@ -53,6 +53,52 @@ fn test_user_stage_oss_latest() -> anyhow::Result<()> {
...
+            format: mt::StageFileFormatType::Json,
+            skip_header: 1024,
+            field_delimiter: "|".to_string(),
+            record_delimiter: "//".to_string(),
+            nan_display: "NaN".to_string(),
+            compression: mt::StageFileCompression::Bz2,
+            escape: "\\".to_string(),
+            row_tag: "row".to_string(),
+        },
+        copy_options: mt::CopyOptions {
```
drmingdrmer, line 86: If UserStageInfo is not affected by this change, you can just test FileFormatOptions

ariesdevil: As the `user.proto` version already increased, I think testing full UserStageInfo is OK here.

drmingdrmer (approved):

PsiACE (approved):

src/common/io/tests/it/options_deserializer.rs
```diff
@@ -74,6 +76,10 @@ fn default_field_delimiter() -> String {
     ",".to_string()
 }
 
+fn default_nan_display() -> String {
```
youngsofun, line 79: not need default_nan_display

src/meta/proto-conv/src/user_from_to_protobuf_impl.rs
```diff
@@ -494,11 +494,18 @@ impl FromToProto for mt::FileFormatOptions {
...
+            "".to_string()
```
youngsofun, line 498: the if is useless...

src/query/formats/src/file_format_type.rs
```diff
@@ -335,6 +345,9 @@ fn final_csv_options(options: &mut FileFormatOptionsExt) -> Result<()> {
     if options.quote.is_empty() {
         options.quote = "\"".to_string();
     }
+    if options.stage.nan_display.is_empty() {
```
youngsofun, line 348: other formats need to deal with it too. because they will use `nan_display` later
and `nan_display=""` will be store in meta and cause problem.

src/query/formats/src/field_decoder/csv.rs
```diff
@@ -48,7 +47,7 @@ impl FieldDecoderCSV {
                 true_bytes: TRUE_BYTES_LOWER.as_bytes().to_vec(),
                 false_bytes: FALSE_BYTES_LOWER.as_bytes().to_vec(),
                 null_bytes: NULL_BYTES_ESCAPE.as_bytes().to_vec(),
-                nan_bytes: NAN_BYTES_LOWER.as_bytes().to_vec(),
+                nan_bytes: options.stage.nan_display.as_bytes().to_vec(),
```
youngsofun, line 50: other formats need to use stage.nan_display later.
we should change them now.
so we can tests it`s default values properly handled (should not change to NaN or "") after this pr.

youngsofun:
another related issue is inf,
it should be changed right after this pr.  or we may see "NaN" and "inf" in the same csv.


and we`d better record the solid reason to change "nan" to "NaN"?  maybe provide a link in the source code comment too.

https://github.com/datafuselabs/databend/discussions/8934

youngsofun:
we can add a test in tests/suites/1_stateful/05_formats/05_02_csv/

youngsofun:
let`s merge it pr first, I will refactor it to solve the remains issues better.

src/query/formats/src/file_format_type.rs
ariesdevil: other formats using consts right now, will modify file format one by one.

PR #8963, opened 2022-11-25
feat(query): Improve fuse source.

RinChanNOWWW:
Split one part into multiple blocks.

## Summary

Limit the output size of one `DataBlock` produced by fuse source.


RinChanNOWWW:
Fuse engine was refactored and disabled prewhere filter.

PR #8966, opened 2022-11-25
feat(query): support bulk insert without exprssion

### LINKED ISSUE #8911: Feature: improve bulk insert speed
**Summary**
```
insert into tb values(1,'a','databend'),(2,'a','databend'),(3,'a','databend'),...
```

Maybe we could add settings to support inserts that don't degenerate into single inserts.

cc @sundy-li

b41sh:

## Summary

`insert_values_disable_expression` is used to identify whether to turn off expression parsing when inserting values.
If `insert_values_disable_expression` is set to 1, values will be inserted as `SyncSource`, and only the raw value will be inserted, an expression will cause an error.

Closes #8911


tests/logictest/suites/base/03_dml/03_0016_insert_into_values
```diff
@@ -52,6 +52,15 @@ create table if not exists st1(a string);
 statement error 1303
 insert into st1 values ('a', 'b');
 
+statement ok
+insert into st1 values (upper('a'));
+
+statement ok
+set insert_values_disable_expression = 1;
```
youngsofun, line 59: need set global (and set back)?

src/query/settings/src/lib.rs
```diff
@@ -471,6 +471,16 @@ impl Settings {
...
                 possible_values: None,
             },
+            SettingValue {
+                default_value: UserSettingValue::UInt64(0),
+                user_setting: UserSetting::create(
+                    "insert_values_disable_expression",
+                    UserSettingValue::UInt64(0),
+                ),
+                level: ScopeLevel::Session,
+                desc: "If enable insert values disable expression, default value: 0.",
```
youngsofun, line 481:
```suggestion
                desc: "Whether to disable expression when insert values, default value: 0.",
```

tests/logictest/suites/base/03_dml/03_0016_insert_into_values
youngsofun, line 59:
```suggestion
set global insert_values_disable_expression = 1;
```

tests/logictest/suites/base/03_dml/03_0016_insert_into_values
```diff
@@ -52,6 +52,15 @@ create table if not exists st1(a string);
...
+
+statement error 4000
+insert into st1 values (upper('a'));
```
youngsofun, line 62:
```suggestion
insert into st1 values (upper('a'));

statement ok
set global insert_values_disable_expression = 0;
```

src/query/service/src/interpreters/interpreter_insert_v2.rs
```diff
@@ -425,6 +486,39 @@ impl ValueSource {
...
+        if !reader.ignore_byte(b'(') {
+            return Err(ErrorCode::BadDataValueType(
+                "Must start with parentheses".to_string(),
+            ));
+        }
+        for col_idx in 0..col_size {
+            let _ = reader.ignore_white_spaces();
+            let deser = desers
+                .get_mut(col_idx)
+                .ok_or_else(|| ErrorCode::BadBytes("Deserializer is None"))?;
```
youngsofun, line 511:
```suggestion
                .ok_or_else(|| ErrorCode::Internal("Deserializer is None"))?;
```

youngsofun: /// Internal means this is the internal error that no action
    /// can be taken by neither developers or users.
    /// In most of the time, they are code bugs.

youngsofun (approved):

tests/logictest/suites/base/03_dml/03_0016_insert_into_values
BohuTANG: We should add a test to check `insert_values_disable_expression = 1` works:
```
insert into st1 values ('a'),('b'), ('c');
select * from st1;
```

src/query/settings/src/lib.rs
```diff
@@ -471,6 +471,16 @@ impl Settings {
                 desc: "How many hours will the COPY file metadata expired in the metasrv, default value: 24*7=7days",
                 possible_values: None,
             },
+            SettingValue {
+                default_value: UserSettingValue::UInt64(0),
+                user_setting: UserSetting::create(
+                    "insert_values_disable_expression",
```
BohuTANG, line 477: Let's introduce a new setting: `insert_values_enable_expression`, default is false. in most cases, there is no expression, so the user no need to change anything they can use it.

youngsofun: make sense in production,
but for now, many value need to be expressed with the expression? like `parse_json`?
some  logic tests may be affected?

BohuTANG: Not sure if `insert_values_enable_expression=false`, if the insert value has an expression, can we return an error let the user `enable insert_values_enable_expression`?

sundy-li: > Let's introduce a new setting: insert_values_enable_expression, default is false.

It may take a bad effect on users because it will not be compatible with the script which already works fine.

BohuTANG: It should be ok, bulk insert mostly has no expressions, it is the raw sql that is dumped by mysqldump, are there any other cases?
The key point here is that: if he has expression in the bulk insert, we give an error hint to let him enable expression on.

b41sh: We can make the default value of `insert_values_enable_expression` as true, this is compatible with the current logic, and users can choose to turn it off if they want better performance.

BohuTANG: Ok.

src/query/settings/src/lib.rs
```diff
@@ -471,6 +471,16 @@ impl Settings {
...
                 possible_values: None,
             },
+            SettingValue {
+                default_value: UserSettingValue::UInt64(1),
+                user_setting: UserSetting::create(
+                    "insert_values_enable_expression",
+                    UserSettingValue::UInt64(1),
+                ),
+                level: ScopeLevel::Session,
+                desc: "Whether to enable expression when insert values, default value: 1.",
```
BohuTANG, line 481:
```suggestion
                desc: "Whether to enable expression when inserting values, if your values do not have expressions please disable this setting to improve write performance, default value: 1.",
```

sundy-li (approved):

src/query/settings/src/lib.rs
b41sh: done

BohuTANG:
Conflicting files
src/query/service/tests/it/storages/testdata/system-tables.txt

wubx:
When I tested it, I found no performance improvement, and it actually decreased.
cc @b41sh 

the tb_order.sql use mysqldump dump from MySQL.

insert_values_enable_expression=1
```
time mysql -h 127.0.0.1 -P3307 -uroot default <./tb_order.sql
real    1m2.701s
user    0m8.683s
sys     0m0.612s

time mysql -h 127.0.0.1 -P3307 -uroot default <./tb_order.sql

real    1m2.907s
user    0m8.649s
sys     0m0.355s
```

insert_values_enable_expression=0
```
wu@fz001:~$ time mysql -h 127.0.0.1 -P3307 -uroot default <./tb_order.sql
real    1m3.242s
user    0m8.648s
sys     0m0.403s

wu@fz001:~$ time mysql -h 127.0.0.1 -P3307 -uroot default <./tb_order.sql

real    1m3.796s
user    0m8.499s
sys     0m0.409s

```

PR #8967, opened 2022-11-25
feat: alter table flashback

### LINKED ISSUE #8943: Feature: sql statement that revert fuse table snapshot
**Summary**

A new sql statement, let's say `alter table revert to SNAPSHOT_ID` which reverts the fuse table to the given snapshot.

This command is supposed to be part of the emergency recovery toolkit.

dantengsky:

## Summary

`alter table T flashback to [travel point]`

e.g.

~~~
alter table t flashback to (snapshot => '12dbc77c1a8c443081964e99599664e7');
alter table t flashback to (TIMESTAMP => '2022-12-06 04:35:17.856848'::TIMESTAMP);
~~~

which essentially updates the table metadata that resides in the metastore, and lets it point to the given travel point. 
- no data in the object storage will be mutated
  except that the `last_snapshot_location_hint` will be updated if flashback succeed
- some checkings will be performed before updating the metadata (of metastore)
  - the snapshot being flashbacked to should be loadable (at the time of execution)
  - the snaphost being flashbacked to is a valid snapshot according to the table’s current state (i.e. it is one of the snapshot of the snapshot chain at the time of execution)
  -  BUT, segment / block that referenced by the snapshot being reverted to are NOT checked

NOTE:

- this statement is intended to be used in the emergency recovery case
   please do not use it to navigate the history, use `select ... at [travel point]` instead.
- once the table has been flashback to a previous snapshot, it can NOT be flash-forward
    I think reverting to arbitrary snapshot is also needed for emergency recovery, and like to impl it in another dedicated PR.

Closes #8943


BohuTANG:
`alter table revert to [travel point]` -- Is `revert` a one-time operation? If it is not a one-time operation, `alter table flashback to [travel point]` is better?

dantengsky:
> `alter table revert to [travel point]` -- Is `revert` a one-time operation? If it is not a one-time operation, `alter table flashback to [travel point]` is better?

does "one-time operation" mean that once a table has been reverted to some point, the table could not be reverted anymore?

if it does, then no.   The reverted table can be reverted / flashback[forward] many times as long as the target point is valid.

I am ok with both names, although the flashback seems to be a little bit "oracleish" to me.

besides,  a force-mode of revert operation (alter table revert/flashback ... FORCE) is planning to be added (in this PR), which does not check if the `point` being reverted to is valid for the table's history (but should check if the snapshot being pointed exist and loadable)

dantengsky:
@BohuTANG  this PR is ready for review, PTAL.

the functionality of flashing back table to arbitrary point is NOT implemented in this PR, I’d like to implement it in another dedicated PR.

the summary of PR updated accordingly.

BohuTANG (approved):

BohuTANG:
@soyeric128 For documentation, thanks.

PR #8979, opened 2022-11-26
fix(storage): compact oom and write not retried

zhyass:

## Summary

Add retry operation when write data. 
Fix out of memory bug.

Closes #issue

### DIFF
--- a/src/query/storages/fuse/fuse/src/operations/mutation/compact/compact_transform.rs
+++ b/src/query/storages/fuse/fuse/src/operations/mutation/compact/compact_transform.rs
@@ -37,6 +37,7 @@ use opendal::Operator;
 use super::compact_meta::CompactSourceMeta;
 use super::compact_part::CompactTask;
 use super::CompactSinkMeta;
+use crate::io::write_data;
 use crate::io::BlockReader;
 use crate::io::TableMetaLocationGenerator;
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):

src/query/storages/fuse/fuse/src/operations/mutation/compact/compact_transform.rs
```diff
@@ -111,7 +112,7 @@ impl CompactTransform {
         thresholds: BlockCompactThresholds,
     ) -> Result<ProcessorPtr> {
         let settings = ctx.get_settings();
-        let max_memory_usage = settings.get_max_memory_usage()?;
+        let max_memory_usage = (settings.get_max_memory_usage()? as f64 * 0.95) as u64;
```
BohuTANG, line 115: `get_max_memory_usage` by default is 0.8 ratio of the total memory:
https://github.com/datafuselabs/databend/blob/3b7872050b0abea7f12570b5dc59da20b06dbff3/src/query/settings/src/lib.rs#L168

```
let max_memory_usage = (settings.get_max_memory_usage()? as f64 * 0.95) as u64;
```
is 0.76, it's make sense.

BohuTANG (approved):

PR #9006, opened 2022-11-29
fix: fix share db bug, create DatabaseIdToName if need

lichuang:

## Summary

fix: Some database has been created before share, so they has no `DatabaseIdToName` for share, create DatabaseIdToName if need.

Closes #issue

### DIFF
--- a/src/meta/api/src/share_api_impl.rs
+++ b/src/meta/api/src/share_api_impl.rs
@@ -600,6 +600,16 @@ impl<KV: KVApi> ShareApi for KV {
                     txn_op_put(&id_key, serialize_struct(&share_meta)?), /* (share_id) -> share_meta */
                     txn_op_put(&object, serialize_struct(&share_ids)?),  /* (object) -> share_ids */
                 ];
+                // Some database has been created before `DatabaseIdToName`, so create it if need.
+                create_db_name_to_id_key_if_need(
+                    self,
+                    &seq_and_id,
+                    &share_name_key.tenant,
+                    &req.object,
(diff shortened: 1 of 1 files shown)


BohuTANG:
Can we add some tests(ut or stateless test) for this PR?

lichuang:
> Can we add some tests(ut or stateless test) for this PR?

Because the case is occur if the shared database is created before share feature implement,  these database leak `DatabaseIdToName` info. After share feature implement, every time create a new database, it will create `DatabaseIdToName ` info.

So it cannot build a test case that create a database leak `DatabaseIdToName`.

BohuTANG:
> > Can we add some tests(ut or stateless test) for this PR?
> 
> Because the case is occur if the shared database is created before share feature implement, these database leak `DatabaseIdToName` info. After share feature implement, every time create a new database, it will create `DatabaseIdToName ` info.
> 
> So it cannot build a test case that create a database leak `DatabaseIdToName`.

How about some stateless tests? 
Likely we don't have any stateless test for share?

lichuang:
> How about some stateless tests? Likely we don't have any stateless test for share?

https://github.com/datafuselabs/databend/blob/main/tests/logictest/suites/base/06_show/06_0009_show_shares

BohuTANG:
I found this stateless test but it only a show function test, can we add a full share function test for it? For example:
1, create a share from A for B
2, B can read from A'share
3, B can not write the A'share
4, A drop the share what's the B will happen?

drmingdrmer (approved):

PR #9009, opened 2022-11-29
feat(query): Support select * exclude [column_name | (col_name, col_name,...)]

### LINKED ISSUE #8964: Feature: selecttarget QualifiedName support EXCLUDE(col)
**Summary**

select a.* [exclude | except] col1 from xx;

select * [exclude | except] (col1, col2, ...)

TCeason:

## Summary

Now support query like this:

```sql
select * exclude col1 from t;
select * exclude (col1, col2) from t;
```

These solution will return err:

1. if col_name in except is duplicate. e.g. `except (id, id)`
2. col_name is not exists in table. e.g. `t.* except (id) from t -- table t does not have id column`
3. col_name is ambiguous. e.g. `except (id) from (select col1 as id, col2 as id)`

Closes #8964


sundy-li:
Keyword `EXCEPT` should not be allowed to use here. It's ambiguous with  union expect query,  let's just use EXCLUDE.

src/query/sql/src/planner/binder/project.rs
```diff
@@ -150,27 +153,150 @@ impl<'a> Binder {
...
+                    exclude,
+                } => {
                     // Handle qualified name as select target
+                    let mut exclude_cols: HashSet<String> = HashSet::new();
+                    if let Some(exclude) = exclude {
+                        match exclude {
+                            ExcludeCol::Col(col) => {
+                                exclude_cols.insert(col.name.clone());
+                            }
+                            ExcludeCol::Cols(cols) => {
```
xudong963, line 167: Why `ExcludeCol::Col` and `ExcludeCol::Cols` are concomitant?  

When `Cols` has only one column, isn't it `Col`?

TCeason: I think we need judge `* exclude (id, id) ` the col is duplicate.

src/query/ast/src/ast/query.rs
```diff
@@ -119,9 +119,31 @@ pub enum SelectTarget<'a> {
         alias: Option<Identifier<'a>>,
     },
 
-    // Qualified name, e.g. `SELECT t.a, t.* FROM t`.
+    // Qualified name, e.g. `SELECT t.a, t.* exclude t.a FROM t`.
     // For simplicity, wildcard is involved.
-    QualifiedName(QualifiedName<'a>),
+    QualifiedName {
+        qualified: QualifiedName<'a>,
+        exclude: Option<ExcludeCol<'a>>,
```
b41sh, line 126: It seems too complicated, can we use `exclude: Option<Vec<Identifier<'a>>>`?

TCeason: I think it's ok.

src/query/sql/src/planner/binder/project.rs
```diff
@@ -150,27 +152,143 @@ impl<'a> Binder {
...
+                    // Pre-check exclude_col is legal
+                    let precheck_exclude_cols = |input_context: &BindContext,
+                                                 exclude_cols: &HashSet<String>,
+                                                 db_name: Option<&Identifier>,
+                                                 table_name: Option<&Identifier>|
+                     -> Result<()> {
+                        let all_columns_bind = input_context.all_column_bindings();
+                        let mut qualified_cols_name: HashMap<String, u8> = HashMap::new();
+                        for i in all_columns_bind {
+                            if let (None, None) = (db_name, table_name) {
```
xudong963, line 180: I recommend to use `match (db_name, table_name) {}`

TCeason: I use this match first. But I think this arm is unnecessary.

```
match  (db_name, table_name) {
   (Some(db_name), None ) => {}
}
```

So I use `if let`.

Xuanwo: > But I think this arm is unnecessary.

Although there is an unnecessary arm, they are more readable than three `if let` cases.

TCeason: > > But I think this arm is unnecessary.
> 
> Although there is an unnecessary arm, they are more readable than three `if let` cases.

Agreed, it's better in terms of readability. Let's change it.

src/query/sql/src/planner/binder/project.rs
sundy-li: Seems  `ExcludeCol::Cols(cols)` can cover `ExcludeCol::Col(col) `, so I recommend to keep `Cols`.

An extra judge for single item vec is trivial, but the code could be cleaner.

TCeason: > Seems `ExcludeCol::Cols(cols)` can cover `ExcludeCol::Col(col) `, so I recommend to keep `Cols`.
> 
> An extra judge for single item vec is trivial, but the code could be cleaner.

Now it already change to

```rust
QualifiedName {
  qualified: QualifiedName<'a>,
  exclude: Option<Vec<Identifier<'a>>>,
},
```

So we can drop struct ExcludeCol

src/query/sql/src/planner/binder/project.rs
TCeason: > I recommend to use `match (db_name, table_name) {}`

Done

src/query/sql/src/planner/binder/project.rs
```diff
@@ -150,27 +152,151 @@ impl<'a> Binder {
...
+                    // Pre-check exclude_col is legal
+                    let precheck_exclude_cols = |input_context: &BindContext,
+                                                 exclude_cols: &HashSet<String>,
+                                                 db_name: Option<&Identifier>,
+                                                 table_name: Option<&Identifier>|
+                     -> Result<()> {
+                        let all_columns_bind = input_context.all_column_bindings();
+                        let mut qualified_cols_name: HashMap<String, u8> = HashMap::new();
+                        for i in all_columns_bind {
+                            match (db_name, table_name) {
```
b41sh, line 180: better to use `match_column_binding` to match columns.

TCeason: Can't use `match_column_binding`. But it gives me inspiration. Now we use match wrap for loop and extract a function named `fill_qualified_cols`

xudong963 (approved):
The `normalize_select_list` method is toooo tedious :/

Better to split it (Doing this in the next PR also makes sense to me)

TCeason:
> The `normalize_select_list` method is toooo tedious :/
> 
> Better to split it (Doing this in the next PR also makes sense to me)

Try to merge * and table.* in the newest commit. Please review it.

src/query/sql/src/planner/binder/project.rs
```diff
@@ -290,3 +308,79 @@ impl<'a> Binder {
...
+) -> Result<()> {
+    let all_columns_bind = input_context.all_column_bindings();
+    let mut qualified_cols_name: HashMap<String, u8> = HashMap::new();
+
+    fn fill_qualified_cols(
+        qualified_cols_name: &mut HashMap<String, u8>,
+        column_bind: &ColumnBinding,
+    ) {
+        let col_name = column_bind.column_name.clone();
+        if qualified_cols_name.contains_key(col_name.as_str()) {
```
b41sh, line 327: we can return `ambiguous column name '{exclude_col}'` here, without contiune check.

TCeason: > we can return `ambiguous column name '{exclude_col}'` here, without contiune check.

Done

src/query/sql/src/planner/binder/project.rs
```diff
@@ -290,3 +308,75 @@ impl<'a> Binder {
...
+    input_context: &BindContext,
+    exclude_cols: &HashSet<String>,
+    db_name: Option<&Identifier>,
+    table_name: Option<&Identifier>,
+) -> Result<()> {
+    let all_columns_bind = input_context.all_column_bindings();
+    let mut qualified_cols_name: HashMap<String, u8> = HashMap::new();
+
+    fn fill_qualified_cols(
+        qualified_cols_name: &mut HashMap<String, u8>,
```
b41sh, line 323: use `HashSet`

b41sh, line 323: use `HashSet`

TCeason: Truly, the value is no used. HashSet is better

TCeason: > use `HashSet`

Fixed

src/query/ast/src/parser/query.rs
```diff
@@ -83,21 +83,48 @@ pub fn with(i: Input) -> IResult<With> {
...
+pub fn exclude_col(i: Input) -> IResult<Vec<Identifier>> {
+    let var = map(
+        rule! {
+            #ident
+        },
+        |col| vec![col],
+    );
+    let vars = map(
+        rule! {
+            "(" ~ ^#comma_separated_list1(ident) ~ ")"
```
andylokandy, line 95:
```suggestion
            "(" ~ ^#comma_separated_list1(ident) ~ ^")"
```

src/query/ast/src/parser/query.rs
```diff
@@ -83,21 +83,48 @@ pub fn with(i: Input) -> IResult<With> {
...
+                exclude: opt_exclude.map(|(_, exclude)| exclude),
+            },
+            Some((fst, _, None)) => SelectTarget::QualifiedName {
+                qualified: vec![Indirection::Identifier(fst), Indirection::Star],
+                exclude: opt_exclude.map(|(_, exclude)| exclude),
+            },
+            None => SelectTarget::QualifiedName {
+                qualified: vec![Indirection::Star],
+                exclude: opt_exclude.map(|(_, exclude)| exclude),
+            },
```
andylokandy, line 127:
```suggestion
        |(res, _, opt_exclude)| {
            let exclude = opt_exclude.map(|(_, exclude)| exclude);
            match res {
                Some((fst, _, Some((snd, _)))) => SelectTarget::QualifiedName {
                    qualified: vec![
                        Indirection::Identifier(fst),
                        Indirection::Identifier(snd),
                        Indirection::Star,
                   ],
                    exclude,
                },
                Some((fst, _, None)) => SelectTarget::QualifiedName {
                    qualified: vec![Indirection::Identifier(fst), Indirection::Star],
                    exclude,
                },
                None => SelectTarget::QualifiedName {
                    qualified: vec![Indirection::Star],
                    exclude,
                },
            }
```

src/query/ast/src/parser/query.rs
TCeason: Done

src/query/ast/src/parser/query.rs
TCeason: Done

b41sh (approved):

TCeason:
cc @soyeric128 Now we can add doc in https://databend.rs/doc/sql-commands/query-syntax/dml-select

PR #9041, opened 2022-11-30
feat(query): push down filter to remain reader.

RinChanNOWWW:

## Summary

After get the bitmap after filtering the prewhere columns, we can push down this bitmap to remain reader.

It help us:

- Skip decompression of some pages. (Not make sense yet because one parquet file only have one page for each column now in databend.)
- Skip decoding of some rows.


RinChanNOWWW:
Fuse engine was refactored and disabled prewhere filter.

PR #9042, opened 2022-11-30
chore(planner): Add `NameAndDataTypes` to replace some usage of `DataSchema`

leiysky:

## Summary

In the new expression, we will use `SchemaDataType` in `DataSchema` to represent table schema, which causes we cannot use `DataSchema` in crates other than storage-related crates.

This PR introduces a new struct `NameAndDataType` to replace `DataSchema` in some crates.


BohuTANG:
Many test failure :/

PR #9050, opened 2022-12-01
feat(fuse): change max_storage_io_requests default value from 1000 to 64

BohuTANG:

## Summary

We already have `adjust_io_request` for the fuse engine, but if we only select a few columns, the max io requests will to 1000. This needs more memory usage, so let's change it to small: 64.

Closes #issue


sundy-li:
@BohuTANG  the failed ut should be fixed.

PR #9055, opened 2022-12-01
fix: try to fix table name case sensibility

### LINKED ISSUE #9049: Bug: table name can't find
**Summary**

`select A.* from (WITH source2 AS (select 1 as e) SELECT * FROM source2) A;`

`A` will transfer to `a`, so the SQL won't find table `A`.

xudong963:

## Summary

`select A.* from (WITH source2 AS (select 1 as e) SELECT * FROM source2) A;`

`A` will transfer to `a` and add to `ColumnBinding`, so the SQL won't find table `A` when `normalize_select_list`.

Introduce `compare_table_name` method to compare table name correctly.

Btw, the ticket also spilt `normalize_select_list` method

Closes #9049

### DIFF
--- a/src/query/sql/src/planner/binder/project.rs
+++ b/src/query/sql/src/planner/binder/project.rs
@@ -17,6 +17,7 @@ use std::collections::HashSet;
 
 use common_ast::ast::Identifier;
 use common_ast::ast::Indirection;
+use common_ast::ast::QualifiedName;
 use common_ast::ast::SelectTarget;
 use common_exception::ErrorCode;
 use common_exception::Result;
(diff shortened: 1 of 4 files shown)


src/query/sql/src/planner/binder/project.rs
```diff
@@ -307,6 +214,142 @@ impl<'a> Binder {
...
+                        scalar: BoundColumnRef {
+                            column: column_binding.clone(),
+                        }
+                        .into(),
+                        alias: column_binding.column_name.clone(),
+                    });
+                }
+            } else if let Some(name) = &column_binding.table_name {
+                if push_item
+                    && compare_table_name(
```
xudong963, line 257: The main changed place

TCeason (approved):

xudong963:
@mergify update

PR #9058, opened 2022-12-01
chore(mem-tracker): rename tracker related terms to reflect their actual usage

drmingdrmer:

## Related Issues


BohuTANG:
@mergify update

src/common/base/src/base/runtime_tracker.rs
```diff
@@ -88,7 +86,19 @@ impl<'a> Drop for Entered<'a> {
 }
 
 /// A guard that resets the `PANICKING` flag when dropped.
-pub struct Panicking;
+pub(crate) struct Panicking;
```
zhang2014, line 89: Maybe rename to UnLimitMemGuard is better?

zhang2014 (approved):

src/common/base/src/base/runtime_tracker.rs
zhang2014: We can modify it in next pr.

PR #9063, opened 2022-12-02
feat(streaming_load): Support file_format syntax in streaming load insert sql

### LINKED ISSUE #8604: Feature:  support `format  (type=ndjson ...)`
we can use it in streaming load.

`(type=ndjson ...)` is the same with the in StageFormatOptions.

1. the format related are written close to each other
2. not need http header or url parameter (only to be compatible with clickhouse)   
     1 .  http header can not be checked
     2.  url parameter need escape. https://github.com/datafuselabs/databend/issues/8243
3.  avoid supporting None in Settings https://github.com/datafuselabs/databend/issues/8541

ariesdevil:

## Summary

Support COPY INTO FILE_FORMAT syntax for the streaming load to unify usage.

Closes #8604

### DIFF
--- a/src/meta/types/src/user_stage.rs
+++ b/src/meta/types/src/user_stage.rs
@@ -182,6 +182,21 @@ impl FromStr for StageFileFormatType {
     }
 }
 
+impl ToString for StageFileFormatType {
+    fn to_string(&self) -> String {
+        match *self {
+            StageFileFormatType::Csv => "CSV".to_string(),
+            StageFileFormatType::Tsv => "Tsv".to_string(),
+            StageFileFormatType::Json => "Json".to_string(),
+            StageFileFormatType::NdJson => "NdJson".to_string(),
+            StageFileFormatType::Avro => "Avro".to_string(),
+            StageFileFormatType::Orc => "Orc".to_string(),
+            StageFileFormatType::Parquet => "Parquet".to_string(),
+            StageFileFormatType::Xml => "Xml".to_string(),
+        }
+    }
+}
+
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
 #[serde(default)]
 pub struct FileFormatOptions {
--- a/src/meta/types/src/user_stage.rs
+++ b/src/meta/types/src/user_stage.rs
@@ -127,6 +127,24 @@ impl FromStr for StageFileCompression {
     }
 }
 
+impl ToString for StageFileCompression {
+    fn to_string(&self) -> String {
+        match *self {
+            StageFileCompression::Auto => "auto".to_string(),
+            StageFileCompression::Gzip => "gzip".to_string(),
+            StageFileCompression::Bz2 => "bz2".to_string(),
+            StageFileCompression::Brotli => "brotli".to_string(),
+            StageFileCompression::Zstd => "zstd".to_string(),
+            StageFileCompression::Deflate => "deflate".to_string(),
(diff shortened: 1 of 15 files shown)


src/query/ast/src/parser/statement.rs
```diff
@@ -1078,12 +1078,23 @@ pub fn statement(i: Input) -> IResult<StatementMsg> {
 pub fn insert_source(i: Input) -> IResult<InsertSource> {
     let streaming = map(
         rule! {
-            FORMAT ~ #ident ~ #rest_str
-        },
-        |(_, format, (rest_str, start))| InsertSource::Streaming {
-            format: format.name,
-            rest_str,
-            start,
+            FORMAT ~ #ident ~ #rest_str ~ (FILE_FORMAT ~ "=" ~ #options)?
```
youngsofun, line 1081: what about `FORMAT (~ #ident |  "=" ~ #options ) ~ #rest_str ` ? cc @sundy @xuanwo

src/query/settings/src/lib.rs
```diff
@@ -491,6 +492,38 @@ impl Settings {
         }))
     }
 
+    pub fn set_file_format_options(&self, opts: &FileFormatOptions) -> Result<()> {
```
youngsofun, line 495: when we already have FileFormatOptions, should try to avoid using settings (our final goal)

ariesdevil: This is for compatibility, if we decide to remove the old syntax, this will be removed future.

youngsofun: we should only support new syntax internally,  and  try to convert old syntax to new.

youngsofun: FileFormatOptions -> setting is the reverse

src/meta/types/src/user_stage.rs
```diff
@@ -164,6 +182,21 @@ impl FromStr for StageFileFormatType {
     }
 }
 
+impl ToString for StageFileFormatType {
+    fn to_string(&self) -> String {
+        match *self {
```
youngsofun, line 187: what about `format!("{:?}", *self)`

src/query/service/src/servers/http/clickhouse_handler.rs
```diff
@@ -290,6 +290,52 @@ pub async fn clickhouse_handler_post(
...
+                .await
+                .map_err(InternalServerError)?;
+
+            // override settings
+            settings
+                .set_file_format_options(option_settings)
+                .map_err(InternalServerError)?;
+
+            let input_context = Arc::new(
+                InputContext::try_create_from_insert(
```
youngsofun, line 323: we need a new `try_create_from_insert_v2` which do not use settings to pass options.

youngsofun: you can refer to try_create_from_copy

src/query/settings/src/lib.rs
youngsofun: we should only support new syntax internally, and try to convert old syntax to new.

it is also ok to keep both code path for now.
but new syntax should not based on the old,  so we need InputContext::try_create_from_insert_v2

src/query/service/src/servers/http/clickhouse_handler.rs
ariesdevil: good idea, with `try_create_from_insert_v2` we can avoid use settings.

BohuTANG:
Please add more summary about this PR:)

youngsofun (approved):

PR #9074, opened 2022-12-02
fix: Add more tests for hive config loading

Xuanwo:

## Summary

fix: Add more tests for hive config loading

We should not load args during our unit tests, because clap will try to parse the input for `cargo test`.

### DIFF
--- a/src/query/config/src/inner.rs
+++ b/src/query/config/src/inner.rs
@@ -71,6 +71,14 @@ impl Config {
         Ok(cfg)
     }
 
+    /// # NOTE
+    ///
+    /// This function is served for tests only.
+    pub fn load_without_args() -> Result<Self> {
+        let cfg: Self = OuterV0Config::load_without_args()?.try_into()?;
+        Ok(cfg)
+    }
+
     pub fn tls_query_cli_enabled(&self) -> bool {
         !self.query.rpc_tls_query_server_root_ca_cert.is_empty()
             && !self.query.rpc_tls_query_service_domain_name.is_empty()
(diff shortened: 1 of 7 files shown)


src/query/config/src/inner.rs
```diff
@@ -71,6 +71,14 @@ impl Config {
         Ok(cfg)
     }
 
+    /// # NOTE
+    ///
+    /// This function is served for tests only.
+    pub fn load_without_args() -> Result<Self> {
```
ClSlaid, line 77: If this is only used in tests, can we add an `#[cfg(test)]` for it?

src/query/config/src/outer_v0.rs
```diff
@@ -158,6 +158,40 @@ impl Config {
...
+
+    /// Load will load config from file, env and args.
+    ///
+    /// - Load from file as default.
+    /// - Load from env, will override config from file.
+    ///
+    /// # NOTE
+    ///
+    /// This function is only used for tests.
+    pub fn load_without_args() -> Result<Self> {
```
ClSlaid, line 170: can this be configured with `#[cfg(test)]?

src/query/config/src/inner.rs
Xuanwo:
```rust
error[E0599]: no function or associated item named `load_for_test` found for struct `databend_meta::configs::Config` in the current scope
  --> src/meta/service/tests/it/configs.rs:71:27
   |
71 |         let cfg = Config::load_for_test().expect("load must success");
   |                           ^^^^^^^^^^^^^ function or associated item not found in `databend_meta::configs::Config`
```

Not works for tests that not in the same crate.

ClSlaid: > Not works for tests that not in the same crate.

At least, we learn more about rust.

BohuTANG:
We can do more unit tests for config, we can not do it because `clap will try to parse the input for cargo test` in the past.

Issue #9018, opened 2022-11-29
bug: index out of bounds


hanyisong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

8.0.26-v0.8.114-nightly

### What's Wrong?

The case is from sqlancer.
```sql
MySQL [(none)]> select 1 as count FROM t0 NATURAL RIGHT JOIN t1;
ERROR 1105 (HY000): result row write failed: Code: 1104, displayText = index out of bounds: the len is 1 but the index is 2.
```

### How to Reproduce?

```sql
DROP DATABASE IF EXISTS databend1;
CREATE DATABASE databend1;
USE databend1;
CREATE TABLE t0(c0VARCHAR VARCHAR NULL, c1BOOLEAN BOOLEAN NULL, c2BOOLEAN BOOLEAN NULL);
CREATE TABLE t1(c0INT INT64 NULL DEFAULT(-273272402), c1INT INT64 NULL);
select 1 as count FROM t0 NATURAL RIGHT JOIN t1;
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

xudong963:
@lichuang wants to try to fix the issue.

PR #9081, opened 2022-12-03
fix: fix right join bug, issue 9018

lichuang:

## Summary

fix: fix right join bug, issue 9018

Closes https://github.com/datafuselabs/databend/issues/9018


src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_state_impl.rs
```diff
@@ -294,6 +294,9 @@ impl HashJoinState for JoinHashTable {
         }
 
         let rest_block = self.rest_block()?;
+        if rest_block.is_empty() {
```
xudong963, line 297: Why `rest_block` is empty, then you directly finish the method?

xudong963 (approved):

Issue #9065, opened 2022-12-02
bug: segment fault in Settings::default_settings


dantengsky:
**Summary**

A user provided coredump shows that query node SEGMENT FAULTed.

binary: https://repo.databend.rs/databend/v0.8.129-nightly/databend-v0.8.129-nightly-x86_64-unknown-linux-musl.tar.gz
os: centos 7

seems caused by multi-threaded calling  of `getgrgid` ..

tailored bt
~~~
#0  a_crash () at ../src_musl/arch/x86_64/atomic_arch.h:108
#1  unmap_chunk () at ../src_musl/src/malloc/malloc.c:515
#2  free () at ../src_musl/src/malloc/malloc.c:526
#3  0x00007fbc1cf2bbaa in __getgrent_a () at ../src_musl/src/passwd/getgrent_a.c:45
#4  0x00007fbc1cf35fee in __getgr_a () at ../src_musl/src/passwd/getgr_a.c:35
#5  0x00007fbc1cf2ba8f in getgrgid () at ../src_musl/src/passwd/getgrent.c:29
#6  0x00007fbc1ca594c0 in sysinfo::users::get_users_list::{closure#1}::{closure#0} () at src/users.rs:64
....
#31 common_settings::Settings::default_settings () at src/query/settings/src/lib.rs:168
#32 0x00007fbc1b9a6f19 in common_settings::{impl#1}::try_create::{async_fn#0} () at src/query/settings/src/lib.rs:84
#33 core::future::from_generator::{impl#1}::poll<common_settings::{impl#1}::try_create::{async_fn_env#0}> () at /rustc/e631891f7ad40eac3ef58ec3c2b57ecd81e40615/library/core/src/future/mod.rs:91
#34 databend_query::sessions::session_mgr::{impl#0}::create_session::{async_fn#0} () at src/query/service/src/sessions/session_mgr.rs:127
....
#134 0x00007fbc1cf31971 in start () at ../src_musl/src/thread/pthread_create.c:192
#135 0x00007fbc1cf32ea1 in __clone () at ../src_musl/src/thread/x86_64/clone.s:22

~~~

**update**

- for version 7.xxx no such segmentfault occurred
- segmentfault occurred for both v8.123 & 8.129
   and the location of crash seems to be the same  (only the IP left in var message, no other coredump files):
   #0 a_crash () at ../src_musl/arch/x86_64/atomic_arch.h:108

dantengsky:
full bt
~~~
[...]

PR #9086, opened 2022-12-05
fix: unit of max_memory_usage should be B (not kB)

dantengsky:

## Summary

in PR #9065, the unit of total memory changed to kB incorrectly, this PR fix it.

Closes #issue


src/query/settings/src/lib.rs
```diff
@@ -157,7 +157,8 @@ impl Settings {
             },
             // max_memory_usage
             SettingValue {
-                default_value: UserSettingValue::UInt64(memory_info.total * 80 / 100),
+                // unit of memory_info.total is kB
+                default_value: UserSettingValue::UInt64(1024 * memory_info.total * 80 / 100),
```
BohuTANG, line 161: Not sure this display is always KB not depends on OS?

dantengsky: for other platforms, the value will be convert to Kb

dantengsky: e.g.

windows:
https://github.com/FillZpp/sys-info-rs/blob/master/c/windows.c#L106-L113

darwin:

https://github.com/FillZpp/sys-info-rs/blob/master/c/darwin.c#L108-L134

BohuTANG (approved):

PR #9090, opened 2022-12-05
fix(query): fix state merge bug for aggregation

### LINKED ISSUE #9010: bug:  sql generate different result at each attempt  in cluster mode.
### Search before asking

### Version

1340868f7c82eb353fd02e09a008ff864c941a9b

### What's Wrong?

sql may generate different  wrong result in cluster mode at different attempts.  while sql result is right in local cluster.

### How to Reproduce?

sql
```
SELECT count(*) AS sqlrewriter0
FROM
  (SELECT product_name,
          sum(cost_total) AS t
   FROM xxx
   WHERE p_date = '20221110'
   GROUP BY product_name
   HAVING t > 100000) AS a
ORDER BY sqlrewriter0
LIMIT 100
;
```

sundy-li:

## Summary

Summary about this PR

Aggregation state should be aligned to the nested state alignment for null adaptors.

I'm trying to have a test to cover this bug.

Closes #9010


sundy-li:
@mergify update

RinChanNOWWW (approved):

zhang2014 (approved):

PR #9105, opened 2022-12-06
fix: fuse_snapshot with limits may return less items

### LINKED ISSUE #9102: bug: fuse_snapshot with limit return wrong result
### Search before asking

### Version

v0.8.142

### What's Wrong?

```
MySQL [(none)]> select count(*) from fuse_snapshot('default','tb01');
+----------+
| count(*) |
+----------+
|     7326 |
+----------+
1 row in set (11.273 sec)

MySQL [(none)]> select snapshot_id from fuse_snapshot('default','tb01') limit 10;
+----------------------------------+
| snapshot_id                      |
+----------------------------------+
| 7a0cd445833b4cd6801187e0afaf7f23 |
+----------------------------------+
1 row in set (0.853 sec)
```

### How to Reproduce?

_No response_

dantengsky:

## Summary

for limited selection from fuse_snapshot like this
~~~
select snapshot_id from fuse_snapshot('default','tb01') limit 10;
~~~

the `SnapshotIO` being used will collect a limited number of snapshots first, and then, try to build a chain of snapshots from them.  if the snapshots collected are not of a consecutive segment of the history chain, the number of items of the final resultset will not be as expected.

In this PR, for `select ... form fuse_snapshot(..) limit ...`,   `SnapshotHistoryReader` is used instead, which collects the history of snapshots by traversing the snapshot (instead of listing from object storage).

Closes #9102

### DIFF
--- a/src/query/storages/fuse/fuse/src/table_functions/fuse_snapshots/fuse_snapshot.rs
+++ b/src/query/storages/fuse/fuse/src/table_functions/fuse_snapshots/fuse_snapshot.rs
@@ -18,7 +18,11 @@ use common_datablocks::DataBlock;
 use common_datavalues::prelude::*;
 use common_exception::Result;
 use common_storages_table_meta::meta::TableSnapshotLite;
+use futures::stream::StreamExt;
+use futures::stream::TryStreamExt;
 
+use crate::io::MetaReaders;
+use crate::io::SnapshotHistoryReader;
 use crate::io::SnapshotsIO;
 use crate::io::TableMetaLocationGenerator;
 use crate::sessions::TableContext;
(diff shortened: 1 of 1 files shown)


wubx:
LGTM

src/query/storages/fuse/fuse/src/table_functions/fuse_snapshots/fuse_snapshot.rs
```diff
@@ -45,16 +49,45 @@ impl<'a> FuseSnapshot<'a> {
...
+                    Ok(snapshots)
+                }
+                Some(l) => {
+                    // SnapshotHistoryReader (which TableSnapshotReader impls) traverses the history
+                    // of snapshot sequentially, by using the TableSnapshot::prev_snapshot_id, which
+                    // guarantees the snapshot returned can be chained together
+                    let table_snapshot_reader = MetaReaders::table_snapshot_reader(
+                        self.ctx.get_data_operator()?.operator(),
+                    );
+                    table_snapshot_reader
```
BohuTANG, line 77: It seems read all snapshot first one by one and then take the limit? If so, ten thousand snapshots will take a few minutes if limit 10.

dantengsky: not sure about this.  
usually, fetching tens of small files from object storage sequentially takes hundreds of ms. if the value of the limitation is small, maybe traversing them by `prev_snapshot_id` is better?

dantengsky: oh... I misinterpreted your comment.

> If so, ten thousand snapshots will take a few minutes with limit 10.

no, it will not. in this case, only 10 snapshots will be traversed, from the latest one to the 10th. SnapshotHistoryReader returns a stream(which is lazy), and stream::take(10) only takes the first 10 elements from the stream.

BohuTANG (approved):

PR #9111, opened 2022-12-07
fix: prewhere_column empty and predicate is not const will return empty

### LINKED ISSUE #9083: bug: The where clause of the select statement is wrong
### Search before asking

### Version

8.0.26-v0.8.114-nightly

### What's Wrong?

row is not fetched.

### How to Reproduce?

```sql
DROP DATABASE IF EXISTS databend6;
CREATE DATABASE databend6;
USE databend6;
CREATE TABLE t0(c0INT INT32 NOT NULL DEFAULT(1456832334));
INSERT INTO t0(c0int) VALUES (1388388634), (-1943680716);
SELECT t0.c0int FROM t0 WHERE ((((((true)or(('s' NOT IN ('oE', '70', '3r', 'k', '9aRgze')))))AND(true)))or(((NULL)<=(NULL))));
```

TCeason:

## Summary

```sql
select * from t1 where 's' not in ('x', 'b', 'c', 'd') or to_nullable(null);
-- return empty block
```

if in function args len()>3, the filter will be set with [(not(in('s', (x, b, c, d))) or NULL).

Now in this query prewhere_column is empty , the data_block will be empty; and because of NULL the predicate will be nullable(bool).

So, in cast_to_nonull_boolean will return boolean columnRef.

So in filter_exists, the count_zeros and rows is 0. 

the (!filter_exists) will return true, so generate on empty block.

## Fix

I think we have two ways to fix this:

1. if prewhere_column is empty, logical.prewhere = None.

2. In this pr, if count_zeros=rows=0, filter_exists returns true.

But I think the first way is easier and if prewhere_column maybe we shouldn't set the prewhere? 

Closes #9083

### DIFF
--- a/src/query/datablocks/src/kernels/data_block_filter.rs
+++ b/src/query/datablocks/src/kernels/data_block_filter.rs
@@ -33,7 +33,12 @@ impl DataBlock {
         let boolean_col: &BooleanColumn = Series::check_get(&predict_boolean_nonull)?;
         let rows = boolean_col.len();
         let count_zeros = boolean_col.values().unset_bits();
-        Ok(count_zeros != rows)
+        if (count_zeros == 0) && (rows == 0) {
+            // prewhere_column is empty and predicate is nullable
+            Ok(true)
+        } else {
+            Ok(count_zeros != rows)
+        }
     }
 
     pub fn filter_block(block: DataBlock, predicate: &ColumnRef) -> Result<DataBlock> {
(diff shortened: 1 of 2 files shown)


RinChanNOWWW:
Oh, I have fixed this bug in#8963, but this PR is still WIP.

TCeason:
> Oh, I have fixed this bug in #8963, but this PR is still WIP.
> 
> ```rust
> get.prewhere = if prewhere_pred.is_empty() {
>     None
> } else {
>     if prewhere_columns.is_empty() {
>         prewhere_columns = get.columns.clone();
>     }
>     Some(Prewhere {
>         output_columns: get.columns.clone(),
>         prewhere_columns,
>         predicates: prewhere_pred,
>     })
> };
> ```

Ok, let’s close this pr.

RinChanNOWWW:
You can fix it now, because #8963 is stalled now.

PR #9118, opened 2022-12-07
chore: make inner `Config` in crate `databend-query` as a global singeleton.

### LINKED ISSUE #9114: Make inner `Config` in crate `databend-query` as a global singleton.
**Summary**

The inner `Config` of `databend-query`  (src/query/config/src/inner.rs) is immutable during runtime. We can make it as a global singleton.

So that we don't need context when we want to get the config and we can get it everywhere in crate `databend-query`.

RinChanNOWWW:

## Summary

Make inner Config in crate databend-query as a global singeleton.

Get the config by `GlobalConfig::instance()`.

Closes #9114


Xuanwo (approved):

PsiACE (approved):

src/query/config/src/global.rs
```diff
@@ -0,0 +1,40 @@
...
+use once_cell::sync::OnceCell;
+
+use crate::Config;
+
+pub struct GlobalConfig;
+
+static GLOBAL_CONFIG: OnceCell<Singleton<Arc<Config>>> = OnceCell::new();
+
+impl GlobalConfig {
+    pub fn init(config: Config, v: Singleton<Arc<Config>>) -> Result<()> {
```
BohuTANG, line 28: Can we add a unit test for the singleton?

src/query/service/tests/it/tests/sessions.rs
```diff
@@ -512,10 +512,8 @@ impl SingletonImpl<Arc<Config>> for TestGlobalServices {
     fn init(&self, value: Arc<Config>) -> Result<()> {
         match std::thread::current().name() {
             None => panic!("thread name is none"),
-            Some(name) => match self.query_config.lock().entry(name.to_string()) {
-                Entry::Vacant(v) => v.insert(value),
-                Entry::Occupied(_v) => panic!("GlobalConfig set twice in test[{:?}]", name),
-            },
+            // allow overwrite config in test.
```
Xuanwo, line 515: Config should not be overwritten, it must be buggy.

RinChanNOWWW: The unit test `test_abort_on_error` will init multiple times.

src/query/config/src/global.rs
RinChanNOWWW: What behaviors do we need to test?

src/query/service/tests/it/tests/sessions.rs
BohuTANG: The unit test should handle the case. This why I comment above we should add a unit test.

Xuanwo: > The unit test `test_abort_on_error` will init multiple times and will lead to panic.

The test should be refactored.

src/query/config/src/global.rs
BohuTANG: An important test is to initialize multiple times and to ensure that the singleton is correct and works as well.

sundy-li (approved):

PR #9123, opened 2022-12-07
feat(stage): support `select from @a_stage`

youngsofun:

## Summary

done:

1. Add stage as TableReference, see the test.
2. StageTable.read_partiton() return file splits instead of files.
3. copy and select use StageTable.read_data() to build a pipeline

as a POC, there are some hard codes and other simplifications, which will be fixed later in other PRs.

1. only support Parquet.
2. only support name stage.
4. not support push down
5. not support cluster


BohuTANG (approved):

BohuTANG:
Conflicting files
src/query/catalog/src/plan/pushdown.rs

sundy-li (approved):

sundy-li:
we can bench this stage table with read_parquet table function.

Let's see the difference and improvement.

Issue #9038, opened 2022-11-30
Feature:  range filter not support "in"


sandflee:
**Summary**

hive sql use range filter to prune partitions, runing sql like  select * from xx where a in (xx,xx,xx);  range filter could't take affect and may read all partition datas.

cc @sundy-li

sundy-li:
Ok, range filter will be rewritten after we have the new expression by @andylokandy 

Please take your time with this feature.

sandflee:
hi, @andylokandy , is any  expected time to do this?

andylokandy:
It should be done at the time when `expression` branch is merged into main.

sandflee:
so, when the expression branch is merged into main ?

Issue #8196, opened 2022-10-13
Optimize BlockFilter creation


drmingdrmer:
**Summary**

BlockFilter building takes a lot time.

Add benchmark of current xor8 filter building for `i64` and `string` of
length 16 to 32.

Benchmark shows that with `Column::to_values()`, which potentially is an
unnecessary overhead, the time cost for every key is:
```
- i64:    210ns/key
- string: 240ns/key
```

And the benchmark of xor8-filter shows that building a `i64` key takes
about 180 ns/key, which means without modifying xor8-filter, the max
performance boost would be about 25%.

![img_v2_0c93db10-cb5b-44ba-9878-5d1693bc69ag](https://user-images.githubusercontent.com/44069/195539565-34031905-5f35-4257-83cb-b6af01d60c3f.png)

TODO:

- [x] #8225 
- [x] Avoid using `Column::to_values()`, which allocates a lot memory (https://github.com/datafuselabs/databend/pull/9053)
- [ ] #8233

BohuTANG:
I think we can close this issue now :)

PR #9127, opened 2022-12-07
feat(expr): migrate storages/index

andylokandy:

## Summary

- Reimplement `RangeFilter` and `BloomFilter` with the new expression in the `expression` branch,
- Reimplement the tests and benchmark.

The benchmark result is 2x faster than the original implementation which seems too good to be true. 🤔  

Closes https://github.com/datafuselabs/databend/issues/9038

### DIFF
--- a/src/query/storages/index/benches/build_from_block.rs
+++ b/src/query/storages/index/benches/build_from_block.rs
@@ -15,13 +15,9 @@
 #[macro_use]
 extern crate criterion;
 
-use common_datablocks::DataBlock;
-use common_datavalues::prelude::*;
-use common_datavalues::DataField;
-use common_datavalues::DataSchemaRefExt;
-use common_datavalues::Series;
-use common_datavalues::SeriesFrom;
-use common_storages_index::filters::Filter;
+use common_expression::types::number::NumberColumn;
+use common_expression::types::string::StringColumnBuilder;
+use common_expression::Column;
 use common_storages_index::filters::FilterBuilder;
 use common_storages_index::filters::Xor8Builder;
 use criterion::Criterion;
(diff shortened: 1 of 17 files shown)


andylokandy:
The test should fail anyway. So just don't be bothered with it.

src/query/storages/index/src/bloom.rs
```diff
@@ -53,193 +56,185 @@ use crate::SupportedType;
...
+                // create filter per column
+                let mut filter_builder = Xor8Builder::create();
+
+                // ingest the same column data from all chunks
+                for chunk in chunks {
+                    let entry = chunk.get_by_offset(i);
+                    match &entry.value {
+                        Value::Scalar(scalar) => filter_builder.add_key(&scalar),
+                        Value::Column(col) => {
+                            col.iter()
```
sundy-li, line 110: col.iter() will generate temp Enum ScalarRef which may lead to poor performance.

Maybe we could use `siphash` function to generate U64Column and add the u64 keys ?

It could be addressed after new expression is merged.

src/query/storages/index/benches/build_from_block.rs
```diff
@@ -38,86 +34,70 @@ use rand::SeedableRng;
...
 
 fn bench_u64(c: &mut Criterion) {
-    let block = rand_i64_block(1_000_000);
-    let column = block.try_column_by_name("a").unwrap();
-
-    let mut builder = Xor8Builder::create();
-    builder.add_keys(&column.to_values());
-    let filter = builder.build().unwrap();
-
-    for key in column.to_values() {
```
sundy-li, line 50: maybe you could use :

```
column.for_each
```

to update this benchmark on your machine.

sundy-li:
bench in my env
```
main without to_values: 91ns, 108ns
exprindex:    110ns, 106ns
stdhash to u64: 88ns, 92ns
stdhash to u64 and xor filter with no hash:  78ns, 84ns
```

src/query/storages/index/src/bloom.rs
drmingdrmer: > col.iter() will generate temp Enum ScalarRef which may lead to poor performance (hash of enum).

Does the performance penalty is caused by the `match` statement of the enum variants?
Are there any other factors related to the enum that can impact performance?

sundy-li: #8196

sundy-li (approved):
LGTM. But `new_filter_pruner` is not finished, should we enable it in the next pr ?

Issue #8649, opened 2022-11-05
feat: hive release build


BohuTANG:
**Summary**

Add a new release type with hive:
```
databend-hive-v0.8.126-nightly-x86_64-unknown-linux-musl.tar.gz
```

Xuanwo:
Let me take a look.

BohuTANG:
I have updated the issue. Let's build a new hive release type first. cc @Xuanwo

PR #9137, opened 2022-12-07
ci: try to add hive release

PsiACE:

## Summary

- follow the process of linux releases in databend-release.yml, try to add hive release
  - only for x86 & gnu
  - runs-on: [self-hosted, X64, Linux, development]
  - with databend-hive-* prefix
- add hive release build ci to production.yml

Closes #issue

### DIFF
--- a/.github/workflows/databend-release.yml
+++ b/.github/workflows/databend-release.yml
@@ -215,6 +215,73 @@ jobs:
           target: ${{ steps.target.outputs.target }}
           repo_role_arn: ${{ secrets.REPO_ROLE_ARN }}
 
+  publish_hive:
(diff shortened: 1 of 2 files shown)


PsiACE:
Ref https://github.com/datafuselabs/databend/issues/8649, it looks like we need to build the musl version, so I've switched the development and production targets for hive to musl and hope that it all works out.

PsiACE:
Let's start with the hive release of gnu? It looks like we need to use the musl version of the JVM in order to build the musl release successfully.

everpcpc (approved):
I think there's no need for us to provide a musl version, which is only targeted on trial propose.

PR #9141, opened 2022-12-08
feat(query): sink fuse table process in multiple threads

### LINKED ISSUE #9089: performance: build BlockFilter and BlockStatistics in parallel
**Summary**

When inserting data, we will sequentially execute three operations `build BlockFilter`, `build BlockStatistics`, and `serialize_data_blocks` in `FuseTableSink`. If the number of inserted data rows is large, it will be time-consuming.
The performance of insertion can be improved by performing these operations in parallel or asynchronously.

image

b41sh:

## Summary

Build `BloomIndexState` and `BlockStatistics` in separate threads. After optimization, the time of inserting 100M values will be reduced by about 6s.

Closes #9089

### DIFF
--- a/src/query/storages/fuse/fuse/src/operations/fuse_sink.rs
+++ b/src/query/storages/fuse/fuse/src/operations/fuse_sink.rs
@@ -13,7 +13,9 @@
 //  limitations under the License.
 
 use std::any::Any;
+use std::sync::mpsc::channel;
 use std::sync::Arc;
+use std::thread;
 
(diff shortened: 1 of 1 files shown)


BohuTANG:
I have updated the title, avoid to  use `chore`, it will not show in the changelog :)

b41sh:
Since the pipelines are multi-threaded, adding multiple threads here may cause additional performance loss, so close this PR first.

PR #9144, opened 2022-12-08
chore(query): fix nan in range filter

### LINKED ISSUE #9142: bug:  range filter drops datablocks with max value 'Nan'
### Search before asking

### Version

xxx

### What's Wrong?

a parquet column named `a` with min value '1.0', max value 'NaN',  
range filter with express `a > 0.9` drops the block by mistake.

### How to Reproduce?

_No response_

sundy-li:

## Summary

Summary about this PR

fix nan in range filter

Closes #9142

### DIFF
--- a/src/query/storages/index/src/range_filter.rs
+++ b/src/query/storages/index/src/range_filter.rs
@@ -288,10 +288,22 @@ impl StatColumn {
                 single_point = false;
             }
 
-            let min_col = v.data_type().create_constant_column(&stat.min, 1)?;
+            let min = if let DataValue::Float64(f64::NAN) = stat.min {
+                &DataValue::Float64(f64::MIN)
+            } else {
+                &stat.min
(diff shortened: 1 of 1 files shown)


src/query/storages/index/src/range_filter.rs
```diff
@@ -288,10 +288,22 @@ impl StatColumn {
                 single_point = false;
             }
 
-            let min_col = v.data_type().create_constant_column(&stat.min, 1)?;
+            let min = if let DataValue::Float64(f64::NAN) = stat.min {
```
BohuTANG, line 291: Add a stateless test for it?

sundy-li: I just think about how to create `nan` index in stateless test, currently no idea.

BohuTANG (approved):

PR #9150, opened 2022-12-08
feat: introduce rust native sqllogictest framework

xudong963:

## Summary

The ticket aims to implement rust native sqllogictest framework and replace current python version.

**Reasons**:
1. Unify our codebase, all in rust!
2. The maintainer of the python version has left databend, the python version is a little difficult to maintain.
3. [sqllogictest-rs](https://github.com/risinglightdb/sqllogictest-rs) becomes more and more stable, the guys are talented, and we can work with them to make sqllogictest-rs better

**Todos**:
-  [x] refactor code: walking dir to read all test files, remove hard code and etc.
-  [x] change current logictest file's syntax, make it standard: such as `statement query II` -> `query II`.
-  [x] add args to support run specific test dir or file.
-  [x] make all tests can be run under new sqllogictest framework.
-  [x] add it to CI

**After finishing the above todos, I'll make it ready for review (maybe the weekend)**

**Future todos**:

See https://github.com/datafuselabs/databend/issues/9174

Closes #issue


Xuanwo:
I like this!

tests/sqllogictests/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+futures-util = "0.3.25"
+mysql = "23.0.0"
+reqwest = { version = "0.11.13", features= ["json"]}
+serde_json = "1.0.89"
+
+
+
+
+
+
```
Xuanwo, line 23: trim empty lines.

tests/sqllogictests/suits/test.slt
```diff
@@ -0,0 +1,29 @@
+statement ok
```
Xuanwo, line 1: Does this means we need to port all our logic test to `slt`?

Xuanwo:
Another question is how our users (like databend cloud) reuse our logictest suites. They will run sqllogic test by themselves.

tests/sqllogictests/suits/test.slt
xudong963: Just change toriginal file name to .slt and adjust the content syntax

tests/sqllogictests/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+publish = false
+version = "0.1.0"
+
+[dependencies]
+common-exception = { path = "../../src/common/exception" }
+tokio = { version = "1.21.1", features = ["full"] }
+sqllogictest = "0.9.0"
+async-trait = "0.1.53"
+futures-util = "0.3.25"
+mysql = "23.0.0"
```
PsiACE, line 15: mysql_async

xudong963: I have tried it but met some blocks, so I chose the crate.

For tests, I think the two crates don't have too much difference.  Any thoughts? @PsiACE

tests/sqllogictests/suits/test.slt
BohuTANG: '.slt' is strange, can we change it?

Xuanwo: I think it makes sense to support different file extension in `sqllogictest-rs`, we can submit an issue and implement it.

Xuanwo: We do not check file extensions. You can use anything as filename :P

_Originally posted by @xxchan in https://github.com/risinglightdb/sqllogictest-rs/issues/118#issuecomment-1342878713_

xudong963: Yes, both are ok, will use `.test`.

xudong963:
https://github.com/datafuselabs/databend/pull/9150/commits/0b7305afcbe97b3bad970b44cea1e184135e6b17 The commit fixed test files format and is large. So you can ignore it when reviewing.

xudong963:
Support to run specific dir or test files:
```shell
➜  databend git:(new_sqllogictest) ✗ cargo run -p sqllogictests -- --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/sqllogictests --help`
Mysql client starts to run...
Usage: sqllogictests [OPTIONS]

Options:
  -d, --dir <DIR>    Run sqllogictests in specific directory, the arg is optional
  -f, --file <FILE>  Run sqllogictests in specific test file, the arg is optional
  -h, --help         Print help information
```

xudong963:
> > Support to run specific dir or test files:
> 
> sqllogictest has a binary which supports pg, can we share our work with them? I think we only need to maintain the HTTP API part.

Do you mean let sqllogictest-rs binary support mysql?

Xuanwo:
> Do you mean let sqllogictest-rs binary support mysql?

Yes, we can run the binary directly instead of `cargo run` which need to build the runner everytime.

xudong963:
> > Do you mean let sqllogictest-rs binary support mysql?
> 
> Yes, we can run the binary directly instead of `cargo run` which need to build the runner everytime.

This can be advanced afterwards, perhaps until their next release, my efforts now are mainly focused on getting the new framework running inside the CI.

I opened an issue last night: https://github.com/risinglightdb/sqllogictest-rs/issues/120. Should be part of your proposal

tests/sqllogictests/Cargo.toml
xudong963: The upstream seems to fix the bug,  I'll try again. https://github.com/blackbeam/mysql_async/releases/tag/v0.31.2

Xuanwo:
image

Respect! :bow:

Xuanwo:
I think `add ability to process Regex` can be moved to next PR.

xudong963:
> I think `add ability to process Regex` can be moved to next PR.

Yes, it's very tricky, I think we should find a more suitable test framework to support it.

xudong963:
> image
> 
> Respect! 🙇

Still fighting! HTTP and ck HTTP client also need to check, I think there are still many blocks.

scripts/ci/ci-run-sqllogic-tests.sh
```diff
@@ -7,23 +7,21 @@ set -e
 echo "Starting standalone DatabendQuery and DatabendMeta"
 ./scripts/ci/deploy/databend-query-standalone.sh
 
-SCRIPT_PATH="$(cd "$(dirname "$0")" >/dev/null 2>&1 && pwd)"
-cd "$SCRIPT_PATH/../../tests/logictest" || exit
-
-TEST_HANDLERS=${TEST_HANDLERS:-"mysql,http,clickhouse"}
+echo "Install rust"
+./scripts/setup/dev_setup.sh
```
xudong963, line 11: I need to use `Cargo` in the file, is it a good place to install `./scripts/setup/dev_setup.sh`? cc @everpcpc

PsiACE: I think we can assume that the environment is already set up

Xuanwo: It's better to call `carge` directly and run `sqllogic-tests` with build-tools like unit test:

https://github.com/datafuselabs/databend/blob/main/.github/actions/test_unit/action.yml#L8-L27

Xuanwo: Since we don't need python anymore, we can remove the usage of `datafuselabs/build-tool:sqllogic`.

xudong963: Let me have a try. Thanks

Xuanwo: I'm willing to provide any help you need. I can update the workflow files if that blocks you. Just leave me a call~ :love_letter:

.github/actions/test_sqllogic_standalone_linux/action.yml
```diff
@@ -26,20 +26,14 @@ runs:
...
         target: ${{ inputs.target }}
-
-    - name: Run sqllogic Tests with Standalone mode
+    - name: Setup Build Tool
+      uses: ./.github/actions/setup_build_tool
+      with:
+        bypass_env_vars: RUSTFLAGS,RUSTDOCFLAGS,RUST_TEST_THREADS,RUST_LOG,RUST_BACKTRACE
+    - name:  Run sqllogic tests with Standalone mode
       shell: bash
+      run: cargo run -p sqllogictests -- --skip_dir cluster
```
xudong963, line 35: I didn't follow the handler and suit separately, maybe we can do that later? cc @Xuanwo

xudong963:
image

It seems like a flaky test, but it can't reproduce in my local... (Btw, it's normal in old logictest)

BohuTANG:
@xudong963 

If this one test is blocked for a while, we can file a tracking issue and skip it in this PR.

xudong963 (commented):
Now the ticket is ready for review! Its changes contain three parts: **CI**, **logictest files** and **new sqllogictest framework**. The following is a summary of the ticket(hope it will help reviewers):
1. For CI, the main changes are removing some files used by old framework and changing the commands that run new logictests. cc @Xuanwo @everpcpc @PsiACE 
2. For test files, the changes are as follows
    -  Remove the semicolon after the query.
    -  Remove the blank line between the query and `----`.
    -  Replace the comment symbol from `--` to `##`.
    -  For those results where the output is empty, such as `select '', 1`; the first cell, ` ` will be displayed instead of `(empty)`, which can reduce confusion.
    -  Temporarily commenting out queries that contain dynamic results in their output, such as `db1 t1 FUSE $ANYTHING $DATE NULL 0 0 0 0 0`, we need a better way to handle them.
3. For new sqllogictest framework, it's under `sqllogictests` dir. 
   - It's based on [sqllogictest-rs](https://github.com/risinglightdb/sqllogictest-rs) crate.
   - It supports three clients. ping @youngsofun for review
   - Walk dir part needs to review carefully


FYI, the future todos: https://github.com/datafuselabs/databend/issues/9174

BohuTANG (approved):

BohuTANG:
Let's merge first, since it with a big diff.

tests/sqllogictests/src/client/http_client.rs
```diff
@@ -0,0 +1,120 @@
...
+        let url = "http://127.0.0.1:8000/v1/query".to_string();
+        let mut query = HashMap::new();
+        query.insert("sql", serde_json::to_value(sql).unwrap());
+        if let Some(session) = &self.session {
+            query.insert("session", serde_json::to_value(session).unwrap());
+        }
+        let mut response = self.response(sql, &url, true).await?;
+        // Set session from response to client
+        // Then client will same session for different queries.
+        self.session = serde_json::from_value(
```
Xuanwo, line 58: How about read the whole content via `response.bytes()` and deserialize as json instead of calling ` response.as_object()` multiple times?

Or maybe just `response.json()`?

tests/sqllogictests/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+    WalkDir(WalkDirError),
+    // Error from serde json
+    #[error("Serde json error: {0}")]
+    SerdeJson(SerdeJsonError),
+    // Error from databend sqllogictests
+    #[error("Databend sqllogictests error: {0}")]
+    SelfError(String),
+}
+
+impl From<TestError> for DSqlLogicTestError {
```
Xuanwo, line 50: We can use `Disconnect(#[from] io::Error)` to avoid those code.

tests/sqllogictests/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+        if let Some(mysql_client) = &mut self.mysql_client {
+            println!("Running sql with mysql client: [{}]", sql);
+            return mysql_client.query(sql);
+        }
+        if let Some(http_client) = &mut self.http_client {
+            println!("Running sql with http client: [{}]", sql);
+            return http_client.query(sql).await;
+        }
+        println!("Running sql with clickhouse client: [{}]", sql);
+        self.ck_client.as_mut().unwrap().query(sql).await
```
Xuanwo, line 72: Why we don't need to check this `ck_client`?

Xuanwo: If we must chose on from those three engines, I prefer to use an `enum` here for better reading.

tests/sqllogictests/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+
+#[tokio::main]
+pub async fn main() -> Result<()> {
+    env_logger::init();
+    let args = SqlLogicTestArgs::parse();
+    if let Some(handlers) = &args.handlers {
+        for handler in handlers.iter() {
+            match handler.as_str() {
+                "mysql" => {
+                    println!("Mysql client starts to run...");
```
Xuanwo, line 94: Use log instead of `println` so users have more control on this.

tests/sqllogictests/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+        "http"
+    }
+}
+
+#[tokio::main]
+pub async fn main() -> Result<()> {
+    env_logger::init();
+    let args = SqlLogicTestArgs::parse();
+    if let Some(handlers) = &args.handlers {
+        for handler in handlers.iter() {
```
Xuanwo, line 91: Maybe it's simpler to use `mysql,http,clickhouse` as the default value of handlers?

Xuanwo:
cc @xudong963, here some review you may be interested.

xudong963:
> cc @xudong963, here are some review you may be interested.

Oh, I forgot it, will take a look later! Thanks @Xuanwo

PR #9153, opened 2022-12-08
feat: implement show tables (from|in catalog.database)

ClSlaid:

## Summary

This PR implements show tables (from | in catalog.database)

### changes

- A new column named `catalog` is add into `system.tables`


src/query/service/tests/it/storages/system.rs
```diff
@@ -341,38 +341,38 @@ async fn test_tables_table() -> Result<()> {
...
-        r"\| system             \| tables_with_history \| SystemTables       \|            \| \d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{3} [\+-]\d{4} \| NULL     \| NULL      \| NULL                 \| NULL       \|",
-        r"\| system             \| tracing             \| SystemTracing      \|            \| \d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{3} [\+-]\d{4} \| NULL     \| NULL      \| NULL                 \| NULL       \|",
-        r"\| system             \| users               \| SystemUsers        \|            \| \d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{3} [\+-]\d{4} \| NULL     \| NULL      \| NULL                 \| NULL       \|",
-        r"\+--------------------\+---------------------\+--------------------\+------------\+-------------------------------\+----------\+-----------\+----------------------\+------------\+",
+        r"\+---------\+--------------------\+---------------------\+--------------------\+------------\+-------------------------------\+----------\+-----------\+----------------------\+------------\+",
```
Xuanwo, line 344: Maybe we can move this test to glodenfiles?

Xuanwo:
Mostly LGTM!

BohuTANG (approved):

BohuTANG:
@mergify update

Xuanwo (approved):

src/query/service/tests/it/storages/system.rs
ClSlaid: Once move it to the golden files, since currently system table tests' golden files does not accept regex expressions, some strongly time related column like `CREATED_ON` will ruin the whole test.

I'll revert it.

This part could be improved by:
## reading strings in blocks as RegExps:
```
| system | tracing | [0-9]+ | \d{4}-\d{2}-\d{2} \d{1,2}-\d{2}-\d{2} (\+|\-)\d{4} |
```
read as 4 RegExps:
```
`system`  (case sensitive)
`tracing` (case sensitive)
[0-9]+    (1 or more digits)
\d{4}-\d{2}-\d{2} \d{1,2}-\d{2}-\d{2} (\+|\-)\d{4} (date time)
```
Only '|'s at the start or end of line, or with spaces surrounded, will be seen as block splits.
```c
"| ",
" | ",
"|\0",
```

If each blocks in the row matches those RegExps fully, the row passes the test.

## regenerate golden file with column types:
```rust
// column_tys: Vec<Option<ColumnType>>
// type of each column
for (col, ty) in columns.iter().zip(column_tys.iter()) {
    if let Some(ty) = ty {
        // write with type assuring regex
        let reg = gen_reg(ty);
        write!(file, "{}", reg).unwrap();
    } else {
        // exactly as is
        // for example, "|p|p|p" should only matches "|p|p|p"
        // so RegExp escaping is needed.
        let col = reg_escape(col);
        write!(file, "{}", col).unwrap();
    }
}
```

PR #9159, opened 2022-12-08
feat(storage): use distinct count calculated by the xor hash function

b41sh:

## Summary

Xor filter calculates the hash value for each value in the calculation process. Using this hash value, we can get an approximate number of unique values, which can be used to replace the distinct count function `calc_column_distinct_of_values`. This optimization can improve the performance of data insertion.

Closes #issue


BohuTANG:
Is there some performance test after this PR?

lichuang:
test may need add some case that `gen_columns_statistics` second param is not None.

b41sh:
> Is there some performance test after this PR?

The time to insert 100m values was reduced by about 4s, and the time to build `BlockStatistics` was reduced to less than 1s.

image

src/query/storages/index/src/bloom.rs
```diff
@@ -134,9 +140,11 @@ impl BlockFilter {
                 }
 
                 let filter = filter_builder.build()?;
+                if let Some(len) = filter.len() {
```
dantengsky, line 143: does `filter.len()` return the NDV or just the number of values?

b41sh: the len is the number of distinct values.

dantengsky: :ok_hand:

dantengsky (approved):
LGTM

lichuang (approved):

PR #9160, opened 2022-12-09
fix(format): support ASCII control code hex as format field delimiter

### LINKED ISSUE #9155: bug: field_delimiter not support hex char
**Summary**

```
set format_field_delimiter='\x01';
```

But the check treats it as a string and fails:
https://github.com/datafuselabs/databend/blob/00fce4199834b578ff773cfb1306153fdd311b17/src/query/formats/src/format_option_checker.rs#L222

BohuTANG:

## Summary

support hex as a format field delimiter, such as ASCII control code: `\x01`, this is the hive default delimiter.
**Notes**:
`\x01` is a hex char([SOH](https://en.wikipedia.org/wiki/C0_and_C1_control_codes#SOH)) unable to display not a string `"\x01"` in your CSV.
image

 
Closes #9155


BohuTANG:
cc @wubx @TCeason

src/query/formats/src/format_option_checker.rs
```diff
@@ -138,14 +138,14 @@ impl FormatOptionChecker for TSVFormatOptionChecker {
...
 
-    fn check_quote(&self, quote: &mut String) -> Result<()> {
-        check_quote(quote, "\'")
-    }
-
     fn check_escape(&self, escape: &mut String) -> Result<()> {
         check_escape(escape, "\\")
     }
 
+    fn check_quote(&self, quote: &mut String) -> Result<()> {
```
BohuTANG, line 145: This change makes the order the same as the trait defined.

youngsofun (approved):

PR #9181, opened 2022-12-09
feat(query): Support timezone config in config file and sys env

### LINKED ISSUE #9173: Feature: Support timezone config in config file
**Summary**
Databend query need specify a timezone startup
support env and databend-query.toml support timezone

TCeason:

## Summary

Support timezone config in config file and sys env

Now in toml

```
[settings]
timezone="UTC"
```

in sys env

```
SETTINGS_TIMEZONE='Asia/Shanghai' RUST_LOG=info ./target/debug/databend-query -c databend-query.toml
```

## Priority level

global > sys env > toml.

And the value will be set in function default_settings. override default_value and user_setting.

If query set global timezone='xx' and restart with `SETTINGS_TIMEZONE='yy' RUST_LOG=info ./target/debug/databend-query -c databend-query.toml`

The value of timezone is 'xx', the level is GLOBAL.

Until execute sql `unset timezone`, the value of timezone will be set to 'yy', and level is SESSION.

So there should be no compatibility issues

Closes #9173

### DIFF
--- a/src/query/config/src/inner.rs
+++ b/src/query/config/src/inner.rs
@@ -43,6 +43,9 @@ pub struct Config {
     // Query engine config.
     pub query: QueryConfig,
 
+    // Settings config.
+    pub settings: SettingsConfig,
+
     pub log: LogConfig,
 
     // Meta Service config.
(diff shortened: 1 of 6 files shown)


src/query/config/src/inner.rs
```diff
@@ -43,6 +43,9 @@ pub struct Config {
     // Query engine config.
     pub query: QueryConfig,
 
+    // Settings config.
+    pub settings: SettingsConfig,
```
BohuTANG, line 47: No need new config section. Putting the `timezone` in the `QueryConfig` section is ok.

BohuTANG: Ok, I see the discussion(https://github.com/datafuselabs/databend/issues/9173#issuecomment-1344018671), that's make sense.

sundy-li:
Not only timezone, can we have other settings in this pr?

sundy-li:
`[settings]` is a JSON struct.

TCeason:
Maybe we need stick to use set global. More info in  #9173

PR #9185, opened 2022-12-09
fix(query): fix create table like view empty schema

sundy-li:

## Summary

Summary about this PR

Create table like view will have an empty schema, this pr fix that.

Closes #issue


TCeason (approved):

BohuTANG:
@mergify update

PR #9188, opened 2022-12-10
feat: custom csv record delimiter

BohuTANG:

## Summary

* CSV input with custom record delimiter, but can not have more than two chars

Closes #issue


BohuTANG:
@mergify update

PR #9190, opened 2022-12-11
refactor: move string func from base/io to base/string_func

BohuTANG:

## Summary

* Move string functions from `src/common/io/utils.rs` to `src/common/base/string.rs`
* Rename `src/common/io/utils.rs` to `src/common/io/src/serialization.rs`

Closes #issue


BohuTANG:
Only refactor, let's merge!

PR #9197, opened 2022-12-11
feat(query): enable runtime feature detection

sundy-li:

## Summary

Summary about this PR

enable runtime feature detection for `target_features`.

performance :

insert 10000_0000  random numbers  into hashmap 

```
// pure run without target_features
cargo test --package common-hashtable --lib --release -- b::test_hashmap --exact --nocapture
// insert elasped 5072 ms
// find elasped 1955 ms, total: 4999999950000000


// run with sse4.2 features
RUSTFLAGS="-C target-cpu=x86-64 -C target-feature=+sse4.2" cargo test --package common-hashtable --lib --release -- b::test_hashmap --exact --nocapture
// insert elasped 3006 ms
// find elasped 1288 ms, total: 4999999950000000


// run runtime detection (with this pr)
cargo test --package common-hashtable --lib --release -- b::test_hashmap --exact --nocapture
// insert elasped 3788 ms
// find elasped 2037 ms, total: 4999999950000000
```

Runtime dispatch will share an atomic flag, so it's not overhead. And we can also enjoy the compile-time dispatch using `target-feature` flags.

Closes #issue

### DIFF
--- a/src/common/hashtable/src/traits.rs
+++ b/src/common/hashtable/src/traits.rs
@@ -16,6 +16,7 @@ use std::hash::BuildHasher;
 use std::mem::MaybeUninit;
 use std::num::NonZeroU64;
 
+use multiversion::multiversion;
 use ordered_float::OrderedFloat;
 use primitive_types::U256;
 use primitive_types::U512;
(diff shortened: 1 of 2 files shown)


sundy-li:
Closing because it may have different hash results in heterogeneous arch deployment.

Let's still use none-simd version by default.

PR #9199, opened 2022-12-12
feat: push down filter to parquet reader.

### LINKED ISSUE #9135: Feature: improve table function `read_parquet`.
**Summary**

Table function `read_parquet` is introduced by #9080. We need more improvement on it.

- [x] Don't read metadata from files when `read_partitions`, but do this before reading data. #9154
- [x] Support reading parquet files with multiple row groups. #9154
- [ ] Support pushing down filter to parquet reading. #9199

RinChanNOWWW:

## Summary

- Push down filter to parquet reader.
- Convert `ArrowSchema` to `DataSchema` only when convert `ArrowChunk` to `DataBlock`.

Closes #9135

### DIFF
--- a/src/query/storages/parquet/src/parquet_reader/mod.rs
+++ b/src/query/storages/parquet/src/parquet_reader/mod.rs
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 mod deserialize;
+mod filter;
 mod meta;
 mod read;
+
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::sync::Arc;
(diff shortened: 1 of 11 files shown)


PsiACE (approved):

src/query/storages/parquet/src/parquet_reader/mod.rs
```diff
@@ -38,16 +43,24 @@ pub struct ParquetReader {
...
+    /// select a, b, a from t;
     /// ```
     columns_to_read: HashSet<usize>,
     /// The schema of the [`common_datablocks::DataBlock`] this reader produces.
-    projected_schema: DataSchemaRef,
+    output_schema: DataSchemaRef,
+    /// The actual schema used to read parquet. It will be converted to [`common_datavalues::DataSchema`] when output [`common_datablocks::DataBlock`].
+    ///
+    /// The reason of using [`ArrowSchema`] to read parquet is that
+    /// There are some types that Databend not support such as Timestmap of nanoseconds.
```
sundy-li, line 61: good comments.

sundy-li (approved):

Issue #9162, opened 2022-12-09
bug: downcast column error, column type: "Const(String)", expected column: "common_datavalues::columns::string::StringColumn".


hanyisong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

8.0.26-v0.8.114-nightly

### What's Wrong?

downcast column error, column type: "Const(String)", expected column: "common_datavalues::columns::string::StringColumn".

### How to Reproduce?

```sql
DROP DATABASE IF EXISTS databend6;
CREATE DATABASE databend6;
USE databend6;
CREATE TABLE t0(c0VARCHAR VARCHAR NOT NULL DEFAULT('XVk'));
CREATE TABLE t1(c0VARCHAR VARCHAR NULL);
CREATE TABLE t2(c0INT SMALLINT NULL);
CREATE TABLE t3(c0VARCHAR VARCHAR NOT NULL DEFAULT('at'));
INSERT INTO t1(c0varchar) VALUES ('gxZu6');
DELETE FROM t0 WHERE ((NULL)<=(NULL));
EXPLAIN SELECT DISTINCT ((NULL)<=(NULL)) FROM t1, t2, t0, t3 HAVING ((((((t2.c0int)%(t2.c0int)))-(((t2.c0int)%(1617114512))))) IS NOT NULL) LIMIT 1964819055;
INSERT INTO t2(c0int) VALUES (1617114512);
INSERT INTO t0(c0varchar) VALUES ('AUJC5e');
INSERT INTO t3(c0varchar) VALUES ('kx');
INSERT INTO t2(c0int) VALUES (1617114512);
CREATE VIEW v0 AS SELECT (t3.c0varchar NOT IN (t0.c0varchar, '', '')), t0.c0varchar, t2.c0int FROM t3, t1, t2, t0 WHERE ((1.617114496E9) IS NULL) GROUP BY '' HAVING (('kDIT3') IS NOT NULL) LIMIT 194625690 OFFSET 165596766;
INSERT INTO t1(c0varchar) VALUES ('E5'), ('');
INSERT INTO t2(c0int) VALUES (-90628224), (1357686063);
INSERT INTO t1(c0varchar) VALUES ('');
DROP VIEW v0;

[...]

PR #9217, opened 2022-12-12
fix: fix ConstColumn(String) cast to StringColumn bug

lichuang:

## Summary

fix: downcast column error, column type: "Const(String)", expected column: "common_datavalues::columns::string::StringColumn"

But the logic test will panic in cluster mode, see: https://github.com/datafuselabs/databend/issues/9222
So split the logic test out as a bug issue.

Closes https://github.com/datafuselabs/databend/issues/9162


src/query/datavalues/tests/it/columns/const_.rs
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common_datavalues::prelude::*;
+use common_exception::Result;
+
+#[test]
+fn test_cast_const_string_to_stringcolumn() -> Result<()> {
```
sundy-li, line 19: should be removed.

sundy-li (approved):

PR #9228, opened 2022-12-13
feat(parquet): prune row groups before reading.

RinChanNOWWW:

## Summary

Prune row groups before reading.

If there are column statistics in the parquet file, we can do row group pruning by these statistics.
Otherwise, we read the whole row groups.


src/query/storages/parquet/src/parquet_reader/meta.rs
```diff
@@ -43,4 +52,80 @@ impl ParquetReader {
...
+                ));
+            }
+
+            let field = &schema.fields[*index];
+            let column_stats = pread::statistics::deserialize(field, rgs)?;
+            stats_of_row_groups.insert(*index, BatchStatistics::from(column_stats));
+        }
+
+        for (rg_idx, _) in rgs.iter().enumerate() {
+            let mut cols_stats = HashMap::new();
```
Xuanwo, line 83: How about `HashMap::with_capacity(stats.capacity())`

src/query/storages/pruner/src/range_pruner.rs
```diff
@@ -55,6 +55,9 @@ impl RangePruner for RangeFilter {
     }
 }
 
+/// Create a new [`RangePruner`] from expression and schema.
```
Xuanwo, line 58: I prefer to have a `RangePruner::new()`

src/query/storages/fuse/fuse/src/pruning/pruning_executor.rs
```diff
@@ -79,7 +79,7 @@ impl BlockPruner {
             .and_then(|p| p.limit);
 
         // prepare the limiter. in case that limit is none, an unlimited limiter will be returned
-        let limiter = limiter::new_limiter(limit);
+        let limiter = limiter_pruner::new_limiter(limit);
```
Xuanwo, line 82: I prefer to have a `LimiterPruner::new()` API which carries more context about this struct. `new_limiter` could be anything if imported directly.

src/query/storages/pruner/src/range_pruner.rs
RinChanNOWWW: But `RangePruner` is a trait.

src/query/storages/fuse/fuse/src/pruning/pruning_executor.rs
RinChanNOWWW: `LimiterPruner` is also a trait.

Xuanwo (approved):

dantengsky:
looks good to me. It will be better if there were unit or stateless tests

PR #8517, opened 2022-10-28
refactor(query): INFORMATION_SCHEMA to lowwer case

TCeason:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

cast INFORMATION_SCHEMA to lowwer case

Fixes #8516

PR #9233, opened 2022-12-13
chore: clean error and useless tests

xudong963:

## Summary

When I work on https://github.com/datafuselabs/databend/pull/9150, I found some strange and useless tests... 😅 

I have some suggestions for tests:

1. For reviewers, we should review tests carefully too.
2. For tests writers, we should focus on test points to avoid useless tests.
3. Consider if tests that you write will be influenced by other files and even become flaky tests.

Closes #issue


tests/logictest/suites/base/05_ddl/05_0003_ddl_alter_database
```diff
@@ -38,75 +38,43 @@ statement ok
 ALTER DATABASE IF EXISTS c RENAME TO a;
 
 statement query T
-SELECT name FROM system.databases;
```
xudong963, line 41: Can't guarantee previous tests file don't create databases and drop it.

So the test will be flaky..., from the context, I guess we only should test if database `a` and `b` are created.

So I changed it to `SELECT * FROM system.databases where name = 'a' or name = 'b';` which is more focused.

tests/logictest/suites/base/05_ddl/05_0003_ddl_alter_database
```diff
@@ -38,75 +38,43 @@ statement ok
...
-statement query T
-SELECT name FROM system.databases;
-
-----
-system
-information_schema
-a
-b
-default
+-- don't exist database c, so the result will be empty
```
xudong963, line 53: Ditto

tests/logictest/suites/base/05_ddl/05_0003_ddl_alter_database
```diff
@@ -38,75 +38,43 @@ statement ok
...
+statement query ok
+SELECT * FROM system.databases where name = 'c';
 
 statement error 1002
 ALTER DATABASE IF EXISTS system RENAME TO C;
 
 statement error 1002
 ALTER DATABASE system RENAME TO C;
 
-statement query T
```
xudong963, line 72: I think the test is useless

tests/logictest/suites/base/05_ddl/05_0024_ddl_set_role
```diff
@@ -1,12 +1,6 @@
 statement ok
 CREATE ROLE IF NOT EXISTS 'test-s'
 
-statement query T
```
xudong963, line 4: The test is useless and flaky...

We can't guarantee the current role is `account_admin`.

tests/logictest/suites/base/06_show/06_0003_show_settings
```diff
@@ -1,10 +1,3 @@
-statement query B
```
xudong963, line 1: Can we guarantee the value of `max_threads` or `max_memory_usage` didn't be changed by previous files?

tests/logictest/suites/base/06_show/06_0007_show_roles
```diff
@@ -1,13 +1,13 @@
 statement ok
 CREATE ROLE IF NOT EXISTS 'test';
 
-statement query TIII
-SHOW ROLES;
```
xudong963, line 5: Can we guarantee the current role is `account_admin`?

What we need to test is if the role test is created.

tests/logictest/suites/base/09_fuse_engine/09_0001_remote_insert
```diff
@@ -46,9 +46,6 @@ SELECT * FROM t2;
 1 2021-09-07 21:38:35.000000 2021-09-07
 0 2021-09-07 21:38:35.000000 2021-09-07
 
- 2021-09-07 21:38:35.000000 2021-09-07
```
xudong963, line 49: I don't understand why the test was passed 😅

tests/logictest/suites/base/09_fuse_engine/09_0001_remote_insert_v2
```diff
@@ -47,9 +47,6 @@ SELECT * FROM t2;
 1 2021-09-07 21:38:35.000000 2021-09-07
 0 2021-09-07 21:38:35.000000 2021-09-07
 
- 2021-09-07 21:38:35.000000 2021-09-07
```
xudong963, line 50: ditto

tests/logictest/suites/base/09_fuse_engine/09_0019_issue_8831
```diff
@@ -46,8 +46,10 @@ select count(1) from t1;
 ----
 7
 
-
```
xudong963, line 49: ditto

tests/logictest/suites/base/05_ddl/05_0003_ddl_alter_database
TCeason: Suggest to add where name ='information_schema'. I have change in this issue https://github.com/datafuselabs/databend/pull/8517

tests/logictest/suites/base/05_ddl/05_0003_ddl_alter_database
TCeason: Agree

tests/logictest/suites/base/05_ddl/05_0024_ddl_set_role
TCeason: cc @flaneur2020 Need help, how to test current_role()?

tests/logictest/suites/base/06_show/06_0003_show_settings
TCeason: How about execute `unset max_threads; unset max_memory_usage` before execute this query?

tests/logictest/suites/base/06_show/06_0007_show_roles
TCeason: cc @flaneur2020  Need help .

tests/logictest/suites/base/09_fuse_engine/09_0020_analyze
```diff
@@ -27,11 +27,9 @@ select * from t order by a;
 6
 7
 
-statement query T
+statement ok
```
TCeason, line 30: Maybe this sql need result is empty, not only ok.

tests/logictest/suites/base/20+_others/20_0008_where_optimization
```diff
@@ -32,12 +32,9 @@ select * from t_where_optimizer where a = b order by a;
 
 -- All rows are filtered out
 
-statement query II
+statement ok
```
TCeason, line 35: Maybe this sql need result is empty, not only ok?

tests/logictest/suites/base/05_ddl/05_0003_ddl_alter_database
xudong963: The file is aim to test `alter` database. The test is to test the databases(a / b) that were created above. So I think we don't need to add `where name ='information_schema'`.

tests/logictest/suites/base/05_ddl/05_0024_ddl_set_role
xudong963: 15~16 rows test `current_role()`.  The test is useless

tests/logictest/suites/base/06_show/06_0003_show_settings
xudong963: I think the test is useless

TCeason: It's usefule. Because the max_threads and max_memory_usage it's determined automatically. So we can test it in [src/query/service/tests/it/storages/testdata/system-tables.txt](https://github.com/datafuselabs/databend/pull/8883/files#diff-86bc27fced7f8932a180b8f562dc48d053d34ce76ef5511ce54132da842c7d1a), so extract it into logictest.

sundy-li: settings belong to a session, now we start a new session per file.

If we set global, the setting value will be global, we must remember to unset the settings before end the file.

tests/logictest/suites/base/05_ddl/05_0003_ddl_alter_database
TCeason: Ok. I see same test in tests/logictest/suites/base/01_system/01_0003_information_schema

tests/logictest/suites/base/05_ddl/05_0024_ddl_set_role
TCeason: I think maybe want to test `SELECT current_role()!='test-s';`? I'm not sure.

tests/logictest/suites/base/09_fuse_engine/09_0020_analyze
xudong963: In our new logictest framework, I'll use `query ok` or `query types` without `----` and results. Both ways are ok to me.

tests/logictest/suites/base/06_show/06_0003_show_settings
TCeason: > settings belong to a session, now we start a new session per file.
> 
> If we set global, the setting value will be global, we must remember to unset the settings before end the file.

Agree. 

I think we need to set a rule for test (for now only reviewers and developers should take note).

> 1. All states used in this use case should be assumed to exist and drop them until we start running the actual test content. 
> 2. Clear all states before this test exits.

e.g.

```sql
-- must drop if exists
drop database if exists a;

create database a;
create table a.b(id int);
show tables from a;

-- must clear
drop database a;

` ` `

File: tests/logictest/suites/base/09_fuse_engine/09_0001_remote_insert_v2 (hunk 8 above)
Comment (BohuTANG) [reply tests/logictest/suites/base/09_fuse_engine/09_0001_remote_insert_v2:50]: This will use the default value? ping @dantengsky

**Comment by xudong963** [comment]
Strange..

image

Any thoughts?

**Comment by TCeason** [comment]
> Strange..
> 
> image
> 
> Any thoughts?

Not *. 


statement query T
SELECT name FROM system.databases where name = 'a' or name = 'b';

----
a
b

```sql
'u1'@mysqldb 22:41:26 [default]> SELECT * FROM system.databases where name = 'a' or name = 'b';
+---------+------+
| catalog | name |
+---------+------+
| default | a    |
| default | b    |
+---------+------+
2 rows in set (0.08 sec)
Read 9 rows, 286.00 B in 0.020 sec., 442.51 rows/sec., 13.73 KiB/sec.

'u1'@mysqldb 22:41:32 [default]> SELECT name FROM system.databases where name = 'a' or name = 'b';
+------+
| name |
+------+
| a    |
| b    |
+------+
2 rows in set (0.06 sec)
Read 9 rows, 286.00 B in 0.016 sec., 545.55 rows/sec., 16.93 KiB/sec.

```

tests/logictest/suites/base/05_ddl/05_0024_ddl_set_role
flaneur2020: oh, 'account_admin' is the default role of the user which executing sqlogictest in the before. IMHO this test case can be skipped because this checks a default value but not the behaviour of the `set role` statement.

tests/logictest/suites/base/06_show/06_0007_show_roles
flaneur2020: what's the user sqlogictest uses?

if it's 'root', it should guarantees the `account_admin` role

tests/logictest/suites/base/09_fuse_engine/09_0001_remote_insert_v2
dantengsky: those are more like typos to me.  seems the commit forget to remove the deprecated “expected resultset” after tuning the test case.

do not know how will the “logictest driver” interpret them..

tests/logictest/suites/base/06_show/06_0007_show_roles
xudong963: > what's the user sqlogictest uses?

Maybe the user can't be guaranteed? So we can't guarantees the `account_admin` role?

BohuTANG: Let's add more role tests later. If we are not sure, suggest skipping this case first.

BohuTANG (approved):

PR #9249, opened 2022-12-14
feat(query): support attaching stage for insert values

### LINKED ISSUE #8889: Feature: Support pass a staged file path in the POST /v1/query args
**Summary**

On ingesting huge amount of data, staged file is the first choice, it have the lowest price.

We can consider adding a parameter like `valuesStagePath` which allows take a stage file's path as the payload of the data.

like:

```
{
  "sql": "INSERT INTO books VALUES",
  "valuesStagePath": "~/books.csv"
}
```

The insert SQL can be same as the clickhouse handler or the `insert_sql` in the streaming load api, the code may be reused among these handlers. The only exception is it takes a s3 staged path instead of reading in http streams.

It can auto purge the staged file by default.

everpcpc:

## Summary

Then we can obtain data from stage with the insert operation.

Closes #8889


Xuanwo:
Hi, can you give some explaination in this PR's summary? I don't what does `sideload` means.

src/query/service/src/sessions/query_ctx.rs
```diff
@@ -354,6 +359,11 @@ impl TableContext for QueryContext {
     fn get_processes_info(&self) -> Vec<ProcessInfo> {
         SessionManager::instance().processes_info()
     }
+
+    // Get Sideload Options.
+    fn get_sideload(&self) -> Option<SideloadOptions> {
+        self.shared.get_sideload()
+    }
```
flaneur2020, line 366: can this method only placed in `impl QueryContext` instead of `impl TableContext for QueryContext`?

IMHO, the sideload options makes sense in QueryContext (it takes effect in a query's lifetime), but it may hard to understand as a property in the table context.

everpcpc: The `Binder` only have a `TableContext` when binding statement, I think the name `TableContext` is somehow missleading.

everpcpc: It seems @Xuanwo changed this previously https://github.com/datafuselabs/databend/pull/7600.
Can you help provide some background for this?

Xuanwo: I added this to address circle dependences. `Planner` can't relay on `QueryContext` directly.

flaneur2020: How about name it as BinderContext ?

src/query/catalog/src/table_context.rs
```diff
@@ -54,6 +54,12 @@ pub struct ProcessInfo {
     pub created_time: SystemTime,
 }
 
+#[derive(Debug, Clone)]
+pub struct SideloadOptions {
```
flaneur2020, line 58: can `uri` and `stage` be unified as a uri field? if it's started as `@`, it's a staged file's path

src/query/sql/src/planner/plans/insert.rs
```diff
@@ -31,6 +32,8 @@ pub enum InsertInputSource {
     StreamingWithFileFormat(FileFormatOptions, usize, Option<Arc<InputContext>>),
     // From cloned String and format
     Values(String),
+    // From sideload stage or uri
+    Sideload(Arc<SideloadOptions>),
```
flaneur2020, line 36: How about calls this new source as `StagedValues`? We can get about the data comes from stage by its name

src/query/service/src/sessions/query_ctx.rs
everpcpc: Maybe we can rename it in next PR.

src/query/service/src/interpreters/interpreter_insert_v2.rs
```diff
@@ -101,6 +117,177 @@ impl InsertInterpreterV2 {
...
         Ok(cast_needed)
     }
+
+    fn apply_stage_options(
+        &self,
+        stage: &mut UserStageInfo,
+        params: &BTreeMap<String, String>,
+    ) -> Result<()> {
+        for (k, v) in params.iter() {
+            match k.as_str() {
```
youngsofun, line 127: may we can reuse the parser combinator to do it.
which make it easier to maintain (in one place).

search for `tokenize_sql` in code base, you can see several  code snippets like this.


```
        let tokens = tokenize_sql(sql)?;
        let backtrace = Backtrace::new();
        let exprs = some_parser(&tokens[1..tokens.len()], sql_dialect, &backtrace)?;
```

src/query/service/src/servers/http/v1/query/http_query.rs
```diff
@@ -141,6 +143,12 @@ impl HttpSessionConf {
     }
 }
 
+#[derive(Deserialize, Debug, Clone)]
+pub struct StageAttachmentConf {
+    pub(crate) location: String,
+    pub(crate) params: Option<BTreeMap<String, String>>,
```
youngsofun, line 149: is this `params` only for format_options?
is one location enough?
which form of location do we intend to use? name? internal/external/uri?

src/query/service/src/interpreters/interpreter_insert_v2.rs
```diff
@@ -101,6 +117,177 @@ impl InsertInterpreterV2 {
...
+            let ctx = ctx.clone();
+            let table = table.clone();
+
+            match may_error {
+                Some(error) => {
+                    tracing::error!("insert stage file error: {}", error);
+                    Err(may_error.as_ref().unwrap().clone())
+                }
+                None => {
+                    let append_entries = ctx.consume_precommit_blocks();
```
youngsofun, line 269: we need to clear the tmp stage file.  
do it in the client? or need some GC?

src/query/service/src/interpreters/interpreter_insert_v2.rs
```diff
@@ -101,6 +117,177 @@ impl InsertInterpreterV2 {
...
+            pattern: "".to_string(),
+            files_to_copy: None,
+        };
+
+        let all_source_file_infos = StageTable::list_files(&table_ctx, &stage_table_info).await?;
+
+        // TODO:(everpcpc) color_copied_files
+
+        let mut need_copied_file_infos = vec![];
+        for file in &all_source_file_infos {
```
youngsofun, line 216: I think this is used only for copy, which commit the paths to meta  with `set_on_finished`.
for `insert with stage_attachment.`, I think we do not need it.
the path maybe a tmp uuid.  if upload ok and insert fail, we may retry this insert (dep on the clear error code)
finally the tmp file is cleaned

src/query/service/src/servers/http/v1/query/http_query.rs
everpcpc: > is this params only for format_options?
for both format options and copy options.

> is one location enough?
> which form of location do we intend to use? name? internal/external/uri?

yes, we use format like `@stage_name/path`

src/query/service/src/interpreters/interpreter_insert_v2.rs
everpcpc: The cleanup for tmp stage file would be in next PR, and it should be optional.

youngsofun: ok

src/query/service/src/servers/http/v1/query/http_query.rs
youngsofun: > yes, we use format like @stage_name/path
ok 

> for both format options and copy options.

I prefer to separate them and reuse code in parser

src/query/service/src/interpreters/interpreter_insert_v2.rs
```diff
@@ -101,6 +117,177 @@ impl InsertInterpreterV2 {
         let cast_needed = select_schema != *output_schema;
         Ok(cast_needed)
     }
+
+    fn apply_stage_options(
+        &self,
+        stage: &mut UserStageInfo,
+        params: &BTreeMap<String, String>,
+    ) -> Result<()> {
+        for (k, v) in params.iter() {
```
youngsofun, line 126: we can separate copy options and format options.

and reuse fn like 

```
pub fn parse_copy_file_format_options(
    file_format_options: &BTreeMap<String, String>,
) -> Result<FileFormatOptions>
```

src/query/service/src/servers/http/v1/query/http_query.rs
youngsofun: > yes, we use format like @stage_name/path

we can add a little comments here.

src/query/service/src/interpreters/interpreter_insert_v2.rs
everpcpc: Moved and separated this function to `src/meta/types/src/user_stage.rs`.
`parse_copy_file_format_options` maybe unified in next PR.

src/query/service/src/sessions/query_ctx.rs
everpcpc: https://github.com/datafuselabs/databend/issues/9263

src/query/catalog/src/table_context.rs
everpcpc: Changed to stage only.

src/query/sql/src/planner/plans/insert.rs
everpcpc: updated

ariesdevil:
We need to update docs for this new way of data insertion, maybe next PR.

ariesdevil (approved):

youngsofun (approved):

youngsofun:
need a stateful test, maybe next PR.

PR #9250, opened 2022-12-14
fix: fix unknown database in query without relation to this database

### LINKED ISSUE #9231: Bug: Unknown database in query without relation to this database
**Summary**
```sql
mysql> create database test1;

mysql> use test1;
Database changed
mysql> drop database test1;

mysql> create database test2;

mysql> create table test2.t(a int);

mysql> insert into test2.t values(1);
ERROR 1105 (HY000): Code: 1003, displayText = Unknown database 'test1'.
mysql>
```

It's strange that create table in database test but can't insert data ...  😅 

Ideally, we should avoid drop database in current database, so when `drop database test1;`, we should get error: ` cannot drop the currently open database`;

ariesdevil:

## Summary

Fix unknown database in query without relation to this database.

Thanks to @drmingdrmer and @lichuang for sharing the `share db` knowledge and advice.
Thanks to @Xuanwo for sharing the historical reasons of `TableInfo`.

Closes #9231


src/meta/app/src/schema/table.rs
```diff
@@ -156,6 +156,8 @@ pub struct TableInfo {
 
     pub tenant: String,
 
+    pub database: String,
```
Xuanwo, line 159: I remember there is an assumption that `database` could be changed.

Cc @drmingdrmer for confirmation.

ariesdevil:
@Xuanwo Putting comments here prevent comments from being collapsed when click Resolve conversation:)

I had a long discussion with @drmingdrmer and @lichuang  yesterday afternoon. The summary is:

1. Ideally, operations between `database` and `table` should be isolated and **yes** we do now almost everywhere except:

2. We have `SharedDatabase` and this is not a real db, changes to the table metadata under the shared db involve permissions, which are different from the normal db. So @lichuang put the change operation to a trait and normal,shared db impls this trait respectively.

3. As such modification, we do some modifications for table must through db and we must know which db we will be op on it.

4. Here is a example:

```sql
create database t1;

# after use t1, we set the current database to session-level and it's all right.
use t1;

create database t2;

create table t2.t(a int);

# But when we do this, the FuseTable don't have database field.
# So if we want a db-level operation we use `ctx.get_current_database()` which bug happens.
insert into t2.t values(1);
```

5. Another solution I think we can set a `InsertIntoTable` to query-ctx-level, so we no need add field to FuseTable and we can get it from TableContext.

cc @sundy-li

src/query/storages/fuse/fuse/src/fuse_table.rs
```diff
@@ -94,22 +95,24 @@ impl FuseTable {
...
-            Some(ref from_share) => create_share_table_operator(
+
+        let mut operator = match table_info.db_type.clone() {
+            DatabaseType::ShareDB(share_ident) => create_share_table_operator(
                 ShareTableConfig::share_endpoint_address(),
                 ShareTableConfig::share_endpoint_token(),
-                &from_share.tenant,
-                &from_share.share_name,
+                &share_ident.share_name,
+                &share_ident.share_name,
```
drmingdrmer, line 104: Both of them are `share_name`?

src/query/storages/fuse/fuse/src/operations/truncate.rs
```diff
@@ -72,11 +72,15 @@ impl FuseTable {
...
-                })
+                .update_table_meta(
+                    &tenant,
+                    self.table_info.db_type.clone(),
+                    UpdateTableMetaReq {
+                        table_id,
+                        seq: MatchSeq::Exact(table_version),
+                        new_table_meta,
+                    },
+                )
```
drmingdrmer, line 83: `TableInfo` contains both tenant and `db_type`, it looks like `update_table_meta` should accept a `TableInfo` argument. in next PR 🤔

src/query/storages/fuse/fuse/src/operations/revert.rs
```diff
@@ -54,8 +54,9 @@ impl FuseTable {
 
         // 4. let's roll
         let tenant = ctx.get_tenant();
-        let db_name = &navigation_descriptor.database_name;
-        let reply = catalog.update_table_meta(&tenant, db_name, req).await;
+        let reply = catalog
+            .update_table_meta(&tenant, self.table_info.db_type.clone(), req)
+            .await;
```
drmingdrmer, line 59: Maybe it'd be better to use the `tenant` in `TableInfo` instead of fetching it from ctx, similar to the way `db_type` does. Maybe next PR?

ariesdevil: No, this PR should do this.

drmingdrmer (approved):
💯

lichuang (approved):

sundy-li (approved):

Issue #9216, opened 2022-12-12
 CSV | TSV | NDJSON | PARQUET | XML limitations for COPY INTO


soyeric128:
As reported by @wubx, the documentation for COPY INTO needs some explanations about the limitations of each supported format, such as accepted delimiters. 

SF reference: https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#type-csv

@youngsofun Could you pls help provide some input on this issue?

Thanks.

PR #9255, opened 2022-12-15
docs(copy): update format_options.

youngsofun:

## Summary

1.  desc options format by format
2.  add notices for CSV/TSV
3.  add some options
4.  add 'Data loading only' for some options.
5.  add 'Supported values'  for some options.

Closes https://github.com/datafuselabs/databend/issues/9216

### DIFF
--- a/docs/doc/14-sql-commands/10-dml/dml-copy-into-table.md
+++ b/docs/doc/14-sql-commands/10-dml/dml-copy-into-table.md
@@ -160,34 +160,75 @@ formatTypeOptions ::=
   RECORD_DELIMITER = '<character>'
   FIELD_DELIMITER = '<character>'
   SKIP_HEADER = <integer>
+  QUOTE = '<character>'
+  ESCAPE = '<character>'
+  NAN_DISPLAY = '<string>'
+  ROW_TAG = '<string>'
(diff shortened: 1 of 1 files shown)


youngsofun:
maybe we should have  a page for format options https://clickhouse.com/docs/en/interfaces/formats/

1.  we are likely to have many more options later,
2.  we may need more explanations later to avoid miss use, like clickhouse and mysql https://dev.mysql.com/doc/refman/8.0/en/load-data.html.
3. `####` for each type and `#####` for each option is not easy to read
4.  3 clauses select from stage and copy into stage/table

BohuTANG (approved):

soyeric128:
> maybe we should have a separate page for format options https://clickhouse.com/docs/en/interfaces/formats/
> 
> 1. we are likely to have many more options later,
> 2. we may need more explanations later to avoid miss use, like clickhouse and mysql https://dev.mysql.com/doc/refman/8.0/en/load-data.html.
> 3. `####` for each type and `#####` for each option is not easy to read
> 4. there clauses use it：`select from stage` and `copy into stage/table`
> 
> where in the doc is the right place? @soyeric128 @BohuTANG

Please review: https://github.com/datafuselabs/databend/pull/9315

PR #9269, opened 2022-12-16
fix(base): try fix SIGABRT when catch unwind

zhang2014:

## Summary

fix(base): try fix SIGABRT when catch unwind

Closes #issue


drmingdrmer (commented):
💯

PR #9277, opened 2022-12-18
feat(query): add allow_insecure_tenant config and sandbox_tenant  

sundy-li:
…ings

## Summary

Summary about this PR

Add a new config `allow_insecure_tenant` and a new setting named `sandbox_tenant`

If `allow_insecure_tenant` is on, each session can use `set sandbox_tenant = 'xxx'` to create a new tenant env.

This helps:

1. Now logicaltest runs really slow because the suitcase is executed one by one. With this config, we could execute the suitcase in parallel.

2. The logicaltest may be failed during runtime, this leaves a dirty state which takes effect on other suits. Now we could make each suit run in a virtual sandbox and the catalog is isolated totally.

Closes #issue

### DIFF
--- a/src/query/config/src/outer_v0.rs
+++ b/src/query/config/src/outer_v0.rs
@@ -1314,6 +1314,9 @@ pub struct QueryConfig {
 
     #[clap(skip)]
     quota: Option<TenantQuota>,
+
+    #[clap(long)]
+    pub allow_insecure_tenant: bool,
 }
 
 impl Default for QueryConfig {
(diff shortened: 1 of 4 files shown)


BohuTANG:
@mergify update

src/query/config/src/outer_v0.rs
```diff
@@ -1314,6 +1314,9 @@ pub struct QueryConfig {
 
     #[clap(skip)]
     quota: Option<TenantQuota>,
+
+    #[clap(long)]
+    pub allow_insecure_tenant: bool,
```
Xuanwo, line 1319: How about `internal_enable_sandbox_tenant`? `internal_` to indicate that it's an internal only config.

BohuTANG (approved):

sundy-li:
The `test_default_config` seems redundant, it's testing for testing, so I think we shall remove it.

Xuanwo (approved):

PR #9279, opened 2022-12-18
feat(query): add native format in fuse table

sundy-li:

## Summary

Summary about this PR

1. bump arrow2 version
2. Introduce an experimental native storage_format in fuse

Closes #issue


sundy-li:
how to use:

```
 create table tmp (a int) ENGINE=FUSE STORAGE_FORMAT='native';
```

src/query/datavalues/src/columns/array/mod.rs
```diff
@@ -142,12 +143,15 @@ impl Column for ArrayColumn {
...
-            Box::new(LargeListArray::from_data(
-                arrow_type,
-                self.offsets.clone(),
-                array,
-                None,
-            ))
+            Box::new(
+                LargeListArray::try_new(
+                    arrow_type,
+                    unsafe { OffsetsBuffer::new_unchecked(self.offsets.clone()) },
```
doki23, line 149: We can use OffsetsBuffer::try_from

sundy-li: I do think we should check the offset which drop the performance, we ensure it's valid.

src/query/storages/fuse/fuse/src/operations/fuse_native_source.rs
```diff
@@ -0,0 +1,248 @@
...
+pub struct FuseNativeSource {
+    state: State,
+    ctx: Arc<dyn TableContext>,
+    scan_progress: Arc<Progress>,
+    output: Arc<OutputPort>,
+    output_reader: Arc<BlockReader>,
+
+    prewhere_reader: Arc<BlockReader>,
+    prewhere_filter: Arc<Option<EvalNode>>,
+    remain_reader: Arc<Option<BlockReader>>,
```
RinChanNOWWW, line 60: If we don't pre-filter prewhere chunks, one reader is enough.

sundy-li: Yes, I want to add `page` skip later.

sundy-li:
Average 1.85x in hits dataset

image

BohuTANG:
Based on s3:
image

sundy-li:
> Q7 & Q8 native is faster than parquet, others are not very obvious, and a number of them are even slower than parquet.

Currently, we don't have prewhere optimization like parquet source.  This is an initial implementation that take no bad effect on  default parquet-based fuse engine,  we can have it later.

BohuTANG:
@sundy-li 

Some tests are deleted(like [09_fuse_engine/09_0001_remote_insert_v2](https://github.com/datafuselabs/databend/pull/9279/files#diff-aff2e89d0f348f9ed1363bf41b1821e8fd33460a82fac815e05f958cabb668e0)), this is for?

sundy-li:
> @sundy-li
> 
> Some tests are deleted(like [09_fuse_engine/09_0001_remote_insert_v2](https://github.com/datafuselabs/databend/pull/9279/files#diff-aff2e89d0f348f9ed1363bf41b1821e8fd33460a82fac815e05f958cabb668e0)), this is for?

It's duplicated.

BohuTANG (approved):
🚀

PR #9285, opened 2022-12-19
feat(http handler): enable result timeout by default.

youngsofun:

## Summary

1. rename http_handler_result_timeout_millis to http_handler_result_timeout_secs
2. change default to 60 secs.
3. enable it (it was enabled only when wait_time_secs=0)


BohuTANG:
@mergify update

BohuTANG (approved):

PR #9288, opened 2022-12-19
feat(query): estimate insert value rows

b41sh:

## Summary

Currently, when parse insert values, we use a fixed value(1024) as the capacity of `MutableColumn`. If the number of rows to insert is relatively small, a lot of allocated memory space will not be used. On the contrary, if the number of rows to insert is relatively large, the allocated memory space is not enough, which will lead to frequent resizing and affect performance.

Since we have used the `AhoCorasick` algorithm to find `'` and `\`, we can also find `(` at the same time, and use the number of `(` to estimate the number of rows to insert, so that we can allocate the required memory for insert values and avoid unnecessary resize.

Closes #issue

### DIFF
--- a/src/query/service/src/interpreters/interpreter_insert_v2.rs
+++ b/src/query/service/src/interpreters/interpreter_insert_v2.rs
@@ -395,17 +395,24 @@ impl AsyncSource for ValueSource {
             return Ok(None);
         }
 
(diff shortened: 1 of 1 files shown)


BohuTANG:
@mergify update

sundy-li (approved):

PR #9290, opened 2022-12-19
refactor: replace TableContext in planner with PlannerContext

### LINKED ISSUE #9263: Change TableContext in Binder to PlannerContext
How about name it as BinderContext ?

_Originally posted by @flaneur2020 in https://github.com/datafuselabs/databend/pull/9249#discussion_r1049244952_

everpcpc:

## Summary

* add new `PlannerContext` for planner & binder, then we can split `TableContext` into several small traits with little impact on planner later.
* move `get_stage_attachment` to `PlannerContext`
* operate stage table with `Operator` instead of context to avoid circular dependency

Closes #9263


BohuTANG:
@mergify update

youngsofun (approved):

src/query/storages/stage/src/stage_table.rs
```diff
@@ -73,17 +74,17 @@ impl StageTable {
     }
 
     /// Get operator with correctly prefix.
-    pub fn get_op(ctx: &Arc<dyn TableContext>, stage: &UserStageInfo) -> Result<Operator> {
+    pub fn get_op(dop: &DataOperator, stage: &UserStageInfo) -> Result<Operator> {
         if stage.stage_type == StageType::External {
             Ok(init_operator(&stage.stage_params.storage)?)
         } else {
-            let pop = ctx.get_data_operator()?.operator();
+            let pop = dop.operator();
```
Xuanwo, line 81: Maybe we can use `DataOperator::instance()` here directly?

src/query/storages/stage/src/file.rs
```diff
@@ -29,11 +27,11 @@ use tracing::warn;
 use crate::StageTable;
 
 pub async fn stat_file(
-    table_ctx: Arc<dyn TableContext>,
+    dop: &DataOperator,
```
Xuanwo, line 30: After `StageTable::get_op` has been refactored, we can pass an `Operator` here directly.

src/query/storages/stage/src/stage_table.rs
everpcpc: Is it OK to build a `DataOperator::instance()` for every operation?

everpcpc: hmmm, it seems `get_data_operator` also cloned the shared `DataOperator` every time.

Xuanwo: Don't worry, DataOperator is cheap to clone. We have an internal `Arc`.

BohuTANG (approved):

everpcpc:
@mergify update

PR #9293, opened 2022-12-19
feat: support ANSI quoted identifier with escape

### LINKED ISSUE #9115: Feature: Support ANSI identifier escaping
**Summary**

currently the following statement will create a table called `t1""a`, but `t1"a` is expected:

```
create table "t1""a" (a int);
```

without identifier escaping, it'd be hard for tools in ecosystems to make queries about table with strange characters, may cause execution bugs or sql injection.

flaneur2020:

## Summary

currently the following statement will create a table called t1""a, but t1"a is expected:

```
create table "t1""a" (a int);
```

this PR add the quote escape logic to identifiers, so the SQL-concating tools like sql ui or dbt can quote identifiers safely.

Closes #9115

### DIFF
--- a/src/query/ast/src/util.rs
+++ b/src/query/ast/src/util.rs
@@ -24,6 +24,7 @@ use crate::ast::Identifier;
 use crate::input::Input;
 use crate::input::WithSpan;
 use crate::parser::token::*;
+use crate::parser::unescape::unescape;
 use crate::rule;
 use crate::Error;
 use crate::ErrorKind;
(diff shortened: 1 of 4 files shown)


src/query/ast/tests/it/testdata/statement.txt
```diff
@@ -304,6 +304,46 @@ DescribeTable(
 Some(
     "TabSeparatedWithNamesAndTypes",
 )
+---------- Input ----------
+describe "name""with""quote";
+---------- Output ---------
+DESCRIBE "name"with"quote"
```
andylokandy, line 310: the display may cause bugs when persisting the SQL into a view. We've to escape it at the display stage.

src/query/ast/src/util.rs
```diff
@@ -105,10 +106,17 @@ pub fn stage_name(i: Input) -> IResult<Identifier> {
                     .filter(|c| i.1.is_ident_quote(*c))
                     .is_some()
                 {
+                    let quote = token.text().chars().next().unwrap();
```
BohuTANG, line 109: Here is handle the quoted identifier for the stage name, not the table name?

flaneur2020:
underestimate the complexity about this issue, close this PR yet, may get back this issue again in a few week.

PR #9297, opened 2022-12-20
ci: make tarball before upload failure artifact

everpcpc:

## Summary

It's too slow to upload files one by one.

### DIFF
--- a/.github/actions/artifact_failure/action.yml
+++ b/.github/actions/artifact_failure/action.yml
@@ -7,15 +7,19 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: pack failure artifacts
+      shell: bash
+      run: |
+        tar --exclude='target' \
+            --exclude='website' \
+            --exclude='docs' \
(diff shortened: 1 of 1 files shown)


PsiACE (approved):

.github/actions/artifact_failure/action.yml
```diff
@@ -7,15 +7,19 @@ inputs:
...
-        path: |
-          .
-          !./target
-          !./website
-          !./docs
-          !./docker
-          !./bins
-          !./.git
-          !./tests/suites/0_stateless/13_tpch/data
+        path: failure-${{ inputs.name }}.tar.gz
```
Xuanwo, line 25: BTW, I prefer to use `zstd` for better compression rate (

everpcpc: We do not have a zstd installed on runner currently, maybe we can change it later.
But the artifacts are actually small enough to ignore the compression rate, and `tar.gz` is more convenient for developers.

PR #9318, opened 2022-12-21
feat: purge at given point

### LINKED ISSUE #9284: Feature:  purge by travel point
**Summary**

`optimize table t purge BEFORE ( { SNAPSHOT => <snapshot_id> | TIMESTAMP => <timestamp> } );`

dantengsky:

## Summary

1. introduces `before ([SNAPSHOT|TIEMSTAMP] => ...)` subclause for `optimize table t purge` stmt

- `optimize table t purge before (SNAPSHOT => '$SNAPSHOT_ID')`
- `optimize table t purge before (TIMESTAMP => '$TIMEPOINT'::TIMESTAMP)`

which purges historical data before(exclusive) the give point

2. deprecates the usage of   `optimize table t`

   which is somewhat error-prone.

    now the user needs to specify the optimization action explicitly. i.e. one of the following actions: 
    `all`, `purge [before POINT] ` and `compact [block|segment]`

Closes #9284


src/query/storages/fuse/fuse/src/fuse_table.rs
```diff
@@ -498,13 +498,17 @@ impl Table for FuseTable {
     }
 
     #[tracing::instrument(level = "debug", name = "fuse_table_navigate_to", skip_all)]
-    async fn navigate_to(&self, point: &NavigationPoint) -> Result<Arc<dyn Table>> {
+    async fn navigate_to(
+        &self,
+        point: &NavigationPoint,
+        read_only: bool,
```
BohuTANG, line 504: The new para `read_only` is confusing:
If `read_only = true` what will do, or  `read_only=false` what will do?

Do we consider a more meaningful `enum` here?

dantengsky: if read_only is true, the table instance returned will be read only, i.e. data mutations, like insert / delete, are not allowed(runtime exception).

yes we can replace the bool with an enumeration, such as:

~~~
enum ReadOnly {
  True,
  False,
}
~~~

not sure if it is worth to do it. maybe we can upgrade the type to enum when there are more variants?

zhyass (approved):
LGTM

src/query/storages/fuse/fuse/src/fuse_table.rs
dantengsky: the `read_only` flag of fuse table instance might be over-cautioned, it seems to be enough to let the execution layer to prevention historical data from being mutated. thus, the flag has been removed.

@BohuTANG  PTAL

BohuTANG (approved):

BohuTANG:
cc @soyeric128  for doc

PR #9325, opened 2022-12-21
chore(query): default num_physical_cpus can't be zero

sundy-li:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/src/query/settings/src/lib.rs
+++ b/src/query/settings/src/lib.rs
@@ -522,11 +522,7 @@ impl Settings {
     pub fn get_max_threads(&self) -> Result<u64> {
         let key = "max_threads";
         let value = self.try_get_u64(key)?;
-        if value == 0 {
-            Ok(16)
-        } else {
-            Ok(value)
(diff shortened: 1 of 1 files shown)


src/query/settings/src/lib.rs
```diff
@@ -111,6 +111,9 @@ impl Settings {
         if conf.query.num_cpus != 0 {
             num_physical_cpus = conf.query.num_cpus;
         }
+        if num_physical_cpus == 0 {
```
BohuTANG, line 114: Printing a warn log here would be better if the num_physical_cpus is 0. It means we get it failed.

BohuTANG: image

We can log the settings that reload from the global meta in this PR.

sundy-li: I used to log that, but it'll generate many logs in each new session from @Xuanwo advices.

BohuTANG (approved):

PR #9330, opened 2022-12-21
feat(storage): rangefilter support in

### LINKED ISSUE #9331: improvement: rangefilter support in
rangefilter need support in function.

zhyass:

## Summary

expr in list => max_expr >= min_list and min_expr <= max_list

Closes #9331

### DIFF
--- a/src/query/storages/index/src/range_filter.rs
+++ b/src/query/storages/index/src/range_filter.rs
@@ -33,6 +33,7 @@ use common_sql::evaluator::Evaluator;
 use common_sql::evaluator::ExpressionMonotonicityVisitor;
 use common_sql::executor::ExpressionOp;
 use common_storages_table_meta::meta::StatisticsOfColumns;
+use itertools::Itertools;
 
 #[derive(Clone)]
 pub struct RangeFilter {
@@ -175,10 +176,9 @@ fn inverse_operator(op: &str) -> Result<&str> {
         "<=" => Ok(">="),
         ">" => Ok("<"),
         ">=" => Ok("<="),
-        "like" | "not like" | "ilike" | "not ilike" => Err(ErrorCode::UnknownException(format!(
-            "cannot inverse the operator: {:?}",
-            op
-        ))),
+        "like" | "not like" | "ilike" | "not ilike" | "in" | "not in" => Err(
+            ErrorCode::UnknownException(format!("cannot inverse the operator: {:?}", op)),
+        ),
         _ => Ok(op),
     }
 }
@@ -626,6 +626,38 @@ impl<'a> VerifiableExprBuilder<'a> {
                 }
                 Err(ErrorCode::UnknownException(
                     "Cannot build atom expression by the operator: not like",
+                ))
+            }
+            "in" => {
+                if let Expression::Constant {
+                    value: DataValue::Struct(vals),
+                    data_type: DataTypeImpl::Struct(t),
+                } = &self.args[1]
+                {
+                    // max_left >= min_right and min_left <= max_right
+                    let types = t.types();
+                    let sorted = vals
+                        .iter()
+                        .zip(types.iter())
+                        .sorted_by_key(|&(val, _)| val)
+                        .collect::<Vec<_>>();
+                    let right_min = Expression::Constant {
+                        value: sorted[0].0.clone(),
+                        data_type: sorted[0].1.clone(),
+                    };
+                    let right_max = Expression::Constant {
+                        value: sorted.last().unwrap().0.clone(),
+                        data_type: sorted.last().unwrap().1.clone(),
+                    };
+
+                    let left_min = self.min_column_expr(0)?;
+                    let left_max = self.max_column_expr(0)?;
+                    return left_max
+                        .gt_eq(&right_min)?
+                        .and(&left_min.lt_eq(&right_max)?);
+                }
+                Err(ErrorCode::UnknownException(
+                    "Cannot build atom expression by the operator: in",
                 ))
             }
             other => Err(ErrorCode::UnknownException(format!(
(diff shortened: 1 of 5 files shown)


src/query/storages/index/src/range_filter.rs
```diff
@@ -628,6 +628,38 @@ impl<'a> VerifiableExprBuilder<'a> {
...
+                {
+                    // max_left >= min_right and min_left <= max_right
+                    let types = t.types();
+                    let sorted = vals
+                        .iter()
+                        .zip(types.iter())
+                        .sorted_by_key(|&(val, _)| val)
+                        .collect::<Vec<_>>();
+                    let right_min = Expression::Constant {
+                        value: sorted[0].0.clone(),
```
sundy-li, line 645: should check sorted is empty.

BohuTANG:
It would be better to add a stateless test for the in.

zhyass:
> It would be better to add a stateless test for the in.

testcase has been added in logictest

src/query/storages/index/src/range_filter.rs
```diff
@@ -628,6 +628,41 @@ impl<'a> VerifiableExprBuilder<'a> {
                     "Cannot build atom expression by the operator: not like",
                 ))
             }
+            "in" => {
```
sandflee, line 631: sort the `in values` may decrease the filter result?   
suppose a filter like `a in [ 0, 10000]`  and a column with min-value 100, max value 1000, this column should be filtered out. but now seems not.

sandflee, line 631: is supported `not in`?

sundy-li: "not in" is not commonly used, we can support it later.

sundy-li: I did not get it.

dantengsky: @zhyass are we using the “simpler but less tighter falsifiable expression” strategy here?

x IN(𝑐1,...,𝑐𝑁) ~> min(x) > max(𝑐1, ..., 𝑐𝑁 ) OR max(x) < min(𝑐1, ..., 𝑐𝑁 )

if N is not that large, shall we consider the 

x IN(𝑐1,...,𝑐𝑁) ~> (min(x) > 𝑐1 OR max(x) < 𝑐1 ) AND ... AND (min(x)>𝑐 OR max(x)<𝑐 )

zhyass: I probably understand what you mean. 

Now the approach is similar to treating `expr in list` as `expr >= min_list and expr <= max_list`. In fact initially we had considered converting `in` to `eq or` to do it. But this would introduce some other problems.

I'll use an example to illustrate. 
```
a in (1,2,3,4)
```
If we follow the `eq or` approach, the resulting verifiable_expr would be 
```
(min_a <= 1 and max_a >= 1) or (min_a <= 2 and max_a >= 2) or (min_a <= 3 and max_a >= 3) or (min_a <= 4 and max_a >= 4)
``` 
And that's just for the case where the list has 4 values. This will have a bigger impact on performance when doing range prunning. 

So we have to consider some compromise, and there will inevitably be read amplification as you mentioned.

By the way, if the number of inlist is less than 4, the optimizer will convert the `in` to `eq or` before range_filter.

zhyass: > @zhyass are we using the “simpler but less tighter falsifiable expression” strategy here?
> 
> x IN(𝑐1,...,𝑐𝑁) ~> min(x) > max(𝑐1, ..., 𝑐𝑁 ) OR max(x) < min(𝑐1, ..., 𝑐𝑁 )
> 
> if N is not that large, shall we consider the
> 
> x IN(𝑐1,...,𝑐𝑁) ~> (min(x) > 𝑐1 OR max(x) < 𝑐1 ) AND ... AND (min(x)>𝑐 OR max(x)<𝑐 )

@dantengsky This strategy was discussed with @sundy-li before doing it. We need consider the performance impact. On the other hand, if the number of inlist is less than 4, the optimizer will convert the `in` to `eq or` before range_filter. So it may not be necessary to do so.

dantengsky: > On the other hand, if the number of inlist is less than 4, the optimizer will convert the in to eq or before range_filter. So it may not be necessary to do so.

yeah, sounds reasonable to me

dantengsky: > sort the `in values` may decrease the filter result?
> suppose a filter like `a in [ 0, 10000]` and a column with min-value 100, max value 1000, this column should be filtered out. but now seems not.

@sandflee For this case(number of items lesser than 4),  the tighter form (`eq or`) will be utilized, thus I mark this conversation as resolved. pls feel free to re-open it if any others concerns

sandflee: ok, let's check the actual improvment on the online sql.

sundy-li (approved):

dantengsky (approved):

zhyass:
@mergify update

sandflee:
@zhyass ,   I do a test with online sql, seems does't  improve too much.


|  system   | scan rows  |
|  ----  | ----  |
| databend without `rangefilter suppport in`  | 94B |
| databend with `rangefilter suppport in`  | 94B |
| trino | 12B |


there're no more than 20 items in `in clause`,sql likes:
```
SELECT `p_date`,
       `photo_id`
FROM xxx
WHERE p_date="20221217"
  and photo_id in ("xx",
                   "xx")
```

BohuTANG:
@sandflee 

How is the performance of this SQL?
```
SELECT `p_date`,
       `photo_id`
FROM xxx
WHERE p_date="20221217"
  and (photo_id = 'xx' or photo_id='yy')
```

sandflee:
> How is the performance of this SQL?

have not tested yet,  seems ok if there're just a few photo_ids in filter clause.

PR #9340, opened 2022-12-22
feat(query): purge files after insert with stage attachment

everpcpc:

## Summary

Summary about this PR

Closes #issue


src/query/service/src/interpreters/interpreter_insert_v2.rs
```diff
@@ -113,6 +117,30 @@ impl InsertInterpreterV2 {
...
+    async fn try_purge_files(
+        ctx: Arc<QueryContext>,
+        stage_info: &UserStageInfo,
+        stage_files: &[StageFileInfo],
+    ) {
+        let table_ctx: Arc<dyn TableContext> = ctx.clone();
+        let op = StageTable::get_op(stage_info);
+        match op {
+            Ok(op) => {
+                let file = Files::create(table_ctx, op);
```
ariesdevil, line 129: I think the variable name change to `file_op` will be better.

ariesdevil (approved):

src/query/service/src/interpreters/interpreter_insert_v2.rs
```diff
@@ -197,10 +228,18 @@ impl InsertInterpreterV2 {
                             start.elapsed().as_secs()
                         );
                         table
-                            .commit_insertion(ctx, append_entries, overwrite)
+                            .commit_insertion(ctx.clone(), append_entries, overwrite)
                             .await?;
 
-                        // TODO:(everpcpc) purge copied files
+                        if stage_info.copy_options.purge {
```
youngsofun, line 234: only pure when success?

everpcpc: this whole block seems all under success condition?

youngsofun: yes, I means do we need co consider to pure when fail

youngsofun (approved):

everpcpc:
@mergify update

PR #9350, opened 2022-12-23
chore(logging): disabled debug log to stderr for sled

drmingdrmer:

## Related Issues

### DIFF
--- a/.github/actions/test_unit/action.yml
+++ b/.github/actions/test_unit/action.yml
@@ -17,7 +17,7 @@ runs:
       run: cargo -Z sparse-registry test
       env:
         RUST_TEST_THREADS: "2"
-        RUST_LOG: DEBUG,sled=INFO
+        RUST_LOG: INFO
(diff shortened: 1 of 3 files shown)


Xuanwo (approved):

Xuanwo:
image

Seems not affected. I remember metasrv has its own logging settings.

drmingdrmer:
> Seems not affected. I remember metasrv has its own logging settings.

Logging level in env variable should also be updated too.

BohuTANG:
@drmingdrmer Can we change the unit test log level from DEBUG to INFO? 
DEBUG still has many logs(most are meta srv) but is not helpful most time.

drmingdrmer:
> @drmingdrmer Can we change the unit test log level from DEBUG to INFO? DEBUG still has many logs(most are meta srv) but is not helpful most time.

Should we just stop writing debug-level logs to stderr, and instead only write debug-level logs to log-file?

PR #9353, opened 2022-12-24
refactor(query): split fuse source to read data and deserialize

zhang2014:

## Summary

refactor(query): split fuse source to read data and deserialize

Closes #issue


sandflee:
@zhang2014 , why we are doing this feature? should hive adaptive related logic?

BohuTANG:
@sandflee 

This PR tries to make read data and deserialize data into different transforms and let them more parallel.

zhang2014:
> @zhang2014 , why we are doing this feature? should hive adaptive related logic?

Since hive depends on the metadata of the parquet, there may be no way to split it.

src/query/sql/src/planner/optimizer/heuristic/prewhere_optimization.rs
```diff
@@ -11,133 +11,133 @@
...
-            let children = s_expr
-                .children()
-                .iter()
-                .map(|expr| self.prewhere_optimize(expr.clone()))
-                .collect::<Result<Vec<_>>>()?;
-            Ok(SExpr::create(rel_op.clone(), children, None, None))
-        }
-    }
-}
+//
```
sundy-li, line 14: `prewhere` may be useful for other engines like Parquet engine.

We can turn off the prewhere in catalog:

```
fn support_prewhere(&self) -> bool {
        false
}
```

For fusetable.

src/query/service/src/sessions/query_ctx.rs
```diff
@@ -232,6 +232,22 @@ impl TableContext for QueryContext {
         self.partition_queue.write().pop_front()
     }
 
+    fn try_get_parts(&self, num: usize) -> Vec<PartInfoPtr> {
+        let mut res = Vec::with_capacity(num);
+        let mut partition_queue = self.partition_queue.write();
+
+        for _index in 0..num {
+            if let Some(part) = partition_queue.pop_front() {
```
BohuTANG, line 240:
```suggestion
            if let Some(part) = partition_queue.pop_front() {
                res.push(part);
            } else {
            break;
            }
```

src/query/storages/fuse/src/operations/read/native_data_source_reader.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        Ok(Event::Async)
+    }
+
+    async fn async_process(&mut self) -> Result<()> {
+        let mut parts = Vec::with_capacity(self.batch_size);
+        let mut chunks = Vec::with_capacity(self.batch_size);
+
+        for _index in 0..self.batch_size {
```
BohuTANG, line 133: let parts = self.ctx.try_get_parts(self.batch_size)?

BohuTANG (approved):

src/query/storages/fuse/src/operations/read/native_data_source_reader.rs
BohuTANG: Else LGTM :/

Issue #8995, opened 2022-11-28
Feature(VIP):  unify format options and remove `format_` settings


youngsofun:
# background

we need many options for diff formats https://github.com/datafuselabs/databend/issues/8914

suppose we want to load a file with file format CSV and field delimiter "|".
we have 3 ways:

streaming load handler

```
curl  -u root: \
-H "insert_sql:insert into t1 format CSV"  \
-F "upload=@foo.csv"  \
-H "format_field_delimter=|" \
-XPUT http://127.0.0.1:8000/v1/streaming_load
```

clickhouse handler 

```
 curl 'http://localhost:8123/?query=insert%20into%20t1%20format%20CSV?format_field_delimter=| \
 --data-binary @foo.csv
```

copy SQL

```
copy into t1 from 's3://mybucket/foo.csv'
 FILE_FORMAT = (type = 'CSV'  field_delimiter = '|')
```

for the first two `format_field_delimter` is actually apply to the settings of the session (for the HTTP request).
 we can also pass any other settings the same way.
 

 So to summarize
 
 we have two repr for  "file format with format option"
 
 - SQL clause `FILE_FORMAT = (type = 'CSV'  some_option = '|')``,  for copy
 - SQL clause `FORMAT CSV` and the `format_xxx` settings, for insert with clickhouse and streaming load
 
 
 # what`s wrong with it
 
 
 Firstly, settings is  essentially **not ** suitable to store format option.
 
1.  the ·(type = 'CSV'  some_option = '|')` is a whole, should be set/update atomicly
2. setting must have one default value, 2 problem:
    1. we can not tell if it is set by user on purpose or set (global) by other users or just default.   in contrast, in `FILE_FORMAT = (type = 'CSV'  some_option = '|')`, it is clear weather an option 
  not appeared. and we have the `type = 'CSV` as context.
    2. diff format has diff default value

  
  
 
 
 Secondly, no good for compatiblily.

 1. for streaming load, we are free to design our style.
 2. for clickhouse, we are still not  compatible with
  clickhouse about the settings be cause the set of settings are different.
 
 thirdly, hard to use.
 
 1. not unified
 2. `format_` prefx make it long and easy to forget
 3. there will be many options later, add to a lot of settings.
 2. http header is hard to check and http param need escape.

 
 fourthly, hard to maintain.
 
 1. harder to check and report error
 2. need more code to add an option and test them
 
 
 # propose
 
[...]

PR #9360, opened 2022-12-26
refactor: remove format_xxx settings.

youngsofun:

## Summary

Summary about this PR

Closes https://github.com/datafuselabs/databend/issues/8995

some futher refactor need to be done, after this pr is merged.


Xuanwo (approved):

BohuTANG:
We should document them as deprecated and instead use the new formats.

youngsofun:
> We should document them as deprecated and instead use the new formats.

docs are already updated to the new format last week.

youngsofun:
1. old format is not mentioned in current docs
2. when using old format  will get error message like "Only supports INSERT INTO $table FILE_FORMAT = (type = 'CSV' ..)"

src/query/service/src/servers/http/v1/load.rs
```diff
@@ -229,8 +178,16 @@ pub async fn streaming_load(
                     )),
                 }
             }
+            InsertInputSource::StreamingWithFormat(_, _, _) => Err(poem::Error::from_string(
+                "'INSERT INTO $table FORMAT <type> is now only supported in clickhouse handler,\
+                     please use 'FILE_FORMAT = (type = <type> ...)' instead.",
+                StatusCode::BAD_REQUEST,
+            )),
```
youngsofun, line 185: @BohuTANG refined the error message.

ariesdevil (approved):

Issue #9226, opened 2022-12-13
bug: DISTINCT or UNION logic bug


hanyisong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

8.0.26-v0.8.150-nightly

### What's Wrong?

It is expected that the number of rows returned by the previous two sqls is equal, but they are actually not equal.
```sql
SELECT DISTINCT t2.c0float, t1.c0boolean, t0.c0varchar FROM t0, t2 NATURAL LEFT JOIN t1;
-- cardinality: 8
SELECT t2.c0float, t1.c0boolean, t0.c0varchar FROM t0, t2 NATURAL LEFT JOIN t1 WHERE ((((((((((((((t1.c0boolean)or(false)))AND(t1.c0boolean)))AND(true)))or(true)))and(true)))OR((NOT false))))AND(((((t1.c0boolean)AND(t1.c0boolean)))or(((true)AND(t1.c0boolean)))))) UNION SELECT t2.c0float, t1.c0boolean, t0.c0varchar FROM t0, t2 NATURAL LEFT JOIN t1 WHERE (NOT ((((((((((((((t1.c0boolean)OR(false)))and(t1.c0boolean)))AND(true)))OR(true)))AND(true)))OR((NOT false))))AND(((((t1.c0boolean)AND(t1.c0boolean)))OR(((true)and(t1.c0boolean))))))) UNION SELECT t2.c0float, t1.c0boolean, t0.c0varchar FROM t0, t2 NATURAL LEFT JOIN t1 WHERE ((((((((((((((((t1.c0boolean)OR(false)))AND(t1.c0boolean)))and(true)))OR(true)))AND(true)))OR((NOT false))))and(((((t1.c0boolean)AND(t1.c0boolean)))or(((true)and(t1.c0boolean))))))) IS NULL);
-- cardinality: 4
```

### How to Reproduce?

```sql
DROP DATABASE IF EXISTS databend2;
CREATE DATABASE databend2;
USE databend2;
CREATE TABLE t0(c0VARCHAR VARCHAR NOT NULL DEFAULT('EHFUF'));
CREATE TABLE t1(c0BOOLEAN BOOL NOT NULL);
CREATE TABLE t2(c0FLOAT DOUBLE NULL);
INSERT INTO t0(c0varchar) VALUES (''), ('R6a');
INSERT INTO t1(c0boolean) VALUES (true), (false);
[...]

PR #9361, opened 2022-12-26
fix: the data type on both sides of the union does not match

xudong963:

## Summary

For SQL:

```sql
SELECT t1.c0boolean, t0.c0varchar FROM t0 NATURAL LEFT JOIN t1 UNION SELECT t1.c0boolean, t0.c0varchar FROM t0 NATURAL LEFT JOIN t1 where t1.c0boolean;
```

The left side of union is a **left join**, but the right side of union will be convert to **inner join** due to optimize rule: https://github.com/datafuselabs/databend/blob/5d75f849ca2f118b1fb6a61a744da7b299d46827/src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_filter_join.rs#L151

So the data type on both sides of the union does not match, the left side will be nullable for `t1.c0boolean`, but the right side isn't nullable for `t1.c0boolean`.

Closes https://github.com/datafuselabs/databend/issues/9226


sundy-li (approved):

xudong963:
@mergify update

PR #9381, opened 2022-12-27
feat: external storage support location part prefix

### LINKED ISSUE #8801: feat: fuse external table block location  prefix
**Summary**

Raw name:
image

external table:
```
create table demo(name VARCHAR, age int) 
's3://bohu/fuse/' -- your bucket
connection=(ACCESS_KEY_ID='<your-key>' SECRET_ACCESS_KEY='<your-secret>')
location_prefix='my_';
```

```
my_0207b336bbef4cf6808724f7cd255653_v0.parquet
```

ariesdevil:

## Summary

External storage support location part prefix

Closes #8801


src/query/ast/src/parser/stage.rs
```diff
@@ -79,14 +79,21 @@ pub fn uri_location(i: Input) -> IResult<UriLocation> {
             ~ (CONNECTION ~ "=" ~ #options)?
             ~ (CREDENTIALS ~ "=" ~ #options)?
             ~ (ENCRYPTION ~ "=" ~ #options)?
+            ~ (LOCATION_PREFIX ~ "=" ~ #literal_string)?
```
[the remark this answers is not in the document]
ariesdevil: `LOCATION_PREFIX` for filename prefix, eg: you have a bucket with a lot of files, you just want to using files that starts with `databend_`(databend_xxx.parquet)

tests/suites/1_stateful/02_query/02_0001_create_table_with_external_location.result
```diff
@@ -1,2 +1,3 @@
 888
 1024
+lulu
```
drmingdrmer, line 3: hello lulu

src/query/storages/fuse/src/io/locations.rs
```diff
@@ -39,32 +39,52 @@ static SNAPSHOT_STATISTICS_V0: TableSnapshotStatisticsVersion =
...
+                "{}/{}/{}_{}_v{}.parquet",
+                &self.prefix,
+                FUSE_TBL_BLOCK_PREFIX,
+                &self.part_prefix,
+                part_uuid.as_simple(),
+                DataBlock::VERSION,
+            )
+        };
+
+        ((location_path, DataBlock::VERSION), part_uuid)
```
drmingdrmer, line 87: Could the prefix simply be `my_`?
So that we do not have to concatenate the prefix `my` and part name with a `_`.

drmingdrmer (commented):
The metadata part is fine enough. 💯 

Having a little question about how the prefix is interpreted.

src/query/storages/fuse/src/io/locations.rs
ariesdevil: After discuss with @drmingdrmer , we decide:
1. when user input is `my`, we add a `_` and store in table meta.
2. when user input is `my_`, we no need add `_`.
3. after this we no need concat with `_` here.
cc @BohuTANG any suggestions?

Xuanwo: > * when user input is `my`, we add a `_` and store in table meta.
> * when user input is `my_`, we no need add `_`.

We can't predict user's data layout. User could have data like `myb_v2.parquet`.

In fact, I think this feature doesn't make sense. User rarely separate different data files based on prefix `mya`, `myb`. Instead, they will use `mya/`, `myb/`. This is what `root` designed for.

Users just need to use:

- `s3://bucket/path/to/mya/`
- `s3://bucket/path/to/myb/`

We don't need to support the case like `myaxxx_v2` or `mybxxx_v2`.

ariesdevil: We cannot assume that users store files in separate buckets, and that their upstream and downstream processing data may be based on different files inside the same bucket

Xuanwo: > We cannot assume that users store files in separate buckets

No, they are in the same bucket with different root.

`s3://bucket/path/to/mya/` => bucket: `bucket`, root: `/path/to/mya/`.

ariesdevil: I understand that this FEATURE is rarely used, but as said above, we can't assume user behavior.

Xuanwo: > I understand that this FEATURE is rarely used, but as said above, we can't assume user behavior.

If so, I prefer to not do any modification on user's input `prefix`. Just add them in our path, no matter user input `a` or `a_`, `a__`, `a_/_/_`.

BohuTANG: So far, the fuse engine has no dependency on the block filename, we are saving it to the segment file.

src/query/storages/fuse/src/io/locations.rs
```diff
@@ -39,32 +39,52 @@ static SNAPSHOT_STATISTICS_V0: TableSnapshotStatisticsVersion =
...
+                "{}/{}/{}{}_v{}.parquet",
+                &self.prefix,
+                FUSE_TBL_BLOCK_PREFIX,
+                &self.part_prefix,
+                part_uuid.as_simple(),
+                DataBlock::VERSION,
+            )
+        };
+
+        ((location_path, DataBlock::VERSION), part_uuid)
```
drmingdrmer, line 87: It looks the if and else blocks can be merged into one block.

ariesdevil: Yup

src/query/storages/fuse/src/fuse_table.rs
```diff
@@ -138,9 +138,16 @@ impl FuseTable {
             .cloned()
             .unwrap_or_default();
 
+        let part_prefix = table_info.meta.part_prefix.clone();
+
+        let mut meta_location_generator = TableMetaLocationGenerator::with_prefix(storage_prefix);
+        if !part_prefix.is_empty() {
+            meta_location_generator = meta_location_generator.with_part_prefix(part_prefix);
+        }
```
drmingdrmer, line 146: This if-block is unnecessary too.

drmingdrmer (approved):

BohuTANG (approved):

PR #9391, opened 2022-12-28
feat(sqllogictests): Add no-fail-fast support

xudong963:

## Summary

Summary about this PR

Closes #issue


Xuanwo:
Please add a args for this. We want CI to fail fast so they can exit more quick.

xudong963:
> Please add an args for this. We want CI to fail fast so they can exit more quick.
> 
> For example:
> 
> ```
> > cargo test --help
>     --no-fail-fast              Run all tests regardless of failure
> ```

Make sense. cc @zhang2014 @sundy-li  Any thoughts?

Xuanwo:
We can change the workflow files if we have special needs during refactor. And change it back before merging.

sundy-li (approved):

sundy-li:
>  We want CI to fail fast so they can exit more quickly.
Ok in online ci.

In local dev, we can pass this arg to have all errors.

Xuanwo (approved):

PR #9399, opened 2022-12-28
feat: add more metrics for fuse compact and block write

BohuTANG:

## Summary

* add metrics for block and block index write
* add metrics for compact
* add metrics for pruning
* remove unused metrics

Closes #issue


BohuTANG:
Only add metrics, let's merge.

PR #9400, opened 2022-12-28
feat: run all tests in parallel

xudong963:

## Summary

1. use `sandbox_tenant` to isolate each test file.
2. run all test files in parallel.

I did a simple test, running crdb suit. **Run time will be reduced twice**

Closes #issue


tests/sqllogictests/src/client/clickhouse_client.rs
```diff
@@ -43,6 +43,7 @@ impl ClickhouseHttpClient {
     }
 
     pub async fn query(&mut self, sql: &str) -> Result<DBOutput> {
+        println!("Running sql with clickhouse client: [{}]", sql);
```
Xuanwo, line 46: Please use `debug` or `info`.

xudong963: If we use them, we'll get some noise from sqllogictest-rs.

I think those logs are basic and useful, maybe it's better to print them always.

tests/sqllogictests/src/main.rs
```diff
@@ -142,7 +142,10 @@ async fn create_databend(client_type: &ClientType) -> Result<Databend> {
             client = Client::Clickhouse(ClickhouseHttpClient::create()?);
         }
     }
-    client.create_sandbox().await?;
+    let suites = SqlLogicTestArgs::parse().suites;
+    if suites != "tests/fuse-compat/compat-logictest" {
```
xudong963, line 146: To take care of compatibility testing, `set sandbox_tenant` was not supported in past versions.

Maybe we should find a better way to test compat.  Such as use the traditional test way not using sqllogictest.

cc @Xuanwo @sundy-li

xudong963:
@mergify update

tests/sqllogictests/src/main.rs
sundy-li: ok

BohuTANG (approved):

PR #9402, opened 2022-12-29
feat: migrate hash func to func-v2

TCeason:

## Summary

migrate hash func to func-v2

Closes #issue

### DIFF
--- a/src/query/functions-v2/src/scalars/hash.rs
+++ b/src/query/functions-v2/src/scalars/hash.rs
@@ -1,0 +1,350 @@
+// Copyright 2022 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::hash_map::DefaultHasher;
+use std::hash::Hash;
+use std::hash::Hasher;
+
+use common_expression::types::number::NumberScalar;
+use common_expression::types::number::F32;
+use common_expression::types::number::F64;
+use common_expression::types::BooleanType;
+use common_expression::types::DateType;
+use common_expression::types::NumberDataType;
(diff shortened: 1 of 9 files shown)


src/query/functions-v2/src/scalars/hash.rs
```diff
@@ -0,0 +1,309 @@
...
+                output.commit_row();
+                Ok(())
+            },
+        ),
+    );
+
+    for ty in MAX_NUMERICS_TYPES {
+        with_number_mapped_type!(|NUM_TYPE| match ty {
+            NumberDataType::NUM_TYPE => {
+                registry.register_passthrough_nullable_2_arg::<StringType, NumberType<NUM_TYPE>, StringType, _, _>(
```
sundy-li, line 105: `<StringType, Int64Type, StringType>` is enough here. 

The second arg is mostly constant.

src/query/functions-v2/src/scalars/hash.rs
```diff
@@ -0,0 +1,309 @@
...
+        match self {
+            Scalar::Number(NumberScalar::Int64(v)) => DFHash::hash(v, state),
+            Scalar::Number(NumberScalar::UInt64(v)) => DFHash::hash(v, state),
+            Scalar::Number(NumberScalar::Float64(v)) => DFHash::hash(v, state),
+            Scalar::String(vals) => {
+                for v in vals {
+                    DFHash::hash(v, state);
+                }
+            }
+            _ => {}
```
sundy-li, line 306: what if the other is uint8 ... ?

`with_number_mapped_type` could be used there.

src/query/functions-v2/src/scalars/hash.rs
TCeason: I do a test, and we support this in function v1...

```sql
select sha2('x',224.0);
+----------------------------------------------------------+
| sha2('x', 224)                                           |
+----------------------------------------------------------+
| 54a2f7f92a5f975d8096af77a126edda7da60c5aa872ef1b871701ae |
+----------------------------------------------------------+

```

TCeason:
In function v1, [DFHash](https://github.com/datafuselabs/databend/blob/cdf3bc22c097f0eed8bb79567a00628c63adfbee/src/query/functions/src/scalars/hashes/hash_base.rs#L189) has these:

impl<'a> DFHash for &'a VariantValue 
impl<'a> DFHash for ArrayValueRef<'a> 
impl<'a> DFHash for StructValueRef<'a> 

How to convert it into function-v2?

Need help cc @sundy-li  @andylokandy

sundy-li (approved):

PR #9406, opened 2022-12-29
feat(query): add max_query_row_nums

sandflee:

## Summary

user may run a query `select * from xx`, this may scan too many rows.  add a default limiter if user does't specify limit values.

Closes #issue


sundy-li:
better name to `max_result_rows` .

BohuTANG:
@mergify update

BohuTANG:
@sandflee 

It would be great to add a SQL logic test for it so that we can keep the change always working :)

sandflee:
> It would be great to add a SQL logic test for it so that we can keep the change always working :)

could u provide a example？

BohuTANG:
Sure.
For example:

tests/sqllogictests/suites/base/20+_others/20_0011_setting
```
statement ok
DROP DATABASE IF EXISTS db1

statement ok
CREATE DATABASE db1

statement ok
USE db1

statement ok
CREATE TABLE IF NOT EXISTS t1(a INT) Engine = fuse

statement ok
INSERT INTO t1 (a) values (1), (2), (3)

query I
SELECT COUNT() FROM t1
----
3

query I
SELECT COUNT() FROM t1
----
1

statement ok
DROP TABLE t1

statement ok
DROP DATABASE db1

```

src/query/settings/src/lib.rs
```diff
@@ -388,6 +388,13 @@ impl Settings {
                 desc: "the max number of rows each read from parquet to databend processor",
                 possible_values: None,
             },
+            SettingValue {
+                default_value: UserSettingValue::UInt64(0),
+                user_setting: UserSetting::create("max_result_rows", UserSettingValue::UInt64(0)),
+                level: ScopeLevel::Session,
+                desc: "Auto limit max result rows if user not specify the limit, default is 0 means no limit",
```
BohuTANG, line 395: `Auto limit max result rows if user not specify the limit, default is 0 means no limit`-->`Auto limit max result rows if user specify the limit, default is 0 means no limit`?

sandflee: `max_result_rows` does't take affect  If user specifies the limit ,such as select * from xx limit 10.

BohuTANG: Ok.
It would be better to add a new case to the test: `select count() from (select * from t1 limit 2)` after settings `SET max_result_rows=1`.

sandflee: ok

BohuTANG (approved):
2023

PR #9420, opened 2022-12-30
fix(format): catch unwind when read split.

youngsofun:

## Summary

Summary about this PR

```
mysql> copy into traces from 'fs:///Users/yangxiufeng/work/dec.parquet' file_format = (type="parquet") force=true;
ERROR 1105 (HY000): Code: 1104, displayText = not implemented: data_type: Decimal(38, 9).
```

Closes #issue


BohuTANG:
@mergify update

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
```diff
@@ -203,13 +204,28 @@ pub trait InputFormatPipe: Sized + Send + 'static {
...
                         futs.push(fut);
                     }
                     while let Some(row_batch) = futs.next().await {
-                        if data_tx2.send(row_batch.unwrap()).await.is_err() {
-                            info!("execute_copy_aligned fail to send row_batch")
+                        match row_batch {
+                            Ok(row_batch) => {
+                                let is_err = row_batch.is_err();
+                                if data_tx2.send(row_batch).await.is_err() {
+                                    info!("execute_copy_aligned fail to send row_batch")
```
ariesdevil, line 217: Why use info instead of error here?

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
```diff
@@ -203,13 +204,28 @@ pub trait InputFormatPipe: Sized + Send + 'static {
...
+                                }
+                                if is_err {
+                                    break;
+                                }
+                            }
+                            Err(cause) => {
+                                if data_tx2.send(Err(cause)).await.is_err() {
+                                    info!("execute_copy_aligned fail to send row_batch")
```
ariesdevil, line 225: Same as above

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
```diff
@@ -203,13 +204,28 @@ pub trait InputFormatPipe: Sized + Send + 'static {
                 let splits = splits.to_owned().clone();
                 tokio::spawn(async move {
                     let mut futs = FuturesUnordered::new();
-                    for s in &splits {
-                        let fut = Self::read_split(ctx_clone2.clone(), s);
+                    for s in splits.into_iter() {
```
BohuTANG, line 207: To run batch futures, we can try this way?
https://github.com/datafuselabs/databend/blob/757415a8bb741fed56e9b6307122834d6f8766f3/src/query/storages/fuse/src/io/snapshots.rs#L93-L117

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
youngsofun: most of the time, if the receiver closes first，the processor in the receiver side may encounter error, 
so we in fact do not need the log.  logging it to help the debug.

channel is a workaroud, later we will read in pipeline,  and not need the channel any more.

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
youngsofun: here we need to process a future result as soon as it finishes in stead of try_join_all.

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
youngsofun: Same as above 😄

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
BohuTANG: This info log may flood the log, please remove it.

BohuTANG: > most of the time, if the receiver closes first，the processor in the receiver side may encounter error, so we in fact do not need the log. logging it to help the debug.
> 
> channel is a workaroud, later we will read in pipeline, and not need the channel any more.

This log is not useful here, it looks should be an error and log the error msg.

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
youngsofun: need to wrap a future of  `read_split + send` to use try_spawn_batch.
using semaphore if fully pipelined compared with `splits.chunks(p)`  in which batches is not ovelaped

youngsofun: since this  pr is a bug fix,  can I do the refactor in a separate pr later?

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
youngsofun: done

src/query/pipeline/sources/src/processors/sources/input_formats/input_pipeline.rs
```diff
@@ -203,13 +203,26 @@ pub trait InputFormatPipe: Sized + Send + 'static {
...
+                                    break;
+                                }
+                                if is_err {
+                                    break;
+                                }
+                            }
+                            Err(cause) => {
+                                data_tx2.send(Err(cause)).await.ok();
+                                break;
+                            }
```
BohuTANG, line 225:
```suggestion
match row_batch {
    Ok(row_batch) => {
        if let Err(cause) = data_tx2.send(row_batch).await {
            break;
        }
    }
    Err(cause) => {
        if let Err(send_error) = data_tx2.send(Err(cause)).await {
            break;
        }
    }
}
```

youngsofun: this not help with `on_error=continue`,  because the receiver side will abort the query,  should not try to read/send data any more.

as I said, there are many other kinds of error to handle (not only read row) to better support `on_error=continue` .
compare with lost data silently,  I prefer fail for  errors happend here.

youngsofun: cc @ariesdevil

ariesdevil: Why not:
```rust
if row_batch.ok() {
	if data_tx2.send(row_batch).await.is_err() {
	    break;
	}
} else {
    break;
}
```

youngsofun: why?

youngsofun: done @ariesdevil

ariesdevil (approved):

BohuTANG (approved):

PR #9379, opened 2022-12-27
refactor(query): add uncompressed buffer for parquet reader

zhang2014:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

refactor(query): add uncompressed buffer for parquet reader

Closes #issue

PR #9425, opened 2022-12-30
feat: introduce built-in malloc conf

dantengsky:

## Summary

introduces built-in malloc conf  `percpu_arena:percpu,oversize_threshold:0,muzzy_decay_ms:5000,dirty_decay_ms:5000`.

benchmarks(mainly on the test data set of `hits`) show that the above malloc conf can significantly reduce the system CPU usage (page faults) caused by the large size memory allocations, e.g. during decompression, etc.

thus we make this configuration as default during building. 

during startup of databend-query, malloc conf will be shown like this:

~~~
Databend Query

Version: v0.8.167-nightly-c32f0d1(rust-1.67.0-nightly-2022-12-30T07:38:55.015661976Z)

Logging:
...
Memory:
    limit: unlimited
    allocator config: percpu_arena:percpu,oversize_threshold:0,muzzy_decay_ms:5000,dirty_decay_ms:5000
...
~~~

the MALLOC_CONFI is taken from clickhouse :

https://github.com/ClickHouse/ClickHouse/blob/533c7888038453c047df816f3f65292cca05a54f/contrib/jemalloc-cmake/CMakeLists.txt#L37

we have evaluated this configuration before, with no significant improvement has been observed. 

But after @zhang2014 's PR https://github.com/datafuselabs/databend/pull/9379 has been merged, the above malloc conf finally takes effect.

------------------

NOTE:

- **`$PROJ_ROOT/.cargo/config.toml` is no longer ignored by `.gitingroe`**

if this conflicts with your concurrent dev env setup, please consider moving the current one to `~/.cargo/config.toml` or another parent path, for more details, please see https://doc.rust-lang.org/cargo/reference/config.html 

- compile time conf can be overridden by set `JEMALLOC_SYS_WITH_MALLOC_CONF`
  
   e.g. `JEMALLOC_SYS_WITH_MALLOC_CONF=oversize_threshold:0 cargo b -r `
   
- runtime conf can be  be overridden by set `_RJEM_MALLOC_CONF`

  e.g.  ` _RJEM_MALLOC_CONF="muzzy_decay_ms:15000,dirty_decay_ms:15000" ./target/release/databe-query -c ....`

Closes #issue

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[env]
+JEMALLOC_SYS_WITH_MALLOC_CONF = "percpu_arena:percpu,oversize_threshold:0,muzzy_decay_ms:5000,dirty_decay_ms:5000"
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@
 
 # Rust
 target/
-.cargo/config.toml
 
 # OS related
 ### MacOS ###
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2991,6 +2991,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sharing-endpoint",
+ "tikv-jemalloc-ctl",
  "tokio-stream",
  "tonic",
  "tracing",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,6 +103,9 @@ serde_json = { version = "1.0.85", default-features = false, features = ["preser
 chrono = { version = "0.4.22", features = ["serde"] }
 chrono-tz = "0.6.3"
 
+# jemalloc
+tikv-jemalloc-ctl = { version = "0.5.0", features = ["use_std"] }
+
 # http
 reqwest = { version = "0.11", default-features = false, features = [
     "json",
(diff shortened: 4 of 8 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+JEMALLOC_SYS_WITH_MALLOC_CONF = "percpu_arena:percpu,oversize_threshold:0,muzzy_decay_ms:5000,dirty_decay_ms:5000"
```
Xuanwo, line 2: How about add a link to https://github.com/jemalloc/jemalloc/blob/dev/TUNING.md ?

dantengsky: afraid that the configuration used in this PR is not derived from the official tuning guidance. the essential option `oversize_threshold:0` is not mentioned in TUNING.md

Instead, it is from clickhouse :

https://github.com/ClickHouse/ClickHouse/blob/533c7888038453c047df816f3f65292cca05a54f/contrib/jemalloc-cmake/CMakeLists.txt#L37

we have evaluated this configuration before, but no significant improvement been observed. 

But, after @zhang2014 's PR https://github.com/datafuselabs/databend/pull/9379 has been merged, the above malloc conf finally takes effect.

Let me add the above info the PR's summary

BohuTANG (approved):

PR #8166, opened 2022-10-12
refactor: Remove not used disk cache

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

refactor: Remove not used disk cache

DiskCache is not used in any of our code. Remove it to make space.

Part of https://github.com/datafuselabs/databend/issues/6803

PR #9436, opened 2022-12-31
feat(cache): add generic cache trait for different object reader

BohuTANG:

### Goal 

The goal is for each object cache provider to be more flexible and able to define its cache policy:
```
CachedObjectAccessor -> Cache Provider(ByPassCache/MemoryBytesCache/MemoryItemsCache/FileCache) --> OpenDAL object operations
```

### Usage

Usage(MemoryItemsCache):
```
    let mut builder: Builder = fs::Builder::default();
    builder.root("/tmp");
    let op: Operator = Operator::new(builder.build()?);

    let path = uuid::Uuid::new_v4().to_string();
    let object = op.object(&path);

    // Cache provider.
    let settings = CacheSettings::default();
    let cache = Arc::new(MemoryItemsCache::create(&settings));

    // Cached Object Accessor.
    let accessor = CachedObjectAccessor::create(cache.clone());
    accessor.write(&object, expect.clone()).await?;

    let actual: Arc<TestMeta> = accessor.read(&object, 0, 16).await?;
```

Usage(ByPassCache = No cache, same as object operations via CachedObjectAccessor):
```
    let mut builder: Builder = fs::Builder::default();
    builder.root("/tmp");
    let op: Operator = Operator::new(builder.build()?);

    let path = uuid::Uuid::new_v4().to_string();
    let object = op.object(&path);

    // Cache provider.
    let settings = CacheSettings::default();
    let cache = Arc::new(ByPassCache::create(&settings));

    // Cached Object Accessor.
    let accessor = CachedObjectAccessor::create(cache.clone());
    accessor.write(&object, expect.clone()).await?;

    let actual: Arc<TestMeta> = accessor.read(&object, 0, 16).await?;
```

### Notes

This is the first part. We implement three cache backends:
```
ByPassCache. -- DONE
MemoryBytesCache -- DONE
MemoryItemsCache -- DONE
FileCache -- TODO , for block data
```

Next PR will try to replace the following:
`SegmentInfoCache/TableSnapshotCache/TableSnapshotStatisticCache/BloomIndexCache/BloomIndexMetaCache/FileMetaDataCache` to the new.

Closes #issue


src/query/storages/cache/src/providers/file_cache.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::sync::Arc;
+
+use common_exception::Result;
+use opendal::Object;
+
+use crate::CacheSettings;
+use crate::ObjectCacheProvider;
+
+// TODO(bohu): local disk file based LRU cache
+pub struct FileCache {}
```
PsiACE, line 24: In the past, common-cache included a disk-cache implementation. Removed in #8166, perhaps we could consider taking it back.

src/query/storages/cache/src/providers/memory_bytes_cache.rs
```diff
@@ -0,0 +1,86 @@
...
+use parking_lot::RwLock;
+
+use crate::CacheSettings;
+use crate::ObjectCacheProvider;
+
+type BytesCache = RwLock<LruCache<String, Arc<Vec<u8>>, DefaultHashBuilder, BytesMeter>>;
+
+/// Memory LRU cache for bytes(Vec<u8>).
+pub struct MemoryBytesCache {
+    lru: BytesCache,
```
PsiACE, line 32: Although it is currently Lru, I think we could consider using a name that doesn't point so much to a specific algorithm in MemoryBytesCache & MemoryItemsCache.

BohuTANG: Memory cache should always LRU.

src/query/storages/cache/src/providers/file_cache.rs
BohuTANG: For file cache, there are many to consider, for example: if we have a range cached: <location, 100, 150>, if a request is <location, 101, 140>, we should consider the overlap.

BohuTANG:
Let's merge and continue the work.

PR #9437, opened 2022-12-31
feat: Check connection params

### LINKED ISSUE #8380: bug: external table syntax need key workds check
### Search before asking

### Version

```
MySQL [(none)]> select version();
+------------------------------------------------------------------------------------+
| version()                                                                          |
+------------------------------------------------------------------------------------+
| 8.0.26-v0.8.80-nightly-d0006ac(rust-1.66.0-nightly-2022-10-19T09:51:51.600643433Z) |
+------------------------------------------------------------------------------------+
1 row in set (0.017 sec)

```

### What's Wrong?

```
MySQL [(none)]> create table tb101 (id int ,c1 datetime) 's3://wubx/tb101' connection=(aws_key_id='minioadmin' aws_ssecret_key='minioadmin' endpoint_url='http://127.0.0.1:9900');

ERROR 1105 (HY000): Code: 3901, displayText = current configured storage is not available: config: S3(StorageS3Config { endpoint_url: "http://127.0.0.1:9900", region: "", bucket: "wubx", root: "/", disable_credential_loader: true, enable_virtual_host_style: false, role_arn: "", external_id: "", access_key_id: "******min", secret_access_key: "", security_token: "", master_key: "" }), cause: object error: (op: list, path: /, source: status code: 403, headers: {"accept-ranges": "bytes", "content-length": "259", "content-security
MySQL [(none)]> 
````
this error root cause: aws_secret_key be written into aws_ssecret_key 
So I think  there need  key workds check.

### How to Reproduce?

above

### PR DESCRIPTION
**Pull request by ariesdevil**

## Summary

Check input connection params when do parse URI.

Closes #8380

### DIFF
--- a/src/query/sql/src/planner/binder/copy.rs
+++ b/src/query/sql/src/planner/binder/copy.rs
@@ -100,7 +100,7 @@ impl<'a> Binder {
                     .unwrap_or_else(|| self.ctx.get_current_database());
                 let table = normalize_identifier(table, &self.name_resolution_ctx).name;
 
-                let ul = UriLocation {
+                let mut ul = UriLocation {
                     protocol: uri_location.protocol.clone(),
                     name: uri_location.name.clone(),
                     path: uri_location.path.clone(),
@@ -111,7 +111,7 @@ impl<'a> Binder {
                 self.bind_copy_from_uri_into_table(
                     bind_context,
                     stmt,
-                    &ul,
+                    &mut ul,
                     &catalog_name,
                     &database_name,
                     &table,
@@ -165,7 +165,7 @@ impl<'a> Binder {
                     .unwrap_or_else(|| self.ctx.get_current_database());
                 let table = normalize_identifier(table, &self.name_resolution_ctx).name;
 
-                let ul = UriLocation {
+                let mut ul = UriLocation {
                     protocol: uri_location.protocol.clone(),
                     name: uri_location.name.clone(),
                     path: uri_location.path.clone(),
@@ -179,7 +179,7 @@ impl<'a> Binder {
                     &catalog_name,
                     &database_name,
                     &table,
-                    &ul,
+                    &mut ul,
                 )
                 .await
             }
@@ -194,7 +194,7 @@ impl<'a> Binder {
                 .await
             }
             (CopyUnit::Query(query), CopyUnit::UriLocation(uri_location)) => {
-                let ul = UriLocation {
+                let mut ul = UriLocation {
                     protocol: uri_location.protocol.clone(),
                     name: uri_location.name.clone(),
                     path: uri_location.path.clone(),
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/query/sql/src/planner/binder/location.rs [hunk 1]
```diff
@@ -44,6 +48,91 @@ fn secure_omission(endpoint: String) -> String {
     }
 }
 
+static SCHEME_PARAMS: OnceCell<HashMap<Scheme, HashSet<String>>> = OnceCell::new();
+
+fn check_uri(protocol: &Scheme, conns: &BTreeMap<String, String>) -> Result<()> {
```
Xuanwo, line 53: Compared to a general `check_uri`, how about implementing functions like `parse_s3_uri_location`? In which we will do `parse_s3_uri_location(l: UriLocation) -> Result<StorageParams>`.

Implementing `parse_xxx_uri_location` for every scheme could produce more code but will

- make future maintaining easier (all code related to this scheme belongs here).
- Remove the need of `HashMap<Scheme, HashSet<String>>`, we can have a static hashset for every scheme.
- Make unit tests for easier.

ariesdevil: I initially considered this approach, but

1. it would generate a lot of code
2. there would be a lot of global HashSet

Considering that the number of parameters for each cloud platform is not very large, I finally took a unified approach.

Xuanwo: We do these checks in a dynamic way.

For example, build a `ConnectionChecker` which provides the following API:

- `get(key) -> Value`: store the key in a visited `HashSet` and than fecth value from `conn` HashMap.
- `check() -> Result<()>`: after all keys have been checked, we can compared the visited hashset with user input's `conn`, if user inputs have the key that not in `visisted`, we can return an error with UnknownKeyInputted.

In this way, we don't need to maintain a static hashset for every services.

ariesdevil: If we split up the functions by scheme, it would result in a lot of duplicate code, and the logic would be almost identical except construction of different StorageParams by scheme.

Xuanwo: My idea is 

image

ariesdevil:
A little bit of spit for Rust:

The mutability of rust is unusual, and I'm not sure if it has anything to do with its borrow checker.

In Scala or OCaml, the `HashMap` itself is immutable, but the elements contained in it are mutable.

src/query/sql/src/planner/binder/copy.rs
```diff
@@ -194,15 +194,15 @@ impl<'a> Binder {
                 .await
             }
             (CopyUnit::Query(query), CopyUnit::UriLocation(uri_location)) => {
-                let ul = UriLocation {
+                let mut ul = UriLocation {
```
Xuanwo, line 197: Maybe we can keep `UriLocation` unchanged, and only introduce `Connection` struct during `parse_uri_location` so that we don't need to make `UriLocation` mutable.

But current logic LGTM too.

ariesdevil: emmm... `Connection` is part of `UriLocation`, split it out may be a little bit strange.

Xuanwo (approved):

PR #9446, opened 2023-01-03
fix: Lazy load JWKS

flaneur2020:

## Summary

I runs an integration tests environment which runs a jwks endpoint and databend-query, but in the environment, it's hard to keep the start up order between databend-query and the jwks endpoint.

When databend-query started eariler than the jwks endpoint, it'd panic on start up with the following error:

```
Databend Query start failure, cause: Code: 1051, displayText = Could not download JWKS: error sending request for url (https://jwks.xxx.svc:8080/.well-known/jwks.json): error trying to connect: dns error: no record found for Query { name: Name("jwks.xxx.svc."), query_type: AAAA, query_class: IN }.
```

This PR makes the following changes:

- loading JWKS in a lazy manner, makes the JWKS endpoint not a hard requirement for the databend-query's start up
- implement the interval polling JWKS logic in the todo
- removes the `async` in try_create function and the related callers

### DIFF
--- a/src/query/service/src/auth.rs
+++ b/src/query/service/src/auth.rs
@@ -41,7 +41,7 @@ pub enum Credential {
 }
 
 impl AuthMgr {
-    pub async fn create(cfg: &Config) -> Result<Arc<AuthMgr>> {
+    pub fn create(cfg: &Config) -> Result<Arc<AuthMgr>> {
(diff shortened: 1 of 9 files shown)


flaneur2020:
@Xuanwo plz help take a look?

Xuanwo (approved):
Nice change!

PR #9458, opened 2023-01-04
fix(jwt): restrict exp on JWT verification

flaneur2020:

## Summary

> The most crucial security claim is the "exp" claim. The issuer uses this claim to indicate the expiration date of a JWT. If this expiration date lies in the past, the JWT has expired and must not be used anymore. A typical example use case is an OpenID Connect identity token, which expires after a set period.

from: https://www.pingidentity.com/en/resources/blog/post/jwt-security-nobody-talks-about.html

jwt-simple library did not verify the `exp` claim unless set `max_validity`: https://github.com/jedisct1/rust-jwt-simple/blob/542ad46692ceb63a1c48a967d039ba76bc1db6c1/src/claims.rs#L184


Xuanwo (approved):

flaneur2020:
oh, jwt-simple does check exp in this line: https://github.com/jedisct1/rust-jwt-simple/blob/542ad46692ceb63a1c48a967d039ba76bc1db6c1/src/claims.rs#L196

this PR can be closed

PR #9486, opened 2023-01-06
feat(storage): support nested type in read_parquet.

### LINKED ISSUE #9417: Feature: Support reading snappy compressed parquet file
**read_parquet table function should support different compression options**

Usually parquet files are compressed using snappy or zstd. The read_parquet table function should support compression options to deal with popular compression algorithms.

RinChanNOWWW:

## Summary

- Don't push down filter to parquet reading if the field is a nested type. (arrow2 doesn't support constructing index page reader on nested type).
- Fix schema mismatch problem.
- Fix panic problem when reading complex nested types (such as Array(Tuple)).

Try to fix #9417.

## Note

- Get fields on complex nested types can not succeed. Will be fixed in later PR.

```sql
> desc ttt;
+-------+-----------------------+------+---------+-------+
| Field | Type                  | Null | Default | Extra |
+-------+-----------------------+------+---------+-------+
| t     | ARRAY((INT32, INT32)) | NO   | []      |       |
+-------+-----------------------+------+---------+-------+
1 row in set (0.050 sec)

> select t[0]:a from ttt;
ERROR 1105 (HY000): Code: 1065, displayText = no overload satisfies `get((Int32, Int32) NULL, String)`

has tried possible overloads:
  get(Array(Nothing) NULL, UInt64 NULL) :: NULL     : unable to unify `(Int32, Int32)` with `Array(Nothing)`
  get(Array(T0 NULL), UInt64) :: T0 NULL            : unable to unify `(Int32, Int32) NULL` with `Array(T0 NULL)`
  get(Array(T0 NULL) NULL, UInt64 NULL) :: T0 NULL  : unable to unify `(Int32, Int32)` with `Array(T0 NULL)`
  get(Array(T0), UInt64) :: T0 NULL                 : unable to u
```


src/query/expression/src/schema.rs
```diff
@@ -215,46 +215,6 @@ impl DataSchema {
         Self::new_from(fields, self.meta().clone())
     }
 
-    /// project with inner columns by path.
```
sundy-li, line 218: Is this code not used any more?

RinChanNOWWW: Only `TableSchema` implement these methods.

sundy-li (approved):

src/query/expression/src/schema.rs
sundy-li: Ok, so we should modify `leaf_fields` as above.

BohuTANG (approved):

src/query/expression/src/schema.rs
RinChanNOWWW: It seems we don't need to this for Array types. If we select array[i]:field, it will turn to call `get` method.

Issue #9483, opened 2023-01-06
performance: hits dataset performance degradation in new expression


BohuTANG:
**Summary**

The hits table is created and loaded data with the nightly-v0.8.177:

The benchmark with the same settings:
```
mysql> show settings like '%max%';
+------------------------------------+-------------+-------------+---------+-------------------------------------------------------------------------------------------------------------------+--------+
| name                               | value       | default     | level   | description                                                                                                       | type   |
+------------------------------------+-------------+-------------+---------+-------------------------------------------------------------------------------------------------------------------+--------+
| max_block_size                     | 65536       | 65536       | SESSION | Maximum block size for reading, default value: 65536.                                                             | UInt64 |
| max_execute_time                   | 0           | 0           | SESSION | The maximum query execution time. it means no limit if the value is zero. default value: 0.                       | UInt64 |
| max_inlist_to_or                   | 3           | 3           | SESSION | Max size in inlist expression that will convert to or combinator, default value: 3.                               | UInt64 |
| max_memory_usage                   | 26604358860 | 26604358860 | SESSION | The maximum memory usage for processing single query, in bytes. By default the value is determined automatically. | UInt64 |
| max_result_rows                    | 0           | 0           | SESSION | Auto limit max result rows if user not specify the limit, default is 0 means no limit                             | UInt64 |
| max_storage_io_requests            | 64          | 64          | SESSION | The maximum number of concurrent IO requests. By default the value is determined automatically.                   | UInt64 |
| max_threads                        | 32          | 32          | SESSION | The maximum number of threads to execute the request. By default the value is determined automatically.           | UInt64 |
| storage_io_max_page_bytes_for_read | 524288      | 524288      | SESSION | The maximum bytes of one IO request to read. Default the value is 512KB                                           | UInt64 |
+------------------------------------+-------------+-------------+---------+-------------------------------------------------------------------------------------------------------------------+--------+
```

Bench(3rd(Hot), chart generated by [google sheet](https://docs.google.com/spreadsheets/d/1ObfMrknUbq3tgPdVMjlffjVPlF_ARzMPJGtRj-turXs/edit?usp=sharing)):

Update(After #9500 merged)
![image](https://user-images.githubusercontent.com/172204/211149841-948dd7bc-62b1-47f5-a883-80c736308a7b.png)

bench script:
```
cat run-expr.sh
#!/bin/bash

TRIES=3
QUERY_NUM=1

[...]

PR #9488, opened 2023-01-06
chore(query): use smaller distinct state

sundy-li:

## Summary

Summary about this PR

Related #9483


sundy-li:
It's ok now.


```
 SELECT SearchPhrase, COUNT(DISTINCT UserID) AS u FROM (select * from hits)  WHERE SearchPhrase <> '' GROUP BY SearchPhrase  order by u desc  limit 10;
```

2.71 sec vs 2.68

RinChanNOWWW (approved):

BohuTANG:
Tested and the performance same as nightly-v0.8.177 👍

PR #9489, opened 2023-01-06
feat(meta/upgrade): meta data upgrade program

drmingdrmer:

## Summary
feat(meta-upgrade): meta data upgrade program

**What it does**:

This program load meta-service data from a `raft-dir` and upgrade TableMeta to new version and write them back.

Both in raft-log data and in state-machine data will be converted.

Usage:

- Shut down databend-meta
- Backup the data: https://databend.rs/doc/deploy/metasrv/metasrv-backup-restore
- Build it with `cargo build --bin databend-meta-upgrade-2023-01-15` and run it:

```text
databend-meta-upgrade-2023-01-15 --cmd upgrade --raft-dir "<./your/raft-dir/>"
```

## Related Issues


sundy-li:
can we merge the binary  into `metactl` and use args to dispatch ?

drmingdrmer:
> can we merge the binary into `metactl` and use args to dispatch ?

This program is not meant to be used after the upgrade is done. If, in the future, there is another data format upgrade there should be another standalone program. Putting all of these functions into one program is clumsy.

And this program depends on a former version of databend, which will increase building time of metactl.

Xuanwo:
> This program is not meant to be used after the upgrade is done.

If so, how about changing the name to `databend-upgrader-from-v0.8` or `databend-upgrader-v1`? We can add other upgrade logic if needed (for example, rebuild some table index).

PR #9501, opened 2023-01-07
refactor(query): use ctx to store the function evaluation error

sundy-li:

## Summary

Summary about this PR

1. use ctx to store the function evaluation error 
2. improve `to_string` from numbers using lexical_core

Closes #issue


src/query/functions/src/scalars/arithmetic.rs
```diff
@@ -147,79 +151,39 @@ pub fn register(registry: &mut FunctionRegistry) {
...
                                 "divide",
                                 FunctionProperty::default(),
                                 |_, _| FunctionDomain::MayThrow,
                                 vectorize_with_builder_2_arg::<NumberType<L>, NumberType<R>,  NumberType<T>>(
-                                    |a, b, output, _| {
+                                    |a, b, output, ctx| {
                                         let b = (b.as_() : T);
                                         if std::intrinsics::unlikely(b == 0.0) {
-                                            return Err("/ by zero".to_string());
+                                            ctx.set_error(output.len(), "/ by zero");
```
b41sh, line 162: If an error occurs, does the function need to continue to execute?

sundy-li: Yes, it just continues in a single batch which is not very costly.

src/query/expression/src/function.rs
```diff
@@ -56,13 +57,68 @@ impl Default for FunctionContext {
...
+    // validity from the nullable column
+    pub validity: Option<Bitmap>,
+    pub errors: Option<(MutableBitmap, String)>,
     pub tz: Tz,
 }
 
+impl<'a> EvalContext<'a> {
+    #[inline]
+    pub fn set_error(&mut self, row: usize, error_msg: impl AsRef<str>) {
+        // if the row is in valid, we don't need to set error
```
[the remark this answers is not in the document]
sundy-li: Ok, will be addressed in next pr.

src/query/functions/src/scalars/arithmetic.rs
```diff
@@ -393,25 +337,81 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                                            FORMAT,
+                                        >(
+                                            Native::from(*x), bytes, &options
+                                        )
+                                        .len();
+                                        offset += len;
+                                        builder.offsets.push(offset as u64);
+                                    }
+                                    values.set_len(offset);
+                                    values.shrink_to_fit();
```
andylokandy, line 369: This seems to be unnecessary.

src/query/functions/src/scalars/arithmetic.rs
```diff
@@ -393,25 +337,81 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                                            Native::from(*x),
+                                            bytes,
+                                            &options,
+                                        )
+                                        .len();
+                                    offset += len;
+                                    builder.offsets.push(offset as u64);
+                                }
+                                values.set_len(offset);
+                                values.shrink_to_fit();
```
andylokandy, line 406: ditto

src/query/codegen/src/writes/register.rs
```diff
@@ -736,9 +740,9 @@ pub fn codegen_register() {
 
                 format!(
                     "({arm_pat}) => {{
-                        let nullable_column = func({func_arg} ctx)?.into_column().unwrap();
+                        let nullable_column = func({func_arg} ctx).into_column().unwrap();
```
andylokandy, line 743: Do we need to set `ctx.validity` here?

sundy-li: if it's combine_nullable, we don't set validity to ctx. The inner function should care about the error and nulls.

andylokandy: `combine_nullable` takes `Fn(I1, I2) -> NullalbeType<O>` and then returns `Fn(Nullable<I1>, Nullable<I2>) -> NullableType<O>`, therefore it does the same as `passthrough_nullable` in handling argument's nulls, and the only difference is that `combine_nullable` is able to return a NULL while all arguments are not null.

b41sh (approved):

andylokandy (approved):

PR #9506, opened 2023-01-08
feat(meta): use expressin::TableSchema to replace obsolete datavalues::DataSchema

drmingdrmer:

##### feat(meta): use expressin::TableSchema to replace obsolete datavalues::DataSchema

This is a breaking change.

Meta data version becomes VER=24 in this commit:
- Executable without this patch can not read VER=24 metadata from
  meta-service.
- Executable with this patch can read VER=24 and all old version
  metadata.

So, when upgrading to this version, all old databend-query must be
shutdown before databend-query of this version are brought up.

- Fix: #9498

## Related Issues


src/meta/proto-conv/src/data_from_to_protobuf_impl.rs
```diff
@@ -149,6 +149,7 @@ impl FromToProto for dv::DataTypeImpl {
                     ver: VER,
                     min_reader_ver: MIN_READER_VER,
                     dt: Some(Dt::NullType(pb::Empty {})),
+                    dt24: None,
```
sundy-li, line 152: we should change the file name to help people know it's all about `Dt24Version` table meta.

sundy-li: Maybe we could freeze and remove the datavalues crate  then using the latest commit in cargo file starting from this pr.

drmingdrmer: > we should change the file name to help people know it's all about `Dt24Version` table meta.

What file name? Only part of this file changed.

sundy-li (approved):

PR #9511, opened 2023-01-08
chore(expression): be compatible with version v2 bloom filter index

sundy-li:

## Summary

Summary about this PR

No need to reload the data anymore.

Closes #issue


src/query/expression/src/converts/to.rs
```diff
@@ -37,3 +40,38 @@ pub fn to_schema(schema: &TableSchema) -> common_datavalues::DataSchema {
         .collect();
     common_datavalues::DataSchema::new_from(fields, schema.meta().clone())
 }
+
+pub fn scalar_to_datavalue(scalar: &Scalar) -> DataValue {
```
sundy-li, line 44: @drmingdrmer  this function is re add  because we use DataValue's hash to apply old bloom filter.

drmingdrmer (approved):

src/query/storages/index/src/bloom.rs
```diff
@@ -248,7 +254,15 @@ impl BlockFilter {
         };
 
         let (filter, _size) = Xor8Filter::from_bytes(filter_bytes)?;
-        if filter.contains(&target) {
+
+        let contains = if self.version == 2 {
```
BohuTANG, line 258: can we change the constant `2` to an enum value BlockBloomFilterIndexVersion::V2?

BohuTANG: This way, it will be easier for us to consider it when we refactor it in the future.

sundy-li: ok

drmingdrmer (approved):

src/query/storages/index/tests/it/filters/bloom_filter.rs
```diff
@@ -63,7 +63,7 @@ fn test_bloom_filter() -> Result<()> {
     ];
     let chunks_ref = chunks.iter().collect::<Vec<_>>();
 
-    let index = BlockFilter::try_create(FunctionContext::default(), schema, &chunks_ref)?;
+    let index = BlockFilter::try_create(FunctionContext::default(), schema, 3, &chunks_ref)?;
```
BohuTANG, line 66: 3 --> `V3BloomBlock::VERSION`?

BohuTANG: I guess this test is expected to test the latest version, not `3`?

sundy-li: Yes, 3 is the latest bloom version.

sundy-li: I'll update it, there are two struct named BlockFilter

BohuTANG: So, when we upgrade our version number next time, this test may be missed. Can we define a latest version or another way to archive it?

Xuanwo: I prefer to add a `LastestBlockBloomFilterIndexVersion = BlockBloomFilterIndexVersion::V3`?

BohuTANG:
@mergify update

src/query/storages/index/tests/it/filters/bloom_filter.rs
sundy-li: BlockBloomFilterIndexVersion::V3 is a enum, we need u64 here.

src/query/storages/table-meta/src/meta/versions.rs
```diff
@@ -163,9 +163,13 @@ mod converters {
         type Error = ErrorCode;
         fn try_from(value: u64) -> Result<Self, Self::Error> {
             match value {
-                1 | 2 => Err(ErrorCode::DeprecatedIndexFormat(
+                1 => Err(ErrorCode::DeprecatedIndexFormat(
```
BohuTANG, line 166: Looks better to define constant instead the int `1` `2` `3`.

sundy-li: this is already the lowest level now and it's the same logic as before.

We can define constant field name  `const VersionV1 = 1`, but it's too redundant

BohuTANG:
Test this PR, old bloom filter(V2) also works as new(V3):
```
mysql> SELECT UserID FROM hits_expr WHERE UserID = 435090932899640449; --new
+--------------------+
| userid             |
+--------------------+
| 435090932899640449 |
| 435090932899640449 |
| 435090932899640449 |
| 435090932899640449 |
+--------------------+
4 rows in set (0.10 sec)
Read 779848 rows, 5.95 MiB in 0.074 sec., 10.51 million rows/sec., 80.20 MiB/sec.

mysql> SELECT UserID FROM hits  WHERE UserID = 435090932899640449; -- old
+--------------------+
| userid             |
+--------------------+
| 435090932899640449 |
| 435090932899640449 |
| 435090932899640449 |
| 435090932899640449 |
+--------------------+
4 rows in set (0.13 sec)
Read 496176 rows, 3.79 MiB in 0.117 sec., 4.24 million rows/sec., 32.34 MiB/sec.
```

PR #9516, opened 2023-01-08
refactor(query): refactor map access to support view read tuple inner field

b41sh:

## Summary

* map access support read tuple field inside an array
* map access support view read tuple inner field
* add test for schema inner_project

Closes #issue


src/query/datavalues/tests/it/data_schema.rs
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common_datavalues::prelude::*;
+use common_exception::Result;
+use pretty_assertions::assert_eq;
+
+#[test]
+fn test_project_schema_from_struct() -> Result<()> {
+    let child_field11 = u64::to_data_type();
```
sundy-li, line 21: hmm, there is no to_data_type trait anymore.

BohuTANG:
@mergify update

src/query/datavalues/tests/it/data_schema.rs
b41sh: ok, use `DataTypeImpl::UInt64(UInt64Type)` instead.

tests/sqllogictests/suites/base/09_fuse_engine/09_0021_read_tuple_array
```diff
@@ -41,9 +41,18 @@ SELECT ta[3] FROM t
 NULL
 NULL
 
-## TODO: get from tuple fields
-## SELECT ta[0]:a FROM t
-## SELECT * FROM t where ta[0]:b > 3
+query I
+SELECT ta[1].a FROM t
```
RinChanNOWWW, line 45: Should the syntax be `tuple.field` or `tuple:field`?

b41sh: both are ok.

RinChanNOWWW:
Can add some tests of nullable tuple.

For example:

```sql
create table test (t tuple(a int, b int) null);
insert into test values (NULL), (NULL);
select * from test;
select t:a from test;
insert into test values((1,2)), (NULL), ((2,4));
select * from test;
select t:a from test;
select * from test where t is not null;
select * from test where t:a > 1;
````

These sqls seems not work in current Databend.

File: src/query/sql/src/planner/semantic/type_check.rs [hunk 3]
```diff
@@ -1845,108 +1788,201 @@ impl<'a> TypeChecker<'a> {
...
+            let box (path_value, path_data_type) = self.resolve_literal(&path_lit, None)?;
+            let path_scalar: Scalar = ConstantExpr {
+                value: path_value,
+                data_type: Box::new(path_data_type.clone()),
+            }
+            .into();
+            if let TableDataType::Array(inner_type) = table_data_type {
+                table_data_type = *inner_type;
+            }
+            table_data_type = TableDataType::wrap_nullable(&table_data_type);
```
RinChanNOWWW, line 1876: Why wrap_nullable？

b41sh: `get` function for `variant` and `array` will always return a nullable value, because the index may out of range.

RinChanNOWWW (approved):

PR #9531, opened 2023-01-09
feat(query): add a special dispatch for single simple order field.

RinChanNOWWW:

## Summary

If there is only one single simple order by field, we don't need to construct the arrow2 row converter.

This PR also abstracts `Rows` and `RowConverter` traits for more variations in the future.

Closes #issue


src/query/pipeline/transforms/src/processors/transforms/sort_utils/cursor.rs
```diff
@@ -0,0 +1,82 @@
+//  Copyright 2023 Datafuse Labs.
```
BohuTANG, line 1: Directory name `sort_utils` -> `sort` is better? `*_util` is not a good name.

sundy-li (approved):

PR #9532, opened 2023-01-09
feat: Add oss storage option about presign_endpoint_url

flaneur2020:

## Summary

Opendal added a parameter `presign_endpoint_url` which allows offering a seperated endpoint url for presign in OSS.

likewise we can set a internal endpoint url for network access like `internal.oss.aliyuncs.com` (which can only be accessed in aliyun's IDC) with a public endpoint url like `oss.aliyuncs.com` which allows users outside the aliyun's IDC to upload files into databend.

This pr does not affect the stage options (which is serialized in protobuf and persisted in metasrv), but only affects the config file. For users deploys databend in aliyun, they can take advantage of this option in their config file.


Xuanwo (approved):

flaneur2020:
@Xuanwo please take another look?

Xuanwo (approved):

BohuTANG (approved):

Issue #9480, opened 2023-01-06
Tracking: issues about new expression 


sundy-li:
**Summary**

After we merge #9411, it may have some issues to solve before we release the next version of 0.9.

- [x] Delete `functions`, `datablocks` crate and rename `functions-v2`. #9481
- [x] Support `alter table refresh index` to override the indices in new version, instead we have be compatible index in #9511 
- [x]  #9562
- [x] Improve bloom filter to use vectorized & typed hash function rather than hashing the `ScalarRef` enum #9542
- [x] `read_parquet` support read nested datatype. #9486
- [x] functions support set error which respect an option null mask bitmap  during the evaluation  #9501 
- [x] #9483
- [x] https://github.com/datafuselabs/databend/issues/9497
- [x] #9533

Meta-service:

- [x] #9535 
- [x] #9544 
- [x] #9555
- [x] #9506 
- [x] Meta provides a tool to migrate meta between versions. #9489 
- [x] #9498

drmingdrmer:
Update meta-service related task

sundy-li:
> Update meta-service related task: Save the new schema format TableSchema directly to meta-service, without converting back to obsolete DataSchema.

That's ok I think, we can backup the obsolete meta.

PR #9533, opened 2023-01-09
chore(expression): add try adaptor in function

sundy-li:

## Summary

Summary about this PR

Now all functions can have "try_" prefix as a function adaptor.

`try_xxx` means that if `xxx` function works failed it will transfer the result into null rather than throw the exception.

The drawback is the domain is inaccurate if   overflow cast happens because we are using runtime dispatch instead of a static register.

I don't know whether this is a good way to go, cc @andylokandy 

Related #9480


RinChanNOWWW (approved):

sundy-li:
continue in #9545

sundy-li:
@andylokandy  Do we need this try adaptor for other functions? If not I'll close this pr.

andylokandy:
>  Do we need this try adaptor for other functions? If not I'll close this pr.

We may not need it so far

PR #9535, opened 2023-01-09
feat(meta/proto-conv): add reader-min-msg-ver and msg-min-reader-ver

drmingdrmer:

##### feat(meta/proto-conv): add reader-min-msg-ver and msg-min-reader-ver

This is a compatible change.

There should be two **backward compatible** version defined:
One for the reader to define the min version of message it can read
and the other for the message to define the min version of the reader
that can read it.

A message is written with the min-reader-ver in it, and a reader checks
`reader.ver` against `msg.min-reader-ver` and `msg.ver` and
`reader.min-message-ver` to assert compatibility.

## Related Issues


ClSlaid (approved):
Clear explanation!

PR #9551, opened 2023-01-10
fix(query): fix memory leak for unsized hash table

zhang2014:

## Summary

fixes(query): fix memory leak for unsized hash table

Closes #issue


sundy-li (approved):

src/common/hashtable/src/unsized_hashtable.rs
```diff
@@ -1178,6 +1178,6 @@ where A: Allocator + Clone + Default
         self.table2.clear();
         self.table3.clear();
         self.table4.clear();
-        self.arena.reset();
+        drop(std::mem::take(&mut self.arena));
```
BohuTANG, line 1181: Adding some comments on why we `drop` here would be great.

PR #9559, opened 2023-01-10
fix: use context's runtime in clickhouse handler  while executing interpreter

### LINKED ISSUE #9558: Bug:  databend-query hangs when running sqllogic test parallelly
**Summary**

_on behalf of @sundy-li_ :

databend-query hangs when running sqllogic test parallelly : 

- startup databend-query
~~~
export META_EMBEDDED_DIR="./.databend/meta_embedded"
export STORAGE_TYPE="fs"
./target/debug/databend-query --internal-enable-sandbox-tenant
~~~

- run sql logic test

~~~
cargo run -p sqllogictests -- --handlers clickhouse --parallel 8 --no-fail-fast --run_file 09_0 --enable_sandbox —debug
~~~

then, `databend-query` process will highly likely hang.

- sql logic test case can not run to end
- mysql client can not connect to databend-query

dantengsky:

## Summary

use context's runtime in clickhouse handler while executing interpreter

Closes #9558

### DIFF
--- a/src/query/service/src/servers/http/clickhouse_handler.rs
+++ b/src/query/service/src/servers/http/clickhouse_handler.rs
@@ -112,7 +112,14 @@ async fn execute(
     handle: Option<JoinHandle<()>>,
 ) -> Result<WithContentType<Body>> {
     let format_typ = format.typ.clone();
-    let mut data_stream = interpreter.execute(ctx.clone()).await?;
+    let mut data_stream = ctx
+        .try_spawn({
+            let ctx = ctx.clone();
+            async move { interpreter.execute(ctx.clone()).await }
+        })?
+        .await
+        .map_err(ErrorCode::from_std_error)??;
+
     let table_schema = infer_table_schema(&schema)?;
     let mut output_format = FileFormatOptionsExt::get_output_format_from_clickhouse_format(
         format,
--- a/src/query/service/src/servers/http/clickhouse_handler.rs
+++ b/src/query/service/src/servers/http/clickhouse_handler.rs
@@ -112,13 +112,50 @@ async fn execute(
     handle: Option<JoinHandle<()>>,
 ) -> Result<WithContentType<Body>> {
     let format_typ = format.typ.clone();
+
+    // the reason of spawning new task to execute the interpreter:
+    // (sorry, I failed to describe it in a more concise way)
+    //
+    // - there are executions of interpreters that will block the caller (NOT async wait)
+    //   e.g. PipelineCompleteExecutor::execute, will spawn thread that executes the pipeline,
+    //   and then, join the thread handle.
+    // - async mutex (tokio::sync::Mutex) are used while executing the queries/statements
+    //   An async task may yield while holding the lock of an async mutex. e.g. embedded meta store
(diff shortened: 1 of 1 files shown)


zhang2014:
> @zhang2014 PTAL : by using `ctx.spawn` to execute the interrupter, query process no longer hang. but I am not sure if using ctx’s runtime here is correct.

Look good to me.

src/query/service/src/servers/http/clickhouse_handler.rs
```diff
@@ -112,7 +112,14 @@ async fn execute(
...
 ) -> Result<WithContentType<Body>> {
     let format_typ = format.typ.clone();
-    let mut data_stream = interpreter.execute(ctx.clone()).await?;
+    let mut data_stream = ctx
+        .try_spawn({
+            let ctx = ctx.clone();
+            async move { interpreter.execute(ctx.clone()).await }
+        })?
+        .await
+        .map_err(ErrorCode::from_std_error)??;
```
Xuanwo, line 121: `from_std_error` will lost error context, how about using `map_err(|err| anyhow!("clickhouse handler failed to join interpreter thread: {err:?}"))` for better error message?

src/query/service/src/servers/http/clickhouse_handler.rs
```diff
@@ -112,7 +112,14 @@ async fn execute(
     handle: Option<JoinHandle<()>>,
 ) -> Result<WithContentType<Body>> {
     let format_typ = format.typ.clone();
-    let mut data_stream = interpreter.execute(ctx.clone()).await?;
+    let mut data_stream = ctx
```
BohuTANG, line 115: It would be better to add some comments here to explain why we need run it in context runtime

src/query/service/src/servers/http/clickhouse_handler.rs
dantengsky: yeah, that will be better

src/query/service/src/servers/http/clickhouse_handler.rs
dantengsky: ok

src/query/service/src/servers/http/clickhouse_handler.rs
dantengsky: PTAL, the error message has been updated as you suggested

src/query/service/src/servers/http/clickhouse_handler.rs
dantengsky: PTAL, the reason for spawning is documented, but rather lengthy,  failed to put it in a  concise way :(

src/query/service/src/servers/http/clickhouse_handler.rs
```diff
@@ -112,13 +112,50 @@ async fn execute(
     handle: Option<JoinHandle<()>>,
 ) -> Result<WithContentType<Body>> {
     let format_typ = format.typ.clone();
+
+    // the reason of spawning new task to execute the interpreter:
```
Xuanwo, line 116: Nice!

Xuanwo (approved):

PR #9561, opened 2023-01-10
fix(expr): check_date() and to_string(boolean) may panic

andylokandy:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/src/query/expression/src/types/date.rs
+++ b/src/query/expression/src/types/date.rs
@@ -48,10 +48,7 @@ pub fn check_date(days: i64) -> Result<i32, String> {
     if (DATE_MIN as i64..=DATE_MAX as i64).contains(&days) {
         Ok(days as i32)
     } else {
-        Err(format!(
-            "date `{}` is out of range",
-            date_to_string(days, chrono_tz::Tz::UTC)
-        ))
+        Err("date is out of range".to_string())
     }
 }
 
--- a/src/query/expression/src/function.rs
+++ b/src/query/expression/src/function.rs
@@ -63,9 +63,9 @@ pub struct EvalContext<'a> {
     pub num_rows: usize,
     pub tz: Tz,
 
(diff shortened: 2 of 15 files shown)


src/query/expression/src/types/date.rs
```diff
@@ -48,10 +48,7 @@ pub fn check_date(days: i64) -> Result<i32, String> {
     if (DATE_MIN as i64..=DATE_MAX as i64).contains(&days) {
         Ok(days as i32)
     } else {
-        Err(format!(
-            "date `{}` is out of range",
-            date_to_string(days, chrono_tz::Tz::UTC)
-        ))
+        Err("date is out of range".to_string())
```
sundy-li, line 51: why does this solve the panic of `check_date`  ?

andylokandy: because date_to_string assumed that the date is valid thus will not overflow on addition.

BohuTANG:
Some tests need to be fixed:
```
[SQL] explain select t.number from numbers(1) as t, numbers(1) as t1 where t.number > 1 and 1 < t1.number
[Diff] (-excepted|+actual)
    HashJoin
    ├── join type: CROSS
    ├── build keys: []
    ├── probe keys: []
    ├── filters: []
    ├── estimated rows: 0.11
    ├── Filter(Build)
    │   ├── filters: [lt(1_u8, t1.number (#1))]
    │   ├── estimated rows: 0.33
    │   └── TableScan
    │       ├── table: default.system.numbers
    │       ├── read rows: 1
    │       ├── read bytes: 8
    │       ├── partitions total: 1
    │       ├── partitions scanned: 1
-   │       ├── push downs: [filters: [lt(1_u64, number)], limit: NONE]
+   │       ├── push downs: [filters: [lt(1, number)], limit: NONE]
    │       └── estimated rows: 1.00
    └── Filter(Probe)
```

sundy-li (approved):

PR #9562, opened 2023-01-10
chore(fuse): make column meta to be an enum

sundy-li:

## Summary

Summary about this PR

1. The `ColumnMeta` is an enum now which can hold different kinds of storage's ColumnMeta.
2. Bump `strawboat` and `arrow2` version.

Notice: This change is compatible with the latest tag but not compatible with the latest commit.

Related #9480


dantengsky (approved):
LGTM

PR #9563, opened 2023-01-10
feat(storage): read_parquet page index. 

RinChanNOWWW:

## Summary

- Filter (Prune) pages by column page stats, and get a row selection.
- Add unit tests of pruning row groups and pages.
- Fix bug: page reader should use `nums_row` but not `length`.

### Others

- When creating `RangeFilter`, pass `FunctionContext` to it. (no need to pass the whole `TableContext`).
- Fold prewhere expression like #9600.

Closes #issue


BohuTANG:
@mergify update

sundy-li (approved):

PR #9572, opened 2023-01-11
fix(expr): constant folder should run repeatly until stable

andylokandy:

## Summary

Summary about this PR

Closes #issue


src/query/functions/tests/it/scalars/testdata/boolean.txt
```diff
@@ -337,6 +337,27 @@ evaluation (internal):
 +--------+--------------------------------------------------------------------------+
 
 
+ast            : const_false AND CAST('1000' AS STRING) = 1000
+raw expr       : and(#0::Boolean, eq(CAST("1000" AS String), 1000_u16))
+checked expr   : and<Boolean, Boolean>(#0, eq<Int64, Int64>(CAST("1000" AS Int64), CAST(1000_u16 AS Int64)))
+optimized expr : false
```
andylokandy, line 343: @sundy-li PTAL if this satisfies.

src/query/expression/src/evaluator.rs
```diff
@@ -518,7 +518,23 @@ impl<'a, Index: ColumnIndex> ConstantFolder<'a, Index> {
         }
     }
 
+    /// Fold a single expression, returning the new expression and the domain of the new expression.
     pub fn fold(&self, expr: &Expr<Index>) -> (Expr<Index>, Option<Domain>) {
+        let mut old_expr = expr.clone();
+        loop {
+            let (new_expr, domain) = self.fold_once(&old_expr);
```
sundy-li, line 525: maybe we should add a max count and log the error to avoid bugs of hang.

sundy-li: we can use `1024 * 1024` which common expression depth will not be reached.

src/query/functions/tests/it/scalars/testdata/boolean.txt
sundy-li: Great!

sundy-li (approved):

PR #9583, opened 2023-01-12
feat(query):  add information_schema for sharding-jdbc

### LINKED ISSUE #9582: Feature: Make databend compatible with `Sharding-JDBC`
**Summary**

`sharding-jdbc` needs to load meta data from `information_schema`, For now `databend` lacks table `information_schema.statistics` `information_schema.key_column_usage` and a few columns in `information_schema.columns`.

**Exception**
```
Error creating bean with name 'shardingSphereDataSource' defined in class path resource [org/apache/shardingsphere/spring/boot/ShardingSphereAutoConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [javax.sql.DataSource]: Factory method 'shardingSphereDataSource' threw exception; nested exception is java.sql.SQLException: (conn=64) Code: 1025, displayText = `information_schema.statistics` table is unknown.
```
image

**Propose**
We should consider adding these tables to compatible with `sharding-jdbc`, even if the table is empty.

**Refer**
https://github.com/apache/shardingsphere/blob/3588165dd0240de7c53756413bcd3a7d272297bd/infra/common/src/main/java/org/apache/shardingsphere/infra/metadata/database/schema/loader/dialect/MySQLSchemaMetaDataLoader.java#L54-L66

yufan022:

## Summary

Closes #9582


TCeason (approved):
LGTM, Please make ci pass.

src/query/storages/information-schema/src/key_column_usage_table.rs
```diff
@@ -0,0 +1,60 @@
...
+use common_meta_app::schema::TableIdent;
+use common_meta_app::schema::TableInfo;
+use common_meta_app::schema::TableMeta;
+use common_storages_view::view_table::ViewTable;
+use common_storages_view::view_table::QUERY;
+
+pub struct KeyColumnUsageTable {}
+
+impl KeyColumnUsageTable {
+    pub fn create(table_id: u64) -> Arc<dyn Table> {
```
Xuanwo, line 28: How about adding a sqllogic test to cover this which can make sure we never break your use cases unintendedly?

yufan022: Added test to `01_0003_information_schema` file, or should I add it a separate file for sharding-jdbc?

TCeason: I think that's ok. But if you want to add test better to add in there https://github.com/datafuselabs/databend/tree/main/tests/suites/1_stateful. Add a new fold.

BohuTANG (approved):

PR #9585, opened 2023-01-12
refactor(expr): rearrange common_expression and common_function

andylokandy:

## Summary

- Improve module structure of common_expression
- Improve ConstantFolder: folding with full domain no longer requiring input_domains
- Improve `try_inet_aton` and `try_inet_ntoa` by using `error_to_null`
- Move `rand` and `uuid` to `other.rs`


Xuanwo (approved):

src/query/expression/src/evaluator.rs
```diff
@@ -507,20 +507,50 @@ pub struct ConstantFolder<'a, Index: ColumnIndex> {
...
+            func_ctx,
+            fn_registry,
+        };
+
+        folder.fold_to_stable(expr)
+    }
+
+    /// Fold a single expression with columns' domain, and then return the new expression and the
+    /// domain of the new expression.
+    pub fn fold_with_domain(
```
sundy-li, line 536: I also thought about this. Maybe we can use `Option` here and unwrap to full domain.

andylokandy: Yes but I think  the `None` may seem confusing in `ConstantFolder::fold(&expr, None, func_ctx, &BUILTIN_FUNCTIONS)`

sundy-li (approved):

PR #9588, opened 2023-01-13
feat(geo): add other base geo functions

ariesdevil:

## Summary

Add other base geo functions:
* [x] point_in_polygon
* [x] geohash_encode
* [x] geohash_decode

Closes #issue


src/query/functions/src/scalars/geo.rs
```diff
@@ -133,6 +145,47 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    );
+
+    registry
+        .register_passthrough_nullable_1_arg::<StringType, KvPair<Float64Type, Float64Type>, _, _>(
+            "geohash_decode",
+            FunctionProperty::default(),
+            |_| FunctionDomain::Full,
+            vectorize_with_builder_1_arg::<StringType, KvPair<Float64Type, Float64Type>>(
+                |encoded, builder, ctx| {
+                    let s = std::str::from_utf8(encoded).unwrap();
```
sundy-li, line 176: unwrap is not safe

sundy-li: can use and_then to chain the errors

BohuTANG:
It would be better to add a sql logic test.

sundy-li (approved):

PR #9593, opened 2023-01-13
fix(query): format response schema fields

everpcpc:

## Summary

change schema in query response from `Option<DataSchemaRef>` to  `Vec<QueryResponseField>`

for: https://github.com/databendcloud/bendsql/issues/52


sundy-li (approved):

sundy-li:
Notice that:

nullable type in wrapped_display will be like:  `Nullable(Array(UInt32) )`.

PR #9600, opened 2023-01-14
feat(fuse): add prewhere support in native storage format

sundy-li:

## Summary

Summary about this PR

1. add prewhere support in native storage format
2. add clustering page index in native storage format
3. add page filter, this improves hits Q37 50% and other queries have no bad effects.
4. bump opendal & strawboat

```
-- before
MySQL [(none)]> select  *  from hits_native where URL like '%google%' ignore_result;
Empty set (17.149 sec)


-- after
MySQL [(none)]> select  *  from hits_native where URL like '%google%' ignore_result;
Empty set (10.402 sec)
```

Closes #issue


BohuTANG:
Will the `prewhere` also works for the fuse engine with parquet format?

sundy-li:
> Will the `prewhere` also works for the fuse engine with parquet format?

No, just work on the native format.

Xuanwo:
> -- before
> MySQL [(none)]> select  *  from hits_native where URL like '%google%' ignore_result;
> Empty set (17.149 sec)
> 
> 
> -- after
> MySQL [(none)]> select  *  from hits_native where URL like '%google%' ignore_result;
> Empty set (10.402 sec)

This bench is on local fs?

Xuanwo:
@sundy-li, opendal has released v0.25, can you bump to tagged version instead? Thanks.

sundy-li:
> This bench is on local fs?

Yes，I'm focus on improving native format now.

sundy-li:
> > This bench is on local fs?
> 
> Yes，I'm focus on improving native format now.  But I think it can also take effect on s3.

src/query/service/src/pipelines/pipeline_builder.rs
```diff
@@ -345,7 +345,10 @@ impl PipelineBuilder {
             &aggregate.agg_funcs,
         )?;
 
-        if self.ctx.get_cluster().is_empty()
+        // this has bugs now, so we disable it for now, cc @winter
```
BohuTANG, line 348: File a new open issue for it?

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
```diff
@@ -40,31 +52,103 @@ pub struct NativeDeserializeDataTransform {
...
+    prewhere_filter: Arc<Option<Expr>>,
+
+    prewhere_skipped: usize,
 }
 
 impl NativeDeserializeDataTransform {
     pub fn create(
         ctx: Arc<dyn TableContext>,
         block_reader: Arc<BlockReader>,
+        plan: &DataSourcePlan,
```
BohuTANG, line 70: Can we remove the `plan` to use `pushdwon`?
We have tried to decouple the planner with storage.

sundy-li: we also need the plan.source_info in `NativeDeserializeDataTransform`

src/query/service/src/pipelines/pipeline_builder.rs
sundy-li: https://github.com/datafuselabs/databend/issues/9673

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
BohuTANG: `plan.source_info.schema()` --> `schema`

BohuTANG: We have done much work to decouple `planners` for many crates in the past.

BohuTANG: Oh, i see.
This is common_catalog, not planners.

BohuTANG (approved):

PR #9608, opened 2023-01-15
chore(query): make opendal be workspace dep

sundy-li:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,6 +84,8 @@ members = [
 # databend maintains:
 openraft = { git = "https://github.com/drmingdrmer/openraft", tag = "v0.7.4-alpha.3" }
 sled = { git = "https://github.com/datafuse-extras/sled", tag = "v0.34.7-datafuse.1", default-features = false }
+opendal = "0.24"
+ordered-float = { version = "3.4.0", default-features = false }
 
 # error
 anyhow = { version = "1.0.65" }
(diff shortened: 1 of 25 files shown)


Cargo.toml
```diff
@@ -84,6 +84,8 @@ members = [
 # databend maintains:
 openraft = { git = "https://github.com/drmingdrmer/openraft", tag = "v0.7.4-alpha.3" }
 sled = { git = "https://github.com/datafuse-extras/sled", tag = "v0.34.7-datafuse.1", default-features = false }
+opendal = "0.24"
+ordered-float = { version = "3.4.0", default-features = false }
```
PsiACE, line 88: We might consider adding features here as well, unless it is otherwise restricted.

sundy-li: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

Seems the features could be added in children crates.

sundy-li: [features](https://doc.rust-lang.org/cargo/reference/features.html): These are additive with the features declared in the [workspace.dependencies]

Xuanwo (approved):

PsiACE (approved):

Cargo.toml
PsiACE: You're right, my idea was to tell everyone a basic set of features that are used, but given that each crate has its own needs, it seems more logical to specify them in the crate.

PR #9612, opened 2023-01-15
chore(query): add DistinctToGroupBy

sundy-li:

## Summary

Summary about this PR

```
-- before
MySQL [(none)]> select count_distinct(UserID) from hits_native;
+------------------------+
| count_distinct(userid) |
+------------------------+
|               17630976 |
+------------------------+
1 row in set (2.108 sec)


-- after
MySQL [(none)]> select count_distinct(UserID) from hits_native;
+----------+
| count()  |
+----------+
| 17630976 |
+----------+
1 row in set (0.656 sec)
```

Closes #issue


tests/sqllogictests/suites/mode/standalone/explain/explain.test
```diff
@@ -768,6 +768,38 @@ HashJoin
     ├── push downs: [filters: [], limit: NONE]
     └── estimated rows: 5.00
 
+query T
+explain select count_distinct(a) from t1;
```
BohuTANG, line 772: How about adding a new case for `explain select count(distinct a) from t1`?

BohuTANG (approved):

PR #9614, opened 2023-01-15
chore(query): add bloom filter supported data type

TCeason:

## Summary

Bloom filter can support more base data type:
Data
Timestamp

Closes #issue

### DIFF
--- a/src/query/storages/common/index/src/filters/xor8.rs
+++ b/src/query/storages/common/index/src/filters/xor8.rs
@@ -128,9 +128,11 @@ impl Filter for Xor8Filter {
 
 impl SupportedType for Xor8Filter {
     fn is_supported_type(data_type: &DataType) -> bool {
-        // Bloom index only enabled for String and Integral types for now
         let inner_type = data_type.remove_nullable();
-        matches!(inner_type, DataType::Number(_) | DataType::String)
+        matches!(
+            inner_type,
+            DataType::Number(_) | DataType::String | DataType::Timestamp | DataType::Date
+        )
     }
 }
 


src/query/storages/common/index/src/filters/xor8.rs
```diff
@@ -128,9 +128,16 @@ impl Filter for Xor8Filter {
 
 impl SupportedType for Xor8Filter {
     fn is_supported_type(data_type: &DataType) -> bool {
-        // Bloom index only enabled for String and Integral types for now
         let inner_type = data_type.remove_nullable();
-        matches!(inner_type, DataType::Number(_) | DataType::String)
+        matches!(
+            inner_type,
+            DataType::Number(_)
+                | DataType::String
```
BohuTANG, line 135: We used to support these types, but there were some performance issues(I can't find the PR ), so we removed them? cc @dantengsky

src/query/storages/common/index/src/filters/xor8.rs
```diff
@@ -128,9 +128,16 @@ impl Filter for Xor8Filter {
...
+                | DataType::Timestamp
+                | DataType::Date
+                | DataType::Boolean
```
b41sh, line 138: `Boolean` is better to use `MinMax` index

src/query/storages/common/index/src/filters/xor8.rs
```diff
@@ -128,9 +128,16 @@ impl Filter for Xor8Filter {
...
+                | DataType::Variant
```
b41sh, line 139: `Variant` usually filters on internal fields, so storing an index doesn't make much sense.
 The binary `JSONB` values are relatively large, and calculating the index may cause performance problems.

src/query/storages/common/index/src/filters/xor8.rs
sundy-li: why do we have performance issue about `timestamp,date` ?

BohuTANG: Confirmed, it's my mistake. We only support `string` data type, please ignore my comments.

BohuTANG:
@mergify update

src/query/storages/common/index/src/filters/xor8.rs
dantengsky: can not recall the details :(,  but it is not about performance (no benchmark of different data types has been done)

b41sh (approved):

PR #9619, opened 2023-01-15
fix(storage): fix range filter read stat with index

### LINKED ISSUE #9607: bug: query a column after Array type column fail or panic
### Search before asking

### Version

current version

### What's Wrong?

bug: query a column after Array type column fail or panic

### How to Reproduce?

create a table that has a double type column after array column, select by the double column return empty:

```
mysql> CREATE TABLE `05_0027_at_t0`(a int, b Tuple(Bool, Int64, Float64, String, Date, Timestamp),c Array(UInt8),d double); Query OK, 0 rows affected (1.41 sec)

mysql> INSERT INTO TABLE `05_0027_at_t0` values(1,(true, 100, 12.34, 'abc', '2020-01-01', '2020-01-01 00:00:00'),[101,2],4.0);
Query OK, 1 row affected (0.62 sec)

mysql> select * from `05_0027_at_t0` where d = 4.0;
Empty set (1.81 sec)
Read 0 rows, 0.00 B in 0.432 sec., 0 rows/sec., 0.00 B/sec.
```

create a table that has a int type column after array column, select by the int column make task panic:

```
mysql> CREATE TABLE `05_0027_at_t0`(a int, b Tuple(Bool, Int64, Float64, String, Date, Timestamp),c Array(UInt8),d int);
Query OK, 0 rows affected (1.40 sec)

mysql> INSERT INTO TABLE `05_0027_at_t0` values(1,(true, 100, 12.34, 'abc', '2020-01-01', '2020-01-01 00:00:00'),[101,2],4);
Query OK, 1 row affected (0.57 sec)

mysql> select * from `05_0027_at_t0` where d=4;
ERROR 1105 (HY000): Code: 4000, displayText = block pruning failure, task 1106 panicked(while in query pipeline init).
```

b41sh:

## Summary

- fix range filter read statistics with index
- fix generate column statistics with wrong leaf index when having a tuple type inside the array type
- fix unit test `test_project_schema_from_tuple`

Closes #9607


src/query/storages/common/index/src/range_filter.rs
```diff
@@ -62,10 +67,35 @@ impl RangeFilter {
 
         let (new_expr, _) = ConstantFolder::fold(&conjunction, func_ctx, &BUILTIN_FUNCTIONS);
 
+        let mut fields = VecDeque::new();
+        for field in schema.fields().iter() {
```
andylokandy, line 71:
```suggestion
        for field in schema.fields() {
```

src/query/storages/common/index/src/range_filter.rs
```diff
@@ -62,10 +67,35 @@ impl RangeFilter {
...
+        let mut leaf_index = 0;
+        let mut column_indices: HashMap<String, usize> = HashMap::new();
+        while let Some(field) = fields.pop_front() {
+            if let TableDataType::Tuple {
+                fields_name,
+                fields_type,
+            } = field.data_type()
+            {
+                for (inner_field_name, inner_field_type) in
+                    fields_name.iter().rev().zip(fields_type.iter().rev())
```
andylokandy, line 83:
```suggestion
                    fields_name.iter().zip(fields_type.iter()).rev()
```

sundy-li (approved):

Issue #9442, opened 2023-01-03
use column_id to read and write data instead of index.


lichuang:

sundy-li:
you can do it based on the expression branch, it's new TableSchema now.

PR #9623, opened 2023-01-16
feat: add column id in TableSchema; use column id instead of index when read and write data

lichuang:

## Summary

1. add `next_column_id` in `TableSchema` and `column_id` in `TableField` to save column id info.
2. add `leaf_column_ids` in `ColumnLeaf` to get column id of leaf.
3. use column_id to read and write data instead of index.
4. unit test.

Closes https://github.com/datafuselabs/databend/issues/9442


src/meta/protos/proto/metadata.proto
```diff
@@ -28,6 +28,11 @@ message DataSchema {
 
   // Other optional table info
   map<string, string> metadata = 2;
+
+  uint32 max_column_id = 3;
```
drmingdrmer, line 32: What about naming it `next_column_id`?

`max_column_id` looks like the max value of a valid column id

src/meta/proto-conv/src/util.rs
```diff
@@ -81,6 +81,10 @@ const META_CHANGE_LOG: &[(u64, &str)] = &[
         "2023-01-07: Add: new-schema pb::DataType to/from TableDataType",
     ),
     (25, "2023-01-05: Add: user.proto/OnErrorMode::AbortNum"),
+    (
+        26,
+        "2023-01-16: Add: metadata.proto/DataSchema::max_column_id",
+    ),
```
drmingdrmer, line 87: You'll have to add a test case to ensure the new version works:

https://github.com/datafuselabs/databend/blob/78d127257519bea475ad1361ee0359da764a443b/src/meta/proto-conv/tests/it/v024_table_meta.rs

src/meta/protos/proto/metadata.proto
```diff
@@ -28,6 +28,11 @@ message DataSchema {
...
+
+  // Map column name to column id
+  map<string, uint32> column_id_map = 4;
```
drmingdrmer, line 35: How to get a field instance by column-id?

As `self.fields` is a vec of `field`, is column-id the index of a field in `self.fields`?

BohuTANG:
Conflicting files
src/meta/proto-conv/src/util.rs
src/query/expression/tests/it/data_schema.rs

src/meta/protos/proto/metadata.proto
lichuang: first get field use index, than get column id using field.name in column_id_map.

src/query/expression/src/schema.rs
```diff
@@ -75,6 +76,11 @@ pub struct DataField {
 pub struct TableSchema {
     pub(crate) fields: Vec<TableField>,
     pub(crate) metadata: BTreeMap<String, String>,
+
+    // define new fields as Option for compatibility
+    pub next_column_id: Option<u32>,
+    column_id_map: Option<BTreeMap<String, u32>>,
+    column_id_set: Option<HashSet<u32>>,
```
sundy-li, line 83: what's `column_id_set` ?

Is this set can be generated from  `fields` and `column_id_map` ?

lichuang: yes, `column_id_set` can be generated from fields and column_id_map ?

sundy-li: So I suggest to have a method to get the id set,  the `column_id_set` should not be stored into meta because it may cause inconsistent if set is different to map.

lichuang: Ok, tag `column_id_set` as `skip_serializing`, every time `TableSchema` init, it will be recalculated.

sundy-li: Can we have a uniq id in `TableField` ? I think keeping a `Hashmap` in dataschema is too heavy.

Vec will outperform Hashmap in less than hundreds of elements.
And `TableField` can represent the real meaning of the mapping between `name` and `id`  

```
#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
pub struct TableField {
    name: String,
    default_expr: Option<String>,
	id: ColumnID,
    data_type: TableDataType,
}

```

src/query/expression/src/schema.rs
```diff
@@ -241,18 +250,306 @@ impl TableSchema {
...
+            _ => {
+                *next_column_id += 1;
+            }
         }
     }
 
+    fn build_members_from_fields(
+        fields: &[TableField],
+        column_id_map: Option<BTreeMap<String, u32>>,
+        next_column_id_opt: Option<u32>,
```
drmingdrmer, line 321: This argument can be replaced with a `u32`. When calling this method, just pass `next_column_id_opt.unwrap_or_default()` to it.

src/query/expression/src/schema.rs
```diff
@@ -241,18 +250,306 @@ impl TableSchema {
...
+            Some(column_id_map) => {
+                has_column_id_map_inited = !column_id_map.is_empty();
+                column_id_map
+            }
+            None => BTreeMap::new(),
+        };
+        let has_next_column_id_inited = match next_column_id_opt {
+            Some(next_column_id) => next_column_id > 0,
+            None => false,
+        };
```
drmingdrmer, line 335: `next_column_id_opt.unwrap_or_default()` would be quite enough. You do not really need to care about if it is initialized, IMHO.

src/query/expression/src/schema.rs
```diff
@@ -241,18 +250,306 @@ impl TableSchema {
...
+        // make sure that column_id_map and next_column_id init at the same time
+        assert_eq!(has_column_id_map_inited, has_next_column_id_inited);
```
drmingdrmer, line 337: If an empty `column_id_map` means uninitialized `column_id_map`, there is a chance column id map is empty and next_column_id is initialized, i.e., all fields are removed.

src/meta/proto-conv/tests/it/main.rs
```diff
@@ -32,3 +32,4 @@ mod v012_table_meta;
 mod v023_table_meta;
 mod v024_table_meta;
 mod v025_user_stage;
+mod v26_schema;
```
drmingdrmer, line 35: Keep versions in alphabetic order:

```suggestion
mod v026_schema;
```

src/query/expression/src/schema.rs
lichuang: `BlockMeta.col_metas` has to save what column saved in this block, a field-unique id can not work, only schema-unique id can work.

src/query/expression/src/schema.rs
lichuang: So i check they init at the same time here, or else will panic.

sundy-li: It's not safe to do this panic, we should ensure it be always consistent.

lichuang: Now All `TableSchema` init function will call `build_members_from_fields`,  in this function do assert is my way to do this check,   and `TableSchema` can be be init in other way without these init functions, any better advise?

src/query/expression/src/schema.rs
drmingdrmer: > a field-unique id can not work, only schema-unique id can work.

It is not field-unique, it is still schema scope unique. The same way to build field ids a s it does with a `column_id_map`.

Every piece of information you can extract from `column_id_map` can also be extracted from Fields, does it?

With a `column_id_map`, you'll get a Field instance from `fields: Vec<Field>`, by column-name, and get field id from `column_id_map`, by column-name, too.
Without a `column_id_map`, you can just get a Field instance with field-id directly from `fields: Vec<Field>`, right?

b41sh: Do we need to support modifying the columns inside the `Tuple`? If not, it is possible to just give the top-level `TableField` a `ColumnId` and the inner columns can be found via the `ColumnId` plus a path vector. This would simplify the complexity of the design.

src/query/expression/src/schema.rs
```diff
@@ -1085,3 +1370,53 @@ pub fn infer_table_schema(data_schema: &DataSchemaRef) -> Result<TableSchemaRef>
     }
     Ok(TableSchemaRefExt::create(fields))
 }
+
+// a complex schema to cover all data types tests.
+pub fn create_test_complex_schema() -> TableSchema {
```
b41sh, line 1375: This function is just used for testing, better to move to `src/query/expression/tests/it/schema.rs`

lichuang: ColumnLeaf also need this function to generate test data.

src/query/expression/src/schema.rs
drmingdrmer: > Every piece of information you can extract from `column_id_map` can also be extracted from Fields, does it?
> 
> With a `column_id_map`, you'll get a Field instance from `fields: Vec<Field>`, by column-name, and get field id from `column_id_map`, by column-name, too. Without a `column_id_map`, you can just get a Field instance with field-id directly from `fields: Vec<Field>`, right?

I see. It is a recursive structure of TableDataType, not a recursive TableField, we can not add a field-id into TableDataType.

Then I do not have any question about this.

sundy-li: If the inner column need to be modified, it's better to create another  `Tuple Type`, so all the nested column should have a new ColumnID.

E.g.   Add a Tuple field `Tuple { a, b , c }`, this will assign 3 ids  ----`n, n+1, n+2`, we just need to store the first ID  in 
`DataFiled {  Tuple,  ColumnID:   n }`

drmingdrmer: > If the inner column need to be modified, it's better to create another Tuple Type, so all the nested column should have a new ColumnID.

Is it acceptable at storage level? Assumes that field `t Tuple {a,b}` has column id 1,2; If a user removes tuple field `t:a`, it's going to have to assign a new column id: `t Tuple{b}` with column-id 3, and update the column-id for every parquet file?

drmingdrmer: > If the inner column need to be modified, it's better to create another Tuple Type, so all the nested column should have a new ColumnID.

Is it acceptable at storage level? Assumes that field `t Tuple {a,b}` has column id 1,2; If a user removes tuple field `t:a`, it's going to have to assign a new column id: `t Tuple{b}` with column-id 3, and update the column-id for every parquet file?

RinChanNOWWW: I think only allow creating a brand new nested type is simpler.

That means, `t Tuple {a,b}` and `t Tuple{b}` are two different fields. Users can not query the old `t Tuple {a, b}` after changing it to `t Tuple{b}`.

src/common/storage/src/column_leaf.rs
```diff
@@ -47,30 +53,67 @@ impl ColumnLeaves {
...
+                    child_leaf_column_ids,
+                )
             }
             ArrowType::List(inner_field)
             | ArrowType::LargeList(inner_field)
             | ArrowType::FixedSizeList(inner_field, _) => {
                 let mut child_column_leaves = Vec::with_capacity(1);
                 let mut child_leaf_ids = Vec::with_capacity(1);
-                let child_column_leaf = Self::traverse_fields_dfs(inner_field, leaf_id);
+                let inner_field_name = &format!("{}:0", parent_name);
```
RinChanNOWWW, line 91: Is this needed?

src/common/storage/src/column_leaf.rs
```diff
@@ -103,16 +177,32 @@ pub struct ColumnLeaf {
...
         Self {
             field,
             leaf_ids,
             children,
+            leaf_column_ids,
+        }
+    }
+
+    // Handle the case that leaf_column_ids is empty
+    pub fn leaf_column_id(&self, i: usize) -> u32 {
```
RinChanNOWWW, line 201: Is this for compatibility?

src/common/storage/tests/it/column_leaf.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright 2022 Datafuse Labs.
```
RinChanNOWWW, line 1: `ColumnLeaf` is renamed to `ColumnNode` in #9660. So the file name should be renamed later.

src/query/expression/src/schema.rs
RinChanNOWWW: How about adding a label `#[cfg(test)]`.

src/query/expression/src/schema.rs
sundy-li: > Is it acceptable at storage level? Assumes that field `t Tuple {a,b}` has column id 1,2; If a user removes tuple field `t:a`, it's going to have to assign a new column id: `t Tuple{b}` with column-id 3, and update the column-id for every parquet file?

we are not going to update the removed column in parquet file. If Tuple `t` is removed, column id 1,2 are not used anymore, any parquet file with these two ids is not used to read this `t`.

src/query/storages/parquet/src/parquet_part.rs
```diff
@@ -23,6 +23,7 @@ use common_arrow::parquet::compression::Compression;
 use common_arrow::parquet::indexes::Interval;
 use common_catalog::plan::PartInfo;
 use common_catalog::plan::PartInfoPtr;
+use common_catalog::table::ColumnId;
```
RinChanNOWWW, line 26: Not modifying the codes in crate `common-storages-paruqet` is ok. The schema of `read_parquet` will not be changed.

src/query/expression/tests/it/schema.rs
```diff
@@ -0,0 +1,214 @@
...
+        ("d:0", 6),
+        ("e", 7),
+        ("e:0", 7),
+        ("e:0:0", 7),
+        ("f", 8),
+        ("g", 9),
+        ("h", 10),
+        ("h:a", 10),
+        ("h:b", 11),
+    ];
```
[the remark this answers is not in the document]
lichuang: `build_from_data_type` is a recurse function, if change the behavior as remove `non-leaf column` name, in this case, `b:0` and `b` will be removed, remove `b:0`  is not a big deal, but `b` is the top level column, remove `b` will cause panic.

Also, `TableSchema::to_column_ids` generate ids MUST equal to ColumnNode::leaf_ids, if not the compatibility will be break, see `test_column_leaf_schema_from_struct_of_old_version` test for compatibility.

src/query/expression/src/schema.rs
lichuang: if label `#[cfg(test)]`, cannot compile.

src/meta/protos/proto/metadata.proto
```diff
@@ -28,6 +35,11 @@ message DataSchema {
 
   // Other optional table info
   map<string, string> metadata = 2;
+
+  uint32 next_column_id = 3;
+
+  // Map column name to column id
+  repeated ColumnIdVector field_column_ids = 4;
```
drmingdrmer, line 42: The comment is out of date? This is not a map but a 2-dimension list

src/meta/protos/proto/metadata.proto
```diff
@@ -18,6 +18,13 @@ package databend_proto;
 
 import "datatype.proto";
 
+message ColumnIdVector {
+  uint64 ver = 100;
+  uint64 min_reader_ver = 101;
+
+  repeated uint32 column_ids = 1;
+}
+
```
drmingdrmer, line 27: If `ColumnIdVector` won't be decoded as a standalone message, `ver` and `min_reader_ver` can be removed.

src/meta/proto-conv/tests/it/proto_conv.rs
```diff
@@ -333,5 +336,29 @@ fn test_build_pb_buf() -> anyhow::Result<()> {
...
+            TableField::new("c", TableDataType::Number(NumberDataType::UInt64)),
+        ];
+        let schema = TableSchema::new(fields);
+
+        let p = schema.to_pb()?;
+
+        let mut buf = vec![];
+        common_protos::prost::Message::encode(&p, &mut buf)?;
+        println!("schema:{:?}", buf);
+    }
```
drmingdrmer, line 361: This test is a duplicate of 
[v026_schema.rs](https://github.com/datafuselabs/databend/pull/9623/files#diff-ee4eeb4f2627c5f0629a9c4b0a07f4d5f231b5d9da8be3008fed9c400db9e9ad) and can be removed.

drmingdrmer (commented):
I carefully reviewed the meta/proto part and some related calls, and thoroughly examined the other.

It looks good.

sundy-li (approved):

BohuTANG:
Can we mark all the conversion as solved?  @lichuang

BohuTANG:
Let's merge this PR tomorrow and disabled the released to check it works well enough. 
cc @lichuang @sundy-li

Issue #9580, opened 2023-01-12
doc bug: streaming_load can't work 


wubx:
**Summary**
https://databend.rs/doc/load-data/local

Reference above doc, write:
```
curl -XPUT 'http://wubx:wubxwubx@192.168.1.100:8008/v1/streaming_load' -H 'insert_sql: insert into wubx.tmp_202212 file_format = (type = "CSV" skip_header = 1  field_delimiter = "," record_delimiter = "\n")' -F 'upload=@"./InstanceDetail_1953761221157405_InstanceDetail_202212.csv"'
```

<img width="1312" alt="image" src="https://user-images.githubusercontent.com/320680/212034437-455762ab-b67b-49d5-afdc-5a1f0b0c8870.png">

remove  record_delimiter = "\n" 
<img width="1320" alt="image" src="https://user-images.githubusercontent.com/320680/212034667-74284007-ed56-4122-aa7f-c626a7fc8f03.png">

It's work. 

Let's try: Put the double quotation marks out, use single quotation marks around the parameters
```
curl -XPUT "http://wubx:wubxwubx@192.168.1.100:8008/v1/streaming_load" -H "insert_sql: insert into wubx.tmp_202212 file_format = (type = 'CSV' skip_header = 1  field_delimiter = ',' record_delimiter = '\n')" -F "upload=@./InstanceDetail_1953761221157405_InstanceDetail_202212.csv"
```

![image](https://user-images.githubusercontent.com/320680/212034917-111203fb-26b0-4e7e-b6c5-cbfcad6dc95b.png)

It's work too~~~

todo：
1. Modify the manual to make streaming_load work  @soyeric128 

Question:
**Shall we support the first case so that users can use it more conveniently?**

PR #9626, opened 2023-01-16
fix(format): should not allow ident as value of options.

youngsofun:

## Summary

when running ` copy into t1 from 'fs:///xxx' file_format = (type = 'csv' record_delimiter = "\n")`

currently we get error like 

```
record_delimiter with two chars can only be '\r\n'.
```

but it work if we use single quotes `record_delimiter = ‘\n’`.

the reason is we now use Dialect::PostgreSQL (and not allowed to set it),  ` "\n"` is a  ident instead of a literal string.
and ident is not unescaped.

after this pr, we get a clearer error message.

```
 copy into t1 from 'fs:///xxx' file_format = (type = 'csv' record_delimiter = "\n")
                                                                             ^^^^ expected <QuotedString>, <LiteralInteger>, <MySQLLiteralHex>, or <PGLiteralHex>

```

Closes https://github.com/datafuselabs/databend/issues/9580

### DIFF
--- a/src/query/ast/src/parser/stage.rs
+++ b/src/query/ast/src/parser/stage.rs
@@ -36,7 +36,7 @@ pub fn u64_to_string(i: Input) -> IResult<String> {
 
 pub fn parameter_to_string(i: Input) -> IResult<String> {
     map(
-        rule! { ( #literal_string | #ident_to_string | #u64_to_string ) },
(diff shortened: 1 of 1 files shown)


youngsofun:
cc @wubx

youngsofun:
not work because we need: 
```
(type=csv compression=AUTO record_delimiter=NONE escape="\\") 
```

PR #9629, opened 2023-01-16
fix: ensure there is page index while using page index to filter pages.

RinChanNOWWW:

## Summary

Arrow2 will not throw an error but panic when reading page indexes if there is no page index in the parquet file.

We should check if there is page index first.

### DIFF
--- a/src/query/storages/parquet/src/pruning.rs
+++ b/src/query/storages/parquet/src/pruning.rs
@@ -130,7 +130,11 @@ pub fn prune_and_set_partitions(
                 continue;
             }
 
-            let row_selection = if read_options.prune_pages() {
+            let row_selection = if read_options.prune_pages()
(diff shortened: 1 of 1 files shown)


xudong963 (approved):

xudong963:
> ```
> Starting databend-sqllogic tests
> MySQL client starts to run...
> ```
> 
> Can we print the arguments(like parallel thread numbers) of `databend-sqllogic`? cc @xudong963

It's a fixed value in ci, **8**
```
target/${BUILD_PROFILE}/databend-sqllogictests --handlers ${TEST_HANDLERS} ${RUN_DIR} --enable_sandbox --parallel 8 --debug
```

I doubt http handler will hang sometimes like clickhouse handler

ybcwnXOW1q

PR #9630, opened 2023-01-16
feat(query): support named params for table functions.

RinChanNOWWW:

## Summary

~~- When calling `read_parquet`, cache `FileMetaData`.~~
- Support named parameters for table functions. The syntax is like:

```sql
select * from tbl_func(
    param1, param2, 
    op1=>literal1, op2=literal2, op3=>literal3
); 
```


BohuTANG:
@RinChanNOWWW 
I suggest not adding the cache for parquet file meta in this PR, @dantengsky is working on the new cache, we can do it after the refactoring.

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -931,6 +932,7 @@ impl<'a> TypeChecker<'a> {
                 args[1] = ConstantExpr {
                     value: common_expression::Literal::Int64(1),
                     data_type: Box::new(DataType::Number(NumberDataType::Int64)),
+                    key: None,
```
Xuanwo, line 935: Adding a `key` in `ConstantExpr` seems strange. How about adding a new expr type? like `AssignmentExpr`?

src/query/ast/src/ast/expr.rs
```diff
@@ -318,6 +318,8 @@ pub enum BinaryOperator {
     BitwiseOr,
     BitwiseAnd,
     BitwiseXor,
+    // `=>` operator,
+    FatRArrow,
```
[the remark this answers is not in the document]
andylokandy: It's better to add the named option syntax to `Expr::FunctionCall`'s `args` instead of `BinaryOperator`. The field `key` add to `Scalar::Constant` and the constraint of `left of => must be an identifier and the right must be literal` both reveal that the approach may  not be optimal.

src/query/sql/src/planner/semantic/type_check.rs
RinChanNOWWW: I have tried adding a new expr, but it introduced many modifications. I think make it as a `ConstantExpr` would make things simpler.

src/query/ast/src/ast/expr.rs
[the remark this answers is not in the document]
RinChanNOWWW: We just extract the name of the identifier and convert it into a constant string scalar.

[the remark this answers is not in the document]
andylokandy: The problem is that the current ad-hoc design may add unnescessary complexity to the ast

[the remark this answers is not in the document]
andylokandy: The optimal solution may be like:

```diff
  pub enum Expr<'a> {
      ....
      FunctionCall {
          span: &'a [Token<'a>],
          /// Set to true if the function is aggregate function with `DISTINCT`, like `COUNT(DISTINCT a)`
          distinct: bool,
          name: Identifier<'a>,
          args: Vec<Expr<'a>>,
+         named_args: HashMap<String, Expr<'a>>,
          params: Vec<Literal>,
      },
      ....
  } 
```

Xuanwo (approved):

src/query/ast/src/parser/query.rs
```diff
@@ -231,6 +231,28 @@ pub fn table_reference(i: Input) -> IResult<TableReference> {
...
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum TableFunctionParam<'a> {
+    // func(name => arg)
+    Named { name: String, value: Expr<'a> },
+    // func(arg)
+    Normal(Expr<'a>),
+}
+
+pub fn table_function_args(i: Input) -> IResult<TableFunctionParam> {
```
andylokandy, line 242:
```suggestion
pub fn table_function_param(i: Input) -> IResult<TableFunctionParam> {
```

src/query/ast/src/parser/query.rs
```diff
@@ -407,12 +429,29 @@ impl<'a, I: Iterator<Item = WithSpan<'a, TableReferenceElement<'a>>>> PrattParse
...
+                        _ => None,
+                    })
+                    .collect();
+                let named_params = params
+                    .into_iter()
+                    .filter_map(|p| match p {
+                        TableFunctionParam::Named { name, value } => Some((name, value)),
+                        _ => None,
+                    })
+                    .collect();
```
andylokandy, line 446: Should we check that all named_params are behind normal params by `params.iter().array_windows::<2>().all(...)`

RinChanNOWWW: This is unnecessary.

andylokandy (approved):

PR #9640, opened 2023-01-17
refactor(query): analyze table use bloom filter calc ndv

TCeason:

## Summary

Now analyze table will calc all blocks in segments.

we can use bloom filter index calc ndv.

Closes #issue

### DIFF
--- a/src/query/storages/fuse/src/operations/analyze.rs
+++ b/src/query/storages/fuse/src/operations/analyze.rs
@@ -19,10 +19,17 @@ use common_catalog::table::Table;
 use common_catalog::table_context::TableContext;
 use common_exception::ErrorCode;
 use common_exception::Result;
+use common_expression::Value;
+use storages_common_index::filters::Filter;
+use storages_common_index::filters::Xor8Filter;
+use storages_common_index::BloomIndex;
+use storages_common_table_meta::meta::ColumnId;
+use storages_common_table_meta::meta::Location;
 use storages_common_table_meta::meta::TableSnapshot;
 use storages_common_table_meta::meta::TableSnapshotStatistics;
 use tracing::warn;
 
+use crate::io::BlockFilterReader;
 use crate::io::SegmentsIO;
 use crate::FuseTable;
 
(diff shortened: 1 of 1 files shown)


TCeason:
Has some err. Will debug it.

src/query/storages/fuse/src/operations/analyze.rs
```diff
@@ -49,34 +54,110 @@ impl FuseTable {
...
+                                        filter.filter_schema,
+                                        loc.1,
+                                        &[&filter.filter_block],
+                                    )? {
+                                        block_count_sum += 1;
+                                        row_count_sum += row_count;
+                                        for (col_id, distinct_cnt) in
+                                            bloom_index.column_distinct_count.iter()
+                                        {
+                                            let density = *distinct_cnt as f64 / row_count as f64;
```
sundy-li, line 119: We merge use XorFilter to merge the hashes now.

src/query/storages/fuse/src/operations/analyze.rs
```diff
@@ -49,34 +54,110 @@ impl FuseTable {
...
+                                    ctx.clone(),
+                                    self.operator.clone(),
+                                    &filter_block_cols,
+                                    index_len,
+                                )
+                                .await;
+                            match maybe_filter {
+                                Ok(filter) => {
+                                    let func_ctx = ctx.try_get_function_context()?;
+                                    if let Some(bloom_index) = BlockFilter::try_create(
```
b41sh, line 108: should use `BlockFilter::from_filter_block` here, `BlockFilter::try_create` will recalculate the bloom filter.

src/query/storages/fuse/src/io/read/bloom_index_reader.rs
```diff
@@ -275,9 +276,10 @@ mod util_v1 {
                 Ok((bytes, idx))
             }
         } else {
-            Err(ErrorCode::Internal(format!(
-                "failed to find bloom index column. no such column {col_name}"
-            )))
+            warn!(
+                "failed to find bloom index column. no such column {col_name}, return empty"
+            );
+            Ok((Arc::new(vec![0u8]), 0))
```
TCeason, line 282: This modify is useless. I think we can catch read_filter err and return Ok(). That's can cover the table has data filed but not has bloom filter index.

After this pr, just optimize table or insert a value and then re-run analyze table, the fuse_static will correct.

sundy-li: why not be `vec![]` ?

src/query/storages/fuse/src/operations/analyze.rs
```diff
@@ -46,46 +54,75 @@ impl FuseTable {
...
+                                    let block = filter.filters;
+                                    for (i, filter) in block.iter().enumerate() {
+                                        if let Some(len) = filter.len() {
+                                            let idx = source_schema
+                                                .index_of(source_schema.field(i).name().as_str())
+                                                .unwrap()
+                                                as ColumnId;
+                                            match column_distinct_count.get_mut(&idx) {
+                                                Some(val) => {
+                                                    *val += len as u64;
```
sundy-li, line 95: Need to Merge NDV, not just sum it.

PR #9657, opened 2023-01-18
feat(planner): Add switch to enable bushy join

leiysky:

## Summary

Add a switch `enable_bushy_join` to enable bushy join plan generation.

Since bushy join isn't stable for now, we can not estimate the cost of a bushy join plan well. So the `enable_bushy_join` is set to `0` by default, and won't be exposed to the users in documentation until it's stable.


xudong963 (approved):

leiysky:
@mergify requeue

PR #9661, opened 2023-01-18
fix(query):  fix max_result_rows only limit output results nums

sandflee:

## Summary

https://github.com/datafuselabs/databend/pull/9406  add `max_result_rows`, but we may get wrong sql result if there are subqueries.  let's only limit the output results rows.


sandflee:
@mergify update

sundy-li (approved):

PR #9667, opened 2023-01-18
fix(parser): disable the identifier quote `` for mysql dialect

andylokandy:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/src/query/ast/src/input.rs
+++ b/src/query/ast/src/input.rs
@@ -91,8 +91,7 @@ impl Dialect {
         match self {
             Dialect::MySQL => c == '`',
             Dialect::Hive => c == '`',
-            // TODO: remove '`' quote support once mysql handler correctly set mysql dialect.
-            Dialect::PostgreSQL => c == '"' || c == '`',
+            Dialect::PostgreSQL => c == '"',
         }
     }
 


BohuTANG:
```
MySQL client starts to run...
Running MySQL test for file: tests/sqllogictests/suites/base/05_ddl/05_0000_ddl_create_tables ...
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: MysqlClient(Server(ServerError { code: 1105, message: "Code: 1005, displayText = error: \n  --> SQL:1:5\n  |\n1 | USE `default`;\n  | --- ^^^^^^^^^ expected <Ident>\n  | |    \n  | while parsing `USE <database>`\n\n.", state: "HY000" }))', tests/sqllogictests/src/main.rs:234:69
stack backtrace:
```

src/query/ast/src/input.rs
```diff
@@ -91,8 +91,7 @@ impl Dialect {
         match self {
             Dialect::MySQL => c == '`',
             Dialect::Hive => c == '`',
-            // TODO: remove '`' quote support once mysql handler correctly set mysql dialect.
-            Dialect::PostgreSQL => c == '"' || c == '`',
+            Dialect::PostgreSQL => c == '"',
```
sundy-li, line 94: The title is mysql dialect yet here you change pg dialect.

andylokandy: Yes, it's because the MySQL handler did not properly set Dialect to MySQL but still used the ` quote. So we have to allow the default `Dialect:: PostgreSQL` to support the MySQL style quote.

The fact is that the dialect is still not properly set to `Dialect::MySQL` for mysql handler; the dialect is handled by session setting which is still `Dialect:: PostgreSQL`.

So I may close it for now.

PR #9671, opened 2023-01-18
fix(tyck): fix missing simple_cast

andylokandy:

## Summary

Summary about this PR

Closes #issue


RinChanNOWWW:
Here are some unit tests:

```rust
fn test_cast_to_nested_type(file: &mut impl Write, is_try: bool) {
    let prefix = if is_try { "TRY_" } else { "" };

    run_ast(
        file,
        format!("{prefix}CAST(((1, TRUE), 1) AS Tuple(Tuple(INT, INT), INT))"),
        &[],
    );
    run_ast(
        file,
        format!("{prefix}CAST(((1, 'a'), 1) AS Tuple(Tuple(INT, INT NULL), INT))"),
        &[],
    );
    run_ast(
        file,
        format!("{prefix}CAST(((1, 'a'), 1) AS Tuple(Tuple(INT, INT), INT) NULL)"),
        &[],
    );
    run_ast(
        file,
        format!("{prefix}CAST([(1,TRUE),(2,FALSE)] AS Array(Tuple(INT, INT)))"),
        &[],
    );
    run_ast(
        file,
        format!("{prefix}CAST([(1,'a'),(2,'a')] AS Array(Tuple(INT, INT)) NULL)"),
        &[],
    );
    run_ast(
        file,
        format!("{prefix}CAST([(1,'a'),(2,'a')] AS Array(Tuple(INT, INT NULL)))"),
        &[],
    );
    run_ast(
        file,
        format!("{prefix}CAST([[TRUE], [FALSE, TRUE]] AS Array(Array(INT)))"),
        &[],
    );
    run_ast(
        file,
        format!("{prefix}CAST([['a'], ['b', 'c']] AS Array(Array(INT) NULL))"),
        &[],
    );
}

```

src/query/functions/tests/it/scalars/testdata/cast.txt
```diff
@@ -3137,3 +3204,75 @@ evaluation (internal):
...
+output         : ((1_i32, NULL), 1_i32)
+
+
+ast            : TRY_CAST(((1, 'a'), 1) AS Tuple(Tuple(INT, INT), INT) NULL)
+raw expr       : TRY_CAST(tuple(tuple(1_u8, "a"), 1_u8) AS ((Int32, Int32), Int32) NULL)
+checked expr   : TRY_CAST(tuple<(UInt8, String), UInt8>(tuple<UInt8, String>(1_u8, "a"), 1_u8) AS ((Int32 NULL, Int32 NULL) NULL, Int32 NULL) NULL)
+optimized expr : ((1_i32, NULL), 1_i32)
+output type    : ((Int32 NULL, Int32 NULL) NULL, Int32 NULL) NULL
+output domain  : (({1..=1}, {NULL}), {1..=1})
+output         : ((1_i32, NULL), 1_i32)
```
RinChanNOWWW, line 3231: If this should output (NULL) ?

src/query/functions/tests/it/scalars/testdata/cast.txt
```diff
@@ -3137,3 +3204,75 @@ evaluation (internal):
...
+output         : [(1_i32, 1_i32), (2_i32, 0_i32)]
+
+
+ast            : TRY_CAST([(1,'a'),(2,'a')] AS Array(Tuple(INT, INT)) NULL)
+raw expr       : TRY_CAST(array(tuple(1_u8, "a"), tuple(2_u8, "a")) AS Array((Int32, Int32)) NULL)
+checked expr   : TRY_CAST(array<T0=(UInt8, String)><T0, T0>(tuple<UInt8, String>(1_u8, "a"), tuple<UInt8, String>(2_u8, "a")) AS Array((Int32 NULL, Int32 NULL) NULL) NULL)
+optimized expr : [(1_i32, NULL), (2_i32, NULL)]
+output type    : Array((Int32 NULL, Int32 NULL) NULL) NULL
+output domain  : [({1..=2}, {0..=0} ∪ {NULL})]
+output         : [(1_i32, NULL), (2_i32, NULL)]
```
RinChanNOWWW, line 3249: If this should output [NULL, NULL]?

src/query/functions/tests/it/scalars/testdata/cast.txt
```diff
@@ -3137,3 +3204,75 @@ evaluation (internal):
...
+output         : [[1_i32], [0_i32, 1_i32]]
+
+
+ast            : TRY_CAST([['a'], ['b', 'c']] AS Array(Array(INT) NULL))
+raw expr       : TRY_CAST(array(array("a"), array("b", "c")) AS Array(Array(Int32) NULL))
+checked expr   : TRY_CAST(array<T0=Array(String)><T0, T0>(array<T0=String><T0>("a"), array<T0=String><T0, T0>("b", "c")) AS Array(Array(Int32 NULL) NULL) NULL)
+optimized expr : [[NULL], [NULL, NULL]]
+output type    : Array(Array(Int32 NULL) NULL) NULL
+output domain  : [[{0..=0} ∪ {NULL}]]
+output         : [[NULL], [NULL, NULL]]
```
RinChanNOWWW, line 3276: If this should output [NULL, NULL]?

src/query/functions/tests/it/scalars/testdata/cast.txt
andylokandy: @sundy-li PTAL

sundy-li: I think the output type should be `((Int32 NULL, Int32 NULL) NULL, Int32 NULL)`, we should forbidden the use `Tuple` or `Array` inside in `Nullable`.

sundy-li: If the original type is not matched, so all the nested types should be null.

andylokandy: `TRY_CAST(1 AS TUPLE(INT, INT))` has to output `Null`, therefore the output type of `TRY_CATS(xxx AS TUPLE(INT, INT))` is `TUPLE(Int32 NULL, Int32 NULL) NULL`, which has Tuple inside nullable.

RinChanNOWWW: Nullable nested types are unavoidable in runtime.

sundy-li: Ok, so we should allow this in runtime evaluation.

sundy-li: > Nullable nested types are unavoidable in runtime.

It's forbidden in DDL but can be available in runtime.

andylokandy: There are two types of errors CAST could throw:

1. trying to cast to unsupported type.
2. The cast overflowed.

So there is another solution to this problem: Allowing `TRY_CAST` to throw errors on the first error type.

sundy-li: >  Allowing TRY_CAST to throw errors on the first error type.
That could be better.

sundy-li (approved):

RinChanNOWWW (approved):

PR #9672, opened 2023-01-18
refactor: cache layer

dantengsky:

## Summary

- `storages-common-cache`
   - expose trait `CacheAccessor`, which provides the essential cache operation `put`, `get`, `evict`
   - define trait `StroageCache`, which describes the minimum interface that all the cache providers/backends should implement
     note that it is a "sync style" trait. for the `put` and `evict` methods, they are likely/should to be sync, not sure about `get` yet.
   - implement `StorageCache` for `LruCache`, and defines `LruCache` instances(type level) we currently used
      and also a dummy implementation of `StorageCache` for `DiskCache`, to test the API
   - provides commonly used in-memory cache types, instantiated from `LruCache` parameterized by the type of cached item, size meter, etc.
   - provides a generic `CachedReader` 
     given an implementation of `StorageCache`, e.g. `LruCache` or `DiskCache`, and a proper implementation of `Loader`, it can be instantiated to a concrete CachedReader.  
     
- `storages-common-table-meta::cache`
  - `meta_cache_manager.rs` init all the in-memory cache objects that are currently needed
  - `meta_cache.rs` 
    - declares the type alias of in-memory cache types that currently used to cache table meta
    - provides the `CachedMeta` trait and impls, which 
      - binds the type of meta to the type of cache 
         e.g. `SegmentInfo::cache()` should return the `SegmentInfoCache`
      - requires all the `Cache` returned should impl `CacheAccessor`, with proper type parameters.
      - also the instance of `Cache` grabbed from `CacheManager` is  bound to the meta type at compile time    
-  `common-storages-fuse::io`
   - introduces `ColumnDataLoader`, which implements `Loader` to load range of data from object storage, with the ability to customize cache keys
   - introduces `io::read::BloomIndexColumnReader` which basically instantiated `CachedReader` with  `ColumnDataLoader` and `LruCahe`, caches index column data in memory
   - introduces `CachedMetaWriter` trait, which writes  data and populated the item caches
     implemented for all the type that implements `CachedMeta`

- other related refactor
  in fuse table operation modules, replaces the implicit usage of `CacheManger` to populate/evict caches, with the usage of  `CachedMeta` trait. 
  the way of manipulating the cache is kept as it is, e.g. snapshot will not be cached during the writing, but after committed successfully. 

Closes #issue


src/query/storages/fuse/src/io/read/bloom_index/util.rs
```diff
@@ -0,0 +1,284 @@
+// Copyright 2022 Datafuse Labs.
```
BohuTANG, line 1: Name `util.rs` --> `load.rs`?

BohuTANG:
Cool, will review it today later.

src/query/storages/common/cache/src/lib.rs
```diff
@@ -13,16 +13,99 @@
...
+pub use providers::InMemoryItemCacheHolder;
+pub use read::CacheKey;
+pub use read::DiskCacheReader;
+pub use read::InMemoryBytesCacheReader;
+pub use read::InMemoryItemCacheReader;
+pub use read::LoadParams;
+pub use read::Loader;
+pub use read::LoaderWithCacheKey;
+
+pub trait CacheAccessor<K, V> {
```
BohuTANG, line 36: Prefer to move `CacheAccessor` from `lib.rs` to `cache.rs`, keep the lib.rs clean.

dantengsky: :ok_hand:

src/query/storages/fuse/src/operations/gc.rs
```diff
@@ -393,10 +409,28 @@ impl FuseTable {
...
+        ctx: Arc<dyn TableContext>,
+        locations_to_be_purged: HashSet<String>,
+    ) -> Result<()>
+    where
+        T: CachedMeta<T>,
+    {
+        if let Some(cache) = T::cache() {
+            for loc in locations_to_be_purged.iter() {
+                cache.evict(loc);
+            }
```
BohuTANG, line 428: Can you move the eviction to `try_purge_location_files`?

BohuTANG: Else all LGTM.

dantengsky: looks like we should keep the cache eviction logic in the method
    `async fn try_purge_location_files_and_cache<T: CachedMeta<T>>(...) -> Result<()>`
    
`try_purge_location_files` is for 
- remove files only
- called by removals that (currently) caches are not involved
   e.g. removing the data blocks

src/query/storages/fuse/src/operations/gc.rs
```diff
@@ -393,10 +409,28 @@ impl FuseTable {
     ) -> Result<()> {
         let fuse_file = Files::create(ctx.clone(), self.operator.clone());
         let locations = Vec::from_iter(locations_to_be_purged);
-        self.clean_cache(&locations);
+        // self.clean_cache(&locations);
```
BohuTANG, line 412: This line code no need anymore?

BohuTANG (approved):

src/query/storages/fuse/src/operations/gc.rs
BohuTANG: Will remove it in another PR

PR #9674, opened 2023-01-19
feat: Use global http client for sharing dns cache and connection

Xuanwo:

## Summary

feat: Use global http client for sharing dns cache and connection


Xuanwo:
Not good in benchmark, closing.

PR #9681, opened 2023-01-19
chore(planner): unify format of `EXPLAIN FRAGMENTS`

### LINKED ISSUE #9679: feat: unified EXPLAIN and EXPLAIN FRAGMENTS formats
**Summary**

If we have cluster A with 3 nodes and add another 3 query nodes to cluster A, the query performance is not improved, it's very hard to check where the problem is.

So if we add more stats(like cost estimate for each fragment) to `explain fragment`, it's easy to address where the problem is, and we should choose which exchange method and the join-order.

cc @xudong963 @leiysky

leiysky:

## Summary

Close #9679

### DIFF
--- a/src/query/service/src/schedulers/fragments/query_fragment_actions_display.rs
+++ b/src/query/service/src/schedulers/fragments/query_fragment_actions_display.rs
@@ -84,7 +84,6 @@ impl<'a> Display for QueryFragmentActionsWrap<'a> {
                         .format(self.metadata.clone())
                         .and_then(|node| node.format_pretty_with_prefix("    "))
                         .unwrap();
-                    // .map_err(|_| std::fmt::Error)?;
                     write!(f, "{}", plan_display_string)?;
                 }
             };
--- a/src/query/service/src/schedulers/fragments/query_fragment_actions_display.rs
+++ b/src/query/service/src/schedulers/fragments/query_fragment_actions_display.rs
@@ -15,19 +15,25 @@
 use std::fmt::Display;
 use std::fmt::Formatter;
 
+use common_sql::MetadataRef;
+
 use crate::api::DataExchange;
 use crate::api::FragmentPayload;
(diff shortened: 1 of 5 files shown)


src/query/service/src/schedulers/fragments/query_fragment_actions_display.rs
```diff
@@ -70,7 +80,12 @@ impl<'a> Display for QueryFragmentActionsWrap<'a> {
             let fragment_action = &self.inner.fragment_actions[0];
             match &fragment_action.payload {
                 FragmentPayload::Plan(node) => {
-                    write!(f, "{}", node.format_indent(1))?;
+                    let plan_display_string = node
+                        .format(self.metadata.clone())
+                        .and_then(|node| node.format_pretty_with_prefix("    "))
+                        .unwrap();
+                    // .map_err(|_| std::fmt::Error)?;
```
BohuTANG, line 87: This line no need anymore?

xudong963 (approved):

tests/sqllogictests/suites/mode/cluster/exchange.test
```diff
@@ -177,3 +177,74 @@ Exchange
...
+    ├── destination fragment: [3]
+    └── TableScan
+        ├── table: default.system.numbers
+        ├── read rows: 2
+        ├── read bytes: 16
+        ├── partitions total: 1
+        ├── partitions scanned: 1
+        ├── push downs: [filters: [], limit: NONE]
+        └── estimated rows: 2.00
+(empty)
```
BohuTANG, line 196: What's `(empty)` mean? Can we ignore it when print.

leiysky: It represents an empty string, the line breaks. This can help distinguish between fragments.

BohuTANG (approved):

PR #9684, opened 2023-01-19
feat(storages): `read_parquet` uses OpenDAL for all IO operations.

RinChanNOWWW:

## Summary

- replace `std::File` with `opendal` for all IO operations. (TODO(Another PR): use `opendal` to do "glob").
- optimize prewhere logic. (do `resort` before filter).
- fix scan progress collection.
- TODO(Another PR): support remote location and async operations.

Closes #issue

### DIFF
--- a/src/query/storages/parquet/src/parquet_source.rs
+++ b/src/query/storages/parquet/src/parquet_source.rs
@@ -28,7 +28,6 @@ use common_expression::filter_helper::FilterHelpers;
 use common_expression::types::BooleanType;
 use common_expression::DataBlock;
 use common_expression::DataSchemaRef;
-use common_expression::DataSchemaRefExt;
 use common_expression::Evaluator;
 use common_expression::Expr;
 use common_expression::Value;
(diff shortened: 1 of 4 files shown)


Xuanwo:
only fs and hdfs implemented blocking api. Will we need to test s3 via blocking api too?

RinChanNOWWW:
> only fs and hdfs implemented blocking api. Will we need to test s3 via blocking api too?

I plan to check if the backend supports blocking api and choose which api to use.

But I don't have figured out how to achieve this yet
(reading file meta data is only in sync functions now, maybe we need to use blocking api for these operations?) .

RinChanNOWWW:
Have no complete idea about the remote location. Let's implement it later.

BohuTANG:
> > only fs and hdfs implemented blocking api. Will we need to test s3 via blocking api too?
> 
> I plan to check if the backend supports blocking api and choose which api to use.
> 
> But I don't have figured out how to achieve this yet (reading file meta data is only in sync functions now, maybe we need to use blocking api for these operations?) .

Like this?
https://github.com/datafuselabs/databend/blob/923f3fab8491411a81222917c17a55675f3aa860/src/query/storages/fuse/src/fuse_table.rs#L180

RinChanNOWWW:
Yes. Or we block the file meta data reading?

Xuanwo:
> Let's implement it later.

LGTM, we can discussion the remote file support later.

Xuanwo (approved):

sundy-li (approved):

PR #9694, opened 2023-01-20
fix(query): fix query hang in two level aggregator

zhang2014:

## Summary

fix(query): fix query hang in two level aggregator(hits dataset query 10)

Closes #issue


BohuTANG:
Does this query(q11) hang related to this PR?
```
SELECT MobilePhoneModel, COUNT(DISTINCT UserID) AS u FROM hits WHERE MobilePhoneModel <> '' GROUP BY MobilePhoneModel ORDER BY u DESC LIMIT 10;
```

sundy-li:
> Does this query(q11) hang related to this PR?
 

Yes

sundy-li (approved):

PR #9703, opened 2023-01-22
chore(query): refactor hashtable

sundy-li:

## Summary

Summary about this PR

1. add `key_size` in unsize hashtable, this can give the capacity of StringColumnBuilder
2. reduce useless iteration of hashtable

Closes #issue


src/common/hashtable/src/stack_hashtable.rs
```diff
@@ -193,13 +188,8 @@ where
                 _alignment: [0; 0],
             }));
         }
-        while (self.table.len() + other.table.len()) * 2 > self.table.capacity() {
```
BohuTANG, line 196: No need `while` to loop?

sundy-li: It's duplicated. The inner method will check it.

sundy-li: The inner Table0 will check it. So it's duplicated here.

BohuTANG (approved):

PR #9712, opened 2023-01-23
feat:Feat cache bloom index obj

dantengsky:

## Summary

- cache `Xor8Filter` instead of raw bytes for bloom index cache
- avoid using `infer_schema` while deserializing bloom index data
- introduce new crate `storages-common-cache-manager` 

 note:

1. an unsafe call was introduced during deserializing index filter,
https://github.com/datafuselabs/databend/blob/2229c768d0e94ed9ea9773b4c70bfa9a06746eb6/src/query/storages/fuse/src/io/read/bloom/column_filter_reader.rs#L132-L134
   @sundy-li , need your help, please have a look if it is correct

2. there is a breaking change of the configuration
   `table_cache_bloom_index_data_bytes` has been renamed to `table_cache_bloom_index_filter_count`, with default value 1024 * 1024
   @Xuanwo  need your help, please have a look if it is acceptable

Closes #issue


src/query/storages/common/table-meta/src/caches/meta_cache_manager.rs
```diff
@@ -58,8 +57,11 @@ impl CacheManager {
             let table_snapshot_cache = Self::new_item_cache(config.table_cache_snapshot_count);
             let table_statistic_cache = Self::new_item_cache(config.table_cache_statistic_count);
             let segment_info_cache = Self::new_item_cache(config.table_cache_segment_count);
+            // TODO remove this
+            // let bloom_index_data_cache =
+            //    Self::new_bytes_cache(config.table_cache_bloom_index_data_bytes);
             let bloom_index_data_cache =
-                Self::new_bytes_cache(config.table_cache_bloom_index_data_bytes);
+                Self::new_item_cache(config.table_cache_bloom_index_data_bytes);
```
BohuTANG, line 64: Is the capacity `table_cache_bloom_index_data_bytes` not correct?

src/query/storages/common/table-meta/src/caches/meta_cache.rs
```diff
@@ -30,7 +30,7 @@ pub type TableSnapshotCache = InMemoryItemCacheHolder<TableSnapshot>;
 pub type TableSnapshotStatisticCache = InMemoryItemCacheHolder<TableSnapshotStatistics>;
 /// In memory data cache of bloom index data.
 /// For each indexed data block, the index data of column is cached individually
-pub type BloomIndexCache = InMemoryBytesCacheHolder;
+pub type BloomIndexCache = InMemoryItemCacheHolder<Xor8>;
```
BohuTANG, line 33: `Xor8` is one implementation of `Filter`. Can we use the `Filter` instead of using the Xor8 directly?

src/query/storages/common/table-meta/src/caches/meta_cache_manager.rs
dantengsky: thanks for reminding me

src/query/storages/common/table-meta/src/caches/meta_cache.rs
dantengsky: thanks for reminding me.   A new crate `cache-manager` is being extracted, after that, the BloomIndexCachewill hold Xor8Filter directly (it depends on crate `storages-common-index`)

Xuanwo:
> 2\. there is a breaking change of the configuration
>     `table_cache_bloom_index_data_bytes` has been renamed to `table_cache_bloom_index_filter_count`, with default value 1024 * 1024

LGTM

src/query/storages/fuse/src/io/read/bloom/column_filter_reader.rs
```diff
@@ -0,0 +1,147 @@
...
+        let column_type = self.column_descriptor.descriptor.primitive_type.clone();
+        let filed_name = self.column_descriptor.path_in_schema[0].to_owned();
+        let field = ArrowField::new(filed_name, DataType::Binary, false);
+        let mut array_iter =
+            column_iter_to_arrays(vec![decompressor], vec![&column_type], field, None, 1)?;
+        if let Some(array) = array_iter.next() {
+            let array = array?;
+            let col =
+                Column::from_arrow(array.as_ref(), &common_expression::types::DataType::String);
+            let bytes = unsafe { col.as_string().unwrap().index_unchecked(0) };
```
sundy-li, line 134: it's safe if we ensure it has one element.  But it's better to throw exceptions rather than panic.

```
col.as_string().and_then ....
```

sundy-li (approved):

sundy-li:
Rest looks good to me.

src/query/storages/fuse/src/metrics/fuse_metrics.rs
```diff
@@ -95,17 +95,17 @@ pub fn metrics_inc_block_index_write_milliseconds(c: u64) {
     increment_gauge!(key!("block_index_write_milliseconds"), c as f64);
 }
 
-pub fn metrics_inc_block_index_read_nums(c: u64) {
-    increment_gauge!(key!("block_index_read_nums"), c as f64);
-}
+// pub fn metrics_inc_block_index_read_nums(c: u64) {
```
BohuTANG, line 98: No used anymore?

BohuTANG: Others LGTM

BohuTANG (approved):

src/query/storages/fuse/src/metrics/fuse_metrics.rs
dantengsky: > No used anymore?

yes, replaced by `metrics_inc_cache_access_count` in the `CachedReader`. and the label of it changed to `bloom_index_filter_cache_access_count`

PR #9715, opened 2023-01-24
feat(query): Add optional column names to create/alter view statement

### LINKED ISSUE #6854: Feature: Make Databend's create view syntax closer to mysql
**Summary**
It can be parsed normally in mysql, but the following error is reported in databend
```
mysql> create view v1(a) AS select number from numbers(10);
ERROR 1105 (HY000): Code: 1005, displayText = sql parser error: need `AS` after VIEW NAME.
```

johnhaxx7:

## Summary

Add optional column names to create/alter view statement

CREATE VIEW [IF NOT EXISTS] [db.]view_name **[(\<column\>, ...)]** AS SELECT query
ALTER VIEW [db.]view_name **[(\<column\>, ...)]** AS SELECT query

so that column names can be renamed.

If column names is not empty, the subquery is rewrote to 
```
select * from ({subquery}) {viewname}( names ) 
``` 
When we select/describe from the view, it'll parse from the new subquery to ensure we get the right column names.

Thanks to @sundy-li for the idea!

Closes #6854


src/query/service/src/interpreters/interpreter_view_create.rs
```diff
@@ -69,7 +70,28 @@ impl CreateViewInterpreter {
...
         let mut options = BTreeMap::new();
-        options.insert("query".to_string(), self.plan.subquery.clone());
+
+        let subquery = if self.plan.column_names.is_empty() {
+            self.plan.subquery.clone()
+        } else {
+            let mut planner = Planner::new(self.ctx.clone());
+            let (plan, _, _) = planner.plan_sql(&self.plan.subquery.clone()).await?;
+            if plan.schema().fields().len() != self.plan.column_names.len() {
+                return Err(ErrorCode::BadDataArrayLength(format!(
```
johnhaxx7, line 80: Please advice if `BadDataArrayLength` is appropriate, thanks!

src/query/service/src/interpreters/interpreter_view_alter.rs
```diff
@@ -86,7 +87,27 @@ impl AlterViewInterpreter {
...
+                    "column name length mismatch, expect {}, got {}",
+                    plan.schema().fields().len(),
+                    self.plan.column_names.len(),
+                )));
+            }
+            format!(
+                "select * from ({}) {}({})",
+                self.plan.subquery,
+                self.plan.viewname,
+                self.plan.column_names.join(",")
```
sundy-li, line 106: `join(", ")`

src/query/service/src/interpreters/interpreter_view_create.rs
sundy-li: It's ok

src/query/service/src/interpreters/interpreter_view_alter.rs
johnhaxx7: Done!

sundy-li (approved):

johnhaxx7:
Hi @BohuTANG , seems like I don't have access to merge, can you please help if possible? thanks!

PR #9722, opened 2023-01-25
chore(fuse/metrics): better the metrics name of pruning

BohuTANG:

## Summary
Better the result:
```
truncate table system.metrics;
select * from system.metrics where metric  like '%pruning%' order by metric;
```

Closes #issue


BohuTANG:
@mergify update

PR #9738, opened 2023-01-26
feat(query): add topn runtime filter in native storage format

sundy-li:

## Summary

Summary about this PR

If we push down the sort + limit to storage level, we can have `TopN` opt.

But it must work with runtime predication (prewhere) + Page index optimization , so currently we only have it in native storage format.

Query `select  * from hits_page where URL like '%google%' order by EventTime  desc  limit 10;` can works 5 times faster.


src/query/expression/src/kernels/filter.rs
```diff
@@ -380,3 +314,95 @@ impl Column {
         new.into()
     }
 }
+
+impl FilterHelpers {
```
BohuTANG, line 318: It would be better to move FilterHelpers functions to `expresion` or other places? FilterHelpers is not bound to `kernels/filter`.

BohuTANG:
Good idea.
Can we use `OrderBy(min-max)+ Limit` for fuse block pruning(OrderByt must be cluster key) to reduce the block size?

sundy-li:
> Good idea. Can we use `OrderBy(min-max)+ Limit` in reading fuse block data to reduce the block size?

Unless we have prewhere and page index.

src/query/expression/src/kernels/topk.rs
```diff
@@ -0,0 +1,227 @@
...
+            DataType::Number(NumberDataType::NUM_TYPE) =>
+                self.push_column_internal::<NumberType::<NUM_TYPE>>(col, bitmap),
+            DataType::String => self.push_column_internal::<StringType>(col, bitmap),
+            DataType::Timestamp => self.push_column_internal::<TimestampType>(col, bitmap),
+            DataType::Date => self.push_column_internal::<DateType>(col, bitmap),
+            _ => {}
+        });
+    }
+
+    pub fn push_column_internal<T: ValueType>(&mut self, col: &Column, bitmap: &mut MutableBitmap)
```
BohuTANG, line 55:
```suggestion
    fn push_column_internal<T: ValueType>(&mut self, col: &Column, bitmap: &mut MutableBitmap)
```

BohuTANG (approved):

PR #9754, opened 2023-01-28
make partial aggregator support streaming 

sundy-li:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3089,6 +3089,7 @@ dependencies = [
  "storages-common-cache",
  "storages-common-index",
  "storages-common-table-meta",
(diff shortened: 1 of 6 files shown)


sundy-li:
Strange bug
```
MySQL [(none)]> select to_int64(number) a , to_int64(number % 1 ) b, count(number) c, sum(number) , avg(number) from numbers(10000000) group by a,b order by c desc limit 10;

ERROR 1105 (HY000): Code: 1104, displayText = assertion failed: self.data.len() >= *self.offsets.last().unwrap() as usize.
MySQL [(none)]>
```

PR #9755, opened 2023-01-28
fix(metactl): when import data to meta-service dir, the specified "id" has to be one of the "initial_cluster"

drmingdrmer:

##### fix(metactl): when import data to meta-service dir, the specified "id" has to be one of the "initial_cluster"

Otherwise nothing will be done and error returns, e.g., `Error: node id (0) has to be one of cluster member([1])`

## Related Issues


everpcpc (approved):
:clap:

PR #9765, opened 2023-01-29
chore(mem_allocator): JEAllocator -> AllocatorProxy

BohuTANG:

## Summary

* better the name from `JEAllocator -> AllocatorProxy`

Closes #issue

### DIFF
--- a/src/common/base/src/mem_allocator/je_allocator.rs
+++ b/src/common/base/src/mem_allocator/je_allocator.rs
@@ -16,18 +16,6 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-#[derive(Debug, Clone, Copy, Default)]
-pub struct JEAllocator<T> {
-    #[allow(dead_code)]
-    allocator: T,
-}
-
-impl<T> JEAllocator<T> {
-    pub fn new(allocator: T) -> Self {
-        Self { allocator }
-    }
-}
-
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 pub mod linux_or_macos {
     use std::alloc::AllocError;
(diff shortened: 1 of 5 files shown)


BohuTANG:
`mem_allocator` crate is not clear enough and needs some refactoring work.

src/common/base/src/mem_allocator/je_allocator.rs
```diff
@@ -80,7 +64,7 @@ pub mod linux_or_macos {
         }
     }
 
-    unsafe impl<T: Allocator> Allocator for JEAllocator<T> {
+    unsafe impl<T: Allocator> Allocator for AllocatorProxy<T> {
```
zhang2014, line 67: implementation is jemalloc? maybe JEAllocator is better?

BohuTANG: I think we have two layer wrappers is bit confused here: 
`JEAllocator ` is a general wrapper(not a `jemalloc` wrapper), it is defined as:
```
#[derive(Debug, Clone, Copy, Default)]
pub struct JEAllocator<T> {
    #[allow(dead_code)]
    allocator: T,
}

impl<T> JEAllocator<T> {
    pub fn new(allocator: T) -> Self {
        Self { allocator }
    }
}
```

`MmapAllocator` also define his wrapper(It's a general too):
```
pub struct MmapAllocator<T> {
    allocator: T,
}

impl<T> MmapAllocator<T> {
    pub fn new(allocator: T) -> Self {
        Self { allocator }
    }
}
```

At final, we defined `GlobalAllocator` as follows:
 ```
pub struct GlobalAllocator

type Fallback = JEAllocator<SystemAllocator>;
````
and using the `JEAllocator`:
```
#[global_allocator]
pub static GLOBAL_ALLOCATOR: GlobalAllocator = GlobalAllocator;
```

How can we change the allocator from `jemalloc` to `mmap` or a new custom allocator? 

The better way is we define `GlobalAllocator` as only allocator wrapper:
```
#[derive(Debug, Clone, Copy, Default)]
pub struct GlobalAllocator<T> {
    #[allow(dead_code)]
    allocator: T,
}

impl<T> GlobalAllocator<T> {
    pub fn new(allocator: T) -> Self {
        Self { allocator }
    }
}
```
And we use it as:
```
#[global_allocator]
pub static GLOBAL_ALLOCATOR: GlobalAllocator<JeAllocator> = GlobalAllocator<JeAllocator>;
```

or 
```
#[global_allocator]
pub static GLOBAL_ALLOCATOR: GlobalAllocator<MmapAllocator> = GlobalAllocator<MmapAllocator>;
```

If missing something, please correct me :)

BohuTANG:
Closed, still need to research deeply.

PR #9769, opened 2023-01-29
feat: `read_parquet` supports reading from remote locations.

RinChanNOWWW:

## Summary

- Read meta and prune partitions in `read_partitions`. (Before: in `set_on_init`).
- Distinguish meta reading between sync and async.
- Distinguish data reading between sync and async.
- Use OpenDAL to do glob operations. (Only simple patterns are supported).
- Exposed two APIs for creating `ParquetTable`. (`create` for non-blocking, `blocking_create` for blocking)

P.S. 

`read_parquet` will **NOT** support syntax like:

```sql
select * from read_parquet('s3://bucket/xxx/', connection => xxx);
```

Reading parquets from stage table  will use the kernel of `read_parquet` later, so we can achieve the remote files reading by using stage table.


Xuanwo (commented):
Mostly LGTM, can you add a sqllogictest or statfule test for it?

RinChanNOWWW:
There is no interactive way to test the remote parquet reading. We can test it after `select stage` using the kernel of `read_parquet`.

src/query/storages/parquet/src/table_function/table.rs
```diff
@@ -67,10 +78,31 @@ impl ParquetTable {
...
+    pub fn create(
+        database_name: &str,
+        table_func_name: &str,
+        table_id: u64,
+        table_args: TableArgs,
+        operator: Operator,
+    ) -> Result<Arc<dyn TableFunction>> {
+        if !GlobalConfig::instance().storage.allow_insecure {
+            return Err(ErrorCode::StorageInsecure(
+                "Should enable `allow_insecure` to use table function `read_parquet`",
```
youngsofun, line 103: I guess the `create` is to be called by `select from stage`, (I`m trying something like this locally)

1. should not check it in `create_table_function` here?
2. instead of faking a  table_args outside, what about this


```
    pub fn create_from_files(
        database_name: &str,
        table_func_name: &str,
        table_id: u64,
        file_locations: Vec<String>,
        operator: Operator,
        read_options: ReadOptions,
        table_args: Vec<Scalar>,
    ) -> Result<Arc<dyn TableFunction>> 
```
```

**Comment by Xuanwo** [comment]
> There is no interactive way to test the remote parquet reading. We can test it after `select stage` using the kernel of `read_parquet`.

Oh, I don't like this idea. Can we test this function in stateful tests? Seems it's use the global operator?

File: src/query/storages/parquet/src/table_function/table.rs (hunk 1 above)
Comment (RinChanNOWWW) [reply src/query/storages/parquet/src/table_function/table.rs:103]: I will add other methods for `select from stage`.

**Comment by RinChanNOWWW** [comment]
> Oh, I don't like this idea. Can we test this function in stateful tests? Seems it's use the global operator?

It uses fs operator now. If using the global operator, we can only access the files on the corresponding backend.

**Comment by RinChanNOWWW** [comment]
I think we can add an option like:

````sql
select * from read_parquet('xxx', fs=true);
-- or
select * from read_parquet('xxx', local=true);
```

to let `read_parquet` read local files, otherwise, use the global operator.

sundy-li (approved):

youngsofun:
> It's need we supporting from stage syntax? Please add a issue to tacking it, so all of us will know what's we will work on cc @youngsofun

do not have to support stage in `read parquet`. 
see https://github.com/datafuselabs/databend/pull/9801

youngsofun:
> ```sql
> ```sql
> 
> select * from read_parquet('xxx', local=true);
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> to let read_parquet read local files, otherwise, use the global operator.
> And we can test reading remote file if the storage backend is a remote one.
> ```

```, use the global operator.``` is not enough,  for internal at least need prefix,  and must be legal

youngsofun (approved):

Xuanwo:
How about removing this sql function if we only use it internally after select stage has been supported?

PR #9770, opened 2023-01-29
chore(mem_allocator): remove inner allocator for JEAllocator

BohuTANG:

## Summary

* remove inner allocator for JEAllocator, no need for fallback
* better the allocator file name

Closes #issue


src/common/base/src/mem_allocator/jemalloc.rs
```diff
@@ -224,61 +212,3 @@ pub mod linux_or_macos {
         }
     }
 }
-
-#[cfg(not(any(target_os = "linux", target_os = "macos")))]
```
BohuTANG, line 228: If arm works we can remove these codes, we only support Linux and macOS and arm target.

PR #9771, opened 2023-01-29
feat: add syntax about parsing presign options with content type

flaneur2020:

## Summary

Currently the PRESIGN statement gives a default empty Content-Type header on presigning requests. Unfortunately this behaviour got a trouble on uploading files to OSS in the browser:

- OSS restricts the Content-Type header in request should be same as the Content-Type header in the presigned argument, or it will not pass the authentication.
- browser do not allow make a ajax request without Content-Type header, if the script did not specify the Content-Type header, the browser will add one Content-Type: text/plain automatically, which would fail the PRESIGN authentication.

This PR adds a CONTENT_TYPE option to the PRESIGN statement, so the front end can specify a valid CONTENT_TYPE on uploading files in frontend.

- [x] parser about CONTENT_TYPE
- [x] PRESIGN UPLOAD with content-type https://github.com/datafuselabs/opendal/pull/1252


src/query/ast/src/parser/statement.rs
```diff
@@ -895,16 +895,19 @@ pub fn statement(i: Input) -> IResult<StatementMsg> {
...
+        |(_, action, location, opts)| {
+            let mut presign_stmt = PresignStmt {
                 action: action.unwrap_or_default(),
                 location,
-                expire: expire
-                    .map(|(_, _, v)| Duration::from_secs(v))
-                    .unwrap_or_else(|| Duration::from_secs(3600)),
-            })
+                expire: Duration::from_secs(3600),
+                content_type: None,
```
Xuanwo, line 905: How about use `application/octet-stream` as default value if not input?

flaneur2020: makes sense, but i'm a bit afraid of this may break the existing code in SDKs, if it defaults as None, a request without Content-Type header can be successful, however if we make the default value as application/octet-stream, the code without setting Content-Type to application/octet-stream would fail :(

Xuanwo: Based on our `presign` API design, it's required to add all headers that we returned into requests to make it work.

Xuanwo (approved):

flaneur2020:
@Xuanwo added the plan & interpreter part along with the PR, PTAL

Xuanwo (approved):

src/query/ast/src/parser/statement.rs
BohuTANG: mark the conversation as resolved?

flaneur2020: after talk, keeping the behavior with current is ok

PR #9790, opened 2023-01-30
feat(query): hold table function instances in query context.

RinChanNOWWW:

## Summary

After instantiating a table function in binding phase, store the instance in the current query context. Take the table function instance from the query context when building pipeline.

The table function instances in context are indexed by `{global-table-id}-{uuid}`.

### Implementation in the past.

In the past, table function instantialization will happen both in binding phase and pipeline building phase, which is redundant.

For table function whose creating is heavy (like `read_parquet`), it will affect the performance.


youngsofun:
I`m working on another way

add  DataSourceInfo::ParquetSource
let get_data_source_info return DataSourceInfo::ParquetSource

PR #9793, opened 2023-01-30
feat(copy): support skipfile mode

ariesdevil:

## Summary

Support skipfile_n when do copy into table.

Closes #issue


src/query/expression/src/block.rs
```diff
@@ -66,13 +67,22 @@ pub trait BlockMetaInfo: Debug + Send + Sync {
...
     #[inline]
     pub fn new(columns: Vec<BlockEntry>, num_rows: usize) -> Self {
-        DataBlock::new_with_meta(columns, num_rows, None)
+        DataBlock::new_with_meta(columns, num_rows, None, None)
+    }
+
+    pub fn new_with_belong_to(
+        columns: Vec<BlockEntry>,
+        num_rows: usize,
+        belong_to: String,
```
sundy-li, line 76: Other field must be inside meta, it's too intrusive

sundy-li: DataBlock don't know anything about field_name.

src/query/expression/src/block.rs
```diff
@@ -104,6 +115,11 @@ impl DataBlock {
         DataBlock::new(columns, num_rows)
     }
 
+    #[inline]
+    pub fn attach_filename(self, filename: String) -> Self {
+        Self::new_with_belong_to(self.columns, self.num_rows, filename)
```
sundy-li, line 120: `attach_filename` should not used in block.

ariesdevil:
Last Thursday and Friday, I finished filtering the files by adding a transform layer, but during testing, I found that this would be very easy to OOM, and even though we will explain it in the documentation, the user experience is still very bad.

After thinking about it over the weekend, I decided to use the current method, i.e., put the blocks generated by the same file under the same segment when sink, and then filter the segment when commit.

The key point of `skipfile` is to make sure that each phase contains only the contents of the same file in a single block, otherwise the file cannot be filtered at the end.

src/query/expression/src/block.rs
youngsofun, line 120: should not lost meta?

ariesdevil: done

ariesdevil: if the block has no relation with the original file, we hard to filter block when doing skipfile.

BohuTANG: Can we add it to the block meta?

src/query/service/src/sessions/query_ctx.rs
```diff
@@ -193,8 +196,21 @@ impl QueryContext {
...
 
-    pub fn get_on_error_map(&self) -> Option<HashMap<String, ErrorCode>> {
-        self.shared.get_on_error_map()
+    pub fn get_skipped_file(&self, threshold: usize) -> Option<Vec<String>> {
+        if let Some(on_error_map) = self.get_on_error_map() {
+            if on_error_map.is_empty() {
+                return None;
+            }
+            let mut v = Vec::new();
+            on_error_map.iter().for_each(|x| {
```
youngsofun, line 205: what about rename x, h to words to make it easier to read.

src/query/service/src/sessions/query_ctx_shared.rs
```diff
@@ -79,7 +82,10 @@ pub struct QueryContextShared {
     pub(in crate::sessions) precommit_blocks: Arc<RwLock<Vec<DataBlock>>>,
     pub(in crate::sessions) stage_attachment: Arc<RwLock<Option<StageAttachment>>>,
     pub(in crate::sessions) created_time: SystemTime,
-    pub(in crate::sessions) on_error_map: Arc<RwLock<Option<HashMap<String, ErrorCode>>>>,
+    #[allow(clippy::type_complexity)]
```
youngsofun, line 85: add a comment for key 2-level map?  e.g. the keys?

src/query/storages/fuse/src/io/read/meta/meta_readers.rs
```diff
@@ -133,6 +134,7 @@ async fn bytes_reader(op: &Operator, path: &str, len: Option<u64>) -> Result<Obj
     let len = match len {
         Some(l) => l,
         None => {
+            // toto here.
```
youngsofun, line 137: todo

sundy-li:
Convert to draft and  will complete it in the future.

src/query/pipeline/sources/src/input_formats/source_deserializer.rs
```diff
@@ -111,7 +111,9 @@ impl<I: InputFormatPipe> Processor for DeserializeSource<I> {
         }
         let blocks = self.block_builder.deserialize(self.input_buffer.take())?;
         for b in blocks.into_iter() {
-            self.output_buffer.push_back(b)
+            if !b.is_empty() {
```
youngsofun, line 114: row batches from diff files may be delivered to the same  DeserializeProcessor or DeserializeSource, and merge into one block, which should be forbiddened when use skip_file_num,  is this handled in this  pr?

ariesdevil: Yup, handled here: https://github.com/datafuselabs/databend/pull/9793/files#diff-23dfe539c8b7d8c6d471f64802aa6377c323cb2465c59448e5639d1d91a7111bR544

BohuTANG:
Suggest planning this PR to low-priority, `skipfile_n` does not look like a strong requirement at the moment. cc @sundy-li @ariesdevil

ariesdevil:
> Suggest planning this PR to low-priority, `skipfile_n` does not look like a strong requirement at the moment. cc @sundy-li @ariesdevil↳

We'll talk about it at today's group meeting.

PR #9794, opened 2023-01-30
feat(query): tail-array optimization in aggregate partial

sundy-li:

## Summary

Summary about this PR

If the reduce_factor is too low, we use tail-array in hashtable via `enable_tail_array` method.

Closes #issue

### DIFF
--- a/src/common/hashtable/src/tail_array.rs
+++ b/src/common/hashtable/src/tail_array.rs
@@ -1,0 +1,133 @@
+// Copyright 2021 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::alloc::Allocator;
+
+use super::table0::Entry;
+use super::traits::Keyable;
+use crate::container::Container;
+use crate::container::StackContainer;
+
+const SIZE: usize = 8192;
+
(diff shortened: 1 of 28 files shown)


src/common/hashtable/src/tail_array.rs
```diff
@@ -0,0 +1,133 @@
...
+use std::alloc::Allocator;
+
+use super::table0::Entry;
+use super::traits::Keyable;
+use crate::container::Container;
+use crate::container::StackContainer;
+
+const SIZE: usize = 8192;
+
+pub struct TailArray<K, V, A: Allocator> {
```
BohuTANG, line 24: It's better to add unit test for the tail array.

src/common/hashtable/src/tail_array.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn new(allocator: A) -> Self {
+        Self {
+            datas: vec![],
+            num_items: 0,
+            allocator,
+        }
+    }
+
+    pub fn insert(&mut self, key: K) -> &mut Entry<K, V> {
+        let pos = self.num_items % SIZE;
```
zhang2014, line 44: tail array is ring buffer?

sundy-li: No, it's a `Vector<  [Entrys; N]  >`. 

If we enable the tail array, the new keys will be pushed into this vector without expanding the hashtable.

sundy-li:
Improved a little.


image


Memory went down fast after running the whole bench scripts.

image

src/common/hashtable/src/tail_array.rs
zhang2014: Is it only for the last block?

src/query/service/src/pipelines/processors/transforms/aggregator/aggregator_twolevel.rs
```diff
@@ -213,8 +227,11 @@ where
             return Ok(data_blocks);
         }
 
-        drop(agg.area.take());
-        agg.states_dropped = true;
+        if agg.pass_state_to_final {
```
zhang2014, line 230: drop area if agg.pass_state_to_final == false?

PR #9798, opened 2023-01-31
feat(storage): native storage format support nested data types

b41sh:

## Summary

- native storage format support nested data types
- skip filtered pages using NativeReader `nth` function

Closes #issue


sundy-li:
hits native lz4 loading time from `2:45.25 total` -> `2:59.12 total`

Performance dropped
image

lichuang:
need to add some test case too.

lichuang:
I have impl a suite test for native format of add\drop column, see: https://github.com/datafuselabs/databend/blob/main/tests/sqllogictests/suites/base/05_ddl/05_0029_ddl_alter_table_add_drop_column_native,but at the time native does not support of nested types, so this suite has just add\drop for simple data type.

If this pr which make native format support nested data types has been done, please copy parquet add\drop column  suite test ( see: https://github.com/datafuselabs/databend/blob/main/tests/sqllogictests/suites/base/05_ddl/05_0028_ddl_alter_table_add_drop_column) to native format, to test if native format of add\drop nested data type if ok.

sundy-li:
image


performance dropped a lot.

b41sh:
image
Performance is the same as before the revision, with a slight improvement.

sundy-li (approved):

PR #9810, opened 2023-01-31
chore: change cross join cost model

xudong963:

## Summary

tpch q2 join order will be right, there was a cross join before.

Closes #issue


BohuTANG:
Can we build a SQL logic test for it?

Issue #9814, opened 2023-01-31
Feature: http handler always return JSON


youngsofun:
**Summary**

especially when error.

some error are raised by poem, before into our logic, which are plain text.
need to investigate how to config it to gen json too.

youngsofun:
cc @flaneur2020

flaneur2020:
cc @Chasen-Zhang

PR #9834, opened 2023-02-01
feat: http handler always return JSON body.

youngsofun:

## Summary

Summary about this PR

Closes https://github.com/datafuselabs/databend/issues/9814


Xuanwo (approved):

tests/suites/1_stateful/09_http_handler/09_0001_json_response.result
```diff
@@ -0,0 +1,3 @@
+{"code":1025,"message":"Unknown table 't1'"}
```
Xuanwo, line 1: What's the different between `code` and `error.code`?

youngsofun: they are all .error.code

the first contain fields like .queryid so we only compare .error field

youngsofun:
seem s test_unit is blocked, let me check it

flaneur2020 (approved):

PR #9841, opened 2023-02-01
feat(memory): add new allocator mimalloc

BohuTANG:

## Summary

* add new memory allocator mimalloc

Closes #issue


sundy-li:
mimalloc is added multiple times, but it had bug  hard to track.

BohuTANG:
> mimalloc was added multiple times, but it had bug hard to track.
Right, I have tested and still have many problem s, this PR for test only.

PR #9851, opened 2023-02-02
feat: alter table add/drop column SQL support

lichuang:

## Summary

1. feat: add create_with_schema_from_chunk api in DataBlock to fill default field value
2. add more test in column_node
3. alter table add/drop column SQL support and relate test cases.

Closes https://github.com/datafuselabs/databend/issues/9846 https://github.com/datafuselabs/databend/issues/9869


src/query/expression/src/block.rs
```diff
@@ -342,6 +342,49 @@ impl DataBlock {
 
         Ok(DataBlock::new(cols, arrow_chunk.len()))
     }
+
+    // If use_field_default_vals[i].is_some(), then DataBlock.column[i] = num_rows * DataField.default_value().
+    // else, DataBlock.column[i] = chuck.column.
+    // For example, Schema.field is [a,b,c] and use_field_default_vals is [Some(), None, Some()],
+    // then the return block column will be [a.default_value()*num_rows, chunk.column[0], c.default_value()*num_rows].
+    pub fn create_with_schema_from_chunk<A: AsRef<dyn Array>>(
```
BohuTANG, line 350: Name `create_with_schema_from_chunk` to `create_with_default_value` better?

lichuang: Name `create_with_schema_from_chunk` to `create_with_default_value_and_chunk`

BohuTANG: `create_with_schema_from_chunk` to `create_with_default_value_from_chunk` is better?

src/query/expression/tests/it/fill_field_default_value.rs
```diff
@@ -0,0 +1,86 @@
...
+    let chunks: Chunk<ArrayRef> = chunk_block.clone().try_into().unwrap();
+    let use_field_default_vals = vec![None, Some(()), None, Some(()), None];
+    let new_block: DataBlock = DataBlock::create_with_default_value_and_chunk(
+        &schema,
+        &chunks,
+        &use_field_default_vals,
+        num_rows,
+    )
+    .unwrap();
+    assert_eq!(3, new_block.num_rows());
```
BohuTANG, line 58: How about printing the block with a file instead of the rust code check? We have example in `tests/it/kernel.rs`.

src/query/expression/src/block.rs
```diff
@@ -342,6 +342,49 @@ impl DataBlock {
...
+    // else, DataBlock.column[i] = chuck.column.
+    // For example, Schema.field is [a,b,c] and use_field_default_vals is [Some(), None, Some()],
+    // then the return block column will be [a.default_value()*num_rows, chunk.column[0], c.default_value()*num_rows].
+    pub fn create_with_default_value_and_chunk<A: AsRef<dyn Array>>(
+        schema: &DataSchemaRef,
+        chuck: &ArrowChunk<A>,
+        use_field_default_vals: &[Option<()>],
+        num_rows: usize,
+    ) -> Result<DataBlock> {
+        assert_eq!(schema.fields().len(), use_field_default_vals.len());
```
sundy-li, line 356: debug assert

src/query/expression/src/block.rs
```diff
@@ -342,6 +342,49 @@ impl DataBlock {
...
+        data_block.num_rows = num_rows;
+        let mut chunk_idx: usize = 0;
+        let chunk_columns = chuck.arrays();
+        let schema_fields = schema.fields();
+
+        for (i, use_field_default_val) in use_field_default_vals.iter().enumerate() {
+            let field = &schema_fields[i];
+            let data_type = field.data_type();
+            let block = if use_field_default_val.is_some() {
+                let default_value = data_type.default_value();
```
sundy-li, line 367: should respect the default_expression

src/query/expression/tests/it/kernel.rs
```diff
@@ -212,115 +208,3 @@ pub fn test_pass() {
         3,
     );
 }
-
-fn run_filter(file: &mut impl Write, predicate: Column, block: &DataBlock) {
```
sundy-li, line 216: Why do you delete these codes?

sundy-li:
`create_with_default_value_and_chunk` is a new util method, you can continue to work on this pr and bring `create_with_default_value_and_chunk` into the real query path.

src/query/expression/tests/it/kernel.rs
lichuang: `test_data_block_create_with_default_value_and_chunk` need `run_take` function, so i move all `run_*` test util functions to `common.rs`.

BohuTANG:
> `create_with_default_value_and_chunk` is a new util method, you can continue to work on this pr and bring `create_with_default_value_and_chunk` into the real query path.

@lichuang 
What do you think? Continue working on this PR or next?

BohuTANG:
```
[SQL] SELECT * FROM `05_0029_t0`
[Diff] (-expected|+actual)
+   2 1.0
    1 0.0
-   2 1.0
at tests/sqllogictests/suites/base/05_ddl/05_0029_ddl_alter_table_add_drop_column_native:19
```

src/query/expression/src/block.rs
```diff
@@ -342,6 +345,113 @@ impl DataBlock {
...
+    ) -> Self {
+        let mut data_block = DataBlock::empty();
+        data_block.num_rows = num_rows;
+        let schema_fields = schema.fields();
+
+        for (i, default_val) in default_vals.iter().enumerate() {
+            let field = &schema_fields[i];
+            let data_type = field.data_type();
+
+            let builder = ColumnBuilder::repeat(&default_val.as_ref(), num_rows, data_type);
```
sundy-li, line 408: We should not repeat to a column, should be  `Value::Scalar` .

src/query/service/src/interpreters/interpreter_table_add_column.rs
```diff
@@ -0,0 +1,109 @@
+// Copyright 2021 Datafuse Labs.
```
b41sh, line 1: 2021 -> 2023

lichuang:
@sundy-li @BohuTANG @b41sh @dantengsky @RinChanNOWWW

src/query/storages/fuse/src/io/read/block/block_reader_parquet.rs
```diff
@@ -116,6 +126,70 @@ impl BlockReader {
         deserialized_res
     }
 
+    pub(crate) fn schema_default_vals(
+        data_schema: DataSchema,
+        table: Arc<dyn Table>,
+        ctx: Arc<dyn TableContext>,
+    ) -> Result<Vec<Scalar>> {
+        GlobalIORuntime::instance().block_on(async move {
```
BohuTANG, line 134: why need `async move`?

lichuang: need to call `parse_exprs` function, which call tokio `block_in_place` function to resolve expr:

```
pub fn parse_exprs(
    ctx: Arc<dyn TableContext>,
    table_meta: Arc<dyn Table>,
    unwrap_tuple: bool,
    sql: &str,
) -> Result<Vec<Expr>> {
  // ...
      let exprs = ast_exprs
        .iter()
        .map(|ast| {
            let (scalar, _) =
                *block_in_place(|| Handle::current().block_on(type_checker.resolve(ast, None)))?;
            let expr = scalar.as_expr_with_col_index()?;
            Ok(expr)
        })
        .collect::<Result<_>>()?;
}
```

src/query/storages/fuse/src/io/read/block/block_reader_native.rs
```diff
@@ -163,6 +166,55 @@ impl BlockReader {
...
+            let mut num_rows = 0;
+            for part in parts {
+                let part = FusePartInfo::from_part(part)?;
+                num_rows += part.nums_rows;
+            }
+
+            let data_schema = self.data_schema();
+            let ctx = self.ctx.clone();
+            let table = self.table.clone();
+            let default_vals = Self::schema_default_vals(data_schema, table, ctx)?;
```
sundy-li, line 204: This method must be initialized only once, it's a very cost which calls a heavy block_on!

src/query/storages/fuse/src/io/read/block/block_reader_parquet.rs
sundy-li: Dangerous block_in_place inside a hot loop.

src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -99,6 +99,8 @@ impl AccessChecker for PrivilegeAccess {
                     .await?;
             }
             Plan::RenameTable(_) => {}
+            Plan::AddTableColumn(_) => {}
```
BohuTANG, line 102: Here should add privileges for Add/Drop table column and a SQL logic test for them.

sundy-li:
Still in poor performance.

`schema_default_vals` must be called only once during the whole  SQL lifetime, but now it's called per block.

I suggest initialing a  field:  `Vec<Scalar>` in the block_reader constructor.

src/query/storages/fuse/src/io/read/block/block_reader_parquet.rs
```diff
@@ -126,42 +136,56 @@ impl BlockReader {
...
+                    column_chunks,
+                    num_rows,
+                    column_descriptors,
+                    field,
+                    compression,
+                    uncompressed_buffer
+                        .clone()
+                        .unwrap_or_else(|| UncompressedBuffer::new(0)),
+                )?);
+                need_default_vals.push(None);
```
sundy-li, line 184: Why not use boolean?

src/query/expression/src/block.rs
```diff
@@ -342,6 +345,106 @@ impl DataBlock {
...
+                }
+            };
+
+            columns.push(column);
+        }
+
+        Ok(DataBlock::new(columns, num_rows))
+    }
+
+    pub fn create_with_default_value(
```
sundy-li, line 389: this method could be forwarded into `create_with_default_value_and_chunk`.

sundy-li (approved):
LGTM, ci need to run perf of this pr to see the performance issue.

BohuTANG (approved):

BohuTANG:
> LGTM, ci need to run perf of this pr to see the performance issue.

@everpcpc how can we run the perf test for a special PR?

drmingdrmer:
@mergify update

drmingdrmer:
@mergifyio refresh

everpcpc:
> > LGTM, ci need to run perf of this pr to see the performance issue.
> 
> @everpcpc how can we run the perf test for a special PR?

Adding lable `ci-perf` would enable perf for this PR.

tests/sqllogictests/suites/base/05_ddl/05_0028_ddl_alter_table_add_drop_column
```diff
@@ -0,0 +1,134 @@
...
+query I
+SELECT * FROM `05_0028_at_t0`
+----
+[101,2]
+
+statement error 1109
+ALTER TABLE `05_0028_at_t0` DROP COLUMN c
+
+statement error 1108
+ALTER TABLE `05_0028_at_t0` ADD COLUMN c float
```
dantengsky, line 79: is the error code supposed to be 1108? (`AddColumnExistError(1108)`)

dantengsky: I misunderstand the case, column c is still there

dantengsky:
@zhyass please have a look.  need your help to review :

- if conflict detection mechanism needs to be tweaked as well 
     looks like no need to do that, but not sure if the schema is merged correctly (always use the latest schema in the snapshot meta file)
- if time travel implementation should be adjusted as well
    looks like no need to do that, since during time travel, the schema is switched as well.

BohuTANG:
> LGTM, ci need to run perf of this pr to see the performance issue.

Have a simple benchmark:
```
alter table hits_zstd add column x int null;
select * from hits_zstd ignore_result;
```

From the test, there is no performance issue on this case.

BohuTANG:
It's time to add the documentation for this cool feature, cc @lichuang

zhyass:
> * if conflict detection mechanism needs to be tweaked as well looks like no need to do that, but not sure if the schema is merged correctly (always use the latest schema in the snapshot meta file)

It seems to be no problem, but get the following errors after alter table. I will try to resolve them all together in #9873 

```
mysql> create table t(a int);

Query OK, 2 rows affected (0.04 sec)

mysql> alter table t add column b int;

mysql> update t set a=3 where a=1;
ERROR 1105 (HY000): Code: 1104, displayText = index out of bounds: the len is 2 but the index is 2.
```

> * if time travel implementation should be adjusted as well
>   looks like no need to do that, since during time travel, the schema is switched as well.

I'm not sure, it seems to be no problem.

Issue #2931, opened 2021-11-21
Tracking: Support DataType::Decimal


sundy-li:
**Summary**

Description for this feature.

Now we have introduced Decimal128 and Decimal256 types, let's add more features on decimal types.

- [x] #9894 
- [x] #9856 
- [x] #9926   supports number -> Decimal,  Decimal -> Decimal and arithmetic functions   @sundy-li 
- [x] #10214
- [ ] Group KeysU128 & KeysU256 will be stored as Decimal  @sundy-li 
- [x] #10059 
- [x] #10005
- [x] Add Decimal column stats to segment info (https://github.com/datafuselabs/databend/pull/10113)
- [ ] Decimal literal/encoding @youngsofun 
    - [x] Insert values https://github.com/datafuselabs/databend/pull/10167
         - [x] may fallback to expr of parser
    - [x] Insert/copy with other formats https://github.com/datafuselabs/databend/pull/10176
    - [ ] Cast from string or. allow string directly https://github.com/datafuselabs/databend/pull/10369
    - [x] Used decimal for number literal by default instead of float https://github.com/datafuselabs/databend/issues/10220
    - [x] deal with the float in json.   Float -> string -> decimal. or parse json by ourself
    - [x] binary format used inside.
- [x] #10095 

Below is old comments:

Decimal is Synonymous with Number type.

It represents a range up to 38 digits, with an optional precision and scale:

Precision:
Total number of digits allowed.

Scale:
Number of digits allowed to the right of the decimal point.

By default `Number` means `NUMBER(38,0)` or `Decimal(38, 0)`

|Width | Internal | Size (Bytes) |
|---|---|---|
| 1-4 | INT16 | 2 |
| 5-9 | INT32 | 4 |
| 10-18 | INT64 | 8 |
| 19-38 | INT128 | 16 |

liukun4515:
Hi @sundy-li, I have a question about decimal data type which will be implemented in the feature.
From the [datatype](https://github.com/datafuselabs/databend/blob/f1d0a942dd90ccb91f0949e5c26db94b58e215a1/common/datavalues/src/types/data_type.rs#L26), we know the native data type(i32,i64) is simple. when add the decimal, we need to consider the `precision` and `scale`, refine the data type coercion.

Does the community has the plan to implement it？

sundy-li:
How about making decimal's  physical_type to Int128 ? 
[...]

PR #9856, opened 2023-02-02
feat(expr): add Decimal128 and Decimal256 type

andylokandy:
- refactor
- feat(expr): add Decimal128 and Decimal256 type

## Summary

Summary about this PR

Related to https://github.com/datafuselabs/databend/issues/2931


BohuTANG:
@mergify update

src/meta/proto-conv/src/schema_from_to_protobuf_impl.rs
```diff
@@ -226,6 +226,7 @@ impl FromToProto for ex::TableDataType {
                 let x = n.to_pb()?;
                 new_pb_dt24(Dt24::NumberT(x))
             }
+            TableDataType::Decimal(_) => todo!("decimal"),
```
sundy-li, line 229: `unimplemented!  `  or `unreachable` is better.

Old version dt do not have this path.

src/query/expression/Cargo.toml
```diff
@@ -26,6 +26,7 @@ common-jsonb = { path = "../../common/jsonb" }
 educe = "0.4"
 enum-as-inner = "0.5"
 enum_dispatch = "0.3.8"
+ethnum = { version = "1.3", features = ["serde"] }
```
sundy-li, line 29: Seems this crate is better than `primitive_types`.

https://github.com/nlordell/ethnum-rs/blob/d8a2c42f7bd4c06d076a7bbcbaaaa948c67d2a22/bench/benches/num.rs

sundy-li (approved):
Generally Looks good to me.

src/query/functions/src/scalars/variant.rs
```diff
@@ -126,131 +124,6 @@ pub fn register(registry: &mut FunctionRegistry) {
         }),
     );
 
-    registry.register_passthrough_nullable_1_arg::<BooleanType, VariantType, _, _>(
```
sundy-li, line 129: why do we remove these functions?

andylokandy: Because it will be tedious to implement parse_json for decimals, and also because having only primitive numbers overload is inconsistent.

andylokandy:
@mergify update

andylokandy:
@mergify update

Issue #9847, opened 2023-02-02
Tracking: support `select ... from @<stage>`


youngsofun:
**Summary**

__Features__

- [x] basic select from parquet
- [x] `select ... from @<stage> ( <select_stage_options> )` <alias>
     - [x]  https://github.com/datafuselabs/databend/pull/9877
- [x] distributed query
  - [x] list file in read_partitions https://github.com/datafuselabs/databend/pull/9871
- [x]  https://github.com/datafuselabs/databend/issues/9908
- [x] https://github.com/datafuselabs/databend/issues/9974
- [x] doc https://github.com/datafuselabs/databend/pull/10208
- [x] options
     - [x]  file_format

__Bugs__
- [ ] https://github.com/datafuselabs/databend/issues/9884

__Related__

- [x] [create format clause](https://github.com/datafuselabs/databend/issues/8100), so we can use  format name in `<select_stage_options>`

__Optional__

- [ ] error tolerant (not necessary if not used in copy) 
- [ ] select from json(low priority)
- [ ] select from CSV / TSV(low priority)

youngsofun:
@BohuTANG @sundy-li 
do we need a `desc @stage/path` or  `infer_schema @stage/path`  like `desc table` ?
not know which columns can be used.

youngsofun:
is `select ... from '<uri>'`  necessary?
need to design a syntax to pass `connection = (...) `

BohuTANG:
> @BohuTANG @sundy-li @Xuanwo do we need a `desc @stage/path` or `infer_schema @stage/path` like `desc table` ? not know which columns can be used.

Better to have it, we can use it as:
1. For the cloud, we can infer the parquet schema
2. For debugging the schema

BohuTANG:
> is `select ... from '<uri>'` necessary? need to design a syntax to pass `connection = (...) `

Good idea, it is still better to have, but this task can be considered as low priority.

BohuTANG:
@youngsofun Add `desc @stage/path` to this tracking, really need it.

youngsofun:
since select from stage looks like a table function ( with @stage_name as function name) .
should we put the doc under [table-functions](https://databend.rs/doc/reference/functions/table-functions) too?

BohuTANG:
> since select from stage looks like a table function ( with `@stage_name` as function name) . should we put the doc under [table-functions](https://databend.rs/doc/reference/functions/table-functions) too?

Ok

PR #9871, opened 2023-02-03
refactor: ParquetTable list file in read_partition.

youngsofun:

## Summary

Summary about this PR

Closes #issue

part of https://github.com/datafuselabs/databend/issues/9847

### DIFF
--- a/src/common/storage/src/stage.rs
+++ b/src/common/storage/src/stage.rs
@@ -12,14 +12,35 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::path::Path;
+
+use common_exception::ErrorCode;
 use common_exception::Result;
 use common_meta_types::StageType;
 use common_meta_types::UserStageInfo;
+use futures::TryStreamExt;
 use opendal::layers::SubdirLayer;
+use opendal::ObjectMetadata;
+use opendal::ObjectMode;
 use opendal::Operator;
+use regex::Regex;
 
 use crate::init_operator;
 use crate::DataOperator;
+
+pub struct FileWithMeta {
+    pub path: String,
+    pub meta: ObjectMetadata,
+}
+
+impl FileWithMeta {
+    fn new(path: &str, meta: ObjectMetadata) -> Self {
+        Self {
+            path: path.to_string(),
+            meta,
+        }
+    }
+}
 
 pub fn init_stage_operator(stage_info: &UserStageInfo) -> Result<Operator> {
     if stage_info.stage_type == StageType::External {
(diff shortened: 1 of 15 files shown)


src/common/storage/src/stage.rs
```diff
@@ -29,3 +50,185 @@ pub fn init_stage_operator(stage_info: &UserStageInfo) -> Result<Operator> {
...
+            let pattern = self.get_pattern()?;
+            blocking_list_files_with_pattern(operator, &self.path, pattern, first_only)
+        }
+    }
+}
+
+async fn list_files_with_pattern(
+    operator: &Operator,
+    path: &str,
+    pattern: Option<Regex>,
```
RinChanNOWWW, line 154: Should we use `glob::Pattern` or  `Regex`?

`glob::Pattern` will follow Unix shell style, and it is different from regex.

For example: `*.parquet`,

- glob: will match all files with a suffix '.parquet'.
- regex: invalid.

src/common/storage/src/stage.rs
```diff
@@ -29,3 +50,185 @@ pub fn init_stage_operator(stage_info: &UserStageInfo) -> Result<Operator> {
...
+            }
+            Ok(res)
+        } else {
+            let pattern = self.get_pattern()?;
+            list_files_with_pattern(operator, &self.path, pattern, first_only).await
+        }
+    }
+
+    pub async fn first_file(&self, operator: &Operator) -> Result<FileWithMeta> {
+        let mut files = self.list(operator, true).await?;
```
RinChanNOWWW, line 104: If we should list all files at once and hold in memory? Since network IO is expensive.

src/query/storages/parquet/src/parquet_table/table.rs
```diff
@@ -149,9 +143,9 @@ pub(crate) fn arrow_to_table_schema(mut schema: ArrowSchema) -> TableSchema {
     TableSchema::from(&schema)
 }
 
-pub(super) fn create_parquet_table_info(table_id: u64, schema: ArrowSchema) -> TableInfo {
+pub(super) fn create_parquet_table_info(schema: ArrowSchema) -> TableInfo {
     TableInfo {
-        ident: TableIdent::new(table_id, 0),
+        ident: TableIdent::new(0, 0),
```
RinChanNOWWW, line 148: If it's safe to let the table id be 0? cc @sundy-li

RinChanNOWWW:
Need some tests of selecting from stage with pattern.

RinChanNOWWW (dismissed):
Need some tests of selecting from stage with pattern.

src/common/storage/src/stage.rs
youngsofun: FYI:  we used to use  Regex in copy
https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#optional-parameters

src/query/storages/parquet/src/parquet_table/table.rs
youngsofun: id of table_function and table in meta overlap too.
seem only useful for table in meta.

src/common/storage/src/stage.rs
youngsofun: here we list only one file (one list_dir op for local fs )
I think a pro of listing all file in read_partition (instead of binding->add_table) is we can report some syntax error earlier if read_partition take a long time.

is there other diff?

cc @BohuTANG  @Xuanwo  @sundy-li

youngsofun:
> Need some tests of selecting from stage with pattern.

Is it ok to do it in the next pr?

1.  add parser
2. add tests

src/common/storage/src/stage.rs
BohuTANG: What's the read path of parquet in this patch?

youngsofun: read the first file when binding to infer schema,
read all files in read_partition.

sundy-li: I think we should always do list since the result will change.

RinChanNOWWW (approved):

PR #9876, opened 2023-02-03
fix(query): fix slow planner when ndv error backtrace

zhang2014:

## Summary

fix(query): fix slow planner when ndv error backtrace

Closes #issue


zhang2014:
Maybe we need cache s_expr derive_relational_prop result. CC:@leiysky

PR #9877, opened 2023-02-03
feat: select from stage with files/pattern.

youngsofun:

## Summary

Summary about this PR

```
select * from @s2 (files => ('tuple.parquet'));
select * from @s2 (pattern => '.*parquet');
```

- `files` and `pattern` are the same with those in copy
    - except use `=> ` instead of `=` (learn from [snowflake](https://docs.snowflake.com/en/user-guide/querying-stage.html#example-2-calling-functions-when-querying-a-staged-data-file))
    - pattern use regex  as in copy

Closes #issue

part of https://github.com/datafuselabs/databend/issues/9847


BohuTANG:
It would be better to add more information for the PR, to make the reviewer/user clear what we did.

src/query/ast/src/ast/format/ast_format.rs
```diff
@@ -2199,15 +2199,19 @@ impl<'ast> Visitor<'ast> for AstFormatVisitor {
...
             } => {
                 let mut children = Vec::new();
-                if !files.is_empty() {
+                if let Some(files) = &options.files {
                     let files = files.join(",");
                     let files = format!("files = {}", files);
                     children.push(FormatTreeNode::new(AstFormatContext::new(files)))
                 }
+                if let Some(pattern) = &options.pattern {
+                    let pattern = format!("patter = {}", pattern);
```
BohuTANG, line 2212: `patter` -> `pattern`

BohuTANG (approved):

Issue #9865, opened 2023-02-02
grpc-api-address should not be necessary when restoring meta data.


drmingdrmer:
After the following issue being fixed, it does not have to specify grpc-api-address when restoring a meta-node:
- #9864

Currently, grpc-api-address is required:
https://databend.rs/doc/deploy/metasrv/metasrv-backup-restore#import-data-as-a-new-databend-meta-cluster

PR #9878, opened 2023-02-03
refactor(metactl): when restoring a cluster from backup data, it does not require to specify grpc addresses

drmingdrmer:

##### refactor(metactl): when restoring a cluster from backup data, it does not require to specify grpc addresses

- Fix: #9865 

## Related Issues


everpcpc (approved):
Thanks!

PR #9883, opened 2023-02-03
fix(query): remove finished flag for avoid lost block

zhang2014:

## Summary

fix(query): remove finished flag for avoid lost block

Closes #issue


sundy-li (approved):

BohuTANG:
`Dev Linux / test_sqllogic_query_cluster_linux ` test is hang

PR #9893, opened 2023-02-04
chore(storage): refactor cache to moka based

sundy-li:

## Summary

Summary about this PR

The original cache is using RWLock on the API level (too high), it'll acquire the write lock even in read operation(refresh the lru), this may have performance issue.

Performance:  https://github.com/moka-rs/moka/wiki#admission-and-eviction-policies

Closes #issue


BohuTANG:
This need a test, from my test , moka is not good.

Issue #9849, opened 2023-02-02
bug: delete from d2 where id in (select id from d1) panic


BohuTANG:
**Summary**

```
mysql> create table d2(id int);
Query OK, 0 rows affected (0.11 sec)

mysql> insert into d1 values(1),(2);
Query OK, 2 rows affected (0.15 sec)

mysql> insert into d2 values(1),(2),(3),(4);
Query OK, 4 rows affected (0.11 sec)

mysql> delete from d2 where id in (select id from d1);
ERROR 1105 (HY000): Code: 1068, displayText = Cannot join handle from context's runtime, cause: task 867229 panicked.
```

server log:
```
try/src/github.com-1ecc6299db9ec823/tokio-1.23.1/src/runtime/task/harness.rs", line: 198 }, { fn: ""tokio"::"runtime"::"task"::"harness"::"Harness<T",
"S>"::"poll"", file: "/root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.23.1/src/runtime/task/harness.rs", line: 152 }, { fn: ""tokio"::"runtime"::"task"::"raw"::"RawTask"::"poll"", file: "/root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.23.1/src/runtime/task/raw.rs", line: 200 }, { fn: ""tokio"::"runtime"::"task"::"LocalNotified<S>"::"run"", file: "/root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.23.1/src/runtime/task/mod.rs", line: 459 }, { fn: ""tokio"::"runtime"::"scheduler"::"multi_thread"::"worker"::"Context"::"run_task"::{
   {
      "closure"
   }
[...]

PR #9902, opened 2023-02-05
fix: prevent panic when `delete` with subquery

xudong963:

## Summary

When `delete` with subquery, leverage  `SubqueryRewriter` to flatten subquery to join.

Then make flatten SExpr be `DeletePlan` child.

The ticket only changed planner layer, needs more work to support `delete` with subquery in exectuor layer

Closes https://github.com/datafuselabs/databend/issues/9849


src/query/sql/src/planner/plans/delete.rs
```diff
@@ -25,6 +26,8 @@ pub struct DeletePlan {
     pub database_name: String,
     pub table_name: String,
     pub selection: Option<ScalarExpr>,
+    // The case: selection is subquery
+    pub input_expr: Option<SExpr>,
```
xudong963, line 30: It's also needed by `delete` with using.

Such as `DELETE FROM films USING producers
  WHERE producer_id = producers.id AND producers.name = 'foo';`

`Using` will be transferred to cross join, and `input_expr` will be cross join

xudong963:
@mergify update

BohuTANG (approved):

BohuTANG:
We need a logic test for this PR

PR #9903, opened 2023-02-06
feat(query): support array aggregate functions

b41sh:

## Summary

support array aggregate functions
```
array_sum
array_avg
array_count
array_max
array_min
```

part of #7931

### DIFF
--- a/src/query/functions/src/scalars/array.rs
+++ b/src/query/functions/src/scalars/array.rs
@@ -19,6 +19,8 @@ use std::sync::Arc;
 use common_expression::types::array::ArrayColumnBuilder;
 use common_expression::types::boolean::BooleanDomain;
 use common_expression::types::nullable::NullableDomain;
+use common_expression::types::number::Float64Type;
+use common_expression::types::number::Int64Type;
 use common_expression::types::number::SimpleDomain;
 use common_expression::types::number::UInt64Type;
 use common_expression::types::ArgType;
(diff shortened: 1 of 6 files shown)


src/query/functions/src/scalars/array.rs
```diff
@@ -606,4 +613,381 @@ pub fn register(registry: &mut FunctionRegistry) {
             }
         }),
     );
+
+    for left in ALL_SIGNED_INTEGER_TYPES {
+        with_number_mapped_type!(|NUM_TYPE| match left {
+            NumberDataType::NUM_TYPE => {
```
sundy-li, line 619: can we use `register_factory` and agg combinators?

The DataType could be evaluated as `agg.input_type()`.
Thus we could make the code path cleaner.

b41sh: The `return_type` is related to the type of the aggregate operation, and the type of the operation needs to be obtained by the second argument, so the `return_type` cannot be known in `FunctionSignature`.

src/query/functions/src/scalars/array.rs
```diff
@@ -614,380 +613,142 @@ pub fn register(registry: &mut FunctionRegistry) {
...
-                            output.push(val);
-                        }
-                        Err(err) => {
-                            ctx.set_error(output.len(), err.to_string());
-                        }
-                    }
-                } else {
-                    output.push_null();
-                }
+    registry.register_function_factory("array_max", |_, args_type| {
```
sundy-li, line 727: can we use for loop to register these array_agg functons ?

sundy-li (approved):

src/query/functions/src/scalars/array.rs
```diff
@@ -61,12 +64,25 @@ use itertools::Itertools;
 use siphasher::sip128::Hasher128;
 use siphasher::sip128::SipHasher24;
 
+use crate::aggregates::eval_aggr;
+use crate::AggregateFunctionFactory;
+
+const ARRAY_AGGREGATE_FUNCTIONS: &[(&str, &str); 5] = &[
+    ("array_avg", "avg"),
```
sundy-li, line 71: const ARRAY_AGGREGATE_FUNCTIONS =  ["avg", "min", "max", ...];

and use `concat!("arr_", agg)`

PR #9904, opened 2023-02-06
refactor(query): hashtable grow instead of grow_zero

sundy-li:

## Summary

Summary about this PR

Since `jemalloc` did not support realloc_zero, we use `realloc` instead. And introduce an additional `zeros: Vec<bool>` to indicate whether the slot is available.

Performance improved around 5%.

Closes #issue


sundy-li:
It's hard to control the states, closing.

PR #9920, opened 2023-02-07
chore(query): introduce TzLUT util

sundy-li:

## Summary

Summary about this PR

Add a util struct `TzLUT` which provides fast datetime helper functions.

4 times Faster in this query
```
SELECT extract(minute FROM EventTime) AS m, count(1) from hits group by m order by count(1) desc limit 10;
```

Closes #issue


src/query/expression/src/utils/date_helper.rs
```diff
@@ -16,17 +16,205 @@ use chrono::Date;
...
 
 use crate::types::date::check_date;
 use crate::types::timestamp::check_timestamp;
+use crate::types::timestamp::MICROS_IN_A_SEC;
+
+#[derive(Debug, Clone, Copy)]
+pub struct TzLUT {
+    pub tz: Tz,
+    pub offset_round_hour: bool,
+    pub offset_round_minute: bool,
```
RinChanNOWWW, line 41: Can add some comments to introduce the principle of this optimization.

RinChanNOWWW (approved):

BohuTANG (approved):

PR #9926, opened 2023-02-08
feat(decimal): add operation for decimal

sundy-li:

## Summary

Summary about this PR

1. Support  arithmetic for decimal types
2. Support new datatype in meta.
3. Support casting from number/decimal to decimal.

Closes #issue


drmingdrmer:
Change to protobuf based message requires a new version and a new test:
https://github.com/datafuselabs/databend/blob/9975a7d09a0c88de90ca3a1256624115e509b8cb/src/meta/proto-conv/src/util.rs#L85-L91

src/meta/proto-conv/src/schema_from_to_protobuf_impl.rs
```diff
@@ -322,6 +328,80 @@ impl FromToProto for ex::types::NumberDataType {
...
+        p.ver
+    }
+
+    fn from_pb(p: pb::Decimal) -> Result<Self, Incompatible> {
+        reader_check_msg(p.ver, p.min_reader_ver)?;
+
+        let num = match p.decimal {
+            None => {
+                return Err(Incompatible {
+                    reason: "Invalid Number: .num can not be None".to_string(),
```
drmingdrmer, line 344:
```suggestion
                    reason: "Invalid Number: .decimal can not be None".to_string(),
```

tests/sqllogictests/suites/base/11_data_type/11_0006_data_type_decimal
```diff
@@ -0,0 +1,114 @@
...
+
+## parser overflow
+statement error 1005
+select 2::Decimal(2000, 3);
+
+
+## runtime overflow
+statement error 1001
+select 20000::Decimal(4, 3);
+
```
youngsofun, line 14: should not use code 1001 https://github.com/datafuselabs/databend/blob/main/src/common/exception/src/exception_code.rs#L55

sundy-li: This is runtime exception, any error may happen, we don't store the error codes during evaluation for performance issue.

See  `EvalContext:set_error`

src/query/expression/src/types/decimal.rs
```diff
@@ -63,10 +65,183 @@ pub struct DecimalSize {
...
+
 impl DecimalDataType {
-    pub fn from_size(_size: DecimalSize) -> Result<DecimalDataType> {
-        // todo!("decimal")
-        Err(ErrorCode::Unimplemented("decimal"))
+    pub fn from_size(size: DecimalSize) -> Result<DecimalDataType> {
+        if size.precision < 1 || size.precision > MAX_DECIMAL256_PRECISION {
+            return Err(ErrorCode::Overflow(format!(
+                "Decimal precision must be between 1 and {}",
+                MAX_DECIMAL128_PRECISION
```
youngsofun, line 166:
```suggestion
                MAX_DECIMAL256_PRECISION
```

tests/sqllogictests/suites/base/11_data_type/11_0006_data_type_decimal
sundy-li: Will have another pr to use another error code.

src/query/expression/src/types/decimal.rs
```diff
@@ -63,10 +65,183 @@ pub struct DecimalSize {
...
+    ) -> Result<Self> {
+        let mut scale = a.scale().max(b.scale());
+        let mut precision = a.max_result_precision(b);
+
+        let multiply_precision = a.precision() + b.precision();
+        let divide_precision = a.precision() + b.scale();
+
+        // for addition/subtraction, we add 1 to the width to ensure we don't overflow
+        let plus_min_precision =
+            (a.precision() - a.scale()).max(b.precision() - b.scale()) + scale + 1;
```
youngsofun, line 224: for +/- , 
it is the Integer part (left part) that may overflow and need plus one.
the precision is not related to overflow (or over flow to the int part).
that is  the precision is max(a.p, b.p)
but the scale is  max(a.s, b.s) + 1

in snowflake`s [doc](https://docs.snowflake.com/en/sql-reference/operators-arithmetic.html#list-of-arithmetic-operators), let  L = S - P,  make it easier to understand

src/query/expression/src/types/decimal.rs
```diff
@@ -63,10 +65,183 @@ pub struct DecimalSize {
...
+
+        if is_multiply {
+            scale = a.scale() + b.scale();
```
youngsofun, line 227: need to check max scale?

src/query/expression/src/types/decimal.rs
```diff
@@ -63,10 +65,183 @@ pub struct DecimalSize {
...
+
+    pub fn binary_result_type(
+        a: &Self,
+        b: &Self,
+        is_multiply: bool,
+        is_divide: bool,
+        is_plus_minus: bool,
+    ) -> Result<Self> {
+        let mut scale = a.scale().max(b.scale());
+        let mut precision = a.max_result_precision(b);
```
youngsofun, line 217: rename it max_precision? and need a simliler max_scale = 38/76

src/query/expression/src/types/decimal.rs
sundy-li: We will check it in `from_size` function.

src/query/expression/src/types/decimal.rs
sundy-li: this is mutable `precision `  defaults to be `max_precision`, we will modify it via the operator.

src/query/expression/src/types/decimal.rs
sundy-li: Added `leading_digits` function.

> scale is max(a.s, b.s) + 1

Seems it's wrong.

youngsofun: sorry I am wrong,  I remenber precision as the Fractional part.  ignore me

youngsofun (approved):

drmingdrmer:
It still needs a test: 
image

PR #9933, opened 2023-02-08
Revert "fix(functions): use and_filters instead of and in predicate"

sundy-li:
Reverts datafuselabs/databend#9932


sundy-li:
Revert because :

`false and nullable(bool)` --> ` nullable(false)` which can't be folded into false.

cc @andylokandy

Issue #9908, opened 2023-02-06
Feature: table function `infer_schema`


youngsofun:
**Summary**

Description for this feature.

like [snow](https://docs.snowflake.com/en/sql-reference/functions/infer_schema.html)

PR #9936, opened 2023-02-09
feat(query): new table function infer_schema.

youngsofun:

## Summary

Closes https://github.com/datafuselabs/databend/issues/9908

currently only support named stage, with parquet format.

examples:

```
mysql> select * from infer_schema(location=> '@s2/data/tuple.parquet');
+-------------+---------------------+----------+----------+
| column_name | type                | nullable | order_id |
+-------------+---------------------+----------+----------+
| id          | INT                 |        0 |        0 |
| t           | (A INT32, B STRING) |        0 |        1 |
+-------------+---------------------+----------+----------+
```

also support parameter `pattern`

```
select * from infer_schema(location => '@s2/data/', pattern => '.*parquet');
```

## diff with `desc <table>`

-  it is table function
-  no `Default ` and `Extra`; add  `order_id` column
-  as a table function,  it's results is for processing, so `nullable`/ `order_id` column has type bool/uint64; while in `desc <table>`, all column is string

## diff with `copy`/ `select from stage`

1. currently,  arg value of table function must be const literal.  so  '@s2/data/' in the example is single quoted (as a string).
2. we currently only support infer the first file, so parameter `files` in not needed.


tests/suites/1_stateful/08_select_stage/08_00_parquet/08_00_02_select_files.sh
```diff
@@ -16,9 +16,12 @@ echo "create stage s2 url = '${DATADIR}' FILE_FORMAT = (type = PARQUET);"  | $MY
 cp "$CURDIR"/../../../../data/tuple.parquet ${DATADIR_PATH}/
 cp "$CURDIR"/../../../../data/sample.csv ${DATADIR_PATH}/
 
+echo "select * from infer_schema(location => '@s2/tuple.parquet');" | $MYSQL_CLIENT_CONNECT
```
BohuTANG, line 19: I prefer to add a single logic test for it, the progress like:
```
CREATE STAGE ss;
COPY INTO @ss FROM (SELECT number, number::VARCHAR FROM numbers(100)) FILE_FORMAT = (TYPE = PARQUET);
SELECT * FROM infer_schema(location => '@ss', pattern => '.*parquet')
```

youngsofun: ok

BohuTANG: We need more types built here(int and varchar is for example):
```
SELECT number, number::VARCHAR FROM numbers(100)
```

src/query/storages/fuse/src/table_functions/infer_schema/infer_schema_table.rs
```diff
@@ -0,0 +1,236 @@
+//  Copyright 2021 Datafuse Labs.
```
BohuTANG, line 1: Good from `2021` to `2023` :)

src/query/storages/fuse/src/table_functions/infer_schema/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod infer_schema_table;
```
BohuTANG, line 1: Need license header

youngsofun: fixed

src/query/storages/fuse/src/table_functions/infer_schema/infer_schema_table.rs
youngsofun: fixed

tests/suites/1_stateful/08_select_stage/08_00_parquet/08_00_04_infer_schema.result
```diff
@@ -0,0 +1,41 @@
+--- dir:
+id	INT	0	0
+t	(A INT32, B STRING)	0	1
+--- file:
+id	INT	0	0
+t	(A INT32, B STRING)	0	1
+--- pattern:
+id	INT	0	0
+t	(A INT32, B STRING)	0	1
+--- complex:
```
youngsofun, line 10: @BohuTANG  add a test with `complex.parquet`.

src/common/storage/src/parquet.rs
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common_arrow::arrow::datatypes::Schema as ArrowSchema;
+use common_arrow::arrow::io::parquet::read as pread;
+use common_exception::ErrorCode;
+use common_exception::Result;
+use opendal::Operator;
+
+pub async fn read_parquet_schema_async(operator: &Operator, path: &str) -> Result<ArrowSchema> {
```
BohuTANG, line 21: It seems we no need to add a new file for the `read_parquet_schema_async`

youngsofun: prepare to move more utils here,  such that mainly deps on only arrow and dal (but not context)

youngsofun: parquet table,  fuse, parquet copy(a simpler imply not need projection/pushdown)

BohuTANG (approved):

PR #9945, opened 2023-02-09
feat: support multiple jwt sources in databend query

ZhiHanZ:

## Summary
1. add additonal_jwt_key_files configuration
2. refactored authenticator to use multiple jwt sources instead of a single one.
3. add tests verifies the following behavior :

- [X] signed token could be verified and create user/role with additional_jwt_key_files, 
- [X] signed token could not be verified if the jwt_key_file is not in the support list

Closes #issue

### DIFF
--- a/src/query/config/src/inner.rs
+++ b/src/query/config/src/inner.rs
@@ -170,6 +170,7 @@ pub struct QueryConfig {
     /// If in management mode, only can do some meta level operations(database/table/user/stage etc.) with metasrv.
     pub management_mode: bool,
     pub jwt_key_file: String,
+    pub additional_jwt_key_files: Vec<String>,
     pub async_insert_max_data_size: u64,
     pub async_insert_busy_timeout: u64,
     pub async_insert_stale_timeout: u64,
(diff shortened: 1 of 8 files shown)


src/query/config/src/inner.rs
```diff
@@ -170,6 +170,7 @@ pub struct QueryConfig {
     /// If in management mode, only can do some meta level operations(database/table/user/stage etc.) with metasrv.
     pub management_mode: bool,
     pub jwt_key_file: String,
+    pub additional_jwt_key_files: Vec<String>,
```
flaneur2020, line 173: how about just calls it `jwt_key_files`? in most time we only need one of `jwt_key_file` or `jwt_key_files`.

Xuanwo:
Can you explain the use cases a little bit?

ZhiHanZ:
> Can you explain the use cases a little bit?

In cloud, jwks is wildly adopted by two categories:
1. External OAuth related callbacks
2. Internal OIDC connections

Generally it is not desired to use the same public key store for both external and internal communication, thus far, we need to add multiple sources to validate those tokens separately.

ZhiHanZ:
> There might have a gotcha which need to consider on using kubernetes' OIDC provider on authenticating users:
> 
> our JWT token is specialized with tenant identity, thus the authenticator can verify whether the user within a tenant (an jwt signed with tenant A can NEVER be authenticated into tenant B)
> 
> but taking the kubernetes OIDC jwt key to authenticate an user, we can verify the user comes from which pod, but not verify these tokens's tenant.

agreed, current jwt require special claims for user identity which is not an expected behavior for standard auth process, and need a standard OIDC interface to handle it

PR #9951, opened 2023-02-09
fix(shuffle): parts was not distributed evenly.

youngsofun:

## Summary

1. [fix reshuffle: parts was not distributed evenly.](https://github.com/datafuselabs/databend/commit/08f0623d4f4df701035fda23ded72a51d5a7daf7)
2.  [optimize reshuffle: avoid compute hash of each part for each compare.](https://github.com/datafuselabs/databend/commit/0924d17e7fc1a62a1b77f082e9eb97854e28c3b0)
1.  [feature: enable distributed query for ParquetTable.](https://github.com/datafuselabs/databend/commit/7ddacb7fd2e320f020859e27a87bd2aad9a850ae)

Closes #issue


src/query/catalog/src/plan/partition.rs
```diff
@@ -89,21 +89,22 @@ impl Partitions {
...
             PartitionsShuffleKind::Mod => {
                 // Sort by hash%executor_nums.
-                let mut parts = self.partitions.clone();
-                parts.sort_by(|a, b| {
-                    let hl = a.hash() % executor_nums as u64;
-                    let hr = b.hash() % executor_nums as u64;
-                    hl.cmp(&hr)
-                });
-                parts
+                let mut parts = self
```
youngsofun, line 100: @BohuTANG why do we need to `sort by hash%executor_nums.`

BohuTANG: It will make a part always send to the same node, so we can re-use the cache on the node.

src/query/catalog/src/plan/partition.rs
```diff
@@ -89,21 +89,22 @@ impl Partitions {
...
+                    .partitions
+                    .iter()
+                    .map(|p| p.hash() % num_executors as u64)
+                    .zip(self.partitions.clone().into_iter())
```
RinChanNOWWW, line 104:
```suggestion
                    .map(|p| (p.hash() % num_executors as u64, p.clone()))
```

RinChanNOWWW (approved):

zhang2014 (approved):

Issue #9972, opened 2023-02-10
bug: query sub-column of add tuple return 0


lichuang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

current

### What's Wrong?

query sub-column of add tuple return 0

### How to Reproduce?

```
drop database u;

create database u;

use u;

create table t17_0003(a int);

insert into t17_0003 values(1);

alter table t17_0003 add column b tuple(int, float) default (1,2.0);

select * from t17_0003;

select b.1,b.2 from t17_0003;
```

last query `select b.1,b.2 from t17_0003` should return `1 2.0`, but now return `0 0.0`.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #9973, opened 2023-02-10
fix: fix sub-column of added-tuple column return default 0 bug

lichuang:

## Summary

fix: fix sub-column of added-tuple column return default 0 bug

Closes https://github.com/datafuselabs/databend/issues/9972


BohuTANG:
There's no perf comments, cc @everpcpc

BohuTANG:
@everpcpc `ms ` should be `sec` in the perf result

BohuTANG:
From the result of the perf bot above, it seems there is no perf issue.

b41sh (approved):

BohuTANG (approved):

PR #9976, opened 2023-02-10
ci: disable incremental build in CI

xxchan:

## Summary

> [Incrementally compiled crates cannot be cached.](https://github.com/mozilla/sccache#rust)

@Xuanwo 

> Disable incremental compilation. CI builds often are closer to from-scratch builds, as changes are typically much bigger than from a local edit-compile cycle. For from-scratch builds, incremental adds an extra dependency-tracking overhead. It also significantly increases the amount of IO and the size of ./target, which make caching less effective.

https://matklad.github.io/2021/09/04/fast-rust-builds.html

### DIFF
--- a/.github/workflows/build-tool.yml
+++ b/.github/workflows/build-tool.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
     paths:
-      - "scripts/setup/**"
       - "docker/build-tool/**"
 
 permissions:
--- a/scripts/setup/run_build_tool.sh
+++ b/scripts/setup/run_build_tool.sh
@@ -33,7 +33,7 @@ done
 
 if [[ $ENABLE_SCCACHE == "true" ]]; then
 	env | grep -E "^SCCACHE_" >"${CARGO_HOME}/sccache.env"
-	EXTRA_ARGS="${EXTRA_ARGS} --env RUSTC_WRAPPER=/opt/rust/cargo/bin/sccache --env-file ${CARGO_HOME}/sccache.env"
+	EXTRA_ARGS="${EXTRA_ARGS} --env RUSTC_WRAPPER=/opt/rust/cargo/bin/sccach --env CARGO_INCREMENTAL=0 --env-file ${CARGO_HOME}/sccache.env"
 	COMMAND="${COMMAND} && sccache --show-stats"
 fi
 
--- a/scripts/setup/run_build_tool.sh
+++ b/scripts/setup/run_build_tool.sh
@@ -33,7 +33,7 @@ done
 
 if [[ $ENABLE_SCCACHE == "true" ]]; then
 	env | grep -E "^SCCACHE_" >"${CARGO_HOME}/sccache.env"
-	EXTRA_ARGS="${EXTRA_ARGS} --env RUSTC_WRAPPER=/opt/rust/cargo/bin/sccache --env CARGO_INCREMENTAL=0 --env-file ${CARGO_HOME}/sccache.env"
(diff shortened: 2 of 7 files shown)


xxchan:
How to trigger CI build? I'd like to test that 👀

xxchan:
Oops, just found this is already set *by change* due to a bug 😄

https://github.com/datafuselabs/databend/blob/6ac59a93205f92c1387d9d75f622bea14fad3307/Cargo.toml#L144-L148

So there are no improvements for dev build, but maybe it can help release build

xxchan:
Hmmm, I found sccache is disabled on release mode on main 🤔

Xuanwo:
> Hmmm, I found sccache is disabled on release mode on main

Thanks for the contribution. Disable sccache on main bacuase databend will run dev build on its own hosted runner on aws (with access to s3 bucket), and run release build on main on github hosted runner (which can't access s3 bucket).

scripts/setup/run_build_tool.sh
```diff
@@ -33,7 +33,7 @@ done
 
 if [[ $ENABLE_SCCACHE == "true" ]]; then
 	env | grep -E "^SCCACHE_" >"${CARGO_HOME}/sccache.env"
-	EXTRA_ARGS="${EXTRA_ARGS} --env RUSTC_WRAPPER=/opt/rust/cargo/bin/sccache --env-file ${CARGO_HOME}/sccache.env"
+	EXTRA_ARGS="${EXTRA_ARGS} --env RUSTC_WRAPPER=/opt/rust/cargo/bin/sccache --env CARGO_INCREMENTAL=0 --env-file ${CARGO_HOME}/sccache.env"
```
Xuanwo, line 36: Maybe better to set this in https://github.com/datafuselabs/databend/blob/main/.github/actions/build_linux/action.yml#L19?

xxchan: https://github.com/datafuselabs/databend/blob/72d3b00bc66cb1a123405f31aec3878faa61066a/.github/workflows/dev-linux.yml#L19-L21

What about setting it here? I think we can actually disable incremental build regardless of sccache.

xxchan: > Disable incremental compilation. CI builds often are closer to from-scratch builds, as changes are typically much bigger than from a local edit-compile cycle. For from-scratch builds, incremental adds an extra dependency-tracking overhead. It also significantly increases the amount of IO and the size of ./target, which make caching less effective.

https://matklad.github.io/2021/09/04/fast-rust-builds.html

Xuanwo: LGTM!

Xuanwo (approved):

.github/workflows/build-tool.yml
```diff
@@ -5,7 +5,6 @@ on:
     branches:
       - main
     paths:
-      - "scripts/setup/**"
```
everpcpc, line 8: Why this line remove ? We need to trigger build-tool rebuild when `scripts/setup/rust-toolchain.toml` and `scripts/setup/rust-tools.txt` changed.

everpcpc:
Actually we've already set `CARGO_INCREMENTAL=0` during build tool setup:
https://github.com/datafuselabs/databend/blob/main/.github/actions/setup_build_tool/action.yml#L36
So there's no need to add a line in every workflow file except `dev-macos.yml`.

.github/workflows/build-tool.yml
xxchan: Sorry, it's a mistake

everpcpc: Never mind, we can add it back next time upgrading toolchain.

Issue #9946, opened 2023-02-09
Feature: change default format of stage from CSV to Parquet


youngsofun:
**Summary**

`create stage`:

most of the time, a stage with a format is easier to use and maintain, so 

- use `parquet` if not specified when `create stage`
- explicitly set to None if user really want it.

user stage:

- not need create, 
- the default type is Parquet

`copy`/`select stage`/ infer_schema:

- use the file_format of stage if not specified
   - if the format of stage is None: 
- auto-detect the suffix if 
     -  file_format is not specified in the stmt, and  the format of stage is None: 
     -  file_format is set to None in the stmt

reason:  

- both used as load/unload, but parquet is more useful as storage for query, `select stage` and  `infer_schema` is mainly used for parquet
- CSV should be used more carefully: with correct options.  

related https://github.com/datafuselabs/databend/issues/9749

youngsofun:
cc @flaneur2020 does clients depends on the default value is csv.

PR #9983, opened 2023-02-11
feat: stage table function support option FILE_FORMAT.

youngsofun:

## Summary

```
select * from @stage1  (FILE_FORMAT => 'PARQUET')
```

Closes #issue

### DIFF
--- a/src/query/ast/src/parser/stage.rs
+++ b/src/query/ast/src/parser/stage.rs
@@ -213,5 +213,9 @@ pub fn select_stage_option(i: Input) -> IResult<SelectStageOption> {
             rule! { PATTERN ~ "=>" ~ #literal_string },
             |(_, _, pattern)| SelectStageOption::Pattern(pattern),
         ),
+        map(
+            rule! { FILE_FORMAT ~ "=>" ~ #literal_string },
+            |(_, _, file_format)| SelectStageOption::FileFormat(file_format),
+        ),
     ))(i)
 }
--- a/tests/suites/1_stateful/08_select_stage/08_00_parquet/08_00_00_basic.sh
+++ b/tests/suites/1_stateful/08_select_stage/08_00_parquet/08_00_00_basic.sh
@@ -7,7 +7,7 @@ echo "drop table if exists t1;" | $MYSQL_CLIENT_CONNECT
 echo "CREATE TABLE t1 (id INT, name VARCHAR, age INT);" | $MYSQL_CLIENT_CONNECT
 echo "insert into t1 (id,name,age) values(1,'2',3), (4, '5', 6);" | $MYSQL_CLIENT_CONNECT
 
-echo '--- stage'
+echo '--- named internal stage'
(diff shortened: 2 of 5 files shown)


BohuTANG:
So we can check if the `FILE_FORMAT` is parquet or not instead of checking the stage type(`type = PARQUET`)?

youngsofun:
yes， the format in select can overwrite that in the related stage，same as copy.

src/query/ast/src/parser/stage.rs
```diff
@@ -213,5 +213,9 @@ pub fn select_stage_option(i: Input) -> IResult<SelectStageOption> {
             rule! { PATTERN ~ "=>" ~ #literal_string },
             |(_, _, pattern)| SelectStageOption::Pattern(pattern),
         ),
+        map(
+            rule! { FILE_FORMAT ~ "=>" ~ #literal_string },
```
BohuTANG, line 217: It would be better to use enum instead of string? For example:
```
select * from @stage1  (FILE_FORMAT => PARQUET)
```

youngsofun: we will use this syntax in infer_schema too.
in general (and in snow) the value of `FILE_FORMAT` should be a format_name created by `create format`.
after discussion with @sundy-li, we choose to allow build-in format_name too and forbid use to create format parquet.

infer_schema is a table function,  in which a value of named params (after `=> `) need to be a expr.
while `PARQUET`(without quote) is an ident, can not be parsed here.

tests/suites/1_stateful/08_select_stage/08_00_parquet/08_00_00_basic.sh
```diff
@@ -26,4 +26,9 @@ echo "drop stage if exists s2;" | $MYSQL_CLIENT_CONNECT
 echo "create stage s2 url = '${DATADIR}' FILE_FORMAT = (type = PARQUET);"  | $MYSQL_CLIENT_CONNECT
 echo "select * from @s2;" | $MYSQL_CLIENT_CONNECT
 
+echo '--- file_format'
+echo "drop stage if exists s3;" | $MYSQL_CLIENT_CONNECT
+echo "create stage s3 url = '${DATADIR}' FILE_FORMAT = (type = CSV);"  | $MYSQL_CLIENT_CONNECT
+echo "select * from @s2 (FILE_FORMAT => 'PARQUET');" | $MYSQL_CLIENT_CONNECT
```
BohuTANG, line 32: The file type option is mainly for from URI not stage, right? stage has its type, we no need to constraint them too

youngsofun: https://github.com/datafuselabs/databend/issues/9946

BohuTANG: I see.

Issue #5626, opened 2022-05-27
Feat: catalog.drop_table_by_id


BohuTANG:
> Is there a consistency issue that `drop_table` at line 79 and `truncate()` at line 86 will take place on different tables?
> 
> _Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/pull/5619#discussion_r883231426_

------------------

- [ ] introduce `drop_table_by_id` method to catalog
    which ... drops a table by id :)
- [ ] refine the impl of `drop all`
    use `drop_table_byd_id` instread of drop table -> 
https://github.com/datafuselabs/databend/blob/797c1acb205ea7cdcdc37d5adabf97139580a135/query/src/interpreters/interpreter_table_drop.rs#L79
    just follow the `tbl.get_table_info`, and find the table's id

drmingdrmer:
> /assignme

Thanks man!

LiuYuHui:
Hi @dantengsky, I am confused about why drop_table at line 79 and truncate() at line 86 will take place on different tables? it seems like `drop_table` and `truncate` use the same table name, it should operate on the same table. Could you please help me understand this? Thanks a lot.

drmingdrmer:
> Hi @dantengsky, I am confused about why drop_table at line 79 and truncate() at line 86 will take place on different tables? it seems like `drop_table` and `truncate` use the same table name, it should operate on the same table. Could you please help me understand this? Thanks a lot.

When a table is dropped, another table with the same table can be created, in this interval of time, just before the second usage of this name.

LiuYuHui:
If we need to implement `drop_table_by_id` to catalog, it seems like we need to add `drop_table_by_id` to `SchemaApi`? or my understanding is not correct, there is an easy way to do this?

drmingdrmer:
> If we need to implement `drop_table_by_id` to catalog, it seems like we need to add `drop_table_by_id` to `SchemaApi`?

Yes:)

> or my understanding is not correct, there is an easy way to do this?

This seems to be the only way :(

johnhaxx7:
Hi @BohuTANG , seems like this is still open and no update, can you please help transfer to me if it's ok? Thanks!

BohuTANG:
@dantengsky do we still need this api now?

PR #9990, opened 2023-02-11
feat(storagte): Add drop_table_by_id api

johnhaxx7:

## Summary

Add `drop_table_by_id ` api for a race condition issue in `drop all`.  We were dropping table by name, which may be error prone when there's table with same name created before we actually dropped the data. This change is the part 1/2 to introduce the api, then I'll update `drop all` in a separate change.

The api is simply get tenant, database, table count and MISC info from table_id, so that we can get ready to emit the transaction. The transaction is exactly same as `drop_table`.

ref #5626

### DIFF
--- a/src/meta/api/src/schema_api_impl.rs
+++ b/src/meta/api/src/schema_api_impl.rs
@@ -111,6 +111,7 @@ use common_meta_types::MetaNetworkError;
 use common_meta_types::TxnCondition;
 use common_meta_types::TxnOp;
 use common_meta_types::TxnRequest;
+use common_meta_types::UnknownDatabaseId;
 use common_tracing::func_name;
 use tracing::debug;
 use tracing::error;
@@ -1143,7 +1144,7 @@ impl<KV: kvapi::KVApi<Error = KVAppError>> SchemaApi for KV {
                         txn_cond_seq(&dbid_tbname, Eq, tb_id_seq),
                         // table is not changed
                         txn_cond_seq(&tbid, Eq, tb_meta_seq),
-                        // update table count atomicly
+                        // update table count atomically
                         txn_cond_seq(&tb_count_key, Eq, tb_count_seq),
                     ],
                     if_then: vec![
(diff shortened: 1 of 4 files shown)


BohuTANG:
Hi @johnhaxx7 ,
What's this api purposes?

johnhaxx7:
Hi @BohuTANG , sorry I don't have enough context on this, I'm just skimming through the issue list and see what I can do. For this one my understanding is, the code piece mentioned in https://github.com/datafuselabs/databend/issues/5626 is drop my table name, which may be error prone in a race condition. I wrote up this change up front to make sure I'm able to finish it. To confirm on the issue, I think we need @drmingdrmer 's help.

drmingdrmer:
The issue https://github.com/datafuselabs/databend/issues/5626 is meant to address the consistency issue between a `drop-table` and a `truncated-table`, in my opinion.

The problematic use of `drop` and `truncate` by db-name, table-name should be replaced with `drop` and `truncate` by table-id:
image

johnhaxx7:
Hi, I updated it to `review ready`, please help take a look when you get a chance, thanks!

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1783,6 +1784,147 @@ impl<KV: kvapi::KVApi<Error = KVAppError>> SchemaApi for KV {
...
+                    )));
+                }
+
+                tb_meta.drop_on = Some(Utc::now());
+
+                let txn_req = TxnRequest {
+                    condition: vec![
+                        // db has not to change, i.e., no new table is created.
+                        // Renaming db is OK and does not affect the seq of db_meta.
+                        txn_cond_seq(&DatabaseId { db_id }, Eq, db_meta_seq),
```
drmingdrmer, line 1891: IMHO, dropping a table does not involve any changes to its belonging database.
Is it necessary to assert the database-meta not changed?

drmingdrmer (commented):
Thank you!

One question about the db/table.

src/meta/api/src/schema_api_impl.rs
johnhaxx7: Thanks for taking a look! I think it's is. It's from https://github.com/datafuselabs/databend/pull/5186/files. create/drop/rename a table will cause a db change. Otherwise, if the table is renamed after we receive the request and before we commit the drop transaction, some metadata we get may be staled. Then the question is, do we still want to proceed on this request when there's a table update on db?

drmingdrmer: Oh right. there is a table-name list for every db.

drmingdrmer (approved):

BohuTANG (approved):
Thank you for the contribution!

Issue #8100, opened 2022-10-10
feat: support `CREATE FILE FORMAT`


youngsofun:
**Summary**

- [x] create/drop/show
- [x] check when
- [x] use in 
    - [x]  copy
    - [x] select stage
    - [x] infer_schema
- [ ]  https://github.com/datafuselabs/databend/issues/10035

https://docs.snowflake.com/en/sql-reference/sql/create-file-format.html

youngsofun:
clickhouse handler is no longger tested

PR #10009, opened 2023-02-13
feat: support create file format.

youngsofun:

## Summary

start of https://github.com/datafuselabs/databend/issues/8100

```
mysql> CREATE FILE FORMAT test_format type=csv  record_delimiter='\t' escape='\\';

mysql> show file formats;
+-------------+-------------------------------------------------------------------------------------------------------------------------+
| name        | format_options                                                                                                          |
+-------------+-------------------------------------------------------------------------------------------------------------------------+
| test_format | TYPE = CSV FIELD_DELIMITER = ',' RECORD_DELIMITER = '\t' QUOTE = '\"' ESCAPE = '\\' SKIP_HEADER = 0 NAN_DISPLAY = 'NaN' |
+-------------+-------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.03 sec)


note that in `show file formats`,  all options used by the format type is checked, set to default if not mentioned, and all printed in column `format_options`.
.
mysql> drop file format test_format;
```

- [x] add tests for  create/show/drop file format

Closes #issue


drmingdrmer (commented):
Update `META_CHANGE_LOG` in the `src/meta/proto-conv/src/util.rs`

drmingdrmer (approved):

ariesdevil:
Just a question: how about put different FileFormatPlan to an enum and executing them in one interpreter?

sundy-li (approved):

ariesdevil (approved):

ariesdevil:
> > Just a question: how about put different FileFormatPlan to an enum and executing them in one interpreter?↳
> 
> what `s the benefit?↳

Nothing much to gain after thinking about it, the design is now clear enough.

PR #10014, opened 2023-02-13
rfc: query result cache.

### LINKED ISSUE #10010: Tracking: query result cache
**Summary**

Query result cache can help us to get query results faster if the data is not changed for the query.

This is a tracking issue for query result cache.

## Basic
- [x] #10011
- [x] Write and read result from result cache. #10042
- [x] Support cache of multiple tables query. #10158
- [x] Use AST as the cache key. #10165
- [x] Store result cache as parquet. #10187

## Related Tables
- [x] System table: `system.query_cache`. #10159
- [x] Table function: `SELECT * FROM RESULT_SCAN(LAST_QUERY_ID(-2));` #10256

## Optional

- [ ] Garbage collection for result cache.
- [ ] Different compression methods for result cache.
- [ ] Support result cache for compound SQLs. (`INSERT SELECT`, `COPY FROM`, `CREATE SELECT`)

### LINKED ISSUE #10011: RFC: query result cache
**Summary**

Tracking issue for RFC of query result cache.

The RFC PR: #10014.

RinChanNOWWW:

## Summary

RFC of query result cache.

Closes #10011

### DIFF
--- a/docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
+++ b/docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
@@ -27,26 +27,27 @@ If we execute the full query pipeline every time, the cost may be every expensiv
 
 ## Detail design
 
+### Lifecyle of query result cache
+
+Each result cache has a time-to-live (TTL). Each access to the result cache will refresh the TTL .When the TTL is expired, the result cache will not be used any more.
+
+Besides TTL, when the underlying data is changed (we can infer this by snapshot ids, segment ids and partition locations), the result cache will also be invalidated.
+
 ### Result cache storage
 
 Databend uses key-value pairs to record query result caches. For every query, Databend will construct a key to represent the query, and store ralated information in the value.
 
 Databend will not store the query result directly in the key-value storage. Instead, Databend only stores the location of the result cache file in the value. The actual result cache will be stored in the storage layer (local fs, s3, ...).
 
-#### Key structure
+#### Key
 
-A query result cache key contains such information:
+Query result cache is indexed by its abstract syntax tree (AST). Databend serializes the AST to a string and use hashes it as the key.
 
-- `ast`: the AST of the query.
-- `locations`: the locations of partitions used in the query.
-
-Databend will use the information to construct the key. The algorithm is like:
+The key generation is like:
 
 ```rust
-let ast_str = serialize(ast);
-let part_str = join(sort(locations), ",");
-let raw_key = join([ast_str, part_str], ",");
-let key = "_cache" + hash256(raw_key);
+let ast_str = ast.to_string();
+let key = format!("_cache/{}/{}", tenant, hash(ast_str.as_bytes()));
 ```
 
 #### Value structure
@@ -61,6 +62,14 @@ pub struct ResultCacheValue {
     pub query_time: DateTime<Utc>,
     /// Time-to-live of this query.
     pub ttl: usize,
+    /// The snapshot ids of the underlying data.
+    pub snapshot_ids: Vec<String>,
+    /// The segment ids of the underlying data.
+    pub segment_ids: Vec<String>,
+    /// The partition locations of the underlying data.
+    pub partition_locations: Vec<String>,
+    /// The size of the result cache (bytes).
+    pub result_size: usize,
     /// The location of the result cache file.
     pub location: String,
 }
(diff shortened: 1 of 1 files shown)


docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
```diff
@@ -0,0 +1,137 @@
...
+```rust
+pub struct ResultCacheValue {
+    /// The query SQL.
+    pub sql: String,
+    /// The query time.
+    pub query_time: DateTime<Utc>,
+    /// Time-to-live of this query.
+    pub ttl: usize,
+    /// The location of the result cache file.
+    pub location: String,
```
BohuTANG, line 65: need a:
```
/// The size in bytes of this result
pub result_size: usize,
```

And if `result_size` is larger than a quota no need to sink to the cache.

docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
```diff
@@ -0,0 +1,137 @@
...
+- `ast`: the AST of the query.
+- `locations`: the locations of partitions used in the query.
+
+Databend will use the information to construct the key. The algorithm is like:
+
+```rust
+let ast_str = serialize(ast);
+let part_str = join(sort(locations), ",");
+let raw_key = join([ast_str, part_str], ",");
+let key = "_cache" + hash256(raw_key);
```
Xuanwo, line 49: Databend is a multi-tenant service. Do we need to add a tenant identifier in path here?

docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
RinChanNOWWW: I think we don't need to store `result_size` to meta.

docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
```diff
@@ -0,0 +1,137 @@
...
+
+`databend-meta` has a capability to store and query key-value pairs. Databend use it to store query result cache key-value pairs.
+
+#### Garbage collection
+
+Databend will cache every query result is query result cache is enabled. If the result cache is expired, the cache will not be used any more. To save the disk or object storage space, Databend needs a deamon thread to scan all the query cache periodically and remove the expired ones.
+
+### Related configurations
+
+- `enable_write_result_cache`: whether to cache the query results (default: true).
```
Xuanwo, line 79: This configuration is a bit confusing. `enable_write_result_cache` and `enable_read_result_cache` seem to be a pair, one for `insert` operations and the other for `read` operations. Maybe we only need `enable_result_cache`?

docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
BohuTANG: I think it's ok to stat the result cache file and check its size.

BohuTANG: > I think we don't need to store `result_size` to meta.

Another benefit is `result_size` will help the services to scan the cache files by `result_size desc` to remove them by TTL.

docs/doc/90-contributing/01-rfcs/20230213-query-result-cache.md
RinChanNOWWW: Ok. I split the two options because some other DBs does this. We can just keep one `enable_result_cache` now. 

We can consider splitting the config when there are user requirements.

rschu1ze:
Hi! There seems to be a significant overlap in terms of design, integration and naming of this RFC and the recently introduced [query cache in ClickHouse](https://github.com/ClickHouse/ClickHouse/pull/43797) (which I am the author of). Perhaps allow me to post the original [blog post](https://clickhouse.com/blog/introduction-to-the-clickhouse-query-cache-and-design) and [docs](https://clickhouse.com/docs/en/operations/query-cache/) for reference, thanks.

RinChanNOWWW:
Hi @rschu1ze, thanks for your comments and concerns. 

For some namings and functionalities, we did refer other databases like ClickHouse and snowflake, we will add references in this RFC later.

For the detail design and implementation, Databend is totally different from ClickHouse. For example, ClickHouse use a hash table (has limit) to store cache information and tolerate inconsistent. However, Databend use its meta service (doesn't have limit) to store cache information and doesn't allow inconsistent by default.

You can find the first version of implementation in #10010, and we didn't refer any codes of ClickHouse.

PR #10023, opened 2023-02-13
feat(planner): Support `EXPLAIN ANALYZE` statement to profile query execution

### LINKED ISSUE #9927: Feature: Support `EXPLAIN ANALYZE` statement to display query profile

leiysky:

## Summary

A naive implementation of `EXPLAIN ANALYZE` statement.

In this PR, we introduced a framework to help us profile queries.

The processing time of processors is collected and accumulated and will be displayed along with the corresponding SQL operator.

For example:
image
image

This metric is helpful in figuring out which SQL operator is critical in most scenarios.

More metrics will be added later.

Closes #9927

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/transform_cast_schema.rs
+++ b/src/query/service/src/pipelines/processors/transforms/transform_cast_schema.rs
@@ -68,11 +68,15 @@ where Self: Transform
                 }
             })
             .collect();
-        Ok(Transformer::create(input_port, output_port, Self {
-            func_ctx,
-            insert_schema,
-            exprs,
-        }))
+        Ok(ProcessorPtr::create(Transformer::create(
+            input_port,
+            output_port,
+            Self {
+                func_ctx,
+                insert_schema,
+                exprs,
+            },
+        )))
     }
 }
 
--- a/src/query/sql/src/executor/format.rs
+++ b/src/query/sql/src/executor/format.rs
@@ -18,6 +18,7 @@ use common_exception::Result;
 use common_expression::ConstantFolder;
 use common_expression::FunctionContext;
 use common_functions::scalars::BUILTIN_FUNCTIONS;
+use common_profile::ProfSpanSetRef;
 use itertools::Itertools;
 
 use super::AggregateFinal;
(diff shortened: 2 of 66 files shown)


src/query/sql/src/executor/format.rs
```diff
@@ -152,7 +171,14 @@ fn filter_to_format_tree(plan: &Filter, metadata: &MetadataRef) -> Result<Format
         children.extend(items);
     }
 
-    children.push(to_format_tree(&plan.input, metadata)?);
+    if let Some(prof_span) = prof_span_set.lock().unwrap().get(&plan.plan_id) {
+        let process_time = prof_span.process_time / 1000 / 1000; // milliseconds
+        children.push(FormatTreeNode::new(format!(
+            "total process time: {process_time}ms"
```
BohuTANG, line 177: Can we add `total process time` to `TableScan` as well?
This explains analyze the cost of 16.39sec, most cost is TableScan:
```
mysql> explain analyze SELECT * FROM hits WHERE URL LIKE '%google%' ORDER BY EventTime LIMIT 10;
+----------------------------------------------------------------------------------------------------------------------------------------+
| explain                                                                                                                                |
+----------------------------------------------------------------------------------------------------------------------------------------+
| Limit                                                                                                                                  |
| ├── limit: 10                                                                                                                          |
| ├── offset: 0                                                                                                                          |
| ├── estimated rows: 10.00                                                                                                              |
| ├── total process time: 0ms                                                                                                            |
| └── Sort                                                                                                                               |
|     ├── sort keys: [eventtime ASC NULLS LAST]                                                                                          |
|     ├── estimated rows: 33332499.00                                                                                                    |
|     ├── total process time: 157ms                                                                                                      |
|     └── Filter                                                                                                                         |
|         ├── filters: [like(hits.url (#13), "%google%")]                                                                                |
|         ├── estimated rows: 33332499.00                                                                                                |
|         ├── total process time: 4658ms                                                                                                 |
|         └── TableScan                                                                                                                  |
|             ├── table: default.hits.hits                                                                                               |
|             ├── read rows: 99997497                                                                                                    |
|             ├── read bytes: 15458917407                                                                                                |
|             ├── partitions total: 751                                                                                                  |
|             ├── partitions scanned: 751                                                                                                |
|             ├── pruning stats: [segments: <range pruning: 32 to 32>, blocks: <range pruning: 751 to 751, bloom pruning: 0 to 0>]       |
|             ├── push downs: [filters: [like(hits.url (#13), "%google%")], limit: NONE]                                                 |
|             └── estimated rows: 99997497.00                                                                                            |
+----------------------------------------------------------------------------------------------------------------------------------------+
22 rows in set (16.39 sec)
Read 99997497 rows, 75.78 GiB in 16.231 sec., 6.16 million rows/sec., 4.67 GiB/sec.
```

src/query/service/src/pipelines/processors/transforms/transform_cast_schema.rs
```diff
@@ -68,11 +68,15 @@ where Self: Transform
                 }
             })
             .collect();
-        Ok(Transformer::create(input_port, output_port, Self {
-            func_ctx,
-            insert_schema,
-            exprs,
-        }))
+        Ok(ProcessorPtr::create(Transformer::create(
```
BohuTANG, line 71: Move `ProcessorPtr::create` wrap to `Transformer::create` ？

src/query/storages/memory/src/memory_table.rs
```diff
@@ -233,7 +233,12 @@ impl Table for MemoryTable {
         _: AppendMode,
         _: bool,
     ) -> Result<()> {
-        pipeline.add_sink(|input| Ok(ContextSink::create(input, ctx.clone())))
+        pipeline.add_sink(|input| {
+            Ok(ProcessorPtr::create(ContextSink::create(
```
BohuTANG, line 237: Move `ProcessorPtr::create` wrap to `add_sink` or `ContextSink::create` internal? 
The intention of doing a wrap here is not too clear.

src/query/sql/src/executor/format.rs
leiysky: We can do that, but that will make a lot of modifications to the `Table::read_data`. 

This PR is kind of an experimental implementation, which won't involve the full features of query profiling.

By the way, I think we should measure the cost of the table scans with some I/O metrics instead of `process time` since they are always asynchronous.

src/query/service/src/pipelines/processors/transforms/transform_cast_schema.rs
leiysky: The constructor should returns `impl Processor` or `Box<dyn Processor>` so we can wrap it with `ProfileWrapper`.

`ProcessorPtr` is a high-level data struct, it's better to construct it as soon as we need it.

src/query/storages/memory/src/memory_table.rs
leiysky: Ditto.

zhang2014 (approved):

src/query/sql/src/executor/format.rs
BohuTANG: For I/O metrics, we can add more to https://github.com/datafuselabs/databend/blob/80b5d2d403e9aa31d72ed5b281121ce105bec58d/src/query/catalog/src/plan/pruning_statistics.rs#L16-L28

BohuTANG:
Add doc: https://github.com/datafuselabs/databend/pull/10027

PR #10034, opened 2023-02-14
chore(query): Use 16 sized partitioned hashtable for final aggregation

sundy-li:

## Summary

Summary about this PR

1. Use 16 sized partitioned hashtable for final aggregation
2. Improve fast_hash using `_mm_crc32_u64`
3. Add const `HIGH_BIT` in PartitionedHashMap that indicates how to find the bucket via index.

Performance of Q32 increased ~10%
Closes #issue


everpcpc:
result: https://repo.databend.rs/benchmark/clickbench/pr/10034/4172448986.html
(auto comment would fix later)

BohuTANG:
It looks no help?

main branch: Q32 5.63s
https://repo.databend.rs/benchmark/clickbench/pr/7220305151.html 

This PR: Q32 6.54s 
https://repo.databend.rs/benchmark/clickbench/pr/10034/4172448986.html

src/query/service/src/pipelines/processors/transforms/transform_convert_grouping.rs
```diff
@@ -447,10 +457,24 @@ struct MergeBucketTransform<Method: HashMethod + PolymorphicKeysHelper<Method> +
 
     input_block: Option<DataBlock>,
     output_blocks: Vec<DataBlock>,
+
+    state: MergeBucketTransformState<Method>,
 }
 
-impl<Method: HashMethod + PolymorphicKeysHelper<Method> + Send + 'static>
-    MergeBucketTransform<Method>
+enum MergeBucketTransformState<Method>
```
zhang2014, line 464: Need add comment(why we need this?)

src/query/service/src/pipelines/processors/transforms/group_by/aggregator_polymorphic_keys.rs
```diff
@@ -475,17 +474,50 @@ impl PolymorphicKeysHelper<HashMethodSerializer> for HashMethodSerializer {
...
+            tables.push(self.method.create_hash_table()?);
+        }
+
+        Ok(PartitionedHashMap::<_, BUCKETS_LG2, true>::create(tables))
+    }
+
+    pub fn create_final_hashtable(
+        &self,
+    ) -> Result<PartitionedHashMap<Method::HashTable, BUCKETS_LG2, false>> {
+        let buckets = (1 << BUCKETS_LG2) as usize;
```
zhang2014, line 507: Final BUCKETS_LG2?

zhang2014:
maybe we should need to test memory usage?

src/query/service/src/pipelines/processors/transforms/group_by/aggregator_polymorphic_keys.rs
sundy-li: `BUCKETS_LG2` is template here, it will be injected in construter

sundy-li:
> maybe we should need to test memory usage

The memory usage a little better than before, I just tested Q32.  Both used 10G memory.

1. It used 16-sized partitioned hashtable rather than a single hashtable, but the key size is same, so the memory is same
2. But this pr could be memory efficient because it'll iterator the inner hashtable to result the block and free the memory ASAP.
3. The `BucketAggregator` is streaming by buckets now so the downside operators could work (such as sorting after group by high cardinality columns)

BohuTANG:
> It looks no help?
> 
> main branch: Q32 5.63s https://repo.databend.rs/benchmark/clickbench/pr/7220305151.html
> 
> This PR: Q32 6.54s https://repo.databend.rs/benchmark/clickbench/pr/10034/4172448986.html

Bench it again by hand, the performance is a bit better than before.

sundy-li:
Let's hold this pr to draft.

PR #10040, opened 2023-02-14
fix(privilege): add privileges on system.one to PUBLIC by default

### LINKED ISSUE #10036: Bug: Permission denied, user 'testgatrpzwjo-0-cl1'@'%' requires SELECT privilege on 'default'.'system'.'one'. on executing SELECT 1
**Summary**

maybe `SELECT 1` do not need checking privieleges?

flaneur2020:

## Summary

This PR adds a SELECT privilege on system.one to the PUBLIC role, so the normal users can execute `SELECT 1` without hitting permission denied error.

Closes #10036

### DIFF
--- a/src/query/users/src/role_mgr.rs
+++ b/src/query/users/src/role_mgr.rs
@@ -81,14 +81,23 @@ impl UserApiProvider {
     // 2. PUBLIC, which have no any privilege by default, but every role contains the PUBLIC role.
     //    The data objects which owned by PUBLIC can be accessed by any role.
     pub async fn ensure_builtin_roles(&self, tenant: &str) -> Result<u64> {
-        let mut role_info = RoleInfo::new(BUILTIN_ROLE_ACCOUNT_ADMIN);
-        role_info.grants.grant_privileges(
+        let mut account_admin = RoleInfo::new(BUILTIN_ROLE_ACCOUNT_ADMIN);
+        account_admin.grants.grant_privileges(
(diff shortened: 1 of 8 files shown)


BohuTANG (approved):

src/query/users/src/role_mgr.rs
```diff
@@ -78,17 +79,26 @@ impl UserApiProvider {
...
         );
-        self.add_role(tenant, role_info, true).await?;
-        let role_info = RoleInfo::new(BUILTIN_ROLE_PUBLIC);
-        self.add_role(tenant, role_info, true).await
+        self.add_role(tenant, account_admin, true).await?;
+
+        let mut public = RoleInfo::new(BUILTIN_ROLE_PUBLIC);
+        public.grants.grant_privileges(
+            &GrantObject::Table(
+                "default".to_string(),
```
BohuTANG, line 95: We can define a system database list and each item has what's a role that can grant it, it easy to maintain.

flaneur2020: currently we only have two roles builtin `account_admin` and `public`, these two roles are both special:

1. `account_admin` can access every tables
2. `public` is supposed to not able to access any system table at all.

We translate `SELECT 1` to query an internal table `system.one`, this makes us have to grant a privilege on `system.one` for every user. However the query `SELECT 1` do not really touches any tables in the query semantics. So I guess `system.one` can be viewed as a special corner case?

PR #10047, opened 2023-02-15
chore: `transform_sort_merge` use sort instead of merge sort.

RinChanNOWWW:

## Summary

If use merge sort on multiple blocks, same contents will be `memcpy` logN + 1 times (1 for take by sort indices).

If we concat the blocks and sort the whole block, each block will only be take 1 + 1 times (1 for take by sort indices).

The memory consumption of these two methods are almost the same.


src/query/pipeline/transforms/src/processors/transforms/transform_sort_merge.rs
```diff
@@ -51,23 +46,13 @@ impl Compactor for SortMergeCompactor {
...
-        self.aborting.store(true, Ordering::Release);
-    }
-
     fn compact_final(&self, blocks: &[DataBlock]) -> Result<Vec<DataBlock>> {
         if blocks.is_empty() {
             Ok(vec![])
         } else {
-            let aborting = self.aborting.clone();
-            let aborting: Aborting = Arc::new(Box::new(move || aborting.load(Ordering::Relaxed)));
+            let block = DataBlock::concat(blocks)?;
```
sundy-li, line 53: We should only concat the sorted columns.

1. concat the sorted columns to a block
2. apply sort_indices to this block and get the indices
3. use `take_indices` to take from `blocks` and get the result

sundy-li:
> select number * rand() c ,  (number + 11)::String, (number + 22)::String, (number + 33)::String , (number + 1)::String, (number + 2)::String, (number + 3)::String  from numbers(10000000) order by c desc ignore_result;

A little better.

1.66s --> 1.60s

sundy-li (approved):

Issue #10020, opened 2023-02-13
bug: compact not work as expected with alter table


zhyass:
**Summary**

```
mysql> create table t(a int,b int);
Query OK, 0 rows affected (0.06 sec)

mysql> insert into t values(1,1),(2,2);
Query OK, 2 rows affected (0.03 sec)

mysql> alter table t drop column b;
Query OK, 0 rows affected (0.06 sec)

mysql> alter table t add column b int default 100;
Query OK, 0 rows affected (0.04 sec)

mysql> insert into t values(3,3),(4,4);
Query OK, 2 rows affected (0.04 sec)

mysql> select * from t;
+------+------+
| a    | b    |
+------+------+
|    1 |  100 |
|    2 |  100 |
|    3 |    3 |
|    4 |    4 |
+------+------+
4 rows in set (0.04 sec)
Read 4 rows, 32.00 B in 0.011 sec., 367.63 rows/sec., 2.87 KiB/sec.

mysql> optimize table t compact ;
Query OK, 4 rows affected (0.05 sec)
```

After compact, we get the new col_stats in segment as follow:
```
"col_stats": {
	"1": {
		"min": {
			"Number": {
				"Int32": 1
			}
		},
		"max": {
			"Number": {
				"Int32": 2
			}
		},
		"null_count": 0,
		"in_memory_size": 8,
		"distinct_of_values": 3
	},
	"2": {
		"min": {
			"Number": {
				"Int32": 3
			}
		},
		"max": {
			"Number": {
				"Int32": 4
			}
		},
		"null_count": 0,
		"in_memory_size": 8,
		"distinct_of_values": 0
	},
	"0": {
		"min": {
			"Number": {
				"Int32": 1
			}
		},
		"max": {
			"Number": {
				"Int32": 4
			}
		},
		"null_count": 0,
		"in_memory_size": 16,
		"distinct_of_values": 4
	}
}
```

We need to add processing logic to the compact.

zhyass:
https://github.com/datafuselabs/databend/pull/10051#issuecomment-1432331928

PR #10051, opened 2023-02-15
fix: ignore dropped column statistics by column id when reducing block statistics

lichuang:

## Summary

1. fix: ignore dropped column statistics by column id when reducing block statistics
2. add test case.

Closes https://github.com/datafuselabs/databend/issues/10020

### DIFF
--- a/src/query/expression/src/schema.rs
+++ b/src/query/expression/src/schema.rs
@@ -15,6 +15,7 @@
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use common_arrow::arrow::bitmap::Bitmap;
@@ -382,6 +383,10 @@ impl TableSchema {
         self.fields.remove(i);
 
         Ok(())
+    }
+
+    pub fn to_column_id_set(&self) -> HashSet<ColumnId> {
+        HashSet::from_iter(self.to_column_ids().iter().cloned())
     }
 
     pub fn to_column_ids(&self) -> Vec<ColumnId> {
--- a/src/query/service/src/interpreters/interpreter_table_add_column.rs
+++ b/src/query/service/src/interpreters/interpreter_table_add_column.rs
@@ -58,7 +58,6 @@ impl Interpreter for AddTableColumnInterpreter {
 
         if let Some(table) = &tbl {
             let table_info = table.get_table_info();
-            println!("add column table info: {:?}", table_info);
             if table_info.engine() == VIEW_ENGINE {
                 return Err(ErrorCode::TableEngineNotSupported(format!(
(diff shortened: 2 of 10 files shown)


BohuTANG (approved):

zhyass:
Thank you very much! 
But there may be another problem, the case of the added column is ignored here.
For example:
```
mysql> create table t(a int);

Query OK, 2 rows affected (0.12 sec)

mysql> alter table t add column b int default 0;

mysql> insert into t values(3,3),(4,4);
Query OK, 2 rows affected (0.03 sec)

mysql> select * from t;
+------+------+
| a    | b    |
+------+------+
|    1 |    0 |
|    2 |    0 |
|    3 |    3 |
|    4 |    4 |
+------+------+
4 rows in set (0.04 sec)
Read 4 rows, 32.00 B in 0.012 sec., 337.97 rows/sec., 2.64 KiB/sec.

mysql> select * from t where b=0;
+------+------+
| a    | b    |
+------+------+
|    1 |    0 |
|    2 |    0 |
+------+------+
2 rows in set (0.03 sec)
Read 2 rows, 16.00 B in 0.010 sec., 193.03 rows/sec., 1.51 KiB/sec.

mysql> optimize table t compact ;
Query OK, 4 rows affected (0.04 sec)

mysql> select * from t where b=0;
Empty set (0.04 sec)
Read 0 rows, 0.00 B in 0.010 sec., 0 rows/sec., 0.00 B/sec.
```

The statistics in segment after compact:
```
"col_stats": {
	"0": {
		"min": {
			"Number": {
				"Int32": 1
			}
		},
		"max": {
			"Number": {
				"Int32": 4
			}
		},
		"null_count": 0,
		"in_memory_size": 16,
		"distinct_of_values": 4
	},
	"1": {
		"min": {
			"Number": {
				"Int32": 3
			}
		},
		"max": {
			"Number": {
				"Int32": 4
			}
		},
		"null_count": 0,
		"in_memory_size": 8,
		"distinct_of_values": 3
	}
}
```
The minmax of column `b` isnot [3,4], it's [0,4].

BohuTANG:
I think it's not the compact issue, it's the query:
```
mysql> select * from t where b=0;
+------+------+
| a    | b    |
+------+------+
|    1 |    0 |
|    2 |    0 |
+------+------+
```
It should return empty? Because column b is added after the two values of a.

cc @sundy-li

BohuTANG:
> I think it's not the compact issue, it's the query:
> 
> ```
> mysql> select * from t where b=0;
> +------+------+
> | a    | b    |
> +------+------+
> |    1 |    0 |
> |    2 |    0 |
> +------+------+
> ```
> 
> It should return empty? Because column b is added after the two values of a.
> 
> cc @sundy-li

Oh, just test MySQL, I am wrong, please ignore this comment.

PR #10054, opened 2023-02-15
feat(storage): Use drop_table_by_id api in `drop all`

### LINKED ISSUE #5626: Feat: catalog.drop_table_by_id

johnhaxx7:

## Summary

Use `drop_table_by_id` introduced in https://github.com/datafuselabs/databend/pull/9990, to reduce the chance of duplicate table name race condition 

No new test added, it's covered by `tests/sqllogictests/suites/base/05_ddl/05_0001_ddl_drop_table_full`

Closes #5626


src/query/service/src/interpreters/interpreter_table_drop.rs
```diff
@@ -73,6 +72,11 @@ impl Interpreter for DropTableInterpreter {
                 let latest = tbl.as_ref().refresh(self.ctx.as_ref()).await?;
                 latest.truncate(self.ctx.clone(), purge).await?
             }
+        } else {
+            return Err(ErrorCode::TableInfoError(format!(
```
johnhaxx7, line 76: Hi @BohuTANG, I added this check because I think this branch in if condition is error prone. We don't know what will happen if Option<tbl> is empty, we just return  `Ok(PipelineBuildResult::create())`, which IMHO can be hard to debug in some edge cases. After I add this, a series of tests are failing, like https://github.com/datafuselabs/databend/actions/runs/4181126745/jobs/7242889308, can you pull in some one to explain a bit on, in which case tbl is empty? So that I can add some comment here to help explain the scenario.

johnhaxx7:
@drmingdrmer PTAL, thanks!

drmingdrmer (commented):
The codes looks good! Thank you!
Although, personally speaking, I'd not suggest adding more methods to the catalog. Deleting a table is quite a straight forward operation, thus get a handle to meta-service from context and then just call a meta-client method(SchemaApi) would be simple and leader.

BohuTANG:
> The codes looks good! Thank you! Although, personally speaking, I'd not suggest adding more methods to the catalog. Deleting a table is quite a straight forward operation, thus get a handle to meta-service from context and then just call a meta-client method(SchemaApi) would be simple and leader.

Agree.
Change the `drop_table(plan)` to `drop_table(table_id)`? @drmingdrmer

drmingdrmer:
> Agree. Change the `drop_table(plan)` to `drop_table(table_id)`? @drmingdrmer

Sounds good!

src/query/service/src/catalogs/default/database_catalog.rs
```diff
@@ -325,25 +325,7 @@ impl Catalog for DatabaseCatalog {
...
-            .await?
-        {
-            return self.immutable_catalog.drop_table(req).await;
-        }
-        self.mutable_catalog.drop_table(req).await
-    }
-
-    async fn drop_table_by_id(&self, tb_id: MetaId) -> Result<DropTableReply> {
-        let res = self.mutable_catalog.drop_table_by_id(tb_id).await?;
+        let res = self.mutable_catalog.drop_table(req).await?;
```
drmingdrmer, line 328: The method name would be better to be explicit about what it exactly does, by naming it `drop_table_by_id`

drmingdrmer (commented):
Glad to see codes being removed!

src/query/service/src/catalogs/default/database_catalog.rs
johnhaxx7: updated to `drop_table_by_id`, PTAL, thanks!

src/meta/api/src/schema_api.rs
```diff
@@ -102,7 +102,7 @@ pub trait SchemaApi: Send + Sync {
     /// It returns a list of (table-id, table-meta-seq, table-meta).
     async fn list_all_tables(&self) -> Result<Vec<(TableId, u64, TableMeta)>, KVAppError>;
 
-    async fn drop_table(&self, req: DropTableReq) -> Result<DropTableReply, KVAppError>;
+    async fn drop_table_by_id(&self, req: DropTableReq) -> Result<DropTableReply, KVAppError>;
```
drmingdrmer, line 105: The same convention applies to `DropTableReq -> DropTableByIdReq`.

BohuTANG:
`make lint` in your local will make the `Dev Linux / check` happy:)

drmingdrmer (approved):

BohuTANG (approved):
👍

johnhaxx7:
@mergifyio refresh

johnhaxx7:
@Mergifyio refresh

PR #10059, opened 2023-02-15
feat(query): add sum/avg for decimal types

sundy-li:

## Summary

Summary about this PR

1. Add sum agg for decimal types 
2. Add avg agg  for decimal types

Closes #issue


src/query/expression/src/types/decimal.rs
```diff
@@ -65,17 +66,22 @@ pub struct DecimalSize {
...
 
+    fn max_of_max_precision() -> Self;
+
     fn min_for_precision(precision: u8) -> Self;
     fn max_for_precision(precision: u8) -> Self;
 
     fn from_float(value: f64) -> Self;
 
     fn try_downcast_column(column: &Column) -> Option<(Buffer<Self>, DecimalSize)>;
+    fn try_downcast_builder<'a>(builder: &'a mut ColumnBuilder) -> Option<&'a mut Vec<Self>>;
```
TCeason, line 84: I think try_downcast_builder also need return DecimalSize like try_downcast_column

sundy-li: We just need to get the `Vec<_>` here. 
DecimalSize is already in DataType enum.

TCeason (approved):

PR #10092, opened 2023-02-16
docs: changes alter functions example to make the function body reflect the name and match other doc examples

cameronbraid:

## Summary

changes docs example to make the function body reflect the name and match other doc examples


BohuTANG (approved):
Thanks for your contribution!

PR #10111, opened 2023-02-17
fix(expr): allow auto cast from string and variant

andylokandy:

## Summary

- select '1' + 1
```
+-----------+
| ('1' + 1) |
+-----------+
| 2         |
+-----------+
```
- select parse_json('1') + 1
```
+-----------------------+
| (parse_json('1') + 1) |
+-----------------------+
| 2                     |
+-----------------------+
```
- select parse_json('"1"') + 1
```
+-------------------------+
| (parse_json('"1"') + 1) |
+-------------------------+
| 2                       |
+-------------------------+
```
- select parse_json('"foo"') + 1
```
+---------------------------+
| (parse_json('"foo"') + 1) |
+---------------------------+
| <null>                    |
+---------------------------+
```
- select 'foo' + 1
```
(1105, 'Code: 1001, displayText = unable to parse string to type `UInt8` because invalid digit found in string while evaluating function `to_uint8("foo")`.')
```


src/query/functions/tests/it/scalars/testdata/array.txt
```diff
@@ -7,22 +7,23 @@ output domain  : []
...
-checked expr   : array<T0=String><T0, T0>("a", to_string<UInt8>(1_u8))
-optimized expr : ["a", "1"]
-output type    : Array(String)
-output domain  : [{"1"..="a"}]
-output         : ["a", "1"]
+error: 
+  --> SQL:1:1
+  |
+1 | ['a', 1]
+  | ^^^^^^^^ unable to parse string to type `UInt8` because invalid digit found in string while evaluating function `to_uint8("a")`
```
b41sh, line 14: The arguments in array function should support cast from Integer to String.

src/query/functions/tests/it/scalars/testdata/array.txt
```diff
@@ -7,22 +7,23 @@ output domain  : []
...
+
+
+
+error: 
+  --> SQL:1:1
+  |
+1 | [-1, true]
+  | ^^^^^^^^^^ no overload satisfies `array(Int16, Boolean)`
```
b41sh, line 22: ditto

src/query/functions/tests/it/scalars/testdata/array.txt
andylokandy: I had a discussion with @sundy-li yesterday, and we think it might be better to make string as the bottom type, to avoid potential ambigiouty.

sundy-li: Yes, string is bottom type.

sundy-li (approved):

PR #10118, opened 2023-02-18
refactor: save default values in FuseTable instead of BlockReader

lichuang:

## Summary

1. refactor: add TableContext param in `Catalog::get_table`,`Catalog::get_table_by_info`
2. refactor: save default values in FuseTable instead of BlockReader

Closes #issue


src/query/catalog/src/catalog/interface.rs
```diff
@@ -95,14 +96,21 @@ pub trait Catalog: DynClone + Send + Sync {
     /// Table.
 
     // Build a `Arc<dyn Table>` from `TableInfo`.
-    fn get_table_by_info(&self, table_info: &TableInfo) -> Result<Arc<dyn Table>>;
+    // If `ctx` is not None, then will get table schema field default values when init `dyn Table`.
+    fn get_table_by_info(
+        &self,
+        ctx: Option<Arc<dyn TableContext>>,
```
BohuTANG, line 102: `TableContext` argument is Option looks strange, it's hard to understand when we need it.

BohuTANG: Is there another way to do it instead of adding the TableContext to the function?

sundy-li:
The default values will be parsed every time `catalog.get_table`.

BohuTANG:
@mergify update

Xuanwo:
> @mergify update

The workflow has been changed. Please `rebase` with latest main to avoid update failure.

src/query/catalog/src/catalog/interface.rs
lichuang: How about add other API like `load_table`, this will never init default values of table schema, while `get_table` will init defaul values, use different API name to distinguish between different functions.

>> Is there another way to do it instead of adding the TableContext to the function?

Currently, `parse_exprs` need `TableContext` param, that is the root reason of this change.

lichuang:
After discussion with  @sundy-li , this change would calculate the default value at each `FuseTable` initialization, which has a small performance impact, so turn off this pr.

PR #10153, opened 2023-02-22
refactor: load and run old test suites when test query-meta compatibility

drmingdrmer:

## Changelog

- Improvement

## Related Issues


BohuTANG:
test_compat works! 👍

PR #10158, opened 2023-02-22
feat(query): support multi-table query result cache.

RinChanNOWWW:

## Summary

Store the SHA-256 values of partitions of each `TableScan` of the query in the query context.

Use these SHA-256 values to check if the query result cache is valid.

Tracking in #10010 .

### DIFF
--- a/src/query/catalog/src/table_context.rs
+++ b/src/query/catalog/src/table_context.rs
@@ -84,8 +84,8 @@ pub trait TableContext: Send + Sync {
     fn get_partition(&self) -> Option<PartInfoPtr>;
     fn get_partitions(&self, num: usize) -> Vec<PartInfoPtr>;
     fn set_partitions(&self, partitions: Partitions) -> Result<()>;
-    fn set_paritions_sha(&self, sha: String);
-    fn get_partitions_sha(&self) -> Option<String>;
+    fn add_partitions_sha(&self, sha: String);
+    fn get_partitions_shas(&self) -> Vec<String>;
 
     fn attach_query_str(&self, kind: String, query: &str);
     fn get_query_str(&self) -> String;
--- a/src/query/storages/result_cache/src/read/reader.rs
+++ b/src/query/storages/result_cache/src/read/reader.rs
@@ -31,7 +31,7 @@ pub struct ResultCacheReader {
 
     operator: Operator,
     /// To ensure the cache is valid.
-    partitions_sha: String,
+    partitions_shas: Vec<String>,
 
     /// If true, the cache will be used even if it is inconsistent.
     /// In another word, `partitions_sha` will not be checked.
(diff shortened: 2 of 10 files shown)


src/query/catalog/src/table_context.rs
```diff
@@ -84,8 +84,8 @@ pub trait TableContext: Send + Sync {
     fn get_partition(&self) -> Option<PartInfoPtr>;
     fn get_partitions(&self, num: usize) -> Vec<PartInfoPtr>;
     fn set_partitions(&self, partitions: Partitions) -> Result<()>;
-    fn set_paritions_sha(&self, sha: String);
-    fn get_partitions_sha(&self) -> Option<String>;
+    fn add_partitions_sha(&self, sha: String);
```
BohuTANG, line 87: Great.

How will we handle the hash in cluster mode in the future?

RinChanNOWWW: The partitions SHAs in the context will not be changed when redistributing partitions. We can still get the complete partitions SHAs after building the query pipepline.

RinChanNOWWW: As we can see the logic tests also passed in `test_sqllogic_base_cluster_linux`.

BohuTANG: Each node of the cluster has a partially result cache?  Because the partition in the cluster is segment and then read the partitions on the node not the coordinator query node.

BohuTANG: So here we supposed that the segment hash is not changed the cache results is valid?  The partitions   stored in the meta is segments.

RinChanNOWWW: > Each node of the cluster has a partially result cache?

No. Result cache is only written by coordinator node after getting all the result blocks.

> So here we supposed that the segment hash is not changed the cache results is valid?

Yes.

It seems that the segment information will also be changed when modifications happen?

BohuTANG: Yes.
If the partitions changed, the segment will change too, the `segment -> partitions` relation is immutable.

BohuTANG (approved):

src/query/storages/result_cache/src/read/reader.rs
```diff
@@ -48,11 +48,11 @@ impl ResultCacheReader {
         let tenant = ctx.get_tenant();
         let key = gen_common_key(&sql);
         let meta_key = gen_result_cache_meta_key(&tenant, &key);
-        let partitions_sha = ctx.get_partitions_sha().unwrap();
+        let partitions_shas = ctx.get_partitions_shas();
```
sundy-li, line 51: Order is not stable.

RinChanNOWWW: Solved in #10165.

PR #10165, opened 2023-02-22
chore(query): use formatted AST to generate the result cache key.

RinChanNOWWW:

## Summary

Use formatted AST to generate the result cache key. This make SQLs with different cases can use the same cache.

For example:

```sql
select * from t;
SELECT * FROM t;
```

These two SQLs can use the same cache.

Tacking in #10010

### DIFF
--- a/src/query/service/src/interpreters/interpreter_copy_v2.rs
+++ b/src/query/service/src/interpreters/interpreter_copy_v2.rs
@@ -68,13 +68,14 @@ impl CopyInterpreterV2 {
         path: &str,
         query: &Plan,
     ) -> Result<PipelineBuildResult> {
-        let (s_expr, metadata, bind_context) = match query {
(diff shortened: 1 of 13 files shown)


ariesdevil (approved):

BohuTANG (approved):

BohuTANG:
The `tolerate_inconsistent_result_cache` setting can be moved after this PR? @RinChanNOWWW

sundy-li:
> tolerate_inconsistent_result_cache

It's disabled by default, if users want to have `tolerate_inconsistent_result_cache`, cache will work even data is modified.

RinChanNOWWW:
> The `tolerate_inconsistent_result_cache` setting can be moved after this PR? @RinChanNOWWW

 `tolerate_inconsistent_result_cache` is a feature.

PR #10167, opened 2023-02-22
feat: support insert decimal.

youngsofun:

## Summary

the first step of  `Decimal literal/encoding` in https://github.com/datafuselabs/databend/issues/2931

also fix the decimal display (tested in insert tests)

Closes #issue

### DIFF
--- a/src/query/expression/src/deserializations/decimal.rs
+++ b/src/query/expression/src/deserializations/decimal.rs
@@ -1,0 +1,373 @@
+// Copyright 2023 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::cmp::Ordering;
+
+use common_exception::ErrorCode;
+use common_exception::Result;
+use common_io::prelude::FormatSettings;
+
+use crate::types::decimal::Decimal;
+use crate::types::decimal::DecimalSize;
+use crate::types::DecimalDataType;
+use crate::Column;
+use crate::Scalar;
+use crate::TypeDeserializer;
+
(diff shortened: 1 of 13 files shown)


tests/sqllogictests/suites/base/11_data_type/11_0006_data_type_decimal
```diff
@@ -137,3 +137,30 @@ query I
...
+## insert
+
+statement ok
+drop table if exists t_decimal_4_2;
+
+statement ok
+create table t_decimal_4_2(a Decimal(4,2));
+
+statement ok
+insert into t_decimal_4_2 values (1.1), (0.1), (1.0), (1.1e1), (1.1e-1);
```
BohuTANG, line 150: We need more edge tests for decimal, include negative.

youngsofun: @BohuTANG  ok, currently more tests is in unit test (faster for debug/modify)

BohuTANG: logic test is high-level test.
For example, we can do it as:
1, what's the max decimal define we can do in Databend.
2, Give a defination, like the test we have what's the range of it.


If we have these, it's easy to write documents and understand the decimal type better.

src/query/expression/src/deserializations/decimal.rs
```diff
@@ -0,0 +1,373 @@
...
+        _rows: usize,
+        _format: &FormatSettings,
+    ) -> Result<()> {
+        todo!()
+    }
+
+    fn de_json(&mut self, value: &serde_json::Value, _format: &FormatSettings) -> Result<()> {
+        match value {
+            serde_json::Value::Number(n) => {
+                if n.is_i64() {
```
sundy-li, line 101: Seems it's wrong, we should care about the `precision` ? See `m_integer_to_decimal`

BohuTANG (approved):

src/query/expression/src/deserializations/decimal.rs
youngsofun: @sundy-li  https://github.com/datafuselabs/databend/pull/10167/files#diff-93bbc5f6f54e7d2224ea2ff14e7dcf663af5d93c401dab999039c144f3cbf8d9R238

sundy-li (approved):

src/query/expression/src/deserializations/decimal.rs
sundy-li: Ok, better to add a test to cover this bug.

PR #10191, opened 2023-02-23
feat: replace into statement

dantengsky:

## Summary

`REPLACE INTO test [(column, ..)] ON [CONFLICT] T VALUES | SUBQUERY `

e.g. 

~~~
CREATE TABLE test(c1 int, c2 int);
REPLACE INTO test ON  (c1) VALUES(1,2);
REPLACE INTO test ON(c1)select number, number from numbers(101);
REPLACE INTO test ON(c1) select * from @test_stage (pattern => '.*parquet');
~~~

for more examples https://github.com/datafuselabs/databend/blob/58d6659459b2537167b122b8839e908fdd3db0b2/tests/sqllogictests/suites/base/09_fuse_engine/09_0023_replace_into

NOTE:

-  multiple columns of `ON CONFLICT` not supported yet
   currently, trying to specify more than one column in the `ON [CONFLICT](...)` leads to error 1005
-  cluster key related logic not implemented yet
   currently, replacing data into a table with cluster key definition leads to error 4000
- distributed replace into no supported yet   
   replace into statements can be executed in a cluster deployment, but can not utilize multiple query nodes yet.

Closes #issue


src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -217,6 +217,18 @@ impl AccessChecker for PrivilegeAccess {
...
             }
+            Plan::Replace(plan) => {
+                session
+                    .validate_privilege(
+                        &GrantObject::Table(
+                            plan.catalog.clone(),
+                            plan.database.clone(),
+                            plan.table.clone(),
+                        ),
+                        UserPrivilegeType::Insert, // TODO also check update
```
BohuTANG, line 228: FYI:
```
To use MySQL REPLACE, you must have both the INSERT and DELETE privileges for the table.
```

dantengsky: thanks for reminding

BohuTANG (approved):

BohuTANG:
Let's continue the work

PR #10197, opened 2023-02-24
fix: query result cache can also be written if `ignore_result`.

### LINKED ISSUE #10196: bug: SQLs with `ignore_result` will fail if `enable_query_result_cache`.
**Summary**

```
MySQL [(none)]> select * from t;
+------+
| a    |
+------+
|    1 |
+------+
1 row in set (0.062 sec)

MySQL [(none)]> select * from t ignore_result;
Empty set (0.059 sec)

MySQL [(none)]> set enable_query_result_cache = 1;
Query OK, 0 rows affected (0.024 sec)

MySQL [(none)]> select * from t ignore_result;
ERROR 1105 (HY000): Code: 1001, displayText = Cannot duplicate empty pipe..
```

RinChanNOWWW:

## Summary

Closes #10196

### DIFF
--- a/src/query/service/src/interpreters/interpreter_select_v2.rs
+++ b/src/query/service/src/interpreters/interpreter_select_v2.rs
@@ -23,7 +23,9 @@ use common_pipeline_core::pipe::Pipe;
 use common_pipeline_core::pipe::PipeItem;
 use common_pipeline_core::processors::port::InputPort;
 use common_pipeline_core::processors::port::OutputPort;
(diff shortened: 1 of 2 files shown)


zhang2014:
Maybe no result cache if ignore_result is better.

sundy-li (approved):

sundy-li:
Yes, ignore result looks like we don't care about the result, so no cache is better!

ariesdevil (approved):

PR #10207, opened 2023-02-24
chore: only run subquery rewrite for sql contains subquery

xudong963:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/src/query/sql/src/planner/optimizer/s_expr.rs
+++ b/src/query/sql/src/planner/optimizer/s_expr.rs
@@ -187,34 +187,43 @@ fn find_subquery(rel_op: &RelOperator) -> bool {
         | RelOperator::DummyTableScan(_)
         | RelOperator::Pattern(_) => false,
         RelOperator::Join(op) => {
-            op.left_conditions
-                .iter()
-                .any(|expr| matches!(expr, ScalarExpr::SubqueryExpr(_)))
-                || op
-                    .right_conditions
-                    .iter()
-                    .any(|expr| matches!(expr, ScalarExpr::SubqueryExpr(_)))
-                || op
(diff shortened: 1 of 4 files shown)


src/query/sql/src/planner/optimizer/s_expr.rs
```diff
@@ -172,4 +167,56 @@ impl SExpr {
     pub(crate) fn applied_rule(&self, rule_id: &RuleID) -> bool {
         self.applied_rules.get(rule_id)
     }
+
+    /// Check if contain subquery
+    pub(crate) fn contain_subquery(&self) -> bool {
+        if !find_subquery(&self.plan) {
+            for child in self.children.iter() {
+                return child.contain_subquery();
```
sundy-li, line 175: early return inside a loop.

xudong963:
@mergify update

src/query/sql/src/planner/optimizer/s_expr.rs
```diff
@@ -171,9 +171,7 @@ impl SExpr {
     /// Check if contain subquery
     pub(crate) fn contain_subquery(&self) -> bool {
         if !find_subquery(&self.plan) {
-            for child in self.children.iter() {
-                return child.contain_subquery();
-            }
+            return self.children.iter().any(|child| child.contain_subquery());
```
sundy-li, line 174: should check if children are empty, it will return true.

sundy-li: !self.children.is_empty() && ...

xudong963: An empty iterator returns false

image

PR #10214, opened 2023-02-25
feat(query): add compare op for decimal types

sundy-li:

## Summary

Summary about this PR

### DIFF
--- a/src/query/catalog/src/plan/partition.rs
+++ b/src/query/catalog/src/plan/partition.rs
@@ -185,6 +185,7 @@ impl Default for Partitions {
 pub struct StealablePartitions {
     pub partitions: Arc<RwLock<Vec<VecDeque<PartInfoPtr>>>>,
     pub ctx: Arc<dyn TableContext>,
+    pub disable_steal: bool,
 }
 
 impl StealablePartitions {
@@ -192,7 +193,12 @@ impl StealablePartitions {
         StealablePartitions {
             partitions: Arc::new(RwLock::new(partitions)),
             ctx,
+            disable_steal: false,
         }
+    }
+
+    pub fn disable_steal(&mut self) {
+        self.disable_steal = true;
     }
 
     pub fn steal_one(&self, idx: usize) -> Option<PartInfoPtr> {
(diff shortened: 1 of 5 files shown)


src/query/catalog/src/plan/partition.rs
```diff
@@ -185,16 +185,22 @@ impl Default for Partitions {
 pub struct StealablePartitions {
     pub partitions: Arc<RwLock<Vec<VecDeque<PartInfoPtr>>>>,
     pub ctx: Arc<dyn TableContext>,
+    pub disable_steal: bool,
```
BohuTANG, line 188: This option is used for?

tests/sqllogictests/suites/base/11_data_type/11_0006_data_type_decimal
```diff
@@ -163,6 +163,31 @@ SELECT ANY(CAST(2.34 AS DECIMAL(6, 2)))
 ----
 2.34
 
+
+## compare
+
+query IIIII
+select a > b, a < b, a = b, a <= b, a >= b from (select 3::Decimal(13,2) a , 3.1::Decimal(8,2) b);
```
BohuTANG, line 170: Can we auto cast integer to decimal in the underlying?

sundy-li: Yes, check the following test.

TCeason (approved):
LGTM, now we can compare the same decimal type with diff size.

src/query/catalog/src/plan/partition.rs
sundy-li: It's to improvement `topk` queries, we should disable steal partitions from other threads.

If parts goes into topk in specific order, it could have stable and fastest performance.

BohuTANG: It would be better add some comments for it.

tests/sqllogictests/suites/base/11_data_type/11_0006_data_type_decimal
BohuTANG: I can not find it like: where a = 0.3 or a> 2

BohuTANG: I see :)

PR #10216, opened 2023-02-25
chore(query): make tpch queries run on decimal

sundy-li:

## Summary

Summary about this PR

Closes #issue


src/query/expression/src/evaluator.rs
```diff
@@ -168,30 +170,31 @@ impl<'a> Evaluator<'a> {
...
-                    )
-                    .1,
-                    None,
-                    "domain calculation should not return any domain for expressions that are possible to fail"
-                );
-                *RECURSING.lock().unwrap() = false;
-            }
-        }
-
+        // We can't call this in debug mode, because it will cause infinite recursion.
```
sundy-li, line 173: @andylokandy  do you know how to handle this ?

sundy-li: I comment on these codes now. Could be fixed in another issue.

xudong963 (approved):

PR #10226, opened 2023-02-26
chore(storage): limit the scan threads if there are small parts

sundy-li:

## Summary

Summary about this PR

Closes #issue


sundy-li:
Merged, it's only for native performance improvement in large cores.

PR #10227, opened 2023-02-26
chore(query): add debug assert that reproduces meta inconsistent bug of ALTER

sundy-li:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/src/query/storages/fuse/src/fuse_part.rs
+++ b/src/query/storages/fuse/src/fuse_part.rs
@@ -74,6 +74,12 @@ impl FusePartInfo {
         sort_min_max: Option<(Scalar, Scalar)>,
         range: Option<Range<usize>>,
     ) -> Arc<Box<dyn PartInfo>> {
+        debug_assert!(
+            columns_meta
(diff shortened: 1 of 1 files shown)


sundy-li:
There must be a terrible bug in alter table statement.

It will make column meta inconsistent and cause process panic.

lichuang (approved):

sundy-li:
@lichuang  This is not for merge, it proves that DDL had bugs.

lichuang:
Test DDL:
```
DROP TABLE IF EXISTS `05_0028_at_t3`;
CREATE TABLE `05_0028_at_t3`(a Array(UInt8));
INSERT INTO TABLE `05_0028_at_t3` values([101,2]);
```

after insert `Array` type values, the `total_rows` of array return `num_values`, which in there case is 2, so assert failed.But for a nested data type, it is OK.

sundy-li:
@lichuang  got that, maybe you can add more debug_assert for untested types. This could detect bugs in ci.

PR #10229, opened 2023-02-26
feat(query): set returning function `unnest`.

RinChanNOWWW:

## Summary

Introduce a new physical operator: `Unnest`.

While building physical plan, `EvalScalar` will be built into 

```
Unnest
|_____EvalScalar
          |_______Children
```

if there are `unnest` calls.

Array type data will be unnested in `Unnest` operator.

## Limits:

- Only one unnest is allowed now.

Tracking in #10295.


src/query/functions/src/srf/factory.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl SetReturningFunctionFactory {
+    pub fn create() -> Self {
+        Self {
+            func_creators: HashMap::new(),
+        }
+    }
+
+    pub fn instance() -> &'static SetReturningFunctionFactory {
```
sundy-li, line 48: better to use [ctor] which registers functions during compile.

leiysky:
This function is usually implemented as a `ProjectSet` operator, for example in postgres:
```
leiysky=# select unnest(array[1,2]), unnest(array[1]), unnest(array[1,2,3]);
 unnest | unnest | unnest
--------+--------+--------
      1 |      1 |      1
      2 |        |      2
        |        |      3
(3 rows)

leiysky=# explain select unnest(array[1,2]), unnest(array[1]), unnest(array[1,2,3]);
                   QUERY PLAN
------------------------------------------------
 ProjectSet  (cost=0.00..0.04 rows=3 width=12)
   ->  Result  (cost=0.00..0.01 rows=1 width=0)
(2 rows)
```

And they are only allowed to be placed in `SELECT` clause and `FROM` clause.

src/query/sql/src/evaluator/block_operator.rs
```diff
@@ -75,8 +84,131 @@ impl BlockOperator {
...
+        }
+
+        let (unnest_index, unnest_col) = &unnest_columns[0];
+        let (unnest_values, unnest_offsets) = (&unnest_col.values, &unnest_col.offsets);
+
+        let num_rows = unnest_values.len();
+        // Convert unnest_offsets to take indices.
+        let mut take_indices = Vec::with_capacity(num_rows);
+        let offset_len = unnest_offsets.len() - 1;
+        for i in 0..offset_len {
```
sundy-li, line 169: `iter().window(2)`

tests/sqllogictests/suites/query/02_function/02_0062_function_unnest
```diff
@@ -0,0 +1,97 @@
...
+select unnest(a) from t;
+----
+1
+2
+3
+4
+5
+
+query IT
+select unnest(a), a from t;
```
sundy-li, line 54: so `select  3* (unnest(a) + 1) from t` is not supported now ?

RinChanNOWWW: No, it's not supported now. It will be supported in the future.

`unnest` can only be the top of `ScalarExpr` now.

sundy-li (approved):

andylokandy (approved):

RinChanNOWWW:
@mergify update

Issue #10062, opened 2023-02-15
feat: Map data type


BohuTANG:
**Summary**

- [x] Add Map data type
- [x] Support Create the table with Map data type
- [x] Support insert into / streaming load Map values
- [x] Support Map get function
- [x] Support Map get function push down to storage
- [x] Add bloom index for the `map.values` to fast value `=` query, like `where map['ip'] = '192.168.0.7'`

PR #10238, opened 2023-02-27
feat(query): Support Map data type create table and insert values

b41sh:

## Summary

- Support creating a table with a `Map` data type
- Support Insert `Map` values

Part of #10062


sundy-li (approved):

youngsofun:
better if add test to check unloaded file can be load and is the same
  https://github.com/datafuselabs/databend/pull/10176/files

PR #10245, opened 2023-02-27
docs: lift the decimal type under the number type

BohuTANG:

## Summary

* Add more examples for decimal data type, including min and max
* Lift the decimal order

Closes #issue

### DIFF
--- a/docs/doc/13-sql-reference/10-data-types/index.md
+++ b/docs/doc/13-sql-reference/10-data-types/index.md
@@ -7,36 +7,37 @@ slug: ./
 Databend supports SQL data types in several categories:
 * [Boolean Data Types](00-data-type-logical-types.md)
 * [Numeric Data Types](10-data-type-numeric-types.md)
+* [Decimal Data Types](11-data-type-decimal-types.md)
 * [Date & Time Data Types](20-data-type-time-date-types.md)
 * [String Data Types](30-data-type-string-types.md)
-* [Decimal Data Types](43-data-type-decimal-types.md)
 * [Array(T) Data Types](40-data-type-array-types.md)
 * [Tuple Data Types](41-data-type-tuple-types.md)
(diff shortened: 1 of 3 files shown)


docs/doc/13-sql-reference/10-data-types/index.md
```diff
@@ -7,36 +7,37 @@ slug: ./
...
 Databend supports three Semi-structured types: ARRAY, OBJECT and VARIANT.
 
-| Name    |    Aliases   | Build From Values       | Description
-|---------|--------------|-------------------------|----------------
-| ARRAY   |              | [1,2,3]                 | Zero-based indexed list, each value can have difference data type.
-| TUPLE   |              | ('2023-02-14 08:00:00','Valentine's Day') | Collection of ordered,immmutable, which requires the type of each element to be declared before being used.
-| VARIANT |  JSON        | [1,{"a":1,"b":{"c":2}}] | Collection of elements of different data types., including ARRAY and OBJECT.
+| Name    | Aliases | Build From Values                         | Description                                                                                                 |
+|---------|---------|-------------------------------------------|-------------------------------------------------------------------------------------------------------------|
+| ARRAY   |         | [1,2,3]                                   | Zero-based indexed list, each value can have difference data type.                                          |
```
sundy-li, line 40: VARIANT is the only SEMI struct datatype. `OBJECT` should be removed in the docs, let's use `VARIANT` instead.

sundy-li:
```
Zero-based indexed list, each value can have difference data type.
```

It's wrong,  we have `Array<T>` . It's inner type is fixed.

BohuTANG: Ok, let me remove it.

sundy-li: Also `Tuple` is not semi struct datatype, it's a nested datatype.

Issue #10259, opened 2023-02-28
Upgrade openraft from 0.7.4 to 0.8.0


drmingdrmer:
Openraft 0.8:
https://github.com/datafuselabs/openraft/releases/tag/v0.8.0

It provides more clean APIs and fixed bugs. And test coverage is improved: from 88% to 91%

PR #10261, opened 2023-02-28
refactor(meta): move AppliedState from crate raft-store to meta-types

drmingdrmer:

##### refactor(meta): move AppliedState from crate raft-store to meta-types

This type has to be declared as a raft response type in next PR.
It has to be in the same crate as `LogEntry` etc.

## Related Issues


drmingdrmer:
- This is part of #10259

PR #10296, opened 2023-03-02
refactor(meta): upgrade openraft from v0.7 to v0.8

drmingdrmer:

## Summary

This upgrade is backward compatible with older persistent data. But can not be run together with old meta-service. The network protocol format changed.

To upgrade, shutdown all meta server and restart them.

##### refactor(meta): upgrade openraft from v0.7 to v0.8

Openraft v0.8 provides better API for applications.

- Move logic to `Engine`, which is a event-driven non-async mod, in order to make logic more clear and tests easier: [#291](https://github.com/datafuselabs/openraft/issues/291)
- Migrate to a more generic API design: `NodeId` and `Node` are no longer concrete types but a trait, so it is possible for an application to use any node-id type or node type.
- Make the consensus implementation more generic: it supports both multi-leader per term mode and single-leader per term mode: [feature flag single-term-leader](https://github.com/datafuselabs/openraft/commit/4a85ee936856ce6bfde4c8cc6c58e15b00971257)

Openraft v0.8 change log:
https://github.com/datafuselabs/openraft/blob/release-0.8/change-log.md

- Fix: #10259

## Related Issues


BohuTANG:
> But can not be run together with old meta-service. The network protocol format changed.

If we rolling update the metasrv what will happen?

drmingdrmer:
> > But can not be run together with old meta-service. The network protocol format changed.
> 
> If we rolling update the metasrv what will happen?

Network errors will flood the log file, but rest assured, no data will be lost.

BohuTANG (approved):

Issue #10295, opened 2023-03-02
Tracking: support set returning function `unnest`.


RinChanNOWWW:
**Summary**

`unnest` is a function to 'flatten' array type elements. It accepts a scalar value but returns a vector.

Some use cases:

```sql
select unnest([1,2,3]);
+-------------------+
| unnest([1, 2, 3]) |
+-------------------+
|                 1 |
|                 2 |
|                 3 |
+-------------------+

select unnest([1,2,3]), number from numbers(2);
+-------------------+--------+
| unnest([1, 2, 3]) | number |
+-------------------+--------+
|                 1 |      0 |
|                 2 |      0 |
|                 3 |      0 |
|                 1 |      1 |
|                 2 |      1 |
|                 3 |      1 |
+-------------------+--------+

select unnest(arr), arr from ta;
+-------------+---------+
| unnest(arr) | arr     |
+-------------+---------+
|           1 | [1,2]   |
|           2 | [1,2]   |
|           3 | [3,4,5] |
|           4 | [3,4,5] |
|           5 | [3,4,5] |
+-------------+---------+

select * from unnest([1,2,3]);
+-------------------+
| unnest([1, 2, 3]) |
+-------------------+
|                 1 |
|                 2 |
|                 3 |
+-------------------+

```

## Tasks

- [x] Basic implementation. #10229
- [x] Support multiple `unnest` columns. #10312
- [x] `unnest` table function. #10357
- [x] Advanced `unnest` operations. (Such as `select unnest(a) + 1;`, `select unnest([[1,2],[3,4]]);`. #10329

## References

- [Unnest - PostgreSQL](https://www.postgresql.org/docs/9.2/functions-array.html)
- [Unnest - DuckDB](https://duckdb.org/docs/sql/query_syntax/unnest.html)
- [Unnest - Presto](https://prestodb.io/docs/current/sql/select.html#unnest)
- [Even Faster Unnest - Presto](https://prestodb.io/blog/2020/08/20/unnest)
- [arrayJoin - ClickHouse](https://clickhouse.com/docs/en/sql-reference/functions/array-join/)

PR #10312, opened 2023-03-02
feat(query): support multiple unnest columns.

RinChanNOWWW:

## Summary

If the original length of each column is different, the shorter ones will be padded with NULLs.

For example:

```
+---------+---------+
| origin1 | origin2 |
+---------+---------+
| [1,2]   | [1,2,3] |
| [3,4,5] | [4,5]   |
+---------+---------+
 ```

will be unnested to:

```
+---------+---------+
| unnest1 | unnest2 |
+------ --+---------+
| 1       | 1       |  -- From original row 1
| 2       | 2       |  -- From original row 1
| NULL    | 3       |  -- From original row 1
| 3       | 4       |  -- From original row 2
| 4       | 5       |  -- From original row 2
| 5       | NULL    |  -- From original row 2
+---------+---------+
```

Tracking in #10295

### DIFF
--- a/src/query/expression/src/values.rs
+++ b/src/query/expression/src/values.rs
@@ -1362,6 +1362,20 @@ impl Column {
         }
     }
 
+    pub fn wrap_nullable(self) -> Self {
+        match self {
+            col @ Column::Nullable(_) => col,
+            col => {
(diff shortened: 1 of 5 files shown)


sundy-li:
@mergify update

sundy-li (approved):

PR #10275, opened 2023-03-01
feat: parse decimal in expr.

youngsofun:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

trying to split this pr, a drawback is that effect of some changes need this pr to test.

- https://github.com/datafuselabs/databend/pull/10343 vip
- https://github.com/datafuselabs/databend/pull/10338
- https://github.com/datafuselabs/databend/pull/10339
- https://github.com/datafuselabs/databend/pull/10340
- https://github.com/datafuselabs/databend/pull/10341
- https://github.com/datafuselabs/databend/pull/10345

## Summary

- [x] choose the minimal type for numeric literal, large number try to fit into decimal (may truncate trailing digits after the point),  if overflow, use float64
- [x] ` 1 + -1` result in  `plus(1, -1)` instead of `plus(1,  minus(1))`
- [x] add auto cast rule  Decimal->Float32/64
- [x] to_variant(decimal) result in number.

notice:

- [x] agg func like std(1.0),  manually cast in test for now.

not in this pr

- decimal to int (not auto)
- string to decimal

Closes https://github.com/datafuselabs/databend/issues/10220 
https://github.com/datafuselabs/databend/issues/10262

PR #10326, opened 2023-03-03
fix(function): use decimal hash function before float.

youngsofun:

## Summary

this is a rescue for  https://github.com/datafuselabs/databend/pull/10275, should be merged before it.
use a separate pr to make review easier.
only fix one place to avoid conflict with https://github.com/datafuselabs/databend/pull/10275).

Closes #issue


src/query/functions/tests/it/scalars/testdata/hash.txt
```diff
@@ -310,11 +310,11 @@ output         : 12447382539015599056
 
 ast            : siphash64(1.1)
 raw expr       : siphash64(1.1_f64)
-checked expr   : siphash64<Float64>(1.1_f64)
-optimized expr : 14856193058643670405_u64
+checked expr   : siphash64<Decimal(38, 0)>(to_decimal<Float64>(1.1_f64))
```
youngsofun, line 313:
```
raw expr : siphash64(1.1_f64)
```

the 1.1_f64 should be decimal，if fixed in https://github.com/datafuselabs/databend/pull/10275

sundy-li (approved):

youngsofun:
oops, lead to other failurse, fix them will lead to more fail...
so let`s use this pr for review only.

youngsofun:
work well in https://github.com/datafuselabs/databend/pull/10275

PR #10313, opened 2023-03-02
chore(expr): respect overload order for function factories

andylokandy:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR

Closes #issue

PR #10343, opened 2023-03-04
fix(function): fix function register order

youngsofun:

## Summary

- [x] fix decimal divide by zero
- [x] fix cmp of decimal scalar
- [x] add decimal to_float32
- [x] add decimal to_string
- [x] add decimal unary minus
- [x] fix stack overflow ( Decimal->Float32/64 is factory function, which is tried first, before that,  those rules lead to tack overflow  ).   
        - [x]  remove functions like float32_to_float32
        - [x]  func like "to_float32(flao64)" should not be used  auto cast function
 - [x]  auto cast to decimal before float,
     - [x]   function factory respect order https://github.com/datafuselabs/databend/pull/10313
     - [x]  change orders
         - [x]  hash
         - [x] arithmetic
         - [x]  compare
         - [x] to string
         - [x] to float32/float64

Closes #issue


TCeason (approved):

youngsofun:
@sundy-li  need to change decimal in schema of tpch test back to double 
change back after decimal -> float is ready

youngsofun:
> @sundy-li need to change decimal in schema of tpch test back to double change back after decimal -> float is ready current result is based on float->decimal, which is wrong.

instead of revert schema，
https://github.com/datafuselabs/databend/pull/10343/commits/f36dda68a17d4b9ec0b74b29b39877ff7adc59a5

src/query/expression/src/type_check.rs
```diff
@@ -487,7 +487,9 @@ pub fn can_auto_cast_to(
                 .zip(dest_tys)
                 .all(|(src_ty, dest_ty)| can_auto_cast_to(src_ty, dest_ty, auto_cast_rules))
         }
-        (DataType::Number(_), DataType::Decimal(_)) => true,
+        (DataType::Number(n), DataType::Decimal(_)) => {
+            !matches!(n, NumberDataType::Float32 | NumberDataType::Float64)
+        }
```
andylokandy, line 492:
```suggestion
        (DataType::Number(n), DataType::Decimal(_)) if !n.is_float() => true,
```

andylokandy (approved):

youngsofun:
@mergify update

PR #10368, opened 2023-03-06
fix: make `PushDownPrewhere` as a rule.

RinChanNOWWW:

## Summary

`PrewhereOptimizer` is no longer existing. Instead, it will become a rule.

Notice that `PushDownPrewhere` should be after all other rules except `PushDownSortScan`.

After this modification, we don't need to do optimization twice any more. This will also fix the failure of TPC-H Q19 for native format.

### DIFF
--- a/src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
+++ b/src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
@@ -20,7 +20,6 @@ use once_cell::sync::Lazy;
 
 use super::prune_unused_columns::UnusedColumnPruner;
 use crate::optimizer::heuristic::decorrelate::decorrelate_subquery;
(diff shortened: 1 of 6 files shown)


sundy-li (approved):

dusx1981:
Please check what is related to DEFAULT_REWRITE_RULES when HeuristicOptimizer applies rules.###

dusx1981:
### Please check what is related to DEFAULT_REWRITE_RULES when HeuristicOptimizer applies rules.

dusx1981:
I'm investigating whether there is a way to remove DEFAULT_REWRITE_RULES.

RinChanNOWWW:
> I'm investigating whether there is a way to remove DEFAULT_REWRITE_RULES.

Maybe @xudong963 has some idea.

PR #10369, opened 2023-03-06
feat: support function to_decimal(string).

youngsofun:

## Summary

the last pr of  https://github.com/datafuselabs/databend/issues/2931
sorry for delay.

- [x] add some unit tests 

Closes https://github.com/datafuselabs/databend/issues/2931


src/query/expression/src/type_check.rs
```diff
@@ -517,6 +517,7 @@ pub fn can_auto_cast_to(
                 .all(|(src_ty, dest_ty)| can_auto_cast_to(src_ty, dest_ty, auto_cast_rules))
         }
         (DataType::Number(n), DataType::Decimal(_)) if !n.is_float() => true,
+        (DataType::String, DataType::Decimal(_)) => true,
```
sundy-li, line 520: I think this rule should be forbidden.  We will not allow string auto-cast to decimal, even string to float is not allowed.

sundy-li (approved):
Rest lgtm

src/query/expression/src/type_check.rs
youngsofun: ok

youngsofun: fixed

src/query/functions/src/scalars/decimal.rs
```diff
@@ -468,13 +473,72 @@ fn convert_to_decimal(
...
-        return float_to_decimal(arg, ctx, from_type, dest_type);
+}
+
+fn string_to_decimal_internal<T: Decimal>(
+    ctx: &mut EvalContext,
+    string_column: &StringColumn,
+    size: DecimalSize,
+    dest_type: &DecimalDataType,
+) -> DecimalColumn {
+    let mut builder = DecimalDeserializer::<T>::with_capacity(dest_type, string_column.len());
```
b41sh, line 496: better to use `DecimalColumnBuilder`.

src/query/functions/src/scalars/decimal.rs
```diff
@@ -468,13 +473,72 @@ fn convert_to_decimal(
...
+    dest_type: DataType,
+) -> Value<AnyType> {
+    let dest_type = dest_type.as_decimal().unwrap();
+
+    let mut is_scalar = false;
+    let column = match arg {
+        ValueRef::Column(column) => column.clone(),
+        ValueRef::Scalar(s) => {
+            is_scalar = true;
+            let builder = ColumnBuilder::repeat(s, 1, &from_type);
```
b41sh, line 522: Don't need to convert `Scalar` to a `Column`, we can get the value from both `Scalar` and `Column` by function `index_unchecked` or `index`.

src/query/functions/tests/it/scalars/testdata/array.txt
```diff
@@ -1756,7 +1756,7 @@ error:
   --> SQL:1:12
   |
 1 | array_sort(['x', 0, 1.2, 3.4, 5.6, 7.8])
-  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unable to cast type `String` to type `Decimal(2, 1)`
+  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fail to decode string as decimal: Code: 1006, displayText = bad decimal literal: unexpected char. while evaluating function `to_decimal("x")`
```
b41sh, line 1759: Please also add some unit tests for decimal functions.

youngsofun: this pr is about to_decimal(string), 

since we do not allow auto cast from string to decimal, we only need to test manual cast ?  there is a simple one in this pr.  since the core function invoked is  the same as parsing a decimal expr, I did not repeate these tests.

which kind of additional tests do you suggest?

src/query/functions/src/scalars/decimal.rs
youngsofun: you are right,  I just followed the impl of  prev similar functions. 
I will change it.

src/query/functions/src/scalars/decimal.rs
youngsofun: ok

src/query/functions/tests/it/scalars/testdata/array.txt
b41sh: like this `test_to_string` function.
https://github.com/datafuselabs/databend/blob/main/src/query/functions/tests/it/scalars/arithmetic.rs#L108

sundy-li: this pr should not change this file I think,  if string auto cast to decimal is not allowed.

youngsofun: but  string auto cast to  int/float, now, you can try:

```
mysql> select '1' + 11;
+------------+
| ('1' + 11) |
+------------+
|         12 |
+------------+
1 row in set (0.04 sec)
Read 1 rows, 1.00 B in 0.004 sec., 247.45 rows/sec., 247.45 B/sec.

mysql> select '1' + 11.1::float32;
+-----------------------+
| ('1' + 11.1::float32) |
+-----------------------+
|    12.100000381469727 |
+-----------------------+
```

youngsofun:
```
fn register_auto_cast_rules(registry: &mut FunctionRegistry) {
    registry.register_default_cast_rules(GENERAL_CAST_RULES.iter().cloned());
    registry.register_default_cast_rules(CAST_FROM_STRING_RULES.iter().cloned());
```

youngsofun: but CAST_FROM_STRING_RULES does not allow string to decimal of float.

src/query/functions/src/scalars/decimal.rs
youngsofun: I think we mainly optimize of the Column( not the Scalar) case.
it is slower to braching for each iterm?

```
    pub unsafe fn index_unchecked(&'a self, index: usize) -> T::ScalarRef<'a> {
        match self {
            ValueRef::Scalar(scalar) => scalar.clone(),
            ValueRef::Column(c) => T::index_column_unchecked(c, index),
        }
    }
``

File: src/query/functions/tests/it/scalars/testdata/array.txt (hunk 4 above)
Comment (youngsofun) [reply src/query/functions/tests/it/scalars/testdata/array.txt:1759]: > this pr should not change this file I think, if string auto cast to decimal is not allowed.

the super type of  ```['x', 0, 1.2, 3.4, 5.6, 7.8]```  is decimal.

since  string -> unit8 -> decimal is auto

when executing,  try string-> decimal, and got the error

```
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fail to decode string as decimal: Code: 1006, displayText = bad decimal literal: unexpected char. while evaluating function `to_decimal("x")`
```

Comment (sundy-li) [reply src/query/functions/tests/it/scalars/testdata/array.txt:1759]: Oh, I see that now.

File: src/query/functions/src/scalars/decimal.rs (hunk 2 above)
Comment (b41sh) [reply src/query/functions/src/scalars/decimal.rs:496]: This way we can use `Vec<T>` to store the decimal value directly and the `DecimalDeserializer` is not needed

**Comment by youngsofun** [comment]
> You can modify `CAST_FROM_STRING_RULES` to have `String` to `Decimal`.

add `String` to `Float` too?

File: src/query/functions/src/scalars/decimal.rs (hunk 2 above)
Comment (b41sh) [reply src/query/functions/src/scalars/decimal.rs:496]: This way we can use `Vec<T>` to store the decimal value directly and the `DecimalDeserializer` is not needed

Comment (b41sh) [reply src/query/functions/src/scalars/decimal.rs:496]: This way we can use `Vec<T>` to store the decimal value directly and the `DecimalDeserializer` is not needed

File: src/query/functions/src/scalars/decimal.rs (hunk 3 above)
Comment (b41sh) [reply src/query/functions/src/scalars/decimal.rs:522]: `Scalar` only has one item, so it's ok to clone. You can refer to the implementation of this function.
https://github.com/datafuselabs/databend/blob/main/src/query/functions/src/scalars/array.rs#L121

Comment (b41sh) [reply src/query/functions/src/scalars/decimal.rs:522]: `Scalar` only has one item, so it's ok to clone. You can refer to the implementation of this function.
https://github.com/datafuselabs/databend/blob/main/src/query/functions/src/scalars/array.rs#L121

Comment (youngsofun) [reply src/query/functions/src/scalars/decimal.rs:522]: not the clone of scala, I mean the `match`

**Comment by sundy-li** [comment]
> > You can modify `CAST_FROM_STRING_RULES` to have `String` to `Decimal`.
> 
> add `String` to `Float` too?

ok

File: src/query/functions/src/scalars/decimal.rs (hunk 3 above)
Comment (youngsofun) [reply src/query/functions/src/scalars/decimal.rs]: I refactored with a more straightforward way.

File: src/query/functions/tests/it/scalars/testdata/array.txt [hunk 5]
```diff
@@ -1800,15 +1796,13 @@ output domain  : []
...
-  --> SQL:1:12
-  |
-1 | array_sort([1.2, NULL, 3.4, 5.6, '2.2', NULL], 'ASC', 'NULLS FIRST')
-  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no overload satisfies `array(Decimal(2, 1), NULL, Decimal(2, 1), Decimal(2, 1), String, NULL)`
-
-has tried possible overloads:
-  array(T0, T0, T0, T0, T0, T0) :: Array(T0)  : unable to find a common super type for `Decimal(2, 1) NULL` and `String`
-
+ast            : array_sort([1.2, NULL, 3.4, 5.6, '2.2', NULL], 'ASC', 'NULLS FIRST')
+raw expr       : array_sort_asc_null_first(array(1.2_decimal(2, 1), NULL, 3.4_decimal(2, 1), 5.6_decimal(2, 1), "2.2", NULL))
```
b41sh, line 1800: Is the automatic cast `String` to `Decimal` as expected?

b41sh (approved):

youngsofun:
https://github.com/datafuselabs/databend/actions/runs/4344883781/jobs/7588830162 ci failure not related to this pr

youngsofun:
this file only in cluster mode 
https://github.com/datafuselabs/databend/actions/runs/4344883781/jobs/7589074001

```
Test finished, fail fast enabled, 1 out of 3369 records failed to run
0: query result mismatch:
[SQL] SELECT sum(c1),  sum(c2) FROM test;
[Diff] (-expected|+actual)
-   5050 0
+   NULL NULL
at tests/sqllogictests/suites/base/09_fuse_engine/09_0023_replace_into:195

```

PR #10381, opened 2023-03-06
fix(query): fix nested data types encode as NDJSON

### LINKED ISSUE #10367: add test to check unloaded file with map data type
better if add test to check unloaded file can be load and is the same
  https://github.com/datafuselabs/databend/pull/10176/files 

others lgtm

_Originally posted by @youngsofun in https://github.com/datafuselabs/databend/issues/10238#issuecomment-1452756623_

b41sh:

## Summary

- fix nested data types encode as NDJSON
- `FieldEncoderJSON` don't use `FieldEncoderValues` as a nested encoder, because `JSON` strings are encoded in a different format.

Closes #10367


youngsofun (approved):

RinChanNOWWW (approved):

youngsofun:
It suddenly occurred to me:  JSON can only accept string as object key, I guess we don`t have this limit, how to encode?

b41sh:
> It suddenly occurred to me: JSON can only accept string as object key, I guess we don`t have this limit, how to encode?

The legal `JSON Object` key must be a String, maybe we can add an additional cast to convert the key to the corresponding data type?

PR #10382, opened 2023-03-06
chore(test): make tpch work with cluster key

sundy-li:

## Summary

Summary about this PR

Closes #issue


xudong963 (approved):

BohuTANG:
Does the ClickBench TPCH CI also need to be modified? @everpcpc

everpcpc:
> Does the ClickBench TPCH CI also need to be modified? @everpcpc

clickbench CI uses `benchmark/clickbench/tpch/create.sql` which is already updated.

BohuTANG (approved):

PR #10395, opened 2023-03-06
chore: rename UserStageInfo to StageInfo.

youngsofun:

## Summary

- `UserStage` is a special kind of Stage
- no real change to PB data in meta

Closes #issue

### DIFF
--- a/src/meta/proto-conv/src/util.rs
+++ b/src/meta/proto-conv/src/util.rs
@@ -93,7 +93,7 @@ const META_CHANGE_LOG: &[(u64, &str)] = &[
     ),
     (
         30,
-        "2023-02-21: Add: config.proto/WebhdfsStorageConfig; Modify: user.proto/UserStageInfo::StageStorage",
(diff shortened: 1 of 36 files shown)


src/meta/proto-conv/src/util.rs
```diff
@@ -93,7 +93,7 @@ const META_CHANGE_LOG: &[(u64, &str)] = &[
     ),
     (
         30,
-        "2023-02-21: Add: config.proto/WebhdfsStorageConfig; Modify: user.proto/UserStageInfo::StageStorage",
+        "2023-02-21: Add: config.proto/WebhdfsStorageConfig; Modify: user.proto/StageInfo::StageStorage",
```
drmingdrmer, line 96: Never change these record, just append a new one.

drmingdrmer (commented):
Never change ver, test cases in proto-conv.

These are kept to ensure backward compatibility. Only add new test cases.

An old test case can be removed only when an obsolete `VER` is no longer supported.

sundy-li (approved):

PR #10408, opened 2023-03-07
chore: do not cache non-deterministic queries.

RinChanNOWWW:

## Summary

Do not cache non-deterministic queries.

### DIFF
--- a/src/query/catalog/src/table_context.rs
+++ b/src/query/catalog/src/table_context.rs
@@ -87,6 +87,8 @@ pub trait TableContext: Send + Sync {
     fn set_partitions(&self, partitions: Partitions) -> Result<()>;
     fn add_partitions_sha(&self, sha: String);
     fn get_partitions_shas(&self) -> Vec<String>;
+    fn if_can_cache(&self) -> bool;
+    fn cannot_cache(&self);
 
     fn attach_query_str(&self, kind: String, query: &str);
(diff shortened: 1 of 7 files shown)


src/query/catalog/src/table_context.rs
```diff
@@ -87,6 +87,8 @@ pub trait TableContext: Send + Sync {
     fn set_partitions(&self, partitions: Partitions) -> Result<()>;
     fn add_partitions_sha(&self, sha: String);
     fn get_partitions_shas(&self) -> Vec<String>;
+    fn if_can_cache(&self) -> bool;
```
BohuTANG, line 90: Only need one?
if_can_cache is contains cannot_cache?

src/query/catalog/src/table_context.rs
```diff
@@ -87,6 +87,8 @@ pub trait TableContext: Send + Sync {
     fn set_partitions(&self, partitions: Partitions) -> Result<()>;
     fn add_partitions_sha(&self, sha: String);
     fn get_partitions_shas(&self) -> Vec<String>;
+    fn get_cacheable(&self) -> bool;
```
ariesdevil, line 90: How about change the name to `is_cacheable`

ariesdevil (approved):
rest LGTM

sundy-li (approved):

PR #10420, opened 2023-03-08
fix(cluster): use ctx runtime spawn flight action

zhang2014:

## Summary

fix(cluster): use ctx runtime spawn flight action

Closes #issue


zhang2014:
wrong fix, revert.

PR #10424, opened 2023-03-08
fix(query): fix table stats incorrect when create table with exists snapshot

### LINKED ISSUE #10406: bug: incorrect snapshot stats after clone to a new table
**Summary**

```

CREATE TABLE `tt` (
  `Time` TIMESTAMP,
) ENGINE=FUSE TRANSIENT='T' ;


insert into tt(`Time`) select now() from numbers(100);

set hide_options_in_show_create_table = 0;

show create table tt; --- and get the SNAPSHOT_LOCATION

CREATE TABLE `tt_v2` (
  `Time` TIMESTAMP,
) ENGINE=FUSE COMPRESSION='lz4' SNAPSHOT_LOCATION='xxxxx';


select count(1) from tt_v2;  -- it's zero, related to empty stats 

explain select count(1) from tt_v2; 
 
```

TCeason:

## Summary

fix table stats incorrect when create table with exists snapshot

When create table with options SNAPSHOT_LOCATION,

after create table , will judge whether the SNAPSHOT_LOCATION actually exists.

* If exists, will regenerate the table statistic and recreate table.

* if not exists, will drop the table and return err.

Closes #10406


src/query/catalog/src/table.rs
```diff
@@ -225,11 +225,11 @@ pub trait Table: Sync + Send {
         Ok(())
     }
 
-    fn table_statistics(&self) -> Result<Option<TableStatistics>> {
+    async fn table_statistics(&self) -> Result<Option<TableStatistics>> {
         Ok(None)
     }
 
-    async fn column_statistics_provider(&self) -> Result<Box<dyn ColumnStatisticsProvider>> {
+    async fn column_statistics_provider(&self) -> Result<Box<dyn ColumnStatisticsProvider + Send>> {
```
sundy-li, line 232: `ColumnStatisticsProvider: Send`

src/query/storages/fuse/src/fuse_table.rs
```diff
@@ -525,17 +525,27 @@ impl Table for FuseTable {
...
-    fn table_statistics(&self) -> Result<Option<TableStatistics>> {
-        let s = &self.table_info.meta.statistics;
-        Ok(Some(TableStatistics {
-            num_rows: Some(s.number_of_rows),
-            data_size: Some(s.data_bytes),
-            data_size_compressed: Some(s.compressed_data_bytes),
-            index_size: Some(s.index_data_bytes),
-        }))
+    async fn table_statistics(&self) -> Result<Option<TableStatistics>> {
+        let res = if let Some(snapshot) = self.read_table_snapshot().await? {
```
sundy-li, line 529: why not set the `statistics` into `table_info` during create from snapshot?

This changes could lead `read_table_snapshot` to be called every time.

src/query/service/src/interpreters/interpreter_table_create.rs
```diff
@@ -144,7 +152,55 @@ impl CreateTableInterpreter {
...
+        if let Some(snapshot_loc) = self.plan.options.get(OPT_KEY_SNAPSHOT_LOCATION) {
+            let table = self
+                .ctx
+                .get_table(
+                    self.plan.catalog.as_str(),
+                    self.plan.database.as_str(),
+                    self.plan.table.as_str(),
+                )
+                .await?;
+            let fuse_table = FuseTable::try_from_table(table.as_ref())?;
```
BohuTANG, line 166: Putting the fuse table related here is a bit tricky?
Is it possible to check the table engine and get the stats from the table stats API:
```
fn table_statistics()
```

TCeason: Try to do this but if we modify will always read snapshot

TCeason: > Putting the fuse table related here is a bit tricky?
> 
> Is it possible to check the table engine and get the stats from the table stats API:
> 
> ```
> 
> fn table_statistics()
> 
> ```
> 
> 

And now if we use a not exists snapshot loc create table, it can succ. Will err only query it.

BohuTANG: Can we fix this with table API or does it in the internal table engine implementation not put the codes to the interpreter?

TCeason: Like this?

async fn table_statistics(&self) -> Result<Option<TableStatistics>> {
        let res = if let Some(snapshot) = self.read_table_snapshot().await? {
            Some(
                TableStatistics {
                    num_rows: Some(snapshot.summary.row_count),
                    data_size: Some(snapshot.summary.uncompressed_byte_size),
                    data_size_compressed: Some(snapshot.summary.compressed_byte_size),
                    index_size: Some(snapshot.summary.index_size),
                }
            )
        } else {
            let s = self.table_info.meta.statistics.clone();
            Some(TableStatistics {
                num_rows: Some(s.number_of_rows),
                data_size: Some(s.data_bytes),
                data_size_compressed: Some(s.compressed_data_bytes),
                index_size: Some(s.index_data_bytes),
            })
        };
        Ok(res)
    }

TCeason: I think we can do this. because create table XX (id int) snapshot_loc='yy' the 'yy' is from another table named YY and I can not get the table id of YY until the XX table create success.

https://github.com/datafuselabs/databend/issues/10406#issuecomment-1458071911

And the value is incorrect is because read stat in rule_fold_count_aggregate.rs.

How about when the precise_cardinality not equal 0, use this optimize?

```rust
if card != 0 {
    let mut scalars = agg.aggregate_functions;
    for item in scalars.iter_mut() {
        item.scalar = ScalarExpr::ConstantExpr(ConstantExpr {
            value: Literal::UInt64(card),
            data_type: Box::new(item.scalar.data_type()),
        });
    }
    let eval_scalar = EvalScalar { items: scalars };
    let dummy_table_scan = DummyTableScan;
    state.add_result(SExpr::create_unary(
        eval_scalar.into(),
        SExpr::create_leaf(dummy_table_scan.into()),
    ));
}
```

select min/max(id) from XX is ok, because of the column provider is read from the snapshot.

TCeason: After modify fold count:

```sql
mysql> select count(1) from tt_v2;
+----------+
| count(1) |
+----------+
|      100 |
+----------+
1 row in set (0.05 sec)
Read 100 rows, 0.00 B in 0.008 sec., 12.94 thousand rows/sec., 0.00 B/sec.

mysql> explain select count(1) from tt_v2;
+----------------------------------------------------------------------------------------------------------------------------------+
| explain                                                                                                                          |
+----------------------------------------------------------------------------------------------------------------------------------+
| EvalScalar                                                                                                                       |
| ├── expressions: [count(1) (#2)]                                                                                                 |
| ├── estimated rows: 1.00                                                                                                         |
| └── AggregateFinal                                                                                                               |
|     ├── group by: []                                                                                                             |
|     ├── aggregate functions: [count()]                                                                                           |
|     ├── estimated rows: 1.00                                                                                                     |
|     └── AggregatePartial                                                                                                         |
|         ├── group by: []                                                                                                         |
|         ├── aggregate functions: [count()]                                                                                       |
|         ├── estimated rows: 1.00                                                                                                 |
|         └── TableScan                                                                                                            |
|             ├── table: default.default.tt_v2                                                                                     |
|             ├── read rows: 100                                                                                                   |
|             ├── read bytes: 0                                                                                                    |
|             ├── partitions total: 1                                                                                              |
|             ├── partitions scanned: 1                                                                                            |
|             ├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 1 to 1, bloom pruning: 0 to 0>]       |
|             ├── push downs: [filters: [], limit: NONE]                                                                           |
|             ├── output columns: []                                                                                               |
|             └── estimated rows: 0.00                                                                                             |
+----------------------------------------------------------------------------------------------------------------------------------+
21 rows in set (0.04 sec)

mysql> analyze table tt_v2;

mysql> explain select count(1) from tt_v2;
+----------------------------------------+
| explain                                |
+----------------------------------------+
| EvalScalar                             |
| ├── expressions: [count(1) (#2)]       |
| ├── estimated rows: 1.00               |
| └── EvalScalar                         |
|     ├── expressions: [100]             |
|     ├── estimated rows: 1.00           |
|     └── DummyTableScan                 |
+----------------------------------------+
7 rows in set (0.04 sec)

```

dantengsky:
https://github.com/datafuselabs/databend/blob/84ef1b7bd63310a6b905f2e4448b59efb52c6b34/src/query/service/src/interpreters/interpreter_table_create.rs#L193-L200

~~~
            let snapshot = snapshot_reader.read(&params).await;
            match snapshot {
                Ok(snapshot) => {
                     .... 
                }
                Err(e) => { 
                    catalog
                        .drop_table_by_id(DropTableByIdReq {
                            if_exists: true,
                            tb_id: table.get_table_info().ident.table_id,
                        })
                        .await?;
                    return Err(e);
                }
            }
        }
~~~

will a new table be left there with incorrect statistics , if `catalog.drop_table_by_id` failed, let's say, caused by a network error?

--------------------
How about this?

- with `snapshot_loc`, and `MetaReaders::table_snapshot_reader` 
  we can read the `TableSnapshot`
   the version of the table snapshot can be extracted by using this
   https://github.com/datafuselabs/databend/blob/f9badf477bbcb7ff86c490e63dfc743659614bda/src/query/storages/fuse/src/io/locations.rs#L110
- from the `TableSnapshot.summary` we can build the `TableStatictics`
   https://github.com/datafuselabs/databend/blob/77379b5b2240d76ff83cc91247fbf4bd07baba3f/src/query/storages/common/table-meta/src/meta/v2/snapshot.rs#L52

dantengsky:
> https://github.com/datafuselabs/databend/blob/84ef1b7bd63310a6b905f2e4448b59efb52c6b34/src/query/service/src/interpreters/interpreter_table_create.rs#L193-L200
> 
> ```
>             let snapshot = snapshot_reader.read(&params).await;
>             match snapshot {
>                 Ok(snapshot) => {
>                      .... 
>                 }
>                 Err(e) => { 
>                     catalog
>                         .drop_table_by_id(DropTableByIdReq {
>                             if_exists: true,
>                             tb_id: table.get_table_info().ident.table_id,
>                         })
>                         .await?;
>                     return Err(e);
>                 }
>             }
>         }
> ```
> 
> will a new table be left there with incorrect statistics , if `catalog.drop_table_by_id` failed, let's say, caused by a network error?
> 
> How about this?
> 
> * with `snapshot_loc`, and `MetaReaders::table_snapshot_reader`
>   we can read the `TableSnapshot`
>   the version of the table snapshot can be extracted by using this
>   https://github.com/datafuselabs/databend/blob/f9badf477bbcb7ff86c490e63dfc743659614bda/src/query/storages/fuse/src/io/locations.rs#L110
> * from the `TableSnapshot.summary` we can build the `TableStatictics`
>   https://github.com/datafuselabs/databend/blob/77379b5b2240d76ff83cc91247fbf4bd07baba3f/src/query/storages/common/table-meta/src/meta/v2/snapshot.rs#L52

O...  the `operator`

TCeason:
> > https://github.com/datafuselabs/databend/blob/84ef1b7bd63310a6b905f2e4448b59efb52c6b34/src/query/service/src/interpreters/interpreter_table_create.rs#L193-L200
> > 
> > ```
> >             let snapshot = snapshot_reader.read(&params).await;
> >             match snapshot {
> >                 Ok(snapshot) => {
> >                      .... 
> >                 }
> >                 Err(e) => { 
> >                     catalog
> >                         .drop_table_by_id(DropTableByIdReq {
> >                             if_exists: true,
> >                             tb_id: table.get_table_info().ident.table_id,
> >                         })
> >                         .await?;
> >                     return Err(e);
> >                 }
> >             }
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will a new table be left there with incorrect statistics , if `catalog.drop_table_by_id` failed, let's say, caused by a network error?
> > How about this?
> > 
> > * with `snapshot_loc`, and `MetaReaders::table_snapshot_reader`
> >   we can read the `TableSnapshot`
> >   the version of the table snapshot can be extracted by using this
> >   https://github.com/datafuselabs/databend/blob/f9badf477bbcb7ff86c490e63dfc743659614bda/src/query/storages/fuse/src/io/locations.rs#L110
> > * from the `TableSnapshot.summary` we can build the `TableStatictics`
> >   https://github.com/datafuselabs/databend/blob/77379b5b2240d76ff83cc91247fbf4bd07baba3f/src/query/storages/common/table-meta/src/meta/v2/snapshot.rs#L52
> 
> O... the `operator`

We can do this modify

```rust
    async fn table_statistics(&self) -> Result<Option<TableStatistics>> {
        let s = &self.table_info.meta.statistics;
        if s.number_of_rows != 0 {
            Ok(Some(TableStatistics {
                num_rows: Some(s.number_of_rows),
                data_size: Some(s.data_bytes),
                data_size_compressed: Some(s.compressed_data_bytes),
                index_size: Some(s.index_data_bytes),
            }))
        } else {
            // add read snapshot in there
        }
    }
```

If num_rows is 0 means the new table or not stat is incorrect, we can read snapshot.

dantengsky:
> If num_rows is 0 means the new table or not stat is incorrect, we can read snapshot.

not sure if `num_rows is 0` indicates that the stat is incorrect. (seems we should not conclude like this)

besides: there is an implicit constraint for the  `Table::table_statistics`, 
this method is used in **management mode** , which can not access object storage .

TCeason:
> > If num_rows is 0 means the new table or not stat is incorrect, we can read snapshot.
> 
> not sure if `num_rows is 0` indicates that the stat is incorrect. (seems we should not conclude like this)
> 
> besides: there is an implicit constraint for the `Table::table_statistics`, this method is used in **management mode** , which can not access object storage .

If so , how about triggering a backoff: https://github.com/datafuselabs/databend/pull/10424#discussion_r1129135248

dantengsky:
> If so , how about triggering a backoff: [#10424 (comment)](https://github.com/datafuselabs/databend/pull/10424#discussion_r1129135248)

not sure about this, @sundy-li  we need your help

TCeason:
> > If so , how about triggering a backoff: [#10424 (comment)](https://github.com/datafuselabs/databend/pull/10424#discussion_r1129135248)
> 
> not sure about this, @sundy-li we need your help

We can not do this. Because the root cause is stats incorrect, in there we can backoff ,but join or other use stats will also generate err result.

src/query/service/src/interpreters/interpreter_table_create.rs
TCeason: Now we use operator from QueryContext to read_snapshot when create table.

sundy-li (approved):
The current implementation is enough to fix the issue without too much effort.

TCeason:
> The current implementation is enough to fix the issue without too much effort.

A hack usage generates a hack modify pr.. How about supporting https://docs.snowflake.com/en/sql-reference/sql/create-clone in a later version?

dantengsky (approved):

sundy-li:
> > The current implementation is enough to fix the issue without too much effort.
> 
> A hack usage generates a hack modify pr.. How about supporting [docs.snowflake.com/en/sql-reference/sql/create-clone](https://docs.snowflake.com/en/sql-reference/sql/create-clone) in a later version?

That could be great. 

`set hide_options_in_show_create_table = 0;` is hard to use.

PR #10427, opened 2023-03-08
refactor: atomic table commit and upsert_copied_files_info_to_meta

dantengsky:

## Summary

currently, in the last phase of the execution of `copy into`, the following two actions are not "atomic"
 1. Commit data to table.
 2. Upsert files(status with NeedCopy) info to meta.
 
 
https://github.com/datafuselabs/databend/blob/94fb99691c48523dc137a91f92847385cb087564/src/query/service/src/interpreters/interpreter_copy.rs#L427-L439
 
after action 1, the data appended is published, and visible to subsequent tx, but if action 2 fails, the files processed as data appended in action 1 would not be kept in the meta server, and we are taking the risk of processing duplicated files later.

this pr tries to address the above issue by combining the two actions into one atomic action:

- let the `UpdateTableMetaReq`, which is used in action 1 to commit table meta to meta service, carries extra information of files being copied.

- during atomically updating table meta, combined with the action of keeping information of files that have been copied (of this tx), the mini kv tx goes like this:

  ~~~
      if  table_id.version = expected_table_version
            // file_copied_1_key is blended with table id
           // version eq 0, indicates that this file is absent
           && file_copied_1_key.version == 0 
           && file_copied_2_key.version == 0
           ...
            && file_copied_n_key.version == 0
       then 
            put_table_meta(...);
            put(file_copied_1_key, file_copied_1_info);
            .....
            put(file_copied_1_key, file_copiedn1_info);
       else 
           get(table_id)
          

  ~~~
     
     
  if mini kv tx success,  then the changes of table meta and files copied are kept atomically, otherwise,  check the table meta info returned by the `else branch`
  - if the table version mismatch, just return the `TableVersionMismatched` error
  -  otherwise, return the `DuplicatedUpsertFiles` error

NOTE:

- there may be a large number of files being processed in one tx

  tens or hundreds of thousand is not unusual. we need to verify the performance impactions of this PR

- in `force mode` of copy stmt, e.g.
 
  `copy into t from @stage FILES = (....) FORCE = true`
  
  which do not care about the duplication of tables' copied files set.
  
   The logic of  the above mini KV Tx falls back to the current main branch:
   
   instead of checking if files are absent, the tx will check if the files' versions are not changed(and need to read them before construct the mini kv tx), and if changed the mini kv tx will be retried, till success, or max number of retry exceeded.
   
   I might misunderstand the "semantic" of upsert, but the `force` mode can or should be implemented as `mput` IMO.
  

Closes #issue


lichuang (approved):

drmingdrmer:
> * there may be a large number of files being processed in one tx
>   tens or hundreds of thousand is not unusual. we need to verify the performance impactions of this PR

Is it possible to split the `100_000` files into 100 transactions, with each transaction containing 1000 files, instead of committing all the files in one big transaction?

youngsofun:
```
nstead of checking if files are absent, the tx will check if the files' versions are not changed(and need to read them before construct the mini kv tx), and if changed the mini kv tx will be retried, till success of max number of retry exceeded.
```

can you explain more: why we need to `check if the files' versions are not changed`,  can we only check if TableMeta seq changed?

dantengsky:
> > * there may be a large number of files being processed in one tx
> >   tens or hundreds of thousand is not unusual. we need to verify the performance impactions of this PR
> 
> Is it possible to split the `100_000` files into 100 transactions, with each transaction containing 1000 files, instead of committing all the files in one big transaction?

that will be better,  but not sure if it is possible :( 

suppose batching in the following way(2 files a batch):

- files to be committed in tx1
 {f1}
- files to be committed in tx2
 {fa, fb, f1, f2}
 
 - ts_0
   tx1 tries {f1} and success
 - ts_1
   tx2 tries {fa, fb} and success (kept in meta server)
 - ts_2
   tx2 tries {f1, f2} and failed (due to duplication of file f1)
   if we can not revert the previous tx ({fa, fb}), subsequent transactions can not process {fa, fb}

youngsofun:
> ```
> instead of checking if files are absent, the tx will check if the files' versions are not changed(and need to read them before construct the mini kv tx), and if changed the mini kv tx will be retried, till success of max number of retry exceeded.
> ```
> 
> can you explain more: when force=true, why we need to `check if the files' versions are not changed`, can we only check if TableMeta seq changed?

what the meaning of recording the meta of files?  
it is meaningful to record the `lastest copied`(determined by order of commit)  meta of a file? since contents of 2  `insert`  is  accumulated, and order not matters, and we can not replay the 2 copies since we only have one version of file meta.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1926,47 +1930,14 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
                 "upsert_table_copied_file_info"
             );
 
-            let mut condition = vec![txn_cond_seq(&tbid, Eq, tb_meta_seq)];
```
youngsofun, line 1929: upsert_table_copied_file_info in trait Catalog can be removed?

dantengsky: yeah, seems no other components using this method now, let me try rm it

dantengsky:
> ```
> instead of checking if files are absent, the tx will check if the files' versions are not changed(and need to read them before construct the mini kv tx), and if changed the mini kv tx will be retried, till success of max number of retry exceeded.
> ```
> 
> can you explain more: when force=true, why we need to `check if the files' versions are not changed`, can we only check if TableMeta seq changed?

That is the implementation we used currently. frankly speaking, I think in force mod, upsert should be implemented ... by using something like `mput`, without checking the version or retry (see the summary of this pr), but I am not sure if other components rely on the current logic of implementation (things like the gc of "copied files"), thus it is kept "unchanged".

btw, I've discussed this with @lichuang via "side channel" :), seems that the above logic could be simplified (but not in this PR).

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2855,3 +2909,83 @@ async fn list_tables_from_share_db(
...
+    // may return `TxnRetryMaxTimes`.
+    // So now, in case that `TableCopiedFileInfo` has expire time, remove `TableCopiedFileLockKey`
+    // in each function. In this case there is chance that some `TableCopiedFileInfo` may not be
+    // removed in `remove_table_copied_files`, but these data can be purged in case of expire time.
+
+    let mut file_name_infos = req.file_info.clone().into_iter();
+
+    if fail_if_duplicated {
+        // "fail_if_duplicated" mode, assumes files are absent
+        for (file_name, file_info) in file_name_infos {
```
youngsofun, line 2945: maybe we can optimize here:

before copy,  we got a table version with a list  of files to copy, these files not exists as long as  table version  not changed, so no need to check their existence if  table version not changed since then.

dantengsky: hope I get it right, seems that is exactly what this PR does (in non-force mode):

in one kv tx,  we assume that

- the table version does not change
- the files are absent
   these files are collected before constructing the kv tx

~~~
    if  table_id.version = expected_table_version
          // file_copied_1_key is blended with table id
         // version eq 0, indicates that this file is absent
         && file_copied_1_key.version == 0 
         && file_copied_2_key.version == 0
         ...
          && file_copied_n_key.version == 0
     then 
          put_table_meta(...);
          put(file_copied_1_key, file_copied_1_info);
          .....
          put(file_copied_1_key, file_copiedn1_info);
     else 
         get(table_id)
~~~         

if the above tx failed (with a response, that contains the output of `else branch`, i.e. the

~~~
     else 
         get(table_id)
~~~

we check the table meta returned (note , in this case , the kv tx failed):

if the table version is not changed, according to the output of kv tx, then during the executions of this kv tx, there must be some duplicated files. then 
 - if not in `force` mode, we just return with the error `DuplicatedUpsertFiles` **immediately**, without retrying.
 - if in `force` mode, we fallback to the previous logic (before this PR), that will retry

drmingdrmer:
> > > * there may be a large number of files being processed in one tx
> > >   tens or hundreds of thousand is not unusual. we need to verify the performance impactions of this PR
> > 
> > 
> > Is it possible to split the `100_000` files into 100 transactions, with each transaction containing 1000 files, instead of committing all the files in one big transaction?
> 
> that will be better, but not sure if it is possible :(
> 
> suppose batching in the following way(2 files a batch):
> 
> * files to be committed in tx1
>   {f1}
> * files to be committed in tx2
>   {fa, fb, f1, f2}
> * ts_0
>   tx1 tries {f1} and success
> * ts_1
>   tx2 tries {fa, fb} and success (kept in meta server)
> * ts_2
>   tx2 tries {f1, f2} and failed (due to duplication of file f1)
>   if we can not revert the previous tx ({fa, fb}), subsequent transactions can not process {fa, fb}

If such a conflict is possible to happen, e.g., two transactions try to commit a same file to a table, I think the committing them in a single big transaction also has the same problem.

src/meta/api/src/schema_api_impl.rs
dantengsky: if in `force mode`, the files could be put "unconditionally" (except that the table id should exist, and the table version match),  the `if condition` can be adjusted to 

~~~
    if  table_id.version = expected_table_version   // (without check the field are absent)
~~~

but again, I am not sure if this may cause troubles (other components that depend on the current logic of implementation)

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2855,3 +2909,83 @@ async fn list_tables_from_share_db(
...
+            let key = TableCopiedFileNameIdent {
+                table_id,
+                file: file_name.to_owned(),
+            };
+            condition.push(txn_cond_seq(&key, Eq, 0));
```
youngsofun, line 2950: not need `condition.push(txn_cond_seq(&key, Eq, 0));` if  table_verison not changed sinces https://github.com/datafuselabs/databend/blob/main/src/query/service/src/interpreters/interpreter_copy.rs#L276

dantengsky:
> > > > * there may be a large number of files being processed in one tx
> > > >   tens or hundreds of thousand is not unusual. we need to verify the performance impactions of this PR
> > > 
> > > 
> > > Is it possible to split the `100_000` files into 100 transactions, with each transaction containing 1000 files, instead of committing all the files in one big transaction?
> > 
> > 
> > that will be better, but not sure if it is possible :(
> > suppose batching in the following way(2 files a batch):
> > 
> > * files to be committed in tx1
> >   {f1}
> > * files to be committed in tx2
> >   {fa, fb, f1, f2}
> > * ts_0
> >   tx1 tries {f1} and success
> > * ts_1
> >   tx2 tries {fa, fb} and success (kept in meta server)
> > * ts_2
> >   tx2 tries {f1, f2} and failed (due to duplication of file f1)
> >   if we can not revert the previous tx ({fa, fb}), subsequent transactions can not process {fa, fb}
> 
> If such a conflict is possible to happen, e.g., two transactions try to commit a same file to a table, I think the present approach that commits them in a single big transaction also has the same problem.

- files to be committed in tx1

   {f1}

- files to be committed in tx2

   {fa, fb, f1, f2}

-------

-  ts_0
   tx1 tries {f1}, and the kv tx is 
     ~~~
        if table_version_match && all_the_files_are_absent(f1) {
           update_table_meta();
           mput(f1, f2);
        } 
     ~~~
     and success
-  ts_1
   tx2 tries {fa, fb, f1, f2}, and the kv tx is 
     ~~~
        if table_version_match && all_the_files_are_absent(fa, fb, f1, f2) {
           update_table_meta();
           mput(fa, fb, f1, f2);
        } 
     ~~~
     since f1 is already there, the above tx will fail,  table meta and keys of files will not be updated.

src/meta/api/src/schema_api_impl.rs
youngsofun: we can also discuss it later without blocking this pr.

youngsofun:
Rest lgtm as long as the performance is ok.

src/meta/api/src/schema_api_impl.rs
dantengsky: > but this maybe a conflict if we will optimize force = true later out of txn?

not sure.

> not need condition.push(txn_cond_seq(&key, Eq, 0)); if table_verison not changed since

since we have already checked the files during the "coloring" phase? oh, it is reasonable to me, let me try this

dantengsky: oh, it is not safe to do that. 

 since the SchemaApi exposes `upsert_table_copied_file_info`, the table's copied files can be updated without increasing the version of the table.

youngsofun:
why not store copied files in an associated `fuse table` instead of meta from the beginning？
than we can utilize：

1.  parrellel processing/filtering，min max index，bloomfilter
2.  big storage and bandwidth. 
3. versioning and committing like table data.

src/meta/api/src/schema_api_impl.rs
youngsofun: if we need to keep it,  maybe we can add a update to table version in txn?

Issue #10422, opened 2023-03-08
Force to add a valid version number when building


drmingdrmer:
Otherwise, the process will panic because of a `"unknown"` version number:

```
{"timestamp":"2023-03-08T02:10:40.081378Z","level":"ERROR","fields":{"message":"panicked at 'Invalid semver: \"unknown\": unexpected character 'u' while parsing major version number', sr
c/meta/client/src/lib.rs:40:47","backtrace":"Backtrace [
{ fn: \"std::backtrace_rs::backtrace::libunwind::trace\", file: \"/rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/ ../../backtrace/src/backtrace/libunwind.rs\", line: 93 },
{ fn: \"std::backtrace_rs::backtrace::trace_unsynchronized\", file: \"/rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/st d/src/../../backtrace/src/backtrace/mod.rs\", line: 66 },
{ fn: \"std::backtrace::Backtrace::create\", file: \"/rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/backtrace.r s\", line: 332 },
{ fn: \"common_tracing::panic_hook::log_panic\", file: \"/workspace/src/common/tracing/src/panic_hook.rs\", line: 35 },
{ fn: \"common_tracing::panic_hook::set_panic_hook::{{closure}}\", file: \"/workspace/src/common/tracing/src/panic_hook.rs\", line: 30 },
{ fn: \"<alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call\", file: \"/rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/alloc/src/boxed.rs\", line: 2032 },
{ fn: \"std::panicking::rust_panic_with_hook\", file: \"/rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/panicking.rs\", line: 692 },
{ fn: \"std::panicking::begin_panic_handler::{{closure}}\", file: \"/rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/panicking.rs \", line: 579 },
{ fn: \"std::sys_common::backtrace::__rust_end_short_backtrace\", file: \"/rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/sys_common/backtrace.rs\", line : 137 },
[...]

PR #10432, opened 2023-03-08
fix: if git-tag version is not found, stop building

drmingdrmer:

##### fix: if git-tag version is not found, stop building

Network protocol relies on the version number. A `unknown` will panic
the process at runtime.

- Fix: #10422

## Changelog

- Bug Fix

## Related Issues


BohuTANG (approved):

Xuanwo (approved):

drmingdrmer:
In fact, there is a problem when a git tag is required:

https://github.com/datafuselabs/databend/actions/runs/4362125337/jobs/7626670912

It looks like when building `aarch64_musl` it can not find a git tag in the local working dir.

BohuTANG:
> In fact, there is a problem when a git tag is required:
> 
> https://github.com/datafuselabs/databend/actions/runs/4362125337/jobs/7626670912
> 
> It looks like when building `aarch64_musl` it can not find a git tag in the local working dir.

It broke our build, the root cause needs to be addressed.

PR #10435, opened 2023-03-08
feat(query): make regexp initialized at compile time

### LINKED ISSUE #10286: Feature: make regexp initialized at compile time
**Summary**

Description for this feature.

https://github.com/datafuselabs/databend/blob/main/src/query/service/src/servers/federated_helper.rs  will use many regexps to match the query. We can improve this via initializing the regexp at compile time.

xinlifoobar:

## Summary

Initializing the federated regex at compile time to improve runtime matching perf.

Closes #10286


xinlifoobar:
The CI failed due to timeout, could we retry it?

src/query/service/src/servers/mysql/mysql_federated.rs
```diff
@@ -147,121 +212,157 @@ impl MySQLFederated {
...
+            None,
+            // "(?i)^(/\\*!40014 SET(.*) \\*/)$"
+            None,
+            // "(?i)^(/\\*!40000 SET(.*) \\*/)$"
+            None,
+            // "(?i)^(/\\*!40000 ALTER(.*) \\*/)$"
+            None,
         ];
 
+        let rules = MIXED_CHECK_FEDERATED_REGEX_LIST
```
sundy-li, line 361: It's ziped, so why not make `MIXED_CHECK_FEDERATED_REGEX_LIST` as `Vec<(Regex, Option<(TableSchemaRef, DataBlock)> )>`

sundy-li: So we can save the comments

sundy-li (approved):

PR #10439, opened 2023-03-08
feat: do not throw an error if it appears in a short-circuited branch.

### LINKED ISSUE #10409: Feature: support short-circuit evaluation.
**Summary**

Short-circuit evalution can reduce computation and avoid doing some operations under some conditions.

Example:

```sql
select * from t 
where (case when a > 0 then b / a else null end) > c;
```

without short-circuit evaluation, the query will throw an error if a is 0.

RinChanNOWWW:

## Summary

Add a new `Expr`: `Expr::Catch`.

`case when` clause, `if`, and `multi_if`'s branches will be wrapped in `Expr::Catch`.

`Expr::Catch` will catch the error and return `Tuple<T, String NULL>` which contains evaluation results and the error messages (maybe not exists).

The error will be thrown only if it appears in the activated branch.

Closes #10409


zhang2014:
What happen panic(select if(1== 1, 1 / 0, 1 / 1))?

RinChanNOWWW:
> What happen panic(select if(1== 1, 1 / 0, 1 / 1))?

`select if(1 = 1, 1/0, 1/1)` will throw an error.

`select if(1 = 1, 1/1, 1/0)` will not.

sundy-li:
@mergify update

sundy-li (approved):

andylokandy (approved):

Issue #10233, opened 2023-02-26
Feature: support more pg functions


sundy-li:
**Summary**

Description for this feature.
 
We can support more Mathematical operators like [PostgreSQL](https://www.postgresql.org/docs/7.4/functions-math.html)
```
 
Operator | Description | Example | Result
-- | -- | -- | --

^ | exponentiation | 2.0 ^ 3.0 | 8
\|/ | square root | \|/ 25.0 | 5
\|\|/ | cube root | \|\|/ 27.0 | 3
! | factorial | 5 ! | 120
@ | absolute value | @ -5.0 | 5
& | bitwise AND | 91 & 15 | 11
\| | bitwise OR | 32 \| 3 | 35
# | bitwise XOR | 17 # 5 | 20
~ | bitwise NOT | ~1 | -2
<< | bitwise shift left | 1 << 4 | 16
>> | bitwise shift right | 8 >> 2 | 2

```

PR #10444, opened 2023-03-08
feat(query): add more pg style operators

jun0315:

### support  pg operators

image

### support functions

Pg operator has some common functions, so it is also added to functions.

- cbrt (cube root)
- factorial

## Related 
Closes https://github.com/datafuselabs/databend/issues/10233
[postgresql: Mathematical Functions and Operators](https://www.postgresql.org/docs/7.4/functions-math.html)


src/query/functions/src/scalars/arithmetic.rs
```diff
@@ -359,29 +359,28 @@ fn register_binary_arithmetic(registry: &mut FunctionRegistry) {
...
-
-macro_rules! register_pg_square_root{
+macro_rules! register_square_root {
     ( $n:ty, $registry:expr) => {
         type N = $n;
-        type T = <N as ResultTypeOfUnary>::Sum;
+        type T = F64;
         $registry.register_1_arg::<NumberType<N>, NumberType<T>, _, _>(
-            "pgsquareroot",
+            "squareroot",
```
sundy-li, line 367: "sqrt" is better name I think.

sundy-li:
Need sqllogicaltests to cover new functions/operators in `02_0014_function_maths`.

sundy-li (approved):

src/query/sql/src/planner/planner.rs
```diff
@@ -84,6 +84,7 @@ impl Planner {
                 // Step 2: Parse the SQL.
                 let backtrace = Backtrace::new();
                 let (mut stmt, format) = parse_sql(&tokens, sql_dialect, &backtrace)?;
+                println!("stmt: {:?}", stmt);
```
sundy-li, line 87: should be removed.

PR #10448, opened 2023-03-08
refactor(binder): extract util fn normalize_object_identifier_triple.

youngsofun:

## Summary

Closes #issue

### DIFF
--- a/src/query/service/src/interpreters/interpreter_view_alter.rs
+++ b/src/query/service/src/interpreters/interpreter_view_alter.rs
@@ -56,13 +56,13 @@ impl Interpreter for AlterViewInterpreter {
             .await?
             .iter()
             .any(|table| {
-                table.name() == self.plan.viewname.as_str()
+                table.name() == self.plan.view_name.as_str()
                     && table.get_table_info().engine() == VIEW_ENGINE
             })
         {
             return Err(ErrorCode::ViewAlreadyExists(format!(
                 "{}.{} view does not exist",
-                self.plan.database, self.plan.viewname
+                self.plan.database, self.plan.view_name
             )));
         }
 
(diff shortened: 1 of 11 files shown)


src/query/ast/tests/it/testdata/statement.txt
```diff
@@ -6975,6 +6975,163 @@ Copy(
...
 
 
+---------- Input ----------
+COPY INTO mytable
+                FROM (SELECT t.c1
+                    FROM @mystage/path/to/file (FILE_FORMAT=> 'PARQUET' ) t
+                    where t.c2 > 0
+                );
+---------- Output ---------
+COPY INTO mytable FROM (SELECT t.c1 FROM @mystage/path/to/file (FILE_FORMAT => 'PARQUET') AS t WHERE (t.c2 > 0)) SINGLE = false PURGE = false FORCE = false ON_ERROR = 'abort'
```
BohuTANG, line 6985: select `FILE_FORMAT => 'PARQUET'` and copy `FILE_FORMAT = PARQUET` mixed here is confusing, snowflake does not do that, it does not have the options for the SELECT, any better solution?

BohuTANG: `(SELECT t.c1 FROM @mystage/path/to/file (FILE_FORMAT => 'PARQUET')`
in snow will use the stage file option:
`(SELECT t.c1 FROM @mystage/path/to/file)`

BohuTANG: hmm, no need forbidden, let user to do that is ok, like snow did, in his example as above, it doesn't have the option to confuse the user who want transforme the data from a stage.

youngsofun: what about only report error when the same option is specified in both places and is diff?

```
COPY INTO mytable FROM (SELECT t.c1 FROM @mystage/path/to/file(FILE_FORMAT => 'CSV') AS t WHERE (t.c2 > 0)) SINGLE = false PURGE = false FORCE = false ON_ERROR = 'abort' FILE_FORMAT = 'PARQUET'
```

andylokandy (approved):

leiysky (approved):

PR #10451, opened 2023-03-08
fix(expr): fix common_super_type() for decimal

andylokandy:

## Summary

Summary about this PR

Closes #issue


youngsofun (approved):

src/query/expression/src/type_check.rs
```diff
@@ -559,24 +558,24 @@ pub fn common_super_type(
...
-
+        (DataType::Number(num_ty), decimal_ty @ DataType::Decimal(_))
+        | (decimal_ty @ DataType::Decimal(_), DataType::Number(num_ty))
+            if !num_ty.is_float() =>
+        {
+            Some(decimal_ty)
+        }
+        (DataType::Number(num_ty), DataType::Decimal(_))
+        | (DataType::Decimal(_), DataType::Number(num_ty))
+            if num_ty.is_float() =>
```
sundy-li, line 575: if `num_ty` is not float, should be decimal of `max_precision`.

youngsofun: `range > range(num_ty)` is enough?

andylokandy: > if num_ty is not float, should be decimal of max_precision.

if so, we should carefully compare the decimal in `can_auto_cast_to`: https://github.com/datafuselabs/databend/blob/45f3ca1bba14f1d192f9ab0094a9f63e9acfe792/src/query/expression/src/type_check.rs#L521

sundy-li: But it's always true, because `10**38` > `2^64 - 1`

andylokandy: I mean, for `(DataType::Number(n), DataType::Decimal(DecimalSize{precision: 1, scale: 1}))` , `can_auto_cast_to` should return false

sundy-li: Yes!

BohuTANG:
Is this PR still working?

sundy-li (approved):

PR #10457, opened 2023-03-09
feat(storage): Map data type support bloom filter

b41sh:

## Summary

Map data type support bloom filter

Part of #10062

From the test data example, we can see that the bloom filter is effective in reducing the query time if the value of the query does not exist.

```sql
MySQL root@127.0.0.1:default> select count(*) from test;
+----------+
| count(*) |
+----------+
| 100000   |
+----------+
1 row in set
Time: 0.153s
MySQL root@127.0.0.1:default> select * from test where log['ip'] = '205.91.162.148';
+----+----------------------------------------+
| id | log                                    |
+----+----------------------------------------+
| 1  | {'ip':'205.91.162.148','url':'test-1'} |
+----+----------------------------------------+
1 row in set
Time: 1.733s
MySQL root@127.0.0.1:default> select * from test where log['ip'] = '205.91.162.141';
+----+-----+
| id | log |
+----+-----+
+----+-----+
0 rows in set
Time: 0.129s
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8314,6 +8314,7 @@ version = "0.1.0"
 dependencies = [
  "anyerror",
  "cbordata",
+ "common-arrow",
  "common-exception",
  "common-expression",
  "common-functions",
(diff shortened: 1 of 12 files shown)


BohuTANG (approved):

BohuTANG (approved):
LGTM

PR #10464, opened 2023-03-09
docs(query): add access control doc about privilege 

### LINKED ISSUE #10385: tracking: privileges enhancement
**Summary**

- [x] Document: To describe the privileges that are available in the Databend access control model.
- [x] More privileges support(User, Role, Stage, Database, Table, Schema, Catalog)

TCeason:
## Summary

1. add doc: priv

2. fix grant create user/stage/role err

## Questions

1. I try to add priv doc. But I find in [privilege_access.rs
](https://github.com/datafuselabs/databend/blob/main/src/query/service/src/interpreters/access/privilege_access.rs) more plan does not check privilege. 

2. And I think this is unreasonable? Why create view need alter priv?
cc @BohuTANG 
```rust
Plan::CreateView(plan) => {
    session
        .validate_privilege(
            &GrantObject::Database(plan.catalog.clone(), plan.database.clone()),
            UserPrivilegeType::Alter,
        )
        .await?;
}
```

```sql
mysql> show grants;
+-----------------------------------------------------+
| Grants                                              |
+-----------------------------------------------------+
| GRANT CREATE,CREATE ROLE ON *.* TO 'b'@'%'          |
| GRANT SELECT ON 'default'.'system'.'one' TO 'b'@'%' |
+-----------------------------------------------------+
2 rows in set (0.03 sec)

mysql> create view v_t2 as select * from t;
ERROR 1105 (HY000): Code: 1063, displayText = Permission denied, user 'b'@'%' requires ALTER privilege on 'default'.'default'.*.

```

Closes #10385

### DIFF
--- a/docs/doc/04-security/00-access-control/10-access-control-privileges.md
+++ b/docs/doc/04-security/00-access-control/10-access-control-privileges.md
@@ -1,0 +1,78 @@
+---
+title: Access Control Privileges
+sidebar_label: Privileges
+description:
+  Databend Access Control Privileges
+---
+
+This topic describes the privileges that are available in the Databend access control model.
+
+## All Privileges
+
+| Privilege | Object Type | Description |
+| :--                 | :--                  | :--                  |
+| ALL   |  All    | Grants all the privileges for the specified object type. |
+| Alter   |    Database, Table, View   | Privilege to alter databases or tables. |
+| Create   |     Database, Table    | Privilege to create databases or tables. |
+| Delete   |  Table   | Privilege to delete or truncate rows in a table. |
+| Drop       |    Database, Table, View    | Privilege to drop databases or tables or views and undrop databases or tables. |
+| Insert       |   Table       | Privilege to insert rows into tables. |
+| Select       |    Table      | Privilege to select rows from tables . |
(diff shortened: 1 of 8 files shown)


BohuTANG:
> I try to add priv doc. But I find in [privilege_access.rs](https://github.com/datafuselabs/databend/blob/main/src/query/service/src/interpreters/access/privilege_access.rs)
more plan does not check privilege.

Yes,  we need more privileges to add .

> And I think this is unreasonable? Why create view need alter priv?

Not make sense, we can fix it.

docs/doc/04-security/00-access-control/10-access-control-privileges.md
```diff
@@ -0,0 +1,78 @@
+---
```
BohuTANG, line 1: Better to move here?

image

BohuTANG: Before Use cases after SQL Functions

TCeason:
> Yes, we need more privileges to add .

Ok, I will add it in this pr. Now it will be set draft, after this done it will be ready for review.

Issue #10468, opened 2023-03-09
Bug: incorrect type inference of tuple access


leiysky:
```sql
MySQL [(none)]> create table t(a tuple(int null) null);
Query OK, 0 rows affected (1.520 sec)

MySQL [(none)]> insert into t values((1));
Query OK, 1 row affected (0.580 sec)

-- Result should be `INT NULL`
MySQL [(none)]> select typeof(a.1) from t;
+---------------+
| typeof(a.1)   |
+---------------+
| INT NULL NULL |
+---------------+
1 row in set (1.020 sec)
```

PR #10469, opened 2023-03-09
fix(expr): get(tuple) may return nested Nullable

andylokandy:

## Summary

Summary about this PR

Closes https://github.com/datafuselabs/databend/issues/10468

### DIFF
--- a/src/query/functions/src/scalars/tuple.rs
+++ b/src/query/functions/src/scalars/tuple.rs
@@ -92,8 +92,10 @@ pub fn register(registry: &mut FunctionRegistry) {
         Some(Arc::new(Function {
             signature: FunctionSignature {
                 name: "get".to_string(),
-                args_type: vec![DataType::Tuple(fields_ty.to_vec())],
-                return_type: fields_ty[idx].clone(),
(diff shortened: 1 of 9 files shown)


src/query/functions/src/scalars/tuple.rs
```diff
@@ -124,32 +126,45 @@ pub fn register(registry: &mut FunctionRegistry) {
...
                     column: Column::Tuple { fields, .. },
                     validity,
-                })) => Value::Column(Column::Nullable(Box::new(NullableColumn {
-                    column: fields[idx].to_owned(),
-                    validity: validity.clone(),
-                }))),
+                })) => {
+                    let field_col = fields[idx].as_nullable().unwrap();
+                    Value::Column(Column::Nullable(Box::new(NullableColumn {
+                        column: field_col.column.clone(),
```
sundy-li, line 164: Is there any way to check the double-nested nullable datatypes during compile?

andylokandy: We can check it at type_check.

sundy-li (approved):

leiysky (approved):

PR #10480, opened 2023-03-10
fix(ci): make logicaltest ci cover native format

sundy-li:

## Summary

Summary about this PR

Closes #issue


sundy-li:
```
MySQL [(none)]> SELECT b, SUM(a) AS sum FROM test GROUP BY b HAVING b=21 ORDER BY b;
ERROR 1105 (HY000): Code: 1068, displayText = Cannot join handle from context's runtime, cause: task 333 panicked.
```

`b` is binded as `group_item` now,  but it's pushed into storage.  cc @xudong963  @leiysky .

Xuanwo (approved):

native.toml
```diff
@@ -0,0 +1,121 @@
+# Usage:
+# databend-query -c databend_query_config_spec.toml
```
Xuanwo, line 2: I think this file should not be placed in the root directory.

sundy-li: Oh, i forgot to remove this file.

sundy-li:
Timeout is not related to this pr, let's merge.

Issue #9576, opened 2023-01-12
BUG: sqllogic test hangs (cluster mod)


dantengsky:
**Summary**

on behalf of @leiysky 

- https://github.com/datafuselabs/databend/actions/runs/3894810689/jobs/6659648903 
- https://github.com/datafuselabs/databend/actions/runs/3898564614/jobs/6657840741 

**re-produce locally:**

- build & bring up the cluster
~~~
cargo b

./scripts/ci/deploy/databend-query-cluster-3-nodes.sh

~~~

- run the sqllogic test in cluster mode with clickhouse handler

~~~
target/debug/databend-sqllogictests --handlers clickhouse --run_dir ydb --enable_sandbox --parallel 16 --debug
~~~

for a test box (linux) with 16 vCPUs,  hanging is expected to occur within 1-5 runs.

PR #10505, opened 2023-03-11
refactor: use GlobalIORuntime::instance() for reading file in copy.

youngsofun:

## Summary

Summary about this PR

Closes #issue


zhang2014:
Maybe hang? CC: @dantengsky 

ref: https://github.com/datafuselabs/databend/pull/9615/files

dantengsky:
> Maybe hang? CC: @dantengsky
> 
> ref: https://github.com/datafuselabs/databend/pull/9615/files

not sure : (  *seems* to me that it will not re-introduce the bug of https://github.com/datafuselabs/databend/issues/9576 which fixed in pr #9615

BohuTANG:
Does this patch good to have?

dantengsky:
@youngsofun hi, 

2023-03-15-194245_749x241_scrot

If the only concern here is to guarantee that the async closure beginning at line 204 will be executed in the thread pool of `GlobalIORuntime`,  then `tokio::spawn` should do the same job as well, since while spawning, the "current runtime" in the context, is the `GlobalIORuntime`. 

@zhang2014  Am I getting right about the behavior of  `tokio::spawn`?

PR #10024, opened 2023-02-13
feat(planner): Introduce bitmap to record applied rules

dusx1981:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

A draft pr for code review.

closes #9536

PR #10508, opened 2023-03-12
fix(planner): Correct the mapping relationship between RuleID and DEFAULT_REWRITE_RULES

dusx1981:

## Summary
**The new map**
The new mapping relationship between RuleID and DEFAULT_REWRITE_RULES looks like this：
image
**RuleIDs are put together according to the type，so each operator can apply its own rules in the correct order.**

**The scalability**
Considering that each operator only applies its own rules in the end, the new mapping only reduces the number of loops compared to the old implementation, and does not affect the addition of new rules.

**Optimization of the heuristic optimizer**

> 1. Filter: Reduce the number of loops from 22 to 10
> 2. EvalScalar: Reduce the number of loops from 22 to 2
> 3. Limit: Reduce the number of loops from 22 to 6
> 4. Agg: Reduce the number of loops from 22 to 2
> 5. Sort: Reduce the number of loops from 22 to 1
> 6. Scan: Reduce the number of loops from 22 to 0
> 7. DumpTableScan: Reduce the number of loops from 22 to 0

**The sample code**
some operators are not required to apply the rules:
```
impl Operator for Scan {

    // Returns the set of rules that the operator can apply
    fn transformation_candidate_rules(&self) -> roaring::RoaringBitmap {
        RoaringBitmap::new()
    }

impl Operator for DummyTableScan {

    fn transformation_candidate_rules(&self) -> roaring::RoaringBitmap {
        RoaringBitmap::new()
    }

```
Some operators require only a few rules to be applied: 
```
    // EvalScalar
    EliminateEvalScalar,
    MergeEvalScalar,

    // Sort
    PushDownSortScan,

    // EvalScalar
    EliminateEvalScalar,
    MergeEvalScalar,
```

**And with the continuous increase of rules, it may eventually reach hundreds. With recursive calls, the magnitude of the loop may grow exponentially.So I think it's necessary to just apply it's own rules for each operator.**

**In this way, the number of loops per optimization can be greatly reduced.**

Closes #issue


dusx1981:
So Why Bitmap Management Rules Are Deprecated？

dusx1981:
**Because of the precise management of the application of operator rules，please do a benchmark test:)**
one of the implementations：
```
impl Operator for Limit {
    fn transformation_candidate_rules(&self) -> roaring::RoaringBitmap {
        (RuleID::PushDownLimitUnion as u32..(RuleID::PushDownLimitScan as u32) + 1)
            .collect::<RoaringBitmap>()
    }
```

dusx1981:
**Because of the precise management of the application of operator rules，please do a benchmark test:)**

one of the implementations：
```
impl Operator for Limit {
    fn transformation_candidate_rules(&self) -> roaring::RoaringBitmap {
        (RuleID::PushDownLimitUnion as u32..(RuleID::PushDownLimitScan as u32) + 1)
            .collect::<RoaringBitmap>()
    }
```
[AND I'm collating the preprocessing logic that needs to be done from GPORCA](https://www.processon.com/view/link/640d4bd94ae74f77fb88c0e9)

dusx1981:
Is this performance indicator unimportant, or this implementation did not achieve the expected results, because I can't run the benchmark, could you please give me a feedback on the result?

dusx1981:
image

Xuanwo:
> image

cc @everpcpc for help

everpcpc:
@Xuanwo fixed, retry would help

xudong963:
@mergify update

dusx1981:
@mergify update

dusx1981:
> @mergify update

image

dusx1981:
image

xudong963 (approved):
Changes make sense to me. Sorry for late review, thanks @dusx1981 

Any thoughts? @leiysky

src/query/sql/src/planner/optimizer/cascades/cascade.rs
```diff
@@ -29,36 +30,36 @@ use crate::optimizer::cost::DefaultCostModel;
...
 use crate::IndexType;
-use crate::MetadataRef;
 
 /// A cascades-style search engine to enumerate possible alternations of a relational expression and
 /// find the optimal one.
 pub struct CascadesOptimizer {
-    pub(crate) memo: Memo,
-    pub(crate) cost_model: Box<dyn CostModel>,
+    pub memo: Memo,
+    pub explore_rule_set: roaring::RoaringBitmap,
```
leiysky, line 40: I think we should not to expose the `RoaringBitmap` here, `RuleSet`  is better.

src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
```diff
@@ -113,13 +121,27 @@ impl HeuristicOptimizer {
...
+                Some(self.metadata.clone()),
+            )
+        }
+    }
+
     /// Try to apply the rules to the expression.
     /// Return the final result that no rule can be applied.
     fn apply_transform_rules(&self, s_expr: &SExpr) -> Result<SExpr> {
         let mut s_expr = s_expr.clone();
+        let rule_set = self.calc_operator_rule_set(&s_expr.plan);
```
leiysky, line 141: We should distinguish between `RuleSet` and `RuleList`. `RuleSet` is order-insensitive, and `RuleList` is order-sensitive.

In this case, a `RuleList` is required.

src/query/sql/src/planner/optimizer/cascades/cascade.rs
dusx1981: > I think we'd better not expose the `RoaringBitmap` here, should use `RuleSet` instead.

I understand what you mean, I want to transform the RuleSet to use bitmap management, and then submit a PR. If there is no problem with this PR, can it be passed first？

src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
dusx1981: > We should distinguish between `RuleSet` and `RuleList`. `RuleSet` is order-insensitive, and `RuleList` is order-sensitive.
> 
> In this case, a `RuleList` is required.

Please refer to the introduction of my implementation ideas above. If you use RuleList, if you can quickly calculate the rules that can be applied by each operator, it will cause too many loops and reduce performance.

dusx1981: @leiysky 

Any other question？

src/query/sql/src/planner/optimizer/cascades/tasks/explore_expr.rs
```diff
@@ -140,6 +144,18 @@ impl ExploreExprTask {
         }
     }
 
+    fn calc_operator_rule_set(
+        &self,
+        optimizer: &CascadesOptimizer,
+        operator: &RelOperator,
+    ) -> roaring::RoaringBitmap {
+        unsafe {
```
leiysky, line 152: Why do we need `unsafe` here?

src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
```diff
@@ -113,13 +121,27 @@ impl HeuristicOptimizer {
         Ok(result)
     }
 
+    fn calc_operator_rule_set(&self, operator: &RelOperator) -> roaring::RoaringBitmap {
+        unsafe { operator.transformation_candidate_rules() & (&RULE_FACTORY.transformation_rules) }
```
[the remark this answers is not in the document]
dusx1981: > 

```
// read only, so thread safe
pub static mut RULE_FACTORY: Lazy<RuleFactory> = Lazy::new(RuleFactory::create);
```
RULE_FACTORY is read-only, so it is thread-safe. Here, we need to use unsafe to solve the problem of syntax errors.

src/query/sql/src/planner/optimizer/cascades/tasks/explore_expr.rs
```diff
@@ -149,11 +165,11 @@ impl ExploreExprTask {
             .memo
             .group(self.group_index)?
             .m_expr(self.m_expr_index)?;
-        let rule_set = &optimizer.explore_rule_set;
+        let rule_set = self.calc_operator_rule_set(optimizer, &m_expr.plan);
 
         for rule_id in rule_set.iter() {
             let apply_rule_task = ApplyRuleTask::with_parent(
-                rule_id,
+                unsafe { std::mem::transmute::<u8, RuleID>(rule_id as u8) },
```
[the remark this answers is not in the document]
dusx1981: > Transmuting an `enum` to `u8` is dangerous. Why do we need this here?

Ok, first of all, the current RuleID is safe under this transformation. could you all me to handle it with standard casts  in next PR?

dusx1981:
@leiysky 

[Hasn’t the same code been reviewed by you in this PR?](https://github.com/datafuselabs/databend/pull/10024)

leiysky:
> @leiysky
> 
> [Hasn’t the same code been reviewed by you in this PR?](https://github.com/datafuselabs/databend/pull/10024)

I appreciate your effort, but I must admit that I couldn't accept that pull request due to its low quality.

To be frank, I don't have the energy to point out every issue with your code, and it's unnecessary.

Despite its shortcomings, I recognize the value of your work and have accepted it. However, it has caused a significant impact on the project's maintenance, and as a result, I had to refactor it.

This pull request is merely a minor patch on the previous one and even corrupts the latest codebase.

Although there is always room for improvement, I don't think this is the right way.

dusx1981:
> > @leiysky
> > [Hasn’t the same code been reviewed by you in this PR?](https://github.com/datafuselabs/databend/pull/10024)
> 
> I appreciate your effort, but I must admit that I couldn't accept that pull request due to its low quality.
> 
> To be frank, I don't have the energy to point out every issue with your code, and it's unnecessary.
> 
> Despite its shortcomings, I recognize the value of your work and have accepted it. However, it has caused a significant impact on the project's maintenance, and as a result, I had to refactor it.
> 
> This pull request is merely a minor patch on the previous one and even corrupts the latest codebase.
> 
> Although there is always room for improvement, I don't think this is the right way.

So I feel a little weird about a few points：

- You completely negate your original evaluation, there is no logic at all
image
- You completely ignore my implementation ideas above
- [AND in this pr, I see what you call professionalism: ](https://github.com/datafuselabs/databend/pull/10683)
- You completely negate what other people have said about this project
- You completely ignore the benchmark result

src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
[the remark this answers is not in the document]
Xuanwo: > RULE_FACTORY is read-only, so it is thread-safe. Here, we need to use unsafe to solve the problem of syntax errors.

Doesn't make sense to me.

src/query/sql/src/planner/optimizer/cascades/tasks/explore_expr.rs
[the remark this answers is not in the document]
Xuanwo: Doesn't make sense to use `mem::transmute` here. Also, it should be addressed in this PR.

src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
Xuanwo: > And Please refer to the benchmark results

Could you please provide more details about the benchmark? The benefits appear to be relatively low.

hits: from 1.01 to 0.99

image

tpch: from 1.02 to 0.98

image

Although performance is important, I prefer to prioritize code readability and maintainability. It's not worth sacrificing those aspects for the sake of minor performance improvements.

src/query/sql/src/planner/optimizer/cascades/cascade.rs
Xuanwo: > If there is no problem with this PR, can it be passed first？

Please address it in this PR.

src/query/sql/src/planner/optimizer/cascades/cascade.rs
```diff
@@ -29,36 +30,36 @@ use crate::optimizer::cost::DefaultCostModel;
...
+
+    pub cost_model: Box<dyn CostModel>,
+
     /// group index -> best cost context
-    pub(crate) best_cost_map: HashMap<IndexType, CostContext>,
-    pub(crate) explore_rule_set: RuleSet,
-    pub(crate) metadata: MetadataRef,
+    pub best_cost_map: HashMap<IndexType, CostContext>,
+    _ctx: Arc<dyn TableContext>,
```
Xuanwo, line 46: Introducing `TableContext` in `CascadesOptimizer` may not be a good idea. Additionally, I noticed that this context has been named as `_ctx`. If we do not intend to use it, perhaps it would be better to remove it?

src/query/sql/src/planner/optimizer/cascades/explore_rules.rs
```diff
@@ -12,30 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use roaring::RoaringBitmap;
+
 use crate::optimizer::RuleID;
-use crate::optimizer::RuleSet;
 
-pub fn get_explore_rule_set(enable_bushy_join: bool) -> RuleSet {
+pub fn calc_explore_rule_set(enable_bushy_join: bool) -> RoaringBitmap {
```
Xuanwo, line 19: Please refrain from breaking the current abstraction.

Xuanwo:
> Thank you for your feedback, I also roughly know the workflow of this project, and hope to submit more qualified codes later.
> I think there is no need to discuss this PR anymore, I will close it later.

Thanks for your effort!

src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
dusx1981: > 

I disagree with this point of view. With the increase of rules, each rule will increase the number of cycles optimized by unrelated operators, and it is exponential. When I think of it, the performance comparison will be more obvious.

dusx1981:
> 

On the point of performance, I disagree.. With the increase of rules, each rule will increase the number of cycles optimized by unrelated operators, and it is exponential. When I think of it, the performance comparison will be more obvious.

src/query/sql/src/planner/optimizer/heuristic/heuristic.rs
Xuanwo: > With the increase of rules, each rule will increase the number of cycles optimized by unrelated operators, and it is exponential.

You are correct, but for now, it is not worth it. We can optimize when we identify the bottleneck.

Xuanwo:
> The main purpose of my implementation of these codes is to achieve this optimization.

Thanks for your effort again. We acknowledge that there is still room for improvement in Databend. However, based on the performance reports, this particular aspect is not currently a bottleneck. Therefore, we will prioritize maintaining readability and ease of use.

dusx1981:
> > The main purpose of my implementation of these codes is to achieve this optimization.
> 
> Thanks for your effort again. We acknowledge that there is still room for improvement in Databend. However, based on the performance reports, this particular aspect is not currently a bottleneck. Therefore, we will prioritize maintaining readability and ease of use.

Understood，Thank you.

dusx1981:
Please allow me to add an important information, I learned some key information about real database development from this Review, thank you.

PR #10510, opened 2023-03-12
fix(functions):fix function version with MySQL client

### LINKED ISSUE #10342: the function version() has some problem
**Summary**
the deploy mode is single
d3329f524b9f9557e90302022c4daad7

wangjili8417:

## Summary
The modification method is to delete the matching in MySQL Federated, and instead judge the type of the client in the version function to return the corresponding version information. Clickhouse Federated is also a similar problem, which has been changed by the way.

Closes #10342


TCeason:
Please add some test in sql logic

TCeason:
Like this file tests/sqllogictests/suites/query/case_sensitivity/system_table.test, 

onlyif mysql
query I
select "Engine" as engine from system.engines order by engine
----
FUSE
MEMORY
NULL
RANDOM
VIEW

This select query will only check under mysql handler.

TCeason (approved):

BohuTANG:
```
Diff in /workspace/src/query/service/tests/it/servers/mysql/mysql_federated.rs at line 27:
         assert!(result.is_none());
     }
 
-
     // variables
     {
         let query = "select @@tx_isolation, @@session.tx_isolation";
```

b41sh (approved):

PR #10525, opened 2023-03-13
fix: show process status info

dantengsky:

## Summary

show process status info should use the status info of the process/query context being shown, not the query context of statement `show processlist`

Closes #issue


BohuTANG:
Thanks.
Let me continue this PR :)

Issue #10492, opened 2023-03-10
Feature: support `GROUPING SETS` clause.


RinChanNOWWW:
**Summary**

GROUPING SETS perform the same aggregate across different GROUP BY clauses in a single query.

Related clauses:

- [x] Extended group by: grouping sets. #10534
- [x] Extended group by: cube. #10601
- [x] Extended group by: rollup. #10601
- [x] New function: grouping. #10592
- [x] Docs of the features above. #10638
 
## Reference

- https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-GROUPING-SETS
- https://www.postgresql.org/docs/15/functions-aggregate.html
- https://clickhouse.com/docs/en/sql-reference/statements/select/group-by/#grouping-sets-modifier
- https://duckdb.org/docs/sql/query_syntax/grouping_sets.html
- https://doris.apache.org/zh-CN/community/design/grouping_sets_design/

RinChanNOWWW:
cc @sundy-li

sundy-li:
@RinChanNOWWW 
Do you want to take this task or make it a community task?

GroupSets is easy to implement we just need to union the results I think.

RinChanNOWWW:
I can take it.

RinChanNOWWW:
> GroupSets is easy to implement we just need to union the results I think.

The performance will be bad if we just rewrite it to `UNION`s.

We can extend the aggregation pipeline to achieve this feature.

PR #10534, opened 2023-03-13
feat(query): support `GROUP BY GROUPING SETS` clause.

RinChanNOWWW:

## Summary

Basic implementation of `GROUP BY GROUPING SETS`.

Main idea:

Add a new `PhysicalPlan`: `PhysicalPlan::AggregateExpand` and its corresponding `Transform`: `TransformExpandGroupingSets`.

If the query is a `GROUP BY GROUPING SETS` clause, `PhysicalPlan::AggregateExpand` will be added to the front of `PhysicalPlan::AggregatePartial`. And the plan will be built into `TransformExpandGroupingSets`.

`TransformExpandGroupingSets` will expand the input `DataBlock` for each grouping set. It will fill NULLs if the group-by columns are not found in the grouping set.

Tacking in #10492


sundy-li:
> TransformExpandGroupingSets will expand the input DataBlock for each grouping set. It will fill NULLs if the group-by columns are not found in the grouping set.

What if the original column has nulls? Seem they share the same hashmap here.

sundy-li (approved):
Rest Looks good to me.

src/query/sql/src/planner/binder/aggregate.rs
```diff
@@ -74,6 +78,11 @@ pub struct AggregateInfo {
     /// TODO(leiysky): so far we are using `Debug` string of `Scalar` as identifier,
     /// maybe a more reasonable way is needed
     pub group_items_map: HashMap<String, usize>,
+
+    /// Index for virtual column `grouping_id`. It's valid only if `grouping_sets` is not empty.
+    pub grouping_id_index: IndexType,
```
xudong963, line 83: Better to be `Option<IndexType>`? If `grouping_sets` is empty, it's `None`

RinChanNOWWW: Option<IndexType> will consume more memory.

image

xudong963 (approved):

RinChanNOWWW:
image

cc @BohuTANG

PR #10550, opened 2023-03-14
fix: fix comparison of quantile level

ariesdevil:

## Summary

fix comparison of quantile level

Closes #issue

### DIFF
--- a/src/query/functions/src/aggregates/aggregate_quantile_cont.rs
+++ b/src/query/functions/src/aggregates/aggregate_quantile_cont.rs
@@ -21,12 +21,15 @@ use std::sync::Arc;
 use common_arrow::arrow::bitmap::Bitmap;
 use common_exception::ErrorCode;
 use common_exception::Result;
+use common_expression::type_check::check_number;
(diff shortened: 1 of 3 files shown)


src/query/functions/src/aggregates/aggregate_quantile_cont.rs
```diff
@@ -290,29 +290,29 @@ pub fn try_create_aggregate_quantile_function<const TYPE: u8>(
...
                     return Err(ErrorCode::BadDataValueType(format!(
-                        "level range between 0 to 1, got: {:?}",
+                        "level range between [0, 1], got: {:?}",
                         f
                     )));
                 }
                 f
             }
-            Scalar::Number(NumberScalar::UInt64(i)) => {
+            Scalar::Number(NumberScalar::UInt8(i)) => {
```
sundy-li, line 301: Why not use `check_number` here, see `window_funnel.rs`

ariesdevil: done

PR #10561, opened 2023-03-14
feat: support copy into table with transform.

youngsofun:

## Summary

support copy like

```
copy into test_table from (select (t.id+1), age from @test t)  FILE_FORMAT = (type = parquet)  force=true
```

note:

1. only projection with stage_table  is allowed in the select_stmt.
2. not allow options for stage_table,  always put them in copy_stmt.
3. must check colored files before infer_schema (so we infer on the right file), and  infer_schema is required in binding, so list_file and color_file is advanced to the binding phase, but read parquet meta is still in read_partition and list_file stage_table function is not advanced so that user can get error msg earlier instead of waiting for list_file internal.

todo:

- [x] add tests

not in pr:

-  copy  from URI


youngsofun:
file seems not deleted when drop stage.

create a stage with same name may find old files there.

src/query/sql/src/planner/binder/copy.rs
```diff
@@ -464,6 +495,148 @@ impl<'a> Binder {
         })))
     }
 
+    /// Bind COPY INFO <table> FROM <query>
```
ariesdevil, line 498:
```suggestion
    /// Bind COPY INTO <table> FROM <query>
```

ariesdevil (approved):

src/query/sql/src/planner/binder/copy.rs
youngsofun: fixed, thanks

tests/suites/1_stateful/08_select_stage/08_00_parquet/08_00_06_transform.sh
```diff
@@ -0,0 +1,30 @@
...
+DATADIR_PATH="/tmp/08_00_00"
+rm -rf ${DATADIR_PATH}
+DATADIR="fs://$DATADIR_PATH/"
+echo "copy into '${DATADIR}' from t1 FILE_FORMAT = (type = PARQUET);" | $MYSQL_CLIENT_CONNECT
+
+echo "drop stage if exists s2;" | $MYSQL_CLIENT_CONNECT
+echo "create stage s2 url = '${DATADIR}' FILE_FORMAT = (type = PARQUET);"  | $MYSQL_CLIENT_CONNECT
+
+echo '--- copy 1'
+echo "copy into t1 from (select (t.id+1), age from @s2 t)  FILE_FORMAT = (type = parquet);" | $MYSQL_CLIENT_CONNECT
```
BohuTANG, line 19: We only support parquet format, how about add an error test for:
copy into t1 from (select (t.id+1), age from @s2 t)  FILE_FORMAT = (type = csv)?

youngsofun: done

src/query/sql/src/planner/binder/copy.rs
```diff
@@ -464,6 +495,148 @@ impl<'a> Binder {
...
+
+        let start = Instant::now();
+        {
+            let status = "begin to list files";
+            self.ctx.set_status_info(status);
+            info!(status);
+        }
+
+        let operator = init_stage_operator(&stage_info)?;
+        let files = if operator.info().can_blocking() {
```
BohuTANG, line 544: Not sure if the heavy IO in the planning binder is ok or not? cc @leiysky @sundy-li

youngsofun: list s3

sundy-li: It's already in async runtime. we don't need to use `blocking_list` I think.

youngsofun: maybe faster for local fs @RinChanNOWWW ？

youngsofun:
ydb standalone timeout

src/query/sql/src/planner/binder/copy.rs
youngsofun: > Not sure if the heavy IO in the planning binder is ok or not? cc @leiysky @sundy-li

no real harm for now I think.

the only other choice is delay binding and every thin to pipeline building,  which is worse.

sundy-li:
Rest Looks good to me.

sundy-li (approved):

src/query/sql/src/planner/binder/copy.rs
RinChanNOWWW: I think `blocking_list` can let it not yield the CPU and maybe faster for local fs.

RinChanNOWWW:
What if copy from the same stage table with different SQLs?

For example 

```sql
copy into t1 from (select t.id, age from @s2 t)
-- and
copy into t1 from (select (t.id+1), age from @s2 t)
```

Will this trigger `no file need to copy` ?

youngsofun:
> What if copy from the same stage table with different SQLs?
> 
> For example
> 
> ```sql
> copy into t1 from (select t.id, age from @s2 t)
> -- and
> copy into t1 from (select (t.id+1), age from @s2 t)
> ```
> 
> Will this trigger `no file need to copy` ?

yes

youngsofun:
@BohuTANG 

```
This pull request cannot be embarked for merge
The merge queue pull request can't be updated
Details:

Unable to update: user youngsofun is unknown.

Please make sure youngsofun has logged in Mergify dashboard.
```

PR #10562, opened 2023-03-14
feat(planner): plan window function

ariesdevil:

## Summary

initial impl of window function plan.

Closes #issue

### DIFF
--- a/src/query/sql/src/planner/binder/aggregate.rs
+++ b/src/query/sql/src/planner/binder/aggregate.rs
@@ -156,7 +156,9 @@ impl<'a> AggregateRewriter<'a> {
 
             ScalarExpr::AggregateFunction(agg_func) => self.replace_aggregate_function(agg_func),
 
-            ScalarExpr::WindowFunction(_) => unreachable!(),
+            ScalarExpr::WindowFunction(_) => Err(ErrorCode::SemanticError(
+                "bind aggregate functions should not reach to window functions.",
+            )),
         }
     }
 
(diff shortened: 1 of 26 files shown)


src/query/sql/src/planner/binder/aggregate.rs
```diff
@@ -55,6 +57,9 @@ pub struct AggregateInfo {
     /// Arguments of aggregation functions
     pub aggregate_arguments: Vec<ScalarItem>,
 
+    /// infos for window with aggregate function
+    pub window_info: Option<WindowInfo>,
```
leiysky, line 61: There will be multiple windows if there are window functions with different partition definitions? For example:
```sql
SELECT count(*) over(partition by a), count(*) over(partition by b) FROM t
```

src/query/sql/src/planner/binder/window.rs
```diff
@@ -0,0 +1,379 @@
...
+                                ),
+                            });
+                            found = true;
+                            break;
+                        }
+                    }
+
+                    if found {
+                        continue;
+                    }
```
leiysky, line 129: I think we don't need to do this for `order by` in window functions. It's different from `order by` clause in the query body, we can just throw an error if no matched column is found.

src/query/sql/src/planner/binder/aggregate.rs
```diff
@@ -152,6 +152,8 @@ impl<'a> AggregateRewriter<'a> {
             ScalarExpr::SubqueryExpr(_) => Ok(scalar.clone()),
 
             ScalarExpr::AggregateFunction(agg_func) => self.replace_aggregate_function(agg_func),
+
+            ScalarExpr::WindowFunction(_) => unreachable!(),
```
xudong963, line 156: better to return err info

ariesdevil:
If there are no obvious problems, let's merge this pr first and I'll move on to the next.

xudong963 (approved):

sundy-li (approved):

PR #10565, opened 2023-03-14
docs: move performance page to blog

soyeric128:

## Summary

moved performance page to blog.

Closes #issue


soyeric128:
@BohuTANG I will remove https://databend.rs/doc/performance/ from documentation after this blog goes up, as the rs homepage has a link to it:

Screenshot 2023-03-14 at 16 38 25

BohuTANG:
Doc removed the performance page in:https://github.com/datafuselabs/databend/pull/10573

And we should keep the blog slug more human-readable, so I changed the slug too

PR #10575, opened 2023-03-15
refactor(query): delete async insert config

TCeason:

## Summary

As title.

Closes #issue


src/query/config/src/config.rs
```diff
@@ -1254,6 +1242,18 @@ pub struct QueryConfig {
 
     // ----- the following options/args are all deprecated               ----
     // ----- and turned into Option<T>, to help user migrate the configs ----
+    /// OBSOLETED: The maximum memory size of the buffered data collected per insert before being inserted.
+    #[clap(long)]
```
BohuTANG, line 1246: No need to be OBSOLETED, only remove them.
As we know, there is no user used them.

PR #10594, opened 2023-03-15
refactor(expr): make `if` a special builtin function for evalutor

andylokandy:

## Summary

A slightly prettier approach to #10439


sundy-li:
`if` is used more commonly than `multi_if`.

1. How about making the function name to be `if` ?
2. Is this change slow down the `if` function?  

Need benchmark to verify, eg:

```
select sum( if(number % 7 = 0, 1, 2) ) from numbers(1000000000);
```

andylokandy:
After mentally analyzing the `multi_if`, I found that it should perform just as well as the original `if` statement. This was confirmed by benchmark testing.

`if` in master
```
Time: 1.266s
Time: 1.253s
Time: 1.212s
```

`multi_if` in this pr
```
Time: 1.234s
Time: 1.230s
Time: 1.273s
```

andylokandy:
@sundy-li currently, `if` is just an alias of `multi_if`. do you want to rename them all to `if`?

sundy-li:
> @sundy-li currently, `if` is just an alias of `multi_if`. do you want to rename them all to `if`?

Yes, i think it's better to name `if`, `if` contains `if` and `multi_if`.

PR #10596, opened 2023-03-16
chore(storage): const predication returns empty partitions

sundy-li:

## Summary

Summary about this PR

```
select * from table where 1 = 2 
```

will not read segments anymore.

Closes #issue

### DIFF
--- a/src/query/storages/fuse/src/operations/read_partitions.rs
+++ b/src/query/storages/fuse/src/operations/read_partitions.rs
@@ -27,6 +27,8 @@ use common_catalog::plan::TopK;
 use common_catalog::table::Table;
 use common_catalog::table_context::TableContext;
 use common_exception::Result;
+use common_expression::RemoteExpr;
+use common_expression::Scalar;
 use common_expression::TableSchemaRef;
 use common_meta_app::schema::TableInfo;
 use common_storage::ColumnNodes;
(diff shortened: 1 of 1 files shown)


src/query/storages/fuse/src/operations/read_partitions.rs
```diff
@@ -57,6 +59,18 @@ impl FuseTable {
     ) -> Result<(PartStatistics, Partitions)> {
         debug!("fuse table do read partitions, push downs:{:?}", push_downs);
 
+        if let Some(PushDownInfo {
```
BohuTANG, line 62: Is it possible for the optimizer to check and return an empty result when `where 1=2`, even though we know that no records meet this condition?

sundy-li: There's no need to do this optimization. It's fast enough now (0.01s).

BohuTANG: It's not an optimization.
This PR only fixed the fuse engine, not the other engines we have(perhaps have in the future) :)

BohuTANG (approved):

src/query/storages/fuse/src/operations/read_partitions.rs
BohuTANG: mysql> select sum(number) from numbers(100000000000);
+--------------------+
| sum(number)        |
+--------------------+
| 932355974711512064 |
+--------------------+
1 row in set (3.33 sec)
Read 100000000000 rows, 745.06 GiB in 3.319 sec., 30.13 billion rows/sec., 224.47 GiB/sec.

mysql> select sum(number) from numbers(100000000000) where 1=2;
+-------------+
| sum(number) |
+-------------+
|        NULL |
+-------------+
1 row in set (2.40 sec)
Read 100000000000 rows, 745.06 GiB in 2.396 sec., 41.74 billion rows/sec., 310.97 GiB/sec.

BohuTANG:
```
[SQL] explain select * from numbers(1) where number = 0 and false
[Diff] (-expected|+actual)
    Filter
    ├── filters: [false]
    ├── estimated rows: 0.00
    └── TableScan
        ├── table: default.system.numbers
-       ├── read rows: 1
-       ├── read bytes: 8
-       ├── partitions total: 1
-       ├── partitions scanned: 1
+       ├── read rows: 0
+       ├── read bytes: 0
+       ├── partitions total: 0
+       ├── partitions scanned: 0
        ├── push downs: [filters: [false], limit: NONE]
        └── estimated rows: 1.00
at tests/sqllogictests/suites/mode/standalone/explain/select.test:94
```

PR #10609, opened 2023-03-16
refactor: Merge FileWithMetadata into StageFileInfo

Xuanwo:

## Summary

refactor: Merge FileWithMetadata into StageFileInfo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2063,6 +2063,7 @@ dependencies = [
  "common-exception",
  "common-expression",
  "common-meta-app",
+ "flagset",
  "futures",
  "opendal",
  "regex",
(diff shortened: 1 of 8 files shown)


Xuanwo:
> I prefer turn Metadata into StageFileInfo only when necessary.

Great catch! I understand and agree with your suggestion to turn Metadata into StageFileInfo when necessary. However, I will need some time to review the logic and come up with a better solution.

As this pull request does not have any negative impact (since we used to receive full metadata before), how about merging it first and then refining it later?

BohuTANG (approved):

BohuTANG:
Will merge, then we can refine it in another PR.

sundy-li (approved):

PR #10620, opened 2023-03-17
fix(storage): compact wrong row count bug

zhyass:

## Summary

Record location when read segment to build compact tasks.
Add tests

Closes #issue

### DIFF
--- a/src/query/storages/fuse/src/operations/mutation/compact/block_compact_mutator.rs
+++ b/src/query/storages/fuse/src/operations/mutation/compact/block_compact_mutator.rs
@@ -114,18 +114,22 @@ impl BlockCompactMutator {
             // Check the segment to be compacted.
             // Size of compacted segment should be in range R == [threshold, 2 * threshold)
             for (idx, segment) in segment_infos.iter().enumerate() {
-                let segments_vec = checker.add(segment.clone());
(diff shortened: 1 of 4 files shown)


src/query/storages/fuse/src/operations/mutation/compact/block_compact_mutator.rs
```diff
@@ -237,14 +244,19 @@ impl BlockCompactMutator {
...
-                    .map_or((0, vec![]), |(k, v)| (k, vec![v])),
-            );
+                    .map_or((0, vec![]), |(k, v)| (k, vec![v]))
+            };
             blocks.extend(builder.take_blocks());
             if blocks.len() > 1 || builder.check_column_ids(&blocks[0]) {
                 tasks.push_back((index, blocks));
+            } else {
+                // unreachable.
+                unchanged_blocks.insert(index, blocks[0].clone());
```
dantengsky, line 259: is here unreachable?

zhyass: > is here unreachable?

Yes, here must be more than one blocks.

dantengsky: got it

PR #10633, opened 2023-03-18
feat(aggregate function): quantile_cont supports multiple levels

ariesdevil:

## Summary

quantile supports multiple levels

Closes #issue


sundy-li:
Seems the function should be renamed to `quantile_cont ` , because it's intepolated quantile number.

Another function is `quantile_disc ` (alias `quantile`), which returns exact quantile number.

 
image

ariesdevil:
@sundy-li I refactor some codes by using interpolation algorithm to compute quantile, the results are same as duckdb now.

sundy-li (approved):

PR #10314, opened 2023-03-02
chore(planner): Refactor join operator

leiysky:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

PR #10635, opened 2023-03-18
refactor(planner): minimize ScalarExpr

andylokandy:
- refactor(planner): minimize ScalarExpr
- improve

## Summary

Summary about this PR

Closes #issue


sundy-li (approved):
Nice simplification!

xudong963 (approved):
After the ticket, I think tpcds q4 stackoverflow will be solved, please check again @RinChanNOWWW

BohuTANG:
Conflicting files
src/query/sql/src/planner/semantic/type_check.rs

RinChanNOWWW:
> After the ticket, I think tpcds q4 stackoverflow will be solved, please check again @RinChanNOWWW

It seems that it will also stackoverflow, because the recursion counts are the same.

leiysky (approved):

andylokandy:
This PR hit on a weird bug that could be solved by https://github.com/datafuselabs/databend/pull/10314

src/query/sql/src/planner/optimizer/heuristic/subquery_rewriter.rs
```diff
@@ -471,7 +471,7 @@ impl SubqueryRewriter {
                             span: subquery.span,
                             func_name: op.to_func_name().to_string(),
                             params: vec![],
-                            arguments: vec![left_condition, right_condition],
+                            arguments: vec![right_condition, left_condition],
```
RinChanNOWWW, line 474: Why change the order?

andylokandy: It was originally reverted

PR #10640, opened 2023-03-18
feat(front): Add [fetch-databend-releases] plugin and refactor download page

Chasen-Zhang:

## Summary

Releases resources are already obtained when the page is packaged, so the front-end page does not need to send a request again, and the page loading speed is increased by 100%, And don't worry about the limitations of Github API requests！

### Use the following hooks code:
```js
import  useGetReleases  from '@site/src/hooks/useGetReleases';


const {  releasesList, tagName, stargazersCount, formatStargazersCount, repoResource } = useGetReleases();
```

Choose the fields you need.


BohuTANG (approved):
👍

PR #10656, opened 2023-03-19
fix(exclude): column name case issue

SkyFan2002:

## Summary
The results of these two SQL statements were different before, and this PR fixed it
```sql
select * exclude id from t;
select * exclude ID from t;
```

### DIFF
--- a/src/query/sql/src/planner/binder/project.rs
+++ b/src/query/sql/src/planner/binder/project.rs
@@ -165,7 +165,7 @@ impl Binder {
                     let mut exclude_cols: HashSet<String> = HashSet::new();
                     if let Some(cols) = exclude {
                         for col in cols {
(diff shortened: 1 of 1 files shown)


BohuTANG:
It would be better to add a logic SQL test for this PR

sundy-li:
Maybe we should respect:

https://github.com/datafuselabs/databend/blob/5e2090850bc3a4b6b85f3b7bd904c5ad8abab177/src/query/settings/src/lib.rs#L282-L301

sundy-li (approved):

xudong963 (approved):

sundy-li:
@mergify update

sundy-li:
@SkyFan2002  Please run `make lint`

PR #10661, opened 2023-03-20
feat(query): record copied files for insert with stage

everpcpc:

## Summary

after https://github.com/datafuselabs/databend/pull/10427

Closes #issue

### DIFF
--- a/src/query/service/src/interpreters/interpreter_insert.rs
+++ b/src/query/service/src/interpreters/interpreter_insert.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::BTreeMap;
 use std::collections::VecDeque;
 use std::io::BufRead;
 use std::io::Cursor;
@@ -21,6 +22,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use aho_corasick::AhoCorasick;
+use chrono::Utc;
 use common_ast::ast::Expr as AExpr;
 use common_ast::parser::parse_comma_separated_exprs;
 use common_ast::parser::parse_expr;
@@ -52,6 +54,8 @@ use common_io::cursor_ext::ReadCheckPointExt;
 use common_meta_app::principal::FileFormatOptions;
 use common_meta_app::principal::StageFileFormatType;
 use common_meta_app::principal::StageInfo;
+use common_meta_app::schema::TableCopiedFileInfo;
+use common_meta_app::schema::UpsertTableCopiedFileReq;
 use common_pipeline_core::Pipeline;
 use common_pipeline_sources::AsyncSource;
 use common_pipeline_sources::AsyncSourcer;
@@ -72,6 +76,7 @@ use common_sql::MetadataRef;
 use common_sql::NameResolutionContext;
 use common_sql::ScalarBinder;
 use common_storage::StageFileInfo;
+use common_storage::StageFileStatus;
 use common_storage::StageFilesInfo;
 use common_storages_factory::Table;
 use common_storages_fuse::io::Files;
@@ -203,7 +208,8 @@ impl InsertInterpreter {
         let table_ctx: Arc<dyn TableContext> = ctx.clone();
         let source_schema = self.plan.schema();
         let catalog_name = self.plan.catalog.clone();
-        let overwrite = self.plan.overwrite;
+        let database_name = self.plan.database.clone();
+        let table_name = self.plan.table.clone();
 
         let (attachment_data_schema, const_columns) = if attachment.values_str.is_empty() {
             (source_schema.clone(), vec![])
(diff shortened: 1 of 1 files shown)


src/query/service/src/interpreters/interpreter_insert.rs
```diff
@@ -309,6 +334,20 @@ impl InsertInterpreter {
                     let append_entries = ctx.consume_precommit_blocks();
                     // We must put the commit operation to global runtime, which will avoid the "dispatch dropped without returning error" in tower
                     GlobalIORuntime::instance().block_on(async move {
+                        let mut copied_files = BTreeMap::new();
+                        for file in need_copy_files {
+                            // Short the etag to 7 bytes for less space in metasrv.
```
Xuanwo, line 339: This doesn't seem safe.

everpcpc: cc @youngsofun

BohuTANG: Can you convert the `insert ... attach file` to `copy`?

youngsofun: copy will record the file metadata in the meta server.
insert_with_attach  is an internal optimization for insert, 
the file is temporary.

youngsofun:
@everpcpc when does a user need  `insert with stage `  instead of copy?

everpcpc:
> @everpcpc when does a user need `insert with stage ` instead of copy?

Users do not use insert with stage directly, they use insert with batch instead, and we insert the batched values with stage in our drivers, without changing the user sql statement.

flaneur2020:
> @everpcpc when does a user need `insert with stage ` instead of copy?

Insert with stage mainly targets the users on writing application with our SDK, likewise you can write the following code in snowflake sdk:

```
// Create a table containing an integer, float, boolean, and string column.
_, err = db.Exec("create or replace table my_table(c1 int, c2 float, c3 boolean, c4 string)")
...
// Define the arrays containing the data to insert.
intArray := []int{1, 2, 3}
fltArray := []float64{0.1, 2.34, 5.678}
boolArray := []bool{true, false, true}
strArray := []string{"test1", "test2", "test3"}
...
// Insert the data from the arrays and wrap in an Array() function into the table.
_, err = db.Exec("insert into my_table values (?, ?, ?, ?)", Array(&intArray), Array(&fltArray), Array(&boolArray), Array(&strArray))
```

all the user need is prepare the data batch in their favorite language.

on the other hand, the snowflake sdk will batch these data into a temp stage file in the background, the users do not need to care about the details on 1. presign a staged file; 2. upload a staged file; 3. copy into, just a insert with placeholders.

https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Binding_Parameters_to_Array_Variables

youngsofun:
so we do not need to `record copied files`?

https://github.com/datafuselabs/databend/pull/10661#discussion_r1142891637

flaneur2020:
I guess there's still some benefit on allowing `insert with stage` to be idempotent.

like the following:

1. users uploaded the file to stage
2. insert with stage
3. <-- network lost between sdk and databend-query -->

if we got the network issue in step 3, the client cannot get whether the file have ingested successfully or not, a retry is not feasible without the idempotent semantic, as may causes duplications.

youngsofun:
for now,  we do not support ErrorMode of `skip by file`
if tmp files are always purged,
we can check if these files exist (either all exists, or all purged).

youngsofun:
if tmp files are always purged, client can remember the query_id, just check if it  has succeed is enough.
even when the conn is lost

flaneur2020:
> if tmp files are always purged, client can remember the query_id, just check if it has succeed is enough. even when the conn is lost
> 
> if failed, client may choose to retry

this is a feasible approach, we can also get idempotent by remembering the recent committed query ids. I believe idempotent by query id is a more lightweight solution than remembering the file names.

how do we remember the recent committed query ids in databend right now?

youngsofun:
history？

flaneur2020:
can we guarantee the idempotent semantic by query id at the current version of databend? 🤔

youngsofun:
> can we guarantee the idempotent semantic by query id at the current version of databend? 🤔

Thus, the probability to find a duplicate within 103 trillion version-4 UUIDs is one in a billion.


https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions

flaneur2020:
> 

UUID collision is not a problem, but we have to remember the committed query id in some durable storage with some transaction to guarantee the idempotent semantic.

src/query/service/src/interpreters/interpreter_insert.rs
flaneur2020: i guess the insert ... attach file can share the same logic with the `copy` statement ?

youngsofun:
> > 
> 
> UUID collision is not a problem, but we have to remember the committed query id in some durable storage with some transaction to guarantee the idempotent semantic.

yes,  currently history is not persisted? it should be fixed since there may be other things that depend on it? 
cc @BohuTANG

src/query/service/src/interpreters/interpreter_insert.rs
youngsofun: yes, but the detail depends on whether we  `need to record copied files`

BohuTANG:
Hi, @everpcpc 
What's the progress of this PR, we do not seem to be working on it anymore, ready to close?

PR #10650, opened 2023-03-19
fix: avoid dependency loop when selecting from view.

RinChanNOWWW:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Check dependency when selecting from view.

Closes #10625

PR #10670, opened 2023-03-20
fix(query): create view check unknown table

### LINKED ISSUE #10669: bug: need check table exists when create view
Can we check the loop when create view ?

```sql
postgres=# create view v_t as select * from v_t;
ERROR:  relation "v_t" does not exist
LINE 1: create view v_t as select * from v_t;

```

_Originally posted by @TCeason in https://github.com/datafuselabs/databend/pull/10650#discussion_r1141341462_

TCeason:

## Summary

Like this:

```
postgres=# create view v_t as select * from v_t;
ERROR:  relation "v_t" does not exist
LINE 1: create view v_t as select * from v_t;
```

Note:

> QUERY can query table_functions but only DatabaseCatalog impl list_table_functions. HiveCatalog and IcebergCatalog are unimplemented!(). So I try to modify them to return `vec![]`.

Closes #10669

### DIFF
--- a/src/query/service/src/interpreters/interpreter_view_create.rs
+++ b/src/query/service/src/interpreters/interpreter_view_create.rs
@@ -21,6 +21,7 @@ use common_meta_app::schema::CreateTableReq;
 use common_meta_app::schema::TableMeta;
 use common_meta_app::schema::TableNameIdent;
 use common_sql::plans::CreateViewPlan;
+use common_sql::plans::Plan;
 use common_sql::Planner;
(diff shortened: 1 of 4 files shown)


RinChanNOWWW:
So we can remove the codes in #10650.

TCeason:
> So we can remove the codes in #10650.

I think these codes also make sense, so how about keeping them in some last version? This pr is only valid for the next version. It can not cover this case: someone already created a view with an unknown table.

RinChanNOWWW (approved):

PR #10683, opened 2023-03-20
refactor(planner): Refactor set returning function

leiysky:

## Summary

Introduce a framework to register and provide type checking for SRFs.

It also refines the semantic check of SRFs.


src/query/functions/src/srfs/register.rs
```diff
@@ -0,0 +1,186 @@
...
+            "unnest",
+            &arg_types,
+            &[DataType::Nullable(Box::new(DataType::Generic(0)))],
+            unnest_impl,
+        );
+    }
+
+    {
+        // Unnest 9d-array
+        let arg_types = vec![DataType::Array(Box::new(DataType::Array(Box::new(
```
sundy-li, line 151: 8d is missing ?

src/query/functions/src/srfs/evaluator.rs
```diff
@@ -0,0 +1,74 @@
...
+            .map(|arg| evaluator.run(arg))
+            .collect::<Result<Vec<_>>>()?;
+        let arg_result_refs = arg_results
+            .iter()
+            .map(|arg| arg.as_ref())
+            .collect::<Vec<_>>();
+
+        let results = (0..self.input_columns.num_rows())
+            .map(|i| {
+                let args = arg_result_refs
```
sundy-li, line 59: It may work less efficiently than the previous implementation.

src/query/sql/src/executor/physical_plan_visitor.rs
```diff
@@ -233,12 +233,12 @@ pub trait PhysicalPlanReplacer {
         )))
     }
 
-    fn replace_unnest(&mut self, plan: &Unnest) -> Result<PhysicalPlan> {
+    fn replace_unnest(&mut self, plan: &ProjectSet) -> Result<PhysicalPlan> {
```
RinChanNOWWW, line 236:
```suggestion
    fn replace_project_set(&mut self, plan: &ProjectSet) -> Result<PhysicalPlan> {
```

src/query/functions/src/srfs/register.rs
RinChanNOWWW: Why not making a common logic for any dimension?

src/query/functions/src/srfs/evaluator.rs
leiysky: I have no idea about this, do you have any suggestions?

src/query/functions/src/srfs/register.rs
leiysky: > 8d is missing ?

Fixed.

src/query/functions/src/srfs/evaluator.rs
RinChanNOWWW: I think we can let all SRFs return `ArrayColumn` (or column with offsets).

leiysky: > I think we can let all SRFs accept the whole arg columns and return `ArrayColumn` (or column with offsets).

That's a good idea, I will try it out.

src/query/sql/src/evaluator/block_operator.rs
```diff
@@ -88,201 +85,117 @@ impl BlockOperator {
...
-            })
-            .collect::<Vec<_>>();
+                                                Value::<AnyType>::Column(column) => {
+                                                    let nullable_column =
+                                                        column.as_nullable().unwrap().clone();
+                                                    let mut column_builder =
+                                                        NullableColumnBuilder::from_column(
+                                                            *nullable_column,
+                                                        );
+                                                    (0..(max_num_rows - result_set_rows)).for_each(
```
RinChanNOWWW, line 165: Maybe operating the inner column and bitmap is more effective.

src/query/sql/src/evaluator/block_operator.rs
```diff
@@ -88,201 +85,117 @@ impl BlockOperator {
...
-        let meta = input.get_meta().cloned();
+                    for entry in input.columns() {
+                        // Take the i-th row of input data block and add it to the row.
+                        let mut builder =
+                            ColumnBuilder::with_capacity(&entry.data_type, max_num_rows);
+                        let scalar_ref = entry.value.index(i).unwrap();
+                        (0..max_num_rows).for_each(|_| {
+                            builder.push(scalar_ref.clone());
+                        });
+                        row.push(BlockEntry {
```
RinChanNOWWW, line 131: These codes will replicate much useless data (for example: the args of the SRF, the columns need to be removed after projection). How about replace the SRF columns in place.

leiysky: The arguments of SRF won't be put into the `DataBlock`, only the columns that need to be pass through `ProjectSet` will be replicated.

src/query/sql/src/evaluator/block_operator.rs
leiysky: I'll try it out.

Issue #8594, opened 2022-11-02
Feature: Distributed `COPY INTO`


yufan022:
**Summary**

Should we support distributed `copy into from s3 distributed=true` in the future?

If we use `copy into from s3 purge=true` to import data simultaneously in many query-node, we must very careful to set `s3-prefix or pattern` so that different nodes use different directories to avoid repeatedly importing.

like this:
```
#query1
copy into xx from 's3://query1/' pattern ='.*[.]tsv' purge=true;
#query2
copy into xx from 's3://query2/' pattern ='.*[.]tsv' purge=true;
#query3
copy into xx from 's3://query3/' pattern ='.*[.]tsv' purge=true;
```

If we support distributed `copy into`, things will be easier.

we just need to use:
```
copy into xx from 's3://import/' pattern ='.*[.]tsv' purge=true distributed=true;
```

Xuanwo:
https://github.com/datafuselabs/databend/issues/6395 will make COPY run in a cluster so that copy will be distributed automatically.

PsiACE:
https://github.com/datafuselabs/databend/pull/11943

PR #10688, opened 2023-03-21
feat(query): support parser `copy into distributed`

TCeason:

## Summary

 support parser `copy into distributed`

Part of #8594


sundy-li:
I think `distributed` should not be an option , it'll enabled by default in cluster deployment.

Xuanwo:
Hello, why we need a new option to make COPY distributed? Can you elaborate more details?

TCeason:
> I think `distributed` should not be an option , it'll enabled by default in cluster deployment.

Good idea. Will close this pr.

TCeason:
> Hello, why we need a new option to make COPY distributed? Can you elaborate more details?

Base idea is in there:

https://github.com/datafuselabs/databend/issues/8594#issue-1432471562

But now, I think as a default set enable on cluster mode is better.

PR #10693, opened 2023-03-21
fix: fix Nullable(Decimal) division.

### LINKED ISSUE #10691: Bug: "attempt to divide by zero" raised if the value is NULL in `Nullable(Decimal)` column.
**Summary**

```sql
create table test (a int, b decimal(7,2));

insert into test values(1,2),(3,4);

select if(a=1,b,null)/if(a=1,b,null) from test;
ERROR 1105 (HY000): Code: 1104, displayText = attempt to divide by zero.

select if(number=1,number+1,null)/if(number=1,(number+1)::decimal(7,2),null) from numbers(2);
ERROR 1105 (HY000): Code: 1104, displayText = attempt to divide by zero.
```

Expect result:
```
1.0
NULL
```

RinChanNOWWW:

## Summary

- Fix1: not all branches checked if the divisor is `Decimal::zero`.
- Fix2: `wrap_nullable` for function didn't check if the `errors` in `EvalContext` is `Some`.

Closes #10691


tests/sqllogictests/suites/base/11_data_type/11_0006_data_type_decimal
```diff
@@ -890,3 +897,9 @@ drop table if exists t2
 
 statement ok
 drop table if exists desc_decimal
+
+query T
+select if(number=1,number+1,null)/if(number=1,(number+1)::decimal(7,2),null) from numbers(2);
+----
+NULL
+1.00
```
TCeason, line 905: New line

src/query/expression/src/function.rs
```diff
@@ -436,10 +436,19 @@ where F: Fn(&[ValueRef<AnyType>], &mut EvalContext) -> Value<AnyType> {
...
+            .map(|(b, _)| b.clone())
+            .unwrap_or_else(|| constant_bitmap(true, len))
+            .into();
+        ctx.errors = None;
+        bitmap = bitmap & &error_bitmap;
+
         match results {
             Value::Scalar(s) => {
-                if bitmap.get(0) {
+                if bitmap.get(0) && ctx.errors.is_none() {
```
sundy-li, line 451: do not  need to check `ctx.errors` here.

we should update `ctx.errors` by bitmap.

If bitmap is null at some row but it's an error there, the error should be reset.

Evaluator will check `ctx.error` and render it.

sundy-li (approved):

TCeason (approved):

RinChanNOWWW:
@mergify refresh

RinChanNOWWW:
@mergify refresh

RinChanNOWWW:
@mergify refresh

Issue #10400, opened 2023-03-07
Tracking: support set returning function `generate_series `


jun0315:
**Summary**
An SRF is a function that returns a set of rows instead of a single value.

If we are using `generate_series` in FROM clause, it will be treated as a table.

Some use cases:

```sql
SELECT * FROM generate_series(2,4);
 generate_series
-----------------
               2
               3
               4
(3 rows)

SELECT * FROM generate_series(5,1,-2);
 generate_series
-----------------
               5
               3
               1
(3 rows)

SELECT * FROM generate_series(4,3);
 generate_series
-----------------
(0 rows)

SELECT generate_series(1.1, 4, 1.3);
 generate_series 
-----------------
             1.1
             2.4
             3.7
(3 rows)
```

### Related 
1. https://github.com/datafuselabs/databend/issues/8437
2. https://github.com/datafuselabs/databend/issues/10295

### Tasks
- [ ] `generate_series` support table function. https://github.com/datafuselabs/databend/pull/10697

### References
[Set Returning Functions- PostgreSQL](https://postgrespro.com/docs/postgresql/10/functions-srf)

PR #10697, opened 2023-03-21
feat(query): table function generate_series & range

jun0315:

## Summary
generate_series can be a table function.

support Number, Timestamp and Date argument type.

`range` is special case for `generate_series`, it's similar to python's range function.
Yet `generate_series` is inclusive_bound and `range` is exclusive_bound.

Tracking https://github.com/datafuselabs/databend/issues/10400


src/query/service/src/table_functions/srf/generate_series.rs
```diff
@@ -0,0 +1,381 @@
...
+            Ok(Some(DataBlock::new_from_columns(vec![
+                Int64Type::from_data(range(start, end, step).into_iter()),
+            ])))
+        }
+    }
+}
+
+pub fn number_scalars_have_float(scalars: Vec<Scalar>) -> bool {
+    for scalar in scalars {
+        match scalar {
```
sundy-li, line 363: `matches!(...)`

src/query/service/src/table_functions/srf/generate_series.rs
```diff
@@ -0,0 +1,381 @@
...
+        if self.finished {
+            return Ok(None);
+        }
+        let have_float = number_scalars_have_float(vec![
+            self.start.clone(),
+            self.end.clone(),
+            self.step.clone(),
+        ]);
+        if have_float {
+            let dest_type = DataType::Number(NumberDataType::Float64);
```
sundy-li, line 255: 1. better to have a `macro` to make the codes look better.
2. `date`, `timestamp` should be covered.

src/query/service/src/table_functions/srf/generate_series.rs
RinChanNOWWW: How about add a method `is_float` for `Scalar`.

And check `if arg1.is_float() || arg2.is_float() || arg3.is_float()`.

src/query/service/src/table_functions/srf/generate_series.rs
```diff
@@ -0,0 +1,381 @@
...
+                NumberScalar::UInt64(_) => (),
+                NumberScalar::Int8(_) => (),
+                NumberScalar::Int16(_) => (),
+                NumberScalar::Int32(_) => (),
+                NumberScalar::Int64(_) => (),
+                NumberScalar::Float32(_) => return true,
+                NumberScalar::Float64(_) => return true,
+            },
+            Scalar::Decimal(_) => return true,
+            _ => (),
```
RinChanNOWWW, line 377:
```suggestion
        match scalar {
            Scalar::Number(NumberScalar::Float32(_)) | Scalar::Number(NumberScalar::Float32(_)) |  Scalar::Decimal(_) =>  return true,
            _ => {},
```

src/query/service/src/table_functions/srf/generate_series.rs
```diff
@@ -379,3 +358,27 @@ pub fn number_scalars_have_float(scalars: Vec<Scalar>) -> bool {
...
+        table_args.positioned.len(),
+        Some((2, 3)),
+        2,
+    )?;
+
+    if args
+        .iter()
+        .all(|arg| matches!(Scalar::Number(_) | Scalar::Decimal(_)))
+        || args.iter().all(|arg| matches!(Scalar::Date(_)))
+        || args.iter().all(|arg| matches!(Scalar::Timestamp(_)))
```
RinChanNOWWW, line 375:
```suggestion
    if args
        .iter()
        .all(|arg| matches!(Scalar::Number(_) | Scalar::Decimal(_) | Scalar::Date(_) | Scalar::Timestamp(_))
```

PR #10699, opened 2023-03-21
fix: do not build pipeline before checking cache.

### LINKED ISSUE #10687: Bug:  After Query Cache is used, SQL with results is not remove to schedule task
**Summary**

gaivjXP69A

> set global enable_query_result_cache=1;

Repeat execute queries in databend-query cluster, local databend-query node reponsed is very fast, other nodes  still looks like the sql is execute once. 

**Expect:**
The other nodes should not be loaded.

RinChanNOWWW:

## Summary

Do not build pipeline before checking cache.

Building pipeline will activate other cluster nodes.

Closes #10687.


wubx:
LGTM
X1sI6ZFMPQ

PR #10710, opened 2023-03-22
refactor(planner): Refine lowering procedure for `ScalarExpr`

leiysky:

## Summary

Introduce a new trait `LoweringContext` to help lower `ScalarExpr` into `Expr`.

With this trait, we can resolve data types of columns in `PhysicalPlanBuilder`. This will make `ScalarExpr` not data type-aware so we don't need to maintain the validity of `ScalarExpr`s in optimizer anymore.


src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -278,8 +281,45 @@ impl PhysicalPlanBuilder {
...
+                                    DataField::new(field.name(), field.data_type().wrap_nullable())
+                                })
+                                .collect::<Vec<_>>(),
+                        );
+                        probe_schema
+                    }
+
+                    _ => probe_side.output_schema()?,
+                };
+
```
leiysky, line 322: @andylokandy  This will solve the problem that it's hard to measure the nullability of columns with outer joins.

xudong963 (approved):

Issue #10689, opened 2023-03-21
feat: Embedding Model Proposal for Databend(By ChatGPT4)


BohuTANG:
**Summary**

## Introduction

An embedding model is a machine learning technique that maps high-dimensional data into a lower-dimensional vector space. It is used to represent complex data such as words, sentences, or documents in a continuous and fixed-length vector. 

This representation can capture semantic and syntactic relationships, enabling a wide range of applications, including natural language processing, recommendation systems, and anomaly detection.

## Obtaining Embedding Vectors with OpenAI API

To obtain embedding vectors using the OpenAI API, you can use OpenAI's pre-trained language models. 

Here's an example using Python:
```
import openai

openai.api_key = "your_openai_api_key"

def get_embedding(text):
    response = openai.Completion.create(
        engine="davinci-codex",
        prompt=f"Embed the following text: {text}",
        max_tokens=16,
        n=1,
        stop=None,
        temperature=0.5,
    )
    embedding = response.choices[0].text.strip()
    return embedding

text = "Databend warehouse"
embedding = get_embedding(text)
print(embedding)

```

## Storing Embedding Vectors in Databend

To store the embedding vectors returned by the OpenAI API in Databend, you can create a table with a column of `Array(Float)` type to hold the vectors. Assuming you have already connected to a Databend instance:
```
CREATE TABLE embeddings (
    id INT,
    text VARCHAR(255) NOT NULL,
    vector ARRAY(FLOAT) NOT NULL
);

```

## Computing the Distance Between Vectors in Databend

Databend can compute the distance between the query vector and stored vectors using a built-in function called vector_distance, which calculates the distance between two ARRAY(FLOAT) inputs. This function can be used directly in SQL queries.

However, when dealing with large-scale datasets and high-dimensional vectors, calculating the vector distance for every pair of vectors can become computationally expensive and slow. To address this challenge, we propose the following techniques:

* `Approximate Nearest Neighbors (ANN)`: Implement an ANN algorithm like Annoy, HNSW, or Faiss to speed up the search for the most similar vectors. ANN algorithms can significantly reduce query time by sacrificing a small amount of accuracy.
* `Dimensionality Reduction`: Use dimensionality reduction techniques like PCA or t-SNE to reduce the size of the vectors before storing them in Databend. Smaller vectors will require less computation for distance calculations.
[...]

PR #10737, opened 2023-03-23
feat: add cosine_distance for vector similarity compute

BohuTANG:

## Summary

This pull request introduces a new `cosine_distance` function, which takes two input vectors, from and to, represented as slices of f32 values. 

The function calculates the cosine distance between the two vectors and returns a Result<f32> type. If the input vectors have unequal lengths, the function returns an Err variant with an InvalidArgument error code, indicating that the lengths are not equal. 

The implementation utilizes the ArrayView type from the `ndarray` library for efficient calculations.

Part #10689


BohuTANG:
@mergify update

BohuTANG:
@mergify update

PR #10760, opened 2023-03-24
fix(query): create a virtual JWT user instead of ensuring on metasrv

everpcpc:

## Summary

Ensuring a user to MetaSrv is a painful cost, which make our JWT auth take up to 1s+ on every request.
Since we do not really need a user on MetaSrv with JWT auth, this PR

* create a virtual JWT user instead of ensuring on metasrv
* remove hostname for JWT auth & deny root user
* ensure_builtin_roles on GlobalService init for UserApiProvider

Closes #issue


Xuanwo (commented):
If this change safe? We don't need the user info anymore?

everpcpc:
> If this change safe? We don't need the user info anymore?

We do not care about if it is a real user in JWT auth, since it is an external auth method, only granted roles are counted.

BohuTANG (approved):

Issue #10745, opened 2023-03-23
tracking: FlightSQL  handler


youngsofun:
**Summary**

- [ ] query https://github.com/datafuselabs/databend/pull/10732
- [ ] update
- [ ] all types
- [ ] more tests
- [ ] try avoid copy
     - [ ] unsafe
     -  [ ] try `arrow_rs grpc + arrow2 flatbuffer` again:  https://github.com/youngsofun/databend/commit/8713bdf168c5a9386c31125e6f34007b77c94dba
- [ ] ci with JDBC client
- [ ] test with BI tools
    -   [ ] dbeaver

PR #10765, opened 2023-03-24
feat: FlightSQL support update (DDL and insert)

youngsofun:

## Summary

part of https://github.com/datafuselabs/databend/issues/10745.

-  ddl/insert work well with JDBC, see https://github.com/youngsofun/test_flight_sql/blob/main/src/main/java/Test.java.
    - but the official rs client has a bug (which is not used in production AFAIK),  so new unit test cases is commented out for now. however,  it uses a different interface of FlightSQL from JDBC(in JDBC, all statement is implemented as preparedStatement internally ).
- port change from 7000 to 8900 due to https://github.com/datafuselabs/databend/discussions/10741

- still only basic types are supported, I will go back to it after  impl meta related API  https://github.com/datafuselabs/databend/issues/10745.

Closes #issue

### DIFF
--- a/scripts/ci/deploy/config/databend-query-node-1.toml
+++ b/scripts/ci/deploy/config/databend-query-node-1.toml
@@ -29,7 +29,7 @@ http_handler_port = 8000
 
 # Databend Query FlightSQL Handler.
 flight_sql_handler_host = "0.0.0.0"
-flight_sql_handler_port = 50050
+flight_sql_handler_port = 8900
 
 tenant_id = "test_tenant"
 cluster_id = "test_cluster"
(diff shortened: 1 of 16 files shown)


ariesdevil (approved):

youngsofun:
Databend Query start failure, cause: Code: 1068, displayText = {127.0.0.1:4100} Address already in use (os error 98).
😭

youngsofun:
both 8700 & 8600 failed in test_stateful_standalone_linux but suceed with test_stateful_cluster_linux

BohuTANG:
Conflicting files
src/query/service/src/servers/mysql/mysql_interactive_worker.rs

everpcpc:
> both 8700 & 8600 failed in test_stateful_standalone_linux but suceed with test_stateful_cluster_linux

There are two query instance in statefultest, another is a sharing instance: `scripts/ci/deploy/config/databend-query-node-shared.toml`, which would conflict if default port not changed in it.

youngsofun:
> Conflicting files
> src/query/service/src/servers/mysql/mysql_interactive_worker.rs

rebased

TCeason (approved):

youngsofun:
flightsql bug fixed

https://github.com/apache/arrow-rs/pull/3938

PR #10778, opened 2023-03-26
feat(planner): Try add optimizer test framework

dusx1981:

## Summary

As subsequent optimization rules become more and more complex, testing becomes difficult without a single-point debugging framework.
From my current understanding, there is no framework for unit testing the optimizer. I'm trying to add this framework.
Three places were chosen as follow-up paths:

> - **src/query/sql/tests:** The optimization process requires QueryContext data, and the creation of QueryContext depends on the relevant logic of Session. If you add test logic here, you need to rely on the Service module, which will cause circular dependency
> - **src/query/service/tests/it/sql/planner/optimizer:** The example code is copied from src/query/service/tests/it/storages/fuse/operations/optimize.rs, not the real test logic. But because I'm a newcomer to Rust, I don't know why I can't enter debugging after setting a breakpoint.
> - **tests/sqlunittests:** The path corresponding to tests/sqllogictests, the path for unit testing, in which a subpath for unit testing of the optimizer is created: tests/sqlunittests/optimizer. From the current point of view, there is no problem with this path.

Closes #issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3320,6 +3320,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "databend-sqlunittests"
+version = "0.1.0"
+dependencies = [
+ "common-config",
+]
+
+[[package]]
 name = "databend-thrift"
 version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


dusx1981:
@Xuanwo 

Because I just participated in this project, I want to confirm whether it is really necessary to add a test framework for the optimizer from an engineering point of view before actually starting to do it.

Xuanwo:
Thank you for your effort, but we do not require it at this time. And we don't want a test framework outsides `sqllogictests`.

PR #10786, opened 2023-03-27
fix(query): auth jwt with ensure_user on metasrv

everpcpc:

## Summary

Summary about this PR

Closes #issue


src/query/service/src/auth.rs
```diff
@@ -78,34 +80,55 @@ impl AuthMgr {
...
-                        for role in roles.clone().into_iter() {
-                            user_info.grants.grant_role(role);
+                let tenant = session.get_current_tenant();
+                let identity = UserIdentity::new(&user_name, "%");
+                let user = match user_api.get_user(&tenant, identity.clone()).await {
+                    Ok(user_info) => match user_info.auth_info {
+                        AuthInfo::JWT => Ok(user_info),
+                        _ => return Err(ErrorCode::AuthenticateFailure("wrong auth type")),
+                    },
+                    Err(e) => {
```
flaneur2020, line 90: how about short-circuit the logic in the front ?
if e.code() != ErrorCode::UNKNOWN_USER {
  return Err(ErrorCode::AuthenticateFailure(e.message()));
}

src/query/service/src/auth.rs
```diff
@@ -78,34 +80,55 @@ impl AuthMgr {
...
+                        let mut user_info = UserInfo::new(&user_name, "%", AuthInfo::JWT);
+                        if user_info.identity().is_root() {
+                            return Err(ErrorCode::AuthenticateFailure(
+                                "root user is not allowed in jwt auth.",
+                            ));
+                        }
+                        if let Some(ref roles) = ensure_user.roles {
+                            for role in roles.clone().into_iter() {
+                                user_info.grants.grant_role(role);
+                            }
```
flaneur2020, line 111: in the earilier code, there's also a function call to initialize the builtin-roles here (including ACCOUNT_ADMIN and PUBLIC).

src/query/service/src/auth.rs
```diff
@@ -78,34 +80,55 @@ impl AuthMgr {
...
+                            None => return Err(ErrorCode::AuthenticateFailure(e.message())),
+                        };
+                        // create a new user if not exists
+                        let mut user_info = UserInfo::new(&user_name, "%", AuthInfo::JWT);
+                        if user_info.identity().is_root() {
+                            return Err(ErrorCode::AuthenticateFailure(
+                                "root user is not allowed in jwt auth.",
+                            ));
+                        }
```
flaneur2020, line 107: I guess this check about `is_root()` do not needed?

with restricting the users with a JWT token must marked as JWT auth type, the root user's auth_type are less likely to be JWT, thus it also get filtered out.

src/query/service/src/auth.rs
everpcpc: It was moved to the init phase.

flaneur2020 (approved):
LGTM :heart:

src/query/service/src/auth.rs
```diff
@@ -78,34 +80,47 @@ impl AuthMgr {
...
+                        user_api.add_user(&tenant, user_info, true).await?;
+                        user_api.get_user(&tenant, identity).await
                     }
-                }
+                }?;
+
+                let user = match &user.auth_info {
+                    AuthInfo::JWT => user,
+                    _ => return Err(ErrorCode::AuthenticateFailure("wrong auth type")),
+                };
```
flaneur2020, line 113: I guess this check is duplicated with L86: https://github.com/datafuselabs/databend/pull/10786/files#diff-fe9c4d4d86fb6815af28d24728a7e98cc84952c53d2edd270a76bdd7279cd333R86

src/query/service/src/auth.rs
```diff
@@ -78,34 +80,47 @@ impl AuthMgr {
...
+                            .ok_or(ErrorCode::AuthenticateFailure(e.message()))?;
+                        // create a new user if not exists
+                        let mut user_info = UserInfo::new(&user_name, "%", AuthInfo::JWT);
+                        if let Some(ref roles) = ensure_user.roles {
+                            for role in roles.clone().into_iter() {
+                                user_info.grants.grant_role(role);
+                            }
                         }
+                        user_api.add_user(&tenant, user_info, true).await?;
+                        user_api.get_user(&tenant, identity).await
```
flaneur2020, line 106: just returns `user_info` is ok? this may reduce an rpc call

BohuTANG: Mark resolved?

PR #10812, opened 2023-03-28
fix(query): fix range/generate_series schema

sundy-li:

## Summary

Summary about this PR

Closes #issue


sundy-li:
cc @jun0315

BohuTANG (approved):

l1t1:
https://github.com/Mytherin says duckdb's range/generate_series are not parallelized because they are mainly used as debug/data generation functions, rather than for serious usage.

sundy-li:
> [@Mytherin](https://github.com/Mytherin) says duckdb's range/generate_series are not parallelized because they are mainly used as debug/data generation functions, rather than for serious usage.

It's also not parallelized in databend,  it's only used for debugging. Sorry for the inappropriate comparison.

Issue #10803, opened 2023-03-28
Feature: copy with transform support max_files too


youngsofun:
**Summary**

update bind_copy_from_query_into_table

PR #10813, opened 2023-03-28
fix(Copy): max_files was set to 1000 when not provided.

youngsofun:

## Summary

Summary about this PR

Closes #issue


sundy-li (approved):

TCeason (approved):

youngsofun:
just rebased main

TCeason:
LGTM. By the way: if support this https://github.com/datafuselabs/databend/issues/10803, the enum StageFileStatus maybe can be drop?

TCeason (approved):

youngsofun:
> StageFileStatus

make sense to me.

src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -253,45 +250,42 @@ impl CopyInterpreter {
             Some(max_files)
         };
 
-        let mut all_source_file_infos = if force {
+        let all_source_file_infos = if force {
```
BohuTANG, line 253: Don't understand here why we need check the force option.
As my understanding, we should get the max_files files from the list is ok no matter whether the force is true or not.

src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -253,45 +250,42 @@ impl CopyInterpreter {
...
             StageTable::list_files(&stage_table_info, max_files).await?
         } else {
             StageTable::list_files(&stage_table_info, None).await?
         };
+        let num_all_files = all_source_file_infos.len();
 
-        info!("end to list files: {}", all_source_file_infos.len());
+        info!("end to list files: got {} files", num_all_files);
```
BohuTANG, line 260: It would be better to log the max_files.

src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -253,45 +250,42 @@ impl CopyInterpreter {
...
 
-        if !force {
+        let need_copy_file_infos = if force {
```
BohuTANG, line 262: Same as above, if user use max_files option, we assume he absolutely know what he did, it's not related with the force option.

src/query/service/src/interpreters/interpreter_copy.rs
youngsofun: if force = false, files is listed and then filtered by copied files

src/query/service/src/interpreters/interpreter_copy.rs
youngsofun: done

src/query/service/src/interpreters/interpreter_copy.rs
BohuTANG: In the case he only wants force copy max_files files for test or others purposes.
Or, he shouldn't use the max_files.

src/query/service/src/interpreters/interpreter_copy.rs
BohuTANG: I see, thanks.

BohuTANG:
Waut for the CI pass, ready to merge.

youngsofun:
fixing the test

PR #10814, opened 2023-03-28
feat(expression): support Vector data type

### LINKED ISSUE #10769: Introduce a new data type `VECTOR` (alias of `ARRAY<Float32>`) (@b41sh )

b41sh:

## Summary

Support the `Vector` data type, as alias of `Array<Float32>`

Closes #10769

### DIFF
--- a/src/meta/proto-conv/src/schema_from_to_protobuf_impl.rs
+++ b/src/meta/proto-conv/src/schema_from_to_protobuf_impl.rs
@@ -210,6 +210,7 @@ impl FromToProto for ex::TableDataType {
                         ex::TableDataType::Decimal(ex::types::decimal::DecimalDataType::from_pb(x)?)
                     }
                     Dt24::EmptyMapT(_) => ex::TableDataType::EmptyMap,
+                    Dt24::VectorT(_) => ex::TableDataType::Vector,
                 };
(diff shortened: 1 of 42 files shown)


sundy-li:
Why not do it in AST Level ?

BohuTANG (approved):

sundy-li:
@BohuTANG 
Subjectively adding some types by mind may make our code difficult to maintain, because it's hard to change and be compatible with the previous type version.

1. What if we want to change this to `Array<Float64>` in the future ?
2. Vector type is only used for `cosine distance calculations`, we may need a detailed RFC to determine whether and how to make DataBend a vector database.

sundy-li:
> If we can create a index like 'create table XX(v Array(float32) index ivf_pq)' is ok, here only to do the special index.

I do prefer this.  A dynamic index is hard to maintain, considering the old version of the bloom filter index.

BohuTANG:
@b41sh Convert this PR to draft, it's low-prority and need some proposal ok n the future, thanks.

drmingdrmer (approved):

b41sh:
We don't need the `Vector` data type for now, so close this PR. We can reopen it later if needed.

PR #10817, opened 2023-03-28
fix: incorrect index size and perfect_block_count during merge segments

zhyass:

## Summary

incorrect index size and count of perfect blocks during merge segments.
note: block data and index data not affected, only the statistics is incorrect.

Closes #issue


dantengsky:
@BohuTANG please have a look & thanks.

the situation before the PR is :

after appending data into data, 
- at the table level, the index size is incorrect
- but at the segment level, the index size is correct
- and after `optimize table t compact segment`
  the table-level index size will be re-calculated from the segment level, thus the table-level index size will be corrected.
  
  
the situation after the PR is :

 - table level index size will be  correct, after appending data to  table (without re-calculate by using compact segment)

Issue #10810, opened 2023-03-28
Tracking: window function


ariesdevil:
**Summary**

## Main targets

- [x] initial impl window function
	- [x] https://github.com/datafuselabs/databend/pull/10700
    - [x] https://github.com/datafuselabs/databend/pull/10772
    - [x] https://github.com/datafuselabs/databend/pull/10791
- [x] add more tests
- [x] support range frame
	- [x] initial support range frame #10911
	- [x] range frame supports all number types #11002 
- [x] support more complicated window syntax
	- [x] named window #10897 
	- [x] inherited named window #10934  
	- [x] support window function in order by clause #10989 
- [x] add docs #11013
- [x] General window functions. #11148

## TPC-DS Related.

- [x] Support aggregation function in window function (includes `grouping`). #10887

youngsofun:
I`d like to try percent_rank/cume_dist/ntile If no one else is working on them.

ArberSephirotheca:
Hi, can I work the first_value window func ?

ariesdevil:
> Hi, can I work the first_value window func ?↳

Sure and welcome :）

PR #10824, opened 2023-03-29
feat: support general window function row_number/rank/dense_rank.

RinChanNOWWW:

## Summary

- Add basic window function test cases.
- Fix bugs when order by window funciton result.
- Add some missing codes when DFSing scalar tree (in window node).
- Modify plannr to support general window functions (non-aggregation functions).
- Support row_number/rank/dense_rank.
- Fix several bugs.

Tracking in #10810

### DIFF
--- a/src/query/formats/src/file_format_type.rs
+++ b/src/query/formats/src/file_format_type.rs
@@ -137,7 +137,6 @@ impl FileFormatOptionsExt {
 
     pub fn get_output_format(&mut self, schema: TableSchemaRef) -> Result<Box<dyn OutputFormat>> {
         self.check()?;
-        // println!("format {:?} {:?} {:?}", fmt, options, format_settings);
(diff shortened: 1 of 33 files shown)


ariesdevil (approved):
Look great to me

src/query/formats/src/file_format_type.rs
```diff
@@ -137,7 +137,6 @@ impl FileFormatOptionsExt {
 
     pub fn get_output_format(&mut self, schema: TableSchemaRef) -> Result<Box<dyn OutputFormat>> {
         self.check()?;
-        // println!("format {:?} {:?} {:?}", fmt, options, format_settings);
```
ariesdevil, line 140: this line should be removed

RinChanNOWWW: Already removed.

sundy-li (approved):

PR #10831, opened 2023-03-29
fix:  processors used in the impl of `replace into` may lose data

### LINKED ISSUE #10861: bug: `replace into` may lost data being ingested
**Summary**

on behalf of @wubx 

`replace into` may lose data being upserted into:

- lost data that should be inserted into
- lost data that should be updated

as shown in the following log (default settings):

~~~
mysql> create table s(id int);

mysql> --- the source table `t` consist of multiple blocks

mysql> insert into s values(1);
Query OK, 1 row affected (0.11 sec)

mysql> insert into s values(2);
Query OK, 1 row affected (0.12 sec)

mysql> insert into s values(3);
Query OK, 1 row affected (0.51 sec)



mysql> create table t(id int);



mysql> replace into t on(id) select * from s;
Query OK, 3 rows affected (0.16 sec)

mysql> -- the following result is NOT as expected, should be {1,2,3}
mysql> select * from t;
+------+
| id   |
+------+
|    3 |
|    1 |
+------+
2 rows in set (0.04 sec)
Read 2 rows, 8.00 B in 0.005 sec., 367.41 rows/sec., 1.44 KiB/sec.

mysql> replace into t on(id) select * from s;
Query OK, 3 rows affected (0.27 sec)

mysql> -- the following result is NOT as expected, should be {1,2,3}
mysql> select * from t;
+------+
| id   |
+------+
|    3 |
|    3 |
|    1 |
|    2 |
+------+
4 rows in set (0.04 sec)
Read 4 rows, 16.00 B in 0.006 sec., 711.68 rows/sec., 2.78 KiB/sec.
~~~

dantengsky:

## Summary

_broadcast processor_ and _replace_into processor_ used in the impl of `replace into` may lose data blocks,

and if that happened, data may be partially replaced into the target table.

- [x] fix
- [x] sqllogic test

Closes #10861

### DIFF
--- a/src/query/storages/fuse/src/operations/merge_into/mutator/merge_into_mutator.rs
+++ b/src/query/storages/fuse/src/operations/merge_into/mutator/merge_into_mutator.rs
@@ -32,7 +32,7 @@ use storages_common_cache::LoadParams;
 use storages_common_table_meta::meta::BlockMeta;
 use storages_common_table_meta::meta::ColumnStatistics;
 use storages_common_table_meta::meta::Location;
-use tracing::debug;
+use tracing::info;
 
 use crate::io::write_data;
 use crate::io::BlockBuilder;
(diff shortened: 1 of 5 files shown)


zhyass (approved):
LGTM

PR #10835, opened 2023-03-30
feat: add feature "raft-store-defensive" to enable defensive check

drmingdrmer:

##### feat: add feature "raft-store-defensive" to enable defensive check

This feature enables defensive check when accessing the raft store, e.g,
it check if the next log is at index of last index plus one, when
appending logs.

For optimal performance, this feature is disabled by default.

## Changelog

- New Feature

## Related Issues

### DIFF
--- a/src/binaries/Cargo.toml
+++ b/src/binaries/Cargo.toml
@@ -26,6 +26,9 @@ io-uring = [
     "common-meta-raft-store/io-uring",
 ]
 
+# Enable defensive check when accessing raft store.
+raft-store-defensive = ["databend-meta/raft-store-defensive"]
+
 enable-histogram-metrics = [
     "default",
     "common-metrics/enable-histogram",
--- a/src/binaries/metabench/main.rs
+++ b/src/binaries/metabench/main.rs
@@ -41,6 +41,10 @@ use serde::Serialize;
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Parser)]
 #[clap(about, version = &**METASRV_COMMIT_VERSION, author)]
 struct Config {
+    /// The prefix of keys to write.
+    #[clap(long, default_value = "0")]
+    pub prefix: u64,
+
     #[clap(long, default_value = "10")]
     pub client: u32,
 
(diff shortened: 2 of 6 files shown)


Xuanwo:
> This feature enables defensive check when accessing the raft store, e.g,
> it check if the next log is at index of last index plus one, when
> appending logs.

Do you think it's a good idea to hide them under `#[cfg(debug_assertions)]` instead of features?

drmingdrmer:
> > This feature enables defensive check when accessing the raft store, e.g,
> > it check if the next log is at index of last index plus one, when
> > appending logs.
> 
> Do you think it's a good idea to hide them under `#[cfg(debug_assertions)]` instead of features?

Sounds a good idea! I am not quite sure if it is possible to enable debug-assertions for just one crate `databend-meta`? AFAIK, enabling debug-assertions is a workspace level config. Having to enable debug-assertions for all the crates in our project feels a little bit inconvenient.

Xuanwo:
> AFAIK, enabling debug-assertions is a workspace level config. Having to enable debug-assertions for all the crates in our project feels a little bit inconvenient.

Oh, `#[cfg(debug_assertions)]` can be added on block level. Take opendal's code for example:

```rust
pub fn to_flat_pager<A: Accessor, P>(acc: A, path: &str, size: usize) -> ToFlatPager<A, P> {
    #[cfg(debug_assertions)]
    {
        let meta = acc.info();
        debug_assert!(
            !meta.capabilities().contains(AccessorCapability::Scan),
            "service already supports scan, call to_flat_pager must be a mistake"
        );
        debug_assert!(
            meta.capabilities().contains(AccessorCapability::List),
            "service doesn't support list hierarchy, it must be a bug"
        );
    }

    <real work>
}
```

Xuanwo:
> Having to enable debug-assertions for all the crates in our project feels a little bit inconvenient.

Ok, I got it!

Xuanwo (approved):

PsiACE (approved):

Xuanwo:
@mergify update

PR #10843, opened 2023-03-30
refactor: add share endpoint manager to query share endpoint

lichuang:

## Summary

refactor: add share endpoint manager to query share endpoint

Closes #issue


src/query/sharing/src/share_endpoint.rs
```diff
@@ -13,50 +13,170 @@
 // limitations under the License.
 
 use std::collections::BTreeMap;
+use std::sync::Arc;
+use std::sync::RwLock;
```
BohuTANG, line 17: Use `use parking_lot::RwLock` instead? This we can not define a new error code:`RwLockError(1112),`

BohuTANG (approved):

Issue #10829, opened 2023-03-29
bug: query can't sync the meta endpoints after metasrv restart


BohuTANG:
**Summary**
![img_v2_08c774ea-7153-4048-a89b-d01379b2d46g](https://user-images.githubusercontent.com/172204/228584782-9c632ff7-9b4a-4daa-9d51-7283414acfbd.jpg)

drmingdrmer:
Redirect from: https://github.com/datafuselabs/databend/pull/10848#issuecomment-1490082349

> I assume the error is raised here:
> https://github.com/datafuselabs/databend/blob/7f408b4ab45b0f6af49c5aa64186bde26cfc50aa/src/meta/client/src/grpc_client.rs#L560-L562
>
> And the root cause is that `unhealthy_endpoints` did not evict unreachable nodes very quickly, the default eviction timeout is 120 seconds(nodes that are in `unhealthy_endpoints` will not be used):
> https://github.com/datafuselabs/databend/blob/7f408b4ab45b0f6af49c5aa64186bde26cfc50aa/src/meta/client/src/grpc_client.rs#L342
>
> In order to address this issue, we should introduce a new configuration item for the timeout of `unhealthy_endpoints` and reduce it to an appropriate value.

PR #10848, opened 2023-03-30
fix: when saved endpoints is empty, use endpoint config every try_config_endpoints_interval

lichuang:

## Summary

fix: when saved endpoints is empty, use endpoint config every try_config_endpoints_interval

Closes https://github.com/datafuselabs/databend/issues/10829


src/query/config/src/config.rs
```diff
@@ -1647,6 +1647,11 @@ pub struct MetaConfig {
     #[serde(alias = "auto_sync_interval")]
     pub auto_sync_interval: u64,
 
+    /// try_config_endpoints_interval is the interval to try config endpoint when meta node endpoints is empty.
+    #[clap(long = "try-config-endpoints-interval", default_value = "10")]
+    #[serde(alias = "try_config_endpoints_interval")]
+    pub try_config_endpoints_interval: u64,
```
BohuTANG, line 1653: Name to `refresh_meta_endpoints_interval` is better?

lichuang: we now have a `auto_sync_interval` config, which used to sync meta node endpoint internal, `refresh` means it will change endpoints, but this config is used to set internal use config endpoints if meta node is down.

drmingdrmer:
I assume the error is raised here:
https://github.com/datafuselabs/databend/blob/7f408b4ab45b0f6af49c5aa64186bde26cfc50aa/src/meta/client/src/grpc_client.rs#L560-L562

And the root cause is that `unhealthy_endpoints` did not evict unreachable nodes very quickly, the default eviction timeout is 120 seconds(nodes that are in `unhealthy_endpoints` will not be used):
https://github.com/datafuselabs/databend/blob/7f408b4ab45b0f6af49c5aa64186bde26cfc50aa/src/meta/client/src/grpc_client.rs#L342


In order to address this issue, we should introduce a new configuration item for the timeout of `unhealthy_endpoints` and reduce it to an appropriate value.

ariesdevil (approved):

Xuanwo:
@mergify update

PR #10855, opened 2023-03-30
refactor: do not upsert copied-files set if both purge and force are enabled

### LINKED ISSUE #10854: Feature: adjust behavior of  force mode `copy into`
**Summary**

during execution of  `copy into`, if both `purge` and `force` are enabled,  do not upsert the table's copied-files set in meta server

dantengsky:

## Summary

during the execution of `copy-into`, if both purge and force are enabled, do not upsert the table's copied-files set in meta server.

NOTE:

since the table's copied-files set will NOT be upsert in meta service, if both `purge` and `force` are enabled.

later/parallel executions of `copy-into`  are taking the risk of data duplication.

Closes #10854

### DIFF
--- a/src/query/service/src/catalogs/default/mutable_catalog.rs
+++ b/src/query/service/src/catalogs/default/mutable_catalog.rs
@@ -287,7 +287,14 @@ impl Catalog for MutableCatalog {
         req: UpdateTableMetaReq,
     ) -> Result<UpdateTableMetaReply> {
         match table_info.db_type.clone() {
-            DatabaseType::NormalDB => Ok(self.ctx.meta.update_table_meta(req).await?),
+            DatabaseType::NormalDB => {
+                info!(
+                    "updating table meta. table desc: [{}], has copied files: [{}]",
+                    table_info.desc,
+                    req.copied_files.is_some()
+                );
+                Ok(self.ctx.meta.update_table_meta(req).await?)
+            }
             DatabaseType::ShareDB(share_ident) => {
                 let db = self
                     .get_database(&share_ident.tenant, &share_ident.share_name)
--- a/src/query/service/src/catalogs/default/mutable_catalog.rs
+++ b/src/query/service/src/catalogs/default/mutable_catalog.rs
@@ -289,7 +289,7 @@ impl Catalog for MutableCatalog {
         match table_info.db_type.clone() {
             DatabaseType::NormalDB => {
                 info!(
-                    "updating table meta. table desc: [{}], has copied files: [{}]",
+                    "updating table meta. table desc: [{}], has copied files: [{}]?",
                     table_info.desc,
                     req.copied_files.is_some()
                 );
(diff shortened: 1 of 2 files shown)


src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -429,13 +433,23 @@ impl CopyInterpreter {
...
-                expire_hours,
-                copied_file_tree,
-                fail_if_duplicated,
-            );
+            let table_info = to_table.get_table_info();
+            let upsert_copied_files_request = {
+                if stage_info.copy_options.purge && force {
+                    // if `purge-after-copy` is enabled, and in `force` copy mode,
+                    // we do not need to upsert copied files into meta server
+                    info!("[purge] and [force] are both enabled,  will not update copied-files set. ({})", &table_info.desc);
```
BohuTANG, line 441: `table_info.desc` is?

src/query/service/src/catalogs/default/mutable_catalog.rs
```diff
@@ -287,7 +287,14 @@ impl Catalog for MutableCatalog {
         req: UpdateTableMetaReq,
     ) -> Result<UpdateTableMetaReply> {
         match table_info.db_type.clone() {
-            DatabaseType::NormalDB => Ok(self.ctx.meta.update_table_meta(req).await?),
+            DatabaseType::NormalDB => {
+                info!(
+                    "updating table meta. table desc: [{}], has copied files: [{}]",
+                    table_info.desc,
+                    req.copied_files.is_some()
```
youngsofun, line 294: `has copied files: [{true}]` ?

dantengsky:
> seems result of https://github.com/datafuselabs/databend/blob/main/tests/suites/1_stateful/00_copy/00_0004_copy_with_max_files.sh should be updated

thanks for reminding

src/query/service/src/catalogs/default/mutable_catalog.rs
dantengsky: o, i see. that' is better!

src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -429,13 +433,23 @@ impl CopyInterpreter {
...
 
-            let fail_if_duplicated = !force;
-            let upsert_copied_files_request = Self::upsert_copied_files_request(
-                table_id,
-                expire_hours,
-                copied_file_tree,
-                fail_if_duplicated,
-            );
+            let table_info = to_table.get_table_info();
+            let upsert_copied_files_request = {
```
youngsofun, line 437: unnecessary {} out of if else?

youngsofun:
> > seems result of https://github.com/datafuselabs/databend/blob/main/tests/suites/1_stateful/00_copy/00_0004_copy_with_max_files.sh should be updated
> 
> thanks for reminding

rest lgtm as long as the results  of this tests is expected

src/query/service/src/interpreters/interpreter_copy.rs
dantengsky: clippy seems to accept them :)

besides:

 otherwise the line is too long to me 

2023-03-30-211541_983x86_scrot

youngsofun (approved):

Issue #10850, opened 2023-03-30
Add meta-service startup timeout config


drmingdrmer:
Add a configuration to control the max duration that a meta-service can spend attempting to join a cluster before exiting with an error indicating that it was unable to join the cluster.

- In some cases, a meta server is expected to exit if it can not form a cluster before some timeout, so that administrator will find out a node is down and intervenes.

- In other cases, a meta server is expected to keep trying to form a cluster for a very long time.

A node that just started up waits for a leader to send some message to it then it knows that there is a established leader and a healthy cluster.
https://github.com/datafuselabs/databend/blob/7a443445967f41d02d0f0e90566cd7da9f4e1643/src/binaries/meta/main.rs#L180-L190

PR #10877, opened 2023-03-31
feat: add meta join cluster timeout config

lichuang:

## Summary

feat: add meta join cluster timeout config

Closes https://github.com/datafuselabs/databend/issues/10850


src/meta/service/src/configs/outer_v0.rs
```diff
@@ -483,6 +486,10 @@ pub struct RaftConfig {
     /// if not, the default name is used
     #[clap(long, default_value = "foo_cluster")]
     pub cluster_name: String,
+
+    /// Max timeout(in milli seconds) when waiting a cluster leader.
+    #[clap(long, default_value = "10000")]
+    pub wait_leader_timeout: u64,
```
drmingdrmer, line 492: Previously this timeout is 70 seconds(election timeout is heartbeat_interval*7, while this timeout is election_timeout*10). Let's keep the default value unchanged.

PR #10888, opened 2023-03-31
feat: Allow setting connect timeout and timeout for s3

Xuanwo:

## Summary

feat: Allow setting connect timeout and timeout for s3

This PR adds a feature which allow users to set connect timeout and timeout for s3. This feature is internal used only. And we will make it public until we make sure that it works.

- `_DATABEND_INTERNAL_CONNECT_TIMEOUT`: configure the connect timeout, default to 30s
- `_DATABEND_INTERNAL_TIMEOUT`: configure the whole request's timeout, default to 120s

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2253,7 +2253,9 @@ dependencies = [
  "futures",
  "opendal",
  "regex",
+ "reqwest",
  "serde",
+ "ureq",
 ]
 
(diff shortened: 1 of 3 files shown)


BohuTANG:
Some tests always failed :\

BohuTANG:
@mergify update

PR #10916, opened 2023-04-04
feat(debug): Add timeout while reading from s3

Xuanwo:

## Summary

feat(debug): Add timeout while reading from s3

This PR adds a new internal debugging value: `_DATABEND_INTERNAL_READ_TIMEOUT`.

We will make a read timeout if it's hang too long. default to 30s.

NOTE: opendal Accessor's `read` operation is used to sending request to get a reader. Read IO is not happened here.


dantengsky:
LGTM.

note that the timeout setting introduced by this PR, is applied to the reader's construction, not the data reading phase.

dantengsky (approved):

Issue #10376, opened 2023-03-06
Refactor:  change FileFormatOptions to enum internally [as early as possible!].


youngsofun:
**Summary**

## background

currently,  we kind of store the AST  `file_format = (type=<type> .., <key>=<value>..)`.  to meta directly

```
pub struct FileFormatOptions {
    pub format: StageFileFormatType,
    // Number of lines at the start of the file to skip.
    pub skip_header: u64,
    pub field_delimiter: String,
    pub record_delimiter: String,
    pub nan_display: String,
    pub escape: String,
    pub compression: StageFileCompression,
    pub row_tag: String,
    pub quote: String,
    pub name: Option<String>,
}
```
note ta
to make the behavior entirely determined, we fill the options used by `<type>` to a default value of `<type>` (diff type may have diff value for the same option) before use/store it.

for the options not used by the `<type>` we set a `trivial default value`

## problem

but that is not safe enough.
for example,  option `escape`  is not supported yet by `json` not,  when we may  add it later,  the  `trivial default value`  may cause problem.

```
struct CSVFileFormat{
      pub skip_header: u64,
    pub field_delimiter: String,
  ....
}

pub struct FileFormatConfig {
     CSV{CSVFileFormatConfig}.
   .... 
}
```

this is also much much more clear (about which options an format support)

## change it now

`create file format` is newly added, and not widely used yet,  FileFormatOptions is only stored as part of `Stage`
so it is easier to upgrade the meta(omit  old file_format object created,   upgrade/tolerant  FileFormatOptions in Stage object)

cc @sundy-li @Xuanwo @drmingdrmer @ariesdevil

sundy-li:
> omit old file_format object created, upgrade

I think it's ok!

Xuanwo:
> `create file format` is newly added, and not widely used yet, FileFormatOptions is only stored as part of `Stage`
> so it is easier to upgrade the meta(omit old file_format object created, upgrade(on get?)/tolerant FileFormatOptions field in Stage object)

LGTM

ariesdevil:
LGTM

youngsofun:
> upgrade(on get?)/tolerant FileFormatOptions field in Stage object

@drmingdrmer 

How do you recommend doing this update in meta?

```
message StageInfo {
  FileFormatOptions file_format_options = 4;
}
```
to 

```
message StageInfo {
  FileFormatInfo file_format_info = 9;
}
```
```

drmingdrmer:
[...]

PR #10919, opened 2023-04-04
refactor(meta): extract file_format.proto and stage.proto.

youngsofun:

## Summary

I am working on https://github.com/datafuselabs/databend/issues/10376.

it will make codes much more clear to do this refactoring before adding new structures (about file formats) .
and it is easier to review in a separate pr.

nothing except `paths of PB messages in code level` should be changed.

format-related Messages were defined inside StageInfo,  they are refactored to the top level.

not PB tags affected,  so should not cause any compatibility problem.


src/meta/protos/proto/file_format.proto
```diff
@@ -0,0 +1,77 @@
...
+  Gzip = 1;
+  Bz2 = 2;
+  Brotli = 3;
+  Zstd = 4;
+  Deflate = 5;
+  RawDeflate = 6;
+  Lzo = 7;
+  Snappy = 8;
+  // We used to assign `None` as 9, it's a mistake.
+  // Please change this value to 0 instead in next version bump.
```
ariesdevil, line 42: Should this be modified in this PR?

ariesdevil (approved):
Codes LGTM, it is recommended to describe in detail the compatibility involved.

youngsofun:
> Codes LGTM, it is recommended to describe in detail the compatibility involved.

not PB tags affected, so should not cause any compatibility problem.

drmingdrmer (approved):
Looks like there are just file moves. LGTM!

PR #10931, opened 2023-04-04
refactor(format): add FileFormatParams.

youngsofun:

## Summary

step 2  of  https://github.com/datafuselabs/databend/issues/10376

1. add enum FileFormatParams which is going to replace FileFormatOptions,  it`s corresponding PB def in meta, and the convs
2. func to trans FileFormatAst to FileFormatParams.
3. rename `skip_header` to `headers`, since it will be also used for OutputFormat later.

FileFormatParams is not really used, so this pr is safe.

There may be changes to  FileFormatParams in the following pr, but they won't be significant.

### DIFF
--- a/src/meta/app/src/principal/file_format.rs
+++ b/src/meta/app/src/principal/file_format.rs
@@ -1,0 +1,203 @@
+// Copyright 2023 Datafuse Labs.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


drmingdrmer:
If prototypes changes, a new version must be added, see:
https://github.com/datafuselabs/databend/blob/20d12bc1d109a2c3920a5c68ff34a957855dad76/src/meta/proto-conv/src/util.rs#L97-L102

youngsofun:
new proto def add but not use in this pr，i.e. will not be writen to meta.

I guess it is ok to bump version in next pr？

drmingdrmer:
> new proto def add but not use in this pr，i.e. will not be writen to meta.
> 
> I guess it is ok to bump version in next pr？

It is recommended that the unit test and the corresponding code be included in the same commit/PR.
In this PR it should tests these new types can be serialized/deserialized correctly.
The next PR should tests the containing types using these new types can be serialized/deserialized correctly.

drmingdrmer:
> new proto def add but not use in this pr，i.e. will not be writen to meta.
> 
> I guess it is ok to bump version in next pr？

If proto type files change, it is a new version and the version should be increased.

youngsofun:
@drmingdrmer  please review again

PR #10821, opened 2023-03-29
fix: EmptyShareEndpointConfig  error on any SHOW TABLES

flaneur2020:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Currently SHOW TABLES is a rewrite to selecting `system.tables`.

However in `system.tables`, it tries to list all the tables from every databases. If any databases raises EmptyShareEndpointConfig, all SHOW TABLES to any database will met this error.

This PR tries to swallow the EmptyShareEndpointConfig on SHOW TABLES statement.

PR #10978, opened 2023-04-07
fix: swallow all the errors about sharing database in system.tables

flaneur2020:

## Summary

this pr could be considered as a follow up of #10821. 

in #10821, we swallowed the error EMPTY_SHARE_ENDPOINT_CONFIG, but there's still other errors like UNKNOWN_SHARE_ENDPOINT_CONFIG on sharing database may break all the SHOW TABLES queries, and SHOW TABLES is an important query for client side to reflect the table fields, if it fails, the queries will have less chance to be sent to the backend instances.

Currently SHOW TABLES is a rewrite to `SELECT FROM system.tables`, and system.tables tries to list all the tables at first. If list table on any database fails, the SHOW TABLES statement on ALL databases will fail.

The system.tables is not considered easy to be broken with bad user input, as it'd mostly to just read the metasrv and list the tables. But when it comes with sharing db, the risks on errors may increase. This PR tries to swallow all the errors when it comes with sharing databases.


src/query/storages/system/src/tables_table.rs
```diff
@@ -113,11 +113,20 @@ where TablesTable<T>: HistoryAware
...
                 let tables = match Self::list_tables(&ctl, tenant.as_str(), name).await {
                     Ok(tables) => tables,
-                    Err(err) if err.code() == ErrorCode::EMPTY_SHARE_ENDPOINT_CONFIG => {
-                        tracing::warn!("list tables failed on db {}: {}", db.name(), err);
-                        continue;
+                    Err(err) => {
+                        // Swallow the errors related with sharing. Listing tables in a shared database
+                        // is easy to get errors with invalid configs, but system.tables should not be
+                        // affected by it.
+                        if db.get_db_info().meta.from_share.is_some() {
```
flaneur2020, line 120: I dunno is this a feasible approach to check the database is created from sharing or not, @lichuang any suggestion?

lichuang: yes, when `db_info.from_share` is_some, it means that this db is created from a share

lichuang (approved):

Issue #10981, opened 2023-04-07
Use pipeline to refactor index analysis


zhang2014:
**Summary**

Use pipeline to refactor index analysis

PR #10982, opened 2023-04-07
fix(query): parquet don't need sort by meta infos

sundy-li:

## Summary

Summary about this PR

Partof #10981


RinChanNOWWW (approved):
Why parquet don't need this top k optimization?

PR #11002, opened 2023-04-10
feat: range frame supports all number types.

RinChanNOWWW:

## Summary

Range frame supports all number types:

- Integers
- Floats
- Decimals
- Date
- Timestamp

### Planner

For RANGE with offsets, there should be exactly one ORDER BY, like:

```sql
-- offset1 and offset2 are literals
SELECT FUNC() OVER  (ORDER BY o BEWTEEN offset1 PRECEDING AND offset2 FOLLOWING);
```

Then we should unify the types of `o`, `offset1` and `offset2`. And add a `CAST` operator to them.

### Transform

Rewrite `TransformWindow` to a generic type `TransformWindow<T>`.

Split the method implementations into 3 parts:

- Common: `impl<T> for TransformWindow<T>`.
- ROWS only: `impl TransformWindow<u64>`.
- RANGE and ROWS: `impl<T: Number + ResultTypeOfUnary> for TransformWindow<T>`.


ariesdevil (approved):
LGTM :)

Issue #10976, opened 2023-04-07
Fix bugs for tpcds in cluster mode


xudong963:
**Summary**
![iBPSgq2LQ2](https://user-images.githubusercontent.com/41979257/230570805-7c8055b9-74cf-46ac-b189-7ed2d12015ac.jpg)

xudong963:
```sql
 WITH
     max_store_sales AS
  (SELECT max(csales) tpcds_cmax
   FROM
     (SELECT c_customer_sk,
             sum(ss_quantity*ss_sales_price) csales
      FROM store_sales,
           customer
      GROUP BY c_customer_sk) sq2),
     best_ss_customer AS
  (SELECT *
        from max_store_sales)

  SELECT tpcds_cmax
   FROM
        best_ss_customer
   UNION ALL SELECT tpcds_cmax
   FROM
        best_ss_customer
  ;

```

PR #11008, opened 2023-04-10
fix: union panic in cluster

xudong963:

## Summary

Summary about this PR

Closes https://github.com/datafuselabs/databend/issues/10976


RinChanNOWWW (approved):

sundy-li (approved):

src/query/sql/src/planner/optimizer/property/enforcer.rs
```diff
@@ -57,12 +57,19 @@ pub fn require_property(
...
-            continue;
+            // Check if exists `Merge` in child, if not exits, wrap it with `Exchange`
+            if optimized_expr
+                .children()
+                .iter()
+                .all(|child| !check_merge(child))
+            {
+                let enforced_child =
+                    enforce_property(optimized_expr.child(index)?, &RequiredProperty {
+                        distribution: Distribution::Random,
```
leiysky, line 68: `Distribution::Random` is used to describe derived property and shouldn't be used in `RequiredProperty`.

src/query/sql/src/planner/optimizer/property/enforcer.rs
```diff
@@ -108,3 +115,17 @@ pub fn enforce_distribution(distribution: &Distribution, s_expr: &SExpr) -> Resu
         )),
     }
 }
+
+fn check_merge(s_expr: &SExpr) -> bool {
```
leiysky, line 119: I think what you need is checking if the derived distribution of a `SExpr` is `Distribution::Serial`.

xudong963: In fact, I want to check if union's child plan has `Exchange::Merge`, if has, there will no `ExchangeSource`.So we don't make union shuffle.

PR #11012, opened 2023-04-10
chore(query): make embedding_vector returns default if not valid

### LINKED ISSUE #10985: bug: function in update will eval for all rows
**Summary**

There is only one row:
image

When run this query:
```
update elon set embedding = ai_embedding_vector(content) where path='2022-07-05 17:55:09+00:00';
```

The `ai_embedding_vector` will evaluate for all rows:

image

sundy-li:

## Summary

Summary about this PR

Closes #10985


BohuTANG (approved):

zhyass:
Can link with the issue #10985?

PR #11013, opened 2023-04-10
docs: add window function docs

ariesdevil:

## Summary

add window function docs

Tracking: #10810


BohuTANG:
Some suggestions:
1. Add a index.md for the window functions
2. Put to a table list is not suite for windows function, because we need some examples for each function, suggest put each function to a single markdown page with syntax and example.

sundy-li (approved):

PR #11016, opened 2023-04-10
feat(planner): Refactor outer join elimination with constraint

leiysky:

## Summary

Use constraint to eliminate outer joins.

### DIFF
--- a/src/query/constraint/src/problem.rs
+++ b/src/query/constraint/src/problem.rs
@@ -22,10 +22,6 @@ use z3::Solver;
 
 use crate::declare::is_not_null_int;
 
-/// NOTICE: This function is only valid for the predicates that only
-/// contain a single variable. For example, `a > 0` is valid, but
-/// `a + b > 0` is not.
-///
 /// Assert that an integer is not null with a given solver.
 /// We will check this by adding a new constraint to the solver:
 ///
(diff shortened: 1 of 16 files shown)


src/query/constraint/src/problem.rs
```diff
@@ -57,3 +58,30 @@ pub fn assert_int_is_not_null(
     solver.pop(1);
     result
 }
+
+/// The same as `assert_int_is_not_null`, but for multiple variables.
+pub fn assert_int_is_not_null_multiple_variables(
```
andylokandy, line 63: so we can remove assert_int_is_not_null because assert_int_is_not_null_multiple_variables covers the use case.

andylokandy: and rename it to assert_all_ints_are_not_null

src/query/sql/src/planner/optimizer/property/constraint.rs
```diff
@@ -0,0 +1,291 @@
...
+            .iter()
+            .map(|c| as_z3_ast(&context, c))
+            .collect::<Option<Vec<Dynamic>>>();
+
+        if let Some(z3_asts) = z3_asts {
+            let conjunctions = z3_asts
+                .iter()
+                .map(|a| is_true(&context, a))
+                .collect::<Vec<_>>();
+            let proposition = Bool::and(&context, &conjunctions.iter().collect::<Vec<_>>());
```
andylokandy, line 92:
```suggestion
            let proposition = Bool::and(&context, &conjunctions);
```

andylokandy (approved):

src/query/sql/src/planner/optimizer/property/constraint.rs
```diff
@@ -0,0 +1,291 @@
...
+            .any(|scalar| scalar.used_columns().contains(variable))
+        {
+            // If the variable isn't used by any constraint, then it's unconstrained.
+            return false;
+        }
+
+        let context = Context::new(&Config::new());
+        let variable = Int::new_const(&context, variable.to_string().as_str());
+
+        let z3_asts = self
```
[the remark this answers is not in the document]
leiysky: Then it will become a self-referential structure. I don't want to expose the z3 context, so I have to construct the solver on the fly.

xudong963 (approved):

PR #11017, opened 2023-04-10
refactor(storage): optimize purge

zhyass:

## Summary

Do purge during read snapshots, to avoid oom.
The data during the retention period is not purge.

Closes #issue


tests/suites/0_stateless/17_altertable/17_0002_alter_table_purge_before.sh
```diff
@@ -54,10 +54,14 @@ echo "drop table t17_0002 all" | $MYSQL_CLIENT_CONNECT
 echo "create table t17_0002(c int)" | $MYSQL_CLIENT_CONNECT
 ## - 1st snapshot contains 2 rows, 1 block, 1 segment
 echo "insert into t17_0002 values(1),(2)" | $MYSQL_CLIENT_CONNECT
+## take a nap
+sleep 0.002
```
dantengsky, line 58: without these`sleep`s, will this test suite fail?

zhyass: > without these`sleep`s, will this test suite fail?

There is a certain chance of failure, because navigate needs to rely on the modified time of the snapshot file. If do not add sleep, there is a possibility that  the modified time of the snapshot file are not in order, resulting in a situation where the number of snapshot files after purge is not correct (but will not affect the data).

dantengsky (approved):

PR #11022, opened 2023-04-11
refactor(query): better ai functions

sundy-li:

## Summary

Summary about this PR

Closes #issue


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1966,112 +1963,6 @@ impl<'a> TypeChecker<'a> {
                     Err(e) => Err(e),
                 })
             }
-            ("ai_embedding_vector", args) => {
```
BohuTANG, line 1969: 👍 

Also need to remove from `all_rewritable_scalar_function`:
```diff
            "ai_embedding_vector",
            "ai_text_completion",
```

BohuTANG (approved):

PR #11088, opened 2023-04-16
feat: lazy materialize according to virtual column `_row_id`.

### LINKED ISSUE #11126: bug: TopN query with `_row_id` will panic if the storage is native format in debug build.
**Summary**

```sql
select b, _row_id from t order by a limit 10; -- storage of `t` is native format.
```

### LINKED ISSUE #11138: bug: tpcds load_data.sh typo
### Search before asking

### Version

https://github.com/datafuselabs/databend/tree/main/benchmark/tpcds

### What's Wrong?

the env.sh in the file should be shell_env.sh

### How to Reproduce?

open the file and search env.sh

RinChanNOWWW:

## Summary

- Add a new transform `TransformNativeRowsFetcher`. It fetches and deserializes specific columns.
- Analyze if a query can be optimized by lazy materialization in binder. Currently, only simple Top-N query on Fuse engine can be optimized. (single table, no aggregation, no cte).

```sql
select ... from t ... where ... order by ... limit ... ;
```

- Add a new setting value `lazy_top_threshold` to control the behavior: Enable lazy materialization and set the limit threshold of Top-N queries. Set the value to 0 to disable this setting.
- Fix #11126
- Fix typo #11138 

Closes #11126
Closes #11138


src/query/catalog/src/table.rs
```diff
@@ -110,8 +110,13 @@ pub trait Table: Sync + Send {
         false
     }
 
-    /// Whether the table engine supports virtual columns optimization.
-    fn support_virtual_columns(&self) -> bool {
+    /// Whether the table engine supports virtual columns optimization for Variant types.
+    fn support_variant_virtual_columns(&self) -> bool {
```
b41sh, line 114: Virtual columns can be used not only for variant type, but also for internal columns of Map type or other nested types.

RinChanNOWWW: I want to distinguish row_id and other virtual columns.

b41sh: To distinguish it from a `VirtualColumn`, the `row_id` can be treated as an `InternalColumn`, for example in the definition of `ColumnEntry`. https://github.com/datafuselabs/databend/blob/main/src/query/sql/src/planner/metadata.rs#L413

src/query/catalog/src/table_context.rs
```diff
@@ -101,6 +101,9 @@ pub trait TableContext: Send + Sync {
     fn get_cacheable(&self) -> bool;
     fn set_cacheable(&self, cacheable: bool);
 
+    fn set_snapshot(&self, location: String, ver: u64);
+    fn get_snapshot(&self) -> Option<(String, u64)>;
```
lichuang, line 105: I think type `Location` is better to represent `location: String, ver: u64` tuple

src/query/sql/src/planner/binder/select.rs
```diff
@@ -667,6 +693,96 @@ impl Binder {
...
+        select_list: &SelectList,
+        where_scalar: &Option<ScalarExpr>,
+        order_by: &[OrderItem],
+        limit: usize,
+    ) -> Result<()> {
+        // Only simple single table Top-N query is supported.
+        // e.g.
+        // SELECT ... FROM t WHERE ... ORDER BY ... LIMIT ...
+        if stmt.group_by.is_some()
+            || stmt.having.is_some()
```
sundy-li, line 712: no `unnest`

src/query/service/src/pipelines/pipeline_builder.rs
```diff
@@ -992,6 +1006,24 @@ impl PipelineBuilder {
         })
     }
 
+    fn build_row_fetch(&mut self, row_fetch: &RowFetch) -> Result<()> {
+        debug_assert!({
```
zhang2014, line 1010:
```suggestion
        debug_assert!(matches!(&*row_fetch.input, PhysicalPlan::Limit(limit) if matches!(*limit.input, PhysicalPlan::Sort(_)));
```

zhang2014 (approved):

src/query/storages/fuse/src/operations/read/fuse_row_fetcher.rs
```diff
@@ -0,0 +1,283 @@
...
+            ver,
+            self.fetcher.reader().operator.clone(),
+            self.table_schema.clone(),
+            &self.projection,
+        )
+        .await?;
+        Ok(())
+    }
+
+    // #[async_backtrace::framed]
```
b41sh, line 220: unused comment?

RinChanNOWWW: uncommented.

sundy-li (approved):

src/query/sql/src/executor/format.rs
```diff
@@ -602,6 +604,40 @@ fn limit_to_format_tree(
     Ok(FormatTreeNode::with_children("Limit".to_string(), children))
 }
 
+fn row_fetch_to_format_tree(
```
b41sh, line 607: How about adding an `explain` test case?

RinChanNOWWW:
@mergify refresh

RinChanNOWWW:
Thank @zhang2014 very very much for helping to solve the stack overflow issue!!!

Issue #11092, opened 2023-04-17
Feature: add query enterprise binary and databend-enterprise feature


lichuang:
**Summary**

add query enterprise binary, enterprise features only support in enterprise binary.

PR #11093, opened 2023-04-17
feat: add query enterprise binary and databend-enterprise feature

lichuang:

## Summary

feature: add query enterprise binary and databend-enterprise feature

enterprise feature is feature under `databend-enterprise`

Closeshttps://github.com/datafuselabs/databend/issues/11092

### DIFF
--- a/src/binaries/query-enterprise/Cargo.toml
+++ b/src/binaries/query-enterprise/Cargo.toml
@@ -1,0 +1,88 @@
+[package]
+name = "databend-query-enterprise"
+description = "databend command line tools"
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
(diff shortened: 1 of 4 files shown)


src/binaries/query-enterprise/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+name = "databend-query-enterprise"
+description = "databend command line tools"
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+publish = { workspace = true }
+edition = { workspace = true }
+
+[features]
+default = ["simd", "jemalloc", "databend-enterprise"]
```
sundy-li, line 11: I don't think we should add an extra cargo file.

Why not use features to control that ?

PR #11094, opened 2023-04-17
feat(function): support window function percent_rank.

youngsofun:

## Summary

percent_rank operate on the entire partition.

>>> Standard SQL specifies that window functions that operate on the entire partition should have no frame clause. MySQL permits a frame clause for such functions but ignores it. These functions use the entire partition even if a frame is specified

see https://dev.mysql.com/doc/refman/8.0/en/window-functions-frames.html

Postgres do exactly the same.

Tracking in https://github.com/datafuselabs/databend/issues/10810

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
+++ b/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
@@ -90,6 +90,7 @@ pub struct TransformWindow<T: Number> {
     partition_start: RowPtr,
     partition_end: RowPtr,
     partition_ended: bool,
+    partition_size: usize,
 
     // Frame: [`frame_start`, `frame_end`). `frame_end` is excluded.
     frame_unit: WindowFuncFrameUnits,
@@ -109,8 +110,6 @@ pub struct TransformWindow<T: Number> {
     frame_end: RowPtr,
     frame_started: bool,
     frame_ended: bool,
-    // for percent_rank/cume_dist
-    frame_size: usize,
 
     // Can be used to optimize window frame sliding.
     prev_frame_start: RowPtr,
@@ -229,6 +228,7 @@ impl<T: Number> TransformWindow<T> {
                 return;
             }
             self.partition_end.row += 1;
+            self.partition_size += 1;
         }
 
         assert_eq!(self.partition_end.row, block_rows);
@@ -474,10 +474,10 @@ impl<T: Number> TransformWindow<T> {
                 )));
             }
             WindowFunctionImpl::PercentRank => {
-                let percent = if self.frame_size <= 1 {
+                let percent = if self.partition_size <= 1 {
                     0_f64
                 } else {
-                    ((self.current_rank - 1) as f64) / ((self.frame_size - 1) as f64)
+                    ((self.current_rank - 1) as f64) / ((self.partition_size - 1) as f64)
                 };
                 builder.push(ScalarRef::Number(NumberScalar::Float64(percent.into())));
             }
@@ -538,6 +538,7 @@ impl TransformWindow<u64> {
             partition_start: RowPtr::default(),
             partition_end: RowPtr::default(),
             partition_ended: false,
+            partition_size: 0,
             frame_unit: WindowFuncFrameUnits::Rows,
             start_bound,
             end_bound,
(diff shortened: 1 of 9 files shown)


ariesdevil (approved):

ariesdevil:
Better to add same logic test for the named window.

youngsofun:
> Better to add same logic test for the named window.

done

tests/sqllogictests/suites/query/window_function/named_window_basic.test
```diff
@@ -1,278 +1,293 @@
...
+develop 5200 3
+develop 5200 3
+develop 6000 5
+personnel 3500 1
+personnel 3900 2
+sales 4800 1
+sales 4800 1
+sales 5000 3
+
+# percent_rank
```
youngsofun, line 173: only add this test.
no idea why lead to so many diff.

youngsofun: image

RinChanNOWWW: Maybe the file encoding format is changed?

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -109,6 +109,8 @@ pub struct TransformWindow<T: Number> {
     frame_end: RowPtr,
     frame_started: bool,
     frame_ended: bool,
+    // for percent_rank/cume_dist
+    frame_size: usize,
```
RinChanNOWWW, line 113: We can use partition rows.

The definition in DuckDB:

```
The relative rank of the current row: (rank() - 1) / (total partition rows - 1).
```

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -734,6 +746,33 @@ where T: Number + ResultTypeOfUnary
...
+            for block in (start.block + 1)..end.block {
+                count += self.blocks[block].block.num_rows();
+            }
+            count
+        }
+    }
+
+    fn count_rows_in_frame(&mut self) {
+        self.frame_size = self.count_rows(&self.frame_start, &self.frame_end)
+    }
```
RinChanNOWWW, line 763: It seems like this is not needed. We only need to count the partition size while `advance_partition`.

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -734,6 +746,33 @@ where T: Number + ResultTypeOfUnary
...
+
+    fn compute_on_frame(&mut self) -> Result<()> {
+        match &self.func {
+            WindowFunctionImpl::Aggregate(agg) => self.apply_aggregate(agg),
+            WindowFunctionImpl::PercentRank => {
+                self.count_rows_in_frame();
```
RinChanNOWWW, line 769: There will not be reached if partition is not ended. We can add a assertion here, like `debug_assert!(self.partition_eneded`.

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
youngsofun: seems snowflake just only this in SQL but still use whole partitition for compute

RinChanNOWWW: The frame definition will not affect the result. Just like `row_number` and `rank`.

youngsofun: done.

I mean snowflake not allow ` range between 1 preceding and 1 following`,
while mysql and pg just allow any frame clause by ignore it.

RinChanNOWWW: Ok.

RinChanNOWWW (approved):
LGTM

youngsofun:
[build_x86_64_musl](https://github.com/datafuselabs/databend/actions/runs/4718922217/jobs/8369094092#logs)
failed 24 minutes ago in 28m 36s. [View latest attempt](https://github.com/datafuselabs/databend/actions/runs/4718922217).

ariesdevil:
@youngsofun should add doc for this function like: https://databend.rs/doc/sql-functions/window-functions/window-function-rank and put it here: https://databend.rs/doc/sql-functions/window-functions/#list-of-functions-that-support-windows

PR #11109, opened 2023-04-18
chore: remove broadcast join limitation

xudong963:

## Summary

Summary about this PR

Closes #issue


xudong963:
A little improvement
image

We need to continue to improve the scalability of cluster :)
image

src/query/sql/src/planner/plans/join.rs
```diff
@@ -423,8 +423,6 @@ impl Operator for Join {
                     | JoinType::RightSemi
                     | JoinType::RightMark
             )
-            && probe_physical_prop.distribution != Distribution::Broadcast
```
xudong963, line 426: This code was added when fixing a bug in tpcds, but now it seems to be redundant.

Dousir9 (approved):

PR #11123, opened 2023-04-19
fix: copy from parquet oom

youngsofun:

## Summary

see 2 commits.


src/query/pipeline/sources/src/input_formats/input_pipeline.rs
```diff
@@ -215,6 +214,7 @@ pub trait InputFormatPipe: Sized + Send + 'static {
             "copy read {max_splits} splits in parallel, according to max_memory={max_memory}, num_fields={per_split_io}, max_storage_io_requests={max_storage_io_requests}, max_split_size={}",
             sizes[0]
         );
+        Self::build_pipeline_aligned(&ctx, data_rx, pipeline, max_splits)?;
```
ariesdevil, line 217: Why using `max_splits` here and the param in `build_pipeline_aligned` is `num_threads`, what is the correspondence between these two here?

ariesdevil (approved):

sundy-li (approved):

PR #11142, opened 2023-04-20
feat(function): Support `json_path_query` using JSON path

### LINKED ISSUE #10304: Support for more JSON path functions
**Summary**

In OLAP usecases data locality tkaes precedence over data normalization the usage of nested data is very common practice. People use arrays and arrays of structs to hold data together with main data.

Eg: If we need to design an account table with multiple addresses we create address as JSON and store array of address as JSON objects.

To better handle these nested datatypes specially list of structs we need a better JSON Path handling. PostgreSQL has functions like 

jsonb_path_query_first
jsonb_path_query
jsonb_path_query_array

These type of functions will be useful for handling nested structures.

Integrating https://github.com/besok/jsonpath-rust library to expose these type functions will greatly help usability.

b41sh:

## Summary

Support JSON path functions:
- `jsonb_path_query_first`
- `jsonb_path_query`
- `jsonb_path_query_array`

Closes #10304


src/query/expression/src/function.rs
```diff
@@ -77,8 +77,11 @@ pub enum FunctionEval {
     SRF {
         /// Given multiple rows, return multiple sets of results
         /// for each input row, along with the number of rows in each set.
-        eval:
-            Box<dyn Fn(&[ValueRef<AnyType>], usize) -> Vec<(Value<AnyType>, usize)> + Send + Sync>,
+        eval: Box<
+            dyn Fn(&[ValueRef<AnyType>], usize, &mut EvalContext) -> Vec<(Value<AnyType>, usize)>
```
sundy-li, line 81: why do we need `num_rows` as arg?

b41sh: We can get `num_rows` from the `EvalContext`, it looks like `num_rows` can be removed.

src/query/functions/tests/it/scalars/testdata/variant.txt
```diff
@@ -24,13 +24,12 @@ output domain  : Undefined
...
-raw expr       : parse_json(" \t")
-checked expr   : parse_json<String>(" \t")
-optimized expr : 0x2000000000000000
-output type    : Variant
-output domain  : Undefined
-output         : null
+error: 
+  --> SQL:1:1
+  |
+1 | parse_json('    ')
```
andylokandy, line 30: is this expected?

b41sh: yes, empty string is not valid json

PR #11168, opened 2023-04-21
feat: gc orphan segment and block files

lichuang:

## Summary

feat: gc orphan segment and block files

test cases include:
```
    //  this test case test that:
    //  - all the files referenced by snapshot(no matter if or not is an orphan snapshot),
    //    if or not within retention time, will not be purged;
    //  - all the orphan files that not referenced by any snapshots:
    //      - if within retention, will not be purged
    //      - if outof retention, will be purged
```

benchmark memory test:
* env: Mac m1,minio
* memory stat:
```
//table: 
create database db;

create table db.t (
id1 int,
id2 int,
id3 int,
id4 int,
id5 int,
id6 int,
id7 int,
id8 int,
id9 int,
id10 int,
id11 int,
id12 int,
id13 int,
id14 int,
id15 int,
id16 int,
str1 String,
str2 String,
str3 String,
dt timestamp
);
```

script: ./bench8023 --thread-nums=40 --max-allowed-thread-nums=40 --host=127.0.0.1 --user=root --password=root --port=3307 --batch-rows=1 --total-rows=10000

so there will be one segmeng per {block|block index}.

when insert data has `10167` number of segment,block,index, the memory stat is:
```
411249392(memory) 451584(rss)
``` 

when insert data has `20104` number of segment,block,index, the memory stat is:
```
412429040(memory) 1249152(rss) 
``` 

so, when segment\block\index increase, the memory avg is `(412429040 - 411249392) / (20104 - 10167) = 118 byte per segment`, the avg rss is `(1249152 - 451584) / (20104 - 10167) = 80 byte per segment`.

Closes #issue


src/query/storages/fuse/src/io/snapshots.rs
```diff
@@ -226,11 +256,276 @@ impl SnapshotsIO {
         }
     }
 
+    // return from files that last modified time bigger than timestamp
+    #[tracing::instrument(level = "debug", skip_all)]
+    #[async_backtrace::framed]
+    pub async fn get_within_time_files(
```
zhyass, line 262: How about using [list_files](https://github.com/datafuselabs/databend/blob/main/src/query/storages/fuse/src/operations/navigate.rs#L232) directly?

src/query/storages/fuse/src/io/snapshots.rs
```diff
@@ -226,11 +256,276 @@ impl SnapshotsIO {
...
+                file_set.insert(file);
+            }
+        }
+
+        Ok(file_set)
+    }
+
+    #[tracing::instrument(level = "debug", skip_all)]
+    #[async_backtrace::framed]
+    async fn read_segment_blocks(
```
zhyass, line 330: Use [get_block_locations](https://github.com/datafuselabs/databend/blob/main/src/query/storages/fuse/src/operations/gc.rs#L430) ?

BohuTANG:
Nice work! 👍 
However, we still need to add more tests to ensure higher data security before we can merge this PR.

src/query/storages/fuse/src/io/snapshots.rs
lichuang: `get_within_time_files` input a set of files, and filter out files that within time, but `list_files` cannot do that.

zhyass: I mean the fn `list_files` in [navigate.rs](https://github.com/datafuselabs/databend/blob/main/src/query/storages/fuse/src/operations/navigate.rs#L232-L238), add a new arg `prefix`.
```
impl FuseTable {
    #[async_backtrace::framed]
    async fn list_files<P>(&self, time_point: DateTime<Utc>, prefix: String, mut pred: P) -> Result<Vec<String>>
    where P: FnMut(String) {
            let op = self.operator.clone();
            ... ...
     }

        async fn get_orphan_files_to_be_purged(&self, time_point, referenced_files: BTreeSet<String>, ... ) {
               // 1. Filter out all the with in files.
               let files_to_be_purged = self.list_files(time_point, |_| {}).await?;
               // 2. Filter out all the referenced files.
               files_to_be_purged.retain(|location| !referenced_files.contains(location));
               Ok(files_to_be_purged)
        }
}
```

src/query/storages/fuse/src/io/snapshots.rs
```diff
@@ -43,10 +43,34 @@ use crate::io::MetaReaders;
...
+        }
+        files
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct SnapshotLiteExtended {
     pub format_version: u64,
     pub snapshot_id: SnapshotId,
+    pub prev_snapshot_id: Option<SnapshotId>,
```
zhyass, line 73: `prev_snapshot_id` can be removed?

BohuTANG:
Need to redesign this feature with `vacuum`, convert to draft state.

PR #11169, opened 2023-04-21
chore: refine the default column value error

BohuTANG:

## Summary

* better error msg when the default column value is not constant expression
* better the doc for create table and alter table add/drop column

Closes #issue


BohuTANG:
Only change a error message and improve some documents related to the create table, let's merge.

PR #11209, opened 2023-04-24
docs: make the sql reference related docs self-contained

flaneur2020:

## Summary

This PR tries to make the three chapter avoid referencing any other chapters, so these three chapters can be cherry-picked as a standalone docs in other places.

- move manage-settings.md from `10-deploy` to `13-sql-reference`
- use absolute img position
- removes a link to stream-api


BohuTANG (approved):
💯

PR #11210, opened 2023-04-24
fix: fix purge test bugs

lichuang:

## Summary

fix: fix purge test bugs

bug reason: in function `navigate_for_purge` use `snapshot.timestamp` as purge timestamp, but in `list_files` use `last_modified` time to get file list, `last_modified` sometimes may bigger than `snapshot.timestamp`, so some files be ignored.

use `do_purge`  instead of sql.

Closes #issue


zhyass (approved):
LGTM

PR #11212, opened 2023-04-24
fix:  the map `build_segment_id_map returns` should contain format ver

dantengsky:

## Summary

- the map `build_segment_id_map` returns should contain format ver
  i.e. instead of `HashMap<String, usize>`, we should use `HashMap<Location, usize>`
- `TransformRowsFetcher` should use the format version of segment to load the segment, not the version of the snapshot
- some minor refactor (avoid unnecessary cloning)

Closes #issue


RinChanNOWWW:
LGTM

PR #11216, opened 2023-04-24
refactor(optimizer): add constant folding

andylokandy:

## Summary

Summary about this PR

Closes #issue


src/query/sql/src/planner/optimizer/rule/rewrite/rule_fold_constant.rs
```diff
@@ -0,0 +1,184 @@
...
+    ) -> Result<()> {
+        let mut new_plan = s_expr.plan().clone();
+        match &mut new_plan {
+            RelOperator::EvalScalar(eval_scalar) => {
+                for scalar in eval_scalar.items.iter_mut() {
+                    scalar.scalar = self.fold_constant(&scalar.scalar)?;
+                }
+            }
+            RelOperator::Filter(filter) => {
+                for predicate in filter.predicates.iter_mut() {
```
sundy-li, line 146: For `RelOperator::Filter`, could combine with `and` operator and `cast_expr_to_non_null_boolean`.

Then apply `fold_constant`, thus we don't need do it again in physical_plan_builder.

src/query/sql/src/planner/optimizer/rule/rewrite/rule_fold_constant.rs
```diff
@@ -0,0 +1,184 @@
...
+                for cond in join.right_conditions.iter_mut() {
+                    *cond = self.fold_constant(cond)?;
+                }
+                for cond in join.non_equi_conditions.iter_mut() {
+                    *cond = self.fold_constant(cond)?;
+                }
+            }
+            _ => unreachable!(),
+        }
+        dbg!(s_expr.plan());
```
sundy-li, line 173: remove these debug codes

src/query/sql/src/planner/optimizer/rule/rewrite/rule_fold_constant.rs
leiysky: It relies on the pre-split conjunctions to perform some pushdown optimization, I think it's better not to combine the conjunctions here into a single `ScalarExpr`.

src/query/sql/src/planner/optimizer/rule/rewrite/rule_fold_constant.rs
```diff
@@ -0,0 +1,184 @@
...
+pub struct RuleFoldConstant {
+    id: RuleID,
+    patterns: Vec<SExpr>,
+    func_ctx: FunctionContext,
+}
+
+impl RuleFoldConstant {
+    pub fn new(func_ctx: FunctionContext) -> Self {
+        Self {
+            id: RuleID::NormalizeScalarFilter,
```
leiysky, line 38:
```suggestion
            id: RuleID::FoldConstant,
```

src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -1132,9 +1114,8 @@ impl PhysicalPlanBuilder {
 
                     let expr = cast_expr_to_non_null_boolean(expr)?;
 
-                    let (expr, _) = ConstantFolder::fold(&expr, &self.func_ctx, &BUILTIN_FUNCTIONS);
```
sundy-li, line 1135: We can't remove these codes because the `rule_fold_constants` did not fold the whole expr (it only fold the sub expr).

leiysky (approved):

sundy-li (approved):

src/query/sql/src/planner/optimizer/rule/rewrite/rule_fold_constant.rs
```diff
@@ -0,0 +1,182 @@
...
+                        }
+                        .into(),
+                    ),
+                ),
+            ],
+            func_ctx,
+        }
+    }
+
+    fn fold_constant(&self, scalar: &ScalarExpr) -> Result<ScalarExpr> {
```
sundy-li, line 120: `as_expr` will turn agg into ColumnRef.
`from_expr` will not turn back into agg.

PR #11221, opened 2023-04-24
docs: load data from stage o bucket

soyeric128:

## Summary

added "loading from user stage" + "loading from external stage"
refactor "loading from a bucket"

Closes #issue


docs/doc/12-load-data/00-stage.md
```diff
@@ -1,201 +1,209 @@
...
+## Tutorial 1: Loading from User Stage
 
-Download [books.csv](https://datafuse-1253727613.cos.ap-hongkong.myqcloud.com/data/books.csv)
+Follow this tutorial to upload the sample file to the user stage and load data from the staged file into Databend.
 
-</TabItem>
+### Step 1: Upload Sample File
 
-<TabItem value="parquet" label="Parquet">
+1. Use cURL to make a request to the [File Upload API](../11-integrations/00-api/10-put-to-stage.md):
```
BohuTANG, line 43: It would be better to add we can upload files to stage with some Databend driver.

soyeric128: bendsql?

BohuTANG: We could do it in the another PR, also we should add a command 'PUT STAGE' to https://databend.rs/doc/sql-commands/ddl/stage/ with bendsql or others driver.

PR #11231, opened 2023-04-25
refactor: optimize reading small parquet.

youngsofun:

## Summary

normally, reading a parquet file on s3 is broken down into several steps
1.  read file meta when  `read_partitions`
2. read column chunks needed 

sometimes, there are many small files,  it is more efficient to read each file as a whole. 

details

1.  files smaller then setting `parquet_read_whole_file_threshold` is counted as small file, which is 0 by default, i.e. not enabled by default.
2.  up to `num_columns_to_read` small files are grouped together as a Partition,  to read in parallel. 
3.  all pruning logic is reserved and reused for each small file, thanks to the Memory backend in Opendal.  even if a file is small, it may still contain many rows.

some other changes:

1.  to avoid OOM, each source now read one split each time, instead of storage_fetch_part_num  splits, see https://github.com/datafuselabs/databend/pull/11165

Closes #issue


BohuTANG:
How about enabling this feature by default, as it will be beneficial for our users who are using the DatabendCloud? 
This way, they won't need to know how to configure it and can enjoy a faster experience within the Databend kernel.

Xuanwo (approved):

youngsofun:
> How about enabling this feature by default(or later), as it will be beneficial for our users who are using the DatabendCloud? This way, they won't need to know how to configure it and can enjoy a faster experience within the Databend kernel.
> 
> As well, a **benchmark** still need between enable this feature with small parquet files.

maybe we need diff default value for diff backend? cc @Xuanwo

Xuanwo:
> maybe we need diff default value for diff backend? cc @Xuanwo

I believe it's always a win even on local fs.

sundy-li:
a later benchmark results could lead us set a better default value for it.

PR #11238, opened 2023-04-25
feat: add table lock for compact

zhyass:

## Summary

1. Add `TableLock` before update/delete. Drop the `TableLock` when the pipeline finished.
   This is similar to add a table-level lock. When a mutation operation is executed, other mutation operations will wait to get the lock.
2. Compact doesnot add `TableLock` at the beginning,  it will check if the `TableLock` exists. And if exist lock, return an error directly. Only get `TableLock` when commit server. This is to try to avoid the compact blocking update/delete operations.
3. The process of requiring a lock is similar to [etcd](https://www.sobyte.net/post/2022-08/distributed-lock/).  The requested `TableLock` returns a global incremental revision, listing all existing revisions, and if the current revision is the smallest, the lock is acquired successfully. Otherwise, listen to the deletion event of the previous revision in a loop until get lock success.
4. `table_lock_expire_secs` is setting to sets the seconds that the table lock will expire in.  Add `TableLockHeartbeat` to prevent `TableLock`  from being expired before pipeline is complete.
5. The `TableLock` is available as a ee feature.

Closes #issue


drmingdrmer:
Can `AddTableMutationLockReply` serve as a guard that automatically releases the lock upon being dropped? Alternatively, if the process holding the lock crashes, who will be responsible for releasing the lock?

zhyass:
> Can `AddTableMutationLockReply` serve as a guard that automatically releases the lock upon being dropped? Alternatively, if the process holding the lock crashes, who will be responsible for releasing the lock?

Use `expire_at` in `AddTableMutationLockReq` to automatically releases.

drmingdrmer:
> > Can `AddTableMutationLockReply` serve as a guard that automatically releases the lock upon being dropped? Alternatively, if the process holding the lock crashes, who will be responsible for releasing the lock?
> 
> Use `expire_at` in `AddTableMutationLockReq` to automatically releases.

Certainly! That is another way. 

Is there a case the lock timed out while the lock holder is still running?

src/query/service/src/interpreters/interpreter_table_optimize.rs
```diff
@@ -45,81 +46,88 @@ impl Interpreter for OptimizeTableInterpreter {
...
+        let table = self
+            .ctx
+            .get_table(&self.plan.catalog, &self.plan.database, &self.plan.table)
+            .await?;
+        let table_info = table.get_table_info();
+        let catalog = self.ctx.get_catalog(&self.plan.catalog)?;
+        let res = catalog.get_table_mutation_lock(table_info).await?;
+        if res.locked {
+            return Err(ErrorCode::TableMutationAlreadyLocked(format!(
+                "table '{}' already locked for mutation",
```
ZhiHanZ, line 102: Consider change to 
table '{}' is under mutation, please retry compaction later
?

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2314,6 +2322,172 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+                table_meta = debug(&tb_meta),
+                "add_table_mutation_lock"
+            );
+
+            let lock_key = TableMutationLockKey { table_id };
+
+            let lock = TableMutationLock {};
+
+            let condition = vec![
+                txn_cond_seq(&tbid, Eq, tb_meta_seq),
```
dantengsky, line 2388: do we need to check the table mete sequence (version) in KV transaction? seems lock by table id is enough.  
the `retry` logic also seems to be unnecessary.

dantengsky:
> > > Can `AddTableMutationLockReply` serve as a guard that automatically releases the lock upon being dropped? Alternatively, if the process holding the lock crashes, who will be responsible for releasing the lock?
> > 
> > 
> > Use `expire_at` in `AddTableMutationLockReq` to automatically releases.
> 
> Certainly! That is another way.
> 
> Is there a case the lock timed out while the lock holder is still running?

:+1:  yes, that is possible, e.g. if the lock holder failed to maintain the heartbeat(which refreshes the TTL of the key) on time.

this may cause subsequent modification operations on the table to fail, and the existing conflict detection mechanism will continue to work (rollback the tx if necessary).

as I understand it, the lock introduced in this PR is only used to mitigate (but not completely avoid) the rollbacks caused by conflicts between user update/delete/replace operations and backend compact operations.

@zhyass hope I didn't misunderstand.

src/meta/api/src/schema_api_test_suite.rs
```diff
@@ -3879,6 +3883,105 @@ impl SchemaApiTestSuite {
...
+            ..TableMeta::default()
+        };
+
+        let created_on = Utc::now();
+
+        let req = CreateTableReq {
+            if_not_exists: true,
+            name_ident: db_table_name_ident.clone(),
+            table_meta: table_meta(created_on),
+        };
```
drmingdrmer, line 3923: use `Util` to simplify test setting up:
https://github.com/datafuselabs/databend/blob/b77f871a4c01094ade34c3db23f5ec9f11ffbedc/src/meta/api/src/schema_api_test_suite.rs#L3309-L3317

src/meta/app/src/schema/table.rs
```diff
@@ -741,6 +741,47 @@ pub struct TableCopiedFileLockKey {
...
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct ListTableLockRevReq {
+    pub table_id: u64,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct ListTableLockRevReply {
+    pub revisions: Vec<u64>,
+    pub prefix: String,
```
drmingdrmer, line 761: Is `prefix` necessary? I think the caller always knows what the `prefix` is.

src/meta/protos/proto/table.proto
```diff
@@ -165,3 +165,8 @@ message TableCopiedFileLock {
   uint64 ver = 100;
   uint64 min_reader_ver = 101;
 }
+
+message TableLock {
+  uint64 ver = 100;
+  uint64 min_reader_ver = 101;
+}
```
drmingdrmer, line 172: If it is empty, is it possible using `Empty` message as the lock value?

src/query/catalog/src/database.rs
```diff
@@ -174,4 +178,26 @@ pub trait Database: DynClone + Sync + Send {
...
+    #[async_backtrace::framed]
+    async fn delete_table_lock_rev(
+        &self,
+        _req: DeleteTableLockRevReq,
+    ) -> Result<DeleteTableLockRevReply> {
+        Err(ErrorCode::Unimplemented(format!(
+            "UnImplement delete_table_lock_rev in {} Database",
+            self.name()
+        )))
+    }
```
drmingdrmer, line 202: Locking/unlocking a table could be just an operation of a `Table`, because it uses a `TableId` to locate a Lock record. It does not need `database` to access a table lock.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,162 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+
+            let revision = if let Some(rev) = req.revision {
+                rev
+            } else {
+                // generate a new revision.
+                fetch_id(self, IdGenerator::table_lock_id()).await?
+            };
+
+            let lock_key = TableLockKey { table_id, revision };
+
```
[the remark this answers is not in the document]
zhyass: The scenario when we need to update the expire in `TableLockHeartbeat`
https://github.com/datafuselabs/databend/blob/76f5278c5e7c0906a93b041b1525d2343fa12db3/src/query/ee-features/table-lock/src/table_lock_heartbeat.rs#L72

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,162 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+            let (lock_key_seq, _): (_, Option<TableLock>) = get_pb_value(self, &lock_key).await?;
+
+            if lock_key_seq > 0 && req.revision.is_none() {
+                // lock key has been created by other process.
+                continue;
+            }
+
+            let lock = TableLock {};
+
+            let condition = vec![txn_cond_seq(&lock_key, Eq, lock_key_seq)];
```
lichuang, line 2601: if update lock need to assure that table is exist?

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,162 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+                "delete_table_lock_rev"
+            );
+
+            let lock_key = TableLockKey { table_id, revision };
+            let (lock_key_seq, _): (_, Option<TableLock>) = get_pb_value(self, &lock_key).await?;
+            if lock_key_seq == 0 {
+                return Ok(DeleteTableLockRevReply {});
+            }
+
+            let condition = vec![txn_cond_seq(&lock_key, Eq, lock_key_seq)];
```
lichuang, line 2665: if delete lock need to assure that table is exist?

src/meta/api/src/schema_api_impl.rs
zhyass: > if update lock need to assure that table is exist?

Do we need to check the table mete sequence (version)  here? @dantengsky

src/query/ee/src/table_lock/table_lock_handler.rs
```diff
@@ -0,0 +1,109 @@
...
+                ErrorCode::TableLockExpired("the acquired table lock has expired".to_string()),
+            )?;
+
+            if position == 0 {
+                // The lock is acquired by current session.
+                break;
+            }
+
+            // Get the previous revision, watch the delete event.
+            let key = format!("{}/{}", reply.prefix, reply.revisions[position - 1]);
```
lichuang, line 67: i think it is better to encapture a funciton such as `table_lock_key` to clarity this operation.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,162 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
         Ok(CountTablesReply { count })
     }
 
+    async fn list_table_lock_revs(
+        &self,
+        req: ListTableLockRevReq,
+    ) -> Result<ListTableLockRevReply, KVAppError> {
+        let prefix = format!("{}/{}", TableLockKey::PREFIX, req.table_id);
```
lichuang, line 2540: i think it is better to encapture a funciton such as table_lock_key to clarity this operation.

zhyass: > i think it is better to encapture a funciton such as table_lock_key to clarity this operation.

It's a prefix of the `TableLockKey`

sundy-li:
> The TableLock is available as a ee feature.

Does it mean that compaction/mutation will work without this feature? There will be more collision possibility.

zhyass:
> Does it mean that compaction/mutation will work without this feature? There will be more collision possibility.

Yes, it is possible to failure due to conflict when commit server.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,149 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+            trials.next().unwrap()?;
+
+            let tbid = TableId { table_id };
+            let (_tb_meta_seq, tb_meta) = get_table_by_id_or_err(self, &tbid, ctx).await?;
+
+            debug!(
+                ident = display(&tbid),
+                table_meta = debug(&tb_meta),
+                "upsert_table_lock_rev"
+            );
```
drmingdrmer, line 2576: Such a debug logging is already included in `get_table_by_id_or_err`

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,149 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+
+            let revision = if let Some(rev) = req.revision {
+                rev
+            } else {
+                // generate a new revision.
+                fetch_id(self, IdGenerator::table_lock_id()).await?
+            };
```
drmingdrmer, line 2583: Fetching an id can be moved out of the transaction try-loop.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,149 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+
+            let lock_key = TableLockKey { table_id, revision };
+
+            let (lock_key_seq, _): (_, Option<EmptyProto>) = get_pb_value(self, &lock_key).await?;
+
+            if lock_key_seq > 0 && req.revision.is_none() {
+                // lock key has been created by other process.
+                continue;
+            }
```
drmingdrmer, line 2592: It is impossible there is a duplicate lock_key, since the created revision is globally unique.

It looks like there should be two methods for creating a lock and for extending a lock lease respectively.

When creating a lock, it just assert the lock seq to be `0`.
When extending the lease, it asserts a specific lock seq.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -2525,6 +2533,149 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+                ident = display(&tbid),
+                table_meta = debug(&tb_meta),
+                "delete_table_lock_rev"
+            );
+
+            let lock_key = TableLockKey { table_id, revision };
+            let (lock_key_seq, _): (_, Option<EmptyProto>) = get_pb_value(self, &lock_key).await?;
+            if lock_key_seq == 0 {
+                return Ok(DeleteTableLockRevReply {});
+            }
```
drmingdrmer, line 2652: Deleting a lock does not need to fetch a TableMeta IMHO.

lichuang (approved):

drmingdrmer (approved):

PR #11247, opened 2023-04-26
feat(parser): support ANSI quoted identifier

### LINKED ISSUE #9115: Feature: Support ANSI identifier escaping

flaneur2020:

## Summary

currently the following statement will create a table called `t1""a`, but `t1"a` is expected:

```
create table "t1""a" (a int);
```

this PR add the quote escape logic to identifiers, so the SQL-concating tools like sql UI or DBT can quote identifiers safely.

This PR also fixed an issue about displaying the string literals, the string literal should be displayed with escape, like `'abc\'c'` is expected but currently it's displayed as `'abc'c'` which contains incorrect literal syntax.

Closes #9115


src/query/ast/tests/it/testdata/statement.txt
```diff
@@ -5926,6 +5970,51 @@ Query(
 )
 
 
+---------- Input ----------
+select 'stringwith''quote'''
+---------- Output ---------
+SELECT 'stringwith'quote''
```
flaneur2020, line 5976: here is a bad case about string's quote (not introduced in this PR), maybe the string literal needs some escaping

sundy-li (approved):

Issue #10632, opened 2023-03-18
bug: information_schema.column and system.columns will generate recursion 


TCeason:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

main version

### What's Wrong?

information_schema.columns is a view that will query system.columns

if query information_schema.columns will call dump_table_columns and information_schema also in CATALOG_DEFAULT, will recursion

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #11254, opened 2023-04-27
feat(query): show columns support display view engine

### LINKED ISSUE #11136: Feature: hope  to support view use "show full COLUMNS from view_xx form xx"
use view can not show COLUMNS .
There is no problem with the display in mysql view
image
There is no problem with the display in table
image
image

TCeason:

## Summary

Because of #10632, now show columns can not query the view engine field. 

In this pr, add ShowColumnInterpreter to support this.

We will generate a data block and if `show columns` have a limit, will filter the data block.

Closes #11136


src/query/service/src/interpreters/interpreter_show_columns.rs
```diff
@@ -0,0 +1,258 @@
...
+impl ShowColumnsInterpreter {
+    pub fn try_create(ctx: Arc<QueryContext>, plan: ShowColumnsPlan) -> Result<Self> {
+        Ok(ShowColumnsInterpreter { ctx, plan })
+    }
+}
+
+#[async_trait::async_trait]
+impl Interpreter for ShowColumnsInterpreter {
+    fn name(&self) -> &str {
+        "ShowCreateTableInterpreter"
```
b41sh, line 67: ShowColumnsInterpreter

youngsofun (approved):

src/query/service/src/interpreters/interpreter_show_columns.rs
```diff
@@ -0,0 +1,258 @@
...
+                if let Some(expr) = field.default_expr() {
+                    default_expr = expr.to_string();
+                }
+                exprs.push(default_expr.into_bytes());
+                if field.is_nullable() {
+                    is_nullables.push("YES".to_string().into_bytes());
+                } else {
+                    is_nullables.push("NO".to_string().into_bytes());
+                }
+                extras.push("".to_string().into_bytes());
```
b41sh, line 128: this column seems always empty, how about use `BlockEntry` with a `Scalar` value. like:
```rust
BlockEntry {
    data_type: DataType::String,
    value: Value::Scalar(Scalar::String(vec![])),
}
```

src/query/service/src/interpreters/interpreter_show_columns.rs
```diff
@@ -0,0 +1,258 @@
...
+                StringType::from_data(keys),
+                StringType::from_data(exprs),
+                StringType::from_data(extras),
+                StringType::from_data(privileges),
+                StringType::from_data(comments),
+            ])
+        } else {
+            // | Field | Type | Null | Key  | Default | Extra |
+            DataBlock::new_from_columns(vec![
+                StringType::from_data(names),
```
b41sh, line 154: It seems to have some repetitive code that could be simplified.

TCeason: I think we can not do. Need ensure that the sequence is the same as the schema

src/query/service/src/interpreters/interpreter_show_columns.rs
TCeason: If do this, can not ensure that the sequence is the same as the schema?

sundy-li:
`system.columns`  will not display view engine, but `show columns` will.

PR #11262, opened 2023-04-28
feat: covert like expression to comparison

xudong963:

## Summary

For example, converting `p_type like 'PROMO%` to `p_type >= 'PROMO' AND p_type < 'PROMP'`, can improve selectivity (like this form is difficult to predict selectivity), and potential filter efficiency,  `like regular matching` seems to be not very efficient.

Add constant fold to selectivity

Closes #issue


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1461,6 +1461,17 @@ impl<'a> TypeChecker<'a> {
                     data_type,
                 )))
             }
+            BinaryOperator::Like => {
+                // Convert `Like` to compare function , such as `p_type like PROMO%` will be converted to `p_type >= PROMO and p_type < PROMP`
+                if let Expr::Literal { lit, .. } = right {
+                    if let Literal::String(str) = lit {
+                        return self.resolve_like(op, span, left, right, str).await;
+                    }
+                }
```
xudong963, line 1470: Maybe we can directly return error in else? Seems right must be String?

leiysky: It can be an expression as well.

leiysky (approved):

sundy-li:
#11265

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -2955,3 +3010,42 @@ pub fn validate_function_arg(
...
+        if like_str.chars().nth(i - 1).unwrap() != '%' {
+            break;
+        }
+        i -= 1;
+    }
+    if i == like_str.len() {
+        return false;
+    }
+    for j in (0..i).rev() {
+        if like_str.chars().nth(j).unwrap() == '%' || like_str.chars().nth(j).unwrap() == '_' {
```
Dousir9, line 3046: Only need `like_str.chars().nth(j).unwrap() == '_'`, because the interval [0, i) has no `%`.

Dousir9 (approved):

PR #11271, opened 2023-04-28
fix: page reader should use num_values of chunk.

youngsofun:

## Summary

for List Type,  `num_values` of a leaf column may be larger then num_rows.

page reader currently use `num_values` to decide whether all pages is readn.

this bug may lead to some pages not read, and result in less rows in this column than others and report error.

see: 
https://github.com/jorgecarleitao/parquet2/blob/main/src/read/page/reader.rs#L53
https://github.com/jorgecarleitao/parquet2/blob/main/src/read/page/reader.rs#L179

Closes #issue

### DIFF
--- a/src/query/storages/parquet/src/parquet_part.rs
+++ b/src/query/storages/parquet/src/parquet_part.rs
@@ -32,6 +32,7 @@ use common_expression::Scalar;
 pub struct ColumnMeta {
     pub offset: u64,
     pub length: u64,
+    pub num_values: i64,
     pub compression: Compression,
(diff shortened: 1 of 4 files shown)


RinChanNOWWW:
`num_values` of each `ColumnMeta` should be all the same, and it is equal to the one in `BlockMeta`.

RinChanNOWWW (approved):

Xuanwo (approved):

sundy-li (approved):

Issue #11083, opened 2023-04-15
bug: create table option not check


BohuTANG:
**Summary**

```
create table t1(a int) xx=yy;
```

PR #11283, opened 2023-04-30
fix(create): fix create table not check

JackTan25:

## Summary

Summary about this PR

Closes #issue
#11083


BohuTANG:
Thanks for the contribution. 
It's better to add a SQL logic test to check whether this PR works as expected.

PR #11303, opened 2023-05-04
feat: add license parsing function

ZhiHanZ:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/src/common/license/src/license.rs
+++ b/src/common/license/src/license.rs
@@ -1,3 +1,17 @@
+// Copyright 2021 Datafuse Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use serde::Deserialize;
 use serde::Serialize;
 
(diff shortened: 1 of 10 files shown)


src/common/license/src/license.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub const LICENSE_PUBLIC_KEY: &str = r#"-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEGsKCbhXU7j56VKZ7piDlLXGhud0a
+pWjW3wxSdeARerxs/BeoWK7FspDtfLaAT8iJe4YEmR0JpkRQ8foWs0ve3w==
+-----END PUBLIC KEY-----"#;
+
+#[derive(Serialize, Deserialize)]
+pub struct LicenseInfo {
+    #[serde(rename = "type")]
+    pub type_: Option<String>,
```
everpcpc, line 12: this could be `#type`?

src/common/license/src/license.rs
```diff
@@ -0,0 +1,15 @@
+use serde::Deserialize;
+use serde::Serialize;
+
+pub const LICENSE_PUBLIC_KEY: &str = r#"-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEGsKCbhXU7j56VKZ7piDlLXGhud0a
+pWjW3wxSdeARerxs/BeoWK7FspDtfLaAT8iJe4YEmR0JpkRQ8foWs0ve3w==
+-----END PUBLIC KEY-----"#;
```
flaneur2020, line 7: where doese the public key comes from?

src/common/license/src/license.rs
flaneur2020: or naming it as `kind` to avoid conflict with the `type` keyword?

src/common/license/src/license.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct LicenseInfo {
+    #[serde(rename = "type")]
+    pub type_: Option<String>,
+    pub org: Option<String>,
+    pub tenants: Option<Vec<String>>,
```
flaneur2020, line 14: I guess just `Vec<String>` is fine in this case, as an empty vector can already represents the Empty case.

ZhiHanZ: option is needed. otherwise the input must similar to 
```json
{
  tenants: []
}
```
since the jwt-simple would not parse nil to Vec

PR #11305, opened 2023-05-04
feat: support to use config override parquet_fast_read_bytes and max_storage_io_requests settings

ZhiHanZ:

## Summary

1. Support to config on session settings through config for example:
```toml
 [query]
 parquet_fast_read_bytes = 52428800
 max_storage_io_requests = 128
```
2. support session settings override, load priority:
    1. Default value
    2. Config overrided value
    3. Global Settings
    4. Client Settings on Http Session API or other protocol set command


src/query/config/src/config.rs
```diff
@@ -1215,6 +1215,16 @@ pub struct QueryConfig {
...
 
+    /// add session scope settings override
+    /// For example:
+    /// ```toml
+    /// [query.session_settings]
+    /// parquet_fast_read_bytes = "52428800"
+    /// max_storage_io_requests = "128"
+    /// ```
+    #[clap(skip)]
+    pub session_settings: HashMap<String, String>,
```
flaneur2020, line 1226: I guess these fields can just put into the [query] or [storage] section instead of introducing a [session_settings] section.

The [session_settings] section may introduce some duplicated configuration entries with the [query] section or [storage] section.

BohuTANG:
This feature will be applied to specific settings with different warehouse size(small/medium/large) in the Cloud, such as `max_storage_io_requests`. cc @sundy-li @TCeason

TCeason (approved):

src/common/storage/src/config.rs
```diff
@@ -46,7 +46,7 @@ use serde::Serialize;
 pub struct StorageConfig {
     pub num_cpus: u64,
     pub allow_insecure: bool,
-
+    pub max_storage_io_requests: Option<u64>,
```
flaneur2020, line 49: it seems StorageConfig mostly concerns about the connection locations of different storage providers, maybe max_storage_io_requests can also be considered as a QueryConfig ?

src/query/config/src/config.rs
```diff
@@ -1214,6 +1214,15 @@ pub struct QueryConfig {
 
     #[clap(long, default_value = "10000")]
     pub max_query_log_size: usize,
+    /// Parquet file with smaller size will be read as a whole file, instead of column by column.
+    /// For example:
+    /// parquet_fast_read_bytes = 52428800
+    /// will let databend read whole file for parquet file less than 50MB and read column by column
+    /// if file size is greater than 50MB
+    pub parquet_fast_read_bytes: Option<u64>,
```
flaneur2020, line 1222: missed a #[clap] header here?

flaneur2020: if 52428800 is a feasible default config for the most workloads, we can add a default_value here?

ZhiHanZ: I think it depends on the storage io bandwidth for exampl this setup only take significant value for R2, otherwise it seems cannot introduce significant changes.

flaneur2020: IMO this parameter is benefical for the common workload, not only for the r2 workload.

@youngsofun @BohuTANG Is this understanding correct?

flaneur2020 (approved):
LGTM

PR #11320, opened 2023-05-05
feat: add vacuum impl

lichuang:

## Summary

feat: add vacuum impl

Closes #issue


src/binaries/query/oss_main.rs
```diff
@@ -47,5 +49,7 @@ fn main() {
 async fn main_entrypoint() -> Result<()> {
     let conf: InnerConfig = InnerConfig::load()?;
     init_services(&conf).await?;
+    // init oss license manager
+    OssLicenseManager::init()?;
```
flaneur2020, line 53: this line can be moved into the function init_services?

lichuang: `init_services` will be called in both `oss` and `ee` main, it is used to init all `ee` and `oss` version services, see `ee_main.rs`.

src/query/interface-manager/src/interface_mgr.rs
```diff
@@ -0,0 +1,63 @@
...
+use std::sync::Arc;
+
+use chrono::DateTime;
+use chrono::Utc;
+use common_base::base::GlobalInstance;
+use common_catalog::table_context::TableContext;
+use common_exception::Result;
+use common_storages_fuse::FuseTable;
+
+/// All `ee` interface need to add a fn in `InterfaceManager` trait
```
flaneur2020, line 24: Instead of a big interface which contains all the ee functions, I guess it'd better to seperate the different ee interfaces into different places to make each module more domain-specific.

src/binaries/query/oss_main.rs
flaneur2020: it's ok to be called in different places, the `ee` code can choose to override multiple singleton managers on their place.

ZhiHanZ (approved):

src/query/interface-manager/src/interface_mgr.rs
ZhiHanZ: Agreed, it could be an effective way to ensure no config breaking changes occur

flaneur2020 (changes requested):
The interface manager design needs some further discussion.

ZhiHanZ:
Can we add some integration tests to vacuum command?

src/query/interface-manager/src/interface_mgr.rs
```diff
@@ -0,0 +1,63 @@
...
+        retention_time: DateTime<Utc>,
+    ) -> Result<()>;
+}
+
+pub struct InterfaceManagerWrapper {
+    manager: Box<dyn InterfaceManager>,
+}
+
+unsafe impl Send for InterfaceManagerWrapper {}
+unsafe impl Sync for InterfaceManagerWrapper {}
```
drmingdrmer, line 41: Why does it need manually implementation of `Send + Sync`?

src/binaries/query/oss_main.rs
lichuang: global instance cannot be override, otherwise, it will `panicked at 'value has been set in global data registry'`

src/query/ee/src/storages/fuse/io/snapshots.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use common_exception::Result;
+use common_storages_fuse::io::SnapshotLiteExtended;
+use common_storages_fuse::io::SnapshotsIO;
+use storages_common_table_meta::meta::Location;
+use tracing::info;
+
+#[allow(clippy::too_many_arguments)]
+#[async_backtrace::framed]
+pub async fn get_snapshot_referenced_segments<T>(
```
flaneur2020, line 29: these two functions is not very business specific, maybe it's also feasible to be placed in the apache licensed part?

flaneur2020 (approved):

drmingdrmer (approved):

drmingdrmer (approved):

zhyass (approved):
LGTM

src/query/service/src/interpreters/interpreter_table_vacuum.rs
```diff
@@ -42,8 +47,50 @@ impl Interpreter for VacuumTableInterpreter {
...
+                            .get_table(ctx.get_tenant().as_str(), &db_name, &tbl_name)
+                            .await?;
+                        let fuse_table = FuseTable::try_from_table(table.as_ref())?;
+                        let mgr = get_interface_manager();
+                        mgr.do_vacuum(fuse_table, ctx, retention_time).await
+                    });
+                }
+
+                Err(may_error.as_ref().unwrap().clone())
+            });
```
dantengsky, line 91: do we need to keep the else branch?  will the `main_pipeline` always be empty in this case?

src/query/storages/fuse/src/operations/navigate.rs
```diff
@@ -229,13 +247,17 @@ impl FuseTable {
...
-            FUSE_TBL_SNAPSHOT_PREFIX,
-        );
+    pub async fn list_files<P>(
+        &self,
+        prefix: String,
+        time_point: DateTime<Utc>,
+        mut pred: P,
+        filter: impl for<'a> Fn(&'a String) -> bool + Copy + Send + Sync,
+        return_error: bool,
+    ) -> Result<Vec<String>>
```
dantengsky, line 257: can args like `filter`, `return_error` ... be squashed into to predicate `P`

src/query/ee/src/storages/fuse/io/snapshots.rs
BohuTANG: We can't place ee functions(only used by the ee) follow apache license.

tests/suites/1_stateful/11_vacuum/11_0000_vacuum.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
```
BohuTANG, line 1: Please add a `ee` prefix to the test file name, like:
```
tests/suites/1_stateful/11_vacuum/11_0000_ee_vacuum.py
```

PR #11322, opened 2023-05-05
feat: move table_test_fixture to test_kits crate

lichuang:

## Summary

feat: move table_test_fixture to test_utils crate

since `ee` version need some tests, so need to move `table_test_fixture` to a crate.

Closes #issue


src/query/service/src/test_utils/context.rs
```diff
@@ -23,16 +23,17 @@ use common_meta_app::principal::PasswordHashMethod;
 use common_meta_app::principal::UserInfo;
```
BohuTANG, line 23: `src/query/service/src/test_kits` is better than `src/query/service/src/test_utils`?

BohuTANG (approved):

PR #11328, opened 2023-05-06
chore: refactor snapshot_format_version function, fix possible snapshot version missmatch bug

lichuang:

## Summary

chore: refactor snapshot_format_version function, fix possible snapshot version missmatch bug

for shared-table, 

- the snapshot location and format version can not be determined at the time of construction (due to we do not support async table constructor yet).

- the method `Table::snapshot_loc` will read the table snapshot location from hint file **for each invocation**

- `snapshot_format_version` uses `snapshot_loc` method to get the location of snapshot, and parse the format version

- thus `read_table_snapshot` is taking the following risk

https://github.com/datafuselabs/databend/blob/aa97d2c5b57f1585abc8192e5fb76f9fc4157903/src/query/storages/fuse/src/fuse_table.rs#L252-L278

  the `loc` we get at line 253, may not be the `loc` that is used to parse the format version at line 255 (since hint file may have been changed),  if the format version changed, we may be  reading the snapshot file using incorrect(old) version
  
  

Closes #issue

### DIFF
--- a/src/query/storages/fuse/src/fuse_table.rs
+++ b/src/query/storages/fuse/src/fuse_table.rs
@@ -252,7 +252,7 @@ impl FuseTable {
     pub async fn read_table_snapshot(&self) -> Result<Option<Arc<TableSnapshot>>> {
         if let Some(loc) = self.snapshot_loc().await? {
             let reader = MetaReaders::table_snapshot_reader(self.get_operator());
-            let ver = self.snapshot_format_version().await?;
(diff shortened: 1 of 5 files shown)


zhang2014 (approved):

zhang2014 (approved):

dantengsky:
looks good to me.

@lichuang  I've added some extra comments in the PR's summary section, please have a look, hope I get it right

dantengsky (approved):

PR #11335, opened 2023-05-06
docs: refactor docs about drivers

hantmac:

## Summary

Refactor docs about databend driver according to https://docs.snowflake.com/en/developer-guide/drivers


BohuTANG:
`Docusaurus found broken links!`

`yarn build` on local will tell what's is borken :)

docs/doc/03-develop/01-Python/00-installing.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Installing the Python Connector
```
BohuTANG, line 2: It's better to combine them all to a single page, since each peice is not too long: https://databend.rs/doc/develop/, this is easy for share.
For the chapter, we can add subtitles for them.

hantmac: Make sense.

soyeric128:
Thank you for the great updates!

I have some general comments:

- Node.js: it seems we use the MySQL driver, we'd better explicitly clarify it at the beginning, and ask the user to refer to the official documents for the What & How questions instead of elaborating them on our page.
- Please clarify the added code is for a Tutorial / Example for users to follow along and practice, or a Procedure explaining a general step 2 step way they can put to their own job.
- Some of the developer pages (for example, the JDBC page) now have too many titles to read if you see the right side on the page, maybe you might need to reorganize the content.

soyeric128:
Screenshot 2023-05-06 at 11 42 01 AM

I have one more question. The snapshot says the connection uses a **native** xxxx interface, but on the Node.js page it downloads and uses a MySQL driver. Most cases use the word to imply "sth designed specifically for xx".

soyeric128:
If you dont mind, can i take over this PR from here ? Will re-organize the added content and let you know when its done

hantmac:
> If you dont mind, can i take over this PR from here ? Will re-organize the added content and let you know when its done

Of course, thanks for your help!

soyeric128:
> > If you dont mind, can i take over this PR from here ? Will re-organize the added content and let you know when its done
> 
> Of course, thanks for your help!

DONE.

BohuTANG (approved):

Issue #11302, opened 2023-05-04
CopyInto Need to inject cancel point


ZhiHanZ:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

1.1.26

### What's Wrong?

```sql
select sleep(300);
```

when cancel the previous SQL from databend http api through kill, 
```
show processlist
```
still see the query and computing resource is not released.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

flaneur2020:
@youngsofun help take a look?

sundy-li:
`Duration::from_secs(300)`  was too long, we should panic as soon as possible I think.

ZhiHanZ:
> `Duration::from_secs(300)` was too long to block a thread, we should throw the error as soon as possible.
> 
> Fixed in #11304

Not only sleep could not be killed, copy into SQL could not be killed as well, which may result in non-trivial impact on production use cases

sundy-li:
Ok, CopyInto Need to inject cancel point.

youngsofun:
copy can be killed after the executor is started. 
fail to kill if it is listing file before execut , It will be fixed by check aborting when start executor.

I`ll fix it later.

```
    pub fn kill(&self, cause: ErrorCode) {
        self.set_error(cause.clone());
        self.aborting.store(true, Ordering::Release);

 
        if let Some(executor) = self.executor.read().upgrade() {
            executor.finish(Some(cause));
        }
  }
```

BohuTANG:
Listing file(OpenDAL operation) also need be cancelable, for example:
 Copy from a dir which has many files, the lists may take a long time we can't cancel unitl it finished.

sundy-li:
listing of opendal is in streaming mode, so it could be integrated into the pipeline.

youngsofun:
> listing of opendal is in streaming mode, so it could be integrated into the pipeline.

listing is in read_partition which is before building of pipeline.

PR #11336, opened 2023-05-06
fix: enable kill query before pipeline started.

youngsofun:

## Summary

before this pr,  query can only be killed after pipeline is started(` ctx.set_executor()`).

it will be better if long  binding/read_parttiion can be interrupted, which needs another pr.  

Closes  https://github.com/datafuselabs/databend/issues/11302

### DIFF
--- a/src/query/service/src/interpreters/interpreter.rs
+++ b/src/query/service/src/interpreters/interpreter.rs
@@ -53,6 +53,11 @@ pub trait Interpreter: Sync + Send {
         InterpreterMetrics::record_query_start(&ctx);
         log_query_start(&ctx);
 
+        if let Err(err) = ctx.check_aborting() {
+            log_query_finished(&ctx, Some(err.clone()));
+            return Err(err);
+        }
+
         let mut build_res = match self.execute2().await {
             Ok(build_res) => build_res,
             Err(build_error) => {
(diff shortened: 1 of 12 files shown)


src/query/service/src/pipelines/executor/pipeline_executor.rs
```diff
@@ -397,14 +399,17 @@ impl Drop for PipelineExecutor {
...
-        let mut guard = self.on_finished_callback.lock();
-        if let Some(on_finished_callback) = guard.take() {
-            drop(guard);
-            let cause = Some(ErrorCode::Internal(
-                "Pipeline illegal state: not successfully shutdown.",
-            ));
-            if let Err(cause) = catch_unwind(move || on_finished_callback(&cause)).flatten() {
-                warn!("Pipeline executor shutdown failure, {:?}", cause);
+        let inited = *self.inited.lock();
+        if inited {
```
zhang2014, line 403: This will cause to a session leak. we must call 'on_finished'.

youngsofun: why session leak? drop session in a callback?

youngsofun: is it ok to add on_finished to  PipelineCompleteExecutor/PipelinePullingExecutor, and call it here

```
impl PipelineCompleteExecutor {
   ...
    pub fn on_finished(&self, cause: Option<ErrorCode>) {
        self.executor.on_finished(&cause).ok();
    }
}

```

zhang2014 (approved):
.

src/query/service/src/interpreters/interpreter.rs
```diff
@@ -92,17 +97,24 @@ pub trait Interpreter: Sync + Send {
...
+                return Err(err);
+            }
             complete_executor.execute()?;
-            return Ok(Box::pin(DataBlockStream::create(None, vec![])));
+            Ok(Box::pin(DataBlockStream::create(None, vec![])))
+        } else {
+            let pulling_executor = PipelinePullingExecutor::from_pipelines(build_res, settings)?;
+            ctx.set_executor(Arc::downgrade(&pulling_executor.get_inner()));
+            if let Err(err) = ctx.check_aborting() {
+                pulling_executor.on_finished(Some(err.clone()));
```
zhang2014, line 110:
```suggestion
                pulling_executor.finish(Some(err.clone()));
```

youngsofun:
killed by

https://github.com/datafuselabs/databend/blob/main/src/query/service/src/api/rpc/exchange/statistics_receiver.rs#L93

youngsofun:
@zhang2014 plz review again

Issue #11338, opened 2023-05-06
bug: tonic message too large


ZhiHanZ:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

1.1.29

### What's Wrong?

2001: ConnectionError: source: tonic::status::Status: status: OutOfRange, message: "Error, message length too large: found 4231429 bytes, the limit is: 4194304 bytes", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Sat, 06 May 2023 08:35:34 GMT"} }

### How to Reproduce?

On the cloud platform, with a relatively small table(4G), shown below

CREATE STAGE IF NOT EXISTS r2_stage URL='s3://indexed-xyz/ethereum/decoded/logs/v1.2.0/partition_key=9d/'
CONNECTION = (
  ENDPOINT_URL = 'https://ed5d915e0259fcddb2ab1ce5592040c3.r2.cloudflarestorage.com/'
  REGION = 'auto'
  ACCESS_KEY_ID = '43c31ff797ec2387177cabab6d18f15a'
  SECRET_ACCESS_KEY = 'afb354f05026f2512557922974e9dd2fdb21e5c2f5cbf929b35f0645fb284cf7'
);
CREATE DATABASE IF NOT EXISTS compaction_test;
USE compaction_test;
CREATE TABLE IF NOT EXISTS `contract` (
  `block_time` BIGINT NULL,
  `address` VARCHAR NULL,
  `event_signature` VARCHAR NULL,
  `event_params` ARRAY(STRING NULL) NULL,
  `block_number` BIGINT NULL,
  `block_hash` VARCHAR NULL,
  `log_index` BIGINT NULL,
  `transaction_hash` VARCHAR NULL,
  `transaction_index` BIGINT NULL,
  `data` VARCHAR NULL,
  `topics` VARCHAR NULL,
  `id` VARCHAR NULL
);
COPY INTO compaction_test.contract FROM @r2_stage FILE_FORMAT = (type = PARQUET);
truncate table indexedxyz.contract

- [x] #11344

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #11343, opened 2023-05-06
fix(meta): set max encode/decode size to 16MB for meta-grpc and raft client/server

drmingdrmer:

##### fix(meta): set max encode/decode size to 16MB for meta-grpc and raft client/server

- Part of: #11338

## Related Issues


src/meta/client/src/grpc_client.rs
```diff
@@ -542,9 +544,12 @@ impl MetaGrpcClient {
                     .await;
                     match new_token {
                         Ok(token) => {
-                            return Ok(MetaServiceClient::with_interceptor(c, AuthInterceptor {
-                                token,
-                            }));
+                            let client =
+                                MetaServiceClient::with_interceptor(c, AuthInterceptor { token })
+                                    .max_decoding_message_size(16 * 1024 * 1024)
```
sundy-li, line 549: The default limit is `4G`, do you want to make it smaller?

drmingdrmer: The encoding limit is the max size a client can send and is 4GB, the decoding limit is the max size a server can receive and is 4MB.

src/meta/client/src/grpc_client.rs
```diff
@@ -530,7 +530,9 @@ impl MetaGrpcClient {
             let channel = self.make_channel(Some(addr)).await;
             match channel {
                 Ok(c) => {
-                    let mut client = MetaServiceClient::new(c.clone());
+                    let mut client = MetaServiceClient::new(c.clone())
+                        .max_decoding_message_size(16 * 1024 * 1024)
```
ariesdevil, line 534: This arithmetic expression appears many times and is recommended as a constant.

drmingdrmer: Done. my man

BohuTANG (approved):

PR #11345, opened 2023-05-06
feat(query): Allow insert invalid JSON

b41sh:

## Summary

- JSON parser allow escaped white space, `\t`, `\n`, `\r`, `\x0C`
- Add copy option `DISABLE_VARIANT_CHECK` to allow insert invalid JSON characters. for example:
```sql
COPY INTO test_json FROM 'fs:///tmp/test.csv' FILE_FORMAT = ( field_delimiter = '\t' record_delimiter = '\n' type = 'CSV' ) DISABLE_VARIANT_CHECK = true;
```

Closes #issue


src/query/formats/src/field_decoder/row_based.rs
```diff
@@ -272,10 +272,15 @@ pub trait FieldDecoderRowBased: FieldDecoder {
                 column.commit_row();
             }
             Err(_) => {
-                return Err(ErrorCode::BadBytes(format!(
-                    "Invalid JSON value: {:?}",
-                    String::from_utf8_lossy(&buf)
-                )));
+                if self.common_settings().disable_json_check {
```
sundy-li, line 275: if this setting is enabled, we should commit without JSON parse to improve the performance.

docs/doc/14-sql-commands/40-show/show-settings.md
```diff
@@ -22,6 +22,7 @@ SHOW SETTINGS;
 | name                                  | value        | default      | level   | description                                                                                                                                                                         | type   |
 +---------------------------------------+--------------+--------------+---------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------+
 | collation                             | binary       | binary       | SESSION | Sets the character collation. Available values include "binary" and "utf8".                                                                                                         | String |
+| disable_json_check                    | 0            | 0            | SESSION | Disable check JSON syntax, allow insert invalid JSON.                                                                                                                               | UInt64 |
```
sundy-li, line 25: format options or `settings` ? cc @youngsofun

youngsofun: for safety:    
   copy option (affect a copy OP) >  format options( all copies using this format) > settings (may affect other irrelevant copies in the same session)

src/query/formats/src/field_decoder/row_based.rs
b41sh: valid JSON should convert to jsonb format.

docs/doc/14-sql-commands/40-show/show-settings.md
b41sh: using format options?

youngsofun: using `copy option` more reasonable to me

e.g.

CSV row `abc` can be loaded to  table1(string) or table2(variant)
i.e. disable_json_check is meaningful only for table2, 
in other words, `disable_json_check` is not a property of the CSV format

youngsofun: using `copy option` more reasonable to me

e.g.

CSV row `abc` can be loaded to  table1(string) or table2(variant)
i.e. disable_json_check is meaningful only for table2, 
in other words, `disable_json_check` is not a property of the CSV format

youngsofun:
since in copy context, we have a file format called `Json`,  on the other side the Type of field is called `variant`, not `json`, 
 maybe it is better to rename `disable_json_check` to `disable_variant_check`?

tests/data/invalid_json.csv
```diff
@@ -0,0 +1,2 @@
+1	invalidjson
```
youngsofun, line 1: why `invalidjson` is invalid?

youngsofun (approved):

drmingdrmer (approved):

sundy-li (approved):

PR #11367, opened 2023-05-09
feat: validates the value of max_storage_io_requests being

### LINKED ISSUE #11361: Feature:  validate the value of `max_storage_io_requests` being set
**Summary**

while setting the value of `max_storage_io_requests`, make sure that the value is larger than 0 (otherwise, return an Err)

https://github.com/datafuselabs/databend/blob/aa97d2c5b57f1585abc8192e5fb76f9fc4157903/src/query/settings/src/settings_getter_setter.rs#L126C1-L128

Mehrbod2002:

## Summary

Validate max_storage_io_requests which checks it's greater than 0 or throw an error .

Closes #11361

### DIFF
--- a/src/query/settings/src/settings_getter_setter.rs
+++ b/src/query/settings/src/settings_getter_setter.rs
@@ -125,12 +125,9 @@ impl Settings {
 
     pub fn set_max_storage_io_requests(&self, val: u64) -> Result<()> {
         if val > 0 {
-            self.try_set_u64("max_storage_io_requests", val);
-            Ok(())
+            self.try_set_u64("max_storage_io_requests", val)
         } else {
-            Err(ErrorCode::BadArguments(
-                "Value must be greater than 0".into(),
-            ))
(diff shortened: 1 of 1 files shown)


BohuTANG (approved):
👍

src/query/settings/src/settings_getter_setter.rs
```diff
@@ -124,7 +124,14 @@ impl Settings {
...
     pub fn set_max_storage_io_requests(&self, val: u64) -> Result<()> {
-        self.try_set_u64("max_storage_io_requests", val)
+        if val > 0 {
+            self.try_set_u64("max_storage_io_requests", val);
+            Ok(())
+        } else {
+            Err(ErrorCode::BadArguments(
+                "Value must be greater than 0".into(),
+            ))
+        }
```
BohuTANG, line 134:
```suggestion
        if val > 0 {
            self.try_set_u64("max_storage_io_requests", val)
        } else {
            Err(ErrorCode::BadArguments("Value must be greater than 0"))
        }
```

PR #11380, opened 2023-05-09
chore(query): use prune cache only if enable result query cache

zhang2014:

## Summary

chore(query): use prune cache only if enable result query cache

Closes #issue

### DIFF
--- a/src/query/storages/fuse/src/operations/read_partitions.rs
+++ b/src/query/storages/fuse/src/operations/read_partitions.rs
@@ -122,17 +122,17 @@ impl FuseTable {
             segments_location.len()
         );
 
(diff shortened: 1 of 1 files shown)


src/query/storages/fuse/src/operations/read_partitions.rs
```diff
@@ -122,17 +122,17 @@ impl FuseTable {
             segments_location.len()
         );
 
-        type CacheItem = (PartStatistics, Partitions);
-        let cache_key = format!(
-            "{:x}",
-            Sha256::digest(format!("{:?}_{:?}", segments_location, push_downs))
-        );
```
RinChanNOWWW, line 129: This is not for query result cache.

PR #11386, opened 2023-05-09
fix(storage): update/delete oom

zhyass:

## Summary

Optimize the processes of delete/update/replace/compact/analyze.

1. Read and process segments in chunks during update/delete/replace/analyze to avoid oom.

2. Avoid reading all segments during delete/update/replace/compact/compact segment, only read the affected segments. Calculate the new snapshot summary based on the old snapshot summay.

```
new_snapshot.summary = base_snapshot.summary - affected_segments.summary + new_segments.summary
```

3. Recalculate the column statistics in snapshot during analyze table.

***Note***: After delete/update, the MinMax value of `snapshot.summary.col_stats` may be enlarged.

*Of course we can get an exact minmax, but this comes at the cost of reading all the segments during update/delete/replace.*
*On the one hand the value in snapshot is currently unused, on the other hand it will only be enlarged and not missed.* 

And you can perform analyze statement to correct the MinMax value of `snapshot.summary.col_stats`.

For example:
```
mysql> create table t(a int, b int);

mysql> insert into t values(1,1);
Query OK, 1 row affected (0.06 sec)

mysql> insert into t values(2,2);
Query OK, 1 row affected (0.06 sec)

mysql> update t set b=3 where a=1;
Query OK, 1 row affected (0.08 sec)

//After update, the column `b`'s MinMax in latest snapshot will be [1,3].

mysql> analyze table t;
Query OK, 0 rows affected (0.12 sec)

// After analyze, the column `b`'s MinMax in latest snapshot will be [2,3].
```

Closes #issue


src/query/storages/fuse/src/operations/mutation/mutation_aggregator.rs
```diff
@@ -0,0 +1,289 @@
...
+use crate::io::TableMetaLocationGenerator;
+use crate::operations::mutation::AbortOperation;
+use crate::operations::mutation::Mutation;
+use crate::operations::mutation::MutationSinkMeta;
+use crate::operations::mutation::MutationTransformMeta;
+use crate::statistics::reducers::deduct_statistics_mut;
+use crate::statistics::reducers::merge_statistics_mut;
+use crate::statistics::reducers::reduce_block_metas;
+
+type MutationMap = HashMap<usize, (Vec<(usize, Arc<BlockMeta>)>, Vec<usize>)>;
```
lichuang, line 47: can define `(Vec<(usize, Arc<BlockMeta>)>, Vec<usize>)` as a type to be more clearly?

dantengsky:
LGTM.

for the cocern that @lichuang  mentioned:

@zhyassa PR opened at  

https://github.com/zhyass/databend/pull/8/files

zhyass:
> LGTM.
> 
> for the concern that @lichuang mentioned:
> 
> @zhyass , I opened a PR at
> 
> https://github.com/zhyass/databend/pull/8/files
> 
> hope it helps. (feel free to reject the PR, if it is not what u like)

@dantengsky It just so happened that I was also changing this codes. I directly reused the merge into codes here.

dantengsky (approved):

PR #11391, opened 2023-05-10
feat(query): Support Computed columns

### LINKED ISSUE #11390: Feature: Support computed columns
**Summary**

computed column generates data from other columns by defining a scalar expression.
- A stored computed column computes the data when the row is inserted and stores the result value in the file like other columns.
```sql
column_name <type> AS (<expr>) STORED
```
- A virtual computed column does not store the result value, it is computed at query time.
```sql
column_name <type> AS (<expr>) VIRTUAL
```

b41sh:

## Summary

Support Computed columns

Closes #11390


src/meta/proto-conv/tests/it/v036_table_meta.rs
```diff
@@ -0,0 +1,171 @@
...
+                ce::TableField::new("bool", ce::TableDataType::Boolean),
+                ce::TableField::new("int8", ce::TableDataType::Number(NumberDataType::Int8)),
+                ce::TableField::new("int16", ce::TableDataType::Number(NumberDataType::Int16)),
+                ce::TableField::new("int32", ce::TableDataType::Number(NumberDataType::Int32)),
+                ce::TableField::new("int64", ce::TableDataType::Number(NumberDataType::Int64)),
+                ce::TableField::new("uint8", ce::TableDataType::Number(NumberDataType::UInt8)),
+                ce::TableField::new("uint16", ce::TableDataType::Number(NumberDataType::UInt16)),
+                ce::TableField::new("uint32", ce::TableDataType::Number(NumberDataType::UInt32)),
+                ce::TableField::new("uint64", ce::TableDataType::Number(NumberDataType::UInt64)),
+                ce::TableField::new(
```
drmingdrmer, line 114: Fields that are unrelated to this PR can be removed.

drmingdrmer (approved):

src/query/expression/src/schema.rs
```diff
@@ -237,7 +245,12 @@ impl DataSchema {
     }
 
     pub fn to_arrow(&self) -> ArrowSchema {
-        let fields = self.fields().iter().map(|f| f.into()).collect::<Vec<_>>();
+        let fields = self
```
sundy-li, line 248: So some fields may be missed when converting to arrow.

The schema convert from arrow may be incomplete.

b41sh: The columns stored in parquet do not include virtual computed columns. A virtual computed column is also not required when moving from a parquet back to a data block. Virtual computed columns are converted to an expression when binder like other expressions.

tests/suites/1_stateful/00_copy/00_0009_copy_csv_with_computed_column.sh
```diff
@@ -0,0 +1,36 @@
...
+. "$CURDIR"/../../../shell_env.sh
+
+# Should be <root>/tests/data/
+DATADIR=$(realpath $CURDIR/../../../data/)
+
+echo "drop table if exists sample" | $MYSQL_CLIENT_CONNECT
+
+## Create table
+cat <<EOF | $MYSQL_CLIENT_CONNECT
+CREATE TABLE sample
```
youngsofun, line 13: does it work if virtual columns are not the last ones in schema? 

```
(
    Id     INT,
    City2  VARCHAR AS (reverse(City)) STORED,
    Score2 INT64 AS (Score + 2) VIRTUAL,
    City   VARCHAR,
    Score  INT,
);
```

b41sh: yes

drmingdrmer (approved):

sundy-li (approved):

Issue #11355, opened 2023-05-08
Feature:   Optimize read table meta data


wubx:
**Summary**

select from system.tables, system.columns ( information_schema.tables, information_schema.columns], It slows down as the number of tables increases. 
For example:
```
mysql> select name , data_type as type from system.columns where database='default' and table='hackernews';
+-------------+------------------+
| name        | type             |
+-------------+------------------+
| id          | BIGINT UNSIGNED  |
| deleted     | TINYINT UNSIGNED |
| type        | VARCHAR          |
| author      | VARCHAR          |
| timestamp   | TIMESTAMP        |
| comment     | VARCHAR          |
| dead        | TINYINT UNSIGNED |
| parent      | BIGINT           |
| poll        | BIGINT           |
| children    | ARRAY(INT64)     |
| url         | VARCHAR          |
| score       | INT              |
| title       | VARCHAR          |
| parts       | ARRAY(INT64)     |
| descendants | INT              |
+-------------+------------------+
15 rows in set (0.01 sec)
Read 780 rows, 87.61 KiB in 0.011 sec., 69.05 thousand rows/sec., 7.57 MiB/sec.

mysql> create table hits03 like hits;
Query OK, 0 rows affected (0.02 sec)

mysql> select name , data_type as type from system.columns where database='default' and table='hackernews';
+-------------+------------------+
| name        | type             |
+-------------+------------------+
| id          | BIGINT UNSIGNED  |
| deleted     | TINYINT UNSIGNED |
| type        | VARCHAR          |
| author      | VARCHAR          |
| timestamp   | TIMESTAMP        |
| comment     | VARCHAR          |
| dead        | TINYINT UNSIGNED |
| parent      | BIGINT           |
| poll        | BIGINT           |
| children    | ARRAY(INT64)     |
| url         | VARCHAR          |
| score       | INT              |
| title       | VARCHAR          |
| parts       | ARRAY(INT64)     |
| descendants | INT              |
+-------------+------------------+
15 rows in set (0.01 sec)
Read 885 rows, 98.91 KiB in 0.011 sec., 79.34 thousand rows/sec., 8.66 MiB/sec.
```
Same SQL add hits03 Read rows from 780 to 885.

----
In the fact of the matter is:
```
select count(distinct(table_name)) from information_schema.columns;
+-----------+
| count()    |
+-----------+
| 1108     |
+-----------+
1 row in set (2.14 sec)
[...]

PR #11398, opened 2023-05-10
feat(query): Optimize read system.columns data

TCeason:

## Summary

If query `system.columns` with push down filter Expr FunctionCall eq

and the eq arg has column name 'table' like this:

```
where table='tb1' and table='tb2' and table='tb3' and table = 'tb4'
```

will only get these tables.

BTW, not use field database and Expr or as filter. Because there maybe have some bad case:

```
where database='db1' or table='db2_tb1'
where database='db1' and table='db1_tb1' or table='db2_tb1'
```

Closes #https://github.com/datafuselabs/databend/issues/11355

### DIFF
--- a/src/query/storages/system/src/columns_table.rs
+++ b/src/query/storages/system/src/columns_table.rs
@@ -149,12 +149,18 @@ impl ColumnsTable {
         let catalog = ctx.get_catalog(CATALOG_DEFAULT)?;
 
         let mut tables = Vec::new();
-
+        let mut databases = Vec::new();
         if let Some(push_downs) = push_downs {
             if let Some(filter) = push_downs.filter {
                 let expr = filter.as_expr(&BUILTIN_FUNCTIONS);
-                find_eq_table(&expr, &mut |col_name, scalar| {
-                    if col_name == "table" {
+                find_eq_db_table(&expr, &mut |col_name, scalar| {
+                    if col_name == "database" {
+                        if let Scalar::String(s) = scalar {
+                            if let Ok(database) = String::from_utf8(s.clone()) {
+                                databases.push(database);
+                            }
+                        }
+                    } else if col_name == "table" {
                         if let Scalar::String(s) = scalar {
                             if let Ok(table) = String::from_utf8(s.clone()) {
                                 tables.push(table);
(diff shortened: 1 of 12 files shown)


src/query/storages/system/src/table.rs
```diff
@@ -233,12 +237,21 @@ impl<TTable: 'static + AsyncSystemTable> Table for AsyncOneBlockSystemTable<TTab
     fn read_data(
         &self,
         ctx: Arc<dyn TableContext>,
-        _: &DataSourcePlan,
+        plan: &DataSourcePlan,
         pipeline: &mut Pipeline,
     ) -> Result<()> {
         let inner_table = self.inner_table.clone();
+        let push_downs = plan.push_downs.clone();
+        println!("push_downs is {:?}", push_downs.clone());
```
b41sh, line 245: unused `println`

src/query/storages/system/src/columns_table.rs
```diff
@@ -168,3 +219,28 @@ impl ColumnsTable {
...
+
+pub fn find_eq_table(expr: &Expr<String>, visitor: &mut impl FnMut(&str, &Scalar)) {
+    match expr {
+        Expr::Constant { .. } => {}
+        Expr::ColumnRef { .. } => {}
+        Expr::Cast { expr, .. } => {
+            find_eq_table(expr, visitor);
+        }
+        Expr::FunctionCall { function, args, .. } => {
+            if function.signature.name == "eq" {
```
b41sh, line 231: Only check equal function may produce some unintended results, for example
```sql
where table='tb1' or table !='tb2';
```

TCeason: Yes, this maybe cause some other bad case like :

```sql
where table='tb1' and table in ('tb2', 'tb3', 'tb4', 'tb5')
```

So I think filter only contains `eq` `and` `or` do this optimize.

src/query/storages/system/src/columns_table.rs
```diff
@@ -134,17 +143,74 @@ impl ColumnsTable {
...
+                let expr = filter.as_expr(&BUILTIN_FUNCTIONS);
+                find_eq_db_table(&expr, &mut |col_name, scalar| {
+                    if col_name == "database" {
+                        if let Scalar::String(s) = scalar {
+                            if let Ok(database) = String::from_utf8(s.clone()) {
+                                databases.push(database);
+                            }
+                        }
+                    }
+                    if col_name == "table" {
```
b41sh, line 164: `else if`

src/query/storages/system/src/columns_table.rs
```diff
@@ -134,17 +143,74 @@ impl ColumnsTable {
...
+            } else {
+                let mut res = Vec::new();
+                for table in &tables {
+                    match catalog.get_table(tenant.as_str(), &database, table).await {
+                        Ok(table) => {
+                            res.push(table);
+                        }
+                        Err(_) => {
+                            debug!(
+                                "Can not get table {:?}.{:?}.{:?}",
```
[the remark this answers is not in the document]
TCeason: That's ok, if table not exists we need ignore this err and generate the whole data. The return maybe empty.

b41sh (approved):

Issue #11359, opened 2023-05-08
http handler release session/context after query executer stop


youngsofun:
https://github.com/datafuselabs/databend/pull/8894
[fix(http handler): avoid dropping runtime when task on it not finished. by youngsofun · Pull Request #8894 · datafuselabs/databend (github.com)](https://github.com/datafuselabs/databend/pull/8894)

this pr added a reference in the page manager to avoid the Session from being dropped too early.

it takes by default 60s to release the HttpQuery.

not a problem if the final API is properly called.

PR #10501, opened 2023-03-11
fix: thread leaking

dantengsky:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

to avoid deadlock during destructing the ad-hoc runtime `query-ctx`
- introduce `GlobalQueryRuntime`
- replace the ad-hoc runtime `query-ctx` with global runtime `GlobalQueyRuntime` in `HttpQuery::try_create` 
- replace the ad-hoc runtime `query-ctx` with global runtime `GlobalQueyRuntime` in `StatisticsSender::start` 

p.s. This is more like a temp workaround, we should refactor the execution model (the thread pools, ad-hoc runtimes, and communications between sync/async components).

  

Closes #10374

PR #11403, opened 2023-05-11
chore: http handler release context/session earlier.

youngsofun:

## Summary

there is an extra instance of ctx to avoid dropping runtime when task on it is running,
this is solved better in https://github.com/datafuselabs/databend/pull/10501, so not need it any more.

Closes https://github.com/datafuselabs/databend/issues/11359


youngsofun:
cc @wubx

BohuTANG (approved):

PR #11409, opened 2023-05-11
fix(http): Don't print whole SQL for insert

Xuanwo:

## Summary

fix(http): Don't print whole SQL for insert

### DIFF
--- a/src/query/service/src/servers/http/v1/query/http_query.rs
+++ b/src/query/service/src/servers/http/v1/query/http_query.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::collections::BTreeMap;
+use std::fmt::Debug;
 use std::sync::Arc;
 use std::time::Duration;
 use std::time::Instant;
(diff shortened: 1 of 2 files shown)


flaneur2020 (approved):

src/query/service/src/servers/http/v1/query/http_query.rs
```diff
@@ -64,6 +66,19 @@ pub struct HttpQueryRequest {
     pub stage_attachment: Option<StageAttachmentConf>,
 }
 
+impl Debug for HttpQueryRequest {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("HttpQueryRequest")
+            .field("session_id", &self.session_id)
+            .field("session", &self.session)
+            .field("sql", &short_sql(self.sql.clone()))
```
flaneur2020, line 74: could `short_sql` sanitize the sql's sensitive content like access secrets?

Xuanwo: no, the sql is not parsed and we don't what content is sensitive.

PR #11414, opened 2023-05-11
feat(query): add create/drop index support

ariesdevil:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

add create/drop index support

Closes #issue

PR #11418, opened 2023-05-12
feat: auto query rewriting for aggregation index.

RinChanNOWWW:

## Summary

Pre task: #11414

Use RBO framework to rewrite the query plan.

Next task: introduce index scan to pipeline.

## Limits

Only support query like:

```sql
SELECT ...
FROM ...
[WHERE ...]
[GROUP BY ...]
```

**NOT** support `order by` and `having clause`.

## Match Rules

The first matched index will be picked to use.

### Selection

1. The selection of index should contain the selection of query.

### Filter 

1. Predicates of index should contain predicates of query. For example, `a > 0` contains `a > 1`. To simplify the logic, complex predicates should be totally the same (such as `a + b = c * d`).
2. If the predicate of query cannot be eliminated by index, its columns should be contained by the selection of index.

### Aggregation

1. Group items shoud be the same.


ariesdevil:
#11414 has been merged, thanks for @ZhiHanZ 's help:)
The ee ci env should be https://github.com/datafuselabs/databend/blob/2cff53b811daefecc857c41cf218b8f50c3f8167/.github/workflows/dev-linux.yml#L239

sundy-li (approved):

PR #11421, opened 2023-05-12
feat(planner): remove trivial cast

### LINKED ISSUE #11411: Support translating expression with trivial cast to z3 ast
We only support translating simple expressions to z3 ast, excluding type casts. But it's possible to remove some trivial type casts in some situations to convert more expressions to z3 ast.

A typical case is to cast a small integer to a big integer, e.g. `CAST(1 AS Int64)`.

Dousir9:

## Summary

Summary about this PR
If there is a cast expression in the filter, it will prevent the optimizer from translating the expression into z3 ast, which will affect the conversion of outer join to inner join, so we can eliminate some trivial type cast to convert more outer join into inner join.

Closes #11411


src/query/sql/src/planner/optimizer/property/constraint.rs
```diff
@@ -146,22 +151,25 @@ impl<'ctx> ScalarVisitor for VariableCollector<'ctx> {
 
 /// Transform a logical expression into a z3 ast.
 /// Will return a Nullable Boolean ast.
-fn transform_logical_expr<'ctx>(ctx: &'ctx Context, scalar: &ScalarExpr) -> Option<Dynamic<'ctx>> {
+fn transform_logical_expr<'ctx>(
+    ctx: &'ctx Context,
+    scalar: &mut ScalarExpr,
```
leiysky, line 156: Maybe there is no need to mutate the original `ScalarExpr`. Just returning a rewritten result is enough.

leiysky (approved):
Rest LGTM

src/query/sql/src/planner/optimizer/property/constraint.rs
Dousir9: In `is_null_reject`, we will call `as_z3_ast` and then call `transform_logical_expr`, if we return rewritten result, we need to create an unnecessary Vec.

PR #11429, opened 2023-05-12
chore(planner): postpone type unification for window frame

andylokandy:

## Summary

Summary about this PR

Closes #issue

### DIFF
--- a/src/query/sql/src/executor/physical_plan_builder.rs
+++ b/src/query/sql/src/executor/physical_plan_builder.rs
@@ -32,6 +32,7 @@ use common_expression::types::DataType;
 use common_expression::ConstantFolder;
 use common_expression::DataBlock;
 use common_expression::DataField;
+use common_expression::DataSchema;
 use common_expression::DataSchemaRefExt;
 use common_expression::FunctionContext;
 use common_expression::RemoteExpr;
@@ -46,6 +47,7 @@ use super::AggregateFinal;
 use super::AggregateFunctionDesc;
 use super::AggregateFunctionSignature;
 use super::AggregatePartial;
+use super::EvalScalar;
 use super::Exchange as PhysicalExchange;
 use super::Filter;
 use super::HashJoin;
@@ -59,7 +61,6 @@ use crate::binder::wrap_cast;
 use crate::binder::INTERNAL_COLUMN_FACTORY;
 use crate::executor::explain::PlanStatsInfo;
 use crate::executor::table_read_plan::ToReadDataSourcePlan;
-use crate::executor::EvalScalar;
 use crate::executor::FragmentKind;
 use crate::executor::PhysicalPlan;
 use crate::executor::RuntimeFilterSource;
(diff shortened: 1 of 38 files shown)


RinChanNOWWW:
For join, cc @xudong963

src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -855,8 +862,99 @@ impl PhysicalPlanBuilder {
...
+                        | WindowFuncFrameBound::Following(scalar) => scalar.as_mut(),
+                        _ => None,
+                    };
+                    let mut end = match &mut w.frame.end_bound {
+                        WindowFuncFrameBound::Preceding(scalar)
+                        | WindowFuncFrameBound::Following(scalar) => scalar.as_mut(),
+                        _ => None,
+                    };
+
+                    let mut common_ty = order_by
```
RinChanNOWWW, line 885: Do we need to remove nullable first and wrap nullable later?

src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -855,8 +862,99 @@ impl PhysicalPlanBuilder {
...
+                }
+                let input = if !scalar_items.is_empty() {
+                    let eval_scalar = crate::planner::plans::EvalScalar {
+                        items: scalar_items,
+                    };
+                    SExpr::create_unary(eval_scalar.into(), s_expr.child(0)?.clone())
+                } else {
+                    s_expr.child(0)?.clone()
+                };
+                let input = self.build(&input).await?;
```
RinChanNOWWW, line 956: I think we can add `EvalScalar` plan manually instead of `build` again. (Avoid traversing the plan tree twice)

image

This input plan can be the child of the new `EvalScalar`, and we can reuse the codes for building `EvalScalar`.

src/query/sql/src/executor/physical_plan_builder.rs
andylokandy: Casting to nullable has no effect on Scalar value

src/query/sql/src/executor/physical_plan_builder.rs
andylokandy: The duplicate build is to avoid repeating the code of building physical EvalScalar

RinChanNOWWW: But it will traverse the plan again. I think we can avoid this.

src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -856,7 +842,102 @@ impl PhysicalPlanBuilder {
...
+                        | WindowFuncFrameBound::Following(scalar) => scalar.as_mut(),
+                        _ => None,
+                    }
+                    .unwrap();
+                    let mut end = match &mut w.frame.end_bound {
+                        WindowFuncFrameBound::Preceding(scalar)
+                        | WindowFuncFrameBound::Following(scalar) => scalar.as_mut(),
+                        _ => None,
+                    }
+                    .unwrap();
```
RinChanNOWWW, line 864: Cannot `unwrap` directly. The bound can be `None`.

RinChanNOWWW (approved):
The window function part is LGTM.

andylokandy:
The join part has been removed

PR #11442, opened 2023-05-14
feat(query): add func date_format and str_to_date and str_to_timestamp

### LINKED ISSUE #11125: Feature: hope to support like mysql date_format and str_to_date fuction
In applications, we often need to display the time format as yyyy-MM-dd HH: mm: ss or a custom time format,but databend only support to change  yyyyMM yyyyMMDD  yyyymmddhhmmss format.

TCeason:

## Summary

support function:

date_format(Timestamp, String) :: String NULL
date_format(Timestamp NULL, String NULL) :: String NULL

str_to_date(String, String) :: Date NULL
str_to_date(String NULL, String NULL) :: Date NULL
str_to_timestamp(String, String) :: Timestamp NULL
str_to_timestamp(String NULL, String NULL) :: Timestamp NULL

### Note

* If use str_to_timestamp , the string must contains timezone. Because we need to unique date and time. If not contains timezone will return null.

Closes #11125

### DIFF
--- a/src/query/functions/src/scalars/datetime.rs
+++ b/src/query/functions/src/scalars/datetime.rs
@@ -14,6 +14,7 @@
 
 use std::io::Write;
 
+use chrono::prelude::*;
 use chrono::Datelike;
 use chrono::Utc;
 use common_arrow::arrow::temporal_conversions::EPOCH_DAYS_FROM_CE;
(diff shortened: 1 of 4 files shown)


youngsofun (approved):

BohuTANG:
Which database's style we follow for the function name `date_format ` and `str_to_*`?

tests/sqllogictests/suites/query/02_function/02_0012_function_datetimes
```diff
@@ -893,3 +893,38 @@ select * from t order by b
 statement ok
 drop table t
 
+query T
+select date_format('2022-02-02', '精彩的%Y年，美丽的%mmonth,激动の%dd');
+----
+精彩的2022年，美丽的02month,激动の02d
+
+query T
+select str_to_date('精彩的2022年，美丽的02month,激动の02d', '精彩的%Y年，美丽的%mmonth,激动の%dd');
```
BohuTANG, line 902: The name str_to_date is confusing here, string to date is a parse a string to a date data type.

TCeason: In mysql doc: https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_str-to-date

BohuTANG:
These functions requires additional documentation. Thanks! cc @soyeric128

We already have an powerful date format function in place. cc @wubx

src/query/functions/src/scalars/datetime.rs
```diff
@@ -148,6 +149,59 @@ fn register_string_to_timestamp(registry: &mut FunctionRegistry) {
...
+    registry.register_combine_nullable_2_arg::<StringType, StringType, DateType, _, _>(
+        "str_to_date",
+        |_, _| FunctionDomain::MayThrow,
+        vectorize_with_builder_2_arg::<StringType, StringType, NullableType<DateType>>(
+            |date, format, output, _| {
+                if format.is_empty() {
+                    output.push_null();
+                } else {
+                    match (std::str::from_utf8(date), std::str::from_utf8(format)) {
+                        (Ok(date), Ok(format)) => {
```
sundy-li, line 190: why not throw the error?

TCeason: Refer to MySQL behavior here. Return null directly if there is any invalid

sundy-li: I think we should not follow MySQL's behavior, we should throw if any error happens.

TCeason: > I think we should not follow MySQL's behavior, we should throw if any error happens.

Modify in this pr: https://github.com/datafuselabs/databend/pull/11455

PR #11450, opened 2023-05-15
fix: use operator.scan instead of list.

youngsofun:

## Summary

Summary about this PR

Closes #issue


BohuTANG (approved):

src/common/storage/src/stage.rs
```diff
@@ -285,7 +285,7 @@ fn blocking_list_files_with_pattern(
 
     // path is a dir
     let mut files = Vec::new();
-    let list = operator.list(path)?;
+    let list = operator.scan(path)?;
```
BohuTANG, line 288: BTW, What's the difference between `list` and `scan`?

youngsofun: list in fs is `list_dir` not recursive

sundy-li (approved):

PR #11489, opened 2023-05-17
feat: push down aggregating index info to physcial scan.

RinChanNOWWW:

## Summary

- Push down aggregating index info to scan plan.
- Add explain tests.
- Support single table rewrite in multi-table query.

### Modification of Aggregating Index

1. Use full table name (`{catalog}.{database}.{table}`) to index aggregating indexes instead of using table index. Because we will allocate different table indexes when binding tables even if the tables are the same table.
2. Apply `RuleFoldConstant` to index plan. Because the original query plan will apply this rule before rewritting. If we don't apply this rule to index plan, the two plans may not match as the contants will mistmatch.
3. Rewrite the indexes of base columns in index plan. Tables and columns in query and index will be bound different indexes (the same as 1.). For convenience, we can rewrite the column indexes in index plan by the indexes in query plan (replace the indexes of same name columns). Using index to compare two plans is more simple and efficient.

### Next Task

- Enhance scan source processor to support index scan.


src/query/sql/src/executor/format.rs
```diff
@@ -204,6 +210,36 @@ fn table_scan_to_format_tree(
...
+
+        let agg_sel = agg_index
+            .selection
+            .iter()
+            .map(|expr| expr.as_expr(&BUILTIN_FUNCTIONS).sql_display())
+            .join(", ");
+        let agg_filter = agg_index
+            .filter
+            .as_ref()
+            .map(|f| f.as_expr(&BUILTIN_FUNCTIONS).sql_display());
```
ariesdevil, line 235: filters also need join ", " here?

RinChanNOWWW: They are already combined into one expression by "and".

ariesdevil (approved):

PR #11492, opened 2023-05-17
fix(query): fix use db err because of not normalize ident

### LINKED ISSUE #11491: bug: use db in binder not normalize ident db
**Summary**

https://github.com/datafuselabs/databend/blob/main/src/query/sql/src/planner/binder/binder.rs#L222

TCeason:

## Summary

use db err need normalize ident db.

Closes #11491

### DIFF
--- a/src/query/service/src/servers/mysql/mysql_interactive_worker.rs
+++ b/src/query/service/src/servers/mysql/mysql_interactive_worker.rs
@@ -43,6 +43,7 @@ use rand::RngCore;
 use tracing::error;
 use tracing::info;
 use tracing::Instrument;
+use tracing::log::warn;
 
 use crate::interpreters::Interpreter;
 use crate::interpreters::InterpreterFactory;
@@ -415,16 +416,15 @@ impl<W: AsyncWrite + Send + Unpin> InteractiveWorkerBase<W> {
         if database_name.is_empty() {
             return Ok(());
         }
-        let init_query = format!("USE {};", database_name);
+        let init_query = format!("USE `{}`;", database_name);
 
         let do_query = self.do_query(&init_query).await;
         match do_query {
             Ok((_, _)) => Ok(()),
             Err(error_code) => {
-                if error_code.code() == ErrorCode::SyntaxException("").code()
-                    || error_code.code() == ErrorCode::UNKNOWN_DATABASE
-                {
-                    let init_query = format!("USE `{}`;", database_name);
+                if error_code.code() == ErrorCode::UNKNOWN_DATABASE {
+                    warn!("Unknown Database '`{}`, try to trim quota.'", database_name);
+                    let init_query = format!("USE {};", database_name);
                     self.do_query(&init_query).await?;
                     return Ok(());
                 }
--- a/src/query/service/src/servers/mysql/mysql_interactive_worker.rs
+++ b/src/query/service/src/servers/mysql/mysql_interactive_worker.rs
@@ -420,7 +420,16 @@ impl<W: AsyncWrite + Send + Unpin> InteractiveWorkerBase<W> {
         let do_query = self.do_query(&init_query).await;
         match do_query {
             Ok((_, _)) => Ok(()),
-            Err(error_code) => Err(error_code),
+            Err(error_code) => {
+                if error_code.code() == ErrorCode::SyntaxException("").code()
+                    || error_code.code() == ErrorCode::UNKNOWN_DATABASE
+                {
+                    let init_query = format!("USE `{}`;", database_name);
+                    self.do_query(&init_query).await?;
+                    return Ok(());
+                }
+                Err(error_code)
+            }
         }
     }
 }
--- a/src/query/service/src/servers/mysql/mysql_interactive_worker.rs
+++ b/src/query/service/src/servers/mysql/mysql_interactive_worker.rs
@@ -415,7 +415,7 @@ impl<W: AsyncWrite + Send + Unpin> InteractiveWorkerBase<W> {
         if database_name.is_empty() {
             return Ok(());
         }
-        let init_query = format!("USE `{}`;", database_name);
+        let init_query = format!("USE {};", database_name);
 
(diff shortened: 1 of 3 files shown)


b41sh (approved):

TCeason:
Now we have to settings to determines whether Databend treats identifiers as case-sensitive.

1. unquoted_ident_case_sensitive, with default value 0.
2. quoted_ident_case_sensitive, with default value 1.

If user use mysql client execute use db, may generate two format :+1: 

```
use `X`
use X
```

With different value of setting, the result will be different.

So I think we can set the same Default value.

If there are special requirements, then by the customer's own configuration

sundy-li:
The default behavior is compatible with PG, I suggest @leiysky  review this pr.

src/query/settings/src/settings_default.rs
```diff
@@ -131,7 +131,7 @@ impl DefaultSettings {
                     display_in_show_settings: true,
                 }),
                 ("quoted_ident_case_sensitive", DefaultSettingValue {
-                    value: UserSettingValue::UInt64(1),
+                    value: UserSettingValue::UInt64(0),
```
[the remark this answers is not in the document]
leiysky: In most of the databases, the quoted identifier is used to represent the special column name that is stored and resolved consistently with the user-entered value. For example, both [Snowflake](https://docs.snowflake.com/en/sql-reference/identifiers-syntax#label-identifier-casing) and [PostgreSQL](https://www.postgresql.org/docs/current/sql-syntax-lexical.html) follow this.

So it's natural to keep the original case of quoted identifiers, i.e. make quoted identifiers case-sensitive by default. I think we should not change the default value of this.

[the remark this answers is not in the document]
leiysky: @BohuTANG PTAL

src/query/service/src/servers/mysql/mysql_interactive_worker.rs
```diff
@@ -420,7 +421,15 @@ impl<W: AsyncWrite + Send + Unpin> InteractiveWorkerBase<W> {
         let do_query = self.do_query(&init_query).await;
         match do_query {
             Ok((_, _)) => Ok(()),
-            Err(error_code) => Err(error_code),
+            Err(error_code) => {
+                if error_code.code() == ErrorCode::UNKNOWN_DATABASE {
```
TCeason, line 425: This can resolve :

```
create database Tt;
use Tt; -- error: database Tt not exists.
```

But it also has bad case:

```
MySQL [(none)]> create database Rustlang;
Query OK, 0 rows affected (0.026 sec)

MySQL [(none)]> create database `Rustlang`;
Query OK, 0 rows affected (0.036 sec)

MySQL [(none)]> create table Rustlang.t_low(id int);
Query OK, 0 rows affected (0.044 sec)

MySQL [(none)]> create table `Rustlang`.t_up(id int);
Query OK, 0 rows affected (0.034 sec)

MySQL [(none)]> use Rustlang;
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
MySQL [Rustlang]> show tables; --- use db without quota, in mind the result of show tables should be: t_low
+--------------------+
| tables_in_rustlang |
+--------------------+
| t_up               |
+--------------------+
1 row in set (0.038 sec)

```

TCeason: > This can resolve :
> 
> ```
> create database Tt;
> use Tt; -- error: database Tt not exists.
> ```
> 
> But it also has bad case:
> 
> ```
> MySQL [(none)]> create database Rustlang;
> Query OK, 0 rows affected (0.026 sec)
> 
> MySQL [(none)]> create database `Rustlang`;
> Query OK, 0 rows affected (0.036 sec)
> 
> MySQL [(none)]> create table Rustlang.t_low(id int);
> Query OK, 0 rows affected (0.044 sec)
> 
> MySQL [(none)]> create table `Rustlang`.t_up(id int);
> Query OK, 0 rows affected (0.034 sec)
> 
> MySQL [(none)]> use Rustlang;
> Reading table information for completion of table and column names
> You can turn off this feature to get a quicker startup with -A
> 
> Database changed
> MySQL [Rustlang]> show tables; --- use db without quota, in mind the result of show tables should be: t_low
> +--------------------+
> | tables_in_rustlang |
> +--------------------+
> | t_up               |
> +--------------------+
> 1 row in set (0.038 sec)
> ```

BTW:

This may be confusing for users:

```
MySQL [Rustlang]> use Rustlang;
Database changed
MySQL [Rustlang]> show tables;
+--------------------+
| tables_in_rustlang |
+--------------------+
| t_up               |
+--------------------+
1 row in set (0.036 sec)

MySQL [Rustlang]> show tables from Rustlang;
+--------------------+
| tables_in_rustlang |
+--------------------+
| t_low              |
+--------------------+
1 row in set (0.043 sec)

```

TCeason: cc @leiysky @BohuTANG PTAL

leiysky (approved):

src/query/service/src/servers/mysql/mysql_interactive_worker.rs
leiysky: I think in the main branch this is a bug since the `database` identifier should be normalized. This fix will resolve the issue.

sundy-li (approved):

tests/sqllogictests/suites/base/07_use/07_0000_use_database
```diff
@@ -25,5 +25,53 @@ statement ok
...
+
+onlyif mysql
+statement ok
+set quoted_ident_case_sensitive=0;
+
+onlyif mysql
+statement ok
+use Rustlang;
+
+onlyif mysql
```
TCeason, line 56: Only need test MySQL handler.

wubx (approved):

Issue #11035, opened 2023-04-11
Feature:  Support `Replace Into With Stage Attachement`


hantmac:
**Summary**

## Background
In CDC scenarios, `upsert` mode is a very common mode. In Databend, using "insert with stage"  during upsert operations can effectively save network traffic.

Now Databend can do insert with stage attachment as [this doc](https://databend.rs/doc/sql-commands/dml/dml-insert#insert-with-stage-attachment) shown. We want Databend can do `Replace Into with stage attachment` like `insert`:

### Syntax

```sql
REPLACE INTO [db.]table [(c1, c2, c3)] on(c1) VALUES
```
### Examples
```sql
CREATE TABLE  test(a int null, b int default 2, c float, d varchar default 'd');
```

```
INSERT INTO test(a,c) VALUES (1,1.0);
```

### values.csv

```
1,1.1
2,2.0
```
### Upload values.csv into stages:

```
curl -H "stage_name:my_int_stage" -F "upload=@./values.csv" -XPUT http://root:@localhost:8000/v1/upload_to_stage
```

### Replace Into with the uploaded stage:

```
curl -d '{"sql": "replace into test (a, c)  on(a) values", "stage_attachment": {"location": "@my_int_stage/values.csv", "file_format_options": {}, "copy_options": {}}}' -H 'Content-type: application/json' http://root:@localhost:8000/v1/query
```

### Check if the replace into succeeded:

```
SELECT * FROM test;
+------+------+------+------+
| a    | b    | c    | d    |
+------+------+------+------+
|    1 |    2 |  1.1 | d    |
|    2 |    2 |  2.0 | d    |
+------+------+------+------+
```
### More important

An important point to note is that "replace into with stage" must be atomic and idempotent.

PR #11519, opened 2023-05-19
feat: support replace with stage attachment

ZhiHanZ:

## Summary

1. Refactor insert and replace and move stage related things into `interpreter/common` crate
2. Support replace with stage attachment semnatics on databend http api
3. Add Stateful tests for expected behaviors on replace stage attachment
4. feat introduce a behavior change. 
Previously, insert into would purge whether it success or failed, in this PR insert into would NOT purge the file when failed and try best to purge file when success.
Replace into would behave the same, and would NOT purge the stage files when failed and purge file when success.

Closes [#issue](https://github.com/datafuselabs/databend/issues/11035)


flaneur2020:
LGTM

src/query/service/src/interpreters/interpreter_insert.rs
```diff
@@ -297,22 +208,6 @@ impl InsertInterpreter {
...
                     tracing::error!("insert stage file error: {}", error);
-                    GlobalIORuntime::instance()
-                        .block_on(async move {
-                            if stage_info.copy_options.purge {
-                                info!(
-                                    "insert: try to purge files:{}, elapsed:{}",
-                                    all_source_files.len(),
-                                    start.elapsed().as_secs()
-                                );
-                                Self::try_purge_files(ctx.clone(), &stage_info, &all_source_files)
```
dantengsky, line 308: @everpcpc  please have a look, seems that commit 3f2a530d (fix: also purge files when failure) purges copied files on purpose.  thanks.

everpcpc: We purge files on failure because we treat the 3-step insert as a whole process. When insert fails, we upload file to s3 again in previous design.

ZhiHanZ: I think it might be a redundant step, since we could rewrite the logic to reinsert/rereplace stage files as a whole, since the stage files would not be loaded if they commited in copyinto step

dantengsky (approved):

PR #11520, opened 2023-05-19
feat: add `system.indexes` and `show indexes` command.

RinChanNOWWW:

### Features

- Add a new table system table `system.indexes`. It's like:

```
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│  name  │     type    │                         definition                        │         created_on         │
│ String │    String   │                           String                          │          Timestamp         │
├────────┼─────────────┼───────────────────────────────────────────────────────────┼────────────────────────────┤
│ idx1   │ AGGREGATING │ SELECT SUM(a), b FROM default.t1 WHERE (b > 3) GROUP BY b │ 2023-05-20 02:41:50.143182 │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

- Add a new command `show indexes`. It's equivalent to `select * from system.indexes`.

### Refactor Meta API

- Remove index id list KV mapping (`_fd_index_id_list/<tenant>/<table_id>`). We can get id list by `prefix_list` the name ident key.
- API `get_indexes_by_table_id` --> `list_indexes`.  If the request doens't contain a table id, list all the indexes belong to the tenant; else filter the indexes with the table id. `list_indexes` will return `(index_id, index_name, index_meta)`. If `with_name` is false in `ListIndexReq`, the `index_name` field will be an empty string.

### Bug Fix

- Add `planning_agg_index` to `BindContext` to avoid death loop during planning. And avoid binding index plan when creating index.

### Others

When creating index, we will rewrite the query sql. If the query sql doesn't specify the database, add it.

For example:

```sql
select * from t;
-- rewrite to
select * from default.t;
````

### CODE REVIEW

File: src/meta/proto-conv/tests/it/v037_index_meta.rs [hunk 1]
```diff
@@ -54,7 +62,7 @@ fn test_decode_v37_index() -> anyhow::Result<()> {
     };
 
     common::test_pb_from_to(func_name!(), want())?;
-    common::test_load_old(func_name!(), index_v037.as_slice(), 36, want())?;
+    common::test_load_old(func_name!(), index_v037.as_slice(), 37, want())?;
```
RinChanNOWWW, line 65: No idea why this will be changed to 37. cc @drmingdrmer

src/query/storages/system/src/indexes_table.rs
```diff
@@ -0,0 +1,123 @@
...
+                        .to_vec(),
+                );
+            }
+
+            Ok(DataBlock::new_from_columns(vec![
+                StringType::from_data(names),
+                StringType::from_data(types),
+                StringType::from_data(tables),
+                StringType::from_data(defs),
+                StringType::from_data(created_on),
```
sundy-li, line 88: created_on  could use Timestamp DataType.

RinChanNOWWW: In other system tables (like `system.tables`), it is string type.

sundy-li (approved):

drmingdrmer:
> ```rust
> pub struct IndexMeta {
>     pub ident: IndexNameIdent, // newly added
> 
>     pub table_id: MetaId,
>     pub table_desc: String, // newly added
> 
>     pub index_type: IndexType,
>     pub created_on: DateTime<Utc>,
>     // if used in CreateIndexReq, this field MUST set to None.
>     pub drop_on: Option<DateTime<Utc>>,
>     pub query: String,
> }
> ```

What is `IndexNameIdent` for? And if a index is renamed, this field will be updated too?

With a `IndexNameIdent` in it, the query node has to be very careful about the bidirectional mapping between index name and index id. Is this field necessarily required?

The same for `table_desc`, user may use it to fetch a table, which may not point to the same TableMeta the `table_id` pointing to.

RinChanNOWWW:
> What is IndexNameIdent for? And if a index is renamed, this field will be updated too?

`IndexNameIdent` stores the tenant and the name of the index. We don't allow renaming an index. If we allow, we should update this field. And I think the update operation is unavoidable. (There was a `index_id -> (tenant, index_name)` mapping, I just move it to `IndexMeta`)

> With a IndexNameIdent in it, the query node has to be very careful about the bidirectional mapping between index name and index id. Is this field necessarily required?

For `SHOW INDEXES` command, we should get the index name. If we don't hold the info in `IndexMeta`, we need to do more KV operations, whch will increase the complexity. And there was already a bidirectional mapping (`index_id -> name` and `name -> index_id`), is this mapping ok?

> The same for table_desc, user may use it to fetch a table, which may not point to the same TableMeta the table_id pointing to.

The reason is the same as above, we need to do more KV operations, will this be heavy for `SHOW INDEXES` command? I have an idea to remove `table_desc`, that is to rewrite the table name in `query` with the full table name.

drmingdrmer:
@RinChanNOWWW 
So, will there be index renaming support?

This is critical to the design:

- If there will be renaming index, anything about `index name` should not be stored in `IndexMeta`. 

- If there won't be renaming, we can just replace the two level mapping `index_name -> index_id -> index_meta` with a one level mapping `index_name -> index_meta`. In other words, `IndexIdList` should be a `indexNameList`.

RinChanNOWWW:
> So, will there be index renaming support? 

cc @sundy-li

RinChanNOWWW:
I think supporting rename is ok. We mget `index id -> index name` and `index id -> index_meta`  parallelly

sundy-li:
```
So, will there be index renaming support?
```

renaming is not in plan, we can use drop/create instead.

RinChanNOWWW:
As a result, we decide to keep the current `IndexMeta` structure (without name) and APIs.

sundy-li (approved):

drmingdrmer:
I do not quite get the idea behind this design:

- There is a list of index ids for every tenant: `*/<tenant> -> Vec<IndexId>`;
- There is an index id to index name mapping: `*/<index-id> -> <index-name>`;

My question is, does index-name include a **table-name**? What data will be affected if the table a index belongs to is renamed?

It would be more clear if there is doc to describe the general design of these meta-data and the relations between them.

RinChanNOWWW:
> My question is, does index-name include a table-name?

No. But the table name is contained by `query` in `IndexMeta` (which is a SQL string defining the index, such as `select a, sum(b) from default.t group by a`)

> What data will be affected if the table a index belongs to is renamed?

Just like VIEW, if the table is renamed, the index on it will become **INVALID**.

RinChanNOWWW:
The relations:

```
(tenant, index name) -> index id
index id -> (tenant, index name)
index id -> index meta
tenant -> index id list (consists of index ids)
```

drmingdrmer:
> The relations:
> 
> ```
> (tenant, index name) -> index id
> index id -> (tenant, index name)
> index id -> index meta
> tenant -> index id list (consists of index ids)
> ```

`(tenant, index name) -> index id` is already a list of `index_id`, it looks there is no need to create another mapping `tenant -> index id list`.

Query can `prefix_list()` all index names for a tenant, then mget all index meta.

RinChanNOWWW:
> Query can prefix_list() all index names for a tenant, then mget all index meta.

Thanks for this hint @drmingdrmer

RinChanNOWWW:
I have removed key `__fd_index_id_list`. cc @drmingdrmer

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1043,41 +1019,48 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+        // Get index id list by `prefix_list` "<prefix>/<tenant>"
+        let prefix_key = kvapi::KeyBuilder::new_prefixed(IndexNameIdent::PREFIX)
+            .push_str(&req.tenant)
+            .done();
+
+        let id_list = self.prefix_list_kv(&prefix_key).await?;
+        let mut id_name_list = Vec::with_capacity(id_list.len());
+        for (key, seq) in id_list.iter() {
+            // Safety: the key is derived from the `prefix_key` above.
+            let name_ident = IndexNameIdent::from_str_key(key).unwrap();
```
drmingdrmer, line 1037: There may still be incorrect data responded. It would be better to return an error, such as `MetaNetworkError::InvalidReply`

drmingdrmer (approved):
LGTM.

A minor refactoring suggestion.

PR #11561, opened 2023-05-23
feat: scan data from aggregating index.

RinChanNOWWW:

### Fix query rewrite for aggregation query.

We need to aggregate the aggregation result in each aggregating index again to make the final aggregation results. Therefore, we should guarantee the group columns also appear in index.

We shouldn't compute the expression containing aggregation directly, because the aggregation result read from index is incomplete.

In another word, for plan `EvalScalar_2-Aggregate-EvalScalar_1-(Filter)-Scan`, we need to rewrite `EvalScalar_1` but not `EvalScalar_2`.

### Read and Deserialize and Evaluate index scan

If there is an aggregating index data file for the fuse data block, we read data from the index data file.

In deserializing phase, we deserialize the data into `DataBlock` and evaluate selection expressions on it.

The columns layout in the data block is the same as the one from fuse data block which is be executed by `EvalScalar` plan. This is to make the downstream operators can get the columns successfully. (P.S. original table scan and index scan are in the same pipeline.)

### Related PRs

- Write Aggregating Index. #11634


src/query/sql/src/evaluator/block_operator.rs
```diff
@@ -61,6 +63,15 @@ impl BlockOperator {
     pub fn execute(&self, func_ctx: &FunctionContext, mut input: DataBlock) -> Result<DataBlock> {
         match self {
             BlockOperator::Map { exprs } => {
+                if input
```
ariesdevil, line 66:
```
match self {
  BlockOperator::Map{..} | BlockOperator::MapWithOutput{..} | BlockOperator::Filter { .. } if input.get_meta().and_then(AggIndexMeta::downcast_ref_from).is_some() => {
 return Ok(input);
}
  other match arms... => ...
}
```

RinChanNOWWW:
PTAL @sundy-li

RinChanNOWWW:
SQLlogic tests will be added in next PR.

PR #11565, opened 2023-05-23
docs: computed columns

soyeric128:

## Summary

added doc: computed columns

Closes #issue


sundy-li:
The feature is reverted, so the docs could be later.

wubx:
Computed Columns need mark support version

BohuTANG:
This work can now be continued since the #11590  merged.

soyeric128:
> Computed Columns need mark support version

Marked as v1.1.56

docs/doc/14-sql-commands/00-ddl/20-table/10-ddl-create-table.md
```diff
@@ -231,6 +236,53 @@ SELECT * FROM t_default_value;
...
+```
+
+The following is an example of creating a stored computed column: Whenever the values of the "price" or "quantity" columns are updated, the "total_price" column will automatically recalculate and update its stored value.
+
+```sql
+CREATE TABLE IF NOT EXISTS products (
+  id INT,
+  price FLOAT,
+  quantity INT,
+  total_price FLOAT AS (price * quantity) STORED
```
b41sh, line 261: total_price should use `FLOAT64` type.

b41sh (approved):

PR #11571, opened 2023-05-24
refactor(merge-into-mutator): use dedicate thread pool to decode block data

dantengsky:

## Summary

If data blocks needed to be loaded during the execution of `replace into`,  submit the deserialization task to a dedicated thread pool, preventing async tasks from being blocked.

Closes #issue


dantengsky:
convert to draft. Adding shuffle to the segment partition step.

PR #11578, opened 2023-05-25
fix: fix drop data mask bug

lichuang:

## Summary

fix: fix drop data mask bug

Closes #issue

### DIFF
--- a/src/meta/api/src/data_mask_api_impl.rs
+++ b/src/meta/api/src/data_mask_api_impl.rs
@@ -134,10 +134,20 @@ impl<KV: kvapi::KVApi<Error = MetaError>> DatamaskApi for KV {
 
         loop {
             trials.next().unwrap()?;
-            let (id_seq, id, data_mask_seq, _) =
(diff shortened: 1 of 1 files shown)


src/meta/api/src/data_mask_api_impl.rs
```diff
@@ -134,10 +134,20 @@ impl<KV: kvapi::KVApi<Error = MetaError>> DatamaskApi for KV {
...
-            let (id_seq, id, data_mask_seq, _) =
-                get_data_mask_or_err(self, name_key, format!("drop_data_mask: {}", name_key))
-                    .await?;
 
+            let result =
+                get_data_mask_or_err(self, name_key, format!("drop_data_mask: {}", name_key)).await;
+
+            let (id_seq, id, data_mask_seq, _) = match result {
+                Ok((id_seq, id, data_mask_seq, meta)) => (id_seq, id, data_mask_seq, meta),
+                Err(e) => {
```
RinChanNOWWW, line 143: Should check if the error type is "already exists"

RinChanNOWWW (dismissed):

Issue #11148, opened 2023-04-20
Tracking: general window functions.


RinChanNOWWW:
**Summary**

- [x] row_number #10824
- [x] rank #10824
- [x] dense_rank #10824
- [x] percent_rank #11094
- [x] lag #11244 
- [x] lead #11244 
- [x] first_value #11587 
- [x] last_value #11587 
- [x] first(alias first_value) #11587 
- [x] last(alias last_value) #11587 
- [x] cume_dist #11942 
- [x] ntile #11905 
- [x] nth_value #11600

gitccl:
Hi, I'd like to try `first_value` and `last_value` function.

ariesdevil:
Hi @ArberSephirotheca , do you still working on `first_value` and `last_value` ? If not, these will assign to @gitccl .

PR #11587, opened 2023-05-25
feat: add window function first_value and last_value

gitccl:

## Summary

Add window function `first_value` and `last_value`.

Part of #11148

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
+++ b/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
@@ -551,10 +551,9 @@ impl<T: Number> TransformWindow<T> {
             }
             WindowFunctionImpl::FirstValue(first) => {
                 let value = if self.frame_start < self.frame_end {
-                    let block = self
+                    let block = &self
                         .blocks
                         .get(self.frame_start.block - self.first_block)
-                        .cloned()
                         .unwrap()
                         .block;
                     let col = block.get_by_offset(first.arg).value.as_column().unwrap();
@@ -570,10 +569,9 @@ impl<T: Number> TransformWindow<T> {
                     let last_row = self.goback_row(self.frame_end);
                     debug_assert!(self.frame_start <= last_row);
 
-                    let block = self
+                    let block = &self
                         .blocks
                         .get(last_row.block - self.first_block)
-                        .cloned()
                         .unwrap()
                         .block;
                     let col = block.get_by_offset(last.arg).value.as_column().unwrap();
--- a/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
+++ b/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
@@ -345,6 +345,17 @@ impl<T: Number> TransformWindow<T> {
         row
     }
 
+    /// Goback the row to the preceding one row
+    fn goback_row(&self, mut row: RowPtr) -> RowPtr {
+        if row.row != 0 {
+            row.row -= 1;
+        } else {
+            row.block -= 1;
+            row.row = self.block_rows(&row) - 1;
+        }
+        row
+    }
+
     /// If the two rows are within the same peer group.
     fn are_peers(&self, lhs: &RowPtr, rhs: &RowPtr, for_computing_bound: bool) -> bool {
         if lhs == rhs {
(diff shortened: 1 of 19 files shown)


src/query/sql/src/planner/plans/window.rs
```diff
@@ -270,6 +257,14 @@ impl WindowFuncType {
                 default: default.map(Box::new),
                 return_type: Box::new(return_type),
             })),
+            "first_value" => Ok(WindowFuncType::FirstValue(FirstLastFunction {
```
ariesdevil, line 260: This function `get_general_window_func` is called by type_check and used for lag/lead only cause lag/lead 's return type depends on the default argument.
I see in the `transform_window.rs`, if the `frame_start >= frame_end` the value will be `Scalar::Null`, and the builder's data type is same as `first_value(col)`'s col data type, so if you need push null to the builder, you should change the return type in the `type_check.rs`.

ariesdevil:
The `first_value` can be aliased to `first` and the `last_value` is the same, you can do alias here.

src/query/sql/src/planner/plans/window.rs
gitccl: I think we cannot know whether it returning null in advance, because the range window frame will generate unpredictable frame.

tests/sqllogictests/suites/query/window_function/window_basic.test
```diff
@@ -239,6 +239,92 @@ SELECT salary, lead(salary, 2, salary) OVER (ORDER BY enroll_date) FROM empsalar
...
+develop 4200 4200
+develop 4500 4500
+develop 5200 5200
+develop 5200 5200
+develop 6000 6000
+personnel 3500 3500
+personnel 3900 3900
+sales 4800 4800
+sales 4800 4800
+sales 5000 5000
```
RinChanNOWWW, line 326: The expected results are not right.

src/query/service/src/pipelines/processors/transforms/window/window_function.rs
```diff
@@ -106,6 +114,8 @@ pub enum WindowFunctionImpl {
     PercentRank,
     Lag(WindowFuncLagLeadImpl),
     Lead(WindowFuncLagLeadImpl),
+    FirstValue(WindowFuncFirstLastImpl),
+    LastValue(WindowFuncFirstLastImpl),
```
RinChanNOWWW, line 118: I think we don't need make a `XXXImpl` for `first_value` and `last_value`. I will give a detail explanation below.

tests/sqllogictests/suites/query/window_function/window_bound.test
```diff
@@ -71,6 +71,34 @@ SELECT sum(unique1) over (order by unique1 rows between unbounded preceding and
...
+1 9
+2 9
+3 9
+4 9
+5 9
+6 9
+7 9
+8 9
+9 9
+
```
RinChanNOWWW, line 101: Need more test cases with `partition by`. And better to add test cases without unbounded bound.

tests/sqllogictests/suites/query/window_function/window_basic.test
RinChanNOWWW: And can add the test cases in https://github.com/duckdb/duckdb/blob/master/test/sql/window/test_basic_window.test

RinChanNOWWW (changes requested):
I think we don't need make a `XXXImpl` for `first_value` and `last_value`.

We just need to push the value at `frame_start` (when `frame_started`) and `frame_end - 1` (when `frame_ended`) to the result.

RinChanNOWWW (changes requested):
The method `resolve_general_window_function` in src/query/sql/src/planner/semantic/type_check.rs is for `lag` and `lead` only now, should modify it.

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -541,6 +552,41 @@ impl<T: Number> TransformWindow<T> {
                 let builder = &mut self.blocks[self.current_row.block - self.first_block].builder;
                 builder.push(lead_row.as_ref());
             }
+            WindowFunctionImpl::FirstValue(first) => {
+                let value = if self.frame_start < self.frame_end {
```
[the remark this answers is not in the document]
RinChanNOWWW: see the variable `is_empty_frame`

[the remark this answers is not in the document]
gitccl: It seems that the case `self.frame_start >= self.frame_end` will occur when use range window frame.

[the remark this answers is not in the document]
RinChanNOWWW: Oh, `self.frame_start == self.frame_end` will occur.

src/query/sql/src/planner/plans/window.rs
ariesdevil: Oh, yes, i forget it.

RinChanNOWWW: @gitccl is right. We can make the return type of all the window functions with arguments 
 and frame bound nullable.

RinChanNOWWW (approved):
LGTM

ariesdevil (approved):

PR #11588, opened 2023-05-25
feat(FlightSQL): should not use session ID as token

### LINKED ISSUE #11583: Feature(FlightSQL): should not use session ID as token.
not safe since other user can see it when `show processlist `

jun0315:

## Summary

Encrypt session id with salt

Closes #11583


sundy-li:
We need to consider the decoder.

session_id + salt  ---------- encode ------>  token; 

We save `token` as the key of ` pub sessions: Mutex<ExpiringMap<String, Arc<Session>>>,`

token  ---------- decode ------>   session_id

jun0315:
Perhaps we can get the session-id through the session.

youngsofun:
> Perhaps we can get the session-id through the session.

so token do not need to be related to session？just another uuid is enough？

sundy-li:
I guess `uuid` is enough, no need any encoder.

jun0315:
Ok, I will use `UUID`.

sundy-li (approved):

PR #11589, opened 2023-05-25
fix: fix lag/lead default value when column is nullable

ariesdevil:

## Summary

Summary about this PR
1. fix lag/lead default value when the column is nullable
2. fix count(*) can be used in window function
3. add more window function tests that derived from duckdb
4. fix lag function with partition by

Closes #issue

### DIFF
--- a/src/query/expression/src/values.rs
+++ b/src/query/expression/src/values.rs
@@ -2443,7 +2443,11 @@ impl ColumnBuilder {
                     field.append_column(other_field);
                 }
             }
-            (this, other) => unreachable!("unable append column(data type: {:?}) into builder(data type: {:?})", other.data_type(), this.data_type()),
+            (this, other) => unreachable!(
+                "unable append column(data type: {:?}) into builder(data type: {:?})",
+                other.data_type(),
+                this.data_type()
+            ),
         }
     }
 
--- a/src/query/expression/src/values.rs
+++ b/src/query/expression/src/values.rs
@@ -2443,7 +2443,7 @@ impl ColumnBuilder {
                     field.append_column(other_field);
                 }
             }
-            (this, other) => unreachable!("unable append {other:?} into {this:?}"),
+            (this, other) => unreachable!("unable append column(data type: {:?}) into builder(data type: {:?})", other.data_type(), this.data_type()),
         }
     }
 
--- a/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
+++ b/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
@@ -490,54 +490,50 @@ impl<T: Number> TransformWindow<T> {
                 builder.push(ScalarRef::Number(NumberScalar::Float64(percent.into())));
             }
             WindowFunctionImpl::Lag(lag) => {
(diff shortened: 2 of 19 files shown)


src/query/expression/src/values.rs
```diff
@@ -2443,7 +2443,11 @@ impl ColumnBuilder {
                     field.append_column(other_field);
                 }
             }
-            (this, other) => unreachable!("unable append {other:?} into {this:?}"),
+            (this, other) => unreachable!(
```
sundy-li, line 2446: there are another two places.

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -490,28 +490,29 @@ impl<T: Number> TransformWindow<T> {
...
-                        col.index(self.frame_start.row).unwrap().to_owned()
+                let lag_row = if self.current_row.block == self.partition_start.block
+                    && self.current_row.row < self.partition_start.row + (lag.offset as usize)
+                {
+                    let default_value = match lag.default.clone() {
+                        LagLeadDefault::Null => Scalar::Null,
+                        LagLeadDefault::Index(col) => {
+                            let block =
+                                &self.blocks[self.current_row.block - self.first_block].block;
+                            let col = block.get_by_offset(col).value.as_column().unwrap();
```
sundy-li, line 501:
```
let value = &block.get_by_offset(lag.arg).value;
value.index(self.current_row.row).unwrap().to_owned()
```

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -490,28 +490,29 @@ impl<T: Number> TransformWindow<T> {
...
+                            let value = &block.get_by_offset(col).value;
+                            value.index(self.current_row.row).unwrap().to_owned()
+                        }
                     };
+                    default_value
+                } else {
+                    let block = self
+                        .blocks
+                        .get(self.frame_start.block - self.first_block)
+                        .cloned()
```
RinChanNOWWW, line 510: This clone is not need.

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -490,28 +490,29 @@ impl<T: Number> TransformWindow<T> {
...
-                        let block = self
-                            .blocks
-                            .get(self.frame_start.block - self.first_block)
-                            .cloned()
-                            .unwrap()
-                            .block;
-                        let col = block.get_by_offset(lag.arg).value.as_column().unwrap();
-                        col.index(self.frame_start.row).unwrap().to_owned()
+                let lag_row = if self.current_row.block == self.partition_start.block
+                    && self.current_row.row < self.partition_start.row + (lag.offset as usize)
```
RinChanNOWWW, line 494: We can just check if `self.frame_start == self.frame_end` to judge if the frame is empty (or invalid).

sundy-li (approved):

RinChanNOWWW (approved):

Issue #6994, opened 2022-08-05
Feature: Tracking issue for JSON optimization


b41sh:
**Summary**

Optimize JSON read performance by implementing Virtual Column and JSONB binary format

related #5925

## Task list

- [x] [RFC](https://github.com/datafuselabs/databend/pull/6995)
- [ ] FPGrowth algorithm for statistics
- [x] Extract virtual columns
- [x] Push down JSON key path access
- [x] https://github.com/datafuselabs/databend/issues/7668

Xuanwo:
Putting RFC as the first task is excellent!

haydenflinner:
This will be awesome, it's a great usability feature that it's always felt like is possible but seems to be missing in other db/warehouse/lakehouse solutions!

PR #11590, opened 2023-05-26
feat(query): Support generate virtual columns

b41sh:

## Summary

Support generating virtual columns

Add the following SQLs to create, drop and generate virtual columns:
```sql
create virtual columns (<expr>, ..) for <table>
alter virtual columns (<expr>, ..) for <table>
drop virtual columns for <table>
generate virtual columns for <table>
```

Using a log analysis script to get which JSON internal fields are frequently queried and then calls `create virtual columns ..` to create virtual columns for these fields, and then calls `generate virtual columns ..` to generate virtual columns. Once the virtual columns have been generated, we can read the data from the JSON internal fields directly through the virtual columns.

for example:
```sql
create table test (id int, val json);
insert into test values(1, '{"a":33,"b":44}'),(2, '{"a":55,"b":66}');
create virtual columns (val['a'], val['b']) for test;
generate virtual columns for test;

select val['a'], val['b'] from test;
+----------+----------+
| val['a'] | val['b'] |
+----------+----------+
| 33       | 44       |
| 55       | 66       |
+----------+----------+
```
Part of #6994


RinChanNOWWW:
What does `Using a log analysis script to get which JSON internal fields are frequently queried` mean?

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1071,6 +1080,197 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+                    HashSet::from_iter(virtual_columns.into_iter());
+                let virtual_columns = Vec::from_iter(virtual_columns_set.into_iter());
+
+                VirtualColumnMeta {
+                    table_id: req.name_ident.table_id,
+                    virtual_columns,
+                    created_on: old_virtual_column.created_on,
+                    updated_on: Some(Utc::now()),
+                    drop_on: old_virtual_column.drop_on,
+                }
```
drmingdrmer, line 1113: It looks like `VirtualColumnMeta` should provide methods to operate it, such as
- `VirtualColumnMeta::update_columns(add: HashSet<String>, remove: HashSet<String>)` for update columns and the `updated_on` timestamp.
Then this block of codes will be clear and simple

src/query/storages/fuse/src/io/read/block/block_reader_native.rs
```diff
@@ -153,6 +153,24 @@ impl BlockReader {
             results.insert(index, readers);
         }
 
+        // If virtual column file exists, read the data from the virtual columns directly.
+        if let Some(ref virtual_columns_meta) = part.virtual_columns_meta {
+            let virtual_loc = part.location.replace(".parquet", "_virtual.parquet");
```
sundy-li, line 158: @RinChanNOWWW  I think we should have a folder to store the virtual parquet files.

b41sh:
> What does `Using a log analysis script to get which JSON internal fields are frequently queried` mean?

We want the virtual columns to be generated automatically, without the user having to create them manually. So a Python script is needed to analyze the query log, find the internal JSON fields that are frequently queried, and call the `add virtual columns` and `generate virtual columns` commands to generate the virtual columns. This script can be implemented in the next PR.

src/meta/app/src/schema/virtual_column.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct CreateVirtualColumnReply {}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct DropVirtualColumnReq {
+    pub name_ident: VirtualColumnNameIdent,
+    pub virtual_columns: Vec<String>,
+}
```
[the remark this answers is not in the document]
b41sh: dropping virtual columns means removing some of the virtual columns that already exist, not removing them all.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1071,6 +1080,197 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+
+    async fn drop_virtual_column(
+        &self,
+        req: DropVirtualColumnReq,
+    ) -> Result<DropVirtualColumnReply, KVAppError> {
+        debug!(req = debug(&req), "SchemaApi: {}", func_name!());
+
+        let mut retry = 0;
+        while retry < TXN_MAX_RETRY_TIMES {
+            retry += 1;
```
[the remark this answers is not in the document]
b41sh: ok

src/query/storages/fuse/src/io/read/block/block_reader_native.rs
b41sh: add a folder `_vb` to store virtual column blocks

sundy-li (approved):

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1071,6 +1082,222 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+            {
+                let if_then = vec![
+                    txn_op_put(&req.name_ident, serialize_struct(&virtual_column_meta)?), /* (tenant, table_id) -> virtual_column_meta */
+                ];
+
+                let txn_req = TxnRequest {
+                    condition: vec![],
+                    if_then,
+                    else_then: vec![],
+                };
```
drmingdrmer, line 1130: It still has to assert the seq number to be `0`. otherwise two processes will override each others' data.
I.e.:
```rust
                let condition = vec![
                    txn_cond_seq(req.name_ident, Eq, 0),
                ]
```

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1071,6 +1082,222 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+            // Update virtual column by inserting this record:
+            // (tenant, table_id) -> virtual_column_meta
+            {
+                let if_then = vec![
+                    txn_op_put(&req.name_ident, serialize_struct(&virtual_column_meta)?), /* (tenant, table_id) -> virtual_column_meta */
+                ];
+
+                let txn_req = TxnRequest {
+                    condition: vec![],
+                    if_then,
```
drmingdrmer, line 1192: The condition must assert the seq number does not change when applying this transaction

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1071,6 +1082,222 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+                return Err(KVAppError::AppError(AppError::VirtualColumnNotFound(
+                    VirtualColumnNotFound::new(
+                        req.name_ident.table_id,
+                        format!(
+                            "drop virtual column with tenant: {} table_id: {}",
+                            req.name_ident.tenant, req.name_ident.table_id
+                        ),
+                    ),
+                )));
+            }
```
drmingdrmer, line 1237: Such a routine can be extracted to a function, such as:
https://github.com/datafuselabs/databend/blob/529d184808eecbf1cd09ce31f3629553478f8ef2/src/meta/api/src/util.rs#L417-L437

src/meta/api/src/schema_api_impl.rs
```diff
@@ -1071,6 +1082,202 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
...
+            // Drop virtual column by deleting this record:
+            // (tenant, table_id) -> virtual_column_meta
+            {
+                let condition = vec![txn_cond_seq(&req.name_ident, Eq, seq)];
+                let if_then = vec![txn_op_del(&req.name_ident)];
+                let txn_req = TxnRequest {
+                    condition,
+                    if_then,
+                    else_then: vec![],
+                };
```
drmingdrmer, line 1226: If it just the record with a seq that is just fetched, then it does not have to assert the seq in the `condition`. 
Asserting the `seq` is only useful when the seq is passed in by callers.

drmingdrmer (approved):

src/meta/api/src/schema_api_impl.rs
b41sh: I didn't understand it clearly, let me delete them

BohuTANG:
cc @soyeric128 for documentation, thanks.

PR #11592, opened 2023-05-26
refactor: switch table meta encoding scheme to `msgpack`

dantengsky:

## Summary

`bincode` is excellent, except that it does not support schema evolution, which is an essential feature for us. 

- switch table meta encoding format from bincode to msgpack
- introduce table meta version 4
  - freeze table-meta related types in table meta v3 
  
    e.g. `TableSchema`, `BlockMeta`, `Statistics`, `ColumnMeta`, etc. in mod `v3::frozen`, to make the upcoming meta type evolutions easier. non-frozen corresponding types could be "upgraded" as long as `serde`  can handle the backward compatibility issues, without bumping table meta version.
    
    but not all the types are (or easy to be) frozen in v3, e.g. the `Scalar`. 

- backward compatibility test for releases 1.1.30, 1.1.38, 1.1.39, 1.1.46

    the test scripts need to be refined (to avoid duplications of test cases), will be addressed in dedicated pr.
    
        
for reviewers:

- the "evolution" cases I've come up with
  https://github.com/datafuselabs/databend/blob/b6021c7fdd6a2c0f2da78237d1fc9317ecf3d349/src/query/service/tests/it/storages/fuse/meta/serialization_format_compatability.rs
  
  If any other cases need to be tested, please let me know. 

tasks:

- [x] evaluating [`pot`](https://github.com/khonsulabs/pot) and `msgpack(named)`
   let's switch to msgpack
- [x] bump table meta version to V4
- [x] add fuse-compat test

evaluating  `msgpack(named)`

- benchmarks

  for a fully compacted segment (1000 blocks, 23 columns), 

  time for decoding(decompress + deserialize) :   

    - bincode 8.2859 ms
    - msgpack 18.579 ms
   
  time for encoding( compress + serialize) :   
    - bincode 8.6988 ms
    - msgpack 16.923 ms

  size of binary:
  
    segment_msgpack_bytes / segment_bincode_bytes: 1.5107816711590296

- backward compact test

   `msgpack` 's ["named encoding"](https://docs.rs/rmp-serde/latest/rmp_serde/encode/fn.to_vec_named.html) passes the following backward compact test

    https://github.com/datafuselabs/databend/blob/61ba1a9ff8a819d2cc0f71da60f131794c2e147e/src/query/service/tests/it/storages/fuse/meta/serialization_format_compatability.rs
    
------

benchmark :

https://github.com/datafuselabs/databend/blob/61ba1a9ff8a819d2cc0f71da60f131794c2e147e/src/query/storages/common/table-meta/benches/bench.rs
   
   local result: `cargo bench -p storages-common-table-meta`
   
~~~
Benchmarking decoding/bincode-decode-block-metas: Collecting 100 samples in estimated 5.6910 s (700 iteratdecoding/bincode-decode-block-metas
                        time:   [8.2057 ms 8.2859 ms 8.3715 ms]
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
Benchmarking decoding/msg-pack-decode-block-metas: Collecting 100 samples in estimated 6.5603 s (400 iteradecoding/msg-pack-decode-block-metas
                        time:   [17.628 ms 18.579 ms 19.624 ms]
Found 18 outliers among 100 measurements (18.00%)
  18 (18.00%) high severe
Benchmarking decoding/bincode-segment-deserialization: Collecting 100 samples in estimated 5.1036 s (600 idecoding/bincode-segment-deserialization
                        time:   [8.5979 ms 8.6988 ms 8.8083 ms]
Found 13 outliers among 100 measurements (13.00%)
  8 (8.00%) high mild
  5 (5.00%) high severe
Benchmarking decoding/msg-pack-segment-deserialization: Collecting 100 samples in estimated 6.5793 s (400 decoding/msg-pack-segment-deserialization
                        time:   [16.770 ms 16.923 ms 17.087 ms]
Found 15 outliers among 100 measurements (15.00%)
  14 (14.00%) high mild
  1 (1.00%) high severe
----------------------------------
segment_bincode_bytes: 742
segment_msgpack_bytes: 1121
segment_msgpack_bytes / segment_bincode_bytes: 1.5107816711590296
~~~
   

Closes #issue


src/query/storages/common/table-meta/src/meta/v4/segment.rs
```diff
@@ -0,0 +1,285 @@
...
+    ///
+    /// # Returns
+    ///
+    /// A Result containing the serialized Segment data as a byte vector. If any errors occur during
+    /// encoding, compression, or writing to the byte vector, an error will be returned.
+    pub fn to_bytes(&self) -> Result<Vec<u8>> {
+        self.to_bytes_with_encoding(MetaEncoding::MessagePack)
+    }
+
+    pub fn to_bytes_with_encoding(&self, encoding: MetaEncoding) -> Result<Vec<u8>> {
```
sundy-li, line 147: this should not be `pub` in case someone use `to_bytes_with_encoding(MetaEncoding::Bincode)` directly

.github/actions/fuse_compat/action.yml
```diff
@@ -24,7 +24,10 @@ runs:
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.0.56 base
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.0.56 01_meta_compression 01_flashback
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.30 base
-
+          bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.30 02_meta_compression_v3_to_v4 02_flashback_v3_to_v4
```
sundy-li, line 27: if the ci passed, maybe we could remove some old compact tests later.

src/query/storages/common/table-meta/src/meta/v4/segment.rs
dantengsky: this method is used by the benchmark code. let me comment out the bench_function

sundy-li (approved):

.github/actions/fuse_compat/action.yml
dantengsky: unfortunately,  the compatibility *in general*, is not transitive (yet).

if C is compatible with B, and B compatible with A; C may be incompatible with A.

e.g. if in version C, the types that are supposed be to sealed in A are modified by accident, in the compact test, we only generate data of version B, and test C against version B, we can not detect that C can no longer read data of version A correctly. (since the test will not cover case that migrating from C to A).

this way of backward compatibility testing seems not sustainable though :thinking:

PR #11609, opened 2023-05-29
docs: Lazy topn threshold

soyeric128:

## Summary

added description about how the lazy_topn_threshold setting works.

Closes #issue


docs/doc/14-sql-commands/20-query-syntax/01-query-select.md
```diff
@@ -338,6 +338,24 @@ SELECT number FROM numbers(100000) ORDER BY number LIMIT 2 OFFSET 10;
 +--------+
 ```
 
+For optimizing query performance with large result sets, you can enable the *lazy_topn_threshold* option. This option is specifically designed for queries that involve an ORDER BY clause and a LIMIT clause. When enabled, the optimization is activated for queries where the specified LIMIT number is smaller than the threshold value you set.
```
RinChanNOWWW, line 341: > For optimizing query performance with large result sets

Maybe small result set from large data set?

RinChanNOWWW (approved):

Issue #11605, opened 2023-05-29
bug: The cache  max_bytes don't work


wubx:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

all

### What's Wrong?

The user define cache.disk  max_bytes=161061273500 （~150G)

But after a few days of actual running:
![image](https://github.com/datafuselabs/databend/assets/320680/77d29a70-565d-4bab-badd-d451250f73b2)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

dantengsky:
seems that the items eviction of method `DiskCache::insert_bytes` is buggy: 

https://github.com/datafuselabs/databend/blob/aa97d2c5b57f1585abc8192e5fb76f9fc4157903/src/query/storages/common/cache/src/providers/disk_cache.rs#L198C1-L208

cached items should be kept being evicted, till enough room is made for the new item.

PR #11620, opened 2023-05-30
fix: cached items should be kept being evict, till enough room be made for the new item.

lichuang:

## Summary

fix: cached items should be kept being evict, till enough room be made for the new item.

Closes https://github.com/datafuselabs/databend/issues/11605


dantengsky (approved):
LGTM

src/query/storages/common/cache/src/providers/disk_cache.rs
```diff
@@ -195,7 +195,7 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize>
         }
 
         // check eviction
-        if self.cache.size() + bytes_len > self.cache.capacity() {
+        while self.cache.size() + bytes_len > self.cache.capacity() {
```
BohuTANG, line 198: Does this will block the write?

BohuTANG, line 198: Does this will block the write?

dantengsky: yes, it does. but at this point, the cache is full, other write operations should be blocked(or doing cache eviction as well), imo

BohuTANG: Ok, means this cache is full need to evict will block query or only block the cache write?

dantengsky: got your concern.  

the operation of getting items from the cache will also be blocked in this situation (a write lock is acquired).

https://github.com/datafuselabs/databend/blob/aa97d2c5b57f1585abc8192e5fb76f9fc4157903/src/query/storages/common/cache/src/providers/disk_cache.rs#L352-L359

BohuTANG: Oh, this maybe a problem.
The cache eviction should not block the read query(if need the cache to evict).

dantengsky: > If block the read query, for a case:
> 
> 1. disk cache settings to 1G
> 2. disk cache already full to 1G and need to evict
> 3. a read query scan 200MB data
> 4. we evict 200MB cache for new query and the disk cache still full soon.
> 5. a read query scan 300MB data
> 6. we evict 300MB cache ...
> 7. It seems always to evict, i think this will hurt the query

agree,  especially if the `300MB` in step 5 is not partially cached in step 4.

but it seems that the max size of the cache, and the eviction policy matters more in this situation. 

or put it another way,  even without the above exclusive-access problem, if the size of cache is too small, and using LRU policy, a large volume of data reading (of difference blocks), will effectively make the disk cache almost non-functional.

BohuTANG: We can evict larger size cache size(like half of the cache) one time?

dantengsky: > We can evict larger size(like 1GB) one time?

not sure.  I think there are two categories of concerns here:

1. will this PR degrade the performance?

yes, likely. 

but it is due to that the current implementation(main branch) is buggy : (

for the scenario that triggers the bug, more cache files should be removed (while holding the write lock).

the async cache population mechanism (eviction is a "side effect" of cache population) may mitigate this situation,  but there are definitely negative performance impacts there.

2. can we improve it?

yes. a rough idea I can come up with is:

during the eviction, after figuring out the files need to be removed, release the lock immediately, and return their paths to the caller (the cache population worker),  let the population worker do the deletion task(without holding the lock),  or asyncly if necessary(possibly not).

BohuTANG: We can't have a background worker, so if the cache is full only degraded this query is ok, butwe can not degraded each query.

PR #11621, opened 2023-05-30
feat(query): add range function & aliases in select list

sundy-li:

## Summary

Summary about this PR

1. support `range` function to generate array
2. make aliases work in select list:
   1) the latter expression in select list would respect the previous alias
   2) where/selection will prefer to fetch name from context, other clauses will prefer fetch aliases.
   
  ```
  impl ExprContext {
    pub fn prefer_resolve_alias(&self) -> bool {
        !matches!(self, ExprContext::SelectClause | ExprContext::WhereClause)
    }
}
  ```

Closes #issue


src/query/sql/src/planner/binder/project.rs
```diff
@@ -229,12 +229,18 @@ impl Binder {
                     };
                 }
                 SelectTarget::AliasedExpr { expr, alias } => {
+                    let pre_aliases = output
+                        .items
+                        .iter()
+                        .map(|item| (item.alias.clone(), item.scalar.clone()))
+                        .collect::<Vec<_>>();
```
RinChanNOWWW, line 236: Better to collect `pre_aliases` dynamically during the loop.

RinChanNOWWW (approved):

ariesdevil (approved):

sundy-li:
Dev lint will be fixed in #11629

PR #11661, opened 2023-06-03
fix(storage/s3): Add allow_anonymous support for storage

Xuanwo:

## Summary

fix(storage/s3): Add allow_anonymous support for storage


drmingdrmer (approved):

Xuanwo:
Hi, @soyeric128. This pull request introduces a new option called `allow_anonymous` for the S3 storage system. When `allow_anonymous=true` is enabled, users can query a stage without credentials if the bucket has public read access. Otherwise, Databend will return an error indicating that credentials were not found.

Would like to add this option in our docs? Thanks in advance.

PR #11662, opened 2023-06-04
feat: support distributed Top-N.

RinChanNOWWW:

## Summary

- Fix the original Top-N pushing down logic. Only pattern `Limit-Sort-Exchange` can be optimized.
- Skip the partial sort phase if the sort plan is after an exchange plan. (TBD: can we skip the first merge sort phase for each pipeline? @zhang2014 )


zhang2014 (approved):

zhang2014:
> can we skip the first merge sort phase for each pipeline

It's ok if we don't assume the order of the blocks.

sundy-li (approved):

PR #11669, opened 2023-06-05
feat: support merge join

xudong963:

## Summary

The PR introduces merge join for non-equi join:

1. For sql can't use iejoin, will use merge join instead of cross join, such as only one non-equi condition.
2. Unify merge join and iejoin to range join, they can leverage much same code.

For join in databend, we'll choose different physical join type by conditions:

1. If there is equi condition -> hash join
2. If there are two or more ie conditions -> IEjoin
4. If there is only one ie condition or iejoin can't be used -> Merge join
5. Other cases will leverage hash join to execute cross join.

Closes #issue


leiysky (approved):

src/query/service/src/pipelines/processors/transforms/range_join/range_join_state.rs
```diff
@@ -0,0 +1,255 @@
...
+pub struct RangeJoinState {
+    pub(crate) ctx: Arc<QueryContext>,
+    // The origin data for left/right table
+    pub(crate) left_table: RwLock<Vec<DataBlock>>,
+    pub(crate) right_table: RwLock<Vec<DataBlock>>,
+    pub(crate) right_sorted_blocks: RwLock<Vec<DataBlock>>,
+    // For iejoin, it's L1: sort by the first join key
+    pub(crate) left_sorted_blocks: RwLock<Vec<DataBlock>>,
+    pub(crate) conditions: Vec<RangeJoinCondition>,
+    // pub(crate) join_type: JoinType,
```
Dousir9, line 49: useless code.

Dousir9 (approved):

src/query/service/src/pipelines/processors/transforms/range_join/range_join_state.rs
xudong963: Yeah, I'll make it not comment later

Issue #11585, opened 2023-05-25
Feature:  support `$<col_position>`


youngsofun:
**Summary**

motivation: https://github.com/datafuselabs/databend/issues/11581

need to Clarify where it can be used (may lead to diff impl)

snowflake supports mix usage of  col_position and col_name when possible 

```
select max($1), max(c) from variant_test where $1=true group by $1;
```

scopes to consider:
- statement:  select stage only?  or  any statement?
- clauses： all (where, group by.. )  vs `select only`
- allow mixed usage or not

proposal:

1.  all clauses and  mixed usage when possible 
    1.  table and parquet 
    2. json (only support $1)  (. not sure if it is easy to impl mixed use) 
1. CSV/TSV only support col_position

youngsofun:
cc @BohuTANG  @sundy-li

PR #11672, opened 2023-06-06
feat: support column position like $N.

youngsofun:

## Summary

tests: https://github.com/datafuselabs/databend/pull/11672/files#diff-0843122f9eb1c3be263fdc7c68a5cd9aa5edcea4fcf4b22873a5d351ebc30a84

Closes https://github.com/datafuselabs/databend/issues/11585


src/query/sql/src/planner/binder/bind_context.rs
```diff
@@ -321,6 +323,46 @@ impl BindContext {
         }
     }
 
+    pub fn search_column_position_recursively(
```
sundy-li, line 326: maybe we should not do it recursively,  only match the current bind context otherwise throw errors.

youngsofun: done

I am not sure, but it is no harm to start with the simple way.

sundy-li (approved):

PR #11673, opened 2023-06-06
docs: sql alias

soyeric128:

## Summary

added AS keyword for sql alias

Closes #issue

### DIFF
--- a/docs/doc/14-sql-commands/20-query-syntax/01-query-select.md
+++ b/docs/doc/14-sql-commands/20-query-syntax/01-query-select.md
@@ -42,7 +42,9 @@ SELECT number FROM numbers(3);
 +--------+
 ```
 
-You can also alias a column to make the column name more readable and understandable in the result:
+### AS Keyword 
+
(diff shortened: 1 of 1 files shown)


BohuTANG (approved):

docs/doc/14-sql-commands/20-query-syntax/01-query-select.md
```diff
@@ -68,9 +70,46 @@ SELECT number AS `Total` FROM numbers(3);
 +--------+
 ```
 
-## EXCLUDE Parameter
+If you alias a column in the SELECT clause, you can use the alias in the WHERE, GROUP BY, and HAVING clauses:
+
+```sql
+SELECT number * 2 AS a FROM numbers(3) WHERE (a + 1) % 3 = 0;
```
sundy-li, line 76: alias could also be used in select clause,  the latter expression could use the alias declared before.

```
select number * 2 as a, a * 2 as b from numbers(3)
```

PR #11681, opened 2023-06-06
refactor: improve single pipeline merge sort for Top-N.

RinChanNOWWW:

## Summary

- Add a new data structure `FixedHeap` (a max heap) inspired by https://github.com/rodolphito/fixed_heap. `FixedHeap`'s size is fixed. If a `FixedHeap` is full, it will evict the element with lowest priority when pushing new element into it.
- When doing merge sort by heap, we could drop useless block memory at once.

### Others

- Abstract common codes into one function.
- Simplify some codes.


src/query/pipeline/transforms/src/processors/transforms/transform_sort_merge_limit.rs
```diff
@@ -0,0 +1,225 @@
...
+            order_by_cols,
+            heap: FixedHeap::new(limit),
+            buffer: HashMap::with_capacity(limit),
+            cur_index: 0,
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl<R, Converter> AsyncAccumulatingTransform for TransformSortMergeLimit<R, Converter>
```
zhang2014, line 77: Why need `AsyncAccumulatingTransform`?

RinChanNOWWW: Oh, I should use `AccumulatingTransform`. It's a mistake.

sundy-li (approved):

PR #11700, opened 2023-06-08
docs: use databend as vector sink

soyeric128:

## Summary

Vector natively supports delivering data to Databend as a sink, instead of using clickhouse interface.

Closes #issue


docs/doc/11-integrations/10-data-tool/00-vector.md
```diff
@@ -1,73 +1,61 @@
...
-```
+compression = "none"
+database = "nginx" #Your database
+endpoint = "http://localhost:8000"
+table = "mytable" #Your table
+
+...
 
-```shell
 [sinks.databend_sink.auth]
```
BohuTANG, line 49: I think this does not work, we need a dsn here? cc @everpcpc 
We can try to deploy a nginx and test it follow the <Analyzing Nginx Access Logs with Databend> guide line :)

everpcpc: Vector sink currently use these settings, DSN could be supported later.

BohuTANG (approved):

PR #11704, opened 2023-06-08
feat: add redis cache

lichuang:

## Summary

feat: add redis cache

Closes #issue


src/query/storages/common/cache/src/providers/mod.rs
```diff
@@ -14,6 +14,7 @@
 
 mod disk_cache;
 mod memory_cache;
+mod redis_cache;
```
Xuanwo, line 17: How about implement redis cache via opendal redis service?

PR #11712, opened 2023-06-09
feat: port comparable row format from Apache Arrow to Databend.

RinChanNOWWW:

## Summary

- Port the [comparable row format](https://github.com/apache/arrow-rs/tree/master/arrow-row) from Apache Arrow to Databend.
- As we  generate a order column (comparable rows) in the single-pipeline merge sort phase, we can avoid the generation in the mutli-pipelines merge sort phase.

Closes #issue


src/query/expression/src/row/fixed.rs
```diff
@@ -0,0 +1,151 @@
...
+                let to_write = &mut out.data[start..end];
+                to_write[0] = 1;
+                let mut encoded = val.encode();
+                if !asc {
+                    // Flip bits to reverse order
+                    encoded.as_mut().iter_mut().for_each(|v| *v = !*v)
+                }
+                to_write[1..].copy_from_slice(encoded.as_ref());
+            } else {
+                out.data[start] = null_sentinel(nulls_first);
```
b41sh, line 125: forget to write the `val` to the `out` here?

src/query/expression/src/row/row_converter.rs
```diff
@@ -0,0 +1,238 @@
...
+            with_decimal_type!(|DECIMAL| match col {
+                DecimalColumn::DECIMAL(c, _) => {
+                    fixed::encode(out, c, validity, asc, nulls_first)
+                }
+            })
+        }
+        Column::Timestamp(col) => fixed::encode(out, col, validity, asc, nulls_first),
+        Column::Date(col) => fixed::encode(out, col, validity, asc, nulls_first),
+        Column::String(col) => variable::encode(out, col.iter(), validity, asc, nulls_first),
+        Column::Variant(col) => variable::encode(out, col.iter(), validity, asc, nulls_first),
```
b41sh, line 235: treating `Variant` as `String` may lead to unexpected results, we can treat the `Variant` as an unsupported type and sort it in other methods.

RinChanNOWWW: Ok, we can only support one variant order-by column currently.

Cargo.toml
```diff
@@ -186,19 +186,19 @@ opt-level = "s"
...
-incremental = false
-
-[profile.dev.package]
-addr2line = { opt-level = 3 }
-adler = { opt-level = 3 }
-gimli = { opt-level = 3 }
-miniz_oxide = { opt-level = 3 }
-object = { opt-level = 3 }
-rustc-demangle = { opt-level = 3 }
+# [profile.dev]
```
sundy-li, line 189: what's this change for ?

RinChanNOWWW: It's a mistake and it's for local debugging. Recover it soon,

b41sh (approved):

PR #11718, opened 2023-06-11
chore(sqllogictests): fix column types of MySQLClient

byteink:

## Summary

The length of column types should not be equal to the length of rows.

### DIFF
--- a/tests/sqllogictests/src/client/mysql_client.rs
+++ b/tests/sqllogictests/src/client/mysql_client.rs
@@ -73,11 +73,11 @@ impl MySQLClient {
             }
             parsed_rows.push(parsed_row);
         }
-        // Todo: add types to compare
         let mut types = vec![];
         if !parsed_rows.is_empty() {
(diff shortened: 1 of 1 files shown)


tests/sqllogictests/src/client/mysql_client.rs
```diff
@@ -75,6 +74,10 @@ impl MySQLClient {
             parsed_rows.push(parsed_row);
         }
         // Todo: add types to compare
+        let mut types = vec![];
+        if !parsed_rows.is_empty() {
+            types = vec![DefaultColumnType::Any; parsed_rows[0].len()];
+        }
```
xudong963, line 80:
```suggestion
        let mut types = vec![];
        if !parsed_rows.is_empty() {
            types = vec![DefaultColumnType::Any; parsed_rows[0].len()];
        }
        // Todo: add types to compare
```

xudong963 (approved):

BohuTANG:
Thanks for your contritution 👍 @byteink

PR #11745, opened 2023-06-13
feat(query): Support add/drop computed columns

b41sh:

## Summary

- Support adding virtual computed column.
- Drop column check whether the column is referenced by computed columns.
- Make the computed column as EE feature.

Closes #issue


src/query/ee/src/test_kits/mock_licnese_mgr.rs
```diff
@@ -0,0 +1,134 @@
...
+        tenants: Some(vec!["test".to_string()]),
+    }
+}
+
+pub struct MockLicenseManager {
+    // cache available settings to get avoid of unneeded license parsing time.
+    pub(crate) cache: DashMap<String, JWTClaims<LicenseInfo>>,
+}
+
+impl LicenseManager for MockLicenseManager {
```
ZhiHanZ, line 47: I don't think we need additional MockLicenseManager, since we could use RealLicenseManager::new(public_key) for customizable test suites

b41sh: ok

ZhiHanZ (approved):

sundy-li (approved):

Issue #11728, opened 2023-06-12
Feature:  alter table set options


dantengsky:
**Summary**

~~~
ALTER TABLE  table_name SET OPTIONS( options_list)
~~~

example:

~~~
ALTER TABLE  table_name
SET OPTIONS(snapshot_location = '1/2/_ss/xxxx.mpk', block_per_segment = '500')
~~~

this feature is kind of risky, intended to be used by advanced users(or cloud service),  for data migration, low level performance tuning, etc.

- catalog already have the method `update_table_meta`

https://github.com/datafuselabs/databend/blob/1abffb21d5ee950e719f082127dc536b7b01dd23/src/query/catalog/src/catalog/interface.rs#L188

- parser need to be extended

https://github.com/datafuselabs/databend/blob/45cb7dc00b9de9642f5aac9139b60f6f929ef388/src/query/ast/src/parser/statement.rs#L1785

- we also need to validate the table options (https://github.com/datafuselabs/databend/issues/11726)

PR #11746, opened 2023-06-13
feat: Alter set options

JackTan25:

## Summary

Summary about this PR
support "alter table table_name set options(xxx = xxx,....)";
Closes #issue
#11728


src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -232,6 +232,18 @@ impl AccessChecker for PrivilegeAccess {
...
             }
+            Plan::SetOptions(plan) => {
+                session
+                    .validate_privilege(
+                        &GrantObject::Table(
+                            plan.catalog.clone(),
+                            plan.database.clone(),
+                            plan.table.clone(),
+                        ),
+                        vec![UserPrivilegeType::Alter],
```
dantengsky, line 243: `UserPrivilegeType::Super` would be better, since altering arbitrary table options is risky, if this will not bring too much trouble to unit / integration testings

JackTan25: sure!!

dantengsky: turns out, using `Super` privilege does bring troubles to ut/it : ( let's address it later

JackTan25: ok

JackTan25:
> #11749 has been merged, it would be better, if we validate the value of option `block_per_segment` as well in this PR.

yes

JackTan25:
> 

ok, let me add the check

dantengsky:
> BTW, we can't alter the storage_format options

:+1: right & thanks for reminding.

this feature is "risky", let me remove the pr-feature label (so that the change log will omit it). let's keep this feature for internal usage only (a user may need this feature in the near future, for data migration).

@soyeric128  This feature need not be documented, thanks.

dantengsky (approved):

PR #11749, opened 2023-06-14
feat: block_per_segment check

### LINKED ISSUE #11726: Feature: validate `block_per_segment` setting while creating table
**Summary**

the feature "internal column" assumes that DEFAULT_BLOCK_PER_SEGMENT is less or equal to 1000.

https://github.com/datafuselabs/databend/blob/c8f15c28495fe9210d0b3701f3e0496112ca6fa8/src/query/catalog/src/plan/internal_column.rs#L38

thus, while creating a table with block_per_segment settings, we should check that this assumption is not violated.

  https://github.com/datafuselabs/databend/blob/0d77a053efad201d749d3df38b266208261c2d50/src/query/service/src/interpreters/interpreter_table_create.rs#L241
  
the corresponding table option key is:

https://github.com/datafuselabs/databend/blob/1abffb21d5ee950e719f082127dc536b7b01dd23/src/query/storages/fuse/src/constants.rs#L16

JackTan25:

## Summary

Closes #11726


src/query/service/src/interpreters/interpreter_table_create.rs
```diff
@@ -283,3 +286,17 @@ pub static CREATE_TABLE_OPTIONS: Lazy<HashSet<&'static str>> = Lazy::new(|| {
...
 }
+
+pub fn is_valid_block_per_segment(options: &BTreeMap<String, String>) -> Result<()> {
+    // check block_per_segment is not over 1000.
+    if let Some(value) = options.get(FUSE_OPT_KEY_BLOCK_PER_SEGMENT) {
+        let blocks_per_segment = value.parse::<u64>()?;
+        if blocks_per_segment > 1000 {
+            error!("invalid block_per_segment option, can't be over 1000");
+            return Err(ErrorCode::TableOptionInvalid(format!(
+                "invalid block_per_segment option, can't be over 1000"
```
dantengsky, line 297: shall we introduce an error string, and use it both in the log message and error message?

JackTan25: yes

dantengsky:
cargo clippy complains :)

https://github.com/datafuselabs/databend/actions/runs/5264195570/jobs/9515182514?pr=11749#step:4:4065

dantengsky (approved):

PR #11751, opened 2023-06-14
feat: implement Background service for databend

ZhiHanZ:

## Summary
Databend Internal Storage FuseTable is similar to Apache Iceberg, which is a Log Structured Table and need to compact/recluster/vacuum table regularly to merge small data chunks, sort them by the cluster key or vacuum unneeded branches.

Previously, those implementation rely on different drivers which add some complexity for infrastructure as we need to deploy and maintain additional service to trigger driver events.
Thus we need to introduce the Background Service, which allow to run databend as a background one-shot job or as a background daemon to run cron jobs, 
jobs may trigger table maintainence tasks, for example, automatic compaction/vacuum/recluster by cron based/ critiria based without additional maintainence needed

This implementation contain:
1. complete metasrv schema definition and background_job and background_tasks
2. apis to update and maintain background_job and background_task state on meta-service
3. simplified job scheduler implementation which support `one_shot`, `interval`, `cron` job type.

Not contain:
1. cron job graceful shutdown
2. tests on background job state machine
3. management apis or call procedures to view and create background jobs by user
4. Cluster mode background service, as we need a leader election process to ensure a leader to initialize requests

Closes #issue


src/meta/proto-conv/src/util.rs
```diff
@@ -71,10 +71,12 @@ const META_CHANGE_LOG: &[(u64, &str)] = &[
     (39, "2023-05-22: Add: data_mask.proto", ),
     (40, "2023-05-26: Add: TableMeta add column_mask_policy field", ),
     (41, "2023-05-29: Add: virtual_column.proto", ),
+    (42,"2023-06-03: Add allow_anonymous in S3 Config",),
+    (43,"2023-06-05: Add fields `number_of_segments` and `number_of_blocks` to TableStatistics",),
     (42, "2023-06-03: Add allow_anonymous in S3 Config", ),
     (43, "2023-06-05: Add fields `number_of_segments` and `number_of_blocks` to TableStatistics", ),
```
drmingdrmer, line 77: dup?

src/meta/api/src/background_api_impl.rs
```diff
@@ -0,0 +1,390 @@
...
+            );
+
+            {
+                let meta: BackgroundJobInfo = req.job_info.clone();
+                let condition = vec![txn_cond_seq(name_key, Eq, 0)];
+                let if_then = vec![
+                    txn_op_put(name_key, serialize_u64(id)?), // name -> background_job_id
+                    txn_op_put(&id_key, serialize_struct(&meta)?), // id -> meta
+                ];
+
```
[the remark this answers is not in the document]
ZhiHanZ: By design, BackgroundJob would be exposed to customer with the following SQL in the future
```sql
CREATE [ OR REPLACE ] TASK [ IF NOT EXISTS ] <name>
  [ { WAREHOUSE = <string> } | { USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = <string> } ]
  [ SCHEDULE = '{ <num> MINUTE | USING CRON <expr> <time_zone> }' ]
  [ <session_parameter> = <value> [ , <session_parameter> = <value> ... ] ]
[ WHEN <boolean_expr> ]
AS
  <sql>
```

thus I think we need a relavently complete API for it under metastore
Ref:
https://docs.snowflake.com/en/sql-reference/sql/create-task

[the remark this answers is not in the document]
drmingdrmer: It looks like a task can not be renamed, do we still have to keep the indirect mapping `name -> id -> value`?

https://docs.snowflake.com/en/sql-reference/sql/alter-task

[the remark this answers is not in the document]
ZhiHanZ: That make sense

src/meta/api/src/background_api_impl.rs
```diff
@@ -0,0 +1,368 @@
...
+                if_then,
+                else_then: vec![],
+            };
+            let (succ, _responses) = send_txn(self, txn_req).await?;
+            if succ {
+                break Ok(UpdateBackgroundTaskReply {
+                    last_updated: meta.last_updated.unwrap_or(Default::default()),
+                    expire_at: req.expire_at,
+                });
+            }
```
drmingdrmer, line 286: Is it necessary for the target to exist when performing an `update` operation?

And it looks updating an task is quite simple and can be done with a `KVApi::upsert_kv()`.  There is no need to run a transaction. Refer to a sample usage of `upsert_kv`:

https://github.com/datafuselabs/databend/blob/599d5b0841a2b17d59c265cc0ebe3e525b7f34ce/src/query/management/src/cluster/cluster_mgr.rs#L90-L93

Metadata operation that involves only one record, such as `delete_background_task()`, can always be done with `KVApi::upsert_kv()`, for example:
```rust
self.upsert_kv(UpsertKVReq::new(&key, seq, Operation::Delete, None))
```

src/meta/api/src/background_api_impl.rs
```diff
@@ -0,0 +1,368 @@
...
+                    BackgroundJobId { id }
+                }
+                _ => BackgroundJobId { id },
+            };
+            let meta = req.info.clone();
+            let condition = vec![txn_cond_seq(name_key, Eq, seq)];
+            let if_then = vec![
+                txn_op_put(name_key, serialize_u64(id_key.id)?),
+                txn_op_put(&id_key, serialize_struct(&meta)?),
+            ];
```
drmingdrmer, line 174: For a indirect mapping `name -> id -> value`, an update operation handles the `id -> value` part, thus it's unnecessary to assert the seq of `name`(`txn_cond_seq(name_key, Eq, seq)`), but instead, it should assert that the seq of `id -> value` does not change during running this transaction.

src/meta/api/src/background_api_impl.rs
ZhiHanZ: It is not necessary, since it assume to be a simple operation mainly used for audit purpose

src/meta/api/src/background_api_impl.rs
```diff
@@ -0,0 +1,363 @@
...
+        if resp.job_params == Some(req.params.clone()) {
+            return Ok(UpdateBackgroundJobReply { id });
+        }
+        resp.job_params = Some(req.params);
+        resp.last_updated = Some(Utc::now());
+        let req = UpdateBackgroundJobReq {
+            job_name: name.clone(),
+            info: resp,
+        };
+        self.update_background_job(req).await
```
drmingdrmer, line 226: Is this a non-conditional update? i.e., the state after an update does not have to depend on the state before the update.

Implementing update in this way will get an update lost:
```
initial state    process B           process C
A              
|
|    read
+-------------------------------------> A
|                                       | local update
|    read                               | line 220-225
+----------------> A                    |
|                  | local update       | 
|                  | line 220-225       |
|    write         v                    |
B <--------------- B                    |
|                                       |
|                                       |
|                            write      v
C <-------------------------------------C
```

If this is acceptable, there is no need to use transaction. The simplified way is:

- Load the `name->id` record and the `id->value` record.

- Update the `id->value` record with the new value using the `upsert_kv()`. Set the `UpsertKV.seq` field to `MatchSeq::GE(0)` to update any existing record with a sequence number greater than or equal to 0. This will not create a new record if the `id->value` does not exist.

Such a way is feasible because renaming does not affect the `id->value` record, but only affect the `name->id` record. Therefore we only need to forbid a delete operation during updating, with `MatchSeq::GE(0)`.

src/meta/app/src/background/background_job.rs
```diff
@@ -0,0 +1,417 @@
...
+
+impl Display for BackgroundJobState {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            BackgroundJobState::RUNNING => write!(f, "RUNNING"),
+            BackgroundJobState::FAILED => write!(f, "FAILED"),
+            BackgroundJobState::SUSPENDED => write!(f, "SUSPENDED"),
+        }
+    }
+}
```
drmingdrmer, line 52: It's not a problem at all but such implementation can be written as
```suggestion
impl Display for BackgroundJobState {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        write!(f, "{:?}", self)
    }
}
```

Because the `Debug` formatter just outputs the name of the variant.

Or you can just use crate `derive_more`:
```rust
#[derive(derive_more::Display)]
enum BackgroundJobState {
...
}
```

drmingdrmer (approved):
LGTM

PS display implementation can be simplified

drmingdrmer (approved):

PR #11755, opened 2023-06-14
fix: copy duplicate data when split large block.

youngsofun:

## Summary

bug introduced in
https://github.com/datafuselabs/databend/pull/11123
 https://github.com/datafuselabs/databend/releases/tag/v1.1.7-nightly

It is my fault that the feature of splitting large blocks when copy is not well-tested.

the bug only take effect when:

1. copy CSV with a large `input_read_buffer_size` set (default 1MB).
2. copy parquet when row_group is very large (no common).

Closes #issue


sundy-li (approved):

Xuanwo (approved):

youngsofun:
clippy fixed

zhyass (approved):

PR #11759, opened 2023-06-14
refactor: optimize pruning phase of the  delete statement

dantengsky:

## Summary

During the execution of the delete statement,  after pruning,  there might be blocks that could be further pruned, by using rang index only, and the data of those blocks need not be loaded in the subsequent processing phases.

for example:

`delete from t where t.c > xxx and t.c < yyy`

there might be blocks of table `t` that have min/max range `[a,b]` of column `c`, where `a >= xxx and b <= yyy`,  they could be safely marked as whole-block-deletion,  and we do not need to load the data of column `c` and filter the rows.

in this PR, a new pruning phase is introduced:

after the primary pruning, we got the candidate blocks to perform the deletion, these blocks may be deleted totally or partially, to filter out the blocks that could be deleted totally by using range index only:
       
  1. invert the filter expression
  2. apply the inverse filter expression to the block metas, using the range index
     - for those blocks that need to be deleted totally, they will be filtered out.
     -  blocks that need to be deleted partially will NOT be filtered out.
        
for the blocks marked as whole-block-deletion in the above pruning phase,  we just "remove" them from segments, no need to load data of column `c` and filter rows.

-----------

the above optimization is only applied at the block-meta level in this PR. segment-level optimization will be addressed in another PR.

Closes #issue


sundy-li (approved):

dantengsky:
reminder: 
the above optimization is only applied at the block-meta level in this PR. segment-level optimization will be addressed in another PR.

PR #11763, opened 2023-06-15
feat: support subquery in update statement

xudong963:

## Summary

Support subquery in update's selection position.

Part of https://github.com/datafuselabs/databend/issues/9455

### DIFF
--- a/src/query/service/src/interpreters/interpreter_update.rs
+++ b/src/query/service/src/interpreters/interpreter_update.rs
@@ -12,15 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashSet;
+use std::collections::VecDeque;
 use std::sync::Arc;
 
 use common_base::runtime::GlobalIORuntime;
+use common_exception::ErrorCode;
 use common_exception::Result;
+use common_expression::types::DataType;
+use common_expression::types::NumberDataType;
+use common_expression::ROW_ID_COL_NAME;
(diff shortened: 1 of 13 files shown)


src/query/service/src/interpreters/interpreter_update.rs
```diff
@@ -74,14 +84,65 @@ impl Interpreter for UpdateInterpreter {
             .get_table(self.ctx.get_tenant().as_str(), db_name, tbl_name)
             .await?;
 
-        let (filter, col_indices) = if let Some(scalar) = &self.plan.selection {
+        let selection = if !self.plan.subquery_desc.is_empty() {
+            let support_row_id = tbl.support_row_id_column();
+            if !support_row_id {
+                return Err(ErrorCode::from_string(
+                    "table doesn't support row_id, so it can't use delete with subquery"
```
zhyass, line 91:
```suggestion
                    "table doesn't support row_id, so it can't use update with subquery"
```

src/query/storages/fuse/src/operations/mutation/mutation_source.rs
```diff
@@ -261,6 +261,8 @@ impl Processor for MutationSource {
                                 }
                             }
                             MutationAction::Update => {
+                                // Pop the row_id column
```
zhyass, line 264: Is it necessary to add conditions here.
```
if self.query_row_id_col {
    data_block = data_block.pop_columns(1)?;
}
```

PR #11768, opened 2023-06-15
fix(query): to_date(timestamp) result is equal with timestamp date part

### LINKED ISSUE #11766: to_date  returned value is wrong
```
| timezone                              | Asia/Shanghai | UTC          | GLOBAL  | Sets the timezone.

mysql> select to_timestamp(1512168413::int) ts, to_date(ts);
+----------------------------+-------------+
| ts                         | to_date(ts) |
+----------------------------+-------------+
| 2017-12-02 06:46:53.000000 | 2017-12-01  |
+----------------------------+-------------+
1 row in set (0.01 sec)
Read 1 rows, 1.00 B in 0.001 sec., 803.39 rows/sec., 803.39 B/sec.

mysql> select '2017-12-02 06:46:53.000000'::timestamp ts, to_date(ts);
+----------------------------+-------------+
| ts                         | to_date(ts) |
+----------------------------+-------------+
| 2017-12-02 06:46:53.000000 | 2017-12-01  |
+----------------------------+-------------+
1 row in set (0.00 sec)
Read 1 rows, 1.00 B in 0.001 sec., 1.16 thousand rows/sec., 1.13 KiB/sec.
```

The following SQL calculation result is wrong.
```
select to_date(datetime) as day,
       sum(case when behavior_type = 'pv' then 1 else 0 end) as pv,
       count(distinct user_id) as uv
from user_behavior
group by day
order by day;
```

TCeason:

## Summary

to_date(timestamp) result is equal with timestamp date part

Closes #11766


TCeason:
Now ValueRef::Column is ok, ValueRef::Scalar has some mistakes.

src/query/functions/src/scalars/datetime.rs
```diff
@@ -339,8 +338,19 @@ fn register_timestamp_to_date(registry: &mut FunctionRegistry) {
...
+                .to_timestamp(tz)
+                .naive_local()
+                .signed_duration_since(
+                    NaiveDate::from_ymd_opt(1970, 1, 1)
+                        .unwrap()
+                        .and_hms_opt(0, 0, 0)
+                        .unwrap(),
+                )
+                .num_days() as i32;
+            output.push(res);
```
RinChanNOWWW, line 353: I think the logic here is right. We only consider the time zone when display.

TCeason: For a timestamp expression, the date from the timestamp. So we need  check tz when to_date(timestamp).

https://docs.snowflake.com/en/sql-reference/functions/to_date

If not , The following SQL calculation result is wrong.
```
set timezone='Asia/Shanghai';
select to_date(datetime) as day,
       sum(case when behavior_type = 'pv' then 1 else 0 end) as pv,
       count(distinct user_id) as uv
from user_behavior
group by day
order by day;
```

RinChanNOWWW (approved):

src/query/functions/src/scalars/datetime.rs
```diff
@@ -239,7 +238,28 @@ fn register_date_to_timestamp(registry: &mut FunctionRegistry) {
     ) -> Value<TimestampType> {
         vectorize_with_builder_1_arg::<DateType, TimestampType>(|val, output, _| {
             let ts = (val as i64) * 24 * 3600 * MICROS_IN_A_SEC;
-            output.push(ts);
+            let tz = ctx.func_ctx.tz.tz;
+            let ts = ts.to_timestamp(tz);
```
sundy-li, line 242: why not use `ts.day()` cause it's already implement trait `Datelike`

TCeason: The day() function only return the day-th in the month.

```rust
use chrono::{NaiveDate, NaiveDateTime, Datelike};

let dt: NaiveDateTime = NaiveDate::from_ymd_opt(2015, 9, 25).unwrap().and_hms_opt(12, 34, 56).unwrap();
assert_eq!(dt.day(), 25);
```

src/query/functions/src/scalars/datetime.rs
```diff
@@ -275,7 +295,20 @@ fn register_number_to_timestamp(registry: &mut FunctionRegistry) {
     ) -> Value<TimestampType> {
         vectorize_with_builder_1_arg::<Int64Type, TimestampType>(|val, output, ctx| {
             match int64_to_timestamp(val) {
-                Ok(ts) => output.push(ts),
+                Ok(ts) => {
+                    let res = ts
+                        .to_timestamp(ctx.func_ctx.tz.tz)
```
sundy-li, line 300: don't need to consider of tz for  `to_timestamp(int64)`, we just store the timestamp micros.

tests/sqllogictests/suites/query/02_function/02_0012_function_datetimes_tz
```diff
@@ -22,7 +22,7 @@ set timezone='Asia/Shanghai'
 query T
 select to_timestamp(1630320462000000)
 ----
-2021-08-30 18:47:42.000000
+2021-08-31 02:47:42.000000
```
sundy-li, line 25: this is wrong!

src/query/functions/src/scalars/datetime.rs
TCeason: Ok.

tests/sqllogictests/suites/query/02_function/02_0012_function_datetimes_tz
TCeason: How about support this as TO_TIMESTAMP_LTZ?

If to_timestamp(int) consider tz, I think this is TO_TIMESTAMP_LTZ in sf.

sundy-li: micro-seconds has no business with Time Zone.

src/query/functions/src/scalars/datetime.rs
TCeason: Fix

tests/sqllogictests/suites/query/02_function/02_0012_function_datetimes_tz
TCeason: Fix

src/query/functions/src/scalars/datetime.rs
TCeason: I think we can resolve this conversation.

sundy-li (approved):

andylokandy (approved):

PR #11776, opened 2023-06-16
fix: refactor and fix internal columns binding.

### LINKED ISSUE #11772: Bug: panic when select internal columns in multi-table query
**Summary**

```sql
select a,b,c,_block_name from t inner join t2 on t.a = t2.a;
APIError: ResponseError with 500: internal server error
```

This should not panic but return an error.

RinChanNOWWW:

## Summary

- Do not panic but throw an error when selecting internal columns in multi-table queries. And skip tables bound from aggregating index when checking the table count.
- Refactor and fix the index generation logic for internal columns. The bound index should come from `metadata.columns.len()` but not `bind_context.columns.len()`

Closes #11772


b41sh (approved):

sundy-li (approved):

RinChanNOWWW:
@Mergifyio refresh

RinChanNOWWW:
@mergify update

PR #11777, opened 2023-06-16
feat: introduce distributed deletion

SkyFan2002:

## Summary
Generally, we already have excellent [infrastructure](https://github.com/datafuselabs/databend/blob/27b37758ed66979e388dc8ff9031971d6c198028/src/query/service/src/schedulers/scheduler.rs#L84-L103) to convert a physical plan with `Exchange` into `Fragment`s, and do all the distributed work for us, so all we need to do to scale a DML to multiple nodes is to extract a physical plan tree from our pipeline.

Even better, the boundaries of our [pipeline](https://github.com/datafuselabs/databend/blob/27b37758ed66979e388dc8ff9031971d6c198028/src/query/storages/fuse/src/operations/delete.rs#L68-L75) are very clear, which makes it easier to abstract the operator.

This PR introduces distributed delete:
1. Perform a fast path of delete:(e.g. unconditional deletion, this is fast and doesn't need to be distributed)
2. Do block pruning, and get `partitions` that need to be processed by pipeline
3. Build physical plan, `DeletePlan` is divided into `DeletePartial->Exchange(Merge)->DeleteFinal`
4. Build pipeline, `DeletePartial` performs most of the delete work, whilch will be sent to multi nodes,  `DeleteFinal` commits meta, which is performed at local node.Note: `TableMutationAggregator` can be  moved from `DeleteFinal` to `DeletePartial`, which is planned to introduce in next PR.


zhang2014:
Maybe need add some logic test?

src/query/service/src/api/rpc/exchange/exchange_manager.rs
```diff
@@ -197,8 +198,12 @@ impl DataExchangeManager {
         // TODO: When the query is not executed for a long time after submission, we need to remove it
         match queries_coordinator.get_mut(&packet.query_id) {
             None => Err(ErrorCode::Internal(format!(
-                "Query {} not found in cluster.",
-                packet.query_id
+                "Query {} not found in cluster,invalid Query ids: {:?}.",
```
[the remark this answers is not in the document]
SkyFan2002: OK

BohuTANG:
Cool, please complete the PR summary, like : what's the PR did, how we did it, more is better.

src/query/service/src/schedulers/fragments/fragmenter.rs
```diff
@@ -41,7 +41,13 @@ pub struct Fragmenter {
     query_id: String,
 
     /// A state to track if is visiting a source pipeline.
-    visiting_source_pipeline: bool,
+    state: State,
+}
+
+enum State {
```
BohuTANG, line 47: Some comments need here

src/query/sql/src/executor/physical_plan.rs
```diff
@@ -808,6 +847,8 @@ impl PhysicalPlan {
             PhysicalPlan::ExchangeSink(_) => "Exchange Sink".to_string(),
             PhysicalPlan::ProjectSet(_) => "Unnest".to_string(),
             PhysicalPlan::RuntimeFilterSource(_) => "RuntimeFilterSource".to_string(),
+            PhysicalPlan::DeletePartial(_) => "DeletePartial".to_string(),
+            PhysicalPlan::DeleteFinal(_) => "DistributedFinal".to_string(),
```
BohuTANG, line 851: Why 'DistributedFinal'?

SkyFan2002: oh my god, it's a typo. Thanks!

BohuTANG:
The biggest challenge for this PR is that how can we granted the logic is correct, I think we should add more logic testing(Although we have a sql logic test here, but it's very simple nearly no grantees for this PR logic).

SkyFan2002:
> The biggest challenge for this PR is that how can we granted the logic is correct, I think we should add more logic testing(Although we have a sql logic test here, but it's very simple nearly no grantees for this PR logic).

That's also I'm worrying about. I will try. Do we already have any examples similar to tip 2?

BohuTANG:
> That's also I'm worrying about. I will try. Do we already have any examples similar to tip 2?

I think @dantengsky may has some suggestions, if we support `replace into` to distributation in future, we also need a method to check too.

dantengsky:
> > That's also I'm worrying about. I will try. Do we already have any examples similar to tip 2?
> 
> I think @dantengsky may has some suggestions, if we support `replace into` to distributation in future, we also need a method to check too.
> 
> If i test i will:
> 
> 1. make a table with some data
> 2. delete data with standalone mode and `sum()` as `result1`
> 3. delete data with cluster mode(make sure it's a distributed delete) and `sum()` as `result2`
> 4. make sure `result1` is equal `result2`

:+1:  a scratch, hope it helps

~~~

-- make sure there will be multiple blocks there, by shrink the `row_per_block` 
create table t (id int, c1 int, c2 int) row_per_block=10;

-- generate test data
insert into t select number, number * 10, number * 5 from numbers(1000);

-- generate the ids of rows to be deleted
create table del_id (id int) as select cast(FLOOR(0 + RAND() * 1000), int) from numbers(100);

-- "backup" t
create table t_origin as select * from t;

-- do the deletion (in distributed settings)
delete from t where id in (select id from del_id);

-- check the sum of columns

select sum(id) from (select id from t_origin where id not in (select id from del_id));
select sum(id) from t;

select sum(c1) from (select id from t_origin where id not in (select id from del_id));
select sum(c1) from t;

select sum(c2) from (select id from t_origin where id not in (select id from del_id));
select sum(c2) from t;



~~~

tests/sqllogictests/suites/mode/cluster/distributed_delete.sql
```diff
@@ -0,0 +1,37 @@
...
+statement ok
+drop table if exists t_origin;
+
+# make sure there will be multiple blocks there, by shrink the `row_per_block`
+statement ok
+create table t (id int, c1 int, c2 int) row_per_block=10;
+
+# generate test data
+statement ok
+insert into t select number, number * 10, number * 5 from numbers(1000);
```
dantengsky, line 13: :+1: 

how about splitting this insertion into three?  so that table of multiple segments could be convered.


~~~
insert into t select number, number * 10, number * 5 from numbers(500) where number ;
insert into t select number, number * 10, number * 5 from numbers(1000) where number > 499;
insert into t select number, number * 10, number * 5 from numbers(1500) where number > 1499;
~~~

dantengsky (approved):

dantengsky:
@BohuTANG 

with the newly add sqllogic test case 

[tests/sqllogictests/suites/mode/cluster/distributed_delete.sql](https://github.com/datafuselabs/databend/pull/11777/files/8b38339d5db6a6a9ca363bf2280bf036a41cc7bc#diff-898c866ebb6ce89a237d7d8f7973779ff8c4f1c0c1d28f4d03939e14f937f479)

I think this PR is ready to merge as a new feature of nightly release. Meanwhile, I will try to add some extra test cases of this feature in the upcoming "long-run" test suites.

src/query/service/src/pipelines/pipeline_builder.rs
```diff
@@ -192,10 +197,62 @@ impl PipelineBuilder {
...
     }
 
+    /// The flow of Pipeline that build_delete_partial and build_delete_final build is as follows:
+    /// +---------------+      +-----------------------+
+    /// |MutationSource1| ---> |SerializeDataTransform1|   ------
+    /// +---------------+      +-----------------------+         |      +-----------------------+      +----------+
+    /// |     ...       | ---> |          ...          |   ...   | ---> |TableMutationAggregator| ---> |CommitSink|
+    /// +---------------+      +-----------------------+         |      +-----------------------+      +----------+
+    /// |MutationSourceN| ---> |SerializeDataTransformN|   ------
+    /// +---------------+      +-----------------------+
```
dantengsky, line 213: I think we should adjust this doc, `TableMutationAggreator` and `CommitSink` are not "chained" at this phase?

zhyass (approved):

PR #11782, opened 2023-06-17
fix: Config not loaded after bump serde

Xuanwo:

## Summary

`alias` is not compatible with `flatten`, causing our config loading to break after updating serde. Pinning the serde version to `152` is not a viable solution as it would also affect many other projects that depend on serde.

This pull request removes all field aliases and instead inserts deprecated fields that will generate errors for users attempting to load the config.


Xuanwo:
sqllogic_standalone_ydb_parquet is flaky, seems not related to this PR.

PsiACE (approved):

PR #11785, opened 2023-06-17
feat: Integrate with icelake for iceberg support

Xuanwo:

## Summary

This PR will integrate databend with icelake for iceberg support.

IceLake has implement all features that `iceberg_rs` have, our next step is to make query works.


PsiACE (approved):
🚀

PR #11806, opened 2023-06-20
feat: add rocksdb Disk cache

lichuang:

## Summary

1. introduce new cache type `rockdb` and `rockdbdisk`, while `rockdb` save cache value into rocksdb, and `rockdbdisk` save only cache meta into rocksdb but value into disk.
2. upgrade disk cache version from `v1` to `v2`.

Closes #issue


src/query/storages/common/cache/src/providers/rocksdb_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+// map key -> value
+static KEY2VALUE_COLUMN_PREFIX: &str = "_key2value";
+
+pub struct RocksDbCache {
+    db: rocksdb::TransactionDB,
+    limit: i64,
+}
+
+impl RocksDbCache {
+    pub fn new(path: &str, limit: i64) -> Result<RocksDbCache> {
```
BohuTANG, line 35: If we remove the whole path, the rocksdb is work next time?

lichuang: yes, rocksdb will recreate the whole data dir.

BohuTANG:
@dantengsky @lichuang 

BTW, The benchmark CI above is not enable the cache, FYI.

dantengsky:
> @dantengsky @lichuang
> 
> BTW, The benchmark CI above is not enable the cache, FYI.

got it

PR #11815, opened 2023-06-20
feat: fine tune for background compaction 

ZhiHanZ:

## Summary

1. background compaction would not happend if table compacted after the last update time
3. add updated_on field for system.table and system.background_tasks.

Closes #issue

### DIFF
--- a/src/query/ee/src/background_service/compaction_job.rs
+++ b/src/query/ee/src/background_service/compaction_job.rs
@@ -57,11 +57,21 @@ WHERE t.database != 'system'
     AND t.database != 'information_schema'
     AND t.engine = 'FUSE'
     AND t.num_rows > 1 * 1000 * 1000
-    AND t.data_compressed_size > 1 * 1000 * 1000 * 1000
+    AND t.table_id NOT IN (
+        SELECT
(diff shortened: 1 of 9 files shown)


ZhiHanZ:
cc @TCeason , this PR changed the create_on and drop_on field of system.tables into timestamp, PTAL if you have time, since I am not sure about whether current mysql integrations have special concerns on those fields

TCeason:
> cc @TCeason , this PR changed the create_on and drop_on field of system.tables into timestamp, PTAL if you have time, since I am not sure about whether current mysql integrations have special concerns on those fields

I think this is ok. 

https://github.com/datafuselabs/databend/blob/0018dced004a8bc20e91716ea80747dd2b405d7d/src/query/storages/information-schema/src/tables_table.rs#L61

Now mysql intergration only will access information_schema. And only use tables.create_on as information_schema.tables.create_time.

In MySQL this field `information_schema.tables.create_time` type is timestamp.

```
// | CREATE_TIME     | timestamp                                                          | NO   |     | NULL    |       |
```

TCeason (approved):

PR #11816, opened 2023-06-20
feat: enforce copy CSV on_error=continue 

youngsofun:
skip rows with wrong num of fields.

cc @wubx 

## Summary

Summary about this PR

Closes #issue


TCeason (approved):

ariesdevil (approved):
BTW: If we change the default behavior, we should also change the docs. cc: @soyeric128

PR #11818, opened 2023-06-20
feat(query): Disable arbitrary hostnames when creating users

### LINKED ISSUE #11771: Feature: Create/Alter/Drop User don't supported specifying host
**Summary**

Currently, we use `username@hostname` as the user identity, which can lead to multiple users and passwords for the same username. To simplify use, don't support specifying a host when creating a user, use `%` as the default host.

b41sh:

## Summary

This PR changes the following:
1. Disable specify arbitrary ip as hostname, only `%` can be used. Users previously created with an IP address as hostname will not work. Those users need to be dropped or altered before the upgrade.
2. `root` and `default` users will not be auto-generated, it is recommended to specify them via `IDM` configuration.
3. `root` and `default` users do not check if the login IP is `127.0.0.1` or `localhost`, please set a secure password to ensure security.
4. `IDM` config users are given the admin role after logging in.
5. `IDM` config users cannot create, update, or drop, and cannot modify roles or privileges.
6. Add `root` and `default` users with `no_password` to the config file for testing.

Closes #11771


BohuTANG:
@ZhiHanZ @flaneur2020 PTAL, make sure this is no affect the DatabendCloud users.

src/query/users/src/user_mgr.rs
```diff
@@ -43,7 +48,7 @@ impl UserApiProvider {
                     UserInfo::new_no_auth(&user.username, &user.hostname)
                 }
             };
-            if user.is_localhost() {
+            if client_ip == "127.0.0.1" || client_ip == "localhost" {
```
sundy-li, line 51: client_ip could be removed.

b41sh: allow root user login from any client ip?

src/query/users/src/user_mgr.rs
```diff
@@ -43,24 +43,17 @@ impl UserApiProvider {
...
-                    .option
-                    .set_default_role(Some(BUILTIN_ROLE_ACCOUNT_ADMIN.to_string()));
-                user_info.option.set_all_flag();
-            } else {
-                return Err(ErrorCode::UnknownUser(format!(
-                    "only accept root from localhost, current: '{}'@'{}'",
-                    user.username, user.hostname
-                )));
-            }
+            user_info.grants.grant_privileges(
```
flaneur2020, line 46: I guess checking the root user's original IP to be localhost is still a must have, currently we restricted the root access by localhost only. If this behaviour changed, it might leaks some security risks.

tests/sqllogictests/suites/base/05_ddl/05_0004_ddl_create_user
```diff
@@ -1,53 +1,59 @@
...
+DROP USER IF EXISTS 'test-d'
 
 statement ok
 DROP USER IF EXISTS 'test-e'
 
+statement error 1005
+CREATE USER 'test-a'@'127.0.0.1' IDENTIFIED BY 'password'
+
+statement error 1005
+CREATE USER 'test-a'@'%' IDENTIFIED BY 'password'
```
flaneur2020, line 20: I guess keeping this statement compatibility is ok? just allow '%' and returns error on other values

src/query/ast/src/parser/statement.rs
```diff
@@ -2150,13 +2150,11 @@ pub fn user_option(i: Input) -> IResult<UserOptionItem> {
 pub fn user_identity(i: Input) -> IResult<UserIdentity> {
     map(
         rule! {
-            #parameter_to_string ~ ( "@" ~ #literal_string )?
```
flaneur2020, line 2153: how about `#parameter_to_string ~ ( "@" ~  "'%'" )?` ? 

so this may compatbile with the existed `'user'@'%'` queries, but do not allow any other host values.

sundy-li (approved):

flaneur2020 (approved):

PR #11849, opened 2023-06-25
fix(query): dataframe sort api need bind projection

TCeason:

## Summary

dataframe sort api need bind projection

Closes #issue


JackTan25:
when to merge? I need to resolve test_frame bug in my pr? @TCeason @sundy-li

PR #11913, opened 2023-06-29
feat(query): Reducing the memory used by the SRF function

### LINKED ISSUE #11912: Feature: improve srf function performance
**Summary**

Use less memory to prevent OOM

b41sh:

## Summary

FlatMap ignores unused columns, prevents OOM due to repeating the rows of the original column.

- Closes #11912

### DIFF
--- a/tests/sqllogictests/suites/query/02_function/02_0051_function_semi_structureds_get
+++ b/tests/sqllogictests/suites/query/02_function/02_0051_function_semi_structureds_get
@@ -301,9 +301,9 @@ select id, get(arr, 'a') from t4
 query ITS
 select id, json_path_query(arr, '$[2, 1 to last -1]'), typeof(json_path_query(arr, '$[2, 1 to last -1]')) from t1
 ----
-1 3 VARIANT
-1 2 VARIANT
-1 3 VARIANT
+1 3 VARIANT NULL
+1 2 VARIANT NULL
+1 3 VARIANT NULL
 
 query IT
 select id, json_path_query(arr, '$[*]?(@ > 1 && @ <= 3)') from t1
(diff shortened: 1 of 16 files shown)


leiysky:
askbend:summary

tests/sqllogictests/suites/query/02_function/02_0051_function_semi_structureds_get
```diff
@@ -301,9 +301,9 @@ select id, get(arr, 'a') from t4
 query ITS
 select id, json_path_query(arr, '$[2, 1 to last -1]'), typeof(json_path_query(arr, '$[2, 1 to last -1]')) from t1
 ----
-1 3 VARIANT
-1 2 VARIANT
-1 3 VARIANT
+1 3 VARIANT NULL
```
leiysky, line 304: Is this the expected change of type?

leiysky (approved):

tests/sqllogictests/suites/query/02_function/02_0051_function_semi_structureds_get
b41sh: yes, the type should be nullable.

drmingdrmer:
### PR summary(by Claude):

 Commit message for patch:

This patch makes the following changes:

1. Ignore NULL parameters when evaluating json_path_query(). Previously it would error out.

2. Avoid unnecessarily unnesting unused columns when evaluating FlatMap ProjectSet. This improves performance by not increasing the cardinality of rows.

3. Fix unit test cases related to these changes.

4. Add better explain info for ProjectSet, showing unused columns and actual read bytes instead of estimated rows.

5. Add heuristic to prune unused output columns during optimization, which can then be used during FlatMap evaluation to avoid unnesting.

This should improve performance when unnesting SRF functions over unused columns, while keeping backwards compatibility.

sundy-li (approved):

PR #11922, opened 2023-06-29
chore(planner): improve selectivity

Dousir9:

## Summary

Summary about this PR
Improve selectivity, we will get better join order on TPC-H(SF=100) Q19.

- Closes #issue


drmingdrmer:
### PR summary(by Claude):

 modify git commit message for following patch

Fix selectivity calculation for boolean expressions

This patch fixes an issue where the selectivity of AND / OR boolean expressions were calculated incorrectly. Previously, the selectivity of AND expressions was calculated as the product of the selectivities of the operands, and the selectivity of OR expressions was calculated as the sum of the selectivities. 

This patch changes the selectivity calculation to the following:

- For AND expressions, the selectivity is calculated as the minimum of the selectivities of the operands. This results in a more conservative estimate.
- For OR expressions, the selectivity is calculated as the maximum of the selectivities of the operands.

This has the following effects:

- Join orders where filter predicates have boolean expressions are more accurate. 
- Filter plans have more accurate estimated rows.
- Selectivity updates to column statistics are more accurate.

The testsuite show significant improvements in estimated rows for explain plans with boolean expressions.

This fixes issues where join orders and plans were suboptimal due to incorrect selectivity estimates.

Fixes #123

diff --git a/src/query/sql/src/planner/optimizer/property/selectivity.rs b/src/query/sql/src/planner/optimizer/property/selectivity.rs
index 53795ead623..8224abe53a6 100644

xudong963 (approved):

PR #11924, opened 2023-06-29
fix: improve sort and fix index scan with sort.

RinChanNOWWW:

## Summary

- Introduce projection operator before building sort pipeline. This help avoid useless data copying during sorting.
- Add a setting to control if enable reading from aggregating index.
- Disable reading aggregating index for queries with `order by`.


RinChanNOWWW:
askbend:summary

drmingdrmer:
### PR summary(by Claude):


```

drmingdrmer:
### PR summary(by Claude):

 Here is the summary of main changes made in the pull request:

* **Optimized sort plans for memory efficiency**
   Added a pre-projection optimization before sorting to reduce data copying for irrelevant columns. This can improve performance and memory usage for queries with large amount of columns but requiring a subset of columns after sorting.

* **Enabling aggregating index scan optimization**
  Added a new setting `enable_aggregating_index_scan` to enable scanning aggregating index data. Conditionally enabled/disabled this optimization during planning.

* **Disabled aggregating index scan for sort queries**
   Disabled aggregating index scan optimization if the query has an `ORDER BY` clause to avoid incompatibility between sort and aggregating index.

* **Lazy reading optimization**
  Added a lazy reading optimization to avoid pre-projection for sort queries if the query will be optimized by lazy column reading. This can reduce optimization overhead.

* **More test cases**
   Added more test cases for new features, including the explain result of sort plans with/without pre-projection and the effect of new settings.

sundy-li (approved):

ariesdevil (approved):

PR #11928, opened 2023-06-30
chore: test performance

xudong963:

## Summary

Summary about this PR

- Closes #issue


drmingdrmer:
### PR summary(by Claude):

 Updates join.rs to slightly optimize inner join planning

This patch updates join.rs to only consider broadcasting inner joins for optimization, skipping right/full joins. This optimizes inner join planning by focusing only on applicable joins.

diff --git a/src/query/sql/src/planner/plans/join.rs b/src/query/sql/src/planner/plans/join.rs
index 39e953cd6fd..8c824fa87ef 100644

drmingdrmer:
### PR summary(by Claude):

 Adds condition to enable broadcast join for right mark join type

This patch adds the JoinType::RightMark join type to the conditions that enable broadcast join when the prefer_broadcast_join setting is enabled. This allows right mark joins to take advantage of performance improvements from broadcast joins.

diff --git a/src/query/sql/src/planner/plans/join.rs b/src/query/sql/src/planner/plans/join.rs
index 39e953cd6fd..8c824fa87ef 100644

PR #11941, opened 2023-07-03
feat: support explain query result cache reading.

RinChanNOWWW:

## Summary

- If the query will read result cache, we can get this information by `explain`.
- We can `explain pipeline` for all plans now, as we can get the pipeline by `interpreter.execute2`.


RinChanNOWWW:
askbend:summary

sundy-li (approved):

PR #11942, opened 2023-07-03
feat: cume_dist window function

ariesdevil:

## Summary

Impl `cume_dist` window function.

Tracking: #11148

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
+++ b/src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
@@ -727,6 +727,8 @@ where T: Number + ResultTypeOfUnary
             false
         };
 
+        let need_peer = matches!(func, WindowFunctionImpl::CumeDist);
+
         Ok(Self {
             input,
             output,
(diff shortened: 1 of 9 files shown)


src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -707,6 +756,9 @@ where T: Number + ResultTypeOfUnary
             prev_frame_start: RowPtr::default(),
             prev_frame_end: RowPtr::default(),
             peer_group_start: RowPtr::default(),
+            peer_group_end: RowPtr::default(),
+            peer_group_ended: false,
+            need_peer: true,
```
RinChanNOWWW, line 761: Not all functions need peer.

ariesdevil: Theoretically yes, but currently only `cume_dist` is needed, I'll modify it.

RinChanNOWWW (approved):

PR #11963, opened 2023-07-04
feat: add arg in get_table_history to get all the drop tables

lichuang:

## Summary

feat: add arg in get_table_history to get all the drop tables

add `AllDroppedTables` option to get tables in drop db, and test cases.

- Closes #issue

### DIFF
--- a/src/meta/api/src/schema_api_impl.rs
+++ b/src/meta/api/src/schema_api_impl.rs
@@ -63,6 +63,7 @@ use common_meta_app::schema::DatabaseId;
 use common_meta_app::schema::DatabaseIdToName;
 use common_meta_app::schema::DatabaseIdent;
 use common_meta_app::schema::DatabaseInfo;
+use common_meta_app::schema::DatabaseInfoFilter;
 use common_meta_app::schema::DatabaseMeta;
 use common_meta_app::schema::DatabaseNameIdent;
 use common_meta_app::schema::DatabaseType;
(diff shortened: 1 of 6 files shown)


lichuang:
askbend:summary

BohuTANG:
## PR Summary(By [llmchain.rs](https://github.com/shafishlabs/llmchain.rs)):
* **Table Filtering**
The implementation of the `do_get_table_history` and `list_tables` functions were modified to filter tables based on the request filter. New conditional blocks were added to handle the case where the request filter is `AllDroppedTables`.

* **New Function and Test Cases**
A new function `table_history_filter` was added, along with two new test cases for `AllDroppedTables` with and without filter time.

* **New Enum and Struct Fields**
A new enum `DatabaseInfoFilter` with a variant `IncludeDropped` was added, and the `ListDatabaseReq` struct now has an optional field `filter` of type `Option<DatabaseInfoFilter>`. The `TableInfoFilter` enum was also modified to include new variants `AllDroppedTables` and `All`.

* **Default Value for ListDatabaseReq**
The `ListDatabaseReq` now has a default value of `None` for the `filter` field.

src/meta/api/src/schema_api_impl.rs
```diff
@@ -810,7 +811,9 @@ impl<KV: kvapi::KVApi<Error = MetaError>> SchemaApi for KV {
                             continue;
                         }
                         let db_meta = db_meta.unwrap();
-                        if is_drop_time_out_of_retention_time(&db_meta.drop_on, &now) {
+                        if req.filter.is_none()
+                            && is_drop_time_out_of_retention_time(&db_meta.drop_on, &now)
+                        {
```
drmingdrmer, line 816: Why to skip a db_meta instance when filter is none?

It would be better adding some comment explaining the purpose.

Or explicitly examining the content of `req.filter`, e.g.: `if req.filter == Some(...) `

drmingdrmer (approved):
I found that the schema api request such `req: ListTableReq` does not need `serde::Serialize` thus it is possible using a function as filter in the request.

```rust
pub struct ListTableReq {
    pub inner: DatabaseNameIdent,
    pub filter: Option<TableInfoFilter>,
}
```

PR #11964, opened 2023-07-04
feat(catalog): Add get_catalog support and schema api test

Xuanwo:

## Summary

feat(catalog): Add schema api test suite for create catalog


src/meta/api/src/schema_api_test_suite.rs
```diff
@@ -1290,6 +1299,36 @@ impl SchemaApiTestSuite {
...
+                    storage_params: Box::new(StorageParams::Azblob(StorageAzblobConfig {
+                        container: "container".to_string(),
+                        ..Default::default()
+                    })),
+                }),
+                created_on: Utc::now(),
+            },
+        };
+
+        let res = mt.create_catalog(req).await?;
```
b41sh, line 1326: Do we need to check res is success ?

src/meta/api/src/schema_api_test_suite.rs
```diff
@@ -1290,6 +1299,36 @@ impl SchemaApiTestSuite {
...
+        info!("create catalog res: {:?}", res);
```
drmingdrmer, line 1327: It would be better if a `get_catalog()` method was available to enable checking of the `create_catalog` result.

Xuanwo: added.

src/meta/api/src/schema_api_test_suite.rs
Xuanwo: Fixed.

drmingdrmer (approved):
👍

Xuanwo:
Hi, @b41sh, would you like to take another look? Thanks.

PR #11966, opened 2023-07-04
feat(query): improve like function

Dousir9:

## Summary

Summary about this PR
Improve like function, when we can't use `OrdinalStr`, `StartOfPercent`, `EndOfPercent` and `SurroundByPercent`, if the like expression only includes `%`, e.g. `'A%r%w'`, we will use `SimplePattern` to get better performance: TPC-H(SF=100) Q13: 8.6s -> 6.3s.

```sql
From TPC-H Q13:
select count(*) from orders where o_comment not like '%pending%deposits%';

TPC-H(SF=100): 4.03s -> 1.61s
```
ci-benchmark:
截屏2023-07-05 16 17 32

- Closes #issue


TCeason:
1. Has some invalid commit info. Better to reduce them
2. Could you provide  the performance  improvement

TCeason (approved):

BohuTANG:
We should add some unit tests or logic tests to make sure this PR works as expected and not broken by others in the future.

src/query/functions/src/scalars/comparison.rs
```diff
@@ -776,3 +802,178 @@ fn like(haystack: &[u8], pattern: &[u8]) -> bool {
...
+        }
+        segment_idx += 1;
+    }
+    true
+}
+
+#[test]
+fn test_check_pattern_type() {
+    // OrdinalStr
+    let pattern = "databend".as_bytes();
```
sundy-li, line 911:
```
let test_cases = vec![
    ("databend",  PatternType::OrdinalStr),
 .....
];

for (a, b) in test_cases {
  assert_eq!(
        b,
        check_pattern_type(a, false)
    );
}
```

sundy-li:
Do you have any stateless test cases which cover `SimplePattern ` pattern?

Dousir9:
> SimplePattern

Let me add some tests.

sundy-li (approved):

PR #11979, opened 2023-07-05
fix: wrong memory_size of StringColumn.

youngsofun:

## Summary

should not use `col.data.len()`, the `data` maybe shared(sliced).


src/query/expression/src/values.rs
```diff
@@ -1732,7 +1732,7 @@ impl Column {
             Column::Decimal(DecimalColumn::Decimal128(col, _)) => col.len() * 16,
             Column::Decimal(DecimalColumn::Decimal256(col, _)) => col.len() * 32,
             Column::Boolean(c) => c.as_slice().0.len(),
-            Column::String(col) => col.data.len() + col.offsets.len() * 8,
+            Column::String(col) => col.memory_size(),
```
b41sh, line 1735: The `Variant` type may also have a similar problem and can be changed together.

b41sh (approved):

PR #11984, opened 2023-07-05
fix: TSV support mysql text.

youngsofun:

## Summary

in Clickhouse TSV,  HIVE text,  PG text , `[newline]` in value  is escaped as `\n`
but in MySQL text output, it is escaped as `\[newline]` by default.
need to consider when cutting bytes to rows.

https://dev.mysql.com/doc/refman/8.0/en/select-into.html
https://dev.mysql.com/doc/refman/8.0/en/load-data.html

- Closes #issue


BohuTANG:
To make this Pull Request better, let's add some logic tests to test it works.

youngsofun:
```
sccache: error: Timed out waiting for server startup. Maybe the remote service is unreachable?
```

b41sh (approved):

PR #12000, opened 2023-07-06
chore(query): use append mode for spill writer

zhang2014:

## Summary

chore(query): use append mode for spill writer

- Closes #issue


Xuanwo (approved):
Great!

PR #12001, opened 2023-07-06
fix(meta-service): retry when forward-rpc fails

drmingdrmer:

##### fix(meta-service): retry when forward-rpc fails

When a request is forwarded from a follower to the leader,
it should retry if a `NetworkError` is occurred.
In the case when the leader crashed, internal forwarding should retry
until a new leader is established.

Add tests to ensure meta-grpc-client survives crashed
nodes(leader/follower) and internal rpc survives crashed leader too.

## Related Issues

### DIFF
--- a/src/meta/client/src/grpc_client.rs
+++ b/src/meta/client/src/grpc_client.rs
@@ -529,6 +529,8 @@ impl MetaGrpcClient {
             endpoints
         };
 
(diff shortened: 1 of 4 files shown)


drmingdrmer:
### PR summary(by Claude):


```

drmingdrmer:
### PR summary(by Claude):

 File changes done.

PR #12010, opened 2023-07-06
fix: add ignore_err in get_block_locations

zhyass:

## Summary

add ignore_err in get_block_locations to avoid missing blocks and bloom indexes

- Closes #issue

### DIFF
--- a/src/query/storages/fuse/src/operations/gc.rs
+++ b/src/query/storages/fuse/src/operations/gc.rs
@@ -275,8 +275,9 @@ impl FuseTable {
             Ok(v) => v,
         };
 
+        // root snapshot cannot ignore storage not find error.
         let referenced_locations = self
-            .get_block_locations(ctx.clone(), &root_snapshot.segments, put_cache)
+            .get_block_locations(ctx.clone(), &root_snapshot.segments, put_cache, false)
             .await?;
         let snapshot_lite = Arc::new(SnapshotLiteExtended {
             format_version: ver,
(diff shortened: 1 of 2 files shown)


src/query/storages/fuse/src/operations/gc.rs
```diff
@@ -343,7 +346,9 @@ impl FuseTable {
         let mut count = 0;
         let segment_locations = Vec::from_iter(segments_to_be_purged);
         for chunk in segment_locations.chunks(chunk_size) {
-            let locations = self.get_block_locations(ctx.clone(), chunk, false).await?;
+            let locations = self
```
dantengsky, line 349:
```suggestion
            // since we are purging files, the ErrorCode::STORAGE_NOT_FOUND error can be safely ignored.
            let locations = self
```

src/query/storages/fuse/src/operations/gc.rs
```diff
@@ -305,7 +306,9 @@ impl FuseTable {
         // Purge segments&blocks by chunk size
         let segment_locations = Vec::from_iter(segments_to_be_purged);
         for chunk in segment_locations.chunks(chunk_size) {
-            let locations = self.get_block_locations(ctx.clone(), chunk, false).await?;
+            let locations = self
```
dantengsky, line 309:
```suggestion
            // since we are purging files, the ErrorCode::STORAGE_NOT_FOUND error can be safely ignored.
            let locations = self
```

add a line of doc, hope I get it right.

dantengsky (approved):

PR #12011, opened 2023-07-06
refactor: using croaring instead of roaring-rs

ariesdevil:

## Summary

Using croaring instead of roaring-rs, with this PR, the performance is not related to roaring bitmap implementation.

After [this PR](https://github.com/RoaringBitmap/croaring-rs/pull/112) is merged, I'll replace the patched crate with the official crate.

- Closes #issue


Xuanwo:
> Using croaring instead of roaring-rs.

Hi, is there context for this change? Using a pure rust lib seems a better idea.

ariesdevil:
> Hi, is there context for this change? Using a pure rust lib seems a better idea.↳

The pure rust implementation is just a reference implementation, the performance gap is large, and it's no longer being maintained, only maintaining the c binding implementation.

ariesdevil:
@everpcpc I can build the python binding success on my local linux machine, do you have any ideas of the failed build in CI environment?

src/query/functions/src/scalars/bitmap.rs
```diff
@@ -48,8 +49,8 @@ pub fn register(registry: &mut FunctionRegistry) {
                     result
                 }) {
                 Ok(v) => {
-                    let rb = RoaringTreemap::from_iter(v.iter());
-                    rb.serialize_into(&mut builder.data).unwrap();
+                    let rb = Treemap::from_iter(v);
+                    builder.data = rb.serialize().unwrap();
```
b41sh, line 53: value should be pushed into the builder

everpcpc:
> @everpcpc I can build the python binding success on my local linux machine, do you have any ideas of the failed build in CI environment?

Seems like the gcc is too old in python build image?

b41sh:
@ariesdevil almost ok for me, please fix the ci

ariesdevil:
The failed python binding caused by [this reason](https://github.com/PyO3/maturin-action/issues/197)

sundy-li (approved):

b41sh (approved):

ariesdevil:
Many thanks @everpcpc for helping ci pass.

BohuTANG:
This breaking the env in some places, and we are considering reverting it, cc @ariesdevil

PR #12027, opened 2023-07-08
refactor: refactor copy into stage pipleine to avoid OOM and small row_groups

youngsofun:

## Summary

parquet pipeline:

- LimitFileSizeProcessor * 1:  slice/group block to batches (as a block meta) that are suitable as a file.
- ParquetFileSink * N: simply serialize blocks in each meta to a whole file and write out

in this impl,  big data block is not sliced when unload to parquet, I think it is uncommon, and not helpful to OOM.

CSV/TSV/NDJson pipeline:

- SerializeProcessor * N: serialize each data block to many small byte buffers.
- LimitFileSizeProcessor * 1:  group small byte buffers to batches (as a block meta) that are large enough as a file.
- RowBasedFileSink * N: simply concat small byte buffers to a whole and write out.

the key is lightweight LimitFileSizeProcessor 

add new test `0000_0000_copy_into_stage2.sh` to cover code paths


youngsofun:
cc @BohuTANG @wubx

wubx:
LGTM

BohuTANG (approved):

PR #12029, opened 2023-07-09
chore: correct the system settings

Dousir9:

## Summary

Summary about this PR

- Closes #issue


BohuTANG (approved):
👍

src/query/settings/src/settings_default.rs
```diff
@@ -162,7 +162,7 @@ impl DefaultSettings {
                 }),
                 ("max_execute_time", DefaultSettingValue {
                     value: UserSettingValue::UInt64(0),
-                    desc: "Sets the maximum query execution time in seconds. Setting it to 0 means no limit.",
+                    desc: "Sets the maximum query execution time in milliseconds. Setting it to 0 means no limit.",
```
BohuTANG, line 165: Oh, the code should make the setting*1000 to milliseconds, let's user configure as seconds is more easy.
The more reasonable setting name is 'max_execute_time_seconds'.

Dousir9: What if user need 500 milliseconds.

Dousir9: However, 500 milliseconds is very small, and users generally do not need this.

sundy-li (approved):

PR #12031, opened 2023-07-09
chore: refactor copy removing unused codes

BohuTANG:

## Summary

* remove some unused codes to make simple
* decouple local(`build_local_append_data_pipeline`) copy and distributed(`build_distributed_append_data_pipeline`) copy pipeline build

- Closes #issue


src/query/service/src/pipelines/builders/copy.rs
```diff
@@ -205,23 +252,14 @@ pub fn build_append_data_with_finish_pipeline(
                         // 2. Try to purge copied files if purge option is true, if error will skip.
                         // If a file is already copied(status with AlreadyCopied) we will try to purge them.
                         if purge {
-                            try_purge_files(ctx.clone(), &stage_info_clone, &files).await;
-                        }
-
-                        // Status.
-                        {
-                            info!("all copy finished, elapsed:{}", start.elapsed().as_secs());
```
JackTan25, line 213: I think we need debug info here.

BohuTANG: I will add some metrics for COPY in another PR, it will introduce more information and metrics to measure COPY(including distributed).

src/query/service/src/pipelines/builders/copy.rs
```diff
@@ -117,74 +87,151 @@ pub fn build_append_data_with_finish_pipeline(
...
+    }
+
+    // set finished callback to pipeline.
+    set_pipeline_finish_callback(
+        ctx,
+        main_pipeline,
+        coordinator_node_id,
+        stage_info,
+        vec![],
+        false,
```
JackTan25, line 210: should we get a purge flag here? not direct false?

BohuTANG: The distributed can not purge file, the purge will only do in the coordinator node, so here should false.

BohuTANG: The logic test is good to test my fault. Fixed.

src/query/service/src/pipelines/builders/copy.rs
```diff
@@ -117,74 +180,66 @@ pub fn build_append_data_with_finish_pipeline(
...
+                plan_required_values_schema,
+                AppendMode::Copy,
+            )?
         }
         CopyIntoTableMode::Replace => {}
     }
 
-    if local_id == ctx.get_cluster().local_id {
+    // set finished callback to pipeline.
+    set_pipeline_finish_callback(
```
JackTan25, line 212: please make sure the pruge will only happen on the source node. thanks!! It means only source need to do purge operations, remote node doesn't need, thanks. other is fine for me.

BohuTANG: The check is move to `set_pipeline_finish_callback` , we check here:

https://github.com/datafuselabs/databend/pull/12031/files#diff-8f4130db0584112576635553cf35d4760ea25d2e3740ba2de267c77ab913ce2eR233

PR #12032, opened 2023-07-09
chore: COPY unify

BohuTANG:

## Summary

This PR unify the COPY logic, suite for local or distributed two cases:
* COPY split into two phase: 1. append data to table 2. commit data to table
* The source node will do commit and some finish work like pruging

This refactor is an experiential improvement that will greatly benefit for the distributed `insert/replace` in future if need. 
You can based on this PR to further enhance cc @JackTan25 

- Closes #issue


src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -270,21 +272,43 @@ impl CopyInterpreter {
...
+            ctx.clone(),
             &mut build_res.main_pipeline,
-            plan.clone(),
+            CopyPlanType::CopyIntoTablePlanOption(plan.clone()),
             source_schema,
-            to_table,
-            files,
+            to_table.clone(),
         )?;
+        // if it's replace mode, don't commit
```
BohuTANG, line 282: why `replace` mode is special here?

JackTan25: replace into statement can use copy statement as its source, for this case, copy is just a data source. no need to commit

JackTan25 (commented):
LGTM

PR #12044, opened 2023-07-10
refactor: adjust calc of parallelism when load parquet.

youngsofun:

## Summary

@wubx hope this pr can further reduce the chance of OOM when loading parquet.

1. consider the memory already used.
2. small file can be as large as 100MB,  each ParquetSmallFilesPart should not be too large, let it be at most the largest of the normal  ParquetPart::RowGroup 
3. use compression ratio to estimate the uncompressed size of ParquetSmallFilesPart
4. consider the compressed size when calc num by memory limit,  there are cases the compression ratio is low.

### DIFF
--- a/src/query/storages/parquet/src/pruning.rs
+++ b/src/query/storages/parquet/src/pruning.rs
@@ -75,8 +75,8 @@ pub struct PartitionPruner {
     // /// Limit of this query. If there is order by and filter, it will not be used (assign to `usize::MAX`).
     // pub limit: usize,
     pub parquet_fast_read_bytes: usize,
-
     pub compression_ratio: f64,
+    pub max_memory_usage: u64,
 }
 
 fn check_parquet_schema(
@@ -249,7 +249,14 @@ impl PartitionPruner {
             }
             file_metas
         } else {
-            read_parquet_metas_in_parallel(operator.clone(), large_files.clone(), 16, 64).await?
+            read_parquet_metas_in_parallel(
+                operator.clone(),
+                large_files.clone(),
+                16,
+                64,
+                self.max_memory_usage,
+            )
+            .await?
         };
 
         // 2. Use file meta to prune row groups or pages.
--- a/src/query/storages/parquet/src/pruning.rs
+++ b/src/query/storages/parquet/src/pruning.rs
@@ -276,6 +276,8 @@ impl PartitionPruner {
             stats.read_rows += sub_stats.read_rows;
         }
 
+        let num_large_partitions = partitions.len();
+
         collect_small_file_parts(
             small_files,
             max_compression_ratio,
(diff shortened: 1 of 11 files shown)


b41sh (approved):

src/query/storages/parquet/src/pruning.rs
```diff
@@ -277,6 +276,45 @@ impl PartitionPruner {
             stats.read_rows += sub_stats.read_rows;
         }
 
+        if max_compression_ratio <= 0.0 || max_compression_ratio >= 1.0 {
+            max_compression_ratio = 1.0;
+        }
+        if max_compressed_size == 0 {
+            let max = 128usize >> 20;
```
RinChanNOWWW, line 283: `max` will always be 0.

sundy-li (approved):

src/query/storages/parquet/src/pruning.rs
```diff
@@ -277,6 +276,45 @@ impl PartitionPruner {
...
+                make_small_files_part(mem::take(&mut small_part), part_size);
+                part_size = 0;
+            }
+            small_part.push((path, size));
+            part_size += size;
+        }
+        if !small_part.is_empty() {
+            make_small_files_part(mem::take(&mut small_part), part_size);
+        }
+        assert_eq!(num_small_files, 0);
```
RinChanNOWWW, line 316: Better to add some comments to describe these codes.

youngsofun: added a commit, plz review again, thank you.

RinChanNOWWW (approved):

BohuTANG:
@youngsofun 

From my test in dev:
```
copy into meta.dwd_event_pushevent from @bohu_meta_temp file_format=(type=parquet) force=true;
```
This PR still cause oom, you can try it with the [internal doc](https://datafuselabs.feishu.cn/wiki/Hk7iwYsnWi3wsXkpbW5c9apTn0b), there might be some discoveries.

youngsofun:
> @youngsofun
> 
> From my test in dev:
> 
> ```
> copy into meta.dwd_event_pushevent from @bohu_meta_temp file_format=(type=parquet) force=true;
> ```
> 
> This PR still cause oom, you can try it with the [internal doc](https://datafuselabs.feishu.cn/wiki/Hk7iwYsnWi3wsXkpbW5c9apTn0b), there might be some discoveries.

row groups in the parquet files in `@bohu_meta_temp`  are very small (~900 each file, 200k in memory).  which is fixed by https://github.com/datafuselabs/databend/pull/12027.

small row groups lead to large file meta, when there are many files too, it will cause the OOM when read partition. (2700 * 900 row group info)
tolerant this is not the goal of this pr.  
to read files like this,  you need to add options like `max_files=100`.

PR #12048, opened 2023-07-10
feat: support specify bloom index columns

zhyass:

## Summary

Add `bloom_index_columns` option for fuse engine.  Support for specifying bloom index columns. 

To create table with bloom index:
```sql
CREATE TABLE table_name (
  column_name1 column_type1,
  column_name2 column_type2,
  ...
) ... bloom_index_columns='columnName1[, ...]'
```

To create or modify bloom index for a existing table:
*The existing Bloom index options will be replaced by the new options, it does not create Bloom filters for existing data.*
```sql
ALTER TABLE <db.table_name> SET OPTIONS(bloom_index_columns='columnName1[, ...]');
```

To disable the bloom index:
```sql
ALTER TABLE <db.table_name> SET OPTIONS(bloom_index_columns='');
```

- Closes #issue

### DIFF
--- a/tests/sqllogictests/suites/base/05_ddl/05_0003_ddl_alter_table
+++ b/tests/sqllogictests/suites/base/05_ddl/05_0003_ddl_alter_table
@@ -8,6 +8,12 @@ statement ok
 DROP TABLE IF EXISTS `05_0003_at_t1`
 
 statement ok
+DROP TABLE IF EXISTS `05_0003_at_t2`
+
+statement ok
+DROP TABLE IF EXISTS `05_0003_at_t3`
+
+statement ok
 CREATE TABLE `05_0003_at_t0`(a int)
 
 statement ok
@@ -72,3 +78,33 @@ SELECT b FROM `05_0003_at_t2`
 
 statement ok
 DROP TABLE IF EXISTS `05_0003_at_t2`
+
+statement ok
+set hide_options_in_show_create_table=0
+
+statement ok
+CREATE TABLE `05_0003_at_t3`(a int, b int, c int) bloom_index_columns='a,b,c' COMPRESSION='zstd' STORAGE_FORMAT='native'
+
+query TT
+SHOW CREATE TABLE `05_0003_at_t3`
+----
+05_0003_at_t3 CREATE TABLE `05_0003_at_t3` (   `a` INT,   `b` INT,   `c` INT ) ENGINE=FUSE BLOOM_INDEX_COLUMNS='a,b,c' COMPRESSION='zstd' STORAGE_FORMAT='native'
+
+statement ok
+ALTER TABLE `05_0003_at_t3` drop column c
+
+query TT
+SHOW CREATE TABLE `05_0003_at_t3`
+----
+05_0003_at_t3 CREATE TABLE `05_0003_at_t3` (   `a` INT,   `b` INT ) ENGINE=FUSE BLOOM_INDEX_COLUMNS='a,b' COMPRESSION='zstd' STORAGE_FORMAT='native'
+
+statement ok
+ALTER TABLE `05_0003_at_t3` rename column b to c
+
+query TT
+SHOW CREATE TABLE `05_0003_at_t3`
+----
+05_0003_at_t3 CREATE TABLE `05_0003_at_t3` (   `a` INT,   `c` INT ) ENGINE=FUSE BLOOM_INDEX_COLUMNS='a,c' COMPRESSION='zstd' STORAGE_FORMAT='native'
+
+statement ok
+DROP TABLE IF EXISTS `05_0003_at_t3`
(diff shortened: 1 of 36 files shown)


sundy-li:
What's the real usage of this pr? I did not get it.

zhyass:
askbend:summary

dantengsky:
> What's the real usage of this pr? I did not get it.


 bloom filter creation takes a significant amount of CPU.
 
 img_v2_938887ae-79cd-4cdd-a2aa-2184d747918g
 
for some  "wide" tables with lots of columns and 
- only a few of them involved in point query
- and the data ingestion is heavy (and time-critical somehow)

sundy-li:
> Creating bloom indexes for all the columns is wasteful.

I agree with that, so let's disable this creation by default?

dantengsky:
> > Creating bloom indexes for all the columns is wasteful.
> 
> I agree with that, so let's disable this creation by default?

I am not sure about this.

*maybe* for those tables that are not heavily appended(and the speed of ingestion is not that critical), and users are not sure about which columns might be involved in the point query, enabling bloom filters by default seems to be not a bad idea.  I do not if "those tables" are the common cases though.

BohuTANG:
I think this PR only for some special case.
Another question: if we rename a column who is in the options, what will happen?

zhyass:
> I think this PR only for some special case. Another question: if we rename a column who is in the options, what will happen?

Fix, drop column will remove the droped column from the bloom index columns. Rename column will replace the bloom index columns with new column name.

```
mysql> create table t(a int, b int, c int) bloom_index_columns='a,b,c';

mysql> alter table t drop column c;

mysql> set hide_options_in_show_create_table=0 ;
Query OK, 0 rows affected (0.00 sec)

mysql> show create table t;
+-------+-----------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                |
+-------+-----------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` INT,
  `b` INT
) ENGINE=FUSE BLOOM_INDEX_COLUMNS='a,b' COMPRESSION='zstd' STORAGE_FORMAT='parquet' |
+-------+-----------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.02 sec)
Read 0 rows, 0.00 B in 0.012 sec., 0 rows/sec., 0.00 B/sec.

mysql> alter table t rename column b to c;
Query OK, 0 rows affected (0.11 sec)

mysql> show create table t;
+-------+-----------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                |
+-------+-----------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` INT,
  `c` INT
) ENGINE=FUSE BLOOM_INDEX_COLUMNS='a,c' COMPRESSION='zstd' STORAGE_FORMAT='parquet' |
+-------+-----------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.06 sec)
Read 0 rows, 0.00 B in 0.049 sec., 0 rows/sec., 0.00 B/sec.
```

tests/sqllogictests/suites/base/05_ddl/05_0003_ddl_alter_table
```diff
@@ -71,4 +77,32 @@ SELECT b FROM `05_0003_at_t2`
...
+INSERT INTO TABLE `05_0003_at_t3` values(1,2,3)
+
+statement ok
+ALTER TABLE `05_0003_at_t3` drop column c
+
+statement ok
+INSERT INTO TABLE `05_0003_at_t3` values(4,5)
+
+statement ok
+ALTER TABLE `05_0003_at_t3` rename column b to c
```
BohuTANG, line 95: Need a check?
show create table t;

BohuTANG, line 95:
```suggestion
ALTER TABLE `05_0003_at_t3` rename column b to c

statement ok
show create table 05_0003_at_t3;
```

BohuTANG:
Conflicting files:
src/query/service/src/interpreters/interpreter_table_create.rs

tests/sqllogictests/suites/mode/standalone/explain/bloom_filter.test
```diff
@@ -158,6 +158,126 @@ EvalScalar
...
+        ├── output columns: [c1]
+        └── estimated rows: 6.00
+
+# alter bloom_index_columns
+statement ok
+alter table bloom_test_alter_t2 set options(bloom_index_columns='c1, c2')
+
+statement ok
+insert into  bloom_test_alter_t2 values(1,1), (3,4), (10,10)
+
```
dantengsky, line 233:
```suggestion

# note that the bloom index of column c1 is NOT created for the first 2 blocks (but created for the last block), 
# thus, while point querying on column c1, only 1 block might be filtered out.
```

dantengsky (approved):
LGTM.  

Added some comments to the test scripts, hope it helps

PR #11850, opened 2023-06-25
feat(storage):  improve optimize and recluster

zhyass:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

1. Add limit for recluster.  

```
ALTER TABLE [IF EXISTS] <name> RECLUSTER [FINAL] [WHERE condition] [LIMIT <segment_count>]
```
The option LIMIT sets the maximum number of segments to be recluster. Databend will select the newest segments. The maximum segment_ount limit is `max_storage_io_requests`.

2. Add memory usage limit for recluster

3. Sort the blockmeta by cluster_statistics during compact

4. after optimize compact, do recluster

5. Optimize recluster, serialize block in parallel during recluster.
   
```
create table test_order (
    id bigint,
    id1 bigint,
    id2 bigint,
    id3 bigint,
    id4 bigint,
    id5 bigint,
    id6 bigint,
    id7 bigint,
    
    s1 varchar,
    s2 varchar,
    s3 varchar,
    s4 varchar,
    s5 varchar,
    s6 varchar,
    s7 varchar,
    s8 varchar,
    s9 varchar,
    s10 varchar,
    s11 varchar,
    s12 varchar,
    s13 varchar,
    
    d1 DECIMAL(20, 8),
    d2 DECIMAL(20, 8),
    d3 DECIMAL(20, 8),
    d4 DECIMAL(20, 8),
    d5 DECIMAL(20, 8),
    d6 DECIMAL(30, 8),
    d7 DECIMAL(30, 8),
    d8 DECIMAL(30, 8),
    d9 DECIMAL(30, 8),
    d10 DECIMAL(30, 8),
    
    insert_time datetime,
    insert_time1 datetime,
    insert_time2 datetime,
    insert_time3 datetime,
    
    i int
) CLUSTER BY(id,insert_time);


create table random_source (
    id bigint,
    id1 bigint,
    id2 bigint,
    id3 bigint,
    id4 bigint,
    id5 bigint,
    id6 bigint,
    id7 bigint,
    
    s1 varchar,
    s2 varchar,
    s3 varchar,
    s4 varchar,
    s5 varchar,
    s6 varchar,
    s7 varchar,
    s8 varchar,
    s9 varchar,
    s10 varchar,
    s11 varchar,
    s12 varchar,
    s13 varchar,
    
    d1 DECIMAL(20, 8),
    d2 DECIMAL(20, 8),
    d3 DECIMAL(20, 8),
[...]

PR #12059, opened 2023-07-12
[DO NOT MERGE]: testing replace + re-cluster

dantengsky:

## Summary

This PR is based on
-  https://github.com/datafuselabs/databend/pull/11850

with the following features:

- auto recluster

  if the table being replaced into has cluster key, it will be automatically compact and reclustered
  
- more metrics (about the execution of `replace into` statement)

UPDATE:

rebase with the current PR #11850; bloom filter index is not disabled, but enabled by default, and can be specified by the user.

- Closes #issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2708,6 +2708,7 @@ dependencies = [
  "storages-common-pruner",
  "storages-common-table-meta",
  "streaming-decompression",
+ "sys-info",
  "table-lock",
  "tracing",
  "typetag",
--- a/src/query/ast/src/ast/format/ast_format.rs
+++ b/src/query/ast/src/ast/format/ast_format.rs
@@ -1403,11 +1403,18 @@ impl<'ast> Visitor<'ast> for AstFormatVisitor {
                 let action_format_ctx = AstFormatContext::new(action_name);
                 FormatTreeNode::new(action_format_ctx)
             }
-            AlterTableAction::ReclusterTable { selection, .. } => {
+            AlterTableAction::ReclusterTable {
+                selection, limit, ..
+            } => {
                 let mut children = Vec::new();
                 if let Some(selection) = selection {
                     self.visit_expr(selection);
                     children.push(self.children.pop().unwrap());
+                }
+                if let Some(limit) = limit {
+                    let name = format!("Limit {}", limit);
+                    let limit_format_ctx = AstFormatContext::new(name);
+                    children.push(FormatTreeNode::new(limit_format_ctx));
                 }
                 let action_name = "Action Recluster".to_string();
                 let action_format_ctx =
--- a/src/query/ast/src/ast/format/syntax/ddl.rs
+++ b/src/query/ast/src/ast/format/syntax/ddl.rs
@@ -183,6 +183,7 @@ pub(crate) fn pretty_alter_table_action(action: AlterTableAction) -> RcDoc<'stat
         AlterTableAction::ReclusterTable {
             is_final,
             selection,
+            limit,
         } => RcDoc::line()
             .append(RcDoc::text("RECLUSTER"))
             .append(if is_final {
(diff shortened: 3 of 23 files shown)


dantengsky:
@zhyass @BohuTANG  PTAL, if there are any concerns about the following long-run test, please let me know.



this PR has been tested by 

https://github.com/dantengsky/rr

using both 

- local dev box
   
  single node deployment, local fs; 65000 iterations, 
  
- and dev-cloud

  medium size WH; 25000 iterations
  

for both settings, no error was detected.


for your convenience, the doc of testing is replicated as follows:



-------------------

# test scenario 


## setup phase

  test table `test_order` will be created. which consists of 36 columns: 
  8 bigint, 13 varchar, 10 decimal, 4 timestamp and 1 int. 
 
  Table `random_source` of the same schema, with `random` engine, will also be created, which will
  be used as the random data source for the test table `test_order`.
  
  
## execution phase

### **`replace into`** routine
    
  **for each iteration, 1000 random rows will be replaced into `test_order`, with the following arrangements**

  - `on conflict` columns specified to 2 columns
   
    one is of type bigint, the other is of type timestamp.
   
  - values of two columns will be set correlated to value of current iteration

  - failures during executing the `replace into` statement will be ignored 
   
    those failures are mainly due to the conflicts of mutations. 
   
    the case driver will record the number of failures thought, so that later in the verification phase, 
    the data can be verified more in detail.


   roughly, the sql looks like the following:
   ~~~.sql
    replace into test_order on(id, insert_time)
       select 
          id,
          {iteration} as id1,
          {iteration * 7} as id2,
          ...
       from random limit 1000; 
   ~~~

 **For each 7 iterations, an extra action will be taken:**

 half of the rows inserted during this iteration will be replaced by them self.

 ~~~.sql
 replace into test_order on(id, insert_time)
   select * from test_order where id1 = {current_iteration} limit 500
 ~~~

 So that we can make sure that during testing there are definitely some data will be updated (not just appended).
 
   
    
   
### table maintenance routine

  table maintenance tasks will be executed **concurrently**, namely the following sqls:

  ~~~.sql
  optimize table test_order compact segment;
  optimize table test_order compact;
  optimize table test_order purge;
  alter table test_order recluster;
  ~~~
  during execution of the above sqls, errors will be ignored (mainly due to the conflicts of mutations). The case driver
  just keep trying executing each of the sqls as soon as possible (to simulate high concurrency scenario).
 
# verification phase

After iterations of `replace into` tasks done, the table maintenance tasks will be stopped.
Then the followings will be verified:

- value of successfully executed replace into statements
 
   1. count of `test_order`should equal to the number of successfully executed `replace into` statements, multiplied by 1000.
  
      **caveat**: there might be false negative in this case
   
      failure of execution the `replace into` statement at **client** side, does not mean that the execution is definitely failed at **server** side(network error, etc).
      
      

   2. count of each distinct value of `id1` should equal to 1000, since for each iteration, 1000 rows are inserted with the same(but distinct) value of `id1`. 


- value of correlated columns 

   for all the rows in `test_order`, value of `id2` should equal to value of `id1` multiplied by 7.
 
- full table scan should work

NOTE:

after the setup phase, at any time, the following invariant should hold:

- `select count() from test_order where id2 != id1 * 7` eq 0
- `select count() from (select count() a, id1 from test_order  group by id1) where a != 1000")` eq 0
- `select * from test_order ignore_result` should work

# how to run

- set DSN 

  Set the environment variable `DATABAEND_DSN` to the connection string of the database to be tested

  e.g. `export DATABEND_DSN="databend://root:@localhost:8000/?sslmode=disable"`

- kick off the test

  `cargo run -r [ITERATION]`

   Where the `[ITERATION]` is the number of iterations to run the test for. 
   If not specified, the default is 1000 iterations.

   At the end of execution, if everything is OK, logs like the following should be printed to the console:

    ```
    ==========================
    ====verify table state====
    ==========================
    CHECK: value of successfully executed replace into statements
    CHECK: value of correlated column
    CHECK: full table scanning
    ===========================
    ======     PASSED      ====
    ===========================
    ```

   Otherwise, error messages will be printed to the console.

PR #12062, opened 2023-07-12
chore: COPY status

BohuTANG:

## Summary

* add status for COPY
* add memory metrics for memory
* refine the code for better readable

- Closes #issue


src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -254,67 +240,120 @@ impl CopyInterpreter {
...
-            .await?;
+        let to_table = ctx.get_table(catalog, database, table).await?;
 
-        let (mut build_res, source_schema, files) = if let Some(query) = &plan.query {
-            let (build_res, source_schema) = self.build_query(query).await?;
-            (
-                build_res,
-                source_schema,
-                plan.stage_table_info
+        let mut build_res;
```
BohuTANG, line 250: I have careful review these changed codes, the logic is no changed, but some stateful test failed with: 
`+ERROR 1105 (HY000) at line 1: Internal. Code: 1001, Text = Cannot resize empty pipe..`
@JackTan25

BohuTANG: https://github.com/datafuselabs/databend/actions/runs/5529492495/jobs/10087872738?pr=12062#step:4:454

src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -254,67 +240,120 @@ impl CopyInterpreter {
...
-                    to_table,
+                CopyPlanType::CopyIntoTablePlanOption(plan.clone()),
+                source_schema,
+                to_table.clone(),
+            )?;
+
+            // Perf
+            {
+                metrics_inc_copy_read_file_cost_milliseconds(start.elapsed().as_millis() as u32);
+                self.set_status(&format!(
```
youngsofun, line 312: why do we need to metric and status for  `build_append_data_pipeline`?
it does not really copy any data, just builds the pipeline.

BohuTANG: How can i measure the copy data cost?

youngsofun: we can get time for 
 
1. special steps when bind and get_partitons
1. the whole pipeline execution time
2. inside each processor for each part/block.

I can refine this in another PR.

src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -254,67 +240,120 @@ impl CopyInterpreter {
...
+                build_res = PipelineBuildResult::create();
+                if !files.is_empty() {
+                    self.build_read_stage_table_data_pipeline(
+                        &mut build_res.main_pipeline,
+                        plan,
+                        to_table.get_block_thresholds(),
+                        files.clone(),
+                    )
+                    .await?;
+                }
```
youngsofun, line 269:
```suggestion
                } else {
                    return Ok(build_res);
                }
```

src/query/service/src/interpreters/interpreter_copy.rs
youngsofun: https://github.com/datafuselabs/databend/pull/12062/files#r1260972866

src/query/service/src/interpreters/interpreter_copy.rs
BohuTANG: I have set it in line260: `build_res = PipelineBuildResult::create();`

BohuTANG: I see, thanks.

PR #12069, opened 2023-07-12
refactor(query): improve json_path_query

Dousir9:

## Summary

Summary about this PR
Tested `JSON_PATH_QUERY` on a small data set, the time spent on `ProjectSet` was optimized from 1082ms to 668ms.

- Closes #issue


src/query/sql/src/evaluator/block_operator.rs
```diff
@@ -151,98 +152,169 @@ impl BlockOperator {
...
+                                        max_nums_per_row[i],
+                                    ))
+                                }
+                                let data_block = DataBlock::concat(&result_data_blocks)?;
+                                debug_assert!(data_block.num_rows() == total_num_rows);
+                                let block_entry = BlockEntry::new(
+                                    data_block.get_by_offset(0).data_type.clone(),
+                                    data_block.get_by_offset(0).value.clone(),
+                                );
+                                if block_is_empty {
```
b41sh, line 306: There are several places where `block_is_empty` is checked, is it possible to put the `block_entry` into a vector first and then construct the `block` at the end?

Dousir9: Due to `unused_indices` , we can't directly get the first `BlockEntry`.

b41sh (approved):

Issue #11219, opened 2023-04-24
Tracking: bitmap functions


ariesdevil:
**Summary**
Scalar Functions:
- [x] to_bitmap #11097 
- [x] bitmap_from_string #11097 
- [x] bitmap_to_string #11097 
- [x] bitmap_count #11304
- [x] build_bitmap #11304
- [x] bitmap_or
- [x] bitmap_and
- [x] bitmap_xor
- [x] bitmap_not
- [x] bitmap_and_not
- [x] bitmap_subset_limit
- [x] bitmap_subset_in_range
- [x] sub_bitmap
- [x] bitmap_contains 
- [x] bitmap_has_all
- [x] bitmap_has_any
- [x] bitmap_max
- [x] bitmap_min

Aggregate Functions:
- [x] bitmap_union
- [x] bitmap_and_count
- [x] bitmap_and_not_count
- [x] bitmap_xor_count
- [x] bitmap_or_count
- [x] intersect_count
- [x] bitmap_interssect

byrantwithyou:
Hi, I'm new to Databend and I want to contribute. I have a working knowledge of database kernel and Rust. Could you assign some of the tasks in this list to me?

byrantwithyou:
> Hi, I'm new to Databend and I want to contribute. I have a working knowledge of database kernel and Rust. Could you assign some of the tasks in this list to me?

If nobody is doing the same thing, I want to do `bitmap_or`, `bitmap_and`, `bitmap_xor`, `bitmap_not`, `bitmap_and_not` together, because they are kinda similar.

ariesdevil:
@byrantwithyou Thanks for being interested in these funcs, When things are done, I'll put the pr address after the corresponding function.

silver-ymz:
Hello, I'm new to Databend and want to have a try for this. I want to try `bitmap_contains`, `bitmap_has_all`, `bitmap_has_any`, `bitmap_max`, `bitmap_min` together firstly. Could you assign them to me?

byrantwithyou:
I'm so sorry that I couldn't do this task because I'm a little bit busy these days. Please assign it to anyone else.

gitccl:
Hi, I'd like to try `bitmap_or` and `bitmap_and`

gitccl:
Hi @sundy-li , Can `bitmap_hash` use `ahash` as the hash function?

sundy-li:
>  Can bitmap_hash use ahash as the hash function?

[...]

PR #12070, opened 2023-07-12
feat: add intersect_count function

akoshchiy:

## Summary
Added `intersect_count` aggregate function.

- Related #11219


akoshchiy:
@ariesdevil Hi! I need some advice, here is the draft version of `intsersect_count`.
1. Is it optimal way to filter the data in columns or it could be better?
2. There is no type coercion in the arguments, and I have no idea how to do it properly, could you give me some hint?

ariesdevil:
> 1. Is it optimal way to filter the data in the column or it could be better?

The current implementation is LGTM.

> 2. There is no type coercion in the arguments, and I have no idea how to do it properly, could you give me some hint?

You can put the `filter_values` to the params instead of arguments, see:
* https://databend.rs/doc/sql-functions/aggregate-functions/aggregate-quantile-cont
* https://github.com/datafuselabs/databend/blob/b3a3b35e8dbc8bb6062c183e802742dac4bc9509/src/query/functions/src/aggregates/aggregate_quantile_cont.rs#L253-L307

ariesdevil (approved):
Great work, LGTM

Issue #12067, opened 2023-07-12
Feature: speed up CTE by materialization


xudong963:
**Summary**

Inlining CTE is a good idea for most cases, but if the CTE is heavy, the cost is sooooo expensive, such as TPCH Q15, I have mentioned in https://github.com/datafuselabs/databend/pull/11998.

So it's better to introduce materialization for expensive CTE.

PR #12090, opened 2023-07-14
feat: speed up CTE by materialization

xudong963:

## Summary

The ticket supports defining materialized CTE. By materialization, CTE only needs to be read once for storage which can reduce IO and speed up queries.

How to describe a CTE as materialized?
> We introduce the `materialized` keyword and will decide if a CTE is materialized at the parser phase.

> Syntax: `with t1 as materialized (...) ....`

At the plan phase, introduce `CteScan` and `MaterializedCte` nodes to extend the plan tree.

Take the tpch Q15 as an example:

```sql
+-----------------------------------------------------------------------+
| explain                                                               |
+-----------------------------------------------------------------------+
| MaterializedCte: 0                                                    |
| ├── Left                                                              |
| │   └── Scan: default.tpch.lineitem, rows: 600572                     |
| └── Right                                                             |
|     └── HashJoin: INNER                                               |
|         ├── Build                                                     |
|         │   └── CteScan: 0                                            |
|         └── Probe                                                     |
|             └── HashJoin: INNER                                       |
|                 ├── Build                                             |
|                 │   └── CteScan: 0                                    |
|                 └── Probe                                             |
|                     └── Scan: default.tpch.supplier, rows: 1000       |
+-----------------------------------------------------------------------+
```
The left of `MaterializedCte` is the CTE, it'll be executed once and then **notify** `CteScan`s to use.

The result of `MaterializedCte` will be saved at `QueryContext`, see `MaterializedCtesBlocks` for details.

For TPCH Q15, the performance improvement is about **twice** as much, which is as expected since it reduces IO by half.

(Will consider cluster mode for materialized CTE in the following tickets)

cc @soyeric128 for doc after the ticket merged.

- Closes https://github.com/datafuselabs/databend/issues/12067

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12090)


sundy-li (approved):

src/query/sql/src/planner/optimizer/cost/cost_model.rs
```diff
@@ -81,6 +82,14 @@ fn compute_cost_join(memo: &Memo, m_expr: &MExpr, plan: &Join) -> Result<Cost> {
     Ok(Cost(cost))
 }
 
+fn compute_materialized_cte(memo: &Memo, m_expr: &MExpr) -> Result<Cost> {
+    let left_group = m_expr.child_group(memo, 0)?;
+    let right_group = m_expr.child_group(memo, 1)?;
+    let card = left_group.stat_info.cardinality + right_group.stat_info.cardinality;
```
Dousir9, line 88: It seems that we only need to get the cost of CTE here.

src/query/sql/src/planner/optimizer/heuristic/prune_unused_columns.rs
```diff
@@ -278,7 +279,28 @@ impl UnusedColumnPruner {
...
+                    .derive_relational_prop_child(0)?
+                    .output_columns
+                    .clone();
+                let right_used_column = RelExpr::with_s_expr(expr)
+                    .derive_relational_prop_child(1)?
+                    .used_columns
+                    .clone();
+                // Get the intersection of `left_used_column` and `right_used_column`
+                let left_required = left_output_column
+                    .intersection(&right_used_column)
```
Dousir9, line 293: Why intersect left_output_column and right_used_column.

xudong963: left side only needs to keep the columns that are used by right side.

Dousir9 (approved):

PR #12118, opened 2023-07-17
fix: support distributed copy from parquets

JackTan25:

## Summary

Summary about this PR

- Closes #issue


src/query/service/src/interpreters/interpreter_copy.rs
```diff
@@ -162,22 +162,22 @@ impl CopyInterpreter {
...
-                    false,
-                )
-                .await?
-        };
-
-        if read_source_plan.parts.len() <= 1 {
-            return Ok(None);
-        }
         if plan.query.is_none() {
+            let read_source_plan = {
```
SkyFan2002, line 166: May I ask what's change here?

JackTan25: for parquet files, no need to build source plan here, we will build in distributed query.

SkyFan2002 (approved):

BohuTANG:
Performance great 👍 

16c32g standalone:
```
copy into test from @bohu_stage2 file_format=(type=parquet) force=true;
20000000 rows affected in (58.436 sec)
```

2* (8c16g) cluster:
```
set enable_distributed_copy_into=1;
copy into test_order from @bohu_stage2 file_format=(type=parquet) force=true;
20000000 rows affected in (16.634 sec)
```

PR #12120, opened 2023-07-17
chore(storage): add length check for string column in bloom index

zhyass:

## Summary

 If the average length of the string column exceeds 256 bytes, the bloom index for the column will not be established.

- Closes #issue

### DIFF
--- a/src/query/storages/common/index/src/bloom_index.rs
+++ b/src/query/storages/common/index/src/bloom_index.rs
@@ -211,6 +211,11 @@ impl BloomIndex {
                         })
                         .collect::<Vec<_>>();
                     let column = Column::concat(&source_columns);
+
+                    if Self::check_large_string(&column) {
+                        continue;
+                    }
+
                     (column, val_type)
                 }
                 _ => {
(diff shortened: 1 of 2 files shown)


src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -465,6 +475,18 @@ impl BloomIndex {
...
+        if let Column::String(v) = &column {
+            let bytes_per_row = v.data().len() / v.len().max(1);
+            if bytes_per_row > 256 {
+                // If the average length of the string column exceeds 256 bytes,
+                // the bloom index for the column will not be established.
+                return true;
+            }
+        }
+        false
+    }
```
BohuTANG, line 489:
```suggestion
/// Checks if the average length of a string column exceeds 256 bytes.
/// If it does, the bloom index for the column will not be established.
fn check_large_string(column: &Column) -> bool {
    if let Column::String(v) = column {
        let bytes_per_row = v.data().len() / v.len().max(1);
        return bytes_per_row > 256;
    }
    false
}

```

dantengsky (approved):

PR #12137, opened 2023-07-19
feat(query): Alter user support set/unset network policy

b41sh:

## Summary

Alter user support set/unset network policy
Check client ip when login if have set network policy

- Closes #issue


drmingdrmer (commented):
Proto changes requires a new version to be added here:
https://github.com/datafuselabs/databend/blob/2a4c6b313b2d463551da3c3bb45a129362ad9765/src/meta/proto-conv/src/util.rs#L81-L86

And there should be a corresponding test added like:

https://github.com/datafuselabs/fuse-query/blob/2a4c6b313b2d463551da3c3bb45a129362ad9765/src/meta/proto-conv/tests/it/v049_network_policy.rs

sundy-li (approved):

PR #12139, opened 2023-07-19
feat: alter table to modify column data type

lichuang:

## Summary

feat: alter table to modify column data type
SQL: `alter table table_name modify column data type column_name type_name[, column_name type_name]`

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12139)


lichuang:
askbend:summary

src/query/service/src/interpreters/interpreter_table_modify_column.rs
```diff
@@ -94,21 +108,143 @@ impl ModifyTableColumnInterpreter {
...
+        table: &Arc<dyn Table>,
+        type_name: &TypeName,
+    ) -> Result<PipelineBuildResult> {
+        let schema = table.schema().as_ref().clone();
+        let table_info = table.get_table_info();
+        if let Ok(i) = schema.index_of(&self.plan.column) {
+            // Check if this column is referenced by computed columns.
+            let field: DataField = schema.field(i).into();
+            if field.computed_expr().is_none() {
+                check_referenced_computed_columns(
```
b41sh, line 148: The `check_referenced_computed_columns` function needs to be rewritten to handle the case where the column data type changes

dantengsky:
:+1: 

If I understand correctly that

- modifying the data type of multiple columns in one statement is not supported, e.g
   
   ALTER TABLE t MODIFY c1 SET DATA TYPE ..,  COLUMN c2  SET DATA TYPE ..
   
   since the implementation will rewrite the whole table, we may need to support multiple column modifications (of types) in one go.
   
-  about the syntax

   seem the implementation follows the syntax of snowflake,  looks good to me.  @BohuTANG any concerns about this? any user that prepares using this feature expects different syntax?
   
besides, if all the new types that the user specified in the   `MODIFY c SET DATA TYPE ... ` are of the same types,  we need not rewrite the table data.

src/query/service/src/interpreters/interpreter_table_modify_column.rs
```diff
@@ -182,39 +348,29 @@ impl Interpreter for ModifyTableColumnInterpreter {
 
         // NOTICE: if we support modify column data type,
         // need to check whether this column is referenced by other computed columns.
```
BohuTANG, line 350: Need to handle the warning: `need to check whether this column is referenced by other computed columns.`?
Also need a test for it.

lichuang: has test for this case, see:
https://github.com/datafuselabs/databend/pull/12139/files#diff-c9f542e7c99e6aecb18356ed72b1e30e567b9e11c219100c9dc2d1cadf86eeb0

src/query/ast/src/parser/statement.rs
```diff
@@ -1970,6 +1970,52 @@ pub fn alter_database_action(i: Input) -> IResult<AlterDatabaseAction> {
...
+    let convert_stored_computed_column = map(
+        rule! {
+            #ident ~ DROP ~ STORED
+        },
+        |(column, _, _)| ModifyColumnAction::ConvertStoredComputedColumn(column),
+    );
+
+    let modify_column_type = map(
+        rule! {
+            DATA ~ TYPE ~ #column_type ~ ("," ~ #column_type)*
```
BohuTANG, line 2001: This `DATA ~ TYPE` keyword is no need, it looks special.

`ALTER TABLE <name> MODIFY  [<col1_name> <type>][,<col1_name> <type>]`

src/query/storages/fuse/src/operations/common/processors/sink_commit.rs
```diff
@@ -117,9 +124,19 @@ where F: SnapshotGenerator + Send + 'static
             input,
             need_lock,
             start_time: Instant::now(),
+            prev_snapshot_id,
```
SkyFan2002, line 127: What about move the snapshot_id into `ConflictResolveContext`: https://github.com/datafuselabs/databend/blob/282cbc25b8009a66d0dab37d0917bb5322b61db2/src/query/storages/fuse/src/operations/common/snapshot_generator.rs#L332
So that if we change the conflict resolve method(i.e. int64 -> int32 may be resovable) later , we can reuse the retry process：https://github.com/datafuselabs/databend/blob/282cbc25b8009a66d0dab37d0917bb5322b61db2/src/query/storages/fuse/src/operations/common/processors/sink_commit.rs#L213

SkyFan2002: Let's fix it when problem actually occurs

SkyFan2002 (approved):

dantengsky (approved):

PR #12140, opened 2023-07-19
fix: improve query rewrite and fix bugs.

RinChanNOWWW:

### Aggregating index

- Allow aggregation function `AVG` in aggregating index.
- In the past, if one group by key is an expression, Databend will not use aggregating index. This PR fixed this.
- The output fields of `TableScan` will the same order as the table schema. (In the past, the fields were sorted by names).
- Fix the memory violation while merging aggregate state.

### Aggregation

- Drop temporary state every time to avoid memory leak.
- Sort the input scalars of aggregation by index.


ariesdevil (approved):

src/query/service/src/pipelines/processors/transforms/aggregator/transform_aggregate_partial.rs
```diff
@@ -200,11 +200,19 @@ impl<Method: HashMethodBounds> TransformPartialAggregate<Method> {
...
-                function.merge(place.next(offset), self.temp_place)?;
+                function.deserialize(state_place, &mut raw_state)?;
+                function.merge(place.next(offset), state_place)?;
+                if function.need_manual_drop_state() {
+                    unsafe {
+                        // State may allocate memory out of the arena,
+                        // drop state to avoid memory leak.
+                        function.drop_state(state_place);
+                    }
+                    function.init_state(state_place);
```
sundy-li, line 214: the state is override by `deserialize`

RinChanNOWWW: `deserialize` will call `place.get::<State>` first. If we do this without `init_state` after `drop_state`, we will violate the memory.

Dousir9:
Need to rerun ci-benchmark.

src/query/ee/tests/it/aggregating_index/index_scan.rs
```diff
@@ -56,6 +56,12 @@ async fn test_index_scan_with_count() -> Result<()> {
     test_index_scan_with_count_impl("native").await
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_index_scan_agg_args_are_expression() -> Result<()> {
+    test_index_scan_agg_args_are_expression_impl("parquet").await?;
+    test_index_scan_agg_args_are_expression_impl("native").await
+}
```
RinChanNOWWW, line 63: A comment to prevent from being merged.

BohuTANG:
> ## ClickBench Report
> * **hits**: https://repo.databend.rs/benchmark/clickbench/pr/12140/5617431933/hits.html
> * **tpch**: https://repo.databend.rs/benchmark/clickbench/pr/12140/5617431933/tpch.html

The report is broken :\

PR #7777, opened 2022-09-21
chore(planner): Deprecate `Project` operator

leiysky:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

We no longer need `Project` operator.

PR #12157, opened 2023-07-20
refactor(query): improve hash join

Dousir9:

## Summary

1. Reduce calls of `row_space.datablocks()` in left join.
2. Add `pre_projections` and `projections` for hash join, `pre_projections` are used to eliminate columns that are only used for hash join key, and `projections` are used for final column elimination: 
  * (1) In the phase of generating hash join physical plan: we will use `pre_projections` to generate `probe_projections` and `build_projections` and build `merged_fields` according to different join types, `merged_fields` contains columns required by `other_conditions`, but columns related to `other_conditions` may also need to be eliminated, Therefore, the output schema of the final hash join is the result of applying `projections` to `merged_fields`. 
  * (2) In the execution phase of hash join: After building the probe key and build key, we apply `probe_projections` and `build_projections` to probe datablock and build datablock, which can help us reduce memory usage and calls of expensive functions (`take_compacted_indices` and `gather`), after processing `other_conditions`, we will use `projections` for final column elimination.
3. Added projections for `EvalScalar` and `Filter`, remove `BlockOperator::MapWithOutput`.
4. Supported dphyp algorithm for delete.
5. For agg index, there may be two types of datablocks, AggIndex and Normal, so their schemas are different. In order to unify the schema, `agg_functions_len` is added to `AggIndexMeta` to prevent the columns of agg index from being eliminated.
6. Prune columns when building physical plan.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12157)


src/query/sql/src/evaluator/block_operator.rs
```diff
@@ -43,30 +42,28 @@ use common_pipeline_transforms::processors::transforms::Transform;
...
+    Map {
         exprs: Vec<Expr>,
         /// The index of the output columns, based on the exprs.
-        output_indexes: Vec<usize>,
+        projections: Option<ColumnSet>,
     },
 
     /// Filter the input [`DataBlock`] with the predicate `eval`.
-    Filter { expr: Expr },
+    Filter { projections: ColumnSet, expr: Expr },
```
RinChanNOWWW, line 58: Why Filter needs projection?

src/query/sql/src/evaluator/cse.rs
```diff
@@ -72,21 +74,50 @@ pub fn apply_cse(
...
+                    //     Some(
+                    //         projections
+                    //             .iter()
+                    //             .filter(|idx| **idx < input_num_columns)
+                    //             .copied()
+                    //             .collect::<ColumnSet>(),
+                    //     )
+                    // } else {
+                    //     None
+                    // };
```
leiysky, line 95: Remove these.

leiysky:
We previously have a `Project` operator, but is too hard to maintain it in optimizer. So it's deprecated in #7777.

IMO, it's better to put the `projections` into `PhysicalPlan`. This can be calculated when building `PhysicalPlan`.

Dousir9:
> We previously have a `Project` operator, but is too hard to maintain it in optimizer. So it's deprecated in #7777.
> 
> IMO, it's better to put the `projections` into `PhysicalPlan`. This can be calculated when building `PhysicalPlan`.

Yes, we currently put `projections` into `PhysicalPlan`(`HashJoin`, `EvalScalar`, `Filter` for now) and generate `projections` when building `PhysicalPlan`.

src/query/service/src/interpreters/interpreter_delete.rs
```diff
@@ -327,16 +330,24 @@ pub async fn subquery_filter(
...
+        let (dp_res, optimized) =
+            DPhpy::new(ctx.clone(), metadata.clone()).optimize(Arc::new(expr.clone()))?;
+        if optimized {
+            expr = (*dp_res).clone();
+            dphyp_optimized = true;
+        }
+    }
+    let mut cascades = CascadesOptimizer::create(ctx.clone(), metadata.clone(), dphyp_optimized)?;
     expr = cascades.optimize(expr)?;
+    expr = heuristic.optimize(expr, &RESIDUAL_RULES)?;
```
xudong963, line 350: heuristic after cascades?

src/query/sql/src/planner/optimizer/optimizer.rs
```diff
@@ -189,7 +190,8 @@ pub fn optimize_query(
     if enable_distributed_query {
         result = optimize_distributed_query(ctx.clone(), &result)?;
     }
-
+    result = heuristic.optimize_expression(&result, &RESIDUAL_RULES)?;
```
xudong963, line 193: Why put it here?

```suggestion
    if ctx.get_settings().get_enable_dphyp()? {
        let (dp_res, optimized) =
            DPhpy::new(ctx.clone(), metadata.clone()).optimize(Arc::new(result.clone()))?;
        if optimized {
            dphyp_optimized = true;
            result = heuristic.optimize_expression(&dp_res, &RESIDUAL_RULES)?;
            result = heuristic.post_optimize(dp_res)?;
        }
    }
```

src/query/sql/src/planner/plans/join.rs
```diff
@@ -141,6 +141,8 @@ impl Display for JoinType {
 /// the probe side.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct Join {
+    pub pre_projections: Vec<IndexType>,
```
xudong963, line 144: Please add notes for `pre_projections` and `projections`

src/query/sql/src/executor/physical_plan.rs
```diff
@@ -561,6 +575,9 @@ pub struct HashJoin {
     /// A unique id of operator in a `PhysicalPlan` tree.
     /// Only used for display.
     pub plan_id: u32,
+    pub projections: ColumnSet,
+    pub probe_projections: ColumnSet,
```
xudong963, line 579: please add notes for `probe_projections` and `build_projections`

tests/sqllogictests/suites/tpch/queries.test
```diff
@@ -1260,13 +1260,13 @@ limit 5;
 ----
 HashJoin: INNER
 ├── Build
-│   └── HashJoin: INNER
-│       ├── Build
-│       │   └── Scan: default.tpch.customer, rows: 15000
-│       └── Probe
-│           └── Scan: default.tpch.orders, rows: 150000
+│   └── Scan: default.tpch.lineitem, rows: 600572
```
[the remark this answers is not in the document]
Dousir9: Yes, there is an eager aggregation caused by cascades, we put heuristic after cascades and apply `CommuteJoin`.

src/query/service/src/interpreters/interpreter_delete.rs
Dousir9: ditto.

src/query/sql/src/planner/optimizer/optimizer.rs
Dousir9: Because we need to consider the `Exchange` operator to avoid the columns required by `Exchange::Hash(Vec<ScalarExpr>)` being pruned.

tests/sqllogictests/suites/mode/cluster/distributed_copy_into_table2.test
```diff
@@ -48,4 +48,4 @@ select block_count from fuse_snapshot('default','products');
 2
 
 statement ok
-set enable_distributed_copy_into = 0;
\ No newline at end of file
+set enable_distributed_copy_into = 0;
```
JackTan25, line 51: can we avoid this modify?

Dousir9: It is automatically generated by sqllogictest.

src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -1487,6 +1383,422 @@ impl PhysicalPlanBuilder {
...
+    }
+
+    #[async_backtrace::framed]
+    async fn build_scan(
+        &mut self,
+        scan: &Scan,
+        required: ColumnSet,
+        stat_info: PlanStatsInfo,
+    ) -> Result<PhysicalPlan> {
+        // add virtual columns to scan
```
xudong963, line 1590: I suggest moving prune-column code into individual function for each relop.

xudong963 (approved):

BohuTANG:
Conflicting files \:

PR #12176, opened 2023-07-24
chore: change service image to debian:bookworm

everpcpc:

## Summary

Summary about this PR

- Closes #issue

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM debian:bookworm
 
 ARG TARGETPLATFORM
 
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM debian:bookworm
```
Xuanwo, line 1: Wow, really new!

everpcpc: Then we could put binaries built by ourselves into the image for test. cc @wubx

Xuanwo (approved):

PR #12186, opened 2023-07-25
chore: refine lazy materialization

Dousir9:

## Summary

Summary about this PR
We need to know all columns as early as possible, which helps us to remove some unused columns, so this pr will generate `_row_id` column index in bind phase.

- Closes #issue


src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -367,8 +367,11 @@ impl PhysicalPlanBuilder {
                     .unwrap();
                 let index = self
                     .metadata
-                    .write()
-                    .add_internal_column(scan.table_index, internal_column.clone());
+                    .read()
+                    .row_id_index_by_table_index(scan.table_index);
+                debug_assert!(index.is_some());
+                // Safe to unwrap.
```
xudong963, line 373: Seems `analyze_lazy_materialization` doesn't always be called.

So here I recommend conservative style:

```rust
let index = if let Some(index) = "try to read internal column" {
    index
} else {
    // add internal column
}

Dousir9: > Seems `analyze_lazy_materialization` doesn't always be called.
> 
> So here I recommend conservative style:
> 
> ```rust
> let index = if let Some(index) = "try to read internal column" {
>     index
> } else {
>     // add internal column
> }
> ```

https://github.com/datafuselabs/databend/blob/2e772c6d202c18c839874fd1d04819f54c2c19a3/src/query/sql/src/executor/physical_plan_builder.rs#L362-L378
if `!metadata.lazy_columns().is_empty()` is true, the `analyze_lazy_materialization` must be called.
https://github.com/datafuselabs/databend/blob/2e772c6d202c18c839874fd1d04819f54c2c19a3/src/query/sql/src/planner/binder/select.rs#L842-L851

xudong963 (approved):
cc @RinChanNOWWW

sundy-li (approved):

PR #12188, opened 2023-07-25
chore: fix left single join

Dousir9:

## Summary

Summary about this PR
If the join type is left single, the `Datablock` on the build side of the hash join need to be Nullable.

- Closes #issue


xudong963 (approved):

BohuTANG:
This fix will affect what's the cases like?

xudong963:
> This fix will affect what's the cases like?

cc @Dousir9 add some cases that failed before this PR.

BohuTANG:
It would be better to fill the summary, we hope to know every improvement/fix with the join :)

PR #12203, opened 2023-07-25
chore: when vacuum dropped table, ignore when TableIdToName data not exist

lichuang:

## Summary

chore: when vacuum dropped table, ignore when TableIdToName data not exist

- Closes #issue


BohuTANG:
In which case will `TableIdToName data not exist`?

PR #12207, opened 2023-07-26
fix: compact_segment_bug

JackTan25:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12207)


JackTan25:
https://github.com/datafuselabs/databend/pull/12207/files#diff-46625998da2d3923cca903cd6ec42cee1114e9440cfcb0652255ec9c8b358da5L55 I think row_idx won't over 100w, need 20bits, so this modify is good. cc @sundy-li @dantengsky

sundy-li:
> I think row_idx won't over 100w

```
pub const DEFAULT_BLOCK_MAX_ROWS: usize = 1000 * 1000;
```

This may be incorrect if there is large block with more than 100w rows.

JackTan25:
> > I think row_idx won't over 100w
> 
> ```
> pub const DEFAULT_BLOCK_MAX_ROWS: usize = 1000 * 1000;
> ```
> 
> This may be incorrect if there is a large block with more than 100w rows.

I think I should limit the row_per_block option when create table?

sundy-li:
maybe we should expand the row_id type to int128.

JackTan25:
> maybe we should expand the row_id type to int128.

this costs deeply, can we accept the memory size? If so I will try it.

sundy-li (approved):

JackTan25:
if use u128, I need to modify many codes, add a new UInt128 type, should we? cc @sundy-li

JackTan25:
> > I think I should limit the row_per_block option when create table?
> 
> Let's have this limit for now.

I think so.

PR #12213, opened 2023-07-26
chore: Polish gc orphan files retention time

lichuang:

## Summary

chore: use min(now - get_retention_period(), retention_time) as gc orphan files retention time to protect files that generated by txn which has not been committed being gc.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12213)


lichuang:
askbend:summary

src/query/ee/src/storages/fuse/operations/vacuum_table.rs
```diff
@@ -375,6 +377,10 @@ pub async fn do_vacuum(
         Ok(Some(purge_files))
     } else {
         debug_assert!(dry_run_limit.is_none());
+        let retention = Duration::hours(ctx.get_settings().get_retention_period()? as i64);
+        // use min(now - get_retention_period(), retention_time) as gc orphan files retention time
+        // to protect files that generated by txn which has not been committed being gc.
+        let retention_time = std::cmp::min(chrono::Utc::now() - retention, retention_time);
```
zhyass, line 383: Should we use `base_snapshot.timestamp - retention` just like purge? @dantengsky

zhyass (approved):

src/query/ee/src/storages/fuse/operations/vacuum_table.rs
dantengsky: local clock may bring troubles, consider a crazy fast lock which is used by `Utc::new()`…

for this PR, let’s temporarily ignore the above issue.

we need to address this later,  by marking the largest visible snapshot in the meta server (and in a “monotinically increasing way”, roughly.

dantengsky (approved):

PR #12223, opened 2023-07-27
refactor: extract common code parquet table for adding arrow-rs impl.

youngsofun:

## Summary

extract these so make it easy to maintain after adding arrow-rs impl (2 impl are expected to co-exists for a long period):

1. Partition
2. processors
3. trait ParquetReader

only leave those arrow2 structure in common code:

1. Compression:  simply convert to/from it from arrow-rs::Compression
5. Interval, Bitmap: not plan to involve page_pruning in the first version of arrow-rs impl, so we can deal with it later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12223)


Xuanwo (commented):
bravo!

Xuanwo (approved):

Issue #12266, opened 2023-07-31
Tracking: enhance materialized cte


xudong963:
**Summary**

- [x] support materialized CTE is used in SQL with different aliases
- [x] fix prune column of `CteScan` or `MaterializedCte` operator
- [x] support materialized CTE is used in set operations, such as union
- [x] support materialized CTE is used in [materialized] CTE.
- [x] change tpcds SQLs with `materialized` cte and add them to test suit

PR #12288, opened 2023-08-01
feat: support materialized cte in set operation and fix prune column

xudong963:

## Summary

- fix prune column of CteScan or MaterializedCte operator
- support materialized CTE is used in set operations, such as union

Part of #12266

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12288)


src/query/sql/src/planner/optimizer/heuristic/prune_unused_columns.rs
```diff
@@ -314,6 +322,9 @@ impl UnusedColumnPruner {
             }
 
             RelOperator::MaterializedCte(cte) => {
+                if self.apply_lazy {
```
Dousir9, line 325: Currently lazy materialization is only used when there is a single table, so seems there is no need to consider `apply_lazy` here.

Dousir9 (approved):

PR #12291, opened 2023-08-01
docs: materialized cte

soyeric128:

## Summary

Introduced "materialized" keyword.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12291)


docs/doc/14-sql-commands/20-query-syntax/02-query-with.md
```diff
@@ -1,8 +1,13 @@
...
+<FunctionDescription description="Introduced or updated: v1.2.38"/>
+
+Databend supports common table expressions (CTEs) with a WITH clause, allowing you to define one or multiple named temporary result sets used by the following query. The term "temporary" implies that the result sets are not permanently stored in the database schema. They act as temporary views only accessible to the following query.
+
+Databend introduces the "Materialized" parameter for CTEs, allowing you to decide whether a CTE should be materialized or remain temporary. It is recommended to use the "Materialized" parameter for CTEs involving complex computations or large datasets to improve query performance by storing results on a physical storage medium. However, for simpler CTEs with limited usages, using "Materialized" may not be necessary. The decision should consider the trade-offs between performance gains and storage utilization.
```
xudong963, line 10: databend decides whether or not to inline a CTE by using the `materialized` keyword, if `materialized`, the CTE will not be inlined, so it will be executed only once, and then the result will be stored in memory, reducing repetitive execution of the CTE. The trade-off is that if `materialized`, i.e. not inline CTE, it will reduce the chance of the whole query predicate being pushed down and join reordered, so the business needs to make a choice based on its own situation.

docs/doc/14-sql-commands/20-query-syntax/02-query-with.md
```diff
@@ -37,6 +35,51 @@ WHERE  city = 'Montréal'
...
+    GROUP BY customer_id
+)
+SELECT customer_id
+FROM customer_orders
+WHERE order_count > 2;
+```
+  </TabItem>
+  <TabItem value="MATERIALIZED" label="MATERIALIZED">
+
+In this case, we use the MATERIALIZED keyword, which means the CTE customer_orders will not be inlined. Instead, the CTE's result will be calculated and stored in memory during the CTE definition's execution. Later, when executing the main query, Databend will directly retrieve the pre-computed result from memory, avoiding the overhead of repeatedly executing the CTE.
```
xudong963, line 65: > avoiding the overhead of repeatedly executing the CTE.

If the example SQL uses the CTE more than once, it is only then that we can state this

soyeric128: Updated to:  If a query uses a CTE more than once, this also helps avoid the overhead of repeatedly executing the CTE.

docs/doc/14-sql-commands/20-query-syntax/02-query-with.md
```diff
@@ -39,42 +39,46 @@ ORDER  BY customername;
...
-FROM customer_orders
-WHERE order_count > 2;
+SELECT co1.customer_id, co1.order_count, co2.order_count AS other_order_count
+FROM customer_orders co1
+JOIN customer_orders co2 ON co1.customer_id = co2.customer_id
+WHERE co1.order_count > 2
+  AND co2.order_count > 5;
 ```
-This can significantly improve performance for complex CTEs or situations where the CTE's result is used multiple times. However, as the CTE is no longer inlined, the query optimizer may find it difficult to push the CTE's conditions into the main query or optimize the join order, potentially leading to decreased overall query performance. 
+This can significantly improve performance for complex CTEs or situations where the CTE's result is used multiple times. However, as the CTE is no longer inlined, the query optimizer may find it difficult to push the CTE's conditions into the main query or optimize the join order, potentially leading to decreased overall query performance.
```
xudong963, line 81: The main point is when CTE is used multiple times, then performance will improve.

So, `for complex CTEs` can be deleted.

docs/doc/14-sql-commands/20-query-syntax/02-query-with.md
```diff
@@ -39,42 +39,46 @@ ORDER  BY customername;
...
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
 <Tabs>
-  <TabItem value="inline" label="Inline" default>
+  <TabItem value="Inline" label="Inline" default>
 
-In this query, the CTE customer_orders will be inlined during query execution. Databend will directly embed the definition of customer_orders within the main query to create a single execution plan.
+In this query, the CTE *customer_orders* will be inlined during query execution. Databend will directly embed the definition of *customer_orders* within the main query to create a single execution plan.
```
xudong963, line 50:
```suggestion
In this query, the CTE *customer_orders* will be inlined during query execution. Databend will directly embed the definition of *customer_orders* within the main query.
```

xudong963 (approved):
LGTM

PR #12279, opened 2023-07-31
feat: add Basic ownership model for permission management

ZhiHanZ:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

1. Add Ownership field for table and database metadata
2. Add `Ownership` keyword in grant privilege type
3. Add verification about ownership during permission check part

Next:
1. Bring ownership during database/table creation
2. Add Scenario oriented Permission tests to guarantee their is no surprising for databend permission model
- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12279)

PR #12299, opened 2023-08-01
feat(rbac): verify privilege with all the roles

flaneur2020:

## Summary

Currently on veirfy_privilege, only the current role is encountered. There's only one current role in one session. In the before, the current role means only the current role's privilege set takes effect in one session.

Now we have introduced OWNERSHIP in #12279, we can revisit the semantics about the current role. Now we can make the current role means about "when I created any database/table, the datbase/table will be owned by the current role by default", and decouple it from affecting the effective privilege set.

In snowflake, there's another concept called SECONDARY ROLES, which allows the user do disable some granted roles, to limit privilege in the session. The SECONDARY ROLES can have a value `ALL`, which means all the privileges set from roles granted to the user are effective.

The new `validate_privilege` in this PR will behave the same as SECONDARY ROLES = ALL. 

However, we do not have explicit requirement about SECONDARY ROLES yet, we can add the actual SECONDARY ROLES later after we got requirements from users.

Now we have #12279 

related:

- #12279

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12299)


tests/suites/0_stateless/20+_others/20_0012_privilege_access.result
```diff
@@ -1,34 +1,34 @@
...
+ERROR 1105 (HY000) at line 1: PermissionDenied. Code: 1063, Text = Permission denied, privilege [Select] is required on 'default'.'system'.'fuse_snapshot' for user 'test-user'@'%' with roles [public].
 1
-ERROR 1105 (HY000) at line 1: PermissionDenied. Code: 1063, Text = Permission denied, privilege [Select] is required on 'default'.'system'.'fuse_segment' for user 'test-user'@'%' with role public.
+ERROR 1105 (HY000) at line 1: PermissionDenied. Code: 1063, Text = Permission denied, privilege [Select] is required on 'default'.'system'.'fuse_segment' for user 'test-user'@'%' with roles [public].
 1
-ERROR 1105 (HY000) at line 1: PermissionDenied. Code: 1063, Text = Permission denied, privilege [Select] is required on 'default'.'system'.'fuse_block' for user 'test-user'@'%' with role public.
+ERROR 1105 (HY000) at line 1: PermissionDenied. Code: 1063, Text = Permission denied, privilege [Select] is required on 'default'.'system'.'fuse_block' for user 'test-user'@'%' with roles [public].
```
TCeason, line 31: Better to add this test:

user foo has two or more roles: test_role1, teest_role2 ... which without select privilege.

And then execute select default'.'system'.'fuse_block check the privilege access.

b41sh (approved):

Issue #12240, opened 2023-07-28
flaky test: mysql auth


xudong963:
**Summary**

https://github.com/datafuselabs/databend/actions/runs/5689016271/job/15420114001

TCeason:
This test is duplicate with 20_0012_privilege_access.sh. 

So we can drop this test: 20_0005_mysql_auth.py cc @youngsofun

youngsofun:
> This test is duplicate with 20_0012_privilege_access.sh.
> 
> So we can drop this test: 20_0005_mysql_auth.py cc @youngsofun

not totoally duplicated, some recent tests such like allowed_ip_list is added to 20_0005_mysql_auth.py.
maybe we can move most cases to `.sh` (maybe need a in comment in `.py`), 
while `.py` seems to be easier to test connecting.
also, maybe they covered different server-side code path, or triger diff failure, just like this flaky test.

TCeason:
20_0005_mysql_auth:                                                     [ FAIL ] - result differs with:
--- /runner/_work/databend/databend/tests/suites/0_stateless/20+_others/20_0005_mysql_auth.result	2023-08-02 07:17:53.185617239 +0000
+++ /runner/_work/databend/databend/tests/suites/0_stateless/20+_others/20_0005_mysql_auth.stdout	2023-08-02 07:20:50.690085741 +0000
@@ -1 +1,2 @@
+u1 is timeout
 u3 is blocked by client ip

cc @b41sh 
https://github.com/datafuselabs/databend/actions/runs/5735753964/job/15544363718?pr=12298

PR #12328, opened 2023-08-02
chore(ci): fix flaky

TCeason:

## Summary

Summary about this PR

- Closes #https://github.com/datafuselabs/databend/issues/12240#issuecomment-1661810274

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12328)


b41sh (approved):
The first connection after `Databend query` reboot is a bit slow and may timeout. After testing we found that setting `connection_time` to 3s will not timeout in most cases. We can keep an eye on subsequent tests to see if there will be any more timeouts.

PR #12330, opened 2023-08-02
fix(rbac): do not check the privilege about SET ROLE

flaneur2020:

## Summary

`SET ROLE` is a session-local statement (have same semantic with the `SET ROLE` in postgres: https://www.postgresql.org/docs/current/sql-set-role.html), it's not considered to be protected via the `GRANT` privilege.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12330)


src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -593,6 +592,7 @@ impl AccessChecker for PrivilegeAccess {
                     .await?;
             }
             // Note: No need to check privileges
+            Plan::SetRole(_) => {}
```
TCeason, line 596: Add more comment?

TCeason (approved):

PR #12335, opened 2023-08-02
refactor: refine replace into, utilizing bloom filter

dantengsky:

## Summary

- refine replace into, utilizing bloom filter.

  during the execution of replace-into,  if the table is clustered, one of the on-conflict columns that have the greatest NDV will be chosen(if the type of column is supported by the bloom index). later during block pruning, if range pruning can not filter the blocks out, and the block has bloom index of the column we choose,  this bloom index will be used.

- tweak metrics

  new metrics `replace_into_block_number_bloom_pruned`
- add a new setting `enable_replace_into_bloom_pruning`

  by default, enabled 

also tested by the "long run" test script.

------

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12335)


src/query/settings/src/settings_default.rs
```diff
@@ -352,6 +352,12 @@ impl DefaultSettings {
                     possible_values: None,
                     display_in_show_settings: true,
                 }),
+                ("enable_replace_into_bloom_pruning", DefaultSettingValue {
```
sundy-li, line 355: maybe we should unified a setting named `enable_bloom_pruning`, which could be used in query/replace/delete/update.

dantengsky: sounds reasonable to me, maybe we can apply this idea later?

for this PR, we just need a flag to toggle the bloom_pruning behavior while executing replace-into.

SkyFan2002 (approved):

dantengsky:
`tests/sqllogictests/suites/mode/cluster/lazy_read.test:69 failed`, not related to this PR, seem to be flaky

https://github.com/datafuselabs/databend/actions/runs/5752463428/job/15593924405?pr=12335#step:4:1016

PR #12351, opened 2023-08-03
refactor(meta): persist snapshot instead of state-machine

drmingdrmer:

##### refactor(meta): persist snapshot instead of state-machine

Make state machine a pure in-memory store and the durability relies on
persistent snapshot.

Upgrade to openraft v0.8.4-alpha.8: with the ability to persist
`committed` log id, so that upon startup, state machine will be able to
replay all committed log to restore to the latest state.

### Compatibility

This commit add a new data version V002, which depends on upgrading upon
startup to provide compatibility.

New binary can run directly with old version data, `V0` or `V001`.
**Rolling upgrade** is allowed.

`databend-metactl` is also upgraded that can import `V0`, `V001` and
`V002` data, and export `V002` data. Old version data will be upgraded
immediately once imported.

#### Incompatible snapshot format

Snapshot format changed thus during rolling upgrading, it requires all
node data to be up to date, ensure there is no need to replicate with
snapshot.

### New State machine

The new state machine is a multi level in-memory map: only the top level
accepts modifications, the lower levels are read-only. This way building
snapshot won't block writing.

### Snapshot

When a new snapshot is built, it is persisted on disk. And the last 3
snapshot will be kept, older snapshot will be removed.

Snapshot file contains lines of json string. The first line is a
data-version header(also a json) to identify the verson of the snapshot.
The snapshot version is the same as the program data-version, e.g.,
`V002`.

### Other changes

- meta-cluster-status(/v1/cluster/status): add 2 fields: snapshot_last_log_id and purged

### Benchmark of upsert table with copied file

In the benchmark, every request upserts a table and writes multiple **copied file** records.

The total size of the written data is ~ 2G(the snapshot size).

### 1 client

| N.O. file/upsert  | 1 client | Total          |
|------------------:|---------:|        --:     |
|                 1 |  270 rps | 270 copies/sec |
|               100 |   80 rps | 8K  copies/sec |
|              1000 |    8 rps | 8K  copies/sec |
|              5000 |    error |                |

### 100 clients

| N.O. file/upsert  | 100 client |          Total |
|------------------:|-----------:|---------------:|
|                 1 |    700 rps |  7K copies/sec |
|               100 |     80 rps |  8K copies/sec |
|               500 |     40 rps | 20K copies/sec |
|              1000 |      error |                |
|              5000 |      error |                |

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12351)


ariesdevil:
This PR should be tested by the "long run" test script, cc @dantengsky

ariesdevil:
*[`src/binaries/metactl/snapshot.rs` line 108 at r1](https://reviewable.io/reviews/datafuselabs/databend/12351#-NawSEUG3RLDKp8qX0d6:-NawSEUG3RLDKp8qX0d7:b-nnojww) ([raw file](https://github.com/datafuselabs/databend/blob/724df61536d0cc46896a53238adfa0cf8671fb28/src/binaries/metactl/snapshot.rs#L108)):*
> ```XML
>         .ok_or_else(|| anyhow::anyhow!("no data to import"))?;
> 
>     let first_line = match first {
> ```

the `first` already checked error above, why need to check again here?

BohuTANG:
Can we have a benchmark compare with the previous version? 
We can do COPY with many files and check.

drmingdrmer (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12351)* status: 0 of 95 files reviewed, 1 unresolved discussion (waiting on @ariesdevil and @lichuang)

___
*[`src/binaries/metactl/snapshot.rs` line 108 at r1](https://reviewable.io/reviews/datafuselabs/databend/12351#-NawSEUG3RLDKp8qX0d6:-NayHQHO7BWCr3rERkhJ:b-p6nkft) ([raw file](https://github.com/datafuselabs/databend/blob/724df61536d0cc46896a53238adfa0cf8671fb28/src/binaries/metactl/snapshot.rs#L108)):*
<details><summary><i>Previously, ariesdevil (Yijun Zhao) wrote…</i></summary><blockquote>

the `first` already checked error above, why need to check again here?
</blockquote></details>

The first check is to ensure there is at least one line, the second is to ensure there is no error when reading the line.

drmingdrmer:
> Can we have a benchmark compare with the previous version? We can do COPY with many files and check.

A simple `upsert_kv` benchmark shows that the new version handles 1000 RPS while the former version handles 600 RPS.

And what's more important is that building snapshot won't block client write anymore.

drmingdrmer (commented):
> Reviewed 95 of 95 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12351)* status: all files reviewed, 1 unresolved discussion (waiting on @ariesdevil and @lichuang)

drmingdrmer (commented):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12351)* status: all files reviewed, 1 unresolved discussion (waiting on @ariesdevil and @lichuang)

drmingdrmer (commented):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12351)* status: all files reviewed, 1 unresolved discussion (waiting on @ariesdevil and @lichuang)

ariesdevil (approved):

lichuang:
@BohuTANG  @flaneur2020 

This PR use a in memory map instead of sled to represent raft application statemachine, 

Advantages:
when build snapshot, it will not block the process.

disdvantages:
1. when meta data become larger, meta server memory will larger too.
2. when startup, meta will sync on disk data to memory map.

drmingdrmer:
There is another problem I found:
Every time after building a snapshot, client-write will be blocked for 1~3 seconds.
It may be caused by heavy load fsync the snapshot file, or purging log after building snapshot.

I can not yet figure out the root cause because the log file is messed up, which looks like caused by concurrent log writes.

Do not merge this patch until these problems are solved.

drmingdrmer:
I'm merging this PR and the blocking issue and the log issue are going to be solved later. 
Before that, databend-meta can be deployed in testing environment but it's not recommended to be used in production env.
@everpcpc @flaneur2020

flaneur2020:
@drmingdrmer @lichuang Is storing the metadata in memory considered a long-term solution? I guess we can have some further talk about this design decision.

drmingdrmer:
> @drmingdrmer @lichuang Is storing the metadata in memory considered as a long-term solution? I guess we can have some further talk about this design decision.

No. If the storage requirements increase significantly, a SSTable based design will be considered

flaneur2020:
> > @drmingdrmer @lichuang Is storing the metadata in memory considered as a long-term solution? I guess we can have some further talk about this design decision.
> 
> No. If the storage requirements increase significantly, a SSTable based design will be considered

Sorry I guess it's pretty dangerous to be rolled out to production :( . I'll pause deploying this new in-memory storage version before we have enough evaluation about the memory consumption.

drmingdrmer:
The copied-file record is about 200 bytes/record. 1 million copied-file takes about 200MB memory.

It looks OK, not?

PR #12353, opened 2023-08-03
feat(query): mysql handler tls support

### LINKED ISSUE #8983: Feature: tls support for mysql handler
**Summary**

The just released opensrv-mysql v0.3.0 includes support for tls and it is time to introduce it for Databend.

Ref:

- https://github.com/datafuselabs/opensrv/pull/34
- https://github.com/GreptimeTeam/greptimedb/pull/641

akoshchiy:

## Summary

Summary about this PR

- Closes #8983

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12353)


akoshchiy:
1. There is no client authentication, is it enough or it's better to add it? As I understood, client auth also requires "no password" login as mysql does.
2. Seems that `ci-run-stateless-tests-cluster-tls.sh` is not using in CI check now. Should I create my own tests with mysql client? Added some integration test in `mysql_handler.rs`, but I think it's also a good idea to test with the mysql client with enabled tls mode.

zhang2014 (approved):

zhang2014:
Thanks for your contribution

> 1. There is no client authentication, is it enough or it's better to add it? As I understood, c also requires "no password" login as mysql does.

yes, support client auth is better(welcome pr for this). BTW: `no password` login is disabled by default in databend

> 2. Seems that `ci-run-stateless-tests-cluster-tls.sh` is not using in CI check now. Should I create my own tests with mysql client? Added some integration test in `mysql_handler.rs`, but I think it's also a good idea to test with the mysql client with enabled tls mode.

integration test is Ok.

PR #12174, opened 2023-07-24
fix: sqllogic unexpected pass error

JackTan25:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR
https://github.com/risinglightdb/sqllogictest-rs/issues/178
fix at https://github.com/risinglightdb/sqllogictest-rs/pull/187
- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12174)

PR #12355, opened 2023-08-04
fix(query): bump sqllogic crate

sundy-li:

## Summary

Summary about this PR

- Continue #12174

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12355)


src/query/sql/src/executor/physical_plan_builder.rs
```diff
@@ -1605,7 +1605,7 @@ impl PhysicalPlanBuilder {
                     .reduce(|lhs, rhs| {
                         ScalarExpr::FunctionCall(FunctionCall {
                             span: None,
-                            func_name: "and".to_string(),
+                            func_name: "and_filters".to_string(),
```
xudong963, line 1608: What's the different between the two funcs?

sundy-li: `and_filters` will judge the mask if the error happens,  `number > 1 and  3 /(number - 1)` will not throw div by zero error because `number - 1` will not be zero in the first condition.

xudong963 (approved):

PR #12376, opened 2023-08-07
feat:  auto re-cluster for clustered tables

dantengsky:

## Summary

- introduces new setting option `enable_recluster_after_write`,  default 1
- auto compact table while executing `copy into`, `replace into`, if
  - table defines cluster key
  - and `enable_recluster_after_write` is set to `

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12376)


BohuTANG:
Does the 'auto recluster' is doing a recluster after replace into?
If so, it's better to change the 'enable_auto_reclustering' to 'enable_recluster_after_write'?

dantengsky:
> Does the 'auto recluster' is doing a recluster after replace into? If so, it's better to change the 'enable_auto_reclustering' to 'enable_recluster_after_write'?

yes, if set to 1,  after both `replace-into` and  `copy-into`, the table will be compacted and re-clustered.

rename 'enable_auto_reclustering' to 'enable_recluster_after_write' just for `replace-into`  statement? so that they could be configed individually?

dantengsky:
tested using https://github.com/dantengsky/rr , iteration 3000, passed

zhyass (approved):

PR #12380, opened 2023-08-08
feat(query): Support values Clause

### LINKED ISSUE #12321: please add  ` (values( ))` syntax into subquery and CTE
### Discussed in https://github.com/datafuselabs/databend/discussions/11576

<div type='discussions-op-text'>

<sup>Originally posted by **l1t1** May 25, 2023</sup>
```
root@localhost:~# duckdb
v0.7.1 b00b93f
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D with t(a,b) as (values(1,1),(2,null),(null,5))
> select grouping(a) gid,a,sum(b)from t group by rollup(a);
┌───────┬───────┬────────┐
│  gid  │   a   │ sum(b) │
│ int64 │ int32 │ int128 │
├───────┼───────┼────────┤
│     1 │       │      6 │
│     0 │     1 │      1 │
│     0 │     2 │        │
│     0 │       │      5 │
└───────┴───────┴────────┘
D select grouping(a) gid,a,sum(b)from (values(1,1),(2,null),(null,5))t(a,b) group by rollup(a);
┌───────┬───────┬────────┐
│  gid  │   a   │ sum(b) │
│ int64 │ int32 │ int128 │
├───────┼───────┼────────┤
│     1 │       │      6 │
│     0 │     1 │      1 │
│     0 │     2 │        │
│     0 │       │      5 │
└───────┴───────┴────────┘
```</div>

### PR DESCRIPTION
**Pull request by b41sh**

## Summary

Support values Clause

- Closes #12321

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12380)

### CODE REVIEW

File: tests/sqllogictests/suites/query/cte.test [hunk 1]
```diff
@@ -416,3 +416,17 @@ query III
 with cte as (select a, a + 1 as X, b + 1 as Y from t1 order by a limit 3) select X,Y,a from (select * from cte order by Y limit 2) order by X limit 1;
 ----
 105 101 104
+
+query IT
+with t(a) as (values(1,'a'),(2,'b'),(null,'c')) select a, col1 from t
```
sundy-li, line 421: `col1` is not specified, why not throw the error if the alias does not match the column number?

src/query/sql/src/planner/binder/table.rs
```diff
@@ -594,7 +601,188 @@ impl Binder {
...
+                }
+                alias
+                    .columns
+                    .iter()
+                    .map(|c| c.to_string())
+                    .collect::<Vec<_>>()
+            }
+            None => vec![],
+        };
+        // For columns whose names are not specified, assigns default column names col0, col2, etc.
```
sundy-li, line 666: we can throw the error if number is not matched.

see: https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-values.html


 If table_alias specifies column names, their number must match the number of expressions per tuple.

b41sh: Both `DuckDB` and `PostgreSQL` support automatically assigning column names to columns that don't have a name, so it's best to be compatible with them.

sundy-li (approved):

xudong963 (commented):
I'm reviewing.

src/query/sql/src/planner/binder/table.rs
```diff
@@ -565,9 +573,191 @@ impl Binder {
...
+        for i in names.len()..num_cols {
+            names.push(format!("col{}", i));
+        }
+
+        let mut bind_context = BindContext::with_parent(Box::new(bind_context.clone()));
```
xudong963, line 633: Seems there is no need to create a new `bind_context`.

tests/sqllogictests/suites/query/select.test
xudong963, line 1:
```
select col0, col1 from (values(1, 'one'), (null, 'two'), (3 + 2, null))
```
I test the case in my local env, and it works. How about  adding it to the test and avoiding drawbacks in the future (if someone changes the scoped code)

src/query/sql/src/executor/profile.rs
```diff
@@ -491,6 +491,7 @@ fn flatten_plan_node_profile(
             plan_node_profs.push(prof);
         }
         PhysicalPlan::MaterializedCte(_) => todo!(),
+        PhysicalPlan::ConstantTableScan(_) => todo!(),
```
xudong963, line 494:
```suggestion
        PhysicalPlan::MaterializedCte(_) | PhysicalPlan::ConstantTableScan(_) => todo!(),
```

src/query/sql/src/planner/plans/constant_table_scan.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn derive_physical_prop(&self, _rel_expr: &RelExpr) -> Result<PhysicalProperty> {
+        Ok(PhysicalProperty {
+            distribution: Distribution::Serial,
+        })
+    }
+
+    fn derive_cardinality(&self, _rel_expr: &RelExpr) -> Result<Arc<StatInfo>> {
+        Ok(Arc::new(StatInfo {
+            cardinality: self.num_rows as f64,
+            // TODO
```
xudong963, line 107: Let's open an issue?

src/query/sql/src/planner/plans/constant_table_scan.rs
```diff
@@ -0,0 +1,121 @@
...
+use crate::optimizer::RequiredProperty;
+use crate::optimizer::StatInfo;
+use crate::plans::Operator;
+use crate::plans::RelOp;
+
+// Constant table is a table with constant values.
+#[derive(Clone, Debug)]
+pub struct ConstantTableScan {
+    pub values: Vec<Column>,
+    pub num_rows: usize,
```
xudong963, line 37: I'm not sure if the field is necessary, seems we can get it from `values[0].len()`.

b41sh: `num_rows` is needed, as `values` may be empty after `UnusedColumnPruner` optimization.
for example
```sql
mysql> select 1 from (values('a'),('b'),('c'));
+------+
| 1    |
+------+
|    1 |
|    1 |
|    1 |
+------+
```

src/query/sql/src/planner/plans/constant_table_scan.rs
b41sh: ok, good idea.

src/query/sql/src/planner/binder/table.rs
b41sh: Can we use `let mut bind_context = bind_context.clone();`? Because we need to return `BindContext`, but the original `bind_context` is `&BindContext`

src/query/sql/src/planner/binder/table.rs
```diff
@@ -565,9 +573,191 @@ impl Binder {
...
+                values,
+                alias,
+            } => self.bind_values(bind_context, *span, values, alias).await,
         }
     }
 
+    #[async_backtrace::framed]
+    pub(crate) async fn bind_values(
+        &mut self,
+        bind_context: &BindContext,
```
xudong963, line 587: You can change it to `bind_context: &mut BindContext`, because `self.bind_values(bind_context, *span, values, alias).await,` the `bind_context` is `&mut`

src/query/sql/src/planner/binder/table.rs
xudong963: https://github.com/datafuselabs/databend/pull/12380#discussion_r1288196126

xudong963 (approved):

BohuTANG:
@soyeric128 for documentation, as DuckDB: https://duckdb.org/docs/sql/query_syntax/values.html

PR #12383, opened 2023-08-08
docs: Update databend-query.toml

wubx:

## Summary
Enable a default account

- Closes #issue

### DIFF
--- a/scripts/distribution/configs/databend-query.toml
+++ b/scripts/distribution/configs/databend-query.toml
@@ -35,9 +35,9 @@ cluster_id = "default"
 
 table_engine_memory_enabled = true
 
-# [[query.users]]
-# name = "root"
-# auth_type = "no_password"
+[[query.users]]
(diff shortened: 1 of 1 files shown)


PsiACE (commented):
Although it may bring convenience, there may be security issues. I think we need to emphasize avoiding this usage as much as possible in the production environment.

PsiACE (approved):
cc @soyeric128 , I think we need to highlight it specifically in the document

b41sh:
@wubx Automatically enabling root user without a password can be potentially risky. How about adding a separate config file like `databend-query-test.toml` with the root user just for testing?

soyeric128:
> Although it may bring convenience, there may be security issues. I think we need to emphasize avoiding this usage as much as possible in the production environment.

I feel the same way.

BohuTANG:
I close this PR because the security issue.

PR #12385, opened 2023-08-08
fix(logging): log messages are lost with heavy load

drmingdrmer:

##### fix(logging): log messages are lost with heavy load

`tracing_appender::non_blocking()` will drop log messages if the channel
is full. Replace it with a `BufWriter<_>`

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12385)

### DIFF
--- a/src/common/tracing/src/minitrace.rs
+++ b/src/common/tracing/src/minitrace.rs
@@ -14,6 +14,8 @@
 
 use std::borrow::Cow;
 use std::fmt;
+use std::io;
+use std::io::BufWriter;
 use std::io::Write;
 use std::time::SystemTime;
 
@@ -23,6 +25,8 @@ use log::LevelFilter;
 use log::Log;
 use minitrace::prelude::*;
 use serde_json::Map;
+use tracing_appender::non_blocking::NonBlocking;
+use tracing_appender::non_blocking::WorkerGuard;
 use tracing_appender::rolling::RollingFileAppender;
 use tracing_appender::rolling::Rotation;
 
@@ -109,11 +113,8 @@ pub fn init_logging(name: &str, cfg: &Config) -> Vec<Box<dyn Drop + Send + Sync 
     // File logger
     if cfg.file.on {
         let (normal_log_file, flush_guard) =
-            tracing_appender::non_blocking(RollingFileAppender::new(
-                Rotation::HOURLY,
-                &cfg.file.dir,
-                format!("databend-query-{name}"),
-            ));
+            new_file_log_writer(&cfg.file.dir, format!("databend-query-{name}"));
+
         guards.push(Box::new(flush_guard));
 
         normal_logger = normal_logger.chain(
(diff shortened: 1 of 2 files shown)


Xuanwo (approved):
Nice catch!

src/common/tracing/src/minitrace.rs
```diff
@@ -108,13 +109,13 @@ pub fn init_logging(name: &str, cfg: &Config) -> Vec<Box<dyn Drop + Send + Sync
...
-                format!("databend-query-{name}"),
-            ));
-        guards.push(Box::new(flush_guard));
+        let rolling = RollingFileAppender::new(
+            Rotation::HOURLY,
+            &cfg.file.dir,
+            format!("databend-query-{name}"),
+        );
+
+        let normal_log_file = io::BufWriter::with_capacity(64 * 1024 * 1024, rolling);
```
Xuanwo, line 118: By the way, without flush guard, we could lost recent logs if application termination happended (at most 64MiB).

Xuanwo: Maybe it's better to wrap non_blocking appender with `BufWriter`.

andylokandy:
`BufWriter` could lead to infrequent but long halt randomly during the execution.

src/common/tracing/src/minitrace.rs
Xuanwo: Or we can increase the `buffered_lines_limit` from 128 Ki to larger, maybe 64 Mi (like we did in this PR)

drmingdrmer: `BufWriter` itself is a guard and will flush the buf in blocking mode when being dropped, I think that would be good enough?
```rust
#[stable(feature = "rust1", since = "1.0.0")]
impl<W: Write> Drop for BufWriter<W> {
    fn drop(&mut self) {
        if !self.panicked {
            // dtors should not panic, so we ignore a failed flush
            let _r = self.flush_buf();
        }
    }
}
```

andylokandy: image
When the buffer capacity is reached, `BufWriter` will perform a blocking flush.

drmingdrmer:
@andylokandy @Xuanwo 
Roger that, it looks a `BufWriter` upon `NonBlocking` would be better!

Thank you my heros!

andylokandy:
Like @Xuanwo has mentioned, we could increase the `buffere_lines_limit` and also set `lossy` to `false`

image

src/common/tracing/src/minitrace.rs
Xuanwo: > `BufWriter` itself is a guard and will flush the buf in blocking mode when being dropped, I think that would be good enough?

Oh, I don't know about this. Lesson learnt.

drmingdrmer:
> Like @Xuanwo has mentioned, we could increase the `buffere_lines_limit` and also set `lossy` to `false`
> 
> image

The default `buffere_lines_limit` is `128_000` which I think is already big enough but I still encountered log loss.

Another issue is that log messages are not sent by `fern` in a whole to the `Writer`. The following is the `strace` output when logging. Too many `write()` will still be a performance issue and should be avoided. Thus `fern -> BufWriter -> NonBlocking -> File` would be better.

0c86d994-df7d-4160-aa1c-cbd4546395bf

In a non `lossy` mode, `NonBlocking::write()` will return an error. It may not be expected.
```rust
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let buf_size = buf.len();
        if self.is_lossy {
            if self.channel.try_send(Msg::Line(buf.to_vec())).is_err() {
                self.error_counter.incr_saturating();
            }
        } else {
            return match self.channel.send(Msg::Line(buf.to_vec())) {
                Ok(_) => Ok(buf_size),
                Err(_) => Err(io::Error::from(io::ErrorKind::Other)),
            };
        }
        Ok(buf_size)
    }
```

andylokandy (approved):

andylokandy (approved):

Issue #12405, opened 2023-08-10
Feature:  meter `BloomIndexFilterCache` by bytes


dantengsky:
**Summary**

currently cache `BloomIndexFilterCache` is metered by number of cached item, which is not convenient for user (to evaluation the memory usage of cache).

https://github.com/datafuselabs/databend/blob/7faf160ab303ad72b9b53b7f4e1f8d10ada9f1dd/src/query/storages/common/cache-manager/src/caches.rs#L49

The metering of `CompactSegmentInfoCache` is better: 

https://github.com/datafuselabs/databend/blob/7faf160ab303ad72b9b53b7f4e1f8d10ada9f1dd/src/query/storages/common/cache-manager/src/caches.rs#L39-L41

PR #12416, opened 2023-08-10
chore: BloomIndexFilterCache meters as cache size instead of item

lichuang:

## Summary

chore: BloomIndexFilterCache meters as cache size instead of item

- Closes https://github.com/datafuselabs/databend/issues/12405

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12416)


src/query/storages/common/cache-manager/src/cache_manager.rs
```diff
@@ -104,8 +106,11 @@ impl CacheManager {
                 CompactSegmentInfoMeter {},
                 "segment_info",
             );
-            let bloom_index_filter_cache =
-                Self::new_item_cache(config.table_bloom_index_filter_count, "bloom_index_filter");
+            let bloom_index_filter_cache = Self::new_in_memory_cache(
+                config.table_bloom_index_filter_count * std::mem::size_of::<Xor8Filter>() as u64,
```
dantengsky, line 110: afraid that we can not simply multiply by the size of type `Xor8Filter`, please see the comment of 
 
 `impl Meter<String, Arc<Xor8Filter>> for BloomIndexFilterMeter`.
 
 
 **maybe** we need to introduce new config item `table_bloom_index_filter_size`, which user can specify the max size (in bytes) of cached bloom index filters? and deprecate this config option?.
 
 but that may introduce incompatible changes, if user has non-default  (explicit) config item of `table_bloom_index_filter_count`.

lichuang: introduce new_item_cache_with_meter, it use meter instead of count to meter cache size

dantengsky (approved):
LGTM, let's add new config option in dedicated PR

PR #12425, opened 2023-08-10
feat: Virtual column support parquet format

b41sh:

## Summary

- Virtual column support read parquet format
- Using `VirtualColumnReader` to read virtual columns
- Ignore reading source columns if all virtual columns are generated

Part of #6994

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12425)


RinChanNOWWW:
This PR is to support store virtual columns as parquet format?

b41sh:
> This PR is to support store virtual columns as parquet format?

The format of the virtual column block is the same as the corresponding table block format. Previously, only table of native type was supported, after this PR table of parquet type is also supported.

sundy-li (approved):

lichuang (approved):

BohuTANG:
The unit test always failed in this PR

PR #12446, opened 2023-08-12
fix: memory leak in unsafe code

### LINKED ISSUE #12445: bug: memory leak in unsafe code
**Summary**
https://github.com/datafuselabs/databend/blob/b6cb56baa276b9c0c794fd83ec3851eeba5c0252/src/query/expression/src/block.rs#L90-L114
The memory is not freed after `Box::into_raw()`

SkyFan2002:

## Summary

Summary about this PR

- Closes #12445

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12446)

### DIFF
--- a/src/query/expression/src/block.rs
+++ b/src/query/expression/src/block.rs
@@ -89,16 +89,15 @@ pub trait BlockMetaInfoDowncast: Sized {
 
 impl<T: BlockMetaInfo> BlockMetaInfoDowncast for T {
     fn downcast_from(boxed: BlockMetaInfoPtr) -> Option<Self> {
-        if !boxed.as_any().is::<T>() {
-            return None;
+        if boxed.as_any().is::<T>() {
+            unsafe {
+                // SAFETY: `is` ensures this type cast is correct
+                let raw_ptr = Box::into_raw(boxed) as *mut dyn BlockMetaInfo;
+                let typed_ptr = raw_ptr as *mut Self;
+                return Some(*Box::from_raw(typed_ptr));
+            }
         }
-        unsafe {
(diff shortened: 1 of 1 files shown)


dantengsky:
`jeprof` find no memory with this PR, for the simulated copy-into test scenario.


2023-08-12-183340_1115x91_scrot

dantengsky (approved):
let's merge

src/query/expression/src/block.rs
```diff
@@ -92,24 +92,16 @@ impl<T: BlockMetaInfo> BlockMetaInfoDowncast for T {
         if boxed.as_any().is::<T>() {
             unsafe {
                 // SAFETY: `is` ensures this type cast is correct
-                let raw_ptr = Box::into_raw(boxed) as *const dyn BlockMetaInfo;
-                return Some(std::ptr::read(raw_ptr as *const Self));
+                let raw_ptr = Box::into_raw(boxed) as *mut dyn BlockMetaInfo;
+                let typed_ptr = raw_ptr as *mut Self;
+                return Some(*Box::from_raw(typed_ptr));
```
andylokandy, line 97: it'll be more consice if we write:
```
let downcast_box = std::mem::transmute::<Box<_>, Box< Self>>(boxed);
*downcast_box
```

SkyFan2002: Good idea! BTW, I want to discuss, what about removing unsafe code by writing somthing like `boxed.into_any().downcast()`

andylokandy: it may be not able to take ownership from Box<Any>, but we can clone it anyway.

PR #12478, opened 2023-08-16
fix(rbac): do not require GRANT privilege on SHOW ROLES statement

flaneur2020:

## Summary

The SHOW ROLES statement displays the available roles for the current session. It's available for every one, do not require the GRANT privilege.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12478)


TCeason (approved):

TCeason:
The doc also need to modify

BohuTANG:
We should add a logic test for 'show role'

PR #12482, opened 2023-08-16
feat(query): improve inner join

Dousir9:

## Summary

1. When the type of hash join is inner join and there is an output column that is the join key of hash join, we don’t need to save this column on the build side, just use the column of probe side to build it, on TPC-H SF100, it will reduce memory usage by 1.1 GB, TPC-H Q18: 20.12s -> 18.82s.
2. Simplify the generation of datablocks during hash join.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12482)


src/query/sql/src/executor/physical_plan.rs
```diff
@@ -615,6 +615,8 @@ pub struct HashJoin {
     pub join_type: JoinType,
     pub marker_index: Option<IndexType>,
     pub from_correlated_subquery: bool,
+    pub probe_to_build: Vec<(usize, (bool, bool))>,
```
xudong963, line 618: Please add comments for the field

Dousir9: Done.

src/query/service/src/pipelines/processors/transforms/hash_join/row.rs
```diff
@@ -25,19 +25,9 @@ use parking_lot::RwLock;
...
-    pub data_block: DataBlock,
-}
-
-impl Chunk {
-    pub fn num_rows(&self) -> usize {
-        self.data_block.num_rows()
-    }
-}
-
 pub struct RowSpace {
```
xudong963, line 28: It feels like this could be changed: `build_schema` is moved out. Then inside HashJoinState is `Arc<RwLock<RowSpace>>`, then `buffer` and `buffer_row_size` don't need locks, and `write_lock` doesn't need one. `gather `is not used as a method for `RowSpace`.

xudong963 (approved):

BohuTANG:
```
Crate:     spin
Version:   0.9.6
Warning:   yanked
```

we can ignore it in: https://github.com/datafuselabs/databend/blob/main/.cargo/audit.toml

PR #12494, opened 2023-08-17
feat: shuffle blocks instead of segments in distributed execution of repalce into

SkyFan2002:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12494)


src/query/service/src/interpreters/interpreter_replace.rs
```diff
@@ -212,19 +213,27 @@ impl ReplaceInterpreter {
             table_level_range_index,
             need_insert: true,
         }));
+        let target = if is_distributed {
```
zhang2014, line 216: different blocks of the same segment will be executed on different nodes? will they generate different segments after replaced?

PR #12500, opened 2023-08-18
feat: block level shuffle of dist `replace into`

dantengsky:

## Summary

instead of (shuffling and) splitting table segments into partitions, and assigning the partitions to executors, 
description of which blocks should be taken care of is assigned to each executor,  in a manner that 
- one block will only be processed by one executor
- blocks of one segment will be evenly spread to executors

consequently,  one executor has to process all the segments of a given table, but for normal cases, large majorities of segments could be pruned at the segment level (which is lightweight, without unpacking the `CompactSegmentInfo`).

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12500)


dantengsky:
tested with "long-run scripts", 3000 iterations, for both standalone and dist setting

SkyFan2002 (approved):

PR #12519, opened 2023-08-21
refactor: eliminate ambiguous cache API

drmingdrmer:

##### Refactor: eliminate ambiguous cache API

In the `Cache` trait, the `get_mut()` and `peek_mut()` methods return
mutable references that permit users to modify the stored value.
However, when the updated value is given back into the cache, the
eviction policy determined by `Meter` is not validated against these
changes. This could potentially lead to risks, such as the cache
surpassing its maximum capacity.

Given that these two APIs are currently unused, my suggestion is to
remove them to prevent potential misuse in the future.

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12519)

### DIFF
--- a/src/common/cache/src/cache.rs
+++ b/src/common/cache/src/cache.rs
@@ -38,25 +38,10 @@ where
         K: Borrow<Q>,
         Q: Hash + Eq + ?Sized;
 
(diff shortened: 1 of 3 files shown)


PsiACE (approved):
lgtm

Issue #12465, opened 2023-08-15
Tracking: refactor parquet table.


RinChanNOWWW:
**Summary**

- [x] Basic refactor of parquet reader. #12424
- [x] Row groups and pages pruning. #12473
- [x] Reading file metas during `read_partitions` to collect stats, check schemas, ... #12569
- [x] Row group level partition. #12569
- [x] Optimize small parquet files reading. #12637
- [x] Introduce Top-K pushdown optimization. #12877
- [x] Specify size for one output block. #12523
- [x] Adjust source parallelism by file sizes and compression ratios. #12637

BohuTANG:
The latest benchmark in Fabric with local parquet TCPH_SF10 parquet files:

<img width="1062" alt="image" src="https://github.com/datafuselabs/databend/assets/172204/427b5b87-75b7-43b2-81f2-8a881fb68f6a">

Q1/Q18/Q19 still have room to improve.

RinChanNOWWW:
The default setting uses `parquet2`. Did this benchmark alter to `parquet_rs`?

BohuTANG:
> The default setting uses `parquet2`. Did this benchmark alter to `parquet_rs`?

This is the notebook code:
```
from databend import SessionContext
ctx_bend = SessionContext("TPCH")
ctx_bend.sql(f"""set use_parquet2=1""").collect()
....
```

sundy-li:
> Q1/Q18/Q19 still have room to improve.

This is not about parquet but query execution.

RinChanNOWWW:
How about `set use_parquet2=0` comparing with `set use_parquet2=1`;

BohuTANG:
Regarding @Dousir9's explanation:

Q1 is related to decimal data type aggregate/groupby.
Q18 is related to join order issue.
Q19 is the filter.

They are not only to parquet file also in fuse engine.

BohuTANG:
> How about `set use_parquet2=0` comparing with `set use_parquet2=1`;

`set use_parquet2=0`:
<img width="1290" alt="image" src="https://github.com/datafuselabs/databend/assets/172204/8c064ba8-8fbf-426d-81ff-3d559ba11741">

Dousir9:
Q1 is also related to aggregation implementation.

BohuTANG:
# The latest version(v1.2.136) benchmark

## set use_parquet2=1
[...]

PR #12523, opened 2023-08-21
refactor: refactor parquet source to support specifying batch size.

RinChanNOWWW:

## Summary

Customize `Processor` instead of `AsyncSource` to support output `DataBlock` by `batch_size`.

Tracking in #12465

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12523)


zhang2014 (approved):

b41sh (approved):

src/query/storages/parquet/src/parquet_rs/parquet_reader.rs
```diff
@@ -123,6 +126,8 @@ impl ParquetRSReader {
             options,
         )?;
 
+        let batch_size = ctx.get_settings().get_max_block_size()? as usize;
```
youngsofun, line 129: it seems `max_block_size` is currently used mainly for join, i.e.  temporary blocks.
but when used in CopyIntoTable, it is better to read with the BlockThreadsHolds of the dest fuse table, to avoid split/compact again before write.

RinChanNOWWW: If there isn't a dest table, how can I determine the batch size?

youngsofun: we can use max_block_size then.
we can even impl choosing batch size for copy in next pr.

youngsofun (approved):

PR #12529, opened 2023-08-21
feat(cluster): allow specify a client timeout on creating flight client

flaneur2020:

## Summary

today I met an issue that it takes a long period for a new databend-query instance to get ready for new connections:

Screenshot 2023-08-21 at 20 18 17

as the logs shows, it toke 2 minutes to finally get a timeout on create a new rpc client.

this pr try to fix this hanging issue by adding a timeout config

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12529)

### DIFF
--- a/src/query/config/src/config.rs
+++ b/src/query/config/src/config.rs
@@ -1364,6 +1364,9 @@ pub struct QueryConfig {
     #[clap(long, default_value = "localhost")]
     pub rpc_tls_query_service_domain_name: String,
 
+    #[clap(long, default_value = "0")]
+    pub rpc_client_timeout_secs: u64,
+
     /// Table engine memory enabled
     #[clap(long, parse(try_from_str), default_value = "true")]
     pub table_engine_memory_enabled: bool,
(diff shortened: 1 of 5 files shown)


zhang2014 (approved):

src/query/config/src/config.rs
```diff
@@ -1364,6 +1364,9 @@ pub struct QueryConfig {
     #[clap(long, default_value = "localhost")]
     pub rpc_tls_query_service_domain_name: String,
 
+    #[clap(long, default_value = "0")]
+    pub rpc_client_timeout_secs: u64,
```
BohuTANG, line 1368: Give a reasable default value?

flaneur2020: this default value 0 is intended to keep the same behaviour as before, because I'm a bit afraid of setting a timeout here may expose some additional errors on the communication between the instances, I would try a good configuration value in my environment, and put it default in another PR

PR #12530, opened 2023-08-21
refactor: bring back replace-into segment level shuffling & tweak `replace_into_bloom_pruning_max_column_number`

dantengsky:

## Summary

  
- bring back segment-level shuffling 

  in PR #12500 , the segment-level shuffling strategy is replaced with block-level shuffling, which is not proper at least for situations where the table has a large number of segments, and mutations are not skewed at the block level.
  
  
- introduce new setting `replace_into_shuffle_strategy`

  which could be set to 0 (block-level shuffling) or 1 (segment-level shuffling), the default value is 0.
  
  before we found the heuristic rules of switching strategies automatically, we left options for the user to choose the shuffling strategy.
  
 -  adjust the default value of setting `replace_into_bloom_pruning_max_column_number` from 2 to 4
 
     which according to recent practices, is better

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12530)


SkyFan2002 (approved):

BohuTANG:
`Dev / linux_hive / test_stateful_hive_standalone ` is not related to this pr, let's merge.

PR #12481, opened 2023-08-16
refactor(cluster): active upstream first in cluster scatter

zhang2014:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

refactor(cluster): active upstream first in cluster scatter

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12481)

PR #12532, opened 2023-08-21
Revert "refactor(cluster): active upstream first in cluster scatter"

zhang2014:
Reverts datafuselabs/databend#12481


zhang2014:
Some deadlocks may be related to #12481, let's revert first.

BohuTANG (approved):

PR #12555, opened 2023-08-23
refactor(hive catalog): rename address to hive_metastore

PsiACE:

## Summary

Summary about this PR

- Closes #issue

### DIFF
--- a/src/meta/app/src/schema/catalog.rs
+++ b/src/meta/app/src/schema/catalog.rs
@@ -63,7 +63,7 @@ impl CatalogOption {
 /// Option for creating a iceberg catalog
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct HiveCatalogOption {
-    pub address: String,
+    pub hive_metastore: String,
     pub storage_params: Option<Box<StorageParams>>,
 }
 
--- a/src/meta/protos/proto/catalog.proto
+++ b/src/meta/protos/proto/catalog.proto
@@ -52,7 +52,7 @@ message HiveCatalogOption {
   uint64 min_reader_ver = 101;
 
   // Hive metastore thrift uri
-  string address = 1;
+  string hive_metastore = 1;
   StorageConfig storage_params = 2;
 }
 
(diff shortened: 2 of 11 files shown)


src/meta/app/src/schema/catalog.rs
```diff
@@ -63,7 +63,7 @@ impl CatalogOption {
 /// Option for creating a iceberg catalog
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct HiveCatalogOption {
-    pub address: String,
+    pub hive_metastore: String,
```
BohuTANG, line 66: If we change the `hive_metastore` to `metastore_address` it will more general, user not care this option only for hive, we can re-use it if we have another catalog like hive in feature.

src/meta/protos/proto/catalog.proto
```diff
@@ -52,7 +52,7 @@ message HiveCatalogOption {
   uint64 min_reader_ver = 101;
 
   // Hive metastore thrift uri
-  string address = 1;
+  string hive_metastore = 1;
```
Xuanwo, line 55: NOTE: this is a breaking change. We should add a new field in proto instead of change it directly. I prefer to keep using `address` internally and only change the sql represents.

src/query/config/src/config.rs
```diff
@@ -430,7 +430,7 @@ pub struct CatalogsHiveConfig {
 #[serde(default)]
 pub struct HiveCatalogConfig {
     #[clap(long = "hive-meta-store-address", default_value_t)]
-    pub address: String,
+    pub hive_metastore: String,
```
Xuanwo, line 433: The same.

src/meta/protos/proto/catalog.proto
BohuTANG: Not sure change the name will break in proto, cc @drmingdrmer

drmingdrmer: It's compatible unless the field index(`1`) is changed

src/query/config/src/config.rs
drmingdrmer: This is also a compatible change: `impl FromToProto for HiveCatalogConfig` will use the same field name.

PR #12565, opened 2023-08-23
fix: only check aggregate functions when create aggregating index

### LINKED ISSUE #12564: Feature: support to_start_of_minute in aggregation index
**Summary**

> Currently create aggregating index just support these aggregate functions: sum, min, max, avg.
> I am not able to create the index due to this error
> My usecase is that I want to create this aggregates over dimensions and ? to_start_of_minute(event_time) aggregation
> My query
> CREATE SYNC AGGREGATING INDEX test_agg AS SELECT
>    id,
>    platform,
>    to_start_of_minute(event_time)) AS ts,
>    sum(1) AS count_value
> FROM events
> WHERE (ts > (now() - toIntervalMinute(15))) AND (event_name IN ('ABC', 'XYZ'))
> GROUP BY
>    id,
>    platform,
>    ts

ariesdevil:

## Summary

* only check aggregate functions when creating aggregating index
* disallow non-deterministic funcs like `now()` in aggregating index.
* fix index read with both scalar and agg func, thx @RinChanNOWWW 

Closes #12564

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12565)


RinChanNOWWW (dismissed):

BohuTANG:
It would be better to add a logic test for it(If we had, this bug can not happens before).

sundy-li (approved):

BohuTANG:
> > It would be better to add a logic test for it(If we had, this bug can not happen).↳
> 
> After adding more tests, more bugs were found :(

 🤔 
 
If the PR introduces a new feature, please ensure there is some SQL logic tests. 
Without it, I think the PR shouldn't be approved. 

Kindly cc: @sundy-li and @RinChanNOWWW

ariesdevil:
> > > It would be better to add a logic test for it(If we had, this bug can not happen).↳
> > 
> > 
> > After adding more tests, more bugs were found :(↳
> 
> 🤔
> 
> If the PR introduces a new feature, please ensure there is some SQL logic tests. Without it, I think the PR shouldn't be approved.↳
> 
> Kindly cc: @sundy-li and @RinChanNOWWW

approve is fine, the original pr was already covered by adding test_fuzz, but when I added sqllogictests, I found more problems that I wanted to fix all in this pr.

BohuTANG:
Ensure your SQL logic is validated using the same as the end user. While `test_fuzz` and `unit test` are not substitutes for this validation, consider using them if you're uncertain about your function's reliability.

src/query/sql/src/planner/binder/ddl/index.rs
```diff
@@ -220,12 +226,18 @@ impl Binder {
         Ok(plan)
     }
 
-    fn check_index_support(query: &Query) -> Result<()> {
+    fn check_index_support(query: &Query, has_now_func: bool) -> Result<()> {
+        let now_func = "now";
```
RinChanNOWWW, line 230: Better way: get the function signature and check if `is_deterministic`. Because there are other functions are not deterministic such as `random`.

ariesdevil: Oh yep, I forgot to modify this, nice capture.

sundy-li (approved):

src/query/sql/src/planner/semantic/aggregating_index_visitor.rs
```diff
@@ -137,3 +149,29 @@ impl VisitorMut for AggregatingIndexRewriter {
...
+        _lambda: &'ast Option<Lambda>,
+    ) {
+        if self.has_no_deterministic_func {
+            return;
+        }
+
+        self.has_no_deterministic_func = match BUILTIN_FUNCTIONS.get_property(&name.name) {
+            Some(property) => property.non_deterministic,
+            None => false,
+        };
```
RinChanNOWWW, line 171:
```suggestion
        self.has_no_deterministic_func = BUILTIN_FUNCTIONS.get_property(&name.name).map(|p| p.non_deterministic).unwrap_or(false);
```

PR #12567, opened 2023-08-23
chore: enable benchmark cache

Dousir9:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12567)


zhang2014:
data cache may cause oom of the ci benchmark instance, may we should disable until it's production ready?

BohuTANG:
We should establish a stable benchmark environment that can measure our benchmarks for both releases and PRs. and avoid making frequent changes. If alterations are made, each release from the past can be regenerated if need.

PR #12575, opened 2023-08-24
fix: replace flaky test

JackTan25:

## Summary
we get time out after default 1s. And we will response a State::running to client, if user uses curl command, it will finish and continue to excute next command. Maybe ci machine is too cheap and so slow.
https://github.com/datafuselabs/databend/actions/runs/5960410578/job/16167897950
Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12575)


SkyFan2002 (approved):

dantengsky:
`wait_time_secs` can be specified at client site:


https://github.com/datafuselabs/databend/blob/c8731db1a1085a4bcf5945ece545b6182861cccb/tests/suites/1_stateful/09_http_handler/09_0000_invalid_utf8.sh#L10

JackTan25:
> `wait_time_secs` can be specified at client site:
> 
> https://github.com/datafuselabs/databend/blob/c8731db1a1085a4bcf5945ece545b6182861cccb/tests/suites/1_stateful/09_http_handler/09_0000_invalid_utf8.sh#L10

ok thanks

youngsofun:
I am not sure this is safe, our clients and gateways may think 2s is too long? cc @flaneur2020 @ZhiHanZ 
`return State::running` is an expected behavior and not a problem at all when the client is correctly implemented.
Curl is only for demo and test. I prefer to change the `wait_time` (in JSON body) when using curl.

dantengsky:
besides, most of (if not all of ) the similar test cases that use "http-handler" protocol and assume after the time of `wait_time_secs`, the query/statement could be executed completely, are fundamentally flaky.

is there a more reliable way of testing the http-hanlder-only statements?   bendsql?  @youngsofun

youngsofun:
> besides, most of (if not all of ) the similar test cases that use "http-handler" protocol and assume after the time of `wait_time_secs`, the query/statement could be executed completely, are fundamentally flaky.
> 
> is there a more reliable way of testing the http-hanlder-only statements? bendsql? @youngsofun

I searched in the tests dir.

curl is mainly used in testing for

1.  streaming load
2.  clickhouse handler
3.  http handler : stage attachment, 

only the 3rd is related to this problem.

dantengsky:
@JackTan25 

thanks for opening this PR. 

I suggest to mitigate the flakiness of the  following situation

https://github.com/datafuselabs/databend/actions/runs/5960410578/job/16167897950

by tweaking the `wait_time_secs` of the test case involved,  like this

https://github.com/datafuselabs/databend/blob/c8731db1a1085a4bcf5945ece545b6182861cccb/tests/suites/1_stateful/

youngsofun (approved):

PR #12585, opened 2023-08-25
refactor(query): improve hash join gather

Dousir9:

## Summary

Summary about this PR

- Closes #issue


src/query/expression/src/types.rs
```diff
@@ -281,6 +281,7 @@ pub trait ValueType: Debug + Clone + PartialEq + Sized + 'static {
 
     fn try_downcast_scalar<'a>(scalar: &'a ScalarRef) -> Option<Self::ScalarRef<'a>>;
     fn try_downcast_column<'a>(col: &'a Column) -> Option<Self::Column>;
+    fn try_downcast_column_ref<'a>(col: &'a Column) -> Option<&Self::Column>;
```
sundy-li, line 284: Column's clone is trivial, is this new method benefit lot?

Dousir9: Yes, I found that some of them are cloned instead of clone.

sundy-li (approved):

xudong963 (approved):

PR #12596, opened 2023-08-28
fix: explain pipeline bug of update/delete

### LINKED ISSUE #12597: bug: explain pipeline panic of update/delete
### Search before asking

### Version

[0c5dd54](https://github.com/datafuselabs/databend/commit/0c5dd54b47a2ab94e213d9dd517d7f52d3598071)

### What's Wrong?

mysql> explain pipeline update t1 set a = 1;
ERROR 1105 (HY000): TokioError. Code: 1068, Text = Cannot join handle from context's runtime, cause: task 262 panicked.
mysql> explain pipeline delete from t1 where a = 2;
ERROR 1105 (HY000): TokioError. Code: 1068, Text = Cannot join handle from context's runtime, cause: task 362 panicked.

### How to Reproduce?

_No response_

JackTan25:

## Summary

Summary about this PR
explain pipeline delete/update will cause panic.

- Closes #12597

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12596)

### DIFF
--- a/src/query/settings/src/settings_default.rs
+++ b/src/query/settings/src/settings_default.rs
@@ -154,6 +154,12 @@ impl DefaultSettings {
                     possible_values: None,
                     display_in_show_settings: true,
                 }),
+                ("disable_join_reorder", DefaultSettingValue {
+                    value: UserSettingValue::UInt64(0),
+                    desc: "Disable join reorder optimization.",
+                    possible_values: None,
+                    display_in_show_settings: false,
+                }),
                 ("enable_runtime_filter", DefaultSettingValue {
                     value: UserSettingValue::UInt64(0),
                     desc: "Enables runtime filter optimization for JOIN.",
--- a/src/query/settings/src/settings_getter_setter.rs
+++ b/src/query/settings/src/settings_getter_setter.rs
@@ -232,6 +232,14 @@ impl Settings {
         self.try_set_u64("enable_cbo", u64::from(val))
     }
 
+    pub fn get_disable_join_reorder(&self) -> Result<bool> {
+        Ok(self.try_get_u64("disable_join_reorder")? != 0)
+    }
+
+    pub fn set_disable_join_reorder(&self, val: bool) -> Result<()> {
+        self.try_set_u64("disable_join_reorder", u64::from(val))
+    }
+
     pub fn get_runtime_filter(&self) -> Result<bool> {
         Ok(self.try_get_u64("enable_runtime_filter")? != 0)
     }
--- a/src/query/sql/src/planner/optimizer/cascades/cascade.rs
+++ b/src/query/sql/src/planner/optimizer/cascades/cascade.rs
@@ -51,9 +51,12 @@ impl CascadesOptimizer {
     pub fn create(
         ctx: Arc<dyn TableContext>,
         metadata: MetadataRef,
-        optimized: bool,
+        mut optimized: bool,
     ) -> Result<Self> {
         let explore_rule_set = if ctx.get_settings().get_enable_cbo()? {
+            if ctx.get_settings().get_disable_join_reorder()? {
+                optimized = true;
+            }
             get_explore_rule_set(optimized)
         } else {
             RuleSet::create()
--- a/src/query/sql/src/planner/optimizer/optimizer.rs
+++ b/src/query/sql/src/planner/optimizer/optimizer.rs
@@ -30,6 +30,7 @@ use crate::optimizer::hyper_dp::DPhpy;
 use crate::optimizer::runtime_filter::try_add_runtime_filter_nodes;
 use crate::optimizer::util::contains_local_table_scan;
 use crate::optimizer::HeuristicOptimizer;
+use crate::optimizer::RuleID;
 use crate::optimizer::SExpr;
 use crate::optimizer::DEFAULT_REWRITE_RULES;
 use crate::optimizer::RESIDUAL_RULES;
@@ -152,7 +153,7 @@ pub fn optimize_query(
     let mut result = heuristic.pre_optimize(s_expr)?;
     result = heuristic.optimize_expression(&result, &DEFAULT_REWRITE_RULES)?;
     let mut dphyp_optimized = false;
-    if ctx.get_settings().get_enable_dphyp()? {
+    if ctx.get_settings().get_enable_dphyp()? && !ctx.get_settings().get_disable_join_reorder()? {
         let (dp_res, optimized) =
             DPhpy::new(ctx.clone(), metadata.clone()).optimize(Arc::new(result.clone()))?;
         if optimized {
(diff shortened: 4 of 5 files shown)


BohuTANG:
What's the bug is? or we is it possible to add a logic test to explain the pipeline after fixed.

SkyFan2002:
FYI, some DML will actually do some data mutation( and commit) in `execute2` instead of just building pipeline, so if you want to explain pipeline of DML, you should carefully check the `execute2`.  

Or for safety, suggest returning an unimplemented error for now , and implementing them in later pr.

JackTan25:
> FYI, some DML will actually do some data mutation( and commit) in `execute2` instead of just building pipeline, so if you want to explain pipeline of DML, you should carefully check the `execute2`.
> 
> Or for safety, suggest returning an unimplemented error for now , and implementing them in later pr.

for this pr, I just give a flag for interpreter, and the old logic is the same

SkyFan2002:
> > FYI, some DML will actually do some data mutation( and commit) in `execute2` instead of just building pipeline, so if you want to explain pipeline of DML, you should carefully check the `execute2`.
> > Or for safety, suggest returning an unimplemented error for now , and implementing them in later pr.
> 
> for this pr, I just give a flag for interpreter, and the old logic is the same. cc @SkyFan2002

I mean, if you call explain pipeline delete, some mutation may be really committed. https://github.com/datafuselabs/databend/blob/0c5dd54b47a2ab94e213d9dd517d7f52d3598071/src/query/storages/fuse/src/operations/delete.rs#L105

JackTan25:
> > > FYI, some DML will actually do some data mutation( and commit) in `execute2` instead of just building pipeline, so if you want to explain pipeline of DML, you should carefully check the `execute2`.
> > > Or for safety, suggest returning an unimplemented error for now , and implementing them in later pr.
> > 
> > 
> > for this pr, I just give a flag for interpreter, and the old logic is the same. cc @SkyFan2002
> 
> I mean, if you call explain pipeline delete, some mutation may be really committed.
> 
> https://github.com/datafuselabs/databend/blob/0c5dd54b47a2ab94e213d9dd517d7f52d3598071/src/query/storages/fuse/src/operations/delete.rs#L105

yes, we need to track this, thanks.

src/query/service/src/interpreters/interpreter_delete.rs
```diff
@@ -78,12 +78,21 @@ use crate::stream::PullingExecutorStream;
 pub struct DeleteInterpreter {
     ctx: Arc<QueryContext>,
     plan: DeletePlan,
+    explain_pipeline: bool,
```
zhyass, line 81: use `enabled_table_lock`

src/query/service/src/interpreters/interpreter_delete.rs
```diff
@@ -106,12 +115,17 @@ impl Interpreter for DeleteInterpreter {
...
         let tbl = self.ctx.get_table(catalog_name, db_name, tbl_name).await?;
         let table_info = tbl.get_table_info().clone();
-
-        // Add table lock heartbeat.
-        let handler = TableLockHandlerWrapper::instance(self.ctx.clone());
-        let mut heartbeat = handler
-            .try_lock(self.ctx.clone(), table_info.clone())
-            .await?;
+        let explain_pipeline = self.explain_pipeline;
+        let mut heartbeat = None;
```
zhyass, line 119: if `enabled_table_lock` is false, create a `DummyTableLock`.

src/query/service/src/interpreters/interpreter_delete.rs
```diff
@@ -251,13 +265,18 @@ impl Interpreter for DeleteInterpreter {
...
         }
-
+        let explain_pipeline = self.explain_pipeline;
         if build_res.main_pipeline.is_empty() {
-            heartbeat.shutdown().await?;
+            if !explain_pipeline {
+                heartbeat.unwrap().shutdown().await?;
+            }
         } else {
             build_res.main_pipeline.set_on_finished(move |may_error| {
```
zhyass, line 274:
```rust
        if build_res.main_pipeline.is_empty() {
            heartbeat.shutdown().await?;
        } else {
            if enable_table_lock {
                build_res.main_pipeline.set_on_finished(move |may_error| {
                    // shutdown table lock heartbeat.
                    GlobalIORuntime::instance().block_on(async move { heartbeat.shutdown().await })?;
                    match may_error {
                        None => Ok(()),
                        Some(error_code) => Err(error_code.clone()),
                    }
                });
            }
        }
```

JackTan25:
I check update/replace/insert/recluster, it's safe. delete will do_truncate, so I add some logics. For compact, if compact block, it's safe, so hook_compact is also safe. But for compact_segment, it's not so, so I add some logic. cc @dantengsky @zhyass @SkyFan2002

SkyFan2002:
Another truncate:
https://github.com/datafuselabs/databend/blob/0c5dd54b47a2ab94e213d9dd517d7f52d3598071/src/query/storages/fuse/src/operations/delete.rs#L128

JackTan25:
> Another truncate:
> 
> https://github.com/datafuselabs/databend/blob/0c5dd54b47a2ab94e213d9dd517d7f52d3598071/src/query/storages/fuse/src/operations/delete.rs#L128

fixed. Thanks @SkyFan2002 , you are more careful.

SkyFan2002 (approved):

src/query/service/src/interpreters/interpreter_factory.rs
```diff
@@ -53,17 +53,25 @@ pub struct InterpreterFactory;
 /// Such as: Plan::Query -> InterpreterSelectV2
 impl InterpreterFactory {
     #[async_backtrace::framed]
-    pub async fn get(ctx: Arc<QueryContext>, plan: &Plan) -> Result<InterpreterPtr> {
+    pub async fn get(
+        ctx: Arc<QueryContext>,
+        plan: &Plan,
+        is_explain_pipeline: bool,
```
sundy-li, line 59: `is_explain_pipeline` could be derived from plan.

```
let is_explain_pipeline = match plan ...
```

JackTan25: that's better, thanks

JackTan25: we can't use this method, when we arrive here, it's the real plan like delete/update, not explain_pipeline plan.

dantengsky:
@JackTan25 

really appreciate it. 

although the way this PR addresses the issue should work, it looks a little bit hacky.

could we instead, make the `update / delete` etc. follow the principle that,  in method `Interpreter::execute2`, no "side effects" but only the pipeline should be constructed?

e.g. the table-locking related actions should only be taken in `Interpreter::execute` (we may need to extend the `Interpreter` trait or introduce other mechanisms, to express that during execution, something should be locked)

I think that is the original idea/principle that impls of `Interpreter::execute2` should follow.  @zhang2014 need your opinion, please.

------
besides

seems that during explaining  `optimize table ...`, the following action should be prohibited as well

https://github.com/datafuselabs/databend/blob/744cf3ae5e6cb5efebe38acf39fa9d69be1efd17/src/query/service/src/interpreters/interpreter_table_optimize.rs#L115-L121

dantengsky:
> could we instead, make the update / delete etc. follow the principle that, in method Interpreter::execute2, no "side effects" but only the pipeline should be constructed?

that would be harder, but if necessary, we can do the refacts step by step.

JackTan25:
stop merge this pr first. Maybe this pr will cause some errors unexpectedly, a little hacky.

PR #12603, opened 2023-08-29
fix(query): fix negative for u64

sundy-li:

## Summary

Summary about this PR

 fix negative for u64

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12603)


lichuang (approved):

src/query/functions/src/scalars/arithmetic.rs
```diff
@@ -498,6 +498,37 @@ fn register_unary_minus(registry: &mut FunctionRegistry) {
...
+                        },
+                        vectorize_with_builder_1_arg::<NumberType<u64>, NumberType<i64>>(
+                            |a, output, ctx| {
+                                let res = (a as i128).wrapping_neg();
+
+                                if res < std::i64::MIN as i128 || res > std::i64::MAX as i128 {
+                                    ctx.set_error(output.len(), "number overflowed");
+                                    output.push(0);
+                                } else {
+                                    output.push(res as i64);
```
andylokandy, line 526:
```suggestion
                                const MAX: u64 = -i64::MIN as u64;

                                if a > MAX {
                                    ctx.set_error(output.len(), "number overflowed");
                                    output.push(0);
                                } else {
                                    output.push(-(a as i64));
```

src/query/functions/src/scalars/arithmetic.rs
```diff
@@ -517,13 +517,11 @@ fn register_unary_minus(registry: &mut FunctionRegistry) {
                         },
                         vectorize_with_builder_1_arg::<NumberType<u64>, NumberType<i64>>(
                             |a, output, ctx| {
-                                let res = (a as i128).wrapping_neg();
-
-                                if res < std::i64::MIN as i128 || res > std::i64::MAX as i128 {
+                                if a > -i64::MIN as u64 {
```
andylokandy, line 520: same as the domain calculation?

src/query/functions/src/scalars/arithmetic.rs
andylokandy: oh, -i64::MIN will overflow...

sundy-li:
```
 if a > -i64::MIN as u64 {
    |                              ^^^^^^^^^ attempt to negate `i64::MIN`, which would overflow
```

src/query/functions/src/scalars/arithmetic.rs
```diff
@@ -498,6 +498,35 @@ fn register_unary_minus(registry: &mut FunctionRegistry) {
...
+                            }
+
+                            FunctionDomain::Domain(SimpleDomain::<i64> {
+                                min: min as i64,
+                                max: max as i64,
+                            })
+                        },
+                        vectorize_with_builder_1_arg::<NumberType<u64>, NumberType<i64>>(
+                            |a, output, ctx| {
+                                if a < std::i64::MIN as i128 || a > std::i64::MAX as i128 {
```
andylokandy, line 520:
```suggestion
                                if a < -(i64::MIN as i128) as u64 {
```
Eventually, llvm will fold it to a u64 constant.

PR #12608, opened 2023-08-29
feat: support hash join build spill

xudong963:

## Summary

In terms of the overall framework, each build processor contains a `Spiller`, and all processors share the `BuildSpillCoordinator`.

image

Build processor state transitions:

image

1. All processors have an initial state of **Running**.
2. check if spill is enabled, if not, normal process, if need to check memory usage, determine if spill is required
3. If no spill is needed, determine if another processor is already waiting for a spill.
4. if spill is needed or there is another processor waiting for spill, determine if it is the last processor, if so, the procesor is responsible for calculating the partition, splitting the spill task to equalize the tasks of each processor as much as possible, and then notify all the processors to start the first round of spill
5. If it is not the last processor, enter the **WaitProbe**.
6. If don't need to wait for spill, determine if it has already spilled, if yes, then directly spill the corresponding block to the corresponding partition file set. Otherwise, the process is normal.
7. After completing the first round of spill, enter Running state, generate hashtable from the block in memory and send it to the probe side, and enter **WaitProbe** state.
8. After probe side completes the spill, it picks a partition and notifies build to read the data of this partition and generate a hashtable to send to probe.
9. When all the partitions have been read, the build finishes.

**Note**: To avoid large PR, the PR only supports hash join build spill, will add probe spill in next PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12608)


BohuTANG:
askbend:summary

xudong963:
> askbend:summary

Summary added.

src/query/service/src/pipelines/processors/transforms/hash_join/build_spill/build_spill_state.rs
```diff
@@ -0,0 +1,363 @@
...
+    pub send_partition_set: AtomicBool,
+}
+
+impl BuildSpillState {
+    pub fn create(
+        _ctx: Arc<QueryContext>,
+        spill_coordinator: Arc<BuildSpillCoordinator>,
+        build_state: Arc<HashJoinBuildState>,
+    ) -> Self {
+        let spill_config = SpillerConfig::create("_hash_join_build_spill".to_string());
```
zhang2014, line 57: {tenant}/_hash_join_build_spill

src/query/service/src/pipelines/processors/transforms/hash_join/build_spill/build_spill_state.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    // Partition input blocks to different partitions.
+    // Output is <partition_id, blocks>
+    fn partition_input_blocks(
+        &self,
+        input_blocks: &[DataBlock],
+    ) -> Result<HashMap<u8, Vec<DataBlock>>> {
+        let mut partition_blocks = HashMap::new();
+        let mut partition_map = HashMap::with_capacity(8);
+        for block in input_blocks.iter() {
```
zhang2014, line 178: Maybe `DataBlock::scatter` is better.

src/query/service/src/pipelines/processors/transforms/hash_join/build_spill/build_spill_state.rs
```diff
@@ -0,0 +1,363 @@
...
+            let mut hashes = Vec::with_capacity(block.num_rows());
+            self.get_hashes(block, &mut hashes)?;
+            for (row_idx, hash) in hashes.iter().enumerate() {
+                let partition_id = *hash as u8 & 0b0000_0011;
```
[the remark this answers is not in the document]
xudong963: Now there are at most four partitions here, and I want to make it configurable afterward. the number of partitions is not the more the better, but too small will easily lead to the data read up afterward still not fitting into memory, so I want to eventually have the maximum number of partitions as configurable, the default is given to four, and then can support recursion spill afterward.

src/query/service/src/pipelines/processors/transforms/hash_join/build_spill/build_spill_state.rs
xudong963: done

src/query/service/src/pipelines/processors/transforms/hash_join/build_spill/build_spill_state.rs
xudong963: done

zhang2014:
performance degration in tpch cluster mode.

BohuTANG:
To check the benchmark, we can:
`repo.databend.rs` -> `d26e2oxucvi4r9.cloudfront.net`
cc @xudong963 @zhang2014

zhang2014 (approved):

PR #12614, opened 2023-08-29
refactor: unify create index check.

ariesdevil:

## Summary

Unify create index check.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12614)


RinChanNOWWW (approved):

sundy-li (approved):

src/query/sql/src/planner/binder/ddl/index.rs
```diff
@@ -65,11 +63,19 @@ impl Binder {
         } = stmt;
 
         // check if query support index
-        let mut agg_index_checker = AggregatingIndexChecker::default();
-        walk_query(&mut agg_index_checker, query);
-        // todo(ariesdevil): do all check using `AggregatingIndexChecker`
-        Self::check_index_support(query, agg_index_checker.has_no_deterministic_func)?;
-
+        {
+            let mut agg_index_checker = AggregatingIndexChecker::default();
```
BohuTANG, line 67: This is a bad practice:
```
let mut agg_index_checker = AggregatingIndexChecker::default();
```
agg_index_checker is mut, we can do:
```
agg_index_checker.not_support = true
```

Better solution:
```
pub struct AggregatingIndexChecker {
    not_support: RwLock<bool>,
}

impl AggregatingIndexChecker {
  pub fn is_support(&self) -> bool {
     self.not_support.read()
  }
  
  ....
}
```

ariesdevil: Ok, but I think the `RwLock` is not needed, the visitor has no concurrent call.

Issue #12632, opened 2023-08-30
sqllogic test for copy


youngsofun:
**Summary**

- [x] share all tests for s3 and fs  https://github.com/datafuselabs/databend/pull/12631
   - [ ] http too? (start a http server?)
- [ ] for cluster and standalone 
- [ ] work around sandbox:  gen SQL from the prepare_script, and run right after `set sandbox_tenant=xxx`.
- [x] share tests for parquet2 and parquet_rs   https://github.com/datafuselabs/databend/pull/12655
- [ ] test "copy into stage"
     - [ ] output of "copy into stage"
     - [ ] able to test copy from unloaded data
         - [ ] option for "copy into table"  to  only show a summary 
         - [ ] optional stable output path (without uuid)
- [ ] migrate existent tests.
- [ ] more tests
    - [ ] all types
    - [ ] all options
    - [ ] error cases
         - [ ]  polish error message.
            - [ ] diff error code?
            - [ ] escape
            - [ ] no file, line in copy output

PR #12655, opened 2023-09-01
ci: logictest for copy parquet.

youngsofun:

## Summary

part of https://github.com/datafuselabs/databend/issues/12632

add suit and ci first with a few cases, so new test cases can be add with logict tests, cc @RinChanNOWWW . 

- run all  cases for both paruqet2 and parquet_rs

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12655)


RinChanNOWWW:
Do we need to clean the stage after test?

youngsofun:
> Do we need to clean the stage after test?

not the `@data` stage is an exact mapping to the `tests/data`, which is read only, we may use other stage if write is needed later.

BohuTANG (approved):

PR #12656, opened 2023-09-01
chore(http): both accept stage_name and stage-name as args on upload_to_stage

flaneur2020:

## Summary

having underscore in the http header seems to be not HTTP standard, and easy to get troubles on various HTTP proxies like nginx.

this PR makes `upload_to_stage` also accepts `stage-name` and `x-databend-stage-name` as arguments, and adjusted the related docs to encourage using `stage-name` instead of `stage_name`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12656)

### DIFF
--- a/docs/doc/12-load-data/00-transform/06-querying-metadata.md
+++ b/docs/doc/12-load-data/00-transform/06-querying-metadata.md
@@ -37,7 +37,7 @@ CREATE STAGE my_internal_stage;
 2. Use cURL to make a request to the File Upload API:
 
 ```shell title='Put books.parquet to stage'
-curl -u root: -H "stage-name:my_internal_stage" -F "upload=@books.parquet" -XPUT "http://localhost:8000/v1/upload_to_stage"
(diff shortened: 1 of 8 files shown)


BohuTANG (approved):

BohuTANG:
Need a test to check the new header works and not braked by other PR in the future.

flaneur2020:
> Need a test to check the new header works and not braked by other PR in the future.

let me add one in this pr before getting merge

flaneur2020:
added tests in https://github.com/datafuselabs/databend/pull/12656/files#diff-0b123a2c66dc6caef99f4d85f948a085079fd897efda785cd1505f72858068b4R11

sundy-li:
The headers in databend should always be `x-databend-xxxx` (or upper case).

Let's just add `x-databend-stage-name` for this pr.

Later we will bump the clients just send `x-databend-stage-name`.

flaneur2020:
headers are case insensitive, so we can have `x-databend-stage-name` for suggested and keep `stage_name` for backward compatibility for a while

flaneur2020:
@sundy-li PTAL

PR #12664, opened 2023-09-02
chore: fix modify column type not null bug

lichuang:

## Summary

chore: fix modify column type not null bug

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12664)


tests/suites/0_stateless/17_altertable/17_0005_alter_table_modify_column_type.result
```diff
@@ -9,7 +9,13 @@ c	INT	NO	0
 ERROR 1105 (HY000) at line 1: Internal. Code: 1001, Text = invalid float literal while evaluating function `to_float32('a')`.
 ERROR 1105 (HY000) at line 1: UnknownColumn. Code: 1058, Text = Cannot find column b.
 0	1
-ERROR 1105 (HY000) at line 1: Internal. Code: 1001, Text = invalid float literal while evaluating function `to_float32('a')`.
+ERROR 1105 (HY000) at line 1: Internal. Code: 1001, Text = error: 
+  --> SQL:1:1
+  |
+1 | alter table test_modify_column_type.c modify column a float default 'a'
+  | ^^^^^^^^^^^^^^^ invalid float literal while evaluating function `to_float32('a')`
```
ariesdevil, line 16: the error hint is not related to the sql statement.

src/query/sql/src/planner/binder/ddl/table.rs
```diff
@@ -702,7 +704,6 @@ impl Binder {
         })))
     }
 
-    #[async_backtrace::framed]
```
ariesdevil, line 705: why delete this attribute?

tests/suites/0_stateless/17_altertable/17_0005_alter_table_modify_column_type.result
lichuang:
```
CREATE table b(a int default 'a');

1 | CREATE table b(a int default 'a')
  | ^^^^^^^^^^^^^ invalid digit found in string while evaluating function `to_int32('a')`
```

the same behaivor when creating table.

src/query/ast/src/parser/statement.rs
```diff
@@ -2090,14 +2075,14 @@ pub fn modify_column_action(i: Input) -> IResult<ModifyColumnAction> {
 
     let modify_column_type = map(
         rule! {
-            #column_type ~ ("," ~ COLUMN ~ #column_type)*
+            #column_def ~ ("," ~ COLUMN ~ #column_def)*
```
b41sh, line 2078: `column_def` may have computed expr, which is not allowed to modify. Need extra check and report a error.

src/query/service/src/interpreters/interpreter_table_modify_column.rs
```diff
@@ -190,22 +190,20 @@ impl ModifyTableColumnInterpreter {
...
-        for field in modify_table_schema.fields() {
-            let column = &field.name;
-            let default_expr_opt = &field.default_expr;
+        for (field, _comment) in field_and_comments {
+            let column = &field.name.to_string();
             let data_type = &field.data_type;
             if let Ok(i) = schema.index_of(column) {
-                if let Some(default_expr) = default_expr_opt {
-                    new_schema.fields[i].data_type = data_type.clone();
+                if let Some(default_expr) = &field.default_expr {
```
b41sh, line 204: As we have analyzed the computed column, we should also set `computed_expr`.
Some tests for this case also need to be added.

ariesdevil (approved):

b41sh (approved):

PR #12676, opened 2023-09-04
refactor(observability): use prometheus_client in storage mod

### LINKED ISSUE #12649: observability: replace metrics-rs with prometheus-client in databend-query's storage

flaneur2020:

## Summary

this pr adds a dual write metrics to prometheus-client in the storage mod and did not remove the existed metrics-rs code yet, the exported metrics are still from `metrics-rs` after this PR merged. we're plan to switch to `prometheus-client` completely after all the metrics instrument have been covered with the prometheus-client code.

there're two changes on merics after we switched to prometheus-client:

1. change the metric `cache_$cache_name_$action` into `cache_$action{cache_name="xxx"}`, the prometheus labels allows users can auto discover the newly introduced cache types, which makes it easier to make a grafana dashboard to show the hit rate on different cache types.
2. prometheus-client seems prefer to add a "_total" suffix to all the counter metrics, this may break the old dashboard configs. However, I guess there're reasons to make this convention in the community (this convention seems come from the open-metrics project), we can follow it.

fixes #12649 

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12676)


dantengsky (approved):
LGTM

BohuTANG:
Conflicting files
src/query/storages/fuse/src/operations/merge_into/mutator/matched_mutator.rs

PR #12683, opened 2023-09-05
feat: support domain calculation for decimal types.

RinChanNOWWW:

### Main Target

Support domain calculations for following functions:

- [x] to_decimal
- [x] to_float.
- [x] comparison functions
- [x] arithmetic functions.
- [x] pass parquet pruning unit tests involving decimal types.
- [x] fix bugs of some decimal functions when the argument type is nullable.

### Other Modification

- Fix decimal functions registry to support domain calculation. Domain calculation will be erased if the type is nullable in the history implementation.
- Refactor codes. Reduce unnecessary type match.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12683)


BohuTANG:
askbend:summary

src/query/functions/src/scalars/decimal.rs
```diff
@@ -460,36 +618,60 @@ pub fn register(registry: &mut FunctionRegistry) {
     });
 }
 
+macro_rules! single_decimal128_to_float {
```
andylokandy, line 621: can we use https://github.com/andylokandy/databend/blob/811ff87f748ae5ba325afa6c930605cc40fa1ab8/src/query/expression/src/types/decimal.rs#L200?

src/query/functions/src/scalars/decimal.rs
```diff
@@ -927,13 +1219,102 @@ fn decimal_to_decimal(
...
+            let x = $from * <$dest_type_name>::one();
+            match x.checked_mul(factor) {
+                Some(x) if x <= max && x >= min => Some(x),
+                _ => None,
+            }
+        }
+    }};
+}
+
+fn single_decimal_256_to_128(
```
andylokandy, line 1245: what's the difference between:
https://github.com/andylokandy/databend/blob/811ff87f748ae5ba325afa6c930605cc40fa1ab8/src/query/functions/src/scalars/decimal.rs#L884

RinChanNOWWW: `single_decimal_256_to_128` only operates on one decimal.

src/query/functions/src/scalars/decimal.rs
RinChanNOWWW: Yes

src/query/functions/src/scalars/decimal.rs
andylokandy: so, can the new one be a wrapper of the old one?

RinChanNOWWW: I‘m afraid not. The new method is a scalar method, and the old one is for vector. If we reuse the codes, we will introduce instructions and branches into the vectorized loop.

sundy-li (approved):

PR #12686, opened 2023-09-05
fix: remove offset after limit push down outer join

xudong963:

## Summary

For the following query, t1 has 10000 rows, t2 has 100 rows:
```
select * from t1 left join t2 on t1.a = t2.b limit 50 offset 10;
```
Before the ticket, its plan, the result only has 40 rows, correctly, it should return 50 rows.

```
│ Limit                                                                                                                      │
│ ├── output columns: [t2.b (#1), t1.a (#0)]                                                                                 │
│ ├── limit: 50                                                                                                              │
│ ├── offset: 10                                                                                                             │
│ ├── estimated rows: 50.00                                                                                                  │
│ └── HashJoin                                                                                                               │
│     ├── output columns: [t2.b (#1), t1.a (#0)]                                                                             │
│     ├── join type: RIGHT OUTER                                                                                             │
│     ├── build keys: [CAST(t1.a (#0) AS UInt64 NULL)]                                                                       │
│     ├── probe keys: [t2.b (#1)]                                                                                            │
│     ├── filters: []                                                                                                        │
│     ├── estimated rows: 5000.00                                                                                            │
│     ├── Limit(Build)                                                                                                       │
│     │   ├── output columns: [t1.a (#0)]                                                                                    │
│     │   ├── limit: 50                                                                                                      │
│     │   ├── offset: 10                                                                                                     │
│     │   ├── estimated rows: 50.00                                                                                          │
│     │   └── TableScan                                                                                                      │
│     │       ├── table: default.default.t1                                                                                  │
│     │       ├── output columns: [a (#0)]                                                                                   │
│     │       ├── read rows: 10000                                                                                           │
│     │       ├── read bytes: 19063                                                                                          │
│     │       ├── partitions total: 1                                                                                        │
│     │       ├── partitions scanned: 1                                                                                      │
│     │       ├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 1 to 1, bloom pruning: 0 to 0>] │
│     │       ├── push downs: [filters: [], limit: 60]                                                                       │
│     │       └── estimated rows: 10000.00                                                                                   │
│     └── TableScan(Probe)                                                                                                   │
│         ├── table: default.default.t2                                                                                      │
│         ├── output columns: [b (#1)]                                                                                       │
│         ├── read rows: 100                                                                                                 │
│         ├── read bytes: 172                                                                                                │
│         ├── partitions total: 1                                                                                            │
│         ├── partitions scanned: 1                                                                                          │
│         ├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 1 to 1, bloom pruning: 0 to 0>]     │
│         ├── push downs: [filters: [], limit: NONE]                                                                         │
│         └── estimated rows: 100.00                                                                                         │
```
That's because after the limit is pushed down scan, the limit is still kept, so there are at most 50 rows for join, and the top
limit will make the final only 40 rows.

Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12686)


BohuTANG:
From the test, the fixed logic appears to be incorrect.

PR #12713, opened 2023-09-06
feat: Ownership Model prototype

ZhiHanZ:

## Summary

1. each object will bring its creator as the owner, the owner has full manipulation privilege on the object.
2. current support database owner and table owner, further object is accompany with GrantObject extension.
3. support to check database and table owner through system table.
4. remove granted ownership as it introduce too much complexity on current built-in role and managed role.
5. support SQL semantic like grant ownership.

Limitation:
Operation such as Renaming is limited on ownership model as current grant object cannot understand table and database id, this bug should be fixed in further PR.
Show grants could not see all of the objects owned by given role as we removed the transaction for granted object which introduced a lot of tricky corner cases.

- Closes #issue


src/query/users/src/role_cache_mgr.rs
```diff
@@ -111,6 +118,63 @@ impl RoleCacheManager {
         Ok(cached_roles.roles.get(role).cloned())
     }
 
+    #[async_backtrace::framed]
+    pub async fn find_object_owner(
```
flaneur2020, line 122: this method seems not quite related with `role_cache`, any other suggested place to put this method?

src/query/management/src/role/role_mgr.rs
```diff
@@ -213,13 +195,17 @@ impl RoleMgr {
                 })
                 .await?;
             let mut table_meta = table_info.meta.clone();
+
             // if current owner is not none and not from, return error
             if table_meta.owner.is_some()
                 && table_meta.owner.as_ref().unwrap().owner_role_name != from.clone()
```
flaneur2020, line 201: this privilege check seems missed the checking about the other available roles.

I guess we can extract this privilege check outside here, and put the privilege check logic into the caller side, thus it might easier for us to keep the privilege check logic consistent.

ZhiHanZ: caller would loop over all available roles until find one owner.

src/query/users/src/role_cache_mgr.rs
ZhiHanZ: UserApiProvider seems to be some api interface provider which may also not make sense to provide this meta service logic on it.

ZhiHanZ: I think the root issue is that we do not want to directly put an api client in Session struct, as it may get leaked to introduce unwanted trouble

tests/suites/0_stateless/20+_others/20_0014_privilege_ownership.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "GRANT OWNERSHIP ON d20_0014.* TO ROLE 'd20_0014_owner'" | $MYSQL_CLIENT_CONNECT
+
+echo "GRANT ROLE 'd20_0014_owner' TO '${TEST_USER_NAME}'" | $MYSQL_CLIENT_CONNECT
+echo "ALTER USER '${TEST_USER_NAME}' WITH DEFAULT_ROLE='20_0014_owner'" | $MYSQL_CLIENT_CONNECT
+
+## owner should have all privileges on the table
+echo "create table d20_0014.table1(i int);" | $TEST_USER_CONNECT
+echo "insert into d20_0014.table1 values(1),(2),(3);" | $TEST_USER_CONNECT
+echo "select * from d20_0014.table1;" | $TEST_USER_CONNECT
+
```
flaneur2020, line 35: add a test about creating a database, and try listing the newly created database's owner?

src/query/users/src/role_cache_mgr.rs
flaneur2020: I see, I guess a seperated `PrivilegeManager` may be a better code layout than floating all the things into `Session`. Can you leave a TODO on top of this function? I'd like to try adding a PrivilegeManager later after this PR get merged, to collect the fragmented privilege logics into one place.

src/query/management/src/role/role_mgr.rs
flaneur2020: if the caller already have the privilege check, we can remove the check here?

this check here will forbid all the requests whom do not have the exact same roles.

flaneur2020 (approved):
LGTM, @TCeason PTAL

TCeason (approved):
LGTM

PR #12734, opened 2023-09-07
chore: add a setting to disable join reorder

xudong963:

## Summary

Add a setting that can disable join reorder.

There are two scenes that need to disable join reorder: testing and debug.

### DIFF
--- a/src/query/sql/src/planner/optimizer/cascades/cascade.rs
+++ b/src/query/sql/src/planner/optimizer/cascades/cascade.rs
@@ -51,9 +51,12 @@ impl CascadesOptimizer {
     pub fn create(
         ctx: Arc<dyn TableContext>,
         metadata: MetadataRef,
-        optimized: bool,
+        mut optimized: bool,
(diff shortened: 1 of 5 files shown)


src/query/sql/src/planner/optimizer/cascades/cascade.rs
```diff
@@ -51,9 +51,12 @@ impl CascadesOptimizer {
     pub fn create(
         ctx: Arc<dyn TableContext>,
         metadata: MetadataRef,
-        optimized: bool,
+        mut optimized: bool,
     ) -> Result<Self> {
         let explore_rule_set = if ctx.get_settings().get_enable_cbo()? {
+            if ctx.get_settings().get_disable_join_reorder()? {
+                optimized = true;
```
Dousir9, line 58: When `optimized` is true, we will still use `rule_set_rs_l1` to cause join reorder.

Dousir9 (approved):

PR #12743, opened 2023-09-07
chore: add query plan to log

Dousir9:

## Summary

Example:
```
2023-09-07T05:38:31.259308Z  WARN databend_query::interpreters::interpreter_select: src/query/service/src/interpreters/interpreter_select.rs:240 Query plan:
HashJoin
├── output columns: [t1.a (#0)]
├── join type: INNER
├── build keys: [t2.b (#3)]
├── probe keys: [t1.b (#1)]
├── filters: []
├── estimated rows: 0.00
├── TableScan(Build)
│   ├── table: default.default.t2
│   ├── output columns: [b (#3)]
│   ├── read rows: 0
│   ├── read bytes: 0
│   ├── partitions total: 0
│   ├── partitions scanned: 0
│   ├── push downs: [filters: [], limit: NONE]
│   └── estimated rows: 0.00
└── TableScan(Probe)
    ├── table: default.default.t1
    ├── output columns: [a (#0), b (#1)]
    ├── read rows: 0
    ├── read bytes: 0
    ├── partitions total: 0
    ├── partitions scanned: 0
    └── estimated rows: 0.00
```

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12743)


xudong963 (approved):

BohuTANG:
There is a panic here:
https://github.com/datafuselabs/databend/actions/runs/6105862898/job/16570175563?pr=12743#step:4:1052

src/query/service/src/interpreters/interpreter_select.rs
```diff
@@ -231,6 +233,10 @@ impl Interpreter for SelectInterpreter {
 
         // 0. Need to build physical plan first to get the partitions.
         let physical_plan = self.build_physical_plan().await?;
+        let query_plan = physical_plan
+            .format(self.metadata.clone(), SharedProcessorProfiles::default())?
+            .format_pretty()?;
+        warn!("Query plan: \n{}", query_plan);
```
flaneur2020, line 239: how about make it a info level ?

flaneur2020 (approved):

src/query/service/src/interpreters/interpreter_select.rs
flaneur2020: also, we can consider attach a query id in this log, thus may make it easier for us to diagnose the related query plan for a failed query

Dousir9: > how about make it a info level ?

That will cause a lot of logs in the unit test.

flaneur2020: `warn` is considered to be a higher level than `info`, wouldn't `warn` introduce logs into the unit test?

Dousir9: > also, we can consider attach a query id in this log, thus may make it easier for us to diagnose the related query plan for a failed query

Done.

Dousir9: > `warn` is considered to be a higher level than `info`, wouldn't `warn` introduce logs into the unit test?

You're right, let me fix it.

PR #12747, opened 2023-09-07
feat: fill default columns for not match

JackTan25:

## Summary

Summary about this PR
this pr is used to stage blocks in not match to avoid concat blocks.
1. support fill default columns
2. stage blocks
- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12747)


src/query/service/src/pipelines/processors/transforms/transform_resort_addon_without_source_schema.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        let func_ctx = self.ctx.get_function_context()?;
+        Ok(CompoundBlockOperator {
+            ctx: func_ctx,
+            operators: vec![BlockOperator::Map {
+                exprs,
+                projections: None,
+            }],
+        })
```
dantengsky, line 90: seems there are duplications with `TransformResortAddOn`?  can we do a refactoring by the way?

JackTan25: No, transformAddOn need specify the source schema, but for merge into, the insert can be various, not consistent

dantengsky (approved):

PR #12748, opened 2023-09-07
fix(query): fix Float and Decimal can't be cast to boolean

### LINKED ISSUE #12696: bug: Float and Decimal can't be cast to boolean
### Search before asking

### Version

v1.2.98-nightly

### What's Wrong?

Float and Decimal can't be cast to boolean, using float and decimal as filter conditions will report an error.

### How to Reproduce?

```sql
mysql> create table test(a int, b float64, c decimal(76,2));

mysql> insert into test values (1, 1.1, 1.11),(0, 0.0, 0.0);
Query OK, 2 rows affected (0.09 sec)

mysql> select * from test;
+------+------+------+
| a    | b    | c    |
+------+------+------+
|    1 |  1.1 | 1.11 |
|    0 |  0.0 | 0.00 |
+------+------+------+
2 rows in set (0.09 sec)
Read 2 rows, 88.00 B in 0.023 sec., 86.81 rows/sec., 3.73 KiB/sec.

mysql> select * from test where a;
+------+------+------+
| a    | b    | c    |
+------+------+------+
|    1 |  1.1 | 1.11 |
+------+------+------+
1 row in set (0.08 sec)
Read 2 rows, 88.00 B in 0.024 sec., 81.94 rows/sec., 3.52 KiB/sec.

mysql> select * from test where b;
ERROR 1105 (HY000): Internal. Code: 1001, Text = unable to cast type `Float64` to type `Boolean NULL`.
mysql> select * from test where c;
ERROR 1105 (HY000): Internal. Code: 1001, Text = unable to cast type `Decimal(76, 2)` to type `Boolean NULL`.
```

b41sh:

## Summary

Support cast float and decimal to boolean, fix float and decimal cant' used as a filter condition.

- Closes #12696

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12748)


sundy-li (approved):

src/query/functions/src/scalars/boolean.rs
```diff
@@ -338,7 +341,109 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+
+    for src_type in ALL_FLOAT_TYPES {
+        with_float_mapped_type!(|NUM_TYPE| match src_type {
+            NumberDataType::NUM_TYPE => {
+                registry.register_1_arg::<NumberType<NUM_TYPE>, BooleanType, _, _>(
+                    "to_boolean",
+                    |_, domain| {
+                        FunctionDomain::Domain(BooleanDomain {
+                            has_false: domain.min <= OrderedFloat(0.0)
+                                && domain.max >= OrderedFloat(0.0),
```
andylokandy, line 355: Is there any difference from:
```suggestion
                            has_false: domain.min == OrderedFloat(0.0)
                                && domain.max == OrderedFloat(0.0),
```

src/query/functions/src/scalars/boolean.rs
```diff
@@ -338,7 +341,109 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                        vectorize_with_builder_1_arg::<
+                            NumberType<NUM_TYPE>,
+                            NullableType<BooleanType>,
+                        >(|val, output, _| {
+                            output.builder.push(val != OrderedFloat(0.0));
+                            output.validity.push(true);
+                        }),
+                    );
+
+                let name = format!("to_{src_type}").to_lowercase();
```
andylokandy, line 386: seems weird. you can rename `src_type` to `float_ty`

andylokandy (approved):

src/query/functions/src/scalars/boolean.rs
b41sh: we have float32 and float64, can't be sure which one is the source type.

PR #12754, opened 2023-09-08
feat: add snapshot timestamp in file name

lichuang:

## Summary

1. feat: add snapshot timestamp in [snapshot|segment|block|index] file name
2. move snapshot v5 to v6 since it was not used.

- Closes https://github.com/datafuselabs/databend/issues/12671

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12754)


JackTan25:
askbend:summary

sundy-li:
Why not change the style of `snapshot_id` to include the timestamp ?

src/query/storages/common/table-meta/src/meta/v5/snapshot.rs
```diff
@@ -103,6 +102,16 @@ impl TableSnapshot {
         // trim timestamp to micro seconds
         let trimmed_timestamp = trim_timestamp_to_micro_second(adjusted_timestamp);
         let timestamp = Some(trimmed_timestamp);
+        let prev_snapshot_id = match prev_id {
+            Some((prev_snapshot_id, prev_format)) => {
+                if prev_format <= 4 {
+                    Some((None, prev_snapshot_id, prev_format))
```
JackTan25, line 108: If we read an old snapshot, we think the timestamp is very old, and we will vacuum it? How to process the default timestamp, or read it from snapshot,right?

zhyass:
> Why not change the style of `snapshot_id` to include the timestamp ?

I agree with @sundy-li ’s suggestion. Perhaps we can consider using ulid to save space.

sundy-li:
https://github.com/oklog/ulid#string-representation

PR #12760, opened 2023-09-08
fix(query): the issue of (lpad,rpad) function getting hang

### LINKED ISSUE #12706: bug: `lpad` and `rpad` function does not respond if the `pad` string is empty
### Search before asking

### Version

v1.2.99-nightly

### What's Wrong?

`lpad` and `rpad` function does not respond if the `pad` string is empty

### How to Reproduce?

```sql
mysql> select lpad('', 1000, '');

mysql> select rpad('', 1000, '');
```

nange:

## Summary

1. fix the issue of (lpad,rpad) function getting hang if the pad string is empty

- Closes #12706

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12760)


src/query/functions/src/scalars/string.rs
```diff
@@ -139,7 +139,9 @@ pub fn register(registry: &mut FunctionRegistry) {
             |s, pad_len, pad, output, _| {
                 let pad_len = pad_len as usize;
                 if pad_len <= s.len() {
-                    output.put_slice(&s[..pad_len])
+                    output.put_slice(&s[..pad_len]);
```
b41sh, line 142: If the `pad_len` is greater than the length of `s` and the `pad` is empty, we should return an error message because it is not possible to fill the string to the `pad_len` with an empty string.

nange: Logically speacking, you are right. It's ok for me. Currently, I have adopted the same behavior as Postgres. I'll fix it.

nange: @b41sh PTAL.

b41sh (approved):
LGTM

PR #12809, opened 2023-09-12
feat: support column stats provider for PaquetTable (parquet_rs)

RinChanNOWWW:

## Summary

- Create column stats provider for parquet table.
- When creating the provider, we will read all the parquet metas into memory. To avoid duplicate IO and computing, we can store the read metas in memory directly. We only use metas before actually executing (`read_partitions`), so we don't need serialize them for cluster mode.
- Some SQL commands don't need to create column stats provider (such as `COPY INTO`), and meta reading will block the entire SQL parsing if we do it when binding table. Therefore, we need to do meta reading lazily. In another word, we only read parquet metas when `column_statistics_provider` is called.  If `column_statistics_provider` is not called, we will read metas in `read_partitions`.
- If ther parquet table is used for a `COPY` command, it will not create a column stats provider.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12809)


src/query/catalog/src/table.rs
```diff
@@ -483,3 +483,30 @@ impl ColumnStatisticsProvider for Parquet2TableColumnStatisticsProvider {
...
+            num_rows,
+        }
+    }
+    pub fn num_rows(&self) -> u64 {
+        self.num_rows
+    }
+}
+
+impl ColumnStatisticsProvider for ParquetTableColumnStatisticsProvider {
+    fn column_statistics(&self, column_id: ColumnId) -> Option<BasicColumnStatistics> {
```
sundy-li, line 509: better change the signature to return ` Option<&BasicColumnStatistics>`

youngsofun (approved):

Dousir9 (approved):

Issue #12791, opened 2023-09-11
bug: cross join select tuple values panic


b41sh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

v1.2.107-nightly

### What's Wrong?

cross join select tuple values panic
```
panicked at 'assertion failed: `(left == right)`
  left: `Tuple([Number(UInt32), Boolean])`,
 right: `Tuple([Number(UInt32), Nullable(Boolean)])`', src/query/expression/src/block.rs:64:21
   0: common_tracing::panic_hook::log_panic
             at ./src/common/tracing/src/panic_hook.rs:35:21
   1: common_tracing::panic_hook::set_panic_hook::{{closure}}
             at ./src/common/tracing/src/panic_hook.rs:30:9
   2: call<(&core::panic::panic_info::PanicInfo), (dyn core::ops::function::Fn<(&core::panic::panic_info::PanicInfo), Output=()> + core::marker::Send + core::marker::Sync), alloc::alloc::Global>
             at /rustc/8b35c0bb0f833c0077dc57006eb317edde2a2d1e/library/alloc/src/boxed.rs:1999:9
   3: rust_panic_with_hook
             at /rustc/8b35c0bb0f833c0077dc57006eb317edde2a2d1e/library/std/src/panicking.rs:709:13
   4: {closure#0}
             at /rustc/8b35c0bb0f833c0077dc57006eb317edde2a2d1e/library/std/src/panicking.rs:597:13
   5: __rust_end_short_backtrace<std::panicking::begin_panic_handler::{closure_env#0}, !>
             at /rustc/8b35c0bb0f833c0077dc57006eb317edde2a2d1e/library/std/src/sys_common/backtrace.rs:151:18
   6: begin_panic_handler
             at /rustc/8b35c0bb0f833c0077dc57006eb317edde2a2d1e/library/std/src/panicking.rs:593:5
   7: panic_fmt
             at /rustc/8b35c0bb0f833c0077dc57006eb317edde2a2d1e/library/core/src/panicking.rs:67:14
   8: assert_failed_inner
   9: core::panicking::assert_failed
             at /rustc/8b35c0bb0f833c0077dc57006eb317edde2a2d1e/library/core/src/panicking.rs:228:5
  10: common_expression::block::BlockEntry::new
             at ./src/query/expression/src/block.rs:64:21
  11: common_sql::evaluator::block_operator::BlockOperator::execute
[...]

PR #12814, opened 2023-09-12
fix: consider tuple type alone when checking type

xudong963:

## Summary

Summary about this PR

- Closes https://github.com/datafuselabs/databend/issues/12791

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12814)


src/query/expression/src/block.rs
```diff
@@ -590,3 +582,26 @@ impl Clone for Box<dyn BlockMetaInfo> {
...
 }
+
+fn check_type(data_type: &DataType, value: &Value<AnyType>) {
+    match value {
+        Value::Scalar(Scalar::Null) => {
+            assert!(data_type.is_nullable_or_null());
+        }
+        Value::Scalar(s) => {
+            match s {
+                Scalar::Tuple(ss) => {
```
b41sh, line 593: other nested types, such as `Array` and `Map` may have the same problem

xudong963: `Array` and `Map` nest `Column`, maybe don't have the problem.

b41sh: ok, `Array` and `Map` are ok
```sql
mysql> SELECT [true, tt1.b] FROM tt1 CROSS JOIN tt2;
+---------------+
| [true, tt1.b] |
+---------------+
| [1,1]         |
| [1,1]         |
| [1,1]         |
| [1,NULL]      |
| [1,NULL]      |
| [1,NULL]      |
| [1,0]         |
| [1,0]         |
| [1,0]         |
+---------------+
9 rows in set (0.04 sec)
```

b41sh (approved):

sundy-li (approved):

src/query/expression/src/block.rs
```diff
@@ -590,3 +582,27 @@ impl Clone for Box<dyn BlockMetaInfo> {
...
+                    assert!(matches!(data_type, DataType::Tuple(_)));
+                    if let DataType::Tuple(dts) = data_type {
+                        for (s, dt) in ss.iter().zip(dts.iter()) {
+                            check_type(dt, &Value::Scalar(s.clone()));
+                        }
+                    }
+                }
+                _ => assert_eq!(s.as_ref().infer_data_type(), data_type.remove_nullable()),
+            }
+        }
```
andylokandy, line 605:
```suggestion
        Value::Scalar(Scalar::Tuple(fields)) => {
            // Check if data_type is Tuple type.
            let data_type = data_type.remove_nullable();
            assert!(matches!(data_type, DataType::Tuple(_)));
            if let DataType::Tuple(dts) = data_type {
                for (s, dt) in fields.iter().zip(dts.iter()) {
                    check_type(dt, &Value::Scalar(s.clone()));
                }
            }
        }
        Value::Scalar(s) => assert_eq!(s.as_ref().infer_data_type(), data_type.remove_nullable()),
```

PR #12815, opened 2023-09-12
fix: fix syntax of Querying Staged Files.

youngsofun:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12815)


wubx (commented):
```sql
CREATE STAGE my_stage FILE_FORMAT = (TYPE = CSV);
```
In cases where you have staged a file in a format different from the specified stage format, you can explicitly specify the file format within the SELECT or COPY INTO statement. Here are examples:
```sql
SELECT * FROM @my_stage (FILE_FORMAT=>'NDJSON');
COPY INTO my_table FROM (SELECT * SELECT @my_stage t) FILE_FORMAT = (TYPE = NDJSON);
````

file_format=> ‘’ need use user custom file_format, there is not have NDJSON.

Issue #12576, opened 2023-08-24
Feature: Using sqlsmith to support sql fuzzy testing


b41sh:
**Summary**

Description for this feature.

PR #12825, opened 2023-09-13
feat(query): Support inserting bitmap string values

b41sh:

## Summary

- Support inserting bitmap string values
- Sqlsmith support insert bitmap values

Part of #12576

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12825)


src/query/formats/src/field_encoder/row_based.rs
```diff
@@ -165,6 +167,20 @@ pub trait FieldEncoderRowBased {
...
+    fn write_bitmap(
+        &self,
+        column: &StringColumn,
+        row_index: usize,
+        out_buf: &mut Vec<u8>,
+        raw: bool,
+    ) {
+        let v = unsafe { column.index_unchecked(row_index) };
+        let rb = RoaringTreemap::deserialize_from(v).unwrap();
+        let vals = rb.into_iter().collect::<Vec<_>>();
```
sundy-li, line 179: this will write a very large string.

sundy-li: if rb only contains one element, the output will still be very large.
```
rb.insert(10000000000);
````



I think binary is better.

b41sh: Sqlsmith needs to generate bitmap test values. How about adding a config item in `CommonSettings` to control whether to generate String or Binary values for Bitmap?

sundy-li (approved):

ariesdevil (approved):

PR #12833, opened 2023-09-13
fix: delete contain internal columns panic

zhyass:

## Summary

```
mysql> delete from t where _row_id=1;
ERROR 1105 (HY000): SemanticError. Code: 1065, Text = Internal column `_row_id` is not allowed in current statement.
```

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12833)


b41sh (approved):

b41sh:
Do we support `_row_id` in a subquery in the where condition, if not, it seems like the relevant code could be removed.

PR #12848, opened 2023-09-14
fix: fix ambiguity in map_element and map_access

### LINKED ISSUE #12695: bug: parse map expr failed
### Search before asking

### Version

v1.2.98-nightly

### What's Wrong?

parse map expr failed if the value is a column or function

### How to Reproduce?

```sql
mysql> create table test (a int, b string);

mysql> select {a:b} from t;
ERROR 1105 (HY000): SyntaxException. Code: 1005, Text = error:
  --> SQL:1:12
  |
1 | select {a:b} from t
  | ------ -   ^ expected `}`, `IS`, `NOT`, `IN`, `EXISTS`, `BETWEEN`, or 69 more ...
  | |      |
  | |      while parsing `{...}`
  | |      while parsing expression
  | while parsing `SELECT ...`

mysql> select {a: to_boolean(1)} from test;
ERROR 1105 (HY000): SyntaxException. Code: 1005, Text = error:
  --> SQL:1:22
  |
1 | select {a: to_boolean(1)} from test
  | ------ -             ^ expected `}` or `:`
  | |      |
  | |      while parsing `{...}`
  | |      while parsing expression
  | while parsing `SELECT ...`
```

ariesdevil:

## Summary

The current implementation has an ambiguity for forms like `a:b`. Since Snowflake supports accessing map objects with `:`, we'll eliminate the ambiguity by restricting map creation to literal support only.

- Closes #12695


b41sh (approved):
LGTM

src/query/ast/src/ast/expr.rs
```diff
@@ -212,7 +212,10 @@ pub enum Expr {
     /// The `Array` expr
     Array { span: Span, exprs: Vec<Expr> },
     /// The `Map` expr
-    Map { span: Span, kvs: Vec<(Expr, Expr)> },
+    Map {
```
andylokandy, line 215: Key in `Expr` indeed causes ambiguity, but why can't values be Expr?

BohuTANG: @ariesdevil

andylokandy: Wouldn't sql like `select {num: number + 1} from numbers(100)` be useful?

ariesdevil: Seems no user using it that way. BTW: duckdb supports using expression for both key/value, but they don't support using `:` to access.

andylokandy: I mean, is there a simple way to construct bunch of maps based on query results. A more practical example would be `INSERT INTO orders_json_data (data) SELECT {order_id: order_id, order_timestamp: timestamp}` from old_order join ....`.

BohuTANG: I think it's better to have, we can open an new issue here(may not in high-priority), also cc @b41sh

PR #12860, opened 2023-09-15
chore: assert number of block not exceed the max value required by virtual column

dantengsky:

## Summary

in case malformed segments are constructed, with a number of blocks that exceed that virtual-column assumes  

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12860)


JackTan25 (approved):
LGTM

PR #12866, opened 2023-09-15
chore(planner): fix infer filter

Dousir9:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12866)


tests/sqllogictests/suites/mode/standalone/explain_native/infer_filter.test
```diff
@@ -541,6 +541,94 @@ HashJoin
...
+    ├── read rows: 0
+    ├── read bytes: 0
+    ├── partitions total: 0
+    ├── partitions scanned: 0
+    ├── push downs: [filters: [t1.a (#0) > 10], limit: NONE]
+    └── estimated rows: 0.00
+
+# false
+query T
+explain select * from t1, t2 where t1.a = t2.a and t1.a > 5 and t2.a < 1;
```
xudong963, line 604: Filter false is pushed down probe side and still need to build hash table, in the future, maybe we can push down filter both build and probe side?

Dousir9: Yes, when the condition is false, we need a dummy operator to replace the original one, but some problems encountered and it will be added in the next PR.

xudong963 (approved):

PR #12889, opened 2023-09-18
chore(query): check group by contains agg function or not

### LINKED ISSUE #12853: bug: `max_if` function with a group by return an error message
### Search before asking

### Version

v1.2.112-nightly

### What's Wrong?

`max_if`/`min` function with a group by returning an error message

### How to Reproduce?

```sql
mysql> create table test(a boolean not null, b uint8 not null);

mysql> insert into test values(true, 10),(false, 20);
Query OK, 2 rows affected (0.03 sec)

mysql> SELECT max_if(1, a), to_timestamp(5140325614630), 1 FROM test GROUP BY to_timestamp(5140325614630), 1;
ERROR 1105 (HY000): BadArguments. Code: 1006, Text = Unable to get field named "18446744073709551615". Valid fields: ["0"].

mysql> SELECT min(NULL), 1, a, b FROM test GROUP BY ROLLUP (1, a, b);
ERROR 1105 (HY000): BadArguments. Code: 1006, Text = Unable to get field named "18446744073709551615". Valid fields: ["0", "1"].
```

sundy-li:

## Summary

Summary about this PR

- Closes #12853

### DIFF
--- a/src/query/sql/src/planner/binder/aggregate.rs
+++ b/src/query/sql/src/planner/binder/aggregate.rs
@@ -692,6 +692,20 @@ impl Binder {
             );
         }
 
+        // Check group by contains aggregate functions or not
+        for item in bind_context.aggregate_info.group_items.iter() {
+            let f = |scalar: &ScalarExpr| matches!(scalar, ScalarExpr::AggregateFunction(_));
+            let finder = Finder::new(&f);
+            let finder = item.scalar.accept(finder)?;
+
+            if !finder.scalars().is_empty() {
+                return Err(ErrorCode::SemanticError(
+                    "GROUP BY items can't contain aggregate functions".to_string(),
+                )
+                .set_span(item.scalar.span()));
+            }
+        }
+
         // If it's `GROUP BY GROUPING SETS`, ignore the optimization below.
         if collect_grouping_sets {
             return Ok(());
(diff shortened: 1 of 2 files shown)


src/query/functions/src/scalars/string_multi_args.rs
```diff
@@ -145,10 +141,7 @@ pub fn register(registry: &mut FunctionRegistry) {
                         }
                     }
 
-                    match len {
-                        Some(_) => Value::Column(Column::String(builder.build())),
-                        _ => Value::Scalar(Scalar::String(builder.build_scalar())),
-                    }
+                    Value::Column(Column::String(builder.build()))
```
b41sh, line 144: If the source argument is a `Scalar`, should the return value also be a `Scalar`?

src/query/sql/src/planner/binder/aggregate.rs
```diff
@@ -692,6 +692,20 @@ impl Binder {
             );
         }
 
+        // Check group by contains aggregate functions or not
+        for item in bind_context.aggregate_info.group_items.iter() {
+            let f = |scalar: &ScalarExpr| matches!(scalar, ScalarExpr::AggregateFunction(_));
```
ariesdevil, line 697: Also needs check window function

src/query/functions/src/scalars/string_multi_args.rs
sundy-li: ok

b41sh: We can't determine whether the source argument is a `Scalar` or a `Column` based on `num_rows`, a `Column` may have only one row of values.

sundy-li: If the column contains only one row, it's ok to return scalar.

We already support scalar and full columns in a `DataBlock`

ariesdevil (approved):

b41sh (approved):

PR #12899, opened 2023-09-18
refactor: http handler return error position in detail field.

youngsofun:

## Summary

Summary about this PR

e.g.

```
{
  "code": 1046,
  "message": "Number of columns in file (2) does not match that of the corresponding table (3)",
  "detail": "at file 'select.csv', line 1"
}
```

ErrorCode is 136 bytes after add another string field ,clippy warn about it, so I let Range use u32 instead of usize to reduce size to 128 bytes, to be more efficent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12899)

### DIFF
--- a/src/common/exception/src/span.rs
+++ b/src/common/exception/src/span.rs
@@ -15,22 +15,19 @@
 use std::fmt::Debug;
 use std::fmt::Display;
 use std::fmt::Formatter;
-use std::ops::Bound;
-use std::ops::RangeBounds;
 
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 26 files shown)


BohuTANG:
cc @Chasen-Zhang for the change

src/common/exception/src/span.rs
```diff
@@ -15,22 +15,19 @@
...
+    pub fn start(&self) -> usize {
+        self.start as usize
     }
-
-    fn end_bound(&self) -> Bound<&usize> {
-        Bound::Excluded(&self.end)
+    pub fn end(&self) -> usize {
+        self.end as usize
     }
 }
```
andylokandy, line 32: I think you can remove them

PR #12904, opened 2023-09-19
ci: sqllogictest match both error code and error message.

youngsofun:

## Summary

1.  format http error message, so the main part of it is the same as that of mysql/clickhouse handler.
2.  make full use of regex to match both code and message. note:   1. need `(?s)` to match multiline. 2. meta characters like `(` needs to be escaped.

if this pattern is ok, I will update the remaining sqllogictest files in next pr.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12904)

### DIFF
--- a/tests/sqllogictests/suites/crdb/and_or
+++ b/tests/sqllogictests/suites/crdb/and_or
@@ -7,7 +7,7 @@ CREATE TABLE t (k INT null, a INT null, b INT null)
 statement ok
 INSERT INTO t VALUES (1, NULL, NULL), (2, NULL, 1), (3, 1, NULL), (4, 2, 0), (5, 3, 3)
 
-statement error 1001
+statement error (?s)1001.*divided by zero while evaluating function `divide\(3, 0\)`
 SELECT a <> 2 AND 3 / b = 1 FROM t ORDER BY k
 
 query I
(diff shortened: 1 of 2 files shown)


tests/sqllogictests/suites/crdb/and_or
```diff
@@ -7,18 +7,18 @@ CREATE TABLE t (k INT null, a INT null, b INT null)
...
+statement error 1001.*divided by zero while evaluating function `divide\(3, 0\)`
 SELECT a <> 2 AND 3 / b = 1 FROM t ORDER BY k
 
 query I
 SELECT a FROM t WHERE a <> 2 AND 3 / b = 1 ORDER BY k
 ----
 3
 
-statement error 1001
+statement error 1001.*divided by zero while evaluating function `divide\(3, 0\)`
```
sundy-li, line 18: can we omit `.*` to be spaces?

And we only use string contains to match the error text.

youngsofun: the string after `error` is regex.
currently error code is not used, but maybe we need to standardize ant test it later.

if we need string match only,   it is better if sqllogictest-rs support `error_substring` directly so we don`t need to escape.

BohuTANG (approved):

PR #12912, opened 2023-09-19
fix(parser): allow expr in map liteal

andylokandy:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12912)


ariesdevil:
I just saw this PR, we now have two ways to create a map, using either `{}` or the map function, and we need to make sure that the two ways are unified.
After this PR, the `{}` supports value as an expression but the key is not.
The map function supports both key/value expressions.
cc @b41sh @BohuTANG

PR #12925, opened 2023-09-19
fix: fix disallow scalar function with window

### LINKED ISSUE #12914: Should scalar function parse window func's over can success?
**Summary**

Now:

```sql
select to_date('2022-02-02') over(order by (to_date('2017-12-04') - to_date('2017-02-02')));
+---------------------------------------------------------------------------------------+
| to_date('2022-02-02') over (order by (to_date('2017-12-04') - to_date('2017-02-02'))) |
+---------------------------------------------------------------------------------------+
| 2022-02-02                                                                            |
+---------------------------------------------------------------------------------------+
1 row in set (0.012 sec)
```

ariesdevil:

## Summary

disallow scalar function with window

- Closes #12914

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12925)

### DIFF
--- a/src/query/sql/src/planner/semantic/type_check.rs
+++ b/src/query/sql/src/planner/semantic/type_check.rs
@@ -696,6 +696,22 @@ impl<'a> TypeChecker<'a> {
                     }
                 }
 
+                // check window function legal
+                {
+                    let supported_window_funcs = AggregateFunctionFactory::instance()
+                        .registered_names()
+                        .into_iter()
+                        .chain(GENERAL_WINDOW_FUNCTIONS.iter().cloned().map(str::to_string))
+                        .collect::<Vec<String>>();
+                    let name = func_name.to_lowercase();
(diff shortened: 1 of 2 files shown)


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -650,7 +650,7 @@ impl<'a> TypeChecker<'a> {
                 params,
                 window,
                 lambda,
-            } => {
+            } if window.is_none() => {
```
b41sh, line 653: It is better to determine the type of the function by its name, and then determine whether the `window` should be `None` or `Some`. Separately handle will result in the window function not being able to find similar function names if the function name is not correct. We have code logic for finding similar function names between lines 662 and 695.
https://github.com/datafuselabs/databend/pull/12925/files#diff-8e24042e5eec79cd29dda9776333d9ecad8cba883a80e4590c31f594ea745795R662-R695

RinChanNOWWW (approved):

PR #12927, opened 2023-09-19
chore(query): support cluster level process list

zhang2014:

## Summary

- chore(query): support cluster level process list
- chore(query): support cluster level kill query

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12927)

### DIFF
--- a/src/query/storages/system/src/processes_table.rs
+++ b/src/query/storages/system/src/processes_table.rs
@@ -45,6 +45,8 @@ pub struct ProcessesTable {
 impl SyncSystemTable for ProcessesTable {
     const NAME: &'static str = "system.processes";
 
+    const IS_LOCAL: bool = false;
+
     fn get_table_info(&self) -> &TableInfo {
         &self.table_info
     }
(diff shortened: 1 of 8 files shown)


sundy-li (approved):

src/query/storages/system/src/processes_table.rs
```diff
@@ -45,6 +45,8 @@ pub struct ProcessesTable {
 impl SyncSystemTable for ProcessesTable {
     const NAME: &'static str = "system.processes";
 
+    const IS_LOCAL: bool = false;
```
BohuTANG, line 48: Need to add the `nodeid` to processes table?

For a query which distributed executing in 2-nodes cluster: query0 and query1:
Does the query id are same in query0 and query1?

BohuTANG: If they are same no need add the `nodeid` to the processes table.

zhang2014: > If they are same no need add the `nodeid` to the processes table.

for a distributed query, it will only show once in processes table.

zhang2014: let's keep it consistent with the standalone mode.

PR #12928, opened 2023-09-19
refactor(query): improve hash join

Dousir9:

## Summary

Summary about this PR

Improve the `kernels`(`take.rs`, `take_chunks.rs`, `take_compact.rs`, `concat.rs`, `filter.rs`) and reduce memory usage:
1. Use low-level operations as much as possible.
2. Before building a `StringColumn`, we first scan `iter` and calculate the space required by it. This can reduce the memory usage of `StringColumn` and avoid the `resize` and `grow` operations of `Vec`; 
3. When the output_schema of hash join includes `StringColumn`, we add `string_items_buf` to avoid allocating memory frequently in `kernels`.
4. `ctx.get_function_context()` and `ctx.get_settings()` have a certain amount of overhead, so we should call them as little as possible when building pipelines.
5. For concat, when merging two Datablocks, we need Vec push num_rows times before, now we only need one copy_nonoverlapping.

The ci-benchmark TPC-H standalone results:
+ Q9:   21.2s -> 19.9s
+ Q10: 12.3s -> 11.7s
+ Q18: 35.4s -> 30.9s
截屏2023-09-23 10 09 55

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12928)


BohuTANG:
Conflicting files
src/query/service/src/pipelines/pipeline_builder.rs

src/common/hashtable/src/hashjoin_hashtable.rs
```diff
@@ -109,7 +109,7 @@ where
 
     fn contains(&self, key_ref: &Self::Key) -> bool {
         let index = key_ref.hash() as usize & self.hash_mask;
-        let mut raw_entry_ptr = self.pointers[index];
```
zhang2014, line 112: Do we really need unsafe?

Dousir9: > Do we really need unsafe?

The index here must be safe, so we can use `get_unchecked` to get a very small performance improvement. 😅

xudong963: I think we should use unsafe as little as possible, usually adding unsafe gives zero perf gain, compiler can do many things. But unsafe can increase vulnerability of our code.

Even if there really might be a performance gain, we should have a separate PR with only unsafe modifications to ensure that the principle of singularity proves whether or not it's a performance gain from unsafe.

Dousir9: Yes, but here we use hash & mask to get the index, it must be safe, I will add some #Safety comments, because the probe_hash_table will be called many times, it is worth doing.

Dousir9:
I will add some `#Safety` comments.

zhang2014:
Can we observe performance changes after remove these unsafe content?

Dousir9:
> Can we observe performance changes after remove these unsafe content?

These codes use pointers in many places, and unsafe cannot be removed. Unsafe will bring us a great performance improvement. For example, concat, when merging two `Datablocks`, we need `Vec` push `num_rows` times, but with pointers, we only need one `copy_nonoverlapping`.

src/query/expression/src/kernels/take.rs
```diff
@@ -144,33 +160,143 @@ impl Column {
...
-            Column::Variant(column) => Self::take_arg_types::<VariantType, _>(column, indices),
+            Column::Variant(column) => VariantType::upcast_column(Self::take_string_types(
+                column,
+                indices,
+                string_items_buf.as_mut(),
+            )),
+        }
+    }
+
+    pub fn take_primitive_types<T, I>(col: &Buffer<T>, indices: &[I]) -> Vec<T>
```
sundy-li, line 191: no need unsafe , just use `Vec::from_iter`

Dousir9: Done.

sundy-li (approved):

zhang2014 (approved):

PR #12934, opened 2023-09-20
feat(query): support inverted_filter to omit the filter executor

### LINKED ISSUE #12960: [bug]: delete from filter null bug
**Summary**

```
create table t(c int);

insert into t values(1),(null);

delete from t where c >= 1;

select * from t;

-- empty set expect to have null left

```

sundy-li:

## Summary

Summary about this PR

Make parquet-rs reader support inverted_filter.

```sql
select count() from lineitem where l_shipdate <= CAST('4998-09-02' AS date);  ---  0.082s --> 0.019s
```

TODO:  introduce inverted_filter into fuse 

- Closes #12960 

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12934)


sundy-li:
askbend:summary

src/query/catalog/src/plan/pushdown.rs
```diff
@@ -63,6 +63,8 @@ pub struct PrewhereInfo {
     /// filter for prewhere
     /// Assumption: expression's data type must be `DataType::Boolean`.
     pub filter: RemoteExpr<String>,
+    /// inverted_filter  for prewhere
+    pub inverted_filter: RemoteExpr<String>,
```
RinChanNOWWW, line 67: `inverted_filter` is not always needed. How about change this to an `Option`.

For example, the `do_mutation_block_pruning` call in `FuseTable::try_add_update_source` doesn't need `inverted_filter`.

src/query/sql/src/planner/expression_parser.rs
```diff
@@ -141,7 +143,7 @@ pub fn parse_to_remote_string_expr(
     ctx: Arc<dyn TableContext>,
     table_meta: Arc<dyn Table>,
     sql: &str,
-) -> Result<RemoteExpr<String>> {
+) -> Result<Filters> {
```
RinChanNOWWW, line 146: Why change the return type to `Filters`? It looks like the new return type doesn't fit the function name.

src/query/storages/fuse/src/operations/update.rs
```diff
@@ -257,8 +270,7 @@ impl FuseTable {
         let (parts, part_info) = self
             .do_mutation_block_pruning(
                 ctx.clone(),
-                filter,
-                None,
+                filters,
```
RinChanNOWWW, line 273: In the original codes, the inverted filter is `None`. However, it is changed to `Some` here.

src/query/storages/parquet/src/parquet_rs/parquet_reader/row_group.rs
```diff
@@ -96,7 +96,7 @@ pub struct InMemoryRowGroup<'a> {
     metadata: &'a RowGroupMetaData,
     page_locations: Option<&'a [Vec<PageLocation>]>,
     column_chunks: Vec<Option<Arc<ColumnChunkData>>>,
-    row_count: usize,
+    pub row_count: usize,
```
RinChanNOWWW, line 99: I think it's better to expose a public method instead of exposing a public member.

src/query/storages/parquet/src/parquet_rs/parquet_table/partition.rs
```diff
@@ -356,11 +356,11 @@ fn prune_and_generate_partitions(
             ..
         } = meta.as_ref();
         part_stats.partitions_total += meta.num_row_groups();
-        let rgs = pruner.prune_row_groups(meta, row_group_level_stats.as_deref())?;
+        let (rgs, omits) = pruner.prune_row_groups(meta, row_group_level_stats.as_deref())?;
```
RinChanNOWWW, line 359: If all omits, we can skip the `prune_pages` below.

src/query/storages/fuse/src/operations/update.rs
sundy-li: here we combine it into one struct named `Filters`.

`delete.rs` also calls this function but it need to pass  the inverted filter

RinChanNOWWW (approved):

src/query/storages/fuse/src/operations/update.rs
sundy-li: cc @JackTan25

JackTan25: it's well, the update won't use this, it will do "read_pruning", and update doesn't need inversed filter, because if one block need to be updated ,we will read the whole block, but for delete, we don't need to do this, just give a delete log for commit meta.

BohuTANG:
Conflicting files
src/query/storages/parquet/tests/it/parquet_rs/prune_row_groups.rs

PR #12965, opened 2023-09-21
refactor: introduce QueryKind.

youngsofun:

## Summary

`query_kind` was originally only used in query_log, so it is just a `String`.
later we found it useful as a lightweight context in some scenarios,  
let us formalize it as an `Enum` for safety.

`common-sql` already depends on `common-catalog`, so no extra deps is introduced.
 
Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12965)


src/query/sql/src/planner/plans/plan.rs
```diff
@@ -302,123 +304,121 @@ pub enum RewriteKind {
     Call,
 }
 
-impl Display for Plan {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+impl Plan {
+    pub fn kind(&self) -> QueryKind {
```
sundy-li, line 308: Notice that QueryKind is not accurate.

Some statements have been rewritten to Query like `ShowTables`

sundy-li: Why not match the AST statement ?

sundy-li: we can cache the  AST statement  in context.


```
pub(in crate::sessions) running_query_stmt: Statment,
```

youngsofun: currently only used for copy.

youngsofun: nn

> we can cache the AST statement in context.
> 
> ```
> pub(in crate::sessions) running_query_stmt: Arc<RwLock<Option<Statement>>>,,
> ```

need to be serialized and sent to other query nodes.

youngsofun: also comparation is faster

youngsofun: > nn
> 
> > we can cache the AST statement in context.
> > ```
> > pub(in crate::sessions) running_query_stmt: Arc<RwLock<Option<Statement>>>,,
> > ```
> 
> need to be serialized and sent to other query nodes.

Statement not impl Serialize for now, even if it does, it it too big for current usage.

src/query/storages/parquet/src/parquet2/parquet_table/partition.rs
```diff
@@ -159,7 +160,7 @@ impl Parquet2Table {
                 &file_locations,
                 ctx.get_settings().get_max_threads()? as usize,
                 &ctx.get_copy_status(),
-                ctx.get_query_kind().eq_ignore_ascii_case("copy"),
+                ctx.get_query_kind() == QueryKind::Copy,
```
ariesdevil, line 163: The convention for rust is to compare enum with `matches!`

src/query/sql/src/planner/plans/plan.rs
sundy-li: Ok, let's respect `RewriteKind` in query plan.

youngsofun: QueryKind is from Plan,  those stmt in RewriteKind are already rewritten to Query at this point.

src/query/storages/parquet/src/parquet2/parquet_table/partition.rs
youngsofun: done

youngsofun:
@sundy-li  plz review again https://github.com/datafuselabs/databend/pull/12965/commits/1d4312f6c3b9163810b78fc3592f3f382a3ff569

sundy-li (approved):

PR #12978, opened 2023-09-22
feat(http): tell the client the settings is UNSET on reply 

flaneur2020:

## Summary

Currently when users executed `UNSET <settings>`, it found the default values on settings. But this default value may mislead on a different cluster with different default configs, like `max_threads = 4` on a 64 cores instance.

This PR added an `is_unset` parameter on the ChangeSettings affect, allowing the client to delete the local memorized session settings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12978)


src/query/service/src/interpreters/interpreter_unsetting.rs
```diff
@@ -92,9 +93,10 @@ impl Interpreter for UnSettingInterpreter {
             }
         }
         self.ctx.set_affect(QueryAffect::ChangeSettings {
-            keys,
+            keys: keys,
```
youngsofun, line 96:
```suggestion
            keys,
```

src/query/service/src/servers/http/v1/query/http_query.rs
```diff
@@ -139,25 +139,35 @@ pub struct HttpSessionConf {
 }
 
 impl HttpSessionConf {
+    // TODO(liyz): this function is not needed? Client side should apply the affect itself,
+    // and pass the db and settings parameters in the "session" field like:
+    // session: {database: "default", settings: {max_threads: "4"}}.
+    // do not actually need to apply these affect again on the server side.
```
youngsofun, line 145: with this, client just need to update local session as a whole,  make it  easier.

youngsofun: the updating role may get complicated, client code not need to update

youngsofun: session by its nature is server side, client side session is a workaound.
only a few change need also apply to client (like database).
so it is better to not add this burden to client.

youngsofun: session by its nature is server side.
only a few change need also apply to client (like database).
so it is better to not add this burden to client.

src/query/service/src/servers/http/v1/query/http_query.rs
```diff
@@ -139,25 +139,35 @@ pub struct HttpSessionConf {
...
             }
             QueryAffect::ChangeSettings {
                 keys,
                 values,
+                is_unset,
                 is_globals: _,
             } => {
+                // do not change the current settings if reset is true
+                if *is_unset {
+                    return ret;
```
youngsofun, line 161:
```suggestion
                    for key in keys.iter() {
                         settings.remove(key);
                    }
```

youngsofun: I think this solve your problem

BohuTANG:
When `unset <setting>` will be `set <setting> = default`,  not a real number.

TCeason:
> When `unset <setting>` will be `set <setting> = default`,  not a real number.
> 
> 
> 
> For example: `unset max_threads ` will be `set max_threads = default`, the `default` is the default value.
> 
> 
> 
> cc @TCeason: `unset max_threads` == `max_threads = default`
> 
> 

If max threads in conf is 0 ,will set conf.max_thread

BohuTANG:
> If max threads in conf is 0 ,will set conf.max_thread

If we do `unset max_threads`, the `max_threads` value will be, CPUs or 0(default)?

BohuTANG:
As discussed, everything is clear. Let's contiue this PR going.

src/query/service/src/servers/http/v1/query/http_query.rs
flaneur2020: > the updating rule may get complicated or change like this pr, client code not need to update
> session by its nature is server side, client side session is a workaound.

no, the client side may also have it's customized settings in their config like:

```
driver = sql.Open("databend://<host>/<db>?max_sessions_limit=1024")
```

this limit might not actually scoped by the sessions' lifetime, but it's considered long as the client's lifetime.

flaneur2020:
> > Currently when users executed `UNSET <settings>`, it found the default values on settings. But this default value may mislead on a different cluster with different default configs, like `max_threads = 4` on a 64 cores instance.
> 
> isn`t it by design?

no, it'd better to handle the unset behaviour in the client side explicitly

src/query/service/src/interpreters/interpreter_unsetting.rs
flaneur2020: github seems got a bug on display the diff... 

when I pulled the code in my laptop, it shows `keys` but not `keys: keys`...

src/query/service/src/servers/http/v1/query/http_query.rs
flaneur2020: I added a `settings.remove(key)` as you suggested, however it's not enough to resolve the issue. the client side still have to learn about when the settings was unsetted, or else the client will keep sending the settings stored in the client on the next time.

youngsofun: >>>  client will keep sending the settings stored in the client on the next time.

this will not happen

```settings.remove(key); ```
this settings is sent to client,  client update local setting as a whole,

youngsofun: the most simple example is in the client for sqllogict test

send https://github.com/datafuselabs/databend/blob/4784d7f31a3222b3e24b03059ec734521e6a4d61/tests/sqllogictests/src/client/http_client.rs#L128

recv
https://github.com/datafuselabs/databend/blob/4784d7f31a3222b3e24b03059ec734521e6a4d61/tests/sqllogictests/src/client/http_client.rs#L113

src/query/service/src/servers/http/v1/query/http_query.rs
youngsofun: I think this is not the common case.

I simple usage, server side do most of the work, client only handle session simply like a blob, no need to know the structure and how to update it.  it also make life much easier when we want to update/fix about the session object..

for advanced clients like you mentioned , you can simply ignore what server return,
just maintain and **construct** the session part as you like and send to server when post query.

youngsofun: `QueryAffect` is the (change) event. 

 `Session` is the state, `reply state`  = `req state` [ + `change event`]. it is in fact redundant (since client know `req state` and  `change event`), we introduce it only to make both client and server side easier.

src/query/service/src/servers/http/v1/query/http_query.rs
flaneur2020: oh, this makes sense. let me rethink about the design of this PR based this api contract.

src/query/service/src/servers/http/v1/query/http_query.rs
flaneur2020: I think it's not a good design, if we'd hope to sync what the things have changed during the SQL to the client, we'd better encapsulate it to a single place.

In my previous understanding, QueryAffect plays this role to tell what had changed in the SQL to the client about the local state. But now there're two places to tell the clients what had changed and have different meanings.

In this scenario, I suppose that it'd better to still make QueryAffect to play its role, tells the client the new completed set of session level settings whenever the session had changed.

flaneur2020:
after discussed with @youngsofun, the related code here is beyond my initial estimation, I'd close this PR and raise a new one.

PR #12787, opened 2023-09-11
feat(observability): replace metrics-rs with prometheus-client

flaneur2020:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

This PR migrated all the metrics into prometheus-client, and:

- [x] change the /metrics endpoint to promtheus-client
- [x] unit test about the metrics endpoint
- [x] change the metrics-table to prometheus-client
- [x] remove metric-rs's recorder
- ~support reset metrics~ (in another PR)

Closes #12688
Closes #12870

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12787)

PR #12982, opened 2023-09-22
fix(query): fix session manager dead lock

zhang2014:

## Summary

fix(query): fix session manager dead lock

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12982)


zhang2014:
root cause https://github.com/datafuselabs/databend/pull/12787/files#diff-7003e8bdd7728475cd4979b7b53a359579076ec297f87d6fef9d3a466463ceb7R174 CC: @flaneur2020

Dousir9 (approved):

BohuTANG:
We should do some comments here, why caused deadlock?

zhang2014:
> We should do some comments here, why caused deadlock?

https://github.com/datafuselabs/databend/pull/12984

Issue #12976, opened 2023-09-22
Bug: failed to COPY a 4 columns CSV into a 5 columns table with one trailing column with DEFAULT


flaneur2020:
**Summary**

```
CREATE TABLE `people` (
  `id` BIGINT,
  `first_name` VARCHAR,
  `last_name` VARCHAR,
  `age` INT
) ENGINE=FUSE;

// add one column with DEFAULT value
ALTER TABLE people ADD COLUMN age2 INT DEFAULT 0;
```

```
COPY INTO eric.`people`
  PATTERN ='temp1.csv'
  FILE_FORMAT = (TYPE = 'CSV' field_delimiter = '|' record_delimiter = '\n' skip_header = 1 )
 DISABLE_VARIANT_CHECK = TRUE ;
```

Sample data in CSV, it have 4 columns.

```
"1363"|"j"|"j"|"10"
"1364"|"k"|"k"|"11"
"1365"|"l"|"l"|"12"
"1366"|"m"|"m"|"13"
"1367"|"n"|"n"|"14"
"1368"|"o"|"o"|"15"
"1369"|"p"|"p"|"16"
"1370"|"q"|"q"|"17"
```

error:

```
1046=>Number of columns in file (4) does not match that of the corresponding table (5)
```

flaneur2020:
when I try:

```
INSERT INTO people VALUES (1, "a", "b", 12);
```

I got the following error:

```
Table columns count is not match, expect 5, input: 4, expr: [Literal { span: Some(1..2), lit: UInt64(1) }, ColumnRef { span: Some(4..8), database: None, table: None, column: Name(Identifier { name: "li", quote: Some('"'), span: Some(4..8) }) }, ColumnRef { span: Some(10..14), database: None, table: None, column: Name(Identifier { name: "jo", quote: Some('"'), span: Some(10..14) }) }, Literal { span: Some(16..18), lit: UInt64(12) }]
```

but postgres could handle the trailing columns with DEFAULT gracefully:

![Screenshot 2023-09-22 at 18 13 51](https://github.com/datafuselabs/databend/assets/129800/1140e07b-07cd-4227-8564-f2c19351e85b)

sundy-li:
how about making copy support  optional column list like insert?

youngsofun:
> how about making copy support optional column list like insert?

do you mean `copy into t(c1, c2) ... `

[...]

PR #12990, opened 2023-09-23
feat: csv add option error_on_column_count_mismatch.

youngsofun:

## Summary

Summary about this PR

- Closes https://github.com/datafuselabs/databend/issues/12976

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12990)

### DIFF
--- a/src/meta/proto-conv/tests/it/v032_file_format_params.rs
+++ b/src/meta/proto-conv/tests/it/v032_file_format_params.rs
@@ -48,6 +48,7 @@ fn test_decode_v32_csv_file_format_params() -> anyhow::Result<()> {
             nan_display: "nan".to_string(),
             escape: "\\".to_string(),
             quote: "\'".to_string(),
+            error_on_column_count_mismatch: true,
         })
     };
     common::test_load_old(func_name!(), file_format_params_v32.as_slice(), 0, want())?;
--- a/src/meta/protos/proto/file_format.proto
+++ b/src/meta/protos/proto/file_format.proto
@@ -104,6 +104,9 @@ message CsvFileFormatParams {
   string escape = 6;
   string quote = 7;
   string null_display = 8;
+  // corresponding to `!error_on_column_count_mismatch`
+  // for we can not set default value in proto3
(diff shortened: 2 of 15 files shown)


src/meta/protos/proto/file_format.proto
```diff
@@ -104,6 +104,9 @@ message CsvFileFormatParams {
   string escape = 6;
   string quote = 7;
   string null_display = 8;
+  // corresponding to `!error_on_column_count_mismatch`
+  // for we can not set default value in proto3
+  bool allow_column_count_mismatch = 9;
```
ariesdevil, line 109: Seems the boolean default value in protobuf3 is false.
https://protobuf.dev/programming-guides/proto3/#default

youngsofun: yes, that is why I do not use error_on_column_count_mismatch here

drmingdrmer (commented):
> Reviewed 12 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12990)* status: 12 of 16 files reviewed, 1 unresolved discussion (waiting on @andylokandy and @ariesdevil)

drmingdrmer (approved):

src/meta/proto-conv/tests/it/v032_file_format_params.rs
```diff
@@ -48,6 +48,7 @@ fn test_decode_v32_csv_file_format_params() -> anyhow::Result<()> {
             nan_display: "nan".to_string(),
             escape: "\\".to_string(),
             quote: "\'".to_string(),
+            error_on_column_count_mismatch: true,
```
Xuanwo, line 51: It would be better to use `false` as the default value for easier understanding of our options. What do you think about using `disable_column_count_check` or something similar?

BohuTANG: This PR we keep the same option and behavior as snowflake.

BohuTANG:
cc @soyeric128  for this option document, thanks.

PR #12996, opened 2023-09-24
fix: cast float64 to decimal panic #12910

### LINKED ISSUE #12910: bug: cast float64 to decimal panic
### Search before asking

### Version

v1.2.118-nightly

### What's Wrong?

cast float64 to decimal panic
```
panicked at 'called `Option::unwrap()` on a `None` value', src/query/expression/src/types/decimal.rs:388:25
```

### How to Reproduce?

```sql
mysql> create table test (a double);

mysql> insert into test values (0.0941881516339449), (0.8833429252655464);
Query OK, 2 rows affected (0.04 sec)

mysql> select try_to_decimal(20, 19)(a) from test;
ERROR 2013 (HY000): Lost connection to MySQL server during query
Connection id:    20
```

zenus:

## Summary

Summary about this PR

- Closes #12910

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12996)


src/query/expression/src/types/decimal.rs
```diff
@@ -542,7 +542,8 @@ impl Decimal for i256 {
     }
 
     fn from_float(value: f64) -> Self {
-        i256::from(value.to_i128().unwrap())
+        let bs = value.to_le_bytes().to_vec();
+        Self::de_binary(&mut bs.as_slice())
```
sundy-li, line 546: we should rewrite the de_binary function to be  :
```
fn de_binary(bytes: &[u8]) -> Self 
```

thus we can omit to `to_vec`

zenus: ok

tests/sqllogictests/suites/query/02_function/02_0066_function_try_to_decimal
```diff
@@ -0,0 +1,14 @@
+statement ok
+CREATE TABLE test (a double) ENGINE=Memory
+
+statement ok
+INSERT INTO test VALUES (0.0941881516339449), (0.8833429252655464)
+
+query F 
+SELECT try_to_decimal(20, 19)(a) FROM test;
```
BohuTANG, line 8: Simple way:

```
select try_to_decimal(20, 19)(a) from (values(0.0941881516339449),(0.8833429252655464))t(a);
```

src/query/expression/src/types/decimal.rs
sundy-li: Yes, maybe we could use rewrite to `fn de_binary(bytes: &[u8]) `, we don't need the mut reference

tests/sqllogictests/suites/query/02_function/02_0066_function_try_to_decimal
b41sh: select values from table is better, as literal value is always parsed as a decimal value

src/query/expression/src/types/decimal.rs
```diff
@@ -378,7 +379,35 @@ impl Decimal for i128 {
     }
 
     fn from_float(value: f64) -> Self {
-        value.to_i128().unwrap()
+       //still needs to be optimized.
+       //An implementation similar to float64_as_i256 obtained from the ethnum library
```
sundy-li, line 383: So why not use `AsI256:: as_i256(value)` directly?

src/query/expression/src/types/decimal.rs
zenus: > Yes, maybe we could use rewrite to `fn de_binary(bytes: &[u8]) `, we don't need the mut reference

I have updated

sundy-li (approved):

src/query/expression/src/types/decimal.rs
```diff
@@ -378,7 +379,34 @@ impl Decimal for i128 {
     }
 
     fn from_float(value: f64) -> Self {
-        value.to_i128().unwrap()
+        // still needs to be optimized.
```
BohuTANG, line 382: It's great if we have some unit test for the from_float bound check, for example:
```

    #[test]
    fn test_from_float_positive() {
        let value = 123.456;
        let result = from_float(value);
        assert_eq!(result, 123); // The fractional part should be truncated
    }

    #[test]
    fn test_from_float_negative() {
        let value = -123.456;
        let result = from_float(value);
        assert_eq!(result, -123); // The fractional part should be truncated
    }

    #[test]
    fn test_from_float_zero() {
        let value = 0.0;
        let result = from_float(value);
        assert_eq!(result, 0); // Zero should be converted to zero
    }

    #[test]
    fn test_from_float_large() {
        let value = 1e20;
        let result = from_float(value);
        assert_eq!(result, i128::MAX); // Large values should be capped to i128::MAX
    }

    #[test]
    fn test_from_float_small() {
        let value = -1e20;
        let result = from_float(value);
        assert_eq!(result, i128::MIN); // Small values should be capped to i128::MIN
    }

    #[test]
    fn test_from_float_sub_one_positive() {
        let value = 0.123;
        let result = from_float(value);
        assert_eq!(result, 0); // Values less than one should be converted to zero
    }

    #[test]
    fn test_from_float_sub_one_negative() {
        let value = -0.123;
        let result = from_float(value);
        assert_eq!(result, 0); // Values less than negative one should be converted to zero
    }

```

zenus: let me add

tests/sqllogictests/suites/query/02_function/02_0066_function_try_to_decimal
```diff
@@ -0,0 +1,14 @@
...
+----
+0.0941881516339448960
+0.8833429252655463424
```
b41sh, line 11: this test is cast a float to decimal128, how about adding a test for decimal256? 
for example:
```sql
SELECT try_to_decimal(72, 70)(a) FROM test;
```

zenus:
```rust
---- decimal::test_float_to_128 stdout ----
thread 'decimal::test_float_to_128' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
<100000000000000000000
>170141183460469231731687303715884105727
```
unit test  can not pass, do you have any ideas?

b41sh: It looks like the expected value is wrong, `i128::MAX` is `170_141_183_460_469_231_731_687_303_715_884_105_727i128`, which is not equal to `1e20`, can we change the test case?

zenus: @b41sh  the code actullay use u64

b41sh: `1e20` is not equal to `i128::MAX`, we need to change the expected value.

b41sh: @zenus Any progress? Maybe we can just delete this test case.

zenus: i will update today.

b41sh (approved):
Hi, @zenus This bug is blocking and we want to merge this PR soon, so I've modified the test for you, thanks for your contribution!

BohuTANG:
Will merge manually after CI passed.

PR #12997, opened 2023-09-25
chore: replace max_io_request

JackTan25:
we need to use thread_nums to limit the permit_nums , in some cases, if the permit num is too large, we will get oom. 
## Summary

Summary about this PR

- Closes #issue

### DIFF
--- a/src/query/ee/src/storages/fuse/io/snapshots.rs
+++ b/src/query/ee/src/storages/fuse/io/snapshots.rs
@@ -49,7 +49,7 @@ where
     }
 
     // 1. Get all the snapshot by chunks, save all the segments location.
-    let max_io_requests = ctx.get_settings().get_max_storage_io_requests()? as usize;
+    let max_io_requests = ctx.get_settings().get_max_threads()? as usize;
 
     let start = Instant::now();
     let mut count = 1;
--- a/src/query/service/tests/it/storages/fuse/operations/mutation/recluster_mutator.rs
+++ b/src/query/service/tests/it/storages/fuse/operations/mutation/recluster_mutator.rs
@@ -388,11 +388,10 @@ async fn generage_segments(
     }
 
     let threads_nums = ctx.get_settings().get_max_threads()? as usize;
-    let permit_nums = ctx.get_settings().get_max_storage_io_requests()? as usize;
     execute_futures_in_parallel(
         tasks,
         threads_nums,
-        permit_nums,
+        threads_nums,
         "fuse-write-segments-worker".to_owned(),
     )
     .await?
--- a/src/query/storages/fuse/src/io/segments.rs
+++ b/src/query/storages/fuse/src/io/segments.rs
@@ -106,11 +106,10 @@ impl SegmentsIO {
         });
 
         let threads_nums = self.ctx.get_settings().get_max_threads()? as usize;
-        let permit_nums = self.ctx.get_settings().get_max_storage_io_requests()? as usize;
         execute_futures_in_parallel(
             tasks,
             threads_nums,
-            permit_nums,
+            threads_nums,
             "fuse-req-segments-worker".to_owned(),
         )
         .await
@@ -143,11 +142,11 @@ impl SegmentsIO {
         });
 
         let threads_nums = self.ctx.get_settings().get_max_threads()? as usize;
-        let permit_nums = self.ctx.get_settings().get_max_storage_io_requests()? as usize;
+
         execute_futures_in_parallel(
             tasks,
             threads_nums,
-            permit_nums,
+            threads_nums,
             "write-segments-worker".to_owned(),
         )
         .await?
--- a/src/query/storages/fuse/src/io/snapshots.rs
+++ b/src/query/storages/fuse/src/io/snapshots.rs
@@ -141,11 +141,11 @@ impl SnapshotsIO {
         });
 
         let threads_nums = self.ctx.get_settings().get_max_threads()? as usize;
-        let permit_nums = self.ctx.get_settings().get_max_storage_io_requests()? as usize;
+
         execute_futures_in_parallel(
             tasks,
             threads_nums,
-            permit_nums,
+            threads_nums,
             "fuse-req-snapshots-worker".to_owned(),
         )
         .await
@@ -173,7 +173,7 @@ impl SnapshotsIO {
         }
 
         // 1. Get all the snapshot by chunks.
-        let max_io_requests = ctx.get_settings().get_max_storage_io_requests()? as usize;
+        let max_io_requests = ctx.get_settings().get_max_threads()? as usize;
         let mut snapshot_lites = Vec::with_capacity(snapshot_files.len());
 
         let start = Instant::now();
@@ -311,11 +311,11 @@ impl SnapshotsIO {
         });
 
         let threads_nums = self.ctx.get_settings().get_max_threads()? as usize;
-        let permit_nums = self.ctx.get_settings().get_max_storage_io_requests()? as usize;
+
         execute_futures_in_parallel(
             tasks,
             threads_nums,
-            permit_nums,
+            threads_nums,
             "fuse-req-snapshots-worker".to_owned(),
         )
         .await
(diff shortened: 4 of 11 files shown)


JackTan25:
use thread_nums as permit_nums to control the concurrency. Do we need to use try_join_all directly without permits? or use thread_nums * 4 as the permit_nums?

src/query/ee/src/storages/fuse/io/snapshots.rs
```diff
@@ -49,7 +49,7 @@ where
     }
 
     // 1. Get all the snapshot by chunks, save all the segments location.
-    let max_io_requests = ctx.get_settings().get_max_storage_io_requests()? as usize;
+    let max_io_requests = ctx.get_settings().get_max_threads()? as usize;
```
BohuTANG, line 52: The name `max_io_requests` is not good :)

src/query/service/tests/it/storages/fuse/operations/mutation/recluster_mutator.rs
```diff
@@ -388,11 +388,10 @@ async fn generage_segments(
     }
 
     let threads_nums = ctx.get_settings().get_max_threads()? as usize;
-    let permit_nums = ctx.get_settings().get_max_storage_io_requests()? as usize;
     execute_futures_in_parallel(
```
BohuTANG, line 391: Here no need to use `execute_futures_in_parallel`?

BohuTANG: Ping @dantengsky  for better suggestion

dantengsky: yeah, @JackTan25 noticed that exceute_futures_in_parallel could be replaced with `try_join_all`,

but need to confirm further if in all the cases `permit_nums` eq `threads_nums` is reasonable.

BohuTANG: We can wrapper a new method for it? Like:
```
pub async fn execute_futures_parallel<Fut>(
    futures: impl IntoIterator<Item = Fut>,
    thread_nums: usize,
    thread_name: String,
) -> Result<Vec<Fut::Output>>
```

And change the old one to:
```
pub async fn execute_futures_parallel_with_permits<Fut>(
    futures: impl IntoIterator<Item = Fut>,
    thread_nums: usize,
    permit_nums: usize,
    thread_name: String,
) -> Result<Vec<Fut::Output>>
```

dantengsky: @zhyass Any concerns?  shall we let the `permit_nums` eq `scale_factor * threads_nums`, where `scale_factor = 4`, in this case (reclustering/compaction) for example?

src/query/storages/fuse/src/io/segments.rs
```diff
@@ -143,11 +142,11 @@ impl SegmentsIO {
         });
 
         let threads_nums = self.ctx.get_settings().get_max_threads()? as usize;
-        let permit_nums = self.ctx.get_settings().get_max_storage_io_requests()? as usize;
+
```
dantengsky, line 145: `SegmentsIO::write_segments` is no longer used, would you please remove this method?

zhang2014:
maybe we should add new settings is better, with default values same to `max_threads`

src/query/storages/fuse/src/io/segments.rs
JackTan25: yes

JackTan25:
> maybe we should add new settings is better, with default values same to `max_threads`

I'm not sure whether we need the factor or just use try_join_all()

zhang2014:
> > maybe we should add new settings is better, with default values same to `max_threads`
> 
> I'm not sure whether we need the factor * thread_nums or just use try_join_all()

it may be necessary when we need to adjust its parallel？

src/query/storages/fuse/src/io/snapshots.rs
```diff
@@ -311,12 +311,12 @@ impl SnapshotsIO {
...
         let threads_nums = self.ctx.get_settings().get_max_threads()? as usize;
-        let permit_nums = self.ctx.get_settings().get_max_storage_io_requests()? as usize;
+
         execute_futures_in_parallel(
             tasks,
             threads_nums,
-            permit_nums,
-            "fuse-req-snapshots-worker".to_owned(),
+            threads_nums,
+            "fuse-req-snsrc/query/storages/fuse/src/io/snapshots.rsmapshots-worker".to_owned(),
```
dantengsky, line 319: typo?

BohuTANG:
> maybe we should add new settings is better, with default values same to `max_threads`

Looks we don't need the new setting, only adjust a factor with CPU is fine.

src/query/service/tests/it/storages/fuse/operations/mutation/recluster_mutator.rs
zhyass: > @zhyass Any concerns? shall we let the `permit_nums` eq `scale_factor * threads_nums`, where `scale_factor = 4`, in this case (reclustering/compaction) for example?

Set `scale_factor * threads_nums`, if use `execute_futures_in_parallel`. Otherwise use `threads_nums` directly.

dantengsky:
ci-benchmark

https://github.com/datafuselabs/databend/pull/12997#issuecomment-1733013292

shows no performance degradations,  which is expected, since all the modifications are table mutation-related only (If I get it right).

but scenarios the ci-benchmark not covered, e.g. deletions/compactions/replace-into, etc, we do not know the impacts yet.

Maybe too conservative, but I am with @zhang2014 ,  that we may need to introduce a new setting(let's say, `max_mutation_io_request` or a better name), with a default value that equals the values of `max_stroage_io_request`, but can be tweaked individually.

JackTan25:
> ci-benchmark
> 
> [#12997 (comment)](https://github.com/datafuselabs/databend/pull/12997#issuecomment-1733013292)
> 
> shows no performance degradations, which is expected, since all the modifications are table mutation-related only (If I get it right).
> 
> but scenarios the ci-benchmark not covered, e.g. deletions/compactions/replace-into, etc, we do not know the impacts yet.
> 
> Maybe too conservative, but I am with @zhang2014 , that we may need to introduce a new setting(let's say, `max_mutation_io_request` or a better name), with a default value that equals the values of `max_stroage_io_request`, but can be tweaked individually.

I agree with that. When we set max_stroage_io_request, we will also set max_mutation_io_request, but we can also set max_mutation_io_request individually. And no need factor, just use max_mutation_io_request as the permit_nums in execute_futures_in_parallel. cc @dantengsky @zhang2014  @BohuTANG  @zhyass  can we be consistent with this?

BohuTANG:
> > ci-benchmark
> > [#12997 (comment)](https://github.com/datafuselabs/databend/pull/12997#issuecomment-1733013292)
> > shows no performance degradations, which is expected, since all the modifications are table mutation-related only (If I get it right).
> > but scenarios the ci-benchmark not covered, e.g. deletions/compactions/replace-into, etc, we do not know the impacts yet.
> > Maybe too conservative, but I am with @zhang2014 , that we may need to introduce a new setting(let's say, `max_mutation_io_request` or a better name), with a default value that equals the values of `max_stroage_io_request`, but can be tweaked individually.
> 
> I agree with that. When we set max_stroage_io_request, we will also set max_mutation_io_request, but we can also set max_mutation_io_request individually. And no need factor, just use max_mutation_io_request as the permit_nums in execute_futures_in_parallel. cc @dantengsky @zhang2014 @BohuTANG @zhyass can we be consistent with this?

Hmm...
This means that we do not do the `CPU*fatctor` right, depends another settings `max_stroage_io_request`.
This PR aims to make less settings.

JackTan25:
> > @zhyass Any concerns? shall we let the `permit_nums` eq `scale_factor * threads_nums`, where `scale_factor = 4`, in this case (reclustering/compaction) for example?
> 
> Set `scale_factor * threads_nums`, if use `execute_futures_in_parallel`. Otherwise use `threads_nums` directly.
use @zhyass 's suggestion.

dantengsky (approved):

PR #12999, opened 2023-09-25
chore(query): cache expr

### LINKED ISSUE #12970: Feature: make `Evaluator` support cache run result for same expr
**Summary**

Description for this feature.

```
pub struct Evaluator<'a> {
    input_columns: &'a DataBlock,
    func_ctx: &'a FunctionContext,
    fn_registry: &'a FunctionRegistry,

    results: HashMap<Expr, Value<AnyType>>
}
```
And reuse the same evaluator in:
https://github.com/datafuselabs/databend/blob/749581715f0801ead08cecda74e9b5e063e66abd/src/query/sql/src/evaluator/block_operator.rs#L94-L98

ariesdevil:

## Summary

Add a local cache to Evaluator, used only in `block_operator.rs`

- Closes #12970 

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12999)


src/query/expression/src/evaluator.rs
```diff
@@ -60,6 +62,8 @@ pub struct Evaluator<'a> {
     input_columns: &'a DataBlock,
     func_ctx: &'a FunctionContext,
     fn_registry: &'a FunctionRegistry,
+
+    cached_values: Mutex<HashMap<Expr, Value<AnyType>>>,
```
sundy-li, line 66: better and a flag to use the `cached_values`.

cache_enable: bool

And only enable it in `block_operator`

src/query/expression/src/evaluator.rs
```diff
@@ -103,6 +145,13 @@ impl<'a> Evaluator<'a> {
         #[cfg(debug_assertions)]
         self.check_expr(expr);
 
+        // try get result from cache
+        if let Some(_) = self.cached_values {
```
sundy-li, line 149: I believe lint will be unhappy.

self.cached_values.map(...

src/query/expression/src/evaluator.rs
```diff
@@ -89,6 +104,33 @@ impl<'a> Evaluator<'a> {
...
     }
 
+    fn cache_expr_value(&self, expr: &Expr, value: Value<AnyType>) {
+        let mut cached_values_ref = self.cached_values.as_ref().unwrap().write();
+        if let Entry::Vacant(v) = cached_values_ref.entry(expr.clone()) {
+            v.insert(value);
+        }
+    }
+
+    fn get_cached_expr_value(&self, expr: &Expr) -> Option<Value<AnyType>> {
```
sundy-li, line 114: toooo complex, this function is read overhead

sundy-li: if let Some(cache_values) = self.cached_values
                .as_ref() {
                
         if let Entry::Occupied(v) = cached_values_ref.entry(expr.clone()) {
             return Some(v);
        }
   }

src/query/expression/src/evaluator.rs
```diff
@@ -103,6 +118,18 @@ impl<'a> Evaluator<'a> {
         #[cfg(debug_assertions)]
         self.check_expr(expr);
 
+        // try get result from cache
+        if let Some(cached_values) = &self.cached_values {
+            let found = cached_values.read().contains_key(expr);
```
BohuTANG, line 123: Why not use `cached_values.read().get` directly?

ariesdevil: I also found this case, in the previous logical if check false then perform a write to the cache. Fixed.

src/query/expression/src/evaluator.rs
```diff
@@ -204,6 +226,15 @@ impl<'a> Evaluator<'a> {
...
             }
         }
+        if let Some(cached_values) = &self.cached_values {
+            if let Ok(r) = &result {
+                let mut cached_values_ref = cached_values.write();
+                if let Entry::Vacant(v) = cached_values_ref.entry(expr.clone()) {
+                    v.insert(r.clone());
+                }
+            }
+        }
```
BohuTANG, line 236: GPT4:
```suggestion
if let Some(cached_values) = &self.cached_values {
    if let Ok(r) = &result {
        // Perform any necessary here before acquiring the write lock.
        let expr_cloned = expr.clone();
        let result_cloned = r.clone();
        
        // Acquire the write lock and insert the value.
        let mut cached_values_ref = cached_values.write();
        if let Entry::Vacant(v) = cached_values_ref.entry(expr_cloned) {
            v.insert(result_cloned);
        }
    }
}
```

This example performs cloning before acquiring the write lock, potentially reducing the duration for which the lock is held.

ariesdevil: Thx gpt4 :)

BohuTANG:
Can we measured how much improved?

ariesdevil:
> Can we measured how much improved?↳

I'll test tpch sf1 q1 on my local machine

ariesdevil:
tpch sf10 q1 3.592 -> 3.296 on my local machine

sundy-li (approved):

BohuTANG:
> tpch sf10 q1 3.683 -> 3.296 on my local machine

Is this as expected? Based on my test it almost perform same as the main branch. Maybe we should do flamegraph to analyze.

sundy-li:
@ariesdevil  @BohuTANG 

I'm really sorry to create this issue (I almost forgot it).

cache expr is already covered by `apply_cse`,  so this pr is no need anymore.

PR #13020, opened 2023-09-26
refactor: introduce `ReadPolicy` to parquet reader.

RinChanNOWWW:

## Summary

There is a drawback of using `arrow-rs` APIs: when we try to prefetching data for prewhere and topk push downs, we cannot reuse the deserialized blocks. That is, we need to deserialize the remain data again even if we have already deserialized it at prefetch stage. It quite wasteful if the many output columns are contained in prefetched columns.

This PR's target is to reuse the prefetched data blocks at the final output stage. However, the codes now are highly coupled (the logic of prewhere, topk, nested type and small files reading are all together). So this PR refactor a lot to decouple the logic.

### Split Row Group Reader and Full File Reader

This PR split `ParquetRSReader` into `ParquetRSRowGroupReader` and `ParquetRSFullReader`. The first reader is only used to read parquet data by row group level, while the second reader is only used to read a whole parquet file.

I put them into a new mod `reader`.

To reduce duplicate computation, these two readers can be built from a same `ParquetRSReaderBuilder` (by `build_row_group_reader` and `build_full_reader`). They can reuse some intermediate variables.

### Introduce `ReadPolicy`

I add a new mod `read_policy`.

Due to prewhere, topk pushdowns and the introduction of `omit_filter` in #12934. Row group reading become very complicated.
This PR introduce `ReadPolicy` to decouple the logic by runtime dynamic polymorphism (`type ReadPolicyImpl = Box<dyn ReadPolicy>`).

I introduced 4 kinds of policy:

```rust
pub const POLICY_NO_PREFETCH: PolicyType = 0; // 0b00
pub const POLICY_PREDICATE_ONLY: PolicyType = 1; // 0b01
pub const POLICY_TOPK_ONLY: PolicyType = 2; // 0b10
pub const POLICY_PREDICATE_AND_TOPK: PolicyType = 3; // 0b11
```

Each policy is responded to one `ReadPolicyImpl` (`POLICY_PREDICATE_ONLY` and `POLICY_PREDICATE_AND_TOPK` can use a same `ReadPolicyImpl`).

In the case of `omit_filter`, we can use simple bit-wise operation (`& !POLICY_PREDICATE_ONLY`) to convert `POLICY_PREDICATE_ONLY` to `POLICY_NO_PREFETCH`, `POLICY_PREDICATE_AND_TOPK` to `POLICY_TOPK_ONLY`.

`ParquetRSRowGroupReader` uses a two-stage way to read row group:

1. Build `ReadPolicyImpl` by `ReadPolicyBuilder` (which is also a trait).
2. Call `read_block` of `ReadPolicyImpl` to produce `DataBlock`.

When creating `ParquetRSRowGroupReader`, it will register all the policy builders it needs:

```rust
match (prewhere_predicate, topk) {
    (Some(_), Some(_)) => {
        // register POLICY_PREDICATE_AND_TOPK
        // register POLICY_TOPK_ONLY (predicate may be omitted)
    }
    (Some(_), None) => {
        // register POLICY_PREDICATE_ONLY
        // register POLICY_NO_PREFETCH (predicate may be omitted)
    }
    (None, Some(_)) => {
        // register POLICY_TOPK_ONLY
    }
    (None, None) => {
        // register POLICY_NO_PREFETCH
    }
};
```

Available `ReadPolicyImpl`s as follow:

#### NoPrefetchPoliy

Used for `POLICY_NOPREFETCH`.

There is no prefetch stage, we can read, deserialize and output data block we need directly.

#### PredicateAndTopkPolicy

Used for `POLICY_PREDICATE_ONLY` and `POLICY_PREDICATE_AND_TOPK`.s

At prefetch stage, prefetch columns needed by prewhere and topk. Deserialize them into `DataBlock` and evaluate it into `RowSelection`. Slice the `DataBlock` by batch size and store the `VecDeque` in memory.

At final stage, read remain columns by `RowSelection` and output `DataBlock`s by batch size, and merge the prefetch block and project the block by `output_schema`.

#### TopkOnlyPolicy

Used for `POLICY_TOPK_ONLY`.

It's similar to `PredicateANdTopkPolicy`, but we only evaluate topk at prefech stage.

### Others

- This PR add `table_statistics` for `parquet_rs` and `parquet2` tables (only `num_rows` is available`).
- Delete some unused variables.
- This PR add sqllogictests for every policy.

### To be improved

- It's hard to reuse inner columns of nested types now. We can improve it in later PRs. One reasonable solution:  store all most inner `Column`s and combine them into `DataBlock` when outputing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13020)


BohuTANG:
This PR summary is so clear and detail, fantastic! 👍

src/query/storages/parquet/src/parquet_rs/parquet_reader/reader/full_reader.rs
```diff
@@ -0,0 +1,199 @@
...
+use parquet::arrow::ProjectionMask;
+
+use crate::parquet_rs::parquet_reader::predicate::ParquetPredicate;
+use crate::parquet_rs::parquet_reader::utils::transform_record_batch;
+use crate::parquet_rs::parquet_reader::utils::transform_record_batch_by_field_paths;
+use crate::parquet_rs::parquet_reader::utils::FieldPaths;
+use crate::ParquetRSPruner;
+
+/// The reader to read a whole parquet file.
+pub struct ParquetFSFullReader {
```
b41sh, line 41: is `ParquetFS` a typo?

b41sh, line 41: is `ParquetFS` a typo?

src/query/storages/parquet/src/parquet_rs/parquet_reader/read_policy/predicate_and_topk.rs
```diff
@@ -0,0 +1,295 @@
...
+                    *selection = selection.and_then(&sel);
+                }
+                None => {
+                    selection = Some(sel);
+                }
+            }
+        }
+
+        // Evaluate topk
+        if let Some(((topk, offset), sorter)) = self.topk.as_ref().zip(sorter.as_mut()) {
```
b41sh, line 194: should we evaluate topk before predicate? topk may filter more rows.

src/query/storages/parquet/src/parquet_rs/parquet_reader/read_policy/topk_only.rs
```diff
@@ -0,0 +1,216 @@
...
+            Some(selection) => {
+                *selection = selection.and_then(&sel);
+            }
+            None => {
+                selection = Some(sel);
+            }
+        }
+
+        // Slice the prefetched block by `batch_size`.
+        num_rows = prefetched.num_rows();
```
b41sh, line 154: This code looks the same as the implementation in `PredicateAndTopkPolicyBuilder`, can we write them in a single function to reduce duplication of code?
https://github.com/datafuselabs/databend/pull/13020/files#diff-11b6a41c2c804bda82298bb8115e50af179135d64aa13384f17ff6ffb05403ceR230-R252

src/query/storages/parquet/src/parquet_rs/parquet_reader/reader/full_reader.rs
RinChanNOWWW: Yes. Thank you.

sundy-li (approved):

PR #13021, opened 2023-09-26
feat: Enable OpenDAL PrometheusClientLayer

Xuanwo:

## Summary

feat: Enable OpenDAL PrometheusClientLayer

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13021)


flaneur2020 (approved):

src/common/storage/src/operator.rs
```diff
@@ -108,7 +112,8 @@ pub fn build_operator<B: Builder>(builder: B) -> Result<Operator> {
         .layer(LoggingLayer::default())
         // Add tracing
         .layer(MinitraceLayer)
-        // TODO(liyz): add PrometheusClientLayer
+        // Add PrometheusClientLayer
+        .layer(PrometheusClientLayer::new(&mut registry))
```
flaneur2020, line 116: how about move the registry inside a code block here? 

```
.layer({
  let mut registry = load_global_prometheus_registry();
  PrometheusClientLayer::new(&mut registry);
})
```

the `registry` is actually a mutex guard, thus we can reduce the scope of the mutex using the code block.

Xuanwo: Nice!

Issue #13025, opened 2023-09-26
Feature: function factory should support return err code


TCeason:
**Summary**

Better to return Invalid Argument.

```sql
MySQL [(none)]> select factorial(21);
ERROR 1105 (HY000): Internal. Code: 1001, Text = error: 
  --> SQL:1:8
  |
1 | select factorial(21)
  |        ^^^^^^^^^^^^^ factorial number is out of range, max is: 20 while evaluating function `factorial(21)`

```

cc @andylokandy @b41sh

PR #13028, opened 2023-09-26
fix: change evaluation error code to BadArgument

andylokandy:

## Summary

Summary about this PR

- Closes https://github.com/datafuselabs/databend/issues/13025

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13028)


src/query/expression/src/function.rs
```diff
@@ -591,7 +591,7 @@ impl<'a> EvalContext<'a> {
                     )
                 };
 
-                Err(ErrorCode::Internal(err_msg).set_span(span))
```
TCeason, line 594: Can we customize error code? default is BadArguments

andylokandy: It requires some major refactor on `FunctionContext`. In what case custom error code is necessary?

b41sh: return `ErrorCode::BadArguments` is okay, as most errors in function are caused by bad arguments.

TCeason (approved):

PR #13030, opened 2023-09-26
fix: Fix alias of target

JackTan25:

## Summary
when give a alias to target_table, we can't match_column_binding, because the database of column_binding is always none, so if there is alias, when we do generate_update_list, we need to set match_column_binding's database parameter is None.

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13030)


dantengsky:
askbend:summary

dantengsky (approved):

PR #13060, opened 2023-09-27
refactor: refactor decompress and add compress utils.

youngsofun:

## Summary

Summary about this PR

1.  refactor decompress codes
2. fix a potential bug in decompress (not always flush data in decoder), enforce and refactor the unit tests ( seems not lead to real bug in production, otherwise will lead to incomplete row and return error,  may due to the way it is used)
3. add codes for compress, to be used for unload(copy into location),  less wrapping (compared with decompress) seems enough for now 
   4. the async-compress is patch again to publish more api  https://github.com/youngsofun/async-compression/tree/public_encoder

todo next pr:

- impl unload with decompress, need to max compressing parallelism, at the same time respect max_file_size

to do later.

- the async-compress lib and our wrapper is overqualified and too complicated  for current usage,  need to be simplified later

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13060)


Xuanwo (approved):
Nice!

src/common/compress/src/decode.rs
```diff
@@ -409,9 +340,56 @@ impl<R: AsyncRead> AsyncRead for DecompressReader<R> {
     }
 }
 
+impl DecompressDecoder {
+    pub fn decompress_all(&mut self, compressed: &[u8]) -> common_exception::Result<Vec<u8>> {
+        let main = self.decompress_batch(compressed)?;
+        let tail = self.decompress_batch(&[])?;
+        if tail.is_empty() {
```
sundy-li, line 347: seems the if condition could be omitted.

src/common/compress/src/decode.rs
```diff
@@ -409,9 +340,56 @@ impl<R: AsyncRead> AsyncRead for DecompressReader<R> {
...
+            Ok(main)
+        } else {
+            let mut all = main;
+            all.extend_from_slice(&tail);
+            Ok(all)
+        }
+    }
+    // need to finish the decoding by adding a empty input
+    pub fn decompress_batch(&mut self, compressed: &[u8]) -> common_exception::Result<Vec<u8>> {
```
sundy-li, line 356: why not expose the writer in this api:
```
decompress_batch(&mut self, compressed: &[u8],  buffer: &mut Vec<u8> )
```

sundy-li: with a Guard to truncate the size if an error happens.

youngsofun: buf is local, no need for guard?

youngsofun: > why not expose the writer in this api:
> 
> ```
> decompress_batch(&mut self, compressed: &[u8],  buffer: &mut Vec<u8> )
> ```

decompress_all is only used for tests;
this pr was originally for add compress utils, when testing it, found the bug for decompress, so let`s fix it first.
we can lever the improvement for api and performance to another issue.

PR #13062, opened 2023-09-27
refactor(http): refactor the unset processing logic in UnsetInterpreter

flaneur2020:

## Summary

this pr is a continued rework of https://github.com/datafuselabs/databend/pull/12978 .

as talked with @youngsofun, the current logic about UNSET statement is:

1. read the default values and set the default values back
2. record the SET default settings into a QueryAffect
3. replay the QueryAffect in the http handler before return, produced the new session settings list.
4. reply the client as the **session settings** to inform the client send the next request with this updated session settings, and not need to care about the returned query affect at all.

this PR try make a refactor about this:

1. just UNSET the local settings of the current session in the UNSET interpreter
2. reply the client with the updated session

to archive this, we have to retrieve the updated session object after ExecutorStopped.

this PR did not change the existed QueryAffect behaviour.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13062)


src/query/service/src/servers/http/v1/query/execute_state.rs
```diff
@@ -112,6 +112,7 @@ pub struct ExecuteStopped {
     pub affect: Option<QueryAffect>,
     pub reason: Result<()>,
     pub stop_time: Instant,
+    pub session: Arc<Session>,
```
youngsofun, line 115: not very sure it is ok,  not releasing session as fast as possible may lead to unexpected behavior.

youngsofun: or even problems for shutdown? cc @zhang2014

flaneur2020: I suppose this might not extend the lifetime much longer. The session object will be recycled when the ref count reaches zero, and it's still very fast to release this ref count on the response.

youngsofun:
```
pub struct HttpQuery {
    pub(crate) id: String,
    pub(crate) session_id: String,
    request: HttpQueryRequest,
    state: Arc<RwLock<Executor>>,
```

Executor not released until HttpQuery is cleared by final or timeout

flaneur2020:
introduced ExecutorSessionState to store the session's final state thus we can release the Arc<Session> earlier on stopped, @youngsofun PTAL

youngsofun (approved):

src/query/service/tests/it/servers/http/http_query_handlers.rs
```diff
@@ -1200,19 +1200,35 @@ async fn test_affect() -> Result<()> {
...
                     ("max_threads".to_string(), "1".to_string()),
                     ("timezone".to_string(), "Asia/Shanghai".to_string()),
                 ])),
             }),
         ),
+        (
+            serde_json::json!({"sql": "unset timezone", "session": {"settings": {"max_threads": "6", "timezone": "Asia/Shanghai"}}}),
+            Some(QueryAffect::ChangeSettings {
+                keys: vec!["timezone".to_string()],
+                values: vec!["UTC".to_string()], /* TODO(liyz): consider to return the complete settings after set or unset */
```
youngsofun, line 1215: >>> consider to return the complete settings after set or unset

that will exactly the same as `HttpSessionConf.settings?`

youngsofun: we can check if we really need `QueryAffect::ChangeSettings` for set/unset.
if not, just remove it.
otherwise,  it should be the exact effect of the current set/unset SQL.
so  better add a "is_unset" flag. 
for unset, `values` can still be retained. client can choose to just use `is_unset` or the value.

flaneur2020: IMO this `ChangeSettings` is not quite functional as it expected, we can remove it gradually.

BohuTANG:
Conflicting files
src/query/service/src/servers/http/v1/query/execute_state.rs
src/query/service/src/servers/http/v1/query/http_query.rs

youngsofun (approved):

flaneur2020:
one failed stateless case:

```
--- /runner/_work/databend/databend/tests/suites/1_stateful/00_stage/00_0011_distributed_copy_into_table_exection_test.result	2023-11-03 04:35:21.272044130 +0000
+++ /runner/_work/databend/databend/tests/suites/1_stateful/00_stage/00_0011_distributed_copy_into_table_exection_test.stdout	2023-11-03 04:36:17.352007313 +0000
@@ -29,6 +29,7 @@
 24
 null
 Error: APIError: ResponseError with 1046: Number of columns in file (2) does not match that of the corresponding table (3)
+at file 'sample_2_columns.csv', line 0
 24
 5
 54
```

IMHO this failure may be not corronspond with this PR. I suppose that this PR is ready to be merged.

#13537 needs a follow up in this PR with the ability to return the current role after query executed.

PR #13070, opened 2023-09-28
chore(planner): fix push down filter evalscalar

Dousir9:

### Summary
If a predicate does not refer to the `items` in `EvalScalar`, we can push it down directly.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13070)


Dousir9:
Already fixed in #13071

PR #13078, opened 2023-10-02
feat(planner): support infer col ref equal predicate

Dousir9:

## Summary

Support inferring column ref predicate: `[A = B and A = C] => [B = C]`, for example: `select * from t1, t2, t3 where t1.A = t2.B and t1.A = t3.C`, we will derive new predicate `t2.B = t3.C`, but we need to prune out redundant predicates.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13078)


Dousir9:
Achieved by https://github.com/datafuselabs/databend/pull/14338.

PR #13089, opened 2023-10-06
feat(query): support DATABEND_DATA_PATH in bendpy and bend-local

### LINKED ISSUE #12992: Python Binding: support custom path to store data and metadata
DuckDB, Hyper and Glaredb all support custom path to storage data, AFAIK databend by default use a hidden local folder, can we have the option to have a custom path

see example here

image

https://colab.research.google.com/drive/11Jz3G1oqf8glye5BmNIq5cOdeUQJjImv#scrollTo=MDTt42NXEL5_

sundy-li:

## Summary

Summary about this PR

 `bend-local` now respect env `DATABEND_DATA_PATH`, if it's set, meta and data will be persistent in that path.
 
 ```bash
❯ DATABEND_DATA_PATH=/tmp/abc/ ./target/debug/databend-query local -q "create table abc(a int); insert into abc values(3);"
❯ DATABEND_DATA_PATH=/tmp/abc/ ./target/debug/databend-query local -q "select * from abc"
3
 ```
 
 
 `bendpy` will also respect this env, but you must set the env before import databend 
 
 ```python
 
import os
os.environ["DATABEND_DATA_PATH"] = "/tmp/def/"

from databend import SessionContext

...
 ```
 
 

- Closes #12992

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13089)

### DIFF
--- a/src/bendpy/src/lib.rs
+++ b/src/bendpy/src/lib.rs
@@ -21,6 +21,7 @@ mod schema;
 mod utils;
 
 use std::env;
+use std::path::Path;
 
 use common_config::Config;
 use common_config::InnerConfig;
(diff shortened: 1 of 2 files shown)


BohuTANG (approved):
LGTM

PR #13097, opened 2023-10-07
refactor: meta-service: re-implement export with Stream

drmingdrmer:

##### refactor: meta-service: re-implement export with Stream

- Updated the `export()` function in `grpc_service.rs` to use crate
  `futures-async-stream`, changed return type of the function to use a
  `BoxStream` that yields a series of JSON strings, instead of returning
  a big `Vec<String>`.

  This allows for more efficient handling of asynchronous streams, which
  can provide performance improvements.

  Note that `export(self: Arc<Sefl>)` takes an `Arc<Self>` as argument
  because the generated `gRPC` stream signature requires a `'static`
  lifetime. As a result, the `export(&self)` can not be used:

  ```
  type ExportStream: Stream<Item = Result<_, _>> + Send + 'static;
  ```

- Add dependency crate `futures-async-stream`;

- Enable unstable feature `#![feature(generators)]` for building a
  `Stream` from a generator(`yield ...`);

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13097)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3768,6 +3768,7 @@ dependencies = [
  "derive_more",
  "env_logger",
  "futures",
+ "futures-async-stream",
  "itertools 0.10.5",
  "lazy_static",
  "log",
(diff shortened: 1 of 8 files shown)


drmingdrmer:
@everpcpc 
This PR could be relevant to the new features that you are working on. Therefore I'd like to invite you to review this PR.

everpcpc (approved):

PR #13104, opened 2023-10-07
refactor: refactor apply_cse by using cached expr

ariesdevil:

## Summary

This PR introduces a new method `run_exprs` and a new field `cached_values` to `Evaluator` to avoid eval the same expression.

After this, we can delete the `apply_cse` to simple codes.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13104)


src/query/sql/src/evaluator/block_operator.rs
```diff
@@ -91,15 +91,14 @@ impl BlockOperator {
...
-                        let evaluator = Evaluator::new(&input, func_ctx, &BUILTIN_FUNCTIONS);
-                        let result = evaluator.run(expr)?;
-                        let col = BlockEntry::new(expr.data_type().clone(), result);
-                        input.add_column(col);
-                    }
+                    // Local variable run in single thread, so `Rc` and `RefCell` are enough.
+                    let evaluator =
+                        Evaluator::new(&input, func_ctx, &BUILTIN_FUNCTIONS).with_cache();
+                    let result = evaluator.run_exprs(exprs)?;
+                    let block = DataBlock::new(result, input.num_rows());
```
sundy-li, line 98: input should add results to build the final block.

ariesdevil:
This PR needs to run benchmark to see if there are any performance regressions

src/query/expression/src/evaluator.rs
```diff
@@ -89,6 +104,25 @@ impl<'a> Evaluator<'a> {
         }
     }
 
+    pub fn run_exprs(&self, exprs: &[Expr]) -> Result<Vec<BlockEntry>> {
+        let mut columns: Vec<BlockEntry> = Vec::with_capacity(exprs.len());
+        for expr in exprs {
+            // try get result from cache
+            if let Some(cached_values) = &self.cached_values {
```
sundy-li, line 111: seems this is already covered in L141

src/query/expression/src/evaluator.rs
```diff
@@ -103,6 +137,13 @@ impl<'a> Evaluator<'a> {
         #[cfg(debug_assertions)]
         self.check_expr(expr);
 
+        // try get result from cache
+        if let Some(cached_values) = &self.cached_values {
```
sundy-li, line 141: if !matches!(expr, Expr::ColumnRef {..} | Expr::Constant{..}) && let Some(cached_values) = &self.cached_values {

src/query/expression/src/evaluator.rs
```diff
@@ -89,6 +104,16 @@ impl<'a> Evaluator<'a> {
         }
     }
 
+    pub fn run_exprs(&self, exprs: &[Expr]) -> Result<Vec<BlockEntry>> {
```
sundy-li, line 107: should return `Result<Vec<Value<AnyType>>>`, keep consistent with `run` function.

sundy-li: seems we don't need this function, better to call `run` function outside.

ariesdevil: thx, fixed

ariesdevil:
The `Expr::Cast` run in different evaluator lead some issue, reproduce sql:
```sql
create table t(a int);
insert into t values(1);
select sum(case when a < 1 then cast(a as decimal(2, 0)) else 0 end) as a_old, sum(case when a >= 1 then cast(a as decimal(2,0)) else 0 end) as a_new from t;
```
output should be `0, 1`, but got `0, null`

sundy-li (approved):

PR #13123, opened 2023-10-08
refactor: using column name instead of column id for agg index schema

ariesdevil:

## Summary

Using column name instead of column id for agg index schema.

Assume we have an agg-index `select sum(a+1), b from t group by b` with table `t(a int, b int)`, after refresh index then do `select * from 'xxx_index.parquet'`

Before:
```
+------+--------------------+
|  2   |          6         |
+------+--------------------+
| NULL |               |
|    1 |               |
+------+--------------------+
```

After this PR:
```
+------+--------------------+
| b    | sum_state((a + 1)) |
+------+--------------------+
| NULL |               |
|    1 |               |
+------+--------------------+
```
- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13123)

### DIFF
--- a/src/query/service/src/interpreters/interpreter_index_refresh.rs
+++ b/src/query/service/src/interpreters/interpreter_index_refresh.rs
@@ -332,7 +332,7 @@ impl Interpreter for RefreshIndexInterpreter {
             .map(|f| {
                 let pos = select_columns
                     .iter()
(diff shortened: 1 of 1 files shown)


ariesdevil:
see #13125

PR #13125, opened 2023-10-08
refactor: using column name instead of column id for agg index schema

ariesdevil:

## Summary

Using column name instead of column id for agg index schema.

Assume we have an agg-index `select sum(a+1), b from t group by b` with table `t(a int, b int)`, after refresh index then do `select * from 'xxx_index.parquet'`

Before:
```
+------+--------------------+
|  2   |          6         |
+------+--------------------+
| NULL |               |
|    1 |               |
+------+--------------------+
```

After this PR:
```
+------+--------------------+
| b    | sum_state((a + 1)) |
+------+--------------------+
| NULL |               |
|    1 |               |
+------+--------------------+
```
- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13125)


src/query/service/src/interpreters/interpreter_index_refresh.rs
```diff
@@ -326,10 +326,23 @@ impl Interpreter for RefreshIndexInterpreter {
...
 
+        let fields = output_schema
+            .fields()
+            .iter()
+            .map(|f| {
+                let pos = select_columns
+                    .iter()
+                    .find(|col| col.index.to_string().eq_ignore_ascii_case(f.name()))
+                    .ok_or_else(|| ErrorCode::Internal("should find the corresponding column"))?;
+                Ok(DataField::new(&pos.column_name, f.data_type().clone()))
```
RinChanNOWWW, line 337: We can create `TableDataField` directly here.

RinChanNOWWW: How about iterate `select_columns` to get the `output_schema`.

RinChanNOWWW (approved):

PR #13127, opened 2023-10-08
refactor: replace once_cell with std

lewiszlw:

## Summary

Replace once_cell with std as much as possible.

P.S. OnceLock dones't provide `wait` method, so still keep once_cell in some places.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13127)


BohuTANG:
Thanks for the contribution.

It's not the right moment to replace `once_cell`. Otherwise, we'll end up intermixing `once_cell` and `OnceLock` usage.

PR #13131, opened 2023-10-08
fix(metrics): avoid leaking meta_client's error messages into metrics

flaneur2020:

## Summary

have a full error message into metrics' labels is not effective to the prometheus storage, and have dangerous for the metrics scraper like this:

```
expected equal, got "expired\"" ("LNAME") while parsing: "databend_meta_grpc_client_request_fail_total{endpoint=\"meta-service-0.meta-service.databend-system.svc:9191\",request=\"PrefixList\",error=\"ConnectionError: source: tonic::status::Status: status: Cancelled, message: \"Timeout expired\""
```

this PR added a `name()` method to MetaError, and put the name into the label instead of `err.to_string()` to reduce the cardinality of the metric labels on error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13131)

### DIFF
--- a/src/meta/client/src/grpc_client.rs
+++ b/src/meta/client/src/grpc_client.rs
@@ -482,7 +482,7 @@ impl MetaGrpcClient {
                     grpc_metrics::incr_meta_grpc_client_request_failed(
                         &current_endpoint,
                         req_name,
-                        &err.to_string(),
+                        err,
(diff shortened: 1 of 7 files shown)


drmingdrmer (approved):
Nice shot

PR #13148, opened 2023-10-09
add loser tree merge

### LINKED ISSUE #11604: Feature: Update sort algorithm using Loser Tree
**Summary**

Currently Databend's sort algorithm using MergSort based on **heap**. The main algorithms of multiple merging include heap sort, winner tree and loser tree. In these three algorithms, each heap adjustment of heap sorting requires comparison with the left and right child nodes, and the comparison times is 2logN, while the comparison times of winner tree and loser tree adjustment are logN. The difference is that the winner tree needs to compare with sibling nodes and update the parent node, while the loser tree only needs to compare with the parent node, and the memory access times are less.

ct20000901:

## Summary

I have changed the heap-based merge sort to a loser tree-based implementation.

- Closes #11604


RinChanNOWWW:
It seems the performance degrades greatly on clickbench.

src/query/pipeline/transforms/src/processors/transforms/transform_sort_merge.rs
```diff
@@ -89,6 +91,52 @@ where
             _r: PhantomData,
         })
     }
+
+    fn adjust(
```
RinChanNOWWW, line 95: It's better to abstract a struct for the loser tree.

ct20000901: OK, I will implement this tonight and, in the process, conduct some testing for the 'order by' feature to identify any areas that may require optimization in the implementation.

JackTan25:
good deal!!

src/query/pipeline/transforms/src/processors/transforms/transform_sort_merge.rs
RinChanNOWWW: We also need unit tests for the loser tree.

RinChanNOWWW (changes requested):
`transform_sort_merge.rs` only used for partial sort (one pipeline), we merge sort all the pipelines in `transform_multi_sort_merge.rs`.

In `transform_sort_merge.rs`, the sorting is not a **"multiple-way merge sort"** so the loser tree almost has no benefit.
We should try to introduce the loser tree for `transform_multi_sort_merge.rs`.

ct20000901:
> `transform_sort_merge.rs` only used for partial sort (one pipeline), we merge sort all the pipelines in `transform_multi_sort_merge.rs`.
> 
> In `transform_sort_merge.rs`, the sorting is not a **"multiple-way merge sort"** so the loser tree almost has no benefit. We should try to introduce the loser tree for `transform_multi_sort_merge.rs`.

Yes, I was initially concerned that the loser tree would require blocking until all the data arrives for sorting, so I didn't modify transform_multi_sort_merge.rs. However, upon further reflection, I realized that my previous concerns were unnecessary. I will work on modifying this part in the coming days.

PR #13152, opened 2023-10-09
feat: generate sync agg indexes after copy into

ariesdevil:

## Summary

* Generates sync agg index after `copy into`.
* Clean code for `hook_refresh_agg_index`.
* Add associated tests.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13152)

### DIFF
--- a/src/query/ee/tests/it/aggregating_index/index_refresh.rs
+++ b/src/query/ee/tests/it/aggregating_index/index_refresh.rs
@@ -319,7 +319,6 @@ async fn test_sync_agg_index_after_insert() -> Result<()> {
         .set_enable_refresh_aggregating_index_after_write(true)?;
     let fixture = TestFixture::new_with_ctx(_guard, ctx).await;
     let ctx = fixture.ctx();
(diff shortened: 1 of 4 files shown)


src/query/ee/tests/it/aggregating_index/index_refresh.rs
```diff
@@ -305,26 +305,39 @@ async fn test_refresh_agg_index_with_limit() -> Result<()> {
...
     .await?;
 
     // Create agg index `index0`
     let index_name = "index0";
 
+    execute_sql(
+        ctx.clone(),
+        &format!("DROP AGGREGATING INDEX IF EXISTS {}", index_name),
+    )
+    .await?;
```
RinChanNOWWW, line 337: Do we need to drop the  table and the index? I think `create_ee_query_context` will create a brand new environment.

RinChanNOWWW (approved):

src/query/ee/tests/it/aggregating_index/index_refresh.rs
ariesdevil: fixed

Issue #11270, opened 2023-04-28
Tracking: JSON functions and operators


b41sh:
**Summary**

JSON Functions:
- [x] json_array
- [x] json_array_elements
- [x] json_each
- [x] json_typeof
- [x] json_strip_nulls
- [x] json_pretty
- [x] json_path_exists
- [x] json_path_match

JSON Operators:
- [x] `->` Extracts JSON Array or Object with index or key, returning a JSON value.
- [x] `->>` Extracts JSON Array or Object with index or key, returning a string.
- [x] `#>` Extracts JSON Array or Object with path, returning a JSON value.
- [x] `#>>` Extracts JSON Array or Object with path, returning a string.
- [x] `@>` Does the first JSON value contain the second?
- [x] `<@` Is the first JSON value contained in the second?
- [x] `?` Does the key string exist within the JSON Object key?
- [x] `?|` Does any of the keys in the array exist as JSON Object keys?
- [x] `?&` Does all the keys in the array exist as JSON Object keys?
- [x] `@?` Does the JSON path return any item for the JSON?
- [x] `@@` Returns the result of a JSON path predicate check for the specified JSON value.
- [x] `||` Concatenates two JSON values.
- [x] `- text` Deletes a key (and its value) from a JSON object, or matching string value(s) from a JSON array.
- [x] `- integer` Deletes the array element with specified index (negative integers count from the end). Throws an error if JSON value is not an array.
- [x] `#- text[]` Deletes the field or array element at the specified path, where path elements can be either field keys or array indexes.

Refer to the [PostgreSQL JSON functions document](https://www.postgresql.org/docs/16/functions-json.html)

ariesdevil:
@byrantwithyou You can start by picking a few implementations instead of all of them.

byrantwithyou:
I'll focus on the bitmap first.

byrantwithyou:
I'm so sorry that I couldn't do this task because I'm a little bit busy these days. Please assign it to anyone else.

b41sh:
> I'm so sorry that I couldn't do this task because I'm a little bit busy these days. Please assign it to anyone else.

@byrantwithyou It's ok, feel free to come back and contribute when you have time.

infdahai:
I'd like to write some works in a few days.

akoshchiy:
[...]

PR #13154, opened 2023-10-09
feat: add json_each & json_array_elements

akoshchiy:

## Summary

Added `json_each` and `json_array_elements`

- Related #11270

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13154)


src/query/functions/src/srfs/mod.rs
```diff
@@ -165,6 +166,68 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+
+    registry.properties.insert(
+        "json_array_elements".to_string(),
+        FunctionProperty::default().kind(FunctionKind::SRF),
+    );
+    registry.register_function_factory("json_array_elements", |_, _arg_types| {
+        Some(Arc::new(Function {
+            signature: FunctionSignature {
+                name: "json_array_elements".to_string(),
+                args_type: vec![DataType::Nullable(Box::new(DataType::Variant))],
```
b41sh, line 178: We can check `args_type` first, so that `variant` type doesn't need to be cast to `nullable variant`
for example
```rust
if args_type.len() != 1 || args_type[0].remove_nullable() != DataType::Variant {
    return None;
}
```
you can also reference this example:
https://github.com/datafuselabs/databend/blob/main/src/query/functions/src/srfs/mod.rs#L66-L74

src/query/functions/src/srfs/mod.rs
```diff
@@ -165,6 +166,68 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+
+    registry.properties.insert(
+        "json_each".to_string(),
+        FunctionProperty::default().kind(FunctionKind::SRF),
+    );
+    registry.register_function_factory("json_each", |_, _arg_types| {
+        Some(Arc::new(Function {
+            signature: FunctionSignature {
+                name: "json_each".to_string(),
+                args_type: vec![DataType::Nullable(Box::new(DataType::Variant))],
```
b41sh, line 208: ditto

src/query/functions/src/srfs/mod.rs
```diff
@@ -278,3 +329,49 @@ fn build_unnest(
...
     }
 }
+
+fn unnest_variant_array(
+    val: &[u8],
+    row: usize,
+    max_nums_per_row: &mut [usize],
+) -> (Value<AnyType>, usize) {
+    let mut len = 0;
+    let mut builder = StringColumnBuilder::with_capacity(0, 0);
```
b41sh, line 339: We only need a `builder` if the `array_values` function returns some value other than `None`, otherwise we can just return `Scalar::Null`.

b41sh (approved):

PR #13153, opened 2023-10-09
chore(query): refactor agg state merge

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13153)

PR #13165, opened 2023-10-10
fix: fix the invalid memory write bug for aggregating index.

RinChanNOWWW:

## Summary

**UPDATE**: the memory bug was already fixed by #13153. This PR becomes only providing some unit tests for the bug case.

The orginal summary:

We should hold a owned memory for deserializing aggregating index state instead of use the memory owned by the hashtable.
Because the hashtable will be moved out when spilling and we will lose the control of the memory.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13165)


ariesdevil (approved):

src/query/service/src/pipelines/processors/transforms/aggregator/transform_aggregate_partial.rs
```diff
@@ -111,8 +142,13 @@ pub struct TransformPartialAggregate<Method: HashMethodBounds> {
...
 
-    /// A temporary place to hold aggregating state from index data.
-    temp_place: StateAddr,
+    /// A temporary memory to transform aggregating state from index data.
+    ///
+    /// **NOTES**: we should create a new [`Area`] to transform the aggregating index data.
+    /// We cannot use the [`Area`] in `hash_table` to hold the temporary memory,
+    /// because the [`Area`] may be moved out when spilling happens.
+    /// And this [`TransformPartialAggregate`] will lose the control of the memory.
+    temp_memory: TempMemory,
```
sundy-li, line 151: it's covered in #13153, we don't need temp_place now.

sundy-li (approved):

zhang2014 (approved):

PR #13171, opened 2023-10-10
fix(users): enable system.users table again

flaneur2020:

## Summary

now we've removed the auth_string infomation in the `system.users` table and `SHOW USERS` statement, it's saft to have `system.users` enabled again. otherwise it would produce a confusing error message on `SHOW USERS` like this:

```
1025=>
error: 
  --> SQL:1:12
  |
1 | show users;
  |              Unknown table `system`.`users` in catalog 'default'
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13171)


BohuTANG (approved):
LGTM

PR #13193, opened 2023-10-11
chore(query): reduce reallocation for aggregator state builders

sundy-li:

## Summary

Summary about this PR

reduce reallocation for aggregator state builders 

1. For `count`, `sum`, `avg`, `min`, `max` we use fixed-int encoding to align the states
2. Introduce `batch_serialize` API in aggregator function

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13193)


sundy-li:
askbend:summary

src/query/service/src/pipelines/processors/transforms/aggregator/transform_aggregate_final.rs
```diff
@@ -136,28 +136,19 @@ where Method: HashMethodBounds
...
-                            let aggr_column =
-                                agg.value.as_column().unwrap().as_string().ok_or_else(|| {
-                                    ErrorCode::IllegalDataType(format!(
-                                        "Aggregation column should be StringType, but got {:?}",
-                                        agg.value
-                                    ))
-                                })?;
-                            states_binary_columns.push(aggr_column);
+                            let col = agg.value.as_column().unwrap();
+                            states_binary_columns.push(col.slice(0..places.len()));
```
RinChanNOWWW, line 140: `places.len()` should equal to `agg.len()`. We can add an assertion and remove the slice operation.

sundy-li: there is `reach_limit` which could make places size shorter than `agg.len()`

src/query/functions/src/aggregates/aggregate_min_max_any.rs
```diff
@@ -120,14 +126,23 @@ where
     fn serialize(&self, place: StateAddr, writer: &mut Vec<u8>) -> Result<()> {
         let state = place.get::<State>();
 
-        serialize_state(writer, state)
+        if self.serialize_size_per_row().is_some() {
+            serialize_fixed_state(writer, state)
```
RinChanNOWWW, line 130: How about add a template parameter to indicate if need fixed state.

src/query/functions/src/aggregates/aggregate_sum.rs
```diff
@@ -342,13 +344,22 @@ where State: SumState
 
     fn serialize(&self, place: StateAddr, writer: &mut Vec<u8>) -> Result<()> {
         let state = place.get::<State>();
-        serialize_state(writer, state)
+        if self.serialize_size_per_row().is_some() {
```
RinChanNOWWW, line 347: Ditto.

PR #13215, opened 2023-10-11
feat: generate sync agg indexes after update

ariesdevil:

## Summary

Generate sync agg indexes after update table.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13215)


ariesdevil:
Update do bind_table first, then use the returned `bind_context` to bind column expr. 

If we have an agg index, then same table will be bound twice, causing bind_context add same column with different index.

As of this behavior, update table will use the last returned `bind_context` with wrong column indexes.

After discussing with @RinChanNOWWW , we better extract `bind_agg_index` logical out of `bind_table`.

So let's do extract first.

PR #13228, opened 2023-10-12
chore: assert that pushdown columns must exist

andylokandy:

## Summary

Summary about this PR

- Closes #issue


sundy-li:
There will be some inner columns,  like  `SELECT a,_row_id FROM `05_0031_t` where _row_id = 18446735277616529408`

PR #13244, opened 2023-10-13
fix(ddl): bitmap default value deserialize failed

### LINKED ISSUE #13242: bug: bitmap default value deserialize failed
### Search before asking

### Version

v1.2.154-nightly

### What's Wrong?

bitmap default value deserialize failed

### How to Reproduce?

```sql
mysql> create table test (a int not null, b bitmap not null);
Query OK, 0 rows affected (0.71 sec)

mysql> insert into test (a) values (1),(2);
ERROR 1105 (HY000): PanicError. Code: 1104, Text = failed to deserialize bitmap: Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }.
```

zenus:

## Summary

Summary about this PR

- Closes #13242

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13244)


src/query/expression/src/values.rs
```diff
@@ -370,7 +370,12 @@ impl Scalar {
                 let col = builder.build();
                 Scalar::Map(col)
             }
-            DataType::Bitmap => Scalar::Bitmap(vec![]),
+            DataType::Bitmap => {
+                let mut bytes = vec![];
+                let rb = RoaringTreemap::new();
+                rb.serialize_into(&mut bytes).unwrap();
+                Scalar::Bitmap(bytes)
```
b41sh, line 377: put empty bytes as the default value is ok, we should check whether it is empty when deserializing the binary bitmap.

zenus: fine

src/query/expression/src/values.rs
```diff
@@ -1972,12 +1972,16 @@ impl ColumnBuilder {
...
-                    RoaringTreemap::deserialize_from(*b).expect("failed to deserialize bitmap");
-                let mut buf = vec![];
-                rb.serialize_into(&mut buf)
-                    .expect("failed to serialize bitmap");
-                ColumnBuilder::Bitmap(StringColumnBuilder::repeat(&buf, n))
+                if b.is_empty() {
+                    ColumnBuilder::Bitmap(StringColumnBuilder::repeat(b, n))
+                } else {
+                    let rb =
+                        RoaringTreemap::deserialize_from(*b).expect("failed to deserialize bitmap");
```
b41sh, line 1979: It seems that deserialize is not necessary, we can just call the repeat function

tests/sqllogictests/suites/base/03_common/03_0039_insert_into_bitmap
```diff
@@ -26,5 +26,11 @@ SELECT id, to_string(v) FROM t1 order by id
 5 7,8
 6 9,10
 
+statement ok
+CREATE TABLE IF NOT EXISTS t2(a int not null, b bitmap not null) Engine = Fuse
+
+statement ok
+insert into t2 (a) values (1),(2);
```
b41sh, line 33: please add a select test case

zenus: fine

src/query/expression/src/values.rs
```diff
@@ -1972,12 +1972,19 @@ impl ColumnBuilder {
...
-                    RoaringTreemap::deserialize_from(*b).expect("failed to deserialize bitmap");
-                let mut buf = vec![];
-                rb.serialize_into(&mut buf)
-                    .expect("failed to serialize bitmap");
-                ColumnBuilder::Bitmap(StringColumnBuilder::repeat(&buf, n))
+                if b.is_empty() {
+                    let rb = RoaringTreemap::new();
+                    let mut buf = vec![];
+                    rb.serialize_into(&mut buf).unwrap();
+                    ColumnBuilder::Bitmap(StringColumnBuilder::repeat(&buf, n))
```
b41sh, line 1979: Using empty `vec![]` as the default value for `Bitmap` will cause a lot of problems, every time we deserialize it needs to check if the value is empty. So after some discussion, we think we can change the Bitmap's default value to an empty `RoaringTreemap`, which is the way you implemented it in the beginning. Sorry for the extra code changes.
This function just copies the value, there is no need to deserialize, we can directly call the `StringColumnBuilder::repeat` functions like `String` and `Variant` types.

tests/sqllogictests/suites/base/03_common/03_0039_insert_into_bitmap
```diff
@@ -27,4 +27,16 @@ SELECT id, to_string(v) FROM t1 order by id
...
+CREATE TABLE IF NOT EXISTS t2(a int not null, b bitmap not null) Engine = Fuse
+
+statement ok
+insert into t2 (a) values (1),(2);
+
+query IT
+SELECT a, to_string(b) FROM t2
+----
+1 0,1
+2 1,2
```
b41sh, line 39: `to_string(b)` should return an empty string, as the default value is empty bitmap

src/query/expression/src/values.rs
zenus: fine

src/query/expression/src/values.rs
```diff
@@ -370,7 +370,12 @@ impl Scalar {
                 let col = builder.build();
                 Scalar::Map(col)
             }
-            DataType::Bitmap => Scalar::Bitmap(vec![]),
+            DataType::Bitmap => {
+                let rb = RoaringTreemap::new();
+                let mut buf = vec![];
+                rb.serialize_into(&mut buf).unwrap();
+                Scalar::Bitmap(vec![])
```
b41sh, line 377: This should be `Scalar::Bitmap(buf)`. It's a good habit to run the code locally before submitting it to verify that it's correct!

src/query/expression/src/values.rs
```diff
@@ -1972,12 +1977,7 @@ impl ColumnBuilder {
             }
             ScalarRef::Map(col) => ColumnBuilder::Map(Box::new(ArrayColumnBuilder::repeat(col, n))),
             ScalarRef::Bitmap(b) => {
-                let rb =
-                    RoaringTreemap::deserialize_from(*b).expect("failed to deserialize bitmap");
-                let mut buf = vec![];
-                rb.serialize_into(&mut buf)
-                    .expect("failed to serialize bitmap");
-                ColumnBuilder::Bitmap(StringColumnBuilder::repeat(&buf, n))
+               ColumnBuilder::Bitmap(StringColumnBuilder::repeat(&buf, n))
```
b41sh, line 1980: This should be `ColumnBuilder::Bitmap(StringColumnBuilder::repeat(&b, n))`

PR #13254, opened 2023-10-13
chore: vectorize columns serialization.

RinChanNOWWW:

## Summary

Serialize each column in batch to leverage SIMD.

We should preallocate the whole memory space for the serializing target, and it could be a huge consumption. 
Therefore, I add a threshold for the memory usage. If it exceed the limit, degrade the serialization to the original non-batch one (not use the memory buffer).

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13254)

### DIFF
--- a/src/common/hashtable/src/arena.rs
+++ b/src/common/hashtable/src/arena.rs
@@ -35,6 +35,10 @@ impl Area {
     pub fn alloc_layout(&mut self, layout: Layout) -> NonNull<u8> {
         self.bump.alloc_layout(layout)
     }
+
+    pub fn reset(&mut self) {
+        self.bump.reset();
+    }
 }
 
 unsafe impl Send for Area {}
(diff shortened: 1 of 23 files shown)


src/query/expression/src/kernels/group_by_hash.rs
```diff
@@ -781,6 +788,104 @@ pub unsafe fn serialize_column_binary(column: &Column, row: usize, row_space: &m
...
 
+/// The batch version of [`serialize_columns_no_batch`]. Try to leverage SIMD to improve performance.
+///
+/// To serialize a column in batch, we need to preallocate a deterministic memory space for each row of the target column at once.
+pub fn serialize_columns_in_batch(
+    columns: &[Column],
+    num_rows: usize,
+    max_bytes_per_row: usize,
+) -> StringColumn {
+    let mut data: Vec<u8> = Vec::with_capacity(num_rows * max_bytes_per_row);
```
sundy-li, line 799: This data should be a filed with type `Vec<Vec<u8>>`, thus we could reuse the data in next batch.

PR #13261, opened 2023-10-14
feat: add json operators: ->, ->>, #>, #>>

akoshchiy:

## Summary

Added `->`, `->>`, `#>`, `#>>` json operators.

- Related #11270

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13261)


src/query/functions/src/scalars/variant.rs
```diff
@@ -228,6 +230,50 @@ pub fn register(registry: &mut FunctionRegistry) {
         }),
     );
 
+    registry
+        .register_combine_nullable_2_arg::<VariantType, ArrayType<StringType>, VariantType, _, _>(
+            "get_by_keypath",
```
b41sh, line 235: We can change this function to a factory function, so that the key paths only need to be parsed once, and we don't need to repeat the parser every time we execute it. See the `json_path_query` and `json_array` implementations.

b41sh (approved):
LGTM @akoshchiy Very well done, thanks for your contribution!

BohuTANG:
Thanks for your contribution 👍  @akoshchiy 
It would be great if we introduce these new JSON operations in our doc 🚀  @soyeric128

PR #13271, opened 2023-10-16
fix: do not reuse expr when function are if

### LINKED ISSUE #13269: bug: make sql divided by zero
### Search before asking

### Version

v1.2.156

### What's Wrong?

```
 select coalesce((case when a = 0 then NULL else 6/a end) ,0,1) aaa from ( select 0 a union all select 3 a ) test;
APIError: ResponseError with 1001: divided by zero while evaluating function `divide(6, 0)`
```

### How to Reproduce?

 select coalesce((case when a = 0 then NULL else 6/a end) ,0,1) aaa from ( select 0 a union all select 3 a ) test;

ariesdevil:

## Summary

`if` is not same as normal function, they eval with validity, we not reuse expr in these two special functions to avoid bug.

- Closes #13269

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13271)

### DIFF
--- a/src/query/sql/src/evaluator/cse.rs
+++ b/src/query/sql/src/evaluator/cse.rs
@@ -124,8 +124,7 @@ pub fn apply_cse(
 /// and stores the count in a HashMap.
 fn count_expressions(expr: &Expr, counter: &mut HashMap<Expr, usize>) {
     match expr {
-        Expr::FunctionCall { function, .. }
-            if function.signature.name == "if" || function.signature.name == "and_filters" => {}
+        Expr::FunctionCall { function, .. } if function.signature.name == "if" => {}
         Expr::FunctionCall { args, .. } => {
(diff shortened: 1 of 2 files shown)


src/query/sql/src/evaluator/cse.rs
```diff
@@ -124,6 +124,8 @@ pub fn apply_cse(
 /// and stores the count in a HashMap.
 fn count_expressions(expr: &Expr, counter: &mut HashMap<Expr, usize>) {
     match expr {
+        Expr::FunctionCall { function, .. }
+            if function.signature.name == "if" || function.signature.name == "and_filters" => {}
```
sundy-li, line 128: `and_filters`  seems no need to add this.

sundy-li (approved):

Dousir9 (approved):

PR #13284, opened 2023-10-17
fix(query): fix precision of timestamp's stats in parquet-rs

### LINKED ISSUE #13270: bug: filter result empty on external parquet file
With data from AWS billing report parquet files:

```python
from databend import SessionContext

ctx = SessionContext()
ctx.register_parquet("report", "./data/", "*.parquet")
ctx.sql("set use_parquet2=0;").collect()
ctx.sql("create table usage like report;").collect()
ctx.sql("insert into usage select * from report;").collect()
ctx.sql("select sum(line_item_unblended_cost) from report where line_item_usage_start_date between '2023-10-13 00:00:00' and '2023-10-14 00:00:00'").show()
ctx.sql("select sum(line_item_unblended_cost) from usage where line_item_usage_start_date between '2023-10-13 00:00:00' and '2023-10-14 00:00:00'").show()
```

The first result from `report` is null, while the latter one from `usage` is normal.

sundy-li:

## Summary

Summary about this PR

fix(query): fix precision of timestamp's stats in parquet-rs

- Closes #13270

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13284)


everpcpc (approved):

tests/suites/1_stateful/08_select_stage/08_00_parquet/08_00_00_basic.sh
```diff
@@ -35,6 +35,16 @@ echo "drop stage if exists s3;" | $MYSQL_CLIENT_CONNECT
 echo "create stage s3 url = '${DATADIR}' FILE_FORMAT = (type = CSV);"  | $MYSQL_CLIENT_CONNECT
 echo "set use_parquet2 = ${USE_PARQUET2} ; select * from @s3 (FILE_FORMAT => 'PARQUET');" | $MYSQL_CLIENT_CONNECT
 
+
```
youngsofun, line 38: for read only tests.
it would be easier to maintain if use sqllogic test.
under suit stage_parquet

src/query/storages/parquet/src/parquet_rs/statistics/column.rs
```diff
@@ -79,7 +79,17 @@ pub fn convert_column_statistics(s: &Statistics, typ: &TableDataType) -> ColumnS
                     TableDataType::Number(NumberDataType::Int64) => {
                         (Scalar::from(max), Scalar::from(min))
                     }
-                    TableDataType::Timestamp => (Scalar::Timestamp(max), Scalar::Timestamp(min)),
+                    TableDataType::Timestamp => {
+                        let multi = match max.checked_ilog10().unwrap_or_default() + 1 {
```
youngsofun, line 83: as a statistic of some data, max can be any value from 0? 
not always for `2023...`?

sundy-li: > max can be any value start from 0

Yes,  if the number of digits is less or equal  than  10, we set the unit to be second.  This could enlarge the range if the real precision is `ms` or others. But it will still work correctly in range pruner.

> data not always 2023...

Yes, the  timestamp value is time seconds(mills, micro, nanos) from epoch.

youngsofun (approved):

PR #13290, opened 2023-10-17
docs: query environment variables 

### LINKED ISSUE #13258: doc: databend ENV
**Summary**

Add `ENV` list to the deployment.

soyeric128:

## Summary

updated the list of query environment variables.

- Closes #13258

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13290)


PsiACE:
After repeatedly checking and comparing our actual implementation with the potential environment variable configuration, I have identified the root cause of #13258 .

Here we take `AWS_ACCESS_KEY_ID` as an example, which is not a formal option appearing in our configuration file, but it can work as an env.

In our implementation, we will check `STORAGE_S3_ACCESS_KEY_ID`, which aligns with our configuration and environment variable mapping. However, if the corresponding configuration is not provided, a rollback will occur. '

- `src/common/storage/src/operator.rs` line 244. 

```
builder.access_key_id(&cfg.access_key_id);
```

The default behavior of opendal in the `access_key_id()` function.

```
    /// - If access_key_id is set, we will take user's input first.
    /// - If not, we will try to load it from environment.
```

Since we did not explicitly set `disable_config_load`, it will be loaded from the environment variables while building S3 Accessor.

```
        // This is our current config.
        let mut cfg = AwsConfig::default();
        if !self.disable_config_load {
            cfg = cfg.from_profile();
            cfg = cfg.from_env();
        }
```

- `AwsConfig` in reqsign:

```
    /// `access_key_id` will be loaded from
    ///
    /// - this field if it's `is_some`
    /// - env value: [`AWS_ACCESS_KEY_ID`]
    /// - profile config: `aws_access_key_id`
    pub access_key_id: Option<String>,
```

Given the user's familiarity with using AWS_ACCESS_KEY_ID, it is acceptable to this environment variable. Users are encouraged to make use of the environment variables they are already accustomed to. 

These environment variables are supported by resign, so I think we need to supplement them together with @Xuanwo  .

Issue #11378, opened 2023-05-09
Feature:  optimize read of small row group in parquet 


youngsofun:
**Summary**

like small file https://github.com/datafuselabs/databend/pull/11231

1. read the whole row group too small 
2. merge io of adjacent chunks

PR #13322, opened 2023-10-18
fix(ddl): inner type of nested array type can not be set as not null

### LINKED ISSUE #13235: bug: inner type of nested type can not be set as not null
### Search before asking

### Version

v1.2.154-nightly

### What's Wrong?

inner type of nested type can not be set as not null

### How to Reproduce?

```sql
mysql> create table test(a int, b array(int not null));
ERROR 1105 (HY000): SyntaxException. Code: 1005, Text = error:
  --> SQL:1:38
  |
1 | create table test(a int, b array(int not null))
  | ------                   - -----     ^^^ expected `NULL`, `UNSIGNED`, `)`, or `(`
  | |                        | |
  | |                        | while parsing type name
  | |                        while parsing `<column name> <type> [DEFAULT <expr>] [AS (<expr>) VIRTUAL] [AS (<expr>) STORED] [COMMENT '<comment>']`
  | while parsing `CREATE TABLE [IF NOT EXISTS] [<database>.]<ta
```

zenus:

## Summary

Summary about this PR

- Closes #13235

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13322)


sundy-li (approved):

src/query/ast/src/parser/expr.rs
```diff
@@ -1387,15 +1387,15 @@ pub fn type_name(i: Input) -> IResult<TypeName> {
...
+            ) ~ (NOT|NULL)? : "type name"
             },
             rule! {
             ( #ty_date
             | #ty_datetime
             | #ty_string
             | #ty_variant
             | #ty_nullable
-            ) ~ NULL? : "type name" },
+            ) ~(NOT|NULL)? : "type name" },
```
b41sh, line 1398: this is incorrect, `not null` will cause a parse error, we need to support both parses `not null` and `null`.

tests/sqllogictests/suites/base/05_ddl/05_0000_ddl_create_tables
```diff
@@ -332,6 +332,12 @@ create table t_with_bloom_index(a int, b int) bloom_index_columns='b'
 statement ok
 drop table if exists t_with_bloom_index;
 
+statement ok
+create table t_nest_type(a int, b array(int not null));
```
b41sh, line 336: test case should include three types: `array(int not null)`, `array(int null)` and `array(int)`

and please also add some unit tests for parser in this file https://github.com/datafuselabs/databend/blob/main/src/query/ast/tests/it/parser.rs

src/query/ast/src/parser/expr.rs
zenus: sure

src/query/ast/src/parser/expr.rs
```diff
@@ -1367,6 +1367,146 @@ pub fn type_name(i: Input) -> IResult<TypeName> {
...
+                        .try_into()
+                        .map_err(|_| ErrorKind::Other("scale is too large"))?
+                } else {
+                    0
+                },
+            })
+        },
+    );
+    let ty_array = map(
+        rule! { ARRAY ~ "(" ~ #basic_type_name ~ NOT? ~ NULL? ~ ")" },
```
b41sh, line 1461: There's too much duplicate code to write this way, and it doesn't handle multiple nested cases and other nested types like map and tuple, for example:
```sql
create table t(a int, b array(array(int not null)));
create table t(a int, b tuple(int not null, string not null));
```
You can write like this example:
```rust
    map_res(
        alt((
            rule! {
            ( #ty_boolean
            | #ty_uint8
            ...
            | #ty_string
            | #ty_variant
            | #ty_nullable
            ) ~ NOT? ~ NULL? : "type name" },
        )),
        |(ty, opt_not, opt_null)| {
            match (opt_not, opt_null) {
                (None, Some(_)) => Ok(ty.wrap_nullable()),
                (Some(_), Some(_)) => Ok(ty),
                (None, None) => Ok(ty),
                (Some(_), None) => Err(ErrorKind::Other("invalid type name")),
            }
        },
    )(i)
```

zenus:
@b41sh Maybe I should use the previous method first and leave the code redundancy problem to be optimized later?

b41sh:
> @b41sh Maybe I should use the previous method first and leave the code redundancy problem to be optimized later?

No, `type_name` is nested, adding a `basic_type_name` function does not solve the problem.
The `nullable_constraint` field in `ColumnDefinition` is also related to this change and needs to be changed together.

zenus:
> field in `ColumnDefinition` is also related to this change and needs to be changed together.

@b41sh  good idea

b41sh:
Hi, @zenus 
You can add an item `Undefined(Box<TypeName>)` to `TypeName` to store types that do not specify `null` or `not null`. And using the `resolve_type_name` function to converts undefined types to `nullable` or `not null` types based on the definition of `self.ctx.get_settings().get_ddl_column_type_nullable()`. This way the `nullable_constraint` can also be removed.

zenus:
@b41sh  thanks, i'm trying

b41sh:
Hi, @zenus Have any recent developments? It looks like it's been quite a while since an update.

zenus:
@b41sh  i'm sorry ,i 'm busy with the #11378 last week, it is time to go on.

b41sh:
Hi @zenus  I'm closing this PR since it's been a long time without any progress, feel free to keep contributing code to other issues.

Issue #12645, opened 2023-08-31
Feature: optimize Inlist to semi join


xudong963:
**Summary**

Description for this feature.

xudong963:
We have materialized cte, so this feature will be relatively easy to implementation.

PR #13399, opened 2023-10-23
feat: convert in list to subquery

xudong963:

## Summary

If `inlist`'s values are large, `inlist` will be inefficient, we can convert it to subquery, and the subquery will be converted to join.

- Closes https://github.com/datafuselabs/databend/issues/12645

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13399)


src/query/settings/src/settings_default.rs
```diff
@@ -155,6 +155,12 @@ impl DefaultSettings {
                     possible_values: None,
                     display_in_show_settings: true,
                 }),
+                ("inlist_to_subquery", DefaultSettingValue {
```
BohuTANG, line 158: No need this setting and check the IN list length to enable or disable?

xudong963: yes, i need some time to find a proper threshold, then remove the setting

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -353,6 +354,20 @@ impl<'a> TypeChecker<'a> {
                 not,
                 ..
             } => {
+                if list.len() >= 1024 {
```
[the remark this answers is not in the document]
xudong963: done

Dousir9 (approved):

PR #13458, opened 2023-10-26
feat(rbac): Centralized object ownership management

### LINKED ISSUE #13440: refactor: add a GrantOwnership object in meta to index all the ownerships
currently the ownership infomation is seperated in different objects, like databases, tables, etc.

this made it hard for us to consolidate the ownership verification logics in one place.

even worse, when it comes with some external objects like warehouse, tasks, this makes things even harder.

another way to deal with it is to consider the ownership as an external index like:

```
/grant-ownerships/:object-type/:object-id -> :GrantOwnershipInfo
```

```
struct GrantOwnershipInfo {
  role_identity RoleIdentity,
  created_at timestamp,
  granted_by UserIdentity,
}
```

flaneur2020:

## Summary

fixes #13440

- seperate validate_privilege and validate_ownership to gradually migrate from GrantObject and GrantObjectByID
- grant ownership statement: validate the only the current owner can transfer ownership, and the new owner have to be one of the available role.
- grant/drop ownership on create/drop database and tables.

known caveats:

- the earlier ownerships will become PUBLIC, user need re-assign the ownerships after this PR merged
- this PR only covered the database/table objects, the other objects like view/tasks will be treated as PUBLIC. (will be fixed soon)

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13458)


tests/suites/0_stateless/20+_others/20_0012_privilege_access.result
```diff
@@ -19,7 +19,7 @@ test -- select
 1
 1
 1
-Error: APIError: ResponseError with 1063: Permission denied, privilege [Select] is required on 'default'.'default2'.'v_t20_0012' for user 'test-user'@'%' with roles [public,test-role1,test-role2]
+1
```
BohuTANG, line 22: Looks it's time to spill out a new dir logic tests for RBAC from `20+_others`, we can do in later PR  :)

tests/suites/0_stateless/20+_others/20_0012_privilege_access.sh
```diff
@@ -146,11 +149,15 @@ echo "drop table if exists default.test_t" |  $BENDSQL_CLIENT_CONNECT
 echo "create table default.test_t(id int not null)" |  $BENDSQL_CLIENT_CONNECT
 echo "show grants for a" |  $BENDSQL_CLIENT_CONNECT
 echo "show databases" | $USER_A_CONNECT
+echo "select 'test -- show tables'" | $MYSQL_CLIENT_CONNECT
```
TCeason, line 152: MYSQL_CLIENT_CONNECT replace to BENDSQL_CLIENT_CONNECT

src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -35,13 +36,95 @@ impl PrivilegeAccess {
...
+            let object_by_id = self.convert_grant_object_by_id(object).await.or_else(|e| {
+                if e.code() == ErrorCode::UNKNOWN_DATABASE
+                    || e.code() == ErrorCode::UNKNOWN_TABLE
+                    || e.code() == ErrorCode::UNKNOWN_CATALOG
+                {
+                    Ok(None)
+                } else {
+                    Err(e.add_message("error on validating access"))
+                }
+            })?;
```
BohuTANG, line 111:
```suggestion
        let object_by_id = self.convert_grant_object_by_id(object).await.or_else(|e| {
            match e.code() {
                ErrorCode::UNKNOWN_DATABASE | ErrorCode::UNKNOWN_TABLE | ErrorCode::UNKNOWN_CATALOG => Ok(None),
                _ => Err(e.add_message("error on validating access")),
            }
        })?;
```

flaneur2020: magic, thank you for the suggestion

src/query/service/src/sessions/session_privilege_mgr.rs
```diff
@@ -231,6 +213,28 @@ impl SessionPrivilegeManager for SessionPrivilegeManagerImpl {
...
+        if !available_roles.iter().any(|r| r.name == owner_role.name) {
+            return Err(ErrorCode::PermissionDenied(
+                "Permission denied, current session do not have the ownership of this object"
+                    .to_string(),
+            ));
+        }
+
+        Ok(())
+    }
+
```
BohuTANG, line 237:
```suggestion
async fn validate_ownership(&self, object: &GrantObjectByID) -> Result<()> {
    let role_mgr = RoleCacheManager::instance();
    let tenant = self.session_ctx.get_current_tenant();

    if let Some(owner_role) = role_mgr.find_object_owner(&tenant, object).await? {
        let available_roles = self.get_all_available_roles().await?;
        if !available_roles.iter().any(|r| r.name == owner_role.name) {
            return Err(ErrorCode::PermissionDenied(
                "Permission denied, current session does not have the ownership of this object".to_string(),
            ));
        }
    }

    Ok(())
}
```

BohuTANG (approved):
Give two suggestions from GPT4, else LGTM.

src/query/service/src/sessions/session_privilege_mgr.rs
flaneur2020: I sometimes prefer to return early on some specified logic path in an function, thus the later code can do not care about the previous logics, sometimes this may make things easier to handle. quq

in this case, `if the object is not owned by any role, then considered as PUBLIC, which is always true` is the case which is ok to return eary by an `return Ok(())`, which means `everything is allowed on a object owned by PUBLIC`, but less obvious in the suggested version without comment.

src/query/service/src/interpreters/access/privilege_access.rs
BohuTANG: I forgot to tell the GPT4 follow our code style, so the style is broken.

BohuTANG:
`[linux_hive / test_stateful_hive_standalone](https://github.com/datafuselabs/databend/actions/runs/6704556844/job/18217599543?pr=13458#logs)` failed is not related to this PR, let's merge.

PR #13463, opened 2023-10-26
chore(query): fix take block vec fast path

Dousir9:

## Summary

We can use fast path only when all columns in ColumnVec are all_true or all_false.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13463)


src/query/expression/src/kernels/take_chunks.rs
```diff
@@ -772,15 +772,25 @@ impl Column {
     pub fn take_block_vec_boolean_types(col: &[Bitmap], indices: &[RowPtr]) -> Bitmap {
         let num_rows = indices.len();
         // Fast path: avoid iterating column to generate a new bitmap.
+        // If all [`Bitmap`] are all_true or all_false and `num_rows <= bitmap.len()`,
+        // we can just slice it.
+        let mut all_true = true;
+        let mut all_false = true;
         for bitmap in col.iter() {
```
sundy-li, line 779: better to use:
```
let mut total_len = 0;
let mut unset_bits = 0;

for bitmap in col.iter() {
    unset_bits += bitmap.unset_bits();
    total_len += bitmap.len();
} 

if total_len == unset_bits || unset_bits == 0 {
   // goes faster path
   
}
```

Dousir9: Done.

src/query/expression/src/kernels/take_chunks.rs
```diff
@@ -772,15 +772,22 @@ impl Column {
     pub fn take_block_vec_boolean_types(col: &[Bitmap], indices: &[RowPtr]) -> Bitmap {
```
BohuTANG, line 772: Do we need an unit test for this function to ensure correctness?

sundy-li (approved):

PR #13471, opened 2023-10-27
chore: add unit test for kernels

Dousir9:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13471)


src/query/expression/src/kernels/tests.rs
```diff
@@ -0,0 +1,263 @@
+// Copyright 2021 Datafuse Labs
```
BohuTANG, line 1: For `pub` functin tests, would be better to put to:
https://github.com/datafuselabs/databend/tree/main/src/query/expression/tests/it

Dousir9: Thx, done.

sundy-li (approved):

PR #13485, opened 2023-10-29
refactor: MapApi should return Result<T, io::Error>

drmingdrmer:

##### refactor: MapApi should return Result<T, io::Error>

The `MapApi` serves as the API for the state machine, which will be
upgraded to support on-disk data storage. Consequently, to ensure
I/O errors are revealed to the calling function, `MapApi` should return
a `Result<T, io::Error>` instead of simply `T`.

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13485)


ariesdevil (approved):
LGTM

PR #13487, opened 2023-10-29
refactor: Alter all command line's value name to VALUE

Xuanwo:

## Summary

This is the behavior that confused our users a lot:

```shell
--storage-oss-access-key-id <OSS_ACCESS_KEY_ID>
    Access key id for OSS storage
    
    [default: ]

--storage-oss-access-key-secret <OSS_ACCESS_KEY_SECRET>
    Access Key Secret for OSS storage
    
    [default: ]

--storage-oss-bucket <OSS_BUCKET>
    Bucket for OSS
    
    [default: ]
```

Our users believe that `OSS_ACCESS_KEY_ID` and `OSS_ACCESS_KEY_SECRET` here serve as the  environment name for this argument. However, it's not. It's just the value name for this arg and default to it's own field type. We are not using the `clap`'s env feature.

To avoid this confusion anymore, I submit this PR to change all args value name to `VALUE`. After this change, the output of help will be:

```shell
--storage-oss-access-key-id <VALUE>
    Access key id for OSS storage
    
    [default: ]

--storage-oss-access-key-secret <VALUE>
    Access Key Secret for OSS storage
    
    [default: ]

--storage-oss-bucket <VALUE>
    Bucket for OSS
    
    [default: ]
```

And it's environment name is the simple mapping from `--storage-oss-bucket` to `STORAGE_OSS_BUCKET`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13487)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -148,7 +148,7 @@ thiserror = { version = "1" }
 semver = "1.0.14"
 
 # CLI
-clap = { version = "4.4.2", features = ["derive", "env"] }
+clap = { version = "4.4.2", features = ["derive"] }
 
 # server
 tonic = { version = "0.10.2", features = ["transport", "codegen", "prost", "tls-roots", "tls"] }
--- a/src/query/config/src/background_config.rs
+++ b/src/query/config/src/background_config.rs
@@ -27,7 +27,7 @@ use serde::Serialize;
 #[derive(Clone, Debug, PartialEq, Default, Eq, Serialize, Deserialize, Args)]
 #[serde(default)]
 pub struct BackgroundConfig {
-    #[clap(long = "enable-background-service")]
+    #[clap(long = "enable-background-service", value_name = "VALUE")]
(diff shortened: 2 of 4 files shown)


Xuanwo:
Also, cc @ZhiHanZ, I found the background services leaks a lot args without correct prefix:

```shell
--compact-mode <COMPACT_MODE>
    [default: one_shot]

--target-tables <TARGET_TABLES>
    

--segment-limit <SEGMENT_LIMIT>
    

--block-limit <BLOCK_LIMIT>
    

--duration-secs <DURATION_SECS>
    [default: 1800]

--cron <CRON>
    [default: ]

--time-zone <TIME_ZONE>
```

We should set the correct `long` for them or just `#[clap(skip)]` if we don't want users to start background services via command line.

everpcpc (approved):

PsiACE (approved):

PR #13491, opened 2023-10-30
chore: change input_read_buffer_size default value from 1MB to 4MB.

youngsofun:

## Summary

it is currently used as buffer size to read CSV/TSV/NDJSON.

1MB is too small for S3.
but should not be too large so that IO/decompress/deserialize can be fully pipelined.

4MB is default BlockSize for JuiceFS.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13491)


Xuanwo (approved):
LGTM

PR #13497, opened 2023-10-30
chore(query): improve filter boolean and introduce fast path

Dousir9:

## Summary

Improved the implementation of filtering the bitmap of nullable column and introduced fast path.
### Benchmark
For example TPC-H SF100 External Parquet Q3, 5.9s -> 5.3s
Flamegraph before:
截屏2023-10-30 18 55 58

Flamegraph after:
截屏2023-10-30 18 55 31
 
- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13497)


src/query/expression/src/kernels/filter.rs
```diff
@@ -424,4 +422,219 @@ impl Column {
...
+                            while continuous_selected > 0 {
+                                if value {
+                                    buf |= BIT_MASK[count % 8];
+                                } else {
+                                    unset_bits += 1;
+                                }
+                                count += 1;
+                                continuous_selected -= 1;
+                                if count % 8 == 0 {
+                                    *builder.get_unchecked_mut(builder_len) = buf;
```
sundy-li, line 520: This could be unknown behavior to access uninitialized value

sundy-li: use ptr write instead

src/query/expression/src/kernels/filter.rs
```diff
@@ -424,4 +423,215 @@ impl Column {
 
         StringColumn::new(data.into(), offsets.into())
     }
+
+    fn copy_continuous_bits(
```
sundy-li, line 427: this function should be unsafe

src/query/expression/src/kernels/filter.rs
```diff
@@ -424,4 +423,215 @@ impl Column {
...
+        let mut builder_idx = 0;
+        let mut unset_bits = 0;
+        let mut buf = 0;
+
+        let (bitmap_slice, bitmap_offset, _) = bitmap.as_slice();
+        let mut bitmap_idx = 0;
+
+        let (mut filter_slice, filter_offset, mut filter_length) = filter.as_slice();
+        unsafe {
+            if filter_offset > 0 {
```
sundy-li, line 497: need a test case to cover this branch, using `slice` to create bitmap with non-zero offset

Dousir9: Done.

sundy-li (approved):

BohuTANG:
@Dousir9 @sundy-li 

There is a panic:
```
--- STDOUT:              common-expression::it kernel::test_filter_boolean ---

running 1 test
test kernel::test_filter_boolean ... FAILED

failures:

failures:
    kernel::test_filter_boolean

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 30 filtered out; finished in 0.02s


--- STDERR:              common-expression::it kernel::test_filter_boolean ---
thread 'kernel::test_filter_boolean' panicked at 'attempt to subtract with overflow', src/query/expression/src/kernels/filter.rs:510:17
```

https://github.com/datafuselabs/databend/actions/runs/6719519748/job/18261330444?pr=13514#step:4:2281

PR #13502, opened 2023-10-31
refactor:  improve encoder/decoder for formats.

youngsofun:

## Summary

1. decuple Decoder/Encoder from formats.  Decoder/Encoder provides standard Text repr (except settings), Format is response for escaping/quoting of the string.
    -  for input TSV/CSV/XML share SeparatedTextDecoder
    -  for output TSV/CSV share FieldEncoderCSV
3. treat map/array/tuple as a special format, not use the same options as outer formats. only do necessary escape when output(`write_quoted_string`), use double-quote to escape `\'`.  use composition instead of inheritance ( making the coding intention more verbose)：
    - for input， TSV/CSV/XML has a NestedDecoder for nested types (with diff settings).
    - for output,  Json decoder has a ValueDecoder for simple types; TSV/CSV has one for simple and one for nested types, with diff settings.  
    - settings for values itself is not really used by not,  it not cost much, makes code more clear.  we reserve it for later usage.
5. distinguish InputCommonSettings and OutputCommonSettings. InputCommonSettings support multi repr for a special value
6.  write_tsv_escaped_string no longer escaped quote.

some fixes:
- string in nested allow double-quote
- fix some test results

add tests:
-  [tsv_types](https://github.com/datafuselabs/databend/pull/13502/files#diff-2143871701849a898ddb71ff8cfad8d9f54309e03ebf6b02b07cffcdfd78c80a) for load unload of complicate type.
-  test csv/tsv for null_display. (rm another stateful test)

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13502)


BohuTANG:
4 failing checks :/

BohuTANG (approved):
LGTM

PR #13503, opened 2023-10-31
fix: build_bitmap for all unsigned integer types

### LINKED ISSUE #13499: Feature: build_bitmap need support array(Uint null)
**Summary**
Databend before every column is not null, so build_bitmap work well. Now we make databend column default is null. 

```
wubx@192.168.1.201:8000/wubx> select list(cast(uid as uint32)),typeof(uid) from (select 1 uid union all select 2 uid union all select null uid) where uid is not null group by all;

SELECT
  list(cast(uid AS uint32)),
  typeof(uid)
FROM
  (
    SELECT
      1 uid
    UNION
    ALL
    SELECT
      2 uid
    UNION
    ALL
    SELECT
      NULL uid
  )
WHERE
  uid IS NOT NULL
GROUP BY
  ALL

-[ RECORD 1 ]-----------------------------------
list(cast(uid as uint32)): [2,1]
              typeof(uid): TINYINT UNSIGNED NULL


wubx@192.168.1.201:8000/wubx> select build_bitmap(list(cast(uid as uint32))),typeof(uid) from (select 1 uid union all select 2 uid union all select null uid) where uid is not null group by all;
error: APIError: ResponseError with 1065: error:
  --> SQL:1:8
  |
1 | select build_bitmap(list(cast(uid as uint32))),typeof(uid) from (select 1 uid union all select 2 uid union all select null uid) where uid is not null group by all
  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no overload satisfies `build_bitmap(Array(UInt32 NULL))`

has tried possible overloads:
  build_bitmap(Array(UInt64)) :: Bitmap            : unable to unify `UInt32 NULL` with `UInt64`
  build_bitmap(Array(UInt64) NULL) :: Bitmap NULL  : unable to unify `UInt32 NULL` with `UInt64`
```

img_v2_f4715437-7cd9-4132-9fa0-7c39aabf360g

ariesdevil:

## Summary

Add `build_bitmap` for all unsigned integer types

- Closes #13499

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13503)


src/query/functions/src/scalars/bitmap.rs
```diff
@@ -62,19 +67,26 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+    for num_type in ALL_UNSIGNED_INTEGER_TYPES {
+        with_integer_mapped_type!(|NUM_TYPE| match num_type {
+            NumberDataType::NUM_TYPE => {
+                registry.register_passthrough_nullable_1_arg::<ArrayType<NullableType<NumberType<NUM_TYPE>>>, BitmapType, _, _>(
+                    "build_bitmap",
+                    |_, _| FunctionDomain::Full,
+                    vectorize_with_builder_1_arg::<ArrayType<NullableType<NumberType<NUM_TYPE>>>, BitmapType>(|arg, builder, _ctx| {
+                        let mut rb = RoaringTreemap::new();
+                        arg.iter().filter(|a| a.is_some()).for_each(|a| {
+                            rb.insert(a.unwrap() as u64);
```
b41sh, line 79: If the value is a negative number, directly casting to u64 may give the wrong result.

b41sh: It looks like it only supports the unsigned integer type, can we also support the signed integer type and report an error when encountering a negative number?

ariesdevil: sounds good, let's do it.

src/query/functions/src/scalars/bitmap.rs
```diff
@@ -62,19 +67,32 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        with_integer_mapped_type!(|NUM_TYPE| match num_type {
+            NumberDataType::NUM_TYPE => {
+                registry.register_passthrough_nullable_1_arg::<ArrayType<NullableType<NumberType<NUM_TYPE>>>, BitmapType, _, _>(
+                    "build_bitmap",
+                    |_, _| FunctionDomain::MayThrow,
+                    vectorize_with_builder_1_arg::<ArrayType<NullableType<NumberType<NUM_TYPE>>>, BitmapType>(|arg, builder, ctx| {
+                        let mut rb = RoaringTreemap::new();
+                        for a in arg.iter() {
+                            if let Some(a) = a {
+                                if a > 0 {
```
b41sh, line 80: `a` can also equal to zero.

b41sh (approved):

PR #13535, opened 2023-11-02
feat: add translate function

### LINKED ISSUE #13432: Feature: add translate() funtion
**Summary**

to reduce the nested calls of replace().
eg.
use
translate('abcde','abc','ABC') instead of
replace(replace(replace('abcde','a','A'),'b','B'),'c','C')
ref https://docs.oracle.com/cd/B13789_01/server.101/b10759/functions181.htm
https://www.postgresql.org/docs/current/functions-string.html

Syntax:
```
TRANSLATE( <subject>, <sourceAlphabet>, <targetAlphabet> )
```

Example:
```
SELECT TRANSLATE('abcde','abc','ABC');
+--------------------------------+
| TRANSLATE('abcde','abc','ABC') |
|--------------------------------|
| ABCdef                         |
+--------------------------------+
```

guojidan:

## Summary

Summary about this PR

- Closes #13432 issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13535)


guojidan:
@TCeason please review this pr : )

src/query/functions/src/scalars/string.rs
```diff
@@ -239,6 +239,30 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        "translate",
+        |_, _, _, _| FunctionDomain::Full,
+        vectorize_with_builder_3_arg::<StringType, StringType, StringType, StringType>(
+            |str, from, to, output, _| {
+            if from.is_empty() || from == to {
+                output.put_slice(str);
+                output.commit_row();
+                return;
+            }
+            let to_len = to.len();
```
BohuTANG, line 252: Should check: `from.len() == to.len()`?

TCeason:
This pr adds a string function translate. Please review this pr. @andylokandy

sundy-li (approved):

src/query/functions/src/scalars/string.rs
guojidan: The length of from and to can be different

Replaces each character in string that matches a character in the from set with the corresponding character in the to set. If from is longer than to, occurrences of the extra characters in from are deleted.

BohuTANG:
When merge with main, there are some mismatched types error: https://github.com/datafuselabs/databend/actions/runs/6806765528/job/18508525426?pr=13535#step:4:7590

andylokandy (approved):

BohuTANG:
Thanks for the contribution.

PR #13537, opened 2023-11-02
feat: allow restricting a current role on starting a query

flaneur2020:

## Summary

An user may have multiple roles, sometimes we hope to restrict the role to limit the privilege we have. It'd be useful when you have a scheduler which hopes to execute multiple tasks in arbitary roles.

- [x] rename auth_role to restricted_role
- [x] add restricted to set_current_role_checked
- [x] add test

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13537)

### DIFF
--- a/src/query/service/src/interpreters/interpreter_role_set.rs
+++ b/src/query/service/src/interpreters/interpreter_role_set.rs
@@ -63,7 +63,7 @@ impl Interpreter for SetRoleInterpreter {
                 .await?;
         } else {
             session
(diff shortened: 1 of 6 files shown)


flaneur2020:
there's still a follow up to return the role in the response after https://github.com/datafuselabs/databend/pull/13062 get merged, it can be added in another pr

ZhiHanZ (approved):
unit test LGTM, but we have to add integration test on session role request behavior into CI once bendsql add role dsn param support

PR #13583, opened 2023-11-05
feat(cloud):add system.task_history table

hantmac:

## Summary

- Since snowflake task has [task history](https://docs.snowflake.com/en/sql-reference/functions/task_history), here we add `task_history` table to system. This table records the information of every task run history.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13583)


ZhiHanZ:
We only need a table named `system.task_history` to retrieve task histories, I don't think we need to add extra SQL command for that requirement cc @hantmac

hantmac:
cc @ZhiHanZ  Would u like to help review this pr? 😄

ZhiHanZ (changes requested):
I think it is under good shape!
could you add some sqllogic tests on the system.task_history table, which is located on https://github.com/datafuselabs/databend/blob/main/tests/sqllogictests/suites/task/task_ddl_test.test.

Plus, I think you need to provide the grpc service implementation around ShowTaskRun API on https://github.com/datafuselabs/databend/blob/main/tests/cloud_control_server/simple_server.py, and as you changed the protobuf, you need to generate the python protobuf file.
using the following command(run on databend root directory
`make genproto`
https://github.com/datafuselabs/databend/blob/main/Makefile#L115

hantmac:
> I think it is under good shape! could you add some sqllogic tests on the system.task_history table, which is located on https://github.com/datafuselabs/databend/blob/main/tests/sqllogictests/suites/task/task_ddl_test.test.
> 
> Plus, I think you need to provide the grpc service implementation around ShowTaskRun API on https://github.com/datafuselabs/databend/blob/main/tests/cloud_control_server/simple_server.py, and as you changed the protobuf, you need to generate the python protobuf file. using the following command(run on databend root directory `make genproto` https://github.com/datafuselabs/databend/blob/main/Makefile#L115

@ZhiHanZ Thank u very much! I have fixed the problems you mentioned above.

ZhiHanZ (approved):
LGTM

Issue #12595, opened 2023-08-27
Feature: Merge Into Optimizations


JackTan25:
Merge Into is supported by 12350, but it's not completed. We have many things to do.

## Optimizations
- [] a more powerful hashTable which support split data block like above.

- [] optimizer for query source.

- [] support not match by source.

- [] insert-only optimization.

- [] pre-compute cols for macth_expr/not match expr.

- [] parallel merge into

- [] distributed merge into

- [] support streaming source

- [] support computed expr

- [] support input schema which is not the same with table schema (filling default columns)

- [] support star "*" function to enhance merge into.

- [] use full outer join to reduce I/O, fetch complete fields for update clause to simplify block split.

- [] matched and unmatched clauses exprs (only target table filed) push down

BohuTANG:
> Hello, may I help with completing this part of the development? I'm quite interested in this aspect.

Sure, thanks.

But we are in the early stage of the merge into design, it's better to wait us some time when the fundamental working is done, we may spill out some tasks, I think then is good.

ct20000901:
> > Hello, may I help with completing this part of the development? I'm quite interested in this aspect.
> 
> Sure, thanks.
> 
> But we are in the early stage of the merge into design, it's better to wait us some time when the fundamental working is done, we may spill out some tasks, I think then is good.

Perhaps this question may seem a bit forward, but can I also participate in the process of basic design and development for 'merge_into'? I plan to attempt an internship with Databend after completing one or two issues, so I would like to have a deeper understanding of Databend's overall design.

BohuTANG:
@ct20000901 Feel free to take any task/issue you are interested :)

PR #13586, opened 2023-11-05
feat: support table as source

JackTan25:

## Summary
Summary about this PR
support table as source.for example:
```sql
merge into t1 using t2 on t1.a = t2.a when matched then delete when not matched then insert *;
```
- Closes #issue
https://github.com/datafuselabs/databend/issues/12595

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13586)


tests/sqllogictests/suites/base/09_fuse_engine/09_0026_merge_into.test
```diff
@@ -622,6 +622,32 @@ select a, b, c from tt2 order by b;
...
+create table t2(a int);
+
+statement ok
+insert into t1 values(1);
+
+statement ok
+insert into t2 values(1),(2);
+
+statement ok
+merge into t1 using t2 on t1.a = t2.a when matched then delete when not matched then insert *;
```
b41sh, line 645: Will an error be reported if the table fields of `t1` and `t2` are different? It would be good to add some tests of this type.

JackTan25: ok

b41sh (approved):

PR #13589, opened 2023-11-05
chore(query): add check_valid during Serializing block and debug_asserts

sundy-li:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13589)

### DIFF
--- a/src/query/expression/src/block.rs
+++ b/src/query/expression/src/block.rs
@@ -47,7 +47,7 @@ pub struct DataBlock {
     meta: Option<BlockMetaInfoPtr>,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct BlockEntry {
     pub data_type: DataType,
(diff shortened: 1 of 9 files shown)


src/query/expression/src/block.rs
```diff
@@ -122,6 +124,29 @@ impl DataBlock {
         }
     }
 
+    fn check_columns_valid(columns: &Vec<BlockEntry>, num_rows: usize) -> Result<()> {
```
zhang2014, line 127:
```suggestion
    #[cfg(debug_assertions)]
    fn check_columns_valid(columns: &Vec<BlockEntry>, num_rows: usize) -> Result<()> {
```

sundy-li: this function is used during serializing, it's enabled not only for debug.

dantengsky (approved):

Dousir9 (approved):

PR #13595, opened 2023-11-06
chore(query): add fuzz test for insert

sundy-li:

## Summary

Summary about this PR

1. fuzz test insert from random
2. improve `CAST_FROM_STRING_RULES` which always cast string into large number types.
3. sql with regexp pattern`RAND_(\d+)_(\d+)` will be replaced by a random number from the range.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13595)

### DIFF
--- a/src/query/functions/src/cast_rules.rs
+++ b/src/query/functions/src/cast_rules.rs
@@ -206,16 +206,10 @@ pub const GENERAL_CAST_RULES: AutoCastRules = &[
 /// used to allow `add_hours('2023-01-01 00:00:00', '1')`. But they should be disabled
 /// for comparison functions, because `1 < '1'` should be an error.
 pub const CAST_FROM_STRING_RULES: AutoCastRules = &[
-    (DataType::String, DataType::Number(NumberDataType::UInt8)),
-    (DataType::String, DataType::Number(NumberDataType::UInt16)),
-    (DataType::String, DataType::Number(NumberDataType::UInt32)),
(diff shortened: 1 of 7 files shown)


src/query/functions/src/cast_rules.rs
```diff
@@ -206,16 +206,10 @@ pub const GENERAL_CAST_RULES: AutoCastRules = &[
 /// used to allow `add_hours('2023-01-01 00:00:00', '1')`. But they should be disabled
 /// for comparison functions, because `1 < '1'` should be an error.
 pub const CAST_FROM_STRING_RULES: AutoCastRules = &[
-    (DataType::String, DataType::Number(NumberDataType::UInt8)),
-    (DataType::String, DataType::Number(NumberDataType::UInt16)),
-    (DataType::String, DataType::Number(NumberDataType::UInt32)),
-    (DataType::String, DataType::Number(NumberDataType::UInt64)),
-    (DataType::String, DataType::Number(NumberDataType::Int8)),
-    (DataType::String, DataType::Number(NumberDataType::Int16)),
-    (DataType::String, DataType::Number(NumberDataType::Int32)),
```
andylokandy, line 215: what's the reason why removing these rules?

andylokandy (approved):

src/query/functions/src/cast_rules.rs
sundy-li: select '333' + 4; 

 will be wrong

PR #13632, opened 2023-11-08
chore: fix cluster key in merge into filter

SkyFan2002:

## Summary

Only when there is more than one cluster key, cluster keys will be parsed as a `Tuple`.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13632)


dantengsky (approved):
:+1:  thanks

PR #13649, opened 2023-11-09
fix: address the slow range() function in the state machine

drmingdrmer:

##### fix: address the slow range() function in the state machine

A State Machine is a structure with multiple levels, with the top level
being writable.

The top writable level cannot be borrowed to create a static stream.
Thus the range result must be copied.

In this commit, we minimize the top level as much as possible.
Additionally, when querying a range from the static shared label, we avoid copying the result.

## Changelog

- Bug Fix

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13649)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2382,6 +2382,7 @@ dependencies = [
  "common-tracing",
  "derive_more",
  "futures",
+ "futures-async-stream",
  "futures-util",
  "hostname",
  "lazy_static",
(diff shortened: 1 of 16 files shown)


drmingdrmer (commented):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13649)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @drmingdrmer)

ariesdevil (approved):

Issue #13544, opened 2023-11-02
feat: seperate primary role and secondary role


flaneur2020:
may looks like: https://docs.snowflake.com/en/sql-reference/sql/use-secondary-roles

- Make secondary roles default as ALL, so every roles' privileges take effect, when you create a database, it's owned by the CURRENT ROLE
- If the secondary roles are NONE, only the privileges of CURRENT ROLE take effect

guojidan:
assign to me : )

BohuTANG:
@guojidan Sorry this not a general issue, it's a complex refactor, we already working on it.

guojidan:
> @guojidan Sorry this not a general issue, it's a complex refactor, we already working on it.

ok, I will choose another one, Do you have any suggestions

BohuTANG:
@guojidan https://github.com/datafuselabs/databend/labels/good%20first%20issue Feel free to discuss and taking :)

PR #13665, opened 2023-11-10
fix(rbac): revert the session role response behaviour in #13062 

flaneur2020:

## Summary

the role response behaviour when combined with #13537, may have a surprised side effect:

1. when user NOT set the restricted role with `session.role`, the http response still returns the current `session.role`
2. the client take the `session.role` returnted from the server, and pass it to the next query
3. the next query get restricted role by `public`, which causes permission denied.

this issue should be finally resolved by the `SECONDARY ROLE` seperation described in https://github.com/datafuselabs/databend/issues/13544

this PR revert the behaviour back before #13062 as a hotfix.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13665)

### DIFF
--- a/src/query/service/src/servers/http/v1/query/http_query.rs
+++ b/src/query/service/src/servers/http/v1/query/http_query.rs
@@ -446,10 +446,18 @@ impl HttpQuery {
             .map(|v| v.keep_server_session_secs)
             .unwrap_or(None);
 
-        // TODO: add current role here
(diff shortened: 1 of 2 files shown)


TCeason (approved):

flaneur2020:
@BohuTANG can you help take a review & merge about this PR? it's intended to hotfix an user's feedback 🙏

PR #13702, opened 2023-11-13
chore(query): fix left anti with conjunct if block size is small

### LINKED ISSUE #13697: bug: left anti join with conjunct wrong result if block size if small
**Summary**

set max_block_size = 1;
```sql
drop table if exists onecolumn;
CREATE TABLE onecolumn (x INT NULL);
INSERT INTO onecolumn(x) VALUES (44), (NULL), (42), (42);
drop table if exists twocolumn;
CREATE TABLE twocolumn (x INT NULL, y INT NULL);
INSERT INTO twocolumn(x, y) VALUES (44,51), (NULL,52), (42,53), (45,45);
select * from onecolumn as a right anti join twocolumn as b on a.x = b.x and a.x > 42 order by b.x;
42 53
45 45
NULL 52
```

Dousir9:

## Summary

Summary about this PR

- Closes #13697

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13702)


Dousir9:
It looks like ci panic is not related to this PR.

tests/sqllogictests/suites/query/join_small_block_size.test
```diff
@@ -718,5 +718,30 @@ drop table t1;
...
+drop table if exists twocolumn;
+
+statement ok
+CREATE TABLE twocolumn (x INT NULL, y INT NULL);
+
+statement ok
+INSERT INTO twocolumn(x, y) VALUES (44,51), (NULL,52), (42,53), (45,45);
+
+query I
+select * from onecolumn as a right anti join twocolumn as b on a.x = b.x and a.x > 42 order by b.x;
```
xudong963, line 740: The test in main has same result as the PR.

```
mysql> select * from onecolumn as a right anti join twocolumn as b on a.x = b.x and a.x > 42 order by b.x;
+------+------+
| x    | y    |
+------+------+
|   42 |   53 |
|   45 |   45 |
| NULL |   52 |
```
So is the test  useless for the PR?

Dousir9: The max block size here is 1.

xudong963: Btw, the test is right-anti-join (I know the real plan will be left-anti-join due to join order), but the tile is fix left-anti, so it's a little confusing.

You can use `select * from twocolumn as b left anti join onecolumn as a on a.x = b.x and a.x > 42 order by b.x;`  to make it clear.

xudong963: > The max block size here is 1.

oh, got

Dousir9: > Btw, the test is right-anti-join (I know the real plan will be left-anti-join due to join order), but the tile is fix left-anti, so it's a little confusing.
> 
> You can use `select * from twocolumn as b left anti join onecolumn as a on a.x = b.x and a.x > 42 order by b.x;` to make it clear.

This test is only for this PR, I will organize them in https://github.com/datafuselabs/databend/pull/13695.

xudong963 (approved):

PR #13734, opened 2023-11-16
fix: fix push down constant filter to join

xudong963:

## Summary

For constant filter from `where` operator which doesn't contain a column, such as `'year' == ''`, `false`, directly push down both sides of join

**Issue from our customer**

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13734)


Dousir9 (approved):

src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_filter_join.rs
```diff
@@ -139,35 +137,13 @@ pub fn try_push_down_filter_join(
     let mut need_push = false;
 
     for predicate in predicates.into_iter() {
-        if is_falsy(&predicate) {
```
Dousir9, line 142: I think we need `is_falsy` and `false_constant()`, if true, we can push `false_constant()` down, otherwise do nothing.

Issue #13753, opened 2023-11-17
read remote parquet file `Result::unwrap()` on an `Err` value: Overflow


BohuTANG:
**Summary**

`3_5M-GPT3_5-Augmented.parquet [3.09 GB]`

```
select * from 
'https://huggingface.co/datasets/Open-Orca/OpenOrca/resolve/main/3_5M-GPT3_5-Augmented.parquet'
limit 1;
```

But `set use_parquet2=0;` is working.

BohuTANG:
cc @youngsofun

PR #13754, opened 2023-11-17
feat: change default value of use_parquet2 to 0.

youngsofun:

## Summary

we do not hope user to use this option unless nessary.

now only affect copy from parquet and select parquet.
later when other component mig to parquet rs,  may resue this setting with diff meaning,  or use a more specific setting name?

if any one feel there are some risk to use parquet_rs as default？we can stop and test and review more.
@b41sh @sundy-li about the data type
@RinChanNOWWW @Dousir9 about the read progress

- Closes https://github.com/datafuselabs/databend/issues/13753

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13754)


src/query/settings/src/settings_default.rs
```diff
@@ -404,7 +404,7 @@ impl DefaultSettings {
                     display_in_show_settings: true,
                 }),
                 ("use_parquet2", DefaultSettingValue {
-                    value: UserSettingValue::UInt64(1),
+                    value: UserSettingValue::UInt64(0),
```
Dousir9, line 407: One point we need to note is that currently the performance of parquet-rs is not as good as parquet2.

BohuTANG: It's very closed now and accecptable.
parquet2 is hard to maintain, this is my benchamrk today:
use parquet2=1;
img_v3_0259_4ed80219-ecce-4c55-9b25-b6b12c82cf5g

use parquet2=0:
img_v3_0259_02262d71-5019-49f5-8dfb-179410f20b3g

Dousir9:
I downloaded the file and located the problem: 
panicked at src/common/arrow/src/arrow/io/parquet/read/deserialize/binary/utils.rs:73:46: called `Result::unwrap()` on an `Err` value: Overflow

Dousir9:
> I downloaded the parquet and located the problem: panicked at src/common/arrow/src/arrow/io/parquet/read/deserialize/binary/utils.rs:73:46: called `Result::unwrap()` on an `Err` value: Overflow

Dousir9:
# 2 VCPU 13.0GB
## parquet2:
截屏2023-11-17 14 20 29

## parquet-rs:
截屏2023-11-17 14 36 48

BohuTANG:
> # 2 vCPU 13.0GB
> ## use parquet2=1:
> 截屏2023-11-17 14 20 29
> ## use parquet2=0:
> 截屏2023-11-17 14 36 48

Looks parquet_rs is still bad in poor hardware.

BohuTANG:
Let's draft this PR.

sundy-li:
> looks parquet_rs is still bad in poor hardware.


I think correctness is better than performance,  parquet2 lacked much features or bugs  which were hard to fix.

PsiACE:
> I think correctness is high priority than performance, parquet2 lacked much features or bugs which were hard to fix.

I agree with this point. In addition, with the further maturity of the Arrow Rust ecosystem, it can be expected that the performance of Parquet will improve, while the maintainability of Parquet2 is almost no longer guaranteed.

BohuTANG:
@sundy-li agreed, I understand your situation. let's continue this PR and keep the use_parquet setting.

youngsofun:
unloading is using parquet2,
parquet_rs reader does not recognize variant types in unloaded parquet.
this is not expected since they are using the same mechanism. 
cc @b41sh

b41sh:
> unloading is using parquet2, parquet_rs reader does not recognize variant types in unloaded parquet. this is not expected since they are using the same mechanism. cc @b41sh

I'll follow up to fix this.

sundy-li (approved):

PR #13740, opened 2023-11-16
fix: flaky test case on test_affect

flaneur2020:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

- Closes #13736

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13740)

PR #13757, opened 2023-11-17
chore(test): use TestHttpQueryRequest on more test cases in http_query_handlers.rs

flaneur2020:

## Summary

this PR is a follow up after #13740 

as more features were added in the HTTP query protocol, the complexities also get increased a lot together in the test cases.

using a simplified mock client in these test cases may have the following benefits:

- unify the adhoc pagination logics among different test cases
- reduce the adhoc helper functions like `post_sql_to_endpoint_with_result_timeout` by a builder pattern.
- make the test cases deterministic without troubling with the timing issue like #13740

this PR still not transformed the entire http_query_handlers.rs yet, but we can make the migration in a phased manner, when adding new test cases, prefer using the unified mock test client instead of building adhoc HTTP requests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13757)


src/query/service/tests/it/servers/http/http_query_handlers.rs
```diff
@@ -69,54 +73,98 @@ use crate::tests::tls_constants::*;
...
         let mut resps = vec![];
 
         let (status, resp) = self.do_request(Method::POST, "/v1/query").await?;
-        let mut next_uri = resp.next_uri.clone();
-        resps.push((status, resp.clone()));
+        self.next_uri = resp.as_ref().and_then(|r| r.next_uri.clone());
+        resps.push((status, resp.clone().unwrap()));
 
-        while next_uri.is_some() {
+        while self.next_uri.is_some() {
```
youngsofun, line 144: do we need a timeout to avoid loop forever?

youngsofun (approved):

PR #13763, opened 2023-11-19
feat: add object cache for stage parquet file 

### LINKED ISSUE #12762: feat: add object cache for stage parquet file
**Summary**

Now, the Databend object cache only supports the FUSE engine table. This query cannot be accelerated:
```
select * from 's3://aa/bb/cc/' (pattern => '.*.parquet')
```

However, we can enable it for the stage files with the parquet format.

zenus:

## Summary

Summary about this PR

- Closes  #12762

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2972,6 +2972,8 @@ dependencies = [
  "opendal",
  "parquet",
  "serde",
+ "storages-common-cache",
+ "storages-common-cache-manager",
  "storages-common-index",
  "storages-common-pruner",
  "storages-common-table-meta",
--- a/src/query/storages/parquet/Cargo.toml
+++ b/src/query/storages/parquet/Cargo.toml
@@ -23,8 +23,10 @@ common-metrics = { path = "../../../common/metrics" }
 common-pipeline-core = { path = "../../pipeline/core" }
 common-pipeline-sources = { path = "../../pipeline/sources" }
 common-storage = { path = "../../../common/storage" }
+storages-common-cache = { path = "../common/cache" }
 storages-common-index = { path = "../common/index" }
 storages-common-pruner = { path = "../common/pruner" }
+storages-common-cache-manager = { path = "../common/cache_manager" }
 storages-common-table-meta = { path = "../common/table_meta" }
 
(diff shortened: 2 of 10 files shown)


BohuTANG:
There is case we should consider:
```
COPY INTO table FROM (SELECT * FROM @stage/parquet)
```
This case we can not do cache, does this PR affect this?

zenus:
@BohuTANG I'm afraid I didn't consider this situation, let me fix it.

src/query/storages/parquet/src/parquet_rs/parquet_table/meta.rs
```diff
@@ -39,6 +39,7 @@ pub async fn read_metas_in_parallel(
     leaf_fields: Arc<Vec<TableField>>,
     num_threads: usize,
     max_memory_usage: u64,
```
BohuTANG, line 41: This doesn't work:)
 We have many write query like:copy/replace/insert/merge into from (select * from @stage/parquet), they are all can read stage parquet files.
For mostly users, write is one query and read is another, so if we cache in the write side, there is no use.

This PR aims to fast read when read a remote parquet many times, but it's not a normal case in facts.

src/query/storages/parquet/src/parquet_rs/parquet_table/partition.rs
```diff
@@ -238,7 +239,11 @@ impl ParquetRSTable {
             let copy_status = copy_status.clone();
             let leaf_fields = self.leaf_fields.clone();
             let topk = topk.clone();
-            let is_copy = matches!(ctx.get_query_kind(), QueryKind::CopyIntoTable);
+            let is_remote_query =
```
BohuTANG, line 242: This does not work: COPY/REPLACE etc. can also read the parquet file from the stage, it's hard to check a query is read only. 
So we should lower this PR's priority if a better method is found.
Thanks.

dantengsky:
Thanks for the work on this PR and for following up on the review feedback. After revisiting the related issue, we no longer see a compelling need for the proposed functionality. The implementation has also diverged substantially from main and would require significant rework. We’re closing this PR and the related issue.

PR #13765, opened 2023-11-20
feat: add downcast_owned_builder

ariesdevil:

## Summary

Add `try_upcast_column_builder` and `try_downcast_owned_builder` for ValueType, so the types like `NullableType` and `ArrayType` can downcast builder.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13765)

### DIFF
--- a/src/query/expression/src/types/decimal.rs
+++ b/src/query/expression/src/types/decimal.rs
@@ -89,6 +89,14 @@ impl<Num: Decimal> ValueType for DecimalType<Num> {
         Num::try_downcast_builder(builder)
     }
 
+    fn try_downcast_owned_builder(builder: ColumnBuilder) -> Option<Self::ColumnBuilder> {
+        Num::try_downcast_owned_builder(builder)
+    }
+
+    fn try_upcast_column_builder(_builder: Self::ColumnBuilder) -> Option<ColumnBuilder> {
+        None
+    }
+
     fn upcast_scalar(scalar: Self::Scalar) -> Scalar {
         Num::upcast_scalar(scalar, Num::default_decimal_size())
     }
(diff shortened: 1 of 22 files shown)


sundy-li (approved):

src/query/expression/src/types/decimal.rs
```diff
@@ -89,6 +89,14 @@ impl<Num: Decimal> ValueType for DecimalType<Num> {
         Num::try_downcast_builder(builder)
     }
 
+    fn try_downcast_owned_builder(builder: ColumnBuilder) -> Option<Self::ColumnBuilder> {
+        Num::try_downcast_owned_builder(builder)
+    }
+
+    fn try_upcast_column_builder(_builder: Self::ColumnBuilder) -> Option<ColumnBuilder> {
+        None
```
andylokandy, line 97: Why no to implement it?

src/query/functions/src/aggregates/aggregator_common.rs
```diff
@@ -167,3 +168,24 @@ pub fn deserialize_state<T: serde::de::DeserializeOwned>(slice: &mut &[u8]) -> R
...
+        let closure_result = panic::catch_unwind(panic::AssertUnwindSafe(|| closure(old_t)));
+
+        match closure_result {
+            Ok(Ok(new_t)) => {
+                ptr::write(mut_ref, new_t);
+                Ok(())
+            }
+            Ok(Err(e)) => Err(e),
+            Err(_) => ::std::process::abort(),
+        }
```
andylokandy, line 189: Why not to use `take_mut`?

ariesdevil: We need return `Result<T>` in closure, so I copy the code and modify it.

andylokandy: fair enough. so common-base may be a better place for it.

src/query/expression/src/types/decimal.rs
ariesdevil: I'm trying to refactor it, but there are huge changes that need to be made, so for now, leave it for future refactor

src/query/functions/src/aggregates/aggregator_common.rs
ariesdevil: Let's move it to common-base

src/query/expression/src/types/number.rs
```diff
@@ -908,6 +956,17 @@ impl Number for u64 {
         builder.as_u_int64_mut()
     }
 
+    fn try_downcast_owned_builder(builder: NumberColumnBuilder) -> Option<Vec<Self>> {
+        match builder {
+            NumberColumnBuilder::UInt64(b) => Some(b),
+            _ => unreachable!(),
```
andylokandy, line 962: Why not return None here?

ariesdevil: Yup, should be None

PR #13770, opened 2023-11-20
feat: limit max copied files to 15000.

youngsofun:

## Summary

committing too many `copied_files` once will lead to 

1.  error when package size exceeds some limit.
2.  performance problem with meta.

before a better solution is found. we can limit the number of files copied.
report error earlier is much better than got error after a long time when data are already copied, just not commited

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13770)

### DIFF
--- a/src/meta/app/src/principal/user_stage.rs
+++ b/src/meta/app/src/principal/user_stage.rs
@@ -57,6 +57,7 @@ use crate::storage::StorageParams;
 // SIZE_LIMIT = <num>
 
 pub const COPIED_FILES_MAX_COMMIT_NUM: usize = 15000;
+pub const COPY_WITH_MAX_FILE_MSG: &str = "copy into table can commit at most 15000 paths once. Please use CopyOption 'max_files=<num>' to limit the number of files and run multi times until result is empty or use 'force=true, purge=true' to avoid committing paths if acceptable.";
 
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
 pub enum StageType {
--- a/src/meta/app/src/principal/user_stage.rs
+++ b/src/meta/app/src/principal/user_stage.rs
@@ -56,8 +56,12 @@ use crate::storage::StorageParams;
 // ON_ERROR = { CONTINUE | SKIP_FILE | SKIP_FILE_<num> | SKIP_FILE_<num>% | ABORT_STATEMENT }
 // SIZE_LIMIT = <num>
 
-pub const COPIED_FILES_MAX_COMMIT_NUM: usize = 15000;
-pub const COPY_WITH_MAX_FILE_MSG: &str = "copy into table can commit at most 15000 paths once. Please use CopyOption 'max_files=<num>' to limit the number of files and run multi times until result is empty or use 'force=true, purge=true' to avoid committing paths if acceptable.";
+/// Maximum files per 'copy into table' commit.
+pub const COPY_MAX_FILES_PER_COMMIT: usize = 15000;
+
+/// Instruction for exceeding 'copy into table' file limit.
+pub const COPY_MAX_FILES_COMMIT_MSG: &str = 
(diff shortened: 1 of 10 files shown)


drmingdrmer (approved):

src/query/ast/src/ast/statements/copy.rs
```diff
@@ -117,6 +122,37 @@ impl CopyIntoTableStmt {
...
+
+        if self.max_files != 0 {
+            copy_options.max_files = self.max_files;
+        }
+
+        if !(copy_options.purge && self.force)
+            && copy_options.max_files > COPIED_FILES_MAX_COMMIT_NUM
+        {
+            return Err(ErrorCode::InvalidArgument(format!(
+                "max_files {} is too large, max_files should be less than {COPIED_FILES_MAX_COMMIT_NUM}",
```
BohuTANG, line 150: We can give a kindly error hints, tell user to copy with max_files option?

Xuanwo (approved):

src/query/ast/src/ast/statements/copy.rs
youngsofun: here user is using `max_files`, but too large, maybe we can further suggest user to copy mutli times?.

`tell user to copy with max_files option` is already done when max_files=0 (no limit)

youngsofun: > here user is using `max_files`, but too large, maybe we can further suggest user to copy mutli times?.

since user are already using  `max_files`, he should have known that.

BohuTANG: Yes，if the COPIED_FILES_MAX_COMMIT_NUM is 15000，if user COPY from a 150000 files stage, we can tell him you should COPY ... MAX_FILES=15000 many times.

youngsofun: done

src/meta/app/src/principal/user_stage.rs
```diff
@@ -56,6 +56,9 @@ use crate::storage::StorageParams;
 // ON_ERROR = { CONTINUE | SKIP_FILE | SKIP_FILE_<num> | SKIP_FILE_<num>% | ABORT_STATEMENT }
 // SIZE_LIMIT = <num>
 
+pub const COPIED_FILES_MAX_COMMIT_NUM: usize = 15000;
+pub const COPY_WITH_MAX_FILE_MSG: &str = "copy into table can commit at most 15000 paths once. Please use CopyOption 'max_files=<num>' to limit the number of files and run multi times until result is empty or use 'force=true, purge=true' to avoid committing paths if acceptable.";
```
BohuTANG, line 60: Only need `purge=true`?

BohuTANG, line 60:
```suggestion
/// Maximum files per 'copy into table' commit.
pub const COPY_MAX_FILES_PER_COMMIT: usize = 15000;

/// Instruction for exceeding 'copy into table' file limit.
pub const COPY_MAX_FILES_COMMIT_MSG: &str = 
    "Limit for 'copy into table': 15,000 files per commit. To handle more files, adjust 'CopyOption' with 'max_files=<num>' and perform several operations until all files are processed.";
```

youngsofun: we used to require `force=true, purge=true`. 
https://github.com/datafuselabs/databend/pull/11981

do you mean filter but not commit when `purge=true ` only.   also makes sense.
I think this is a feature that we can define. better if user have more choices?

`purge=true` is not reliable since delete may fail (and we do not rollback in this case not)?

BohuTANG: `force=true, purge=true` this is not a security way let the user to do, if the purge failed, next time the copy may double the datas.

youngsofun: Limit for 'copy into table' `without specifing max_files` ... ?

youngsofun: what about add another  options `commit_paths = true|false`, which is easier to understand?

BohuTANG:
Some tests failed, need to fix.

PR #13774, opened 2023-11-21
fix(query): fix arrow_rs read variant type

### LINKED ISSUE #13759: Feat: uploading support parquet_rs
> unloading is using parquet2, parquet_rs reader does not recognize variant types in unloaded parquet. this is not expected since they are using the same mechanism. cc @b41sh

I'll follow up to fix this.

_Originally posted by @b41sh in https://github.com/datafuselabs/databend/issues/13754#issuecomment-1817546362_

b41sh:

## Summary

Fix `arrow_rs` read variant type, the schema in `RecordBatch` doesn't contain extended metadata, so we can't determine the type of `Variant` and `Bitmap`. Use an external schema instead to fix this bug.

- Closes #13759

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13774)

### DIFF
--- a/src/query/storages/parquet/src/parquet_rs/parquet_reader/utils.rs
+++ b/src/query/storages/parquet/src/parquet_rs/parquet_reader/utils.rs
@@ -15,13 +15,14 @@
 use arrow_array::BooleanArray;
 use arrow_array::RecordBatch;
 use arrow_array::StructArray;
-use arrow_schema::FieldRef;
 use common_arrow::arrow::array::Arrow2Arrow;
 use common_arrow::arrow::bitmap::Bitmap;
 use common_exception::ErrorCode;
 use common_exception::Result;
 use common_expression::Column;
 use common_expression::DataBlock;
+use common_expression::DataField;
+use common_expression::DataSchema;
 use common_expression::FieldIndex;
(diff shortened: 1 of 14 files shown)


youngsofun (approved):

src/query/storages/parquet/src/parquet_rs/parquet_reader/utils.rs
```diff
@@ -119,7 +121,7 @@ pub fn compute_output_field_paths(
     if !inner_projection {
         return Ok(None);
     }
-    let expected_schema = to_arrow_schema(expected_schema);
+    let expected_schema = DataSchema::from(expected_schema);
```
youngsofun, line 124: @b41sh 
so we infact get the right schema here,  what this pr does is to pass it to needed place.
am I right?

sundy-li (approved):

src/query/storages/parquet/src/parquet_rs/parquet_reader/utils.rs
b41sh: yes, the schema in `RecordBatch` don't have extension type

PR #13795, opened 2023-11-24
chore: add kill reason to http handler

BohuTANG:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13795)

### DIFF
--- a/src/query/service/src/servers/http/v1/http_query_handlers.rs
+++ b/src/query/service/src/servers/http/v1/http_query_handlers.rs
@@ -267,7 +267,7 @@ async fn query_cancel_handler(
         let http_query_manager = HttpQueryManager::instance();
         match http_query_manager.get_query(&query_id).await {
             Some(query) => {
-                query.kill().await;
+                query.kill("http query cancel by handler").await;
                 http_query_manager.remove_query(&query_id).await;
                 StatusCode::OK
             }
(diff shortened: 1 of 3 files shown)


src/query/service/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -267,7 +267,7 @@ async fn query_cancel_handler(
         let http_query_manager = HttpQueryManager::instance();
         match http_query_manager.get_query(&query_id).await {
             Some(query) => {
-                query.kill().await;
+                query.kill("http query cancel by handler").await;
                 http_query_manager.remove_query(&query_id).await;
                 StatusCode::OK
```
BohuTANG, line 272: What's the error will return to the client after killed? @youngsofun

src/query/service/src/servers/http/v1/query/http_query_manager.rs
```diff
@@ -91,7 +91,7 @@ impl HttpQueryManager {
                             warn!("{msg}, but fail to remove");
                         } else {
                             warn!("{msg}");
-                            query.kill().await;
+                            query.kill(msg).await;
```
BohuTANG, line 94: Same as the above.

src/query/service/src/servers/http/v1/http_query_handlers.rs
youngsofun: before this pr,  info will report 404 as since the query id can not be found .
to report timeout to client. we can retain the ids only for a longer time or with a limited count.

youngsofun (approved):

Issue #13752, opened 2023-11-17
not implemented: data_type: Extension("UUID filename of generated images.", Utf8, None)


BohuTANG:
**Summary**

```
SELECT * FROM 'https://huggingface.co/datasets/poloclub/diffusiondb/resolve/main/metadata-large.parquet' limit 1;
```

PR #13796, opened 2023-11-24
chore: add arrow extention data type support

lichuang:

## Summary

chore: add arrow extention data type support

- Closes https://github.com/datafuselabs/databend/issues/13752

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13796)

### DIFF
--- a/src/query/expression/src/values.rs
+++ b/src/query/expression/src/values.rs
@@ -1251,15 +1251,16 @@ impl Column {
         }
     }
 
-    pub fn from_arrow(
+    pub fn from_arrow_by_array_type(
         arrow_col: &dyn common_arrow::arrow::array::Array,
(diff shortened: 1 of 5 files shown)


src/query/expression/src/values.rs
```diff
@@ -1251,15 +1250,14 @@ impl Column {
         }
     }
 
-    pub fn from_arrow(
+    pub fn from_arrow_by_array_type(
```
andylokandy, line 1253:
```suggestion
    pub fn from_arrow_with_array_type(
```

andylokandy: It's better to be moved into `pub fn from_arrow` to avoid any other code using it directly.

PR #13807, opened 2023-11-25
fix(query): remove unexpected strong ref for http query

zhang2014:

## Summary

fix(query): remove unexpected strong ref for http query

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13807)


src/query/service/src/servers/http/v1/query/http_query_manager.rs
```diff
@@ -78,20 +78,30 @@ impl HttpQueryManager {
 
         let self_clone = self.clone();
         let query_id_clone = query_id.to_string();
-        let query_clone = query.clone();
+        let http_query_weak = Arc::downgrade(&query);
```
BohuTANG, line 81: Better to add some comments here.

youngsofun (approved):

PR #13821, opened 2023-11-27
fix(query): external location no need to check priv

TCeason:

## Summary

```sql
select $1 from 'fs:///xx/yy'

copy into t from 'fs:///xx/yy'
```

Not check remote file priv.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13821)


flaneur2020 (approved):

src/meta/app/src/principal/user_stage.rs
```diff
@@ -555,6 +555,7 @@ pub struct StageInfo {
     pub stage_name: String,
     pub stage_type: StageType,
     pub stage_params: StageParams,
+    pub from_uri: bool,
```
BohuTANG, line 558: It seems we no need add the from_uri, we can check it by add a new method for the StageInfo:
impl StageInfo {
  pub fn uri_kind() -> URIKind{}
}

src/meta/protos/proto/stage.proto
```diff
@@ -74,6 +74,8 @@ message StageInfo {
   optional UserIdentity creator = 8;
 
   FileFormatParams file_format_params = 9;
+
+  bool from_uri = 10;
```
Xuanwo, line 78: Maybe we don't need to add this in proto since it will always be `false`?

TCeason: I think you are right.

drmingdrmer (commented):
> Reviewed 6 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13821)* status: 1 of 34 files reviewed, 3 unresolved discussions (waiting on @BohuTANG, @TCeason, @Xuanwo, and @youngsofun)

*[a discussion](https://reviewable.io/reviews/datafuselabs/databend/13821#-NkG1Anq2pgV1sVEVcDC:-NkG1Anq2pgV1sVEVcDD:b-1k31vm) (no related file):*
Does this PR involve any protobuf file changes? no?

___
___
*[`src/meta/app/src/principal/user_stage.rs` line 558 at r1](https://reviewable.io/reviews/datafuselabs/databend/13821#gh-1405973576:-NkFeqCT47O25lQSM20-:b-r35it5) ([raw file](https://github.com/datafuselabs/databend/blob/8cf1c4effe9d4947cfe9180b5602df24313f821b/src/meta/app/src/principal/user_stage.rs#L558)):*
<details><summary><i>Previously, TCeason wrote…</i></summary><blockquote>

We need this flag. Because now, uri convert a external stage.

https://github.com/datafuselabs/databend/blob/548ddcfc3984348d3154edf3e4f65f486c185482/src/query/sql/src/planner/binder/copy_into_table.rs#L556-L575
</blockquote></details>

A bool value should be in form `is_from_uri`. `from_uri` looks like "the source uri from which something is done"

TCeason:
> Does this PR involve any protobuf file changes? no?

I revert it. So not involve any protobuf now.

> A bool value should be in form is_from_uri

Good idea.

TCeason:
@Xuanwo @youngsofun @BohuTANG Please review this pr, thx.

youngsofun:
1.  we do not need to store the flag to meta.
2. prefer `is temporary` to `is_from_uri`,  because we can create an external stage from a uri.

drmingdrmer (approved):
> Reviewed 2 of 16 files at r1, 1 of 20 files at r3, 11 of 11 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13821)* status: 14 of 38 files reviewed, 3 unresolved discussions (waiting on @BohuTANG, @TCeason, @Xuanwo, and @youngsofun)

PR #13842, opened 2023-11-29
refactor: runtime filter

xudong963:

## Summary

**Intro**:
Adaptive derivation of new predicates at runtime is used to filter the join probe side to improve performance.

New predicates generated at runtime are pushed down through the processor to the table scan on the probe side for prune, thus improving performance significantly.

**Simple benchmark**
A simple example that is perfect for runtime filtering
```
select * from t1 join t2 on t1.a = t2.b;
t1: 1_000_000_000
t2: 9999 (inlist runtime filter will not be generated for >10k)
```
cluster:
```
before: 0.04 sec
runtime filter: 0.007 sec. 
```

single node:
```
before: 0.025s
runtime filter: 0.0046 s
```

**Adaptive**:
1. Currently only inlist filter is generated, and only if the total data size of the build is less than 10k. If the inlist is too heavy, it is not efficient, so we may consider supporting bloom filter for data larger than 10k in the future.
2. For clusters, only broadcast join supports runtime filter, because the data size gap between the build and probe sides of broadcast join is relatively large, and better filtering may be achieved.

**Others**:
Runtime filter will be saved into QueryCtx by `HashMap`, key is the table index, and values are filters for the table. Table will get the corresponding filters from ctx by table index to prune before it starts to read data.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13842)


src/query/service/src/pipelines/executor/executor_graph.rs
```diff
@@ -274,6 +278,22 @@ impl ExecutingGraph {
             }
 
             if let Some(schedule_index) = need_schedule_nodes.pop_front() {
+                let current_node = locker.graph[schedule_index].clone();
+                let rt_filters = current_node.processor.get_runtime_filters()?;
```
BohuTANG, line 282: Why add runtime filter logic to the exectuor? Can we add it in the pipeline build phase?

Dousir9:
How to generate the test data of t1 and t2, is it number(1000_000_000)?

src/query/pipeline/core/src/processors/duplicate_processor.rs
```diff
@@ -48,6 +49,8 @@ impl DuplicateProcessor {
     }
 }
 
+impl RuntimeFilter for DuplicateProcessor {}
```
zhang2014, line 52: maybe we can `impl<T:Processor> RuntimeFilter for T`?

src/query/storages/fuse/src/operations/read/runtime_filter_prunner.rs
```diff
@@ -0,0 +1,84 @@
...
+use storages_common_index::statistics_to_domain;
+
+use crate::FusePartInfo;
+
+pub fn runtime_filter_pruner(
+    table_schema: Arc<TableSchema>,
+    part: &PartInfoPtr,
+    filters: &Vec<Expr<String>>,
+    func_ctx: &FunctionContext,
+) -> Result<bool> {
```
BohuTANG, line 36: Do we need adding the runtime filter stats to explain? Now the stats:
```
├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 755 to 755, bloom pruning: 0 to 0>]
```

xudong963: current stats are collected before the pipeline runs, but runtime filter stats will be generated during the pipeline running. Maybe we can try to add runtime filter stats to query profile later.

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_build_state.rs
```diff
@@ -220,6 +220,15 @@ impl HashJoinBuildState {
                 }
             }
 
+            let build_chunks =
```
Dousir9, line 223: The chunks here is not projected, we can collect build_chunks in `build_done`.

src/query/service/src/pipelines/processors/transforms/hash_join/util.rs
```diff
@@ -37,3 +46,85 @@ pub(crate) fn probe_schema_wrap_nullable(probe_schema: &DataSchemaRef) -> DataSc
...
     DataSchemaRefExt::create(nullable_field)
 }
+
+// Construct inlist runtime filter
+pub(crate) fn inlist_filter(
+    func_ctx: &FunctionContext,
+    build_key: &Expr,
+    probe_key: &Expr<String>,
+    build_blocks: &[DataBlock],
+) -> Result<Option<Vec<Expr<String>>>> {
```
sundy-li, line 56:
```
Result<Option<Expr<String>>>
```
is enough?

src/query/storages/fuse/src/operations/read/runtime_filter_prunner.rs
```diff
@@ -0,0 +1,84 @@
...
+use common_expression::FunctionContext;
+use common_expression::Scalar;
+use common_expression::TableSchema;
+use common_functions::BUILTIN_FUNCTIONS;
+use log::info;
+use storages_common_index::statistics_to_domain;
+
+use crate::FusePartInfo;
+
+pub fn runtime_filter_pruner(
```
sundy-li, line 31: This should not be in part's level, because the part is lazy and its `columns_stat` is not accurate enough.

I suggest to apply this in `FusePruner` ( `SegmentPruner` and `block_pruner`)

sundy-li: I mis understand.

sundy-li (approved):

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_build_state.rs
Dousir9, line 223: We can first check if `build_num_rows` is less than 10k and then decide whether to store datablocks into `build_chunks`,

xudong963: For `build_num_rows` is more than 10k, we also want to generate filter, such as bloom filter

xudong963: We can add a check if join is broadcast join, usually, if broadcast join, build side is small

Dousir9: > For `build_num_rows` is more than 10k, we also want to generate filter, such as bloom filter

But now we do not collect information during finalize, and we only support inlist filter now.

Dousir9: > For `build_num_rows` is more than 10k, we also want to generate filter, such as bloom filter

If we collect information during finalize, we can later remove these checks.

Dousir9:
rest LGTM !

Dousir9 (approved):

PR #13845, opened 2023-11-29
fix(query): add some dml prvilege check

### LINKED ISSUE #13727: bug: insert into select need to allow privilege check
### Search before asking

### Version

main

### What's Wrong?

insert into select need to allow privilege check

### How to Reproduce?

```sql
MySQL [(none)]> show grants for a;
+-------------------------------------------------------------+
| Grants                                                      |
+-------------------------------------------------------------+
| GRANT INSERT,DELETE ON 'default'.'default'.'t1' TO 'a'@'%'  |
| GRANT SUPER ON *.* TO 'a'@'%'                               |
| GRANT DELETE ON 'default'.'default'.'t2' TO 'a'@'%'         |
| GRANT ALL ON UDF test_alter_udf TO 'a'@'%'                  |
| GRANT SELECT ON 'default'.'system'.'one' TO 'a'@'%'         |
| GRANT SELECT ON 'default'.'information_schema'.* TO 'a'@'%' |
+-------------------------------------------------------------+

```

user a connection:

```sql
MySQL [(none)]> insert into t1 select * from t2;
Query OK, 0 rows affected (0.503 sec)

MySQL [(none)]>  select * from t2;
ERROR 1105 (HY000): PermissionDenied. Code: 1063, Text = Permission denied, privilege [Select] is required on 'default'.'default'.'t2' for user 'a'@'%' with roles [public].
```

TCeason:

## Summary

Added missing permission checks:

1. insert into table `query`
2. replace table on `query`
3. create table as `select query`
4. copy into table from `query`

The query needs to be checked. If there is no permission, the statement needs to report an error.

After the corresponding permission is added, the command can be executed normally.

This pr is the improvement of the permission check and will not cause breaking

- Closes #13727

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13845)

### DIFF
--- a/src/meta/app/src/principal/user_stage.rs
+++ b/src/meta/app/src/principal/user_stage.rs
@@ -561,7 +561,8 @@ pub struct StageInfo {
     pub stage_name: String,
     pub stage_type: StageType,
     pub stage_params: StageParams,
-    pub is_from_uri: bool,
+    // on `COPY INTO xx FROM 's3://xxx?ak=?&sk=?'`, the URL(ExternalLocation) will be treated as an temporary stage.
+    pub is_temporary: bool,
     pub file_format_params: FileFormatParams,
     pub copy_options: CopyOptions,
     pub comment: String,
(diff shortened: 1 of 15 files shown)


TCeason:
wait https://github.com/datafuselabs/databend/pull/13852 merged.

src/meta/app/src/principal/user_stage.rs
```diff
@@ -561,7 +561,7 @@ pub struct StageInfo {
     pub stage_name: String,
     pub stage_type: StageType,
     pub stage_params: StageParams,
-    pub is_from_uri: bool,
+    pub is_temporary: bool,
```
flaneur2020, line 564:
```suggestion
    // on `COPY INTO xx FROM 's3://xxx?ak=?&sk=?'`, the URL will be treated as an temporary stage.
    pub is_temporary: bool,
```

how about adding a line of comment about the semantics of this field?

TCeason: Done

PR #13846, opened 2023-11-29
feat(query): new filter execution framework

Dousir9:

### The implementation Databend filter in the past

In the past, the execution of Databend Filter was implemented by the `Evaluator`. For the given SQL: `select * from t where a = 3 and (b < 7.2 or b > 12.6) and c < 6;`, The `Evaluator` processes each predicate (`a = 3`, `b < 7.2`, `b > 12.6`, `c < 6`) and generates a bitmap for each. These bitmaps are then pairwise combined using `&` and `|` operations based on the `AND` and `OR`, resulting in a final bitmap called `filter`. Finally, the `Evaluator` uses the `filter` to invoke the `filter_boolean_value` of the DataBlock, generating the filtered DataBlock.

### The disadvantages of old implementation.

1. Frequent construction/destruction of bitmaps will lead to significant memory fragmentation: each comparison operator, such as `a = 3`, will generate a bitmap for result. If the column is nullable, an additional bitmap is generated for validity. In other words, for a where condition like `a = 3 and (b < 7.2 or b > 12.6) and c < 6`, `Evaluator` will generate up to 8 bitmaps during execution for single DataBlock.

2. The independent execution of filtering predicates can result in inefficiencies: If a certain row in a DataBlock has already been filtered out by one predicate, other predicates will still filter it again, leading to unnecessary CPU overhead.

3. using `filter bitmap` to invoke `filter_boolean_value` on the DataBlock to generate the filtered DataBlock may not always be optimal.

### New Filter Execution Framework

**We introduced a groundbreaking concept, defining it as the "Immutable Index". By combining the Immutable Index with the SelectStrategy, we have addressed the drawbacks of DuckDB Filter ! 🚀**, the Immutable Index enables us to avoid generating temporary selection buffer when encountering AND and OR operations. This not only reduces memory fragmentation but It also eliminates the cyclic copying from temporary selection to final selection.

1. New filter execution framework avoid memory allocation by using reusable `true_selection` and `false_selection` (only generating `false_selection` when OR predicates are present) instead of bitmaps..

2. The execution of predicates is dynamically linked through `true_selection` and `false_selection`, ensuring that each row in the DataBlock is filtered only once. This significantly optimizes performance (TPC-H Q12, Q18) and accommodates complex predicates effectively.

3. By employing a heuristic strategy and dynamically choosing between using `take` or `take_range` to generate the DataBlock, this approach is more efficient than using filter bitmap to invoke `filter_boolean_value` on the DataBlock.

### Benchmark
Q19 External Parquet: 14.5s -> 9.7s
before:

after:

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13846)

## Tests
- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


src/query/expression/src/values.rs
```diff
@@ -399,6 +399,42 @@ impl Scalar {
         }
     }
 
+    pub fn data_type(&self) -> DataType {
```
andylokandy, line 402: duplicated with `scalar.as_ref().infer_data_type()`

BohuTANG:
Also need the PR summary to Databend community :)
I think most wonder to know why and what will benfits from the new filter exection framwork.

src/query/expression/src/filter/filter_executor.rs
```diff
@@ -0,0 +1,158 @@
...
+            Ok(data_block)
+        } else if result_count as f64 > data_block.num_rows() as f64 * 0.8
+            && data_block.num_columns() > 1
+        {
+            let range_count = self.build_selection_range(result_count);
+            data_block.take_ranges(&self.selection_range[0..range_count], result_count)
+        } else {
+            // If has_or is true, the order of indices may be changed, so sorting is required only
+            // when has_or is true.
+            if self.keep_order && self.has_or {
```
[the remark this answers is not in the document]
Dousir9: Yeah.

src/query/service/src/pipelines/processors/transforms/transform_filter.rs
```diff
@@ -33,25 +33,36 @@ use crate::pipelines::processors::Processor;
...
+        max_block_size: usize,
     ) -> Box<dyn Processor> {
+        let filter = FilterExecutor::new(
+            select_expr,
+            func_ctx,
+            has_or,
+            max_block_size,
+            Some(projections.clone()),
+            &BUILTIN_FUNCTIONS,
+            false,
```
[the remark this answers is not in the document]
Dousir9: If `has_or` is false, but `keep_order` is true, we don't need to sort, If `has_or` is true, but `keep_order` is false, we also don't need to sort.

src/query/expression/src/filter/select_op.rs
```diff
@@ -0,0 +1,190 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use core::cmp::Ordering;
+
+#[derive(Clone, PartialEq, Debug)]
+pub enum SelectOp {
```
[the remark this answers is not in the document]
Dousir9: Yeah, we can do it later, is distinct from will introduce many code.

src/query/expression/src/types/string.rs
```diff
@@ -156,6 +158,47 @@ impl ValueType for StringType {
     fn column_memory_size(col: &Self::Column) -> usize {
         col.data().len() + col.offsets().len() * 8
     }
+
+    fn cmp(op: &SelectOp) -> fn(Self::ScalarRef<'_>, Self::ScalarRef<'_>) -> bool {
```
sundy-li, line 162: We just need to keep :

```
fn cmp(Self::ScalarRef<'_>, Self::ScalarRef<'_>) -> Ordering {..}
```
The latter functions could be removed, see `register_simple_domain_type_cmp`

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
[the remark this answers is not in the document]
Dousir9: the num_rows of prewhere_block is 1 :)

xudong963 (approved):

src/query/expression/src/types/boolean.rs
```diff
@@ -148,6 +151,52 @@ impl ValueType for BooleanType {
...
+            SelectOp::NotEqual => Self::not_equal,
+            SelectOp::Gt => Self::greater_than,
+            SelectOp::Gte => Self::greater_than_equal,
+            SelectOp::Lt => Self::less_than,
+            SelectOp::Lte => Self::less_than_equal,
+        }
+    }
+
+    #[inline(always)]
+    fn equal(left: Self::ScalarRef<'_>, right: Self::ScalarRef<'_>) -> bool {
```
sundy-li, line 172: these functions are redundant, they will fall back to call `Ord::cmp`, it's already handled in stdlib.

```
 #[stable(feature = "rust1", since = "1.0.0")]
    impl Ord for bool {
        #[inline]
        fn cmp(&self, other: &bool) -> Ordering {
            // Casting to i8's and converting the difference to an Ordering generates
            // more optimal assembly.
            // See <https://github.com/rust-lang/rust/issues/66780> for more info.
            match (*self as i8) - (*other as i8) {
                -1 => Less,
                0 => Equal,
                1 => Greater,
                // SAFETY: bool as i8 returns 0 or 1, so the difference can't be anything else
                _ => unsafe { unreachable_unchecked() },
            }
        }
```

src/query/expression/src/filter/select_value/select_column.rs
```diff
@@ -0,0 +1,370 @@
...
+        mutable_true_idx: &mut usize,
+        mutable_false_idx: &mut usize,
+        select_strategy: SelectStrategy,
+        count: usize,
+    ) -> Result<usize> {
+        let mut true_idx = *mutable_true_idx;
+        let mut false_idx = *mutable_false_idx;
+
+        match select_strategy {
+            SelectStrategy::True => unsafe {
```
sundy-li, line 48: SelectStrategy::true && SelectStrategy::false should be merged into one loop

src/query/expression/src/types.rs
```diff
@@ -344,6 +346,54 @@ pub trait ValueType: Debug + Clone + PartialEq + Sized + 'static {
...
+
+    // Return the comparison function for the given select operation, some data types not support comparison.
+    #[inline(always)]
+    fn compare_operation(_op: &SelectOp) -> fn(Self::ScalarRef<'_>, Self::ScalarRef<'_>) -> bool {
+        unreachable!()
+    }
+
+    // Equal comparison between two scalars, some data types not support it.
+    #[inline(always)]
+    fn equal(_left: Self::ScalarRef<'_>, _right: Self::ScalarRef<'_>) -> bool {
```
sundy-li, line 364: use default trait implementation instead of implementing them in a derived struct.

```
fn equal(_left: Self::ScalarRef<'_>, _right: Self::ScalarRef<'_>) -> bool {
   Self::compare(..).is_equal()
}
```

src/query/expression/src/types.rs
```diff
@@ -344,6 +346,54 @@ pub trait ValueType: Debug + Clone + PartialEq + Sized + 'static {
     fn column_memory_size(col: &Self::Column) -> usize {
         Self::column_len(col) * std::mem::size_of::<Self::Scalar>()
     }
+
+    // Compare two scalars and return the Ordering between them.
+    #[inline(always)]
+    fn compare(_: Self::ScalarRef<'_>, _: Self::ScalarRef<'_>) -> Ordering {
+        unreachable!()
```
sundy-li, line 353: it's dangerous to use `unreachable` , just use : 

```
fn compare(a: Self::ScalarRef<'_>, b: Self::ScalarRef<'_>) -> Ordering {
      a.partial_cmp(b).unwrap_or(Ordering::Equal)
}

since: type Scalar: Debug + Clone + PartialEq;
```

sundy-li (approved):

BohuTANG:
Conflicting files
src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs

PR #13847, opened 2023-11-29
fix: killing in-progress tasks causes query node to crash (Segment Fault)

dantengsky:

## Summary

**Problem trying to address**:

Killing tasks in progress, such as the `replace-into` statement, may cause the query node to crash (SegmentFault).

------

**How to reproduce:**

_Single-node deployment (local fs storage is sufficient)_

- **Test table**: `create table t as select * from numbers(10000000)`
- **Continuously execute**: `replace into t on(number) select * from t`
- **Simultaneously** & continuously use the `kill query $QUERY_ID` command to kill the `replace-into` operations in progress

Usually, within half a minute (or faster), the query node will crash due to SIGSEGV.

------------------------

At least the following issue could lead to the previously described situation:

In the `ProcessorPtr::async_process` method, the returned async task may contain a reference to the `Processor` referenced in `ProcessorPtr`. However, their lifetime might not be aligned, leading to the possibility that this async task accesses a Processor instance that has already been dropped.

In this PR:

The async task (BoxFuture) holds a reference to the ProcessorPtr instance, and it only releases this reference when the task finishes.

-------------------------

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13847)

### DIFF
--- a/src/query/pipeline/core/src/processors/processor.rs
+++ b/src/query/pipeline/core/src/processors/processor.rs
@@ -157,14 +157,12 @@ impl ProcessorPtr {
     }
 
     /// # Safety
-    pub unsafe fn async_process(&self) -> BoxFuture<'static, Result<()>> {
-        let id = self.id();
-        let mut name = self.name();
-        name.push_str("::async_process");
-
-        let task = (*self.inner.get()).async_process();
-
+    pub unsafe fn async_process(self) -> BoxFuture<'static, Result<()>> {
         async move {
+            let id = self.id();
+            let mut name = self.name();
+            name.push_str("::async_process");
+            let task = (*self.inner.get()).async_process();
             let span = Span::enter_with_local_parent(name)
                 .with_property(|| ("graph-node-id", id.index().to_string()));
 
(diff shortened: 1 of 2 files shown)


dantengsky:
@zhang2014  

As you suggested,  `ProcessPtr::async_process` keeps taking `&self` (instead of `self`, and move it into the returned BoxFuture).

have to introduce  things like these:

~~~
struct UnsafeSyncCelledProcessor(UnsafeCell<Box<(dyn Processor)>>);
unsafe impl Sync for UnsafeSyncCelledProcessor {}
pub struct ProcessorPtr {
    ....
    inner: Arc<UnsafeSyncCelledProcessor>,
}
~~~

Please have a look

src/query/pipeline/core/src/processors/processor.rs
```diff
@@ -102,13 +115,13 @@ impl ProcessorPtr {
...
             id: Arc::new(UnsafeCell::new(node_index(0))),
-            inner: Arc::new(UnsafeCell::new(inner)),
+            inner: Arc::new(UnsafeCell::new(inner).into()),
         }
     }
 
     /// # Safety
     pub unsafe fn as_any(&mut self) -> &mut dyn Any {
-        (*self.inner.get()).as_any()
+        (*self.inner.0.get()).as_any()
```
SkyFan2002, line 124: Suggest impl `Deref` for `UnsafeSyncCelledProcessor`, then we don't need add `.0` everywhere.

dantengsky: good point, thanks

zhang2014 (approved):

PR #13852, opened 2023-11-29
chore(query): add settings to control enable stage/udf privilege check

TCeason:

## Summary

add settings to control enable stage/udf privilege check

```rust
("experiment_enable_stage_udf_priv_check", DefaultSettingValue {
                    value: UserSettingValue::UInt64(0),
                    desc: "experiment setting disables stage and udf privilege check(disable by default).",
                    possible_values: None,
                }),
```
- Closes #issue


TCeason:
after owner can control stage and udf, can drop this setting. cc @flaneur2020

BohuTANG:
We will remove this setting when the owner is done cc @flaneur2020

BohuTANG (approved):

Issue #13826, opened 2023-11-27
AggregateStddevPopFunction does not support type 'Decimal(Decimal128(DecimalSize { precision: 10, scale: 2 }))


BohuTANG:
**Summary**

Table setup:
https://github.com/BohuTANG/wizard/blob/main/checksb/sql/selects/bend/setup.sql

```

SQL: -- Query 12: Top 3 products with the most consistent pricing (smallest standard deviation in price) SELECT p.product_id, p.product_name, STDDEV_SAMP(s.net_paid) AS price_stddev FROM products p JOIN sales s ON p.product_id = s.product_id GROUP BY p.product_id, p.product_name ORDER BY price_stddev ASC, p.product_id LIMIT 3; 

1010: error: 
  --> SQL:4:38
  |
1 | 
2 | 
3 | -- Query 12: Top 3 products with the most consistent pricing (smallest standard deviation in price)
4 | SELECT p.product_id, p.product_name, STDDEV_SAMP(s.net_paid) AS price_stddev
  |                                      ^^^^^^^^^^^^^^^^^^^^^^^ AggregateStddevPopFunction does not support type 'Decimal(Decimal128(DecimalSize { precision: 10, scale: 2 }))'


```

PR #13864, opened 2023-11-30
chore: add decimal type agg stddev support

lichuang:

## Summary

chore: add decimal type agg stddev support

- Closes https://github.com/datafuselabs/databend/issues/13826


src/query/functions/src/aggregates/aggregate_stddev.rs
```diff
@@ -40,15 +49,17 @@ use crate::aggregates::AggregateFunction;
 
 const POP: u8 = 0;
 const SAMP: u8 = 1;
+const OVERFLOW_PRECISION: u8 = 18;
+const VARIANCE_PRECISION: u32 = 4;
```
ariesdevil, line 53: why use `u32` here?

src/query/functions/tests/it/aggregates/testdata/agg.txt
```diff
@@ -694,6 +694,16 @@ evaluation (internal):
 +--------+----------------------------------------------------------------------------+
 
 
+ast: stddev_samp(dec)
+evaluation (internal):
```
BohuTANG, line 698: Better to add a test case to window function which in the SQL logic test suites.

ariesdevil (approved):

PR #13878, opened 2023-11-30
feat(query): support take_ranges

### LINKED ISSUE #13873: Feature: support take_range in kernels
**Summary**

In the new filter execution framework, we will eventually generate a selection, but in some cases (the selection selects almost all indexes), calling `DataBlock::take` with selection does not achieve the best performance, a better method is to generate a `selection_range`, and then call `DataBlock::take_range` to generate data block.

Dousir9:

## Summary

Summary about this PR

- Closes #13873

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13878)


src/query/expression/src/kernels/take_ranges.rs
```diff
@@ -0,0 +1,280 @@
...
+use crate::with_decimal_type;
+use crate::with_number_type;
+use crate::BlockEntry;
+use crate::Column;
+use crate::ColumnBuilder;
+use crate::DataBlock;
+use crate::Value;
+
+impl DataBlock {
+    pub fn take_ranges(self, ranges: &Vec<(u32, u32)>, num_rows: usize) -> Result<DataBlock> {
```
BohuTANG, line 45: Better to add comments what's the new function will do.

src/query/expression/src/kernels/take_ranges.rs
```diff
@@ -0,0 +1,282 @@
...
+use crate::with_number_type;
+use crate::BlockEntry;
+use crate::Column;
+use crate::ColumnBuilder;
+use crate::DataBlock;
+use crate::Value;
+
+impl DataBlock {
+    // Generate a new `DataBlock` by the specified indices ranges.
+    pub fn take_ranges(self, ranges: &Vec<(u32, u32)>, num_rows: usize) -> Result<DataBlock> {
```
sundy-li, line 46: seems it's similar to `take_compacted_indices`

sundy-li (approved):

PR #13904, opened 2023-12-02
feat(cache): a variant of sieve, with lazy op

PsiACE:

## Summary

This optimization is inspired by the sieve, which can reduce unnecessary element movements and has a potential filtering effect.

Simply put, we maintain the visited status of each key and try to evict those elements that were inserted a long time ago but have not been accessed yet.

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13904)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1880,6 +1880,7 @@ dependencies = [
  "hashbrown 0.14.0",
  "hashlink",
  "heapsize",
+ "indexmap 1.9.2",
 ]
 
 [[package]]
--- a/src/common/cache/Cargo.toml
+++ b/src/common/cache/Cargo.toml
@@ -19,6 +19,7 @@ heapsize = ["heapsize_"]
 bytes = { workspace = true }
 hashbrown = "0.14"
 hashlink = "0.8"
+indexmap = "1.9.2"
 
 [target.'cfg(not(target_os = "macos"))'.dependencies]
 heapsize_ = { package = "heapsize", version = "0.4.2", optional = true }
--- a/src/common/cache/src/cache.rs
+++ b/src/common/cache/src/cache.rs
@@ -23,7 +23,7 @@ use crate::Meter;
 /// A trait for a cache.
 pub trait Cache<K, V, S, M>
 where
-    K: Eq + Hash,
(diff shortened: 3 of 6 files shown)


PsiACE:
It is no longer Lru and theoretically has better properties. For the sake of evaluation, I did not change the name. You can refer to the unit tests for a quick understanding.

The impact on Databend performance and cache hit rate needs further evaluation.

PsiACE:
This algorithm is storagebackend-friendly. In fact, we only need to maintain a "visited" linkedhashmap as the state. Since the state and storage are decoupled, we can consider refactoring our disk cache and implementing S3 cache in the future.

JackTan25:
image
well, this is the core idea of sieve cache. And the initial state of 'hand' p is null, it will be initialized by the oldest object which is the tail of the queue. And this cache algorithm is good to extended to the existed cache evict algorithms,
image

PsiACE:
IMG_20231204_125916
IMG_20231204_125950
IMG_20231204_130023
IMG_20231204_130052

dantengsky:
:+1: 

Besides the missing of `hand` pointer, are there any other tweaks or improvements worth mentioning?

And about the 'SIEVE is not scan-resistant' thing mentioned in the Sieve paper - any idea how that may affect us?

Also, does `find_evict_candidate` need to iterate through all elements in `visited` every time?

https://github.com/datafuselabs/databend/blob/c8b06a5c86cdba178bbc23ac589db2cb5e32e42f/src/common/cache/src/cache/lru.rs#L163-L174

PsiACE:
Although there seems to be improvement on the hits dataset, it performs similarly to Lru on some public traces and causes a decrease in throughput due to the O(n) traversal. I will try to make further modifications.

1a1a11a:
One possible way to avoid the `O(N)` operations is to track the visited bit using a HashSet. Add to the HashSet upon `get,` and at eviction time, we iterate through the map; if an object has been visited, we put it back. Otherwise, we evict the object. In the worst case, we may have to check N objects, but we can cap this to some value, e.g., 20. There are more optimized solutions, but would need more engineering work.

Issue #13422, opened 2023-10-24
bug: `generate_series` function with timestamp type is time-consuming


b41sh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

v1.2.175-nightly

### What's Wrong?

The start and end times are in seconds, but the step is in microseconds. We need to standardize the units of the parameters for `generate_series` function.

### How to Reproduce?

```sql
mysql> SELECT * FROM generate_series(to_timestamp(1), to_timestamp(10), 2);
+----------------------------+
| generate_series            |
+----------------------------+
| 1970-01-01 00:00:01.000000 |
| 1970-01-01 00:00:01.000002 |
| 1970-01-01 00:00:01.000004 |
| 1970-01-01 00:00:01.000006 |
...

mysql> SELECT * FROM generate_series(to_timestamp(1), to_timestamp(10), 2000000);
+----------------------------+
| generate_series            |
+----------------------------+
| 1970-01-01 00:00:01.000000 |
| 1970-01-01 00:00:03.000000 |
| 1970-01-01 00:00:05.000000 |
| 1970-01-01 00:00:07.000000 |
| 1970-01-01 00:00:09.000000 |
+----------------------------+
5 rows in set (0.02 sec)
Read 5 rows, 40.00 B in 0.001 sec., 4.43 thousand rows/sec., 34.61 KiB/sec.
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #13917, opened 2023-12-04
chore: fix generate_series to_timestamp step unit bug

lichuang:

## Summary

chore: fix generate_series to_timestamp step unit bug

- Closes https://github.com/datafuselabs/databend/issues/13422

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13917)


sundy-li (changes requested):
we should modify function to timestamp

sundy-li:
This is due to `to_timestamp` will have different units by value.  We can take this logic into generate_series function.

b41sh (approved):

PR #13918, opened 2023-12-04
feat(query): Add warnings in the query response

flaneur2020:

## Summary

still WIP

- Closes #issue

### DIFF
--- a/src/query/service/src/servers/http/v1/http_query_handlers.rs
+++ b/src/query/service/src/servers/http/v1/http_query_handlers.rs
@@ -123,6 +123,7 @@ pub struct QueryResponse {
     pub error: Option<QueryError>,
     pub stats: QueryStats,
     pub affect: Option<QueryAffect>,
+    pub warnings: Vec<String>,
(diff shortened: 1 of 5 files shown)


src/query/service/tests/it/storages/fuse/operations/commit.rs
```diff
@@ -494,6 +494,10 @@ impl TableContext for CtxDelegation {
         todo!()
     }
 
+    fn push_warning(&self, warn: String) {
```
youngsofun, line 497:
```suggestion
    fn push_warning(&self, _warn: String) {
```

src/query/service/tests/it/sql/exec/get_table_bind_test.rs
```diff
@@ -507,6 +507,10 @@ impl TableContext for CtxDelegation {
         todo!()
     }
 
+    fn push_warning(&self, warn: String) {
```
youngsofun, line 510:
```suggestion
    fn push_warning(&self, _warn: String) {
```

PR #13931, opened 2023-12-05
fix: max_active_session does not work for http handler.

youngsofun:

## Summary

Summary about this PR

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13931)

### DIFF
--- a/src/query/service/src/servers/http/v1/query/http_query.rs
+++ b/src/query/service/src/servers/http/v1/query/http_query.rs
@@ -235,7 +235,8 @@ impl HttpQuery {
             }
             session
         } else {
-            ctx.get_session(SessionType::HTTPQuery)
+            ctx.upgrade_session(SessionType::HTTPQuery)
+                .map_err(|err| ErrorCode::Internal(format!("{err}")))?
         };
 
         // Read the session variables in the request, and set them to the current session.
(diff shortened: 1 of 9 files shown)


src/query/service/src/servers/http/v1/query/http_query.rs
```diff
@@ -235,7 +235,8 @@ impl HttpQuery {
             }
             session
         } else {
-            ctx.get_session(SessionType::HTTPQuery)
+            ctx.upgrade_session(SessionType::HTTPQuery)
```
BohuTANG, line 238: Better to add a http api unit test to check the max sessions.

zhang2014 (approved):

ariesdevil (approved):

src/query/service/src/servers/http/v1/query/http_query.rs
youngsofun: working on it

youngsofun: done

youngsofun:
1. new impl not longer add dummy session to active sessions unless it is later upgraded. otherwise multi concurrent dummy sessions will occupy the slots for normal sessions, especially make it hard to test .
2. add tests

Issue #13865, opened 2023-11-30
bug: append column panic because of column type mismatch


b41sh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

v1.2.233

### What's Wrong?

append column panic because of column type mismatch

### How to Reproduce?

```sql
mysql> SELECT
    ->     lag([1295465.211, -1348321.148, 2126158.524],
    ->         8,
    ->         [-425330.365, 1081218.009, -2059672.339])
    ->     OVER
    ->     (PARTITION BY 10);
ERROR 1105 (HY000): PanicError. Code: 1104, Text = internal error: entered unreachable code: unable append column(data type: Nullable(Decimal(Decimal128(DecimalSize { precision: 10, scale: 3 })))) into builder(data type: Decimal(Decimal128(DecimalSize { precision: 10, scale: 3 }))).
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #13937, opened 2023-12-06
chore: fix lag default arg bug

lichuang:

## Summary

chore: fix lag default arg bug:

when query with `SELECT lag([456],8,[123]) OVER (PARTITION BY 10);`
```
panicked at src/query/expression/src/values.rs:2582:30:
internal error: entered unreachable code: unable append column(data type: Nullable(Number(UInt16))) into builder(data type: Number(UInt16))
```

because cast default expr `123` with `is_try` true.

- Closes https://github.com/datafuselabs/databend/issues/13865

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13937)

### DIFF
--- a/src/query/sql/src/planner/semantic/type_check.rs
+++ b/src/query/sql/src/planner/semantic/type_check.rs
@@ -1562,18 +1562,16 @@ impl<'a> TypeChecker<'a> {
             _ => unreachable!(),
         };
 
-        let default = if args.len() == 3 {
-            Some(args[2].clone())
+        let (default, return_type) = if args.len() == 3 {
+            (Some(args[2].clone()), arg_types[0].clone())
         } else {
-            None
+            (None, arg_types[0].wrap_nullable())
         };
-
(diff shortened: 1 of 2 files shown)


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1573,7 +1573,7 @@ impl<'a> TypeChecker<'a> {
         let cast_default = default.map(|d| {
             Box::new(ScalarExpr::CastExpr(CastExpr {
                 span: d.span(),
-                is_try: true,
+                is_try: false,
```
ariesdevil, line 1576: It seems `is_try` should be assigned by `return_type.is_nullable()`

ariesdevil:
the logical should be:
```
if default_provided:
    return_type = arg[0].type
else:
    // default is null
    return_type = arg[0].type.wrap_nullable()
```

then, `is_try` should be `return_type.is_nullable()`

ariesdevil (approved):

Issue #13932, opened 2023-12-05
bug: Not Null constraint is not consistent with postgres and snowflake


JackTan25:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

[35f129c](https://github.com/datafuselabs/databend/commit/35f129c23326477094a18d6da888651535e75c66)

### What's Wrong?
```sql
postgres=#  create table t3(a int not null,b text);
CREATE TABLE
Time: 24.030 ms
postgres=# insert into t3(b) values('b');
ERROR:  null value in column "a" of relation "t3" violates not-null constraint
DETAIL:  Failing row contains (null, b).
Time: 5.974 ms
```

snowflake:
<img width="710" alt="image" src="https://github.com/datafuselabs/databend/assets/60096118/a953b151-5c68-4ec4-ab8d-90db7a7bbe93">

### How to Reproduce?
databend:
```sql
mysql> create table t3(a int not null,b string);
Query OK, 0 rows affected (0.05 sec)

mysql> insert into t3(b) values(1);
Query OK, 1 row affected (0.09 sec)
```
_No response_

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

JackTan25:
the reason of the merge into result different from snowflake is above. cc @BohuTANG

PR #13939, opened 2023-12-06
fix: fix non-null constraint

JackTan25:

## Summary

Summary about this PR
1. for snowflake and postgres, it will return a non-null error when fill default value for non-null field.
2. in this pr, when we try to fill default expr for a non-null field, if its default expr is none, we will return the error
- Closes #issue
https://github.com/datafuselabs/databend/issues/13932

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13939)

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/transform_resort_addon_without_source_schema.rs
+++ b/src/query/service/src/pipelines/processors/transforms/transform_resort_addon_without_source_schema.rs
@@ -17,6 +17,7 @@ use std::sync::Arc;
 use common_catalog::table_context::TableContext;
 use common_exception::ErrorCode;
 use common_exception::Result;
+use common_expression::types::DataType;
 use common_expression::BlockMetaInfoDowncast;
 use common_expression::DataBlock;
 use common_expression::DataSchemaRef;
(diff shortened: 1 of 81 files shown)


src/query/service/src/pipelines/processors/transforms/transform_resort_addon_without_source_schema.rs
```diff
@@ -48,6 +50,25 @@ pub fn build_expression_transform(
...
+            // but for pg or snowflake, if a field is non-null, it will return
+            // a non-null error (it means they will give a null as the default value).
+            if !f.is_nullable()
+                && f.default_expr().is_none()
+                && !matches!(f.data_type(), DataType::Bitmap)
+            {
+                // if we have a user-specified default expr, it must satisfy the non-null constraint
+                // in table-create phase. So we just consider default_expr is none.
+                return Err(ErrorCode::BadArguments(format!(
+                    "unable to cast type `null` value to a non-null type field `{}`",
```
sundy-li, line 68: The error msg is wrong, 
```
null value in column "{}" violates not-null constraint
```

sundy-li:
This change is correct, but it is breaking changes. It may cause errors for old scripts or sink processes (insert by other drivers).
cc @BohuTANG

JackTan25:
when run script `tests/suites/0_stateless/14_clickhouse_http_handler/14_0007_http_clickhouse_input_format_diagnostic.sh`,we will
get a stack error, so the reason is below:
```rust
split_tx.send(Ok(Split {
    info: split_info,
    rx: data_rx,
}))
.await
.expect("fail to send split from stream"); // do expect here
```
let me remove the bad test added by my.

b41sh (approved):

TCeason (approved):

JackTan25:
cc @BohuTANG merge into wizard test is passed, can we merge this?

BohuTANG:
A12/A13 need be fixed to make them works for Databend and Snowflake, I will do it this weekend. @JackTan25

BohuTANG:
> A12/A13 need be fixed to make them works for Databend and Snowflake, I will do it this weekend. @JackTan25

Fixex: https://github.com/datafuselabs/wizard/commit/9cd4d96e9f2a075918e252fa68123531bb54ae53

PR #13965, opened 2023-12-09
feat: add setting to disable clickhouse handler

### LINKED ISSUE #13907: add settings to disable ClickHouse hanlder
**Summary**

We prepare to remove the clickhouse handler, before remove better to add a settings:
```
enable_clickhouse_handler
```

default is 0.

guojidan:

## Summary

Summary about this PR

- Closes #13907 issue

add `enable_clickhouse_handler` setting

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13965)


BohuTANG:
We can use settings, default is 0(disable).

guojidan:
> enable_clickhouse_handler

guojidan:
> We can use settings, default is 0(disable).

I think `bool` type better than 0/1. can you explain more detail?

BohuTANG:
We can reference this file:
https://github.com/datafuselabs/databend/blob/ae9dda94b0015e07a23b5d84e73fea77304bf4c6/src/query/settings/src/settings_default.rs

Databend has many settings already, we can add new one like: `enable_clickhouse_handler` default is 0(disable).

guojidan:
test case failed seem like not  I caused it

src/query/service/src/servers/http/clickhouse_handler.rs
```diff
@@ -247,6 +246,16 @@ pub async fn clickhouse_handler_get(
...
             .map_err(BadRequest)?;
 
+        if !settings
+            .get_enable_clickhouse_handler()
+            .map_err(InternalServerError)?
+        {
+            return Ok(Body::from_string(
+                "default settings: enable_clickhouse_handler is 0".to_string(),
+            )
+            .with_content_type("text/abc"));
```
BohuTANG, line 256: What's the `text/abc` mean?

guojidan: emmm, should `text/plain`? sorry I am not familiar with http protocol

youngsofun: should return PoemError (instead of Ok) when not 
what about [405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405)

youngsofun (approved):

Issue #13889, opened 2023-12-01
Tracking: external sort (sort out of core).


RinChanNOWWW:
**Summary**

The new pipeline will be like:

![sort_ooc eng](https://github.com/datafuselabs/databend/assets/33975039/d553d185-b22e-4d0f-86f7-b5296fef201f)

## Tasks

- [x] Refine distributed sort pipeline in cluster mode. #13881
- [x] Refactor `TransformSortMerge(Limit)` to be able to output data to downstream before finishing. #13927
- [x] Implement a new processor to deal with spilling. #13938
- Improve the sort pipeline with the new processor.
  - [x] Improve the full sort pipeline. #13989
  - [x] Improve the TopN pipeline. #14131

PR #13989, opened 2023-12-12
feat: enable sort spilling when memory reach the limit.

RinChanNOWWW:

## Summary

- Use builder pattern to build the merge sort pipeline.
- In `TransformSortMerge`, if memory limit is reached, add meta to the output `DataBlock` to let `TransformSortSpill` spill the blocks.

Tracking in #13889

## Tests
- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13989)


tests/sqllogictests/suites/base/03_common/03_0041_sort_spill.test
```diff
@@ -0,0 +1,402 @@
+statement ok
```
BohuTANG, line 1: It seems that we don't check the spill or not in this test?

BohuTANG: We can check the metrics counter.

RinChanNOWWW: Can we get the metrics by SQL command?

BohuTANG: Yes, 
`select metric, value from system.metrics where metric like '%<metric-name>%';`

zhang2014 (approved):

sundy-li (approved):

PR #13993, opened 2023-12-12
feat: support more runtime filter types

xudong963:

## Summary

Currently we support three runtime filter types: bloom filter, inlist filter and min-max filter.

The inlist filter is only generated if the amount of data on the build side is less than or equal to 1024.

The min-max filter is generated regardless of the data size.

Both inlist filter and min-max filter will be used for partitions prune.

The bloom filter is generated regardless of the data size, but its generation needs to be enabled. `set enable_runtime_filter = 1`;

The bloom filter will filter the data block after deserialization in the parquet fuse engine. In the native fuse engine, it will prune the page first, and then filter the data block together with the prewhere filter.

The bloom runtime filter is useful in certain scenarios, such as when there are heavy computations between scan and join. Because we need to build a bloom filter to scan it, it has some overhead, so currently we decide whether to turn it on or not through the configuration, and in the future we can adapt to it by estimating some statistics (but there are inaccuracies, and the estimation error is relatively large, which needs to be further determined).

## Tests
- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test 

For the bloom runtiem filter, the configuration is turned on for testing in the tpch and tpcds tests
For inlist and min-max filters tests now go to the relevant logic by default.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13993)


src/query/storages/fuse/src/operations/read/parquet_data_source_deserializer.rs
```diff
@@ -62,6 +72,46 @@ pub struct DeserializeDataTransform {
     virtual_reader: Arc<Option<VirtualColumnReader>>,
 
     base_block_ids: Option<Scalar>,
+    cached_runtime_filter: Option<Expr>,
+}
+
+#[macro_export]
+macro_rules! impl_runtime_filter {
```
BohuTANG, line 79: Why use the macro here?

xudong963: `NativeDeserializeDataTransform` and `DeserializeDataTransform` will share the method

src/query/storages/fuse/src/operations/read/parquet_data_source_deserializer.rs
```diff
@@ -62,6 +72,46 @@ pub struct DeserializeDataTransform {
...
+                        Err(ErrorCode::Internal(
+                            "Invalid empty predicate list".to_string(),
+                        ))
+                    })?;
+                self.cached_runtime_filter = Some(filter);
+            }
+            // Using runtime filter to filter data_block
+            let func_ctx = self.ctx.get_function_context()?;
+            let evaluator = Evaluator::new(&data_block, &func_ctx, &BUILTIN_FUNCTIONS);
+            let filter = evaluator
```
Dousir9, line 108: We perform one more filter on all data on the probe side, will this always reduce the overall query time ?

Dousir9: > We perform extra filter on all data on the probe side, will this always reduce the overall query time ?

Dousir9:
In fact, after prewhere, the runtime filter executed in storage layer is equivalent to executing `TransformFilter` in Pipeline.

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
```diff
@@ -795,7 +809,12 @@ impl Processor for NativeDeserializeDataTransform {
                 block
             };
 
-            // Step 7: Add optional virtual columns
+            // Step 7: runtime filter
+            if self.ctx.has_runtime_filters(self.table_index) {
```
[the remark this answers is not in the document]
xudong963: Do you mean to run runtime filter and prewhere filter together?  Maybe putting runtime filter after prewhere is enough. 
https://github.com/datafuselabs/databend/pull/13993/files#diff-5f07a6b7fa92f1af336bb92dd94bad639be187e259a37f4a06e80b20bbcb7b1bR821-R824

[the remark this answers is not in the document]
Dousir9: No, this is not a performance issue, but a correctness issue, the DataBlock filtered by the runtime filter no longer corresponds to the filter generated by prewhere.

src/query/storages/fuse/src/operations/read/parquet_data_source_deserializer.rs
xudong963: Update performance explanation in summary.

xudong963:
Update: After two days of testing, this PR may cause a potential performance degradation. The filtering done here is not as effective as doing it directly in the hashjoin probe. Additionally, if the selectivity is very low, it can introduce the overhead of filter execution. Although it has the potential to improve performance in a distributed environment by reducing the amount of data shuffled, this depends on certain rules to effectively reduce the data volume. In the future, we can formulate these rules first, and then continue with this PR.

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_build_state.rs
```diff
@@ -101,6 +111,10 @@ pub struct HashJoinBuildState {
     pub(crate) send_val: AtomicU8,
     /// Wait all processors finish read spilled data, then go to new round build
     pub(crate) restore_barrier: Barrier,
+    pub(crate) enable_inlist_runtime_filter: bool,
+    pub(crate) enable_min_max_runtime_filter: bool,
+    /// Need to open runtime filter setting.
```
JackTan25, line 116: for inlist and min_max index, we really need a flag to control them? I think they can always be open.

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_build_state.rs
```diff
@@ -229,15 +264,30 @@ impl HashJoinBuildState {
...
-                        .chunks
-                        .clone()
-                };
+            let build_chunks = unsafe {
+                (*self.hash_join_state.build_state.get())
+                    .generation_state
+                    .chunks
+                    .clone()
+            };
+
```
JackTan25, line 273: if it takes place spill, can we hold these all entirely?

src/query/storages/fuse/src/operations/read/parquet_data_source_deserializer.rs
```diff
@@ -110,8 +126,138 @@ impl DeserializeDataTransform {
...
+                            idx += 1;
+                        }),
+                        _ => unreachable!(),
+                    }
+                }
+                _ => unreachable!(),
+            }
+        }
+        data_block.filter_with_bitmap(&bitmap.into())
+    }
```
JackTan25, line 260: The native data source and here have many duplicate codes. Maybe we can reduce that, but it doesn't matter.

JackTan25:
https://github.com/datafuselabs/databend/pull/13993/files#diff-84c18d94fc35afdff4847006353b09aadc9f47326da6e6dc2238d4a4d224700eL241 we need to add offsets here after filter some rows by the filter expr of runtime filter. The same with native source. Thanks cc @xudong963 And please add test for this.

JackTan25:
After this merge, we can start to test the performance of merge-into-runtime-filter branch cc @SkyFan2002

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_build_state.rs
xudong963: Still have some limitations: https://github.com/datafuselabs/databend/pull/13993/files#diff-3a9124588945cc9f4c83fd0bfef409c1fc3ef74b2ccfd49e2504af36bbbb3a9dR144-R157

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_build_state.rs
xudong963: Spill and runtime filter are in conflict. We won't enable runtime filter when spilling.

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
```diff
@@ -505,6 +524,177 @@ impl NativeDeserializeDataTransform {
...
+        // Check if already cached runtime filters
+        if self.cached_bloom_runtime_filter.is_none() {
+            let bloom_filters = self.ctx.get_bloom_runtime_filter_with_id(self.table_index);
+            let bloom_filters = bloom_filters
+                .into_iter()
+                .filter_map(|filter| {
+                    let name = filter.0.as_str();
+                    // Some probe keys are not in the schema, they are derived from expressions.
+                    self.src_schema
+                        .index_of(name)
```
RinChanNOWWW, line 543: There is a case that is not quite discoverable: the probe key is an inner column of a nested type. In another word, `name` could be "a:b:c" or something like that. It's better to add a test case to cover this case.

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
```diff
@@ -799,12 +990,33 @@ impl Processor for NativeDeserializeDataTransform {
...
+                for rf_filter in rf_filters.into_iter().skip(1) {
+                    rf_bitmap = rf_bitmap.bitand(&rf_filter.into());
+                }
+                if let Some(filter) = &filter {
+                    rf_bitmap = rf_bitmap.bitand(filter.as_column().unwrap());
+                }
+                Some(Value::Column(rf_bitmap.into()))
+            } else {
+                filter
+            };
```
RinChanNOWWW, line 1006: Can use `rf_filters.iter().fold()`.

And how about moving these codes into the method `bloom_runtime_filter`?

test.sql
```diff
@@ -0,0 +1,119 @@
+WITH year_total AS
```
RinChanNOWWW, line 1: Wrongly pushed?

RinChanNOWWW (changes requested):

src/query/service/src/pipelines/processors/transforms/hash_join/hash_join_build_state.rs
```diff
@@ -745,4 +785,168 @@ impl HashJoinBuildState {
         }
         Ok(())
     }
+
+    fn bloom_runtime_filter(
```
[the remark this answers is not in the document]
xudong963: No, runtime filter uses a different bloom filter(https://github.com/ayazhafiz/xorf) and hash method.

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
```diff
@@ -505,6 +524,177 @@ impl NativeDeserializeDataTransform {
...
+    ) -> Result<Vec<MutableBitmap>> {
+        let mut local_arrays = vec![];
+        // Check if already cached runtime filters
+        if self.cached_bloom_runtime_filter.is_none() {
+            let bloom_filters = self.ctx.get_bloom_runtime_filter_with_id(self.table_index);
+            let bloom_filters = bloom_filters
+                .into_iter()
+                .filter_map(|filter| {
+                    let name = filter.0.as_str();
+                    // Some probe keys are not in the schema, they are derived from expressions.
```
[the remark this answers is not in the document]
xudong963: These kinds of probe keys won't leverage bloom runtime filters. We can't get their values during the current phase

test.sql
xudong963: yeah

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
```diff
@@ -505,6 +524,177 @@ impl NativeDeserializeDataTransform {
...
+                            arrays.push((*idx, array.clone()));
+                            local_arrays.push((*idx, array.clone()));
+                            self.array_skip_pages.insert(*idx, 0);
+                        }
+                        None => {
+                            return Ok(vec![]);
+                        }
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
xudong963: If prewhere columns are empty, we add probe keys columns to `prewhere_columns`.  Are there some risks involved?

https://github.com/datafuselabs/databend/blob/941fa8afccc1d49f2aa0ac03b42963d5496813c9/src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs#L652-L677

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
xudong963: We don't support nested type for probe key, https://github.com/datafuselabs/databend/pull/13993/files#diff-3a9124588945cc9f4c83fd0bfef409c1fc3ef74b2ccfd49e2504af36bbbb3a9dR802

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
xudong963: done

src/query/storages/fuse/src/operations/read/parquet_data_source_deserializer.rs
xudong963: done

xudong963:
> we need to add offsets here after filter some rows by the filter expr of runtime filter. The same with native source. Thanks cc @xudong963 And please add test for this.

Native source has processed offsets, we only need to process parquet source. I'll open runtime filter setting for merge into tests, because they cover the case.

JackTan25 (approved):

src/query/storages/fuse/src/operations/read/native_data_source_deserializer.rs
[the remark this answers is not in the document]
RinChanNOWWW: Ok, let's keep these codes for now.

RinChanNOWWW:
`TableScan` part is LGTM.

SkyFan2002 (approved):

PR #13997, opened 2023-12-12
refactor: using borsh instead of bincode for serde agg function state

### LINKED ISSUE #13979: feat: consider use borsh to serde aggregation state
**Summary**

The format of [borsh](https://borsh.io/) is exactly the same as the aggregation state.

Since borsh claims it has better performance than bincode crate, we should verify it.

1. Investigate the performance between borsh and bincode in serde  (Sum, Avg, MinState ) using criterion.
2. If it works, use borsh replace bincode.

ariesdevil:

## Summary

using borsh instead of bincode for serde agg function state

BTW: there are three crates that using my own patch version, after these merge to upstream, I'll update them to databend.
* https://github.com/ariesdevil/opensrv/blob/borsh/components/micromarshal/Cargo.toml#L20
* https://github.com/datafuse-extras/streaming_algorithms/pull/1
* https://github.com/nlordell/ethnum-rs/pull/42

- Closes #13979 

## Tests
- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13997)


Cargo.toml
```diff
@@ -138,7 +139,7 @@ clap = { version = "4.4.2", features = ["derive"] }
 dashmap = "5.4.0"
 derive_more = "0.99.17"
 enumflags2 = { version = "0.7.7", features = ["serde"] }
-ethnum = { version = "1.3.2" }
+ethnum = { git = "https://github.com/ariesdevil/ethnum-rs", rev = "4cb05f1" }
```
sundy-li, line 142: better create pr to merge into upstream later

ariesdevil: https://github.com/nlordell/ethnum-rs/pull/42

BohuTANG:
What's the performance compare to bincode?

ariesdevil:
> What's the performance compare to bincode?↳

I'm doing now, will be post to this PR later

ariesdevil:
```
│❯ cargo benchcmp bincode borsh                                                                                                                                                                                                                                                                                              │
│ name                            bincode ns/iter  borsh ns/iter  diff ns/iter   diff %  speedup                                                                                                                                                                                                                             │
│ tests::bench_borsh_bincode_ser  15               12                       -3  -20.00%   x 1.25


│ name                              bincode_de ns/iter  borsh_de ns/iter  diff ns/iter   diff %  speedup                                                                                                                                                                                                                     │
│ tests::bench_borsh_bincode_deser  32                  19                         -13  -40.62%   x 1.68
```

ariesdevil:
I'll go on modify other state ser/deser

Cargo.toml
BohuTANG: We can use the `https://github.com/nlordell/ethnum-rs` with your PR commit.

src/common/io/src/borsh_serialization.rs
```diff
@@ -0,0 +1,40 @@
+// Copyright 2021 Datafuse Labs
```
BohuTANG, line 1: We can rename `io/src/serialization.rs` to `io/src/bincode_serialization.rs`.
We also need a unit test for the borsh_seralization.rs, reference:
https://github.com/datafuselabs/databend/blob/main/src/common/io/tests/it/serialization.rs

BTW, this unit test most parts generated by GPT4.

src/common/io/src/borsh_serialization.rs
```diff
@@ -0,0 +1,40 @@
...
+#[inline]
+pub fn borsh_deserialize_from_slice<T: BorshDeserialize>(slice: &[u8]) -> Result<T> {
+    let value = borsh::from_slice::<T>(slice)?;
+    Ok(value)
+}
+
+#[inline]
+pub fn borsh_deserialize_from_stream<T: BorshDeserialize>(slice: &mut &[u8]) -> Result<T> {
+    let value = borsh::from_slice::<T>(slice)?;
+    Ok(value)
```
BohuTANG, line 39:
```suggestion
    let value = borsh::from_slice::<T>(slice)?;
    
     // Update the slice to point to the remaining bytes.
    *stream = &stream[bytes_read..];
    Ok(value)
```

src/common/io/src/borsh_serialization.rs
```diff
@@ -0,0 +1,40 @@
...
+}
```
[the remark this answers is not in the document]
sundy-li: `from_reader` ensures the stream is consumed.

We should use ` T::deserialize_reader(stream)` instead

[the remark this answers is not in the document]
BohuTANG: Right, I missing this:
https://github.com/near/borsh-rs/blob/master/borsh/src/de/mod.rs#L37-L38

src/query/functions/src/aggregates/aggregator_common.rs
```diff
@@ -154,14 +156,14 @@ pub fn eval_aggr(
...
 ) -> Result<()> {
-    bincode_serialize_into_buf(writer, value)
+    borsh_serialize_into_buf(writer, value)
 }
 
 #[inline]
-pub fn deserialize_state<T: serde::de::DeserializeOwned>(slice: &mut &[u8]) -> Result<T> {
-    bincode_deserialize_from_stream(slice)
+pub fn borsh_deserialize_state<T: BorshDeserialize>(slice: &mut &[u8]) -> Result<T> {
+    borsh_deserialize_from_slice(slice)
```
BohuTANG, line 168: borsh_deserialize_from_slice will set the new position of the slice after read once? bincode_deserialize_from_stream does.

ariesdevil: nop, I realize this, currently borsh not provide a stream decode, we need fork it and impl one.

tests/sqllogictests/suites/query/02_function/02_0000_function_aggregate_state.test
```diff
@@ -1,9 +1,9 @@
 query T
 select length(max_state(number)), typeof(max_state(number)) from numbers(100);
 ----
-3 VARCHAR
+10 VARCHAR
```
BohuTANG, line 4: Does this means the state is larger than before?

sundy-li: Yes, because we are using fixed-size encoding rather than variant encoding.

But this may be trivial, the protocol compression could be effective in the cluster.

ariesdevil: yes, that's the string length.

BohuTANG: > Yes, because we are using fixed-size encoding rather than variant encoding.
> 
> But this may be trivial, the protocol compression could be effective in the cluster.

More concerned about the new state memory usage, the OOM issue.

sundy-li: serialization is only used in cluster queries.

sundy-li (approved):

Cargo.toml
```diff
@@ -254,3 +255,4 @@ parquet2 = { git = "https://github.com/jorgecarleitao/parquet2", rev = "b0e6545"
 metrics = { git = "https://github.com/datafuse-extras/metrics.git", rev = "fc2ecd1" }
 icelake = { git = "https://github.com/icelake-io/icelake", rev = "f06cdf3" }
 sentry = { git = "https://github.com/getsentry/sentry-rust", rev = "6ef6d97" }
+micromarshal = { git = "https://github.com/ariesdevil/opensrv", rev = "6c96813" }
```
BohuTANG, line 258: Why not give the change back and use https://github.com/datafuselabs/opensrv?

ariesdevil: micromarshal also need patched ethnum https://github.com/ariesdevil/opensrv/blob/borsh/components/micromarshal/Cargo.toml#L20 
after ethnum upstream merge, I'll modify this.

Issue #13998, opened 2023-12-13
bug: fix drop_table_by_id when TableIdToName cannot found


lichuang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

current

### What's Wrong?

`drop_table_by_id` when TableIdToName cannot found will return fail

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #14002, opened 2023-12-13
chore: fix drop_table_by_id fail when TableIdToName cannot found

lichuang:

## Summary

chore: fix drop_table_by_id fail when TableIdToName cannot found

- Closes https://github.com/datafuselabs/databend/issues/13998

## Tests
- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14002)

### DIFF
--- a/src/binaries/metabench/main.rs
+++ b/src/binaries/metabench/main.rs
@@ -235,7 +235,11 @@ async fn benchmark_table(client: &Arc<ClientHandle>, prefix: u64, client_num: u6
     let res = client
         .drop_table_by_id(DropTableByIdReq {
             if_exists: false,
-            tenant: tenant(),
+            name_ident: TableNameIdent {
+                tenant: tenant(),
+                db_name: db_name(),
+                table_name: table_name(),
+            },
             tb_id: t.ident.table_id,
         })
         .await;
--- a/src/binaries/metabench/main.rs
+++ b/src/binaries/metabench/main.rs
@@ -235,11 +235,9 @@ async fn benchmark_table(client: &Arc<ClientHandle>, prefix: u64, client_num: u6
     let res = client
         .drop_table_by_id(DropTableByIdReq {
             if_exists: false,
-            name_ident: TableNameIdent {
-                tenant: tenant(),
-                db_name: db_name(),
-                table_name: table_name(),
-            },
+            tenant: tenant(),
+            db_id: db_name(),
+            table_name: table_name(),
             tb_id: t.ident.table_id,
         })
         .await;
--- a/src/binaries/metabench/main.rs
+++ b/src/binaries/metabench/main.rs
@@ -203,6 +203,10 @@ async fn benchmark_table(client: &Arc<ClientHandle>, prefix: u64, client_num: u6
         .await;
 
     print_res(i, "create_db", &res);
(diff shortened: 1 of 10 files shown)


src/query/ee/src/stream/handler.rs
```diff
@@ -176,7 +176,11 @@ impl StreamHandler for RealStreamHandler {
             catalog
                 .drop_table_by_id(DropTableByIdReq {
                     if_exists: plan.if_exists,
-                    tenant,
+                    name_ident: TableNameIdent {
+                        tenant: tenant.clone(),
+                        db_name: plan.database.clone(),
+                        table_name: stream_name.clone(),
+                    },
                     tb_id: table.get_id(),
```
drmingdrmer, line 184: What about passing in a `database_id: DatabaseId`?
Because what is actually required is a `database_id` by `drop_table_by_id`.

In order to decouple logic, 
I would always suggest moving logic(get db_meta) to a higher level(from `SchemaApi::drop_table_by_id()` to this this method `do_drop_stream()`.

drmingdrmer: Another concern is that using `xxxName` internally still has a consistency issue, because `database-name` can be renamed by another transaction.

src/binaries/metabench/main.rs
```diff
@@ -235,11 +235,9 @@ async fn benchmark_table(client: &Arc<ClientHandle>, prefix: u64, client_num: u6
     let res = client
         .drop_table_by_id(DropTableByIdReq {
             if_exists: false,
-            name_ident: TableNameIdent {
-                tenant: tenant(),
-                db_name: db_name(),
-                table_name: table_name(),
-            },
+            tenant: tenant(),
+            db_id: db_name(),
```
drmingdrmer, line 239: `db_name()` returns an ID? 🤔

drmingdrmer (commented):
> Reviewed 2 of 10 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14002)* status: 3 of 10 files reviewed, 3 unresolved discussions (waiting on @lichuang)

___
*[`src/meta/api/src/schema_api_impl.rs` line 2362 at r3](https://reviewable.io/reviews/datafuselabs/databend/14002#-NlatFR405-ttS_6Qeyp:-NlatFR405-ttS_6Qeyq:b-76iuna) ([raw file](https://github.com/datafuselabs/databend/blob/675355d0451315e9b79d3e3cf8a4b304f00a835d/src/meta/api/src/schema_api_impl.rs#L2362)):*
> ```XML
> 
>             let db_id = dbid_tbname.db_id;
>             let tbname = dbid_tbname.table_name.clone();
> ```

It looks like `tbname` is only used for building an error.  Therefore there is no need to pass in a `table_name` in `req`. And in this scenario, the error could be `UnknownTableId` instead of an `UnknownTable`. Because when the table can not be found by `(db_id, table_name)`, the only possible reason is that the table is remove by other transaction.

This way we do not need to be bother using `table_name` in `req`.

What say you?

drmingdrmer (approved):

PR #14010, opened 2023-12-14
chore: merge settings possible_values to range

BohuTANG:

## Summary

* merge `possible_values` into `range`
* add range limit:`[0, 1]` for bool settings like `enable_**`
* add `range` column to `system.settings` table

```
+------------------------------------------------+----------------+----------------+---------------------------------------------------+-----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| Column 0                                       | Column 1       | Column 2       | Column 3                                          | Column 4  | Column 5                                                                                                                                                                              | Column 6 |
+------------------------------------------------+----------------+----------------+---------------------------------------------------+-----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| 'acquire_lock_timeout'                         | '15'           | '15'           | 'None'                                            | 'SESSION' | 'Sets the maximum timeout in seconds for acquire a lock.'                                                                                                                             | 'UInt64' |
| 'aggregate_spilling_bytes_threshold_per_proc'  | '0'            | '0'            | 'None'                                            | 'SESSION' | 'Sets the maximum amount of memory in bytes that an aggregator can use before spilling data to storage during query execution.'                                                       | 'UInt64' |
| 'aggregate_spilling_memory_ratio'              | '0'            | '0'            | 'None'                                            | 'SESSION' | 'Sets the maximum memory ratio in bytes that an aggregator can use before spilling data to storage during query execution.'                                                           | 'UInt64' |
| 'collation'                                    | 'binary'       | 'binary'       | '["binary", "utf8"]'                              | 'SESSION' | 'Sets the character collation. Available values include "binary" and "utf8".'                                                                                                         | 'String' |
| 'create_query_flight_client_with_current_rt'   | '1'            | '1'            | 'None'                                            | 'SESSION' | 'create query flight client with current runtime'                                                                                                                                     | 'UInt64' |
| 'ddl_column_type_nullable'                     | '1'            | '1'            | 'None'                                            | 'SESSION' | 'If columns are default nullable when create or alter table'                                                                                                                          | 'UInt64' |
| 'disable_join_reorder'                         | '0'            | '0'            | 'None'                                            | 'SESSION' | 'Disable join reorder optimization.'                                                                                                                                                  | 'UInt64' |
| 'efficiently_memory_group_by'                  | '0'            | '0'            | 'None'                                            | 'SESSION' | 'Memory is used efficiently, but this may cause performance degradation.'                                                                                                             | 'UInt64' |
```
- Closes #issue

## Tests
- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14010)


src/query/settings/src/settings_default.rs
```diff
@@ -711,56 +677,55 @@ impl DefaultSettings {
...
+                                // Convert the value to lowercase for case-insensitive comparison
+                                let val_lower = v.to_lowercase();
 
-                    for possible_value in possible_values {
-                        if possible_value.eq_ignore_ascii_case(&v) {
-                            checked_value = true;
-                            v = possible_value.to_string();
-                        }
-                    }
+                                range.is_within_string_range(&val_lower)?;
```
zhang2014, line 722:
```suggestion
                                let matched_value = range.is_within_string_range(&val_lower)?;
                                Ok(k, Some(UserSettingValue::String(matched_value))))
```

src/query/settings/src/settings_global.rs
```diff
@@ -96,14 +97,34 @@ impl Settings {
...
-                        if !default_setting_value
-                            .possible_values
-                            .as_ref()
-                            .map(|values| values.iter().any(|v| v.eq_ignore_ascii_case(&val)))
-                            .unwrap_or(true)
-                        {
-                            // the settings may be deprecated
-                            warn!("Ignore invalid global setting {} = {}", name, val);
+                        // Check if the setting value is valid based on the defined range
+                        let validation_result = match &default_setting_value.range {
```
zhang2014, line 101: maybe remove check range is better?

src/query/settings/src/settings_default.rs
BohuTANG: The suggestion is same as the patch code:
```
                                let val_lower = v.to_lowercase();

                                range.is_within_string_range(&val_lower)?;

                                // Return the key and value if valid
                                Ok((k, Some(UserSettingValue::String(v))))
```

when `is_within_string_range` passed, the `v` is the matched value.

BohuTANG:
The test is on fixing.

zhang2014 (approved):

sundy-li (approved):

PR #14013, opened 2023-12-14
refactor: access KVMeta with methods instead of field

drmingdrmer:

##### refactor: access KVMeta with methods instead of field

Make `KVMeta.expire_at` a private field for future refactoring.
Update accessing to it with via methods.

Several methods are added to `KVMeta` and `SeqValue`:

- `eval_expire_at_ms()` evaluate and return a absolute expire time in millisecond. The returned value is set to `u64::MAX` if not expire is set.
- `get_expire_at_ms()` return an `Option` of absolute expire time in millisecond.

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14013)


ariesdevil (approved):

Xuanwo (approved):
Nice change!

PR #14029, opened 2023-12-15
chore: add runtime filter pruned rows to query log

xudong963:

## Summary

Summary about this PR

- Closes #issue

## Tests
- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14029)


BohuTANG:
Can we add it to profiling or `explain analyze`? Add to query_log is not userful, the query log is not for profiling.

xudong963:
> Can we add it to profiling or `explain analyze`? Add to query_log is not userful, the query log is not for profiling.

I'll give a try.

Issue #14070, opened 2023-12-18
Handle error returned from `resolve_data_source()`.


drmingdrmer:
The error should not be ignored and need to be dealt with:
https://github.com/datafuselabs/databend/blob/629ee7bab86bf74538814e794122e1fedbbeb288/src/query/sql/src/planner/binder/table.rs#L244-L265

BohuTANG:
https://github.com/datafuselabs/databend/pull/14097

PR #14076, opened 2023-12-19
chore: handle error returned from `resolve_data_source()`

xudong963:

## Summary

handle error returned from `resolve_data_source()`

Fixes https://github.com/datafuselabs/databend/issues/14070

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test  - covered by current codebase

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [x] Other (please describe): improve error handling

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14076)

### DIFF
--- a/src/query/sql/src/planner/binder/table.rs
+++ b/src/query/sql/src/planner/binder/table.rs
@@ -260,6 +260,12 @@ impl Binder {
                     }
                     parent = bind_context.parent.as_mut();
                 }
+                if e.code() == 1025 {
+                    return Err(ErrorCode::UnknownTable(format!(
+                        "Unknown table `{database}`.`{table_name}` in catalog '{catalog}'"
+                    ))
+                    .set_span(*span));
+                }
                 return Err(e);
             }
         };
--- a/src/query/sql/src/planner/binder/table.rs
+++ b/src/query/sql/src/planner/binder/table.rs
@@ -241,7 +241,7 @@ impl Binder {
             .await
         {
             Ok(table) => table,
-            Err(_) => {
+            Err(e) => {
                 let mut parent = bind_context.parent.as_mut();
                 loop {
                     if parent.is_none() {
@@ -260,10 +260,7 @@ impl Binder {
                     }
                     parent = bind_context.parent.as_mut();
                 }
-                return Err(ErrorCode::UnknownTable(format!(
-                    "Unknown table `{database}`.`{table_name}` in catalog '{catalog}'"
-                ))
-                .set_span(*span));
+                return Err(e);
             }
         };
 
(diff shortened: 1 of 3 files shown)


src/query/sql/src/planner/binder/table.rs
```diff
@@ -260,10 +260,13 @@ impl Binder {
...
-                    "Unknown table `{database}`.`{table_name}` in catalog '{catalog}'"
-                ))
-                .set_span(*span));
+                if e.code() == 1025 {
+                    return Err(ErrorCode::UnknownTable(format!(
+                        "Unknown table `{database}`.`{table_name}` in catalog '{catalog}'"
+                    ))
+                    .set_span(*span));
+                }
+                return Err(e);
```
xudong963, line 269: Need help cc @lichuang 

image

The error is from `resolve_data_source()`.

drmingdrmer (commented):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14076)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @BohuTANG, @lichuang, and @xudong963)

___
*[`src/query/sharing/src/share_endpoint.rs` line 150 at r2](https://reviewable.io/reviews/datafuselabs/databend/14076#-Nm4-hkXCNopfonzufP_:-Nm4-hkXCNopfonzufPa:bgzz59p) ([raw file](https://github.com/datafuselabs/databend/blob/f236a6ac371a6a595c31dbe10e4068a7af25293a/src/query/sharing/src/share_endpoint.rs#L150)):*
> ```XML
>                         share_name, e
>                     ))),
>                 }
> ```

A json decoding error should be considered as internal error IMHO. `UnknownXXX` should be used only when an object does not exist, not when an object is damaged.
___
*[`src/query/sql/src/planner/binder/table.rs` line 263 at r2](https://reviewable.io/reviews/datafuselabs/databend/14076#-Nm3zZHq8okNK2URoodr:-Nm3zZHq8okNK2URoods:b-gpnhyk) ([raw file](https://github.com/datafuselabs/databend/blob/f236a6ac371a6a595c31dbe10e4068a7af25293a/src/query/sql/src/planner/binder/table.rs#L263)):*
> ```XML
>                     parent = bind_context.parent.as_mut();
>                 }
>                 if e.code() == 1025 {
> ```

Use named const `ErrorCode::UNKNOWN_TABLE` instead of integer.

Does it need to deal with other causing errors such as `UnknownDatabase`? There are several other errors needs to be considered: `WrongShare`, `WrongShareObject`, `UnknownTableId`

xudong963:
> A json decoding error should be considered as internal error IMHO. `UnknownXXX` should be used only when an object does not exist, not when an object is damaged.

cc @lichuang, how about opening a new PR to process here?

xudong963:
@lichuang will open a new PR

PR #14111, opened 2023-12-21
chore: replace `BuffedBasicDecompressor` with `BasicDecompressor`

dantengsky:

## Summary

`BuffedBasicDecompressor` assumes single thread access, but not sure if this can be assured anymore. 

Try replacing it with `BasicDecompressor`,  let's see if any performance regressions.

Fixes #[Link the issue here]

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [x] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14111)


BohuTANG:
Seems the performance is not degression.

JackTan25:
we can add a code below
```rust
 info!("One Column Rows Count:{}", pages.count());
```
at https://github.com/datafuselabs/databend/blob/c1e100338dbaf0fd51a9fe5a937188e00bc7570f/src/query/storages/fuse/src/io/read/block/block_reader_parquet_deserialize.rs#L212 , I suspect we have got zero data here. cc @dantengsky

dantengsky:
> we can add a code below
> 
> ```rust
>  info!("One Column Rows Count:{}", pages.count());
> ```
> 
> at
> 
> https://github.com/datafuselabs/databend/blob/c1e100338dbaf0fd51a9fe5a937188e00bc7570f/src/query/storages/fuse/src/io/read/block/block_reader_parquet_deserialize.rs#L212
> 
> , I suspect we have got zero data here. cc @dantengsky

yeah, feeling the same.  Something bad might happen during the decompression process.

let me check this PR  out with @wubx today.

PR #14140, opened 2023-12-25
refactor: make sync agg index default

ariesdevil:

## Summary

Make sync agg index default, change the optional `sync` to `async`

Fixes #[Link the issue here]

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14140)

### DIFF
--- a/src/query/ast/src/parser/statement.rs
+++ b/src/query/ast/src/parser/statement.rs
@@ -924,17 +924,17 @@ pub fn statement(i: Input) -> IResult<StatementWithFormat> {
 
     let create_index = map(
         rule! {
-            CREATE ~ SYNC? ~ AGGREGATING ~ INDEX ~ ( IF ~ ^NOT ~ ^EXISTS )?
+            CREATE ~ ASYNC? ~ AGGREGATING ~ INDEX ~ ( IF ~ ^NOT ~ ^EXISTS )?
             ~ #ident
             ~ AS ~ #query
         },
-        |(_, opt_sync, _, _, opt_if_not_exists, index_name, _, query)| {
+        |(_, opt_async, _, _, opt_if_not_exists, index_name, _, query)| {
             Statement::CreateIndex(CreateIndexStmt {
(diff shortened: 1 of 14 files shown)


src/query/storages/fuse/src/operations/agg_index_sink.rs
```diff
@@ -68,12 +77,36 @@ impl AggIndexSink {
...
             location_data: HashMap::new(),
             blocks: vec![],
+            max_memory_usage,
         });
 
         Ok(ProcessorPtr::create(sinker))
     }
 
-    fn process_block(&mut self, block: &mut DataBlock) {
+    fn check_memory_limit(&self, block: &DataBlock) -> Result<()> {
```
BohuTANG, line 86: We have the soft oom by default, does it also working?

ariesdevil: Yup it works, I have add tests here: https://github.com/datafuselabs/databend/pull/14140/files#diff-4001267ac227fed7b868cb28249ed3020c48bf6c843cb120e213ac816cdbd092

BohuTANG: So we don't need `check_memory_limit`?

ariesdevil: If we don't check memory here, how do we return the corresponding error message?

BohuTANG: No need return, we have many places to return soft oom.

BohuTANG: This is the soft oom error:
image

RinChanNOWWW (approved):

Issue #14069, opened 2023-12-18
Feature: `array_fold` function with lambda expression


sundy-li:
**Summary**

Description for this feature.

similar to  rust's `fold` function: 
```
 let numbers = [1, 2, 3, 4, 5];

 let mut result = 0;

 // for loop:
 for i in &numbers {
     result = result + i;
 }

 // fold:
 let result2 = numbers.iter().fold(0, |acc, &x| acc + x);
```

in databend, for example:
```sql
select array_fold([], (acc, t) -> acc + t);   --- outputs null
select array_fold([1,2,3], (acc, t) -> acc + t);   --- outputs 6
```

b41sh:
The number of arguments should be the same as the number of params. The example SQL should look like this:
```sql
select array_fold(0, [], (acc, t) -> acc + t);   --- outputs null
select array_fold(0, [1,2,3], (acc, t) -> acc + t);   --- outputs 6
```
See the Clickhouse document:
https://clickhouse.com/docs/en/sql-reference/functions/array-functions#arrayfold

sundy-li:
> The number of arguments should be the same as the number of params. The example SQL should look like this:
> 
> ```sql
> select array_fold([], 0, (acc, t) -> acc + t);   --- outputs null
> select array_fold([1,2,3], 0, (acc, t) -> acc + t);   --- outputs 6
> ```
> 
>

Ok, Let's use this, cc @Freejww

sundy-li:
Let's use rust's style, the acc will default to the first element instead of a new arg.

PR #14143, opened 2023-12-25
feat(query): Add array_reduce function with lambda expression

Freejww:

## Summary

_similar to rust's reduce function, we add array_reduce function._

closes: #14069

### example_usage
```
SELECT array_reduce([3, 4],(x,y)->x+y);
+---------------------------------------+
| array_reduce([3, 4], (x, y) -> (x + y)) |
+---------------------------------------+
|                                     7 |
+---------------------------------------+

SELECT array_reduce([1,2,3],(x,y)->x+y-1);
+------------------------------------------------+
| array_reduce([1, 2, 3], (x, y) -> ((x + y) - 1)) |
+------------------------------------------------+
|                                              4 |
+------------------------------------------------

SELECT * FROM t;
+----------+
| col      |
+----------+
| [1]      |
| [1,2]    |
| []       |
| [1,2,3]  |
| [1,NULL] |
+----------+

SELECT array_reduce(col,(x,y)->x+y) FROM t;
+------------------------------------+
| array_reduce(col, (x, y) -> (x + y)) |
+------------------------------------+
|                                  1 |
|                                  3 |
|                               NULL |
|                                  6 |
|                               NULL |
+------------------------------------+
```

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14143)


src/query/expression/src/evaluator.rs
```diff
@@ -1000,6 +1054,11 @@ impl<'a> Evaluator<'a> {
                     },
                     _ => unreachable!(),
                 };
+
+                if func_name == "array_fold" {
+                    return self.run_array_fold(&inner_col, &expr);
```
b41sh, line 1059: `inner_col` can't be used directly in `run_array_fold` because it's an expanded column, the offset information is missing, and we can't construct a result column without offsets.

src/query/expression/src/evaluator.rs
```diff
@@ -951,6 +951,56 @@ impl<'a> Evaluator<'a> {
         unreachable!("expr is not a set returning function: {expr}")
     }
 
+    fn run_array_fold(&self, column: &Column, expr: &Expr) -> Result<Value<AnyType>> {
+        let first_element = column.index(0).unwrap();
```
b41sh, line 955: We need to check the length of the column first to avoid a panic if the first value doesn't exist.

src/query/expression/src/evaluator.rs
```diff
@@ -951,6 +951,56 @@ impl<'a> Evaluator<'a> {
...
+        )?;
+        let mut arg0 = cast_value.as_scalar().unwrap().as_ref();
+        let mut result;
+
+        for i in 1..column.len() {
+            let arg1 = column.index(i).unwrap();
+            let entries = if expr.data_type().is_nullable() {
+                vec![
+                    BlockEntry::new(
+                        if arg0.is_null() {
```
b41sh, line 971: If the argument is null, we should ignore it, or report an error.

src/query/ast/src/util.rs
```diff
@@ -72,6 +72,18 @@ pub fn any_token(i: Input) -> IResult<&Token> {
     }
 }
 
+pub fn lambda_params(i: Input) -> IResult<Vec<Identifier>> {
+    let single_param = map(rule! {#ident}, |param| vec![param]);
+    let multi_params = map(
+        rule! { "(" ~ #ident ~ "," ~ #ident ~ ")" },
```
b41sh, line 78: use `comma_separated_list1` here, because we may have more than two params.

array_fold.txt
```diff
@@ -0,0 +1,38 @@
+select array_fold([],(x,y)->x+y-1);
+
+select array_fold([null],(x,y)->x+y-1);
```
b41sh, line 3: Test cases should be written in this format
```
query T
select array_transform([1, 2, NULL, 3], x -> x + 1)
----
[2,3,NULL,4]
```
You can see more examples of `array_transform` and `array_filter` here
https://github.com/datafuselabs/databend/blob/main/tests/sqllogictests/suites/query/02_function/02_0061_function_array.test#L208
and please delete this text file and write them in the above test file, you can run `make sqllogic-test` in your local environment to see the result.

src/query/expression/src/block.rs
```diff
@@ -624,7 +624,12 @@ fn check_type(data_type: &DataType, value: &Value<AnyType>) {
                 }
             }
         }
-        Value::Scalar(s) => assert_eq!(s.as_ref().infer_data_type(), data_type.remove_nullable()),
-        Value::Column(c) => assert_eq!(&c.data_type(), data_type),
+        Value::Scalar(s) => assert!(
```
b41sh, line 627: Don't modify the assert statement here

src/query/expression/src/evaluator.rs
```diff
@@ -962,6 +1015,12 @@ impl<'a> Evaluator<'a> {
         match &args[0] {
             Value::Scalar(s) => match s {
                 Scalar::Array(c) => {
+                    if func_name == "array_fold" {
+                        let result = self.run_array_fold(c, &expr)?;
+                        let result_col = result.convert_to_full_column(expr.data_type(), 1);
+                        return Ok(Value::Scalar(Scalar::Array(result_col)));
```
b41sh, line 1021: `array_fold` is an aggregation operation where the return value is no longer an array type

src/query/expression/src/evaluator.rs
```diff
@@ -951,13 +951,96 @@ impl<'a> Evaluator<'a> {
         unreachable!("expr is not a set returning function: {expr}")
     }
 
+    fn run_array_fold(&self, column: &Column, expr: &Expr) -> Result<Value<AnyType>> {
```
b41sh, line 954: Return `Result<Scalar>` here, so that we don't need to convert the value back to scalar in `run_lambda`

src/query/expression/src/evaluator.rs
```diff
@@ -951,13 +951,96 @@ impl<'a> Evaluator<'a> {
...
+        if column.data_type().is_null() || column.len() < 1 {
+            return Ok(Value::Scalar(Scalar::Null));
+        }
+
+        let first_element = column.index(0).unwrap();
+        let cast_value = self.run_cast(
```
b41sh, line 960: We can add a `ScalarExpr::CastExpr` for the argument in the `resolve_lambda_function` to convert the original column data to the target type without having to explicitly call the `run_cast` here

src/query/expression/src/evaluator.rs
```diff
@@ -951,13 +951,96 @@ impl<'a> Evaluator<'a> {
...
+        let mut arg0 = cast_value.as_scalar().unwrap().as_ref();
+        let mut result = cast_value.to_owned();
+
+        for i in 1..column.len() {
+            let arg1 = column.index(i).unwrap();
+            let entries = if expr.data_type().is_nullable() {
+                vec![
+                    BlockEntry::new(
+                        if arg0.is_null() {
+                            DataType::Null
```
b41sh, line 976: the data type can get by `column.data_type()`, don't need to check by `arg0` every time.

sundy-li:
Please resolve the conflicts and pass all the tests, and then you can send review requests.  Otherwise, the pr should be taken as draft.

src/query/expression/src/evaluator.rs
```diff
@@ -951,13 +951,75 @@ impl<'a> Evaluator<'a> {
...
+                .run_cast(
+                    None,
+                    &arg0.as_ref().infer_data_type(),
+                    &col_type,
+                    Value::Scalar(arg0.clone()),
+                    None,
+                )?
+                .as_scalar()
+                .unwrap()
+                .clone();
```
b41sh, line 981:
```suggestion
        let result = evaluator.run(expr)?;
        arg0 = self
            .run_cast(
                None,
                expr.data_type(),
                &col_type,
                result,
                None,
            )?
            .as_scalar()
            .unwrap()
            .clone();
```

src/query/expression/src/evaluator.rs
```diff
@@ -951,13 +951,75 @@ impl<'a> Evaluator<'a> {
...
+                Value::Scalar(s) => match s {
+                    Scalar::Array(c) => {
+                        let result = self.run_array_reduce(c, &expr)?;
+                        return Ok(Value::Scalar(result));
+                    }
+                    _ => unreachable!(),
+                },
+                Value::Column(c) => {
+                    let mut builder =
+                        ColumnBuilder::with_capacity(&expr.data_type().wrap_nullable(), c.len());
```
b41sh, line 1005: As we cast the data type of expr to the argument inner data type, the data type of builder should be `c.data_type()`

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1738,7 +1774,8 @@ impl<'a> TypeChecker<'a> {
             .collect::<Vec<_>>();
 
         // TODO: support multiple params
-        if params.len() != 1 {
+        // ARRAY_REDUCE have two params
+        if params.len() != 1 && func_name != "array_reduce" {
```
b41sh, line 1778: We also need to check the length of params equal to 2, if the `func_name` is `array_reduce`

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1761,8 +1798,21 @@ impl<'a> TypeChecker<'a> {
...
+        let inner_tys = if func_name == "array_reduce" {
+            vec![inner_ty.clone(), inner_ty.clone()]
+        } else {
+            vec![inner_ty.clone()]
+        };
+
+        let columns = params
+            .iter()
+            .zip(inner_tys.iter())
+            .map(|col| (col.0.clone(), col.1.clone()))
```
b41sh, line 1811:
```suggestion
            .map(|(col, ty)| (col.clone(), ty.clone()))
```

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1772,10 +1822,30 @@ impl<'a> TypeChecker<'a> {
                     "invalid lambda function for `array_filter`, the result data type of lambda function must be boolean".to_string()
                 ));
             }
+        } else if func_name == "array_reduce" {
+            let max_ty = self.transform_to_max_type(&lambda_type)?;
```
b41sh, line 1826: We should transform the `inner_ty` to max type, not the `lambda_type`

tests/sqllogictests/suites/query/02_function/02_0061_function_array.test
```diff
@@ -264,5 +264,30 @@ select col1 from t where get(array_transform(col1, x -> x + 1), 1) = 2
...
+query T
+select array_reduce(['hello', 'world'], (x,y) -> x || y)
+----
+helloworld
+
+query T
+select array_reduce(col1,(x,y)->x+y+1) from t
+----
+12
+
```
b41sh, line 291: add some unit tests for float type and decimal type

tests/sqllogictests/suites/query/02_function/02_0069_experimental_expr.test
```diff
@@ -15,7 +15,7 @@ OPENAI IS NOTHING WITHOUT ITS PEOPLE
 query IIII
 with t(f) as (select '11|open|22|ai|33|is nothing without sam'.split('|')
         .array_filter(x -> try_cast(x as int) is not null ).array_apply(x -> x::int ).array_apply(x -> x::int64 ))
-	select f.array_reduce('sum'), f.array_reduce('max'), f.array_reduce('min')  from t;
```
b41sh, line 18: `array_reduce` is probably already used by some users, so it's not a good idea to rename it.

b41sh (approved):

sundy-li:
ci is broken.
```
[SQL] select array_reduce([],(x,y)->x+y), array_reduce([null],(x,y)->x+y), array_reduce(null,(x,y)->x+y);
[Diff] (-expected|+actual)
-   [] null null
+   [] NULL NULL
```

sundy-li (approved):

PR #14149, opened 2023-12-26
feat: copy from parquet support missing_field_as.

youngsofun:

## Summary

Sometimes we need to read parquet files with schema evolution, the most common case is adding new columns.
to accepts parquet files with different schema, we can not use one reader for all files, many code logic changes, so we have to introduce a new `ParquetTableForCopy`. the pr is safe for new table is used only when `file_format=(type=parquet missing_field_as='field_default')`, the code for origin ParquetRSTable is not changed.

later, the whole copy/transform from parquet will switch to the new reader,  
the ParquetRSTable will be used only for query. 

the first version missing the following features:

- Transform
- Databend types like JSON/Bitmap
- optimize for small files

as more feature get implemented, we may change related features (e.g. options) and do more refactor,
so `missing_field_as='field_default'` is not written to meta (can not used it when create stage/file_format) for now.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14149)


src/query/storages/parquet/src/parquet_rs/meta.rs
```diff
@@ -142,14 +189,44 @@ pub async fn read_parquet_metas_batch(
...
+            meta,
+            row_group_level_stats: None,
+        }));
+    }
+    check_memory_usage(max_memory_usage)?;
+    Ok(metas)
+}
+
+// TODO(parquet): how to limit the memory when running this method is to be determined.
+fn check_memory_usage(max_memory_usage: u64) -> Result<()> {
```
ariesdevil, line 223: This may be useless, @BohuTANG just told me and after I tried, I found it true.

BohuTANG: I think no need to check this, it is duplicate with the soft oom in the global runtime tracker:
https://github.com/datafuselabs/databend/blob/ecf1c0df3d930d1c140356965f25409e2a2957c9/src/common/base/src/runtime/runtime_tracker.rs#L155-L165

BohuTANG: Oh, please ignore my comments:
`if max_memory_usage as i64 - used < 100 * 1024 * 1024` is more strict than the global checker.

ariesdevil (approved):

wubx:
@soyeric128  There an enhanced parameter is provided here for parquet file loading. file_format=(type=parquet **missing_field_as='field_default'**)

soyeric128:
> @soyeric128 There an enhanced parameter is provided here for parquet file loading. file_format=(type=parquet **missing_field_as='field_default'**)

`missing_field_as` can be assigned other values except "field_default"? @youngsofun

youngsofun:
> > @soyeric128 There an enhanced parameter is provided here for parquet file loading. file_format=(type=parquet **missing_field_as='field_default'**)
> 
> `missing_field_as` can be assigned other values except "field_default"? @youngsofun

`field_default ｜ error`

PR #14163, opened 2023-12-26
refactor(planner): Correct usage of exchange operator

leiysky:

## Summary

This pull request involves the restructuring of logic related to the exchange operator and random distribution.

In a cluster environment, where table data is partitioned across query nodes based on a specified scheme, each `TableScan` operator can inherently offer the `Distribution::Random` property. Consequently, there's no longer a necessity for an `Exchange::Random` operator to consistently enforce the `Distribution::Random` property.

Following the implementation of this pull request, the union-all and broadcast join operators will no longer be required to enforce the `Distribution::Random` for their respective children.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14163)

### DIFF
--- a/src/query/sql/src/planner/plans/join.rs
+++ b/src/query/sql/src/planner/plans/join.rs
@@ -488,42 +488,37 @@ impl Operator for Join {
         let probe_physical_prop = rel_expr.derive_physical_prop_child(0)?;
         let build_physical_prop = rel_expr.derive_physical_prop_child(1)?;
 
-        // if join/probe side is Serial, we use Serial distribution
+        // if join/probe side is Serial or this is a non-equi join, we use Serial distribution
         if probe_physical_prop.distribution == Distribution::Serial
             || build_physical_prop.distribution == Distribution::Serial
+            || (self.left_conditions.is_empty()
+                && self.right_conditions.is_empty()
+                && !self.non_equi_conditions.is_empty())
         {
             // TODO(leiysky): we can enforce redistribution here
             required.distribution = Distribution::Serial;
             return Ok(required);
         }
 
-        // If there is no equi-condition, this is essentially a cross join.
-        if (self.left_conditions.is_empty() && self.right_conditions.is_empty())
-            || self.join_type == JoinType::Cross
-        {
-            if child_index == 0 {
-                required.distribution = Distribution::Any;
-            } else {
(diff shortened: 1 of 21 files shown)


BohuTANG:
```
0: query failed: mysql client error: Server error: `ERROR HY000 (1105): Unimplemented. Code: 1002, Text = Exchange source and no from multiple nodes is unimplemented.
(while in query flight).'
[SQL] WITH
x AS (SELECT number AS a FROM numbers(10)),
y AS (SELECT number AS a FROM numbers(5)),
z AS (SELECT * FROM x WHERE a % 2),
w AS (SELECT * FROM y WHERE a > 0)
SELECT x.a FROM x JOIN y USING (a) WHERE x.a in (SELECT * FROM z)
HAVING x.a <= (SELECT max(a) FROM w)
ORDER BY x.a
at tests/sqllogictests/suites/query/cte.test:239

```

andylokandy (approved):

src/query/sql/src/planner/plans/join.rs
```diff
@@ -486,33 +488,48 @@ impl Operator for Join {
...
+        ) {
             let left_stat_info = rel_expr.derive_cardinality_child(0)?;
             let right_stat_info = rel_expr.derive_cardinality_child(1)?;
             // The broadcast join is cheaper than the hash join when one input is at least (n − 1)× larger than the other
             // where n is the number of servers in the cluster.
-            let broadcast_join_threshold = (ctx.get_cluster().nodes.len() - 1) as f64;
+            let broadcast_join_threshold = if ctx.get_settings().get_prefer_broadcast_join()? {
+                (ctx.get_cluster().nodes.len() - 1) as f64
+            } else {
+                // Use a very large value to prevent broadcast join.
```
xudong963, line 519: Why this?

leiysky: The setting `prefer_broadcast_join` is enabled by default, and should be deprecated in the future.

And by the way, the `prefer_broadcast_join` looks like even it's `false` there is still a chance to use broadcast join, but essentially it will reject broadcast join.

PR #14164, opened 2023-12-27
refactor: build_block method

xudong963:

## Summary

1. No need to assign value to `num_rows` more than once.
2. Add assert to ensure each array has the same length.

Fixes #[Link the issue here]

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14164)

### DIFF
--- a/src/query/storages/fuse/src/io/read/block/block_reader_native.rs
+++ b/src/query/storages/fuse/src/io/read/block/block_reader_native.rs
@@ -244,10 +244,7 @@ impl BlockReader {
                 }
             }
         }
-        match nums_rows {
-            Some(nums_rows) => Ok(DataBlock::new(entries, nums_rows)),
-            None => Ok(DataBlock::empty()),
-        }
+        Ok(DataBlock::new(entries, nums_rows.unwrap_or(0)))
     }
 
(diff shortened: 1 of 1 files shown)


src/query/storages/fuse/src/io/read/block/block_reader_native.rs
```diff
@@ -239,7 +244,10 @@ impl BlockReader {
                 }
             }
         }
-        Ok(DataBlock::new(entries, nums_rows))
+        match nums_rows {
+            Some(nums_rows) => Ok(DataBlock::new(entries, nums_rows)),
+            None => Ok(DataBlock::empty()),
```
b41sh, line 249: If the chunks are empty, we will push `default_val` to `entries`, in this case, `DataBlock` is not empty.

xudong963: fixed

b41sh (approved):

PR #14174, opened 2023-12-27
feat(query): timestamp type can modify output format with new setting TIMESTAMP_OUTPUT_FORMAT

### LINKED ISSUE #14169: Feature: Databend needs to distinguish between DateTime and Timestamp
**Summary**

Generally the format of DateTime type is `2016-06-15 23:00:00` and the Timestamp is `2016-06-15 23:00:00.000000`. So IMHO, Databend should distinguish between these two time type.

TCeason:

## Summary

timestamp type can modify output format with new setting TIMESTAMP_OUTPUT_FORMAT

About new setting

```sql
+-------------------------+-------------------------+-------------------------+----------------------------------------------------+---------+-----------------------------------------------------------------------------------------------------------------------------------------------+--------+
| name                    | value                   | default                 | range                                              | level   | description                                                                                                                                   | type   |
+-------------------------+-------------------------+-------------------------+----------------------------------------------------+---------+-----------------------------------------------------------------------------------------------------------------------------------------------+--------+
| timestamp_output_format | '%Y-%m-%d %H:%M:%S%.6f' | '%Y-%m-%d %H:%M:%S%.6f' | ["%Y-%m-%d %H:%M:%S%.3f", "%Y-%m-%d %H:%M:%S%.6f"] | SESSION | Specifies the display format for the TIMESTAMP data type alias. Available values include '%Y-%m-%d %H:%M:%S%.6f' and '%Y-%m-%d %H:%M:%S%.3f'. | String |
+-------------------------+-------------------------+-------------------------+----------------------------------------------------+---------+-----------------------------------------------------------------------------------------------------------------------------------------------+--------+

```

Fixes #14169 

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14174)

### DIFF
--- a/src/query/expression/src/function.rs
+++ b/src/query/expression/src/function.rs
@@ -35,6 +35,7 @@ use crate::property::FunctionProperty;
 use crate::type_check::try_unify_signature;
 use crate::types::nullable::NullableColumn;
 use crate::types::nullable::NullableDomain;
+use crate::types::timestamp::TIMESTAMP_FORMAT;
 use crate::types::*;
 use crate::values::Value;
 use crate::values::ValueRef;
@@ -91,9 +92,10 @@ pub enum FunctionEval {
     },
 }
 
-#[derive(Clone, Default)]
+#[derive(Clone)]
 pub struct FunctionContext {
     pub tz: TzLUT,
+    pub ts_output_format: String,
     pub rounding_mode: bool,
 
     pub openai_api_chat_base_url: String,
@@ -105,6 +107,24 @@ pub struct FunctionContext {
 
     pub external_server_connect_timeout_secs: u64,
     pub external_server_request_timeout_secs: u64,
+}
+
+impl Default for FunctionContext {
+    fn default() -> Self {
+        Self {
+            tz: Default::default(),
+            ts_output_format: TIMESTAMP_FORMAT.to_string(),
+            rounding_mode: false,
+            openai_api_chat_base_url: "".to_string(),
+            openai_api_embedding_base_url: "".to_string(),
+            openai_api_key: "".to_string(),
+            openai_api_version: "".to_string(),
+            openai_api_embedding_model: "".to_string(),
+            openai_api_completion_model: "".to_string(),
+            external_server_connect_timeout_secs: 0,
+            external_server_request_timeout_secs: 0,
+        }
+    }
 }
 
 #[derive(Clone)]
--- a/src/query/expression/src/types/timestamp.rs
+++ b/src/query/expression/src/types/timestamp.rs
@@ -238,6 +238,7 @@ pub fn string_to_timestamp(ts_str: impl AsRef<[u8]>, tz: Tz) -> Option<DateTime<
 
 #[inline]
 pub fn timestamp_to_string(ts: i64, tz: Tz, ts_format: &str) -> impl Display + '_ {
+    debug_assert!(!ts_format.trim().is_empty());
     // 1. If write a wrong format, .format(ts_format) will display the ts_format directly.
     // let date_time: DateTime<Utc> = Utc.with_ymd_and_hms(2017, 04, 02, 12, 50, 32).unwrap();
     // let formatted = format!("{}", date_time.format("cc"));
--- a/src/query/expression/src/types/timestamp.rs
+++ b/src/query/expression/src/types/timestamp.rs
@@ -237,6 +237,14 @@ pub fn string_to_timestamp(ts_str: impl AsRef<[u8]>, tz: Tz) -> Option<DateTime<
 }
 
 #[inline]
-pub fn timestamp_to_string(ts: i64, tz: Tz) -> impl Display {
-    ts.to_timestamp(tz).format(TIMESTAMP_FORMAT)
+pub fn timestamp_to_string(ts: i64, tz: Tz, ts_format: &str) -> impl Display + '_ {
+    // 1. If write a wrong format, .format(ts_format) will display the ts_format directly.
+    // let date_time: DateTime<Utc> = Utc.with_ymd_and_hms(2017, 04, 02, 12, 50, 32).unwrap();
+    // let formatted = format!("{}", date_time.format("cc"));
+    // assert_eq!(formatted, "cc");
+    // 2. Until now, ts mostly use micros, so %Y-%m-%d %H:%M:%S%.9f or ignore ms part will cause panic.
+    // let micros = ts.timestamp_micros();
+    // check_timestamp(micros)?;
+    // column.push(micros.as_());
+    ts.to_timestamp(tz).format(ts_format)
 }
--- a/src/query/formats/src/file_format_type.rs
+++ b/src/query/formats/src/file_format_type.rs
@@ -15,6 +15,7 @@
 use chrono_tz::Tz;
 use databend_common_exception::ErrorCode;
 use databend_common_exception::Result;
+use databend_common_expression::types::timestamp::TIMESTAMP_FORMAT;
 use databend_common_expression::TableSchemaRef;
 use databend_common_meta_app::principal::FileFormatParams;
 use databend_common_meta_app::principal::StageFileFormatType;
@@ -73,7 +74,6 @@ impl FileFormatOptionsExt {
         settings: &Settings,
     ) -> Result<FileFormatOptionsExt> {
         let timezone = parse_timezone(settings)?;
-        let timestamp_output_format = settings.get_timestamp_output_format()?;
         let mut options = FileFormatOptionsExt {
             ident_case_sensitive: settings.get_unquoted_ident_case_sensitive()?,
             headers: 0,
@@ -82,7 +82,7 @@ impl FileFormatOptionsExt {
             disable_variant_check: false,
             timezone,
             is_select: false,
-            timestamp_output_format,
+            timestamp_output_format: TIMESTAMP_FORMAT.to_string(),
         };
         let suf = &clickhouse_type.suffixes;
         options.headers = suf.headers;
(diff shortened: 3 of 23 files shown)


src/query/expression/src/function.rs
```diff
@@ -94,6 +94,7 @@ pub enum FunctionEval {
 #[derive(Clone, Default)]
 pub struct FunctionContext {
     pub tz: TzLUT,
```
BohuTANG, line 96: Please  use a meaningful  name instead.

TCeason: use timestamp_output_format?

TCeason:
Maybe we can remove this impl in this pr?

https://github.com/datafuselabs/databend/blob/aa97d2c5b57f1585abc8192e5fb76f9fc4157903/src/query/datavalues/src/types/type_timestamp.rs#L37

src/query/expression/src/types/timestamp.rs
```diff
@@ -237,6 +237,16 @@ pub fn string_to_timestamp(ts_str: impl AsRef<[u8]>, tz: Tz) -> Option<DateTime<
...
 
 #[inline]
-pub fn timestamp_to_string(ts: i64, tz: Tz) -> impl Display {
-    ts.to_timestamp(tz).format(TIMESTAMP_FORMAT)
+pub fn timestamp_to_string(ts: i64, tz: Tz, ts_format: &str) -> impl Display + '_ {
+    // If write a wrong format, .format(ts_format) will display the ts_format directly.
+    // let date_time: DateTime<Utc> = Utc.with_ymd_and_hms(2017, 04, 02, 12, 50, 32).unwrap();
+    // let formatted = format!("{}", date_time.format("cc"));
+    // assert_eq!(formatted, "cc");
+    let ts_format = if ts_format != "%Y-%m-%d %H:%M:%S%.6f" && ts_format != "%Y-%m-%d %H:%M:%S%.3f"
```
BohuTANG, line 245: We can not check here, it's a setting check.

TCeason: Ok, I think we can contain this comment, just for the developer.

tests/sqllogictests/suites/base/05_ddl/05_0035_ddl_set_settings.test
```diff
@@ -0,0 +1,35 @@
+#need onlyif mysql tag, because http test can not keep a session
+query T
+show settings like 'timestamp_output_format';
+----
+timestamp_output_format '%Y-%m-%d %H:%M:%S%.6f' '%Y-%m-%d %H:%M:%S%.6f' ["%Y-%m-%d %H:%M:%S%.3f", "%Y-%m-%d %H:%M:%S%.6f"] SESSION Specifies the display format for the TIMESTAMP data type alias. Available values include '%Y-%m-%d %H:%M:%S%.6f' and '%Y-%m-%d %H:%M:%S%.3f'. String
+
+onlyif mysql
```
BohuTANG, line 7: why need `onlyif mysql`?

TCeason: Now sql logic test http handwriting can not store session settings .

But if set global in test file, it maybe cause some flaky test.

tests/sqllogictests/suites/query/case_sensitivity/ddl.test
```diff
@@ -1,3 +1,4 @@
+#need onlyif mysql tag, because http test can not keep a session
```
youngsofun, line 1: http test can keep session settings, each test file is a session.

youngsofun:
date/timestamp format should be part of file formats, not a setting.

TCeason:
> date/timestamp format should be part of file formats, not a setting.
> 
> 1. this is naturally part of a format spec, it should be stable.
> 2. we do not hope users to rely on settings, most used internally (affect how we run it, but not the results).
> 3. during a session,  mutli diff date/timestamp formats may be involved

1. in this issue: https://github.com/datafuselabs/databend/issues/14169 we need support format.
2. sf does the same things. https://docs.snowflake.com/en/sql-reference/data-types-datetime#date-and-time-formats

youngsofun:
> > date/timestamp format should be part of file formats, not a setting.
> > 
> > 1. this is naturally part of a format spec, it should be stable.
> > 2. we do not hope users to rely on settings, most used internally (affect how we run it, but not the results).
> > 3. during a session,  mutli diff date/timestamp formats may be involved
> 
> 1. in this issue: [Feature: Databend needs to distinguish between DateTime and Timestamp #14169](https://github.com/datafuselabs/databend/issues/14169) we need support format.
> 2. sf does the same things. https://docs.snowflake.com/en/sql-reference/data-types-datetime#date-and-time-formats

yes, like sf, we should distinguish `query output` and `format output`,  it is ok for settings to only affect the former.
https://docs.snowflake.com/en/sql-reference/date-time-input-output#label-date-time-input-output-how-snowflake-determines
I think most of the time `query output i`s interpreted to Date/TS type for PY/JAVA/.., instead of used directly, so the ts format for query output is not needed, it is only for transformation and presentation.

tests/sqllogictests/suites/query/case_sensitivity/ddl.test
TCeason: It feels like there are some other unknowns here, and I didn't add the `onlyif mysql` at the beginning. 

But I found that it affected the presentation of other test results.

Let't add it again, and observe the ci result.

BohuTANG:
@youngsofun Yes, I think this is mainly for the third-party SaaS intergration. But now we can not say the superset side issue related to this timestamp format, @hantmac are still invegesting.

src/query/settings/src/settings_default.rs
```diff
@@ -233,6 +233,12 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: Some(SettingRange::String(vec!["PostgreSQL", "MySQL", "Experimental", "Hive"])),
                 }),
+                ("timestamp_output_format", DefaultSettingValue {
+                    value: UserSettingValue::String("%Y-%m-%d %H:%M:%S%.6f".to_owned()),
+                    desc: "Specifies the display format for the TIMESTAMP data type alias. Available values include \'%Y-%m-%d %H:%M:%S%.6f\' and \'%Y-%m-%d %H:%M:%S%.3f\'.",
```
youngsofun, line 238: what does it refer to by  `alias`?

TCeason: Oooo, the alias can be delete.

tests/sqllogictests/suites/base/05_ddl/05_0035_ddl_set_settings.test
```diff
@@ -0,0 +1,34 @@
...
+SET timestamp_output_format='%Y-%m-%d %H:%M:%S%.6f';
+
+
+query T
+select to_timestamp('2022-02-01 00:00:00.123456789');
+----
+2022-02-01 00:00:00.123456
+
+
+statement error 2803
```
youngsofun, line 27: better if check for error message here
current work around `statement error 2803.*part of messges`

TCeason: Err msg in MySQL and http is diff.

src/query/formats/src/file_format_type.rs
```diff
@@ -70,6 +73,7 @@ impl FileFormatOptionsExt {
         settings: &Settings,
     ) -> Result<FileFormatOptionsExt> {
         let timezone = parse_timezone(settings)?;
+        let timestamp_output_format = settings.get_timestamp_output_format()?;
```
youngsofun, line 76: set to TIMESTAMP_FORMAT here too？

TCeason: Ck handler has already disable, why we contain this func?

tests/sqllogictests/suites/base/05_ddl/05_0035_ddl_set_settings.test
youngsofun: yes，the wher string after statement error  is a regrex.
so I use `.*` to handle the diff part of mysql/http handler

src/query/formats/src/file_format_type.rs
youngsofun: disabled in settigns by default, can still be used if enabled

TCeason: Done

src/query/expression/src/types/timestamp.rs
```diff
@@ -237,6 +237,14 @@ pub fn string_to_timestamp(ts_str: impl AsRef<[u8]>, tz: Tz) -> Option<DateTime<
 }
 
 #[inline]
-pub fn timestamp_to_string(ts: i64, tz: Tz) -> impl Display {
-    ts.to_timestamp(tz).format(TIMESTAMP_FORMAT)
+pub fn timestamp_to_string(ts: i64, tz: Tz, ts_format: &str) -> impl Display + '_ {
+    // 1. If write a wrong format, .format(ts_format) will display the ts_format directly.
```
andylokandy, line 241: To avoid uninitialized function context:
```suggestion
    debug_assert!(!ts_format.trim().is_empty());
    // 1. If write a wrong format, .format(ts_format) will display the ts_format directly.
```

BohuTANG:
Confirmed, there is no need add this setting, let's close this PR, thanks all.

PR #14203, opened 2024-01-02
fix(query): http auth query_id should be set into new query ctx

TCeason:

## Summary

http handler will generate a dummy session and query_ctx in auth.

But not reuse this query_id.

Fixes https://github.com/datafuselabs/databend/issues/14172

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14203)

### DIFF
--- a/src/query/service/src/servers/http/middleware.rs
+++ b/src/query/service/src/servers/http/middleware.rs
@@ -204,6 +204,8 @@ impl<E> HTTPSessionEndpoint<E> {
             .map(|id| id.to_str().unwrap().to_string())
             .unwrap_or_else(|| Uuid::new_v4().to_string());
 
+        ctx.set_id(query_id.clone());
+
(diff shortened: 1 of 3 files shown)


youngsofun:
I prefer to ignore such contextID that not  bind to query instead.

TCeason:
> I prefer to ignore such contextID that not bind to query instead (not sure it can work).

If we can do this. I think this way is better. 

This pr is just my temporary idea. Maybe you need to discuss these two issues https://github.com/datafuselabs/databend/issues/14204 https://github.com/datafuselabs/databend/issues/14172 and have a better solution before committing new code.

PR #14235, opened 2024-01-04
refactor: use binary for non-utf8 string type

andylokandy:

## Summary

1. Aggregate state: String -> Binary
2. Bloom index state: String -> Binary
3. Group by hash: String -> Binary
4. Column to arrow: String: LargeBinary -> LargeUtf8
5. `to_base64(String) -> String` -> `to_base64(Binary) -> String`
6. `from_base64(String) -> String` -> `from_base64(String) -> Binary`
7. `to_hex(String) -> String` -> `to_hex(Binary) -> String`
8. `from_hex(String) -> String` -> `from_hex(String) -> Binary`
9. Add utf8 check when: (a) {Binary, LargeBinary, FixLength}Array -> Column::String (2) native format read `DataType::String`

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14235)


src/query/functions/tests/it/scalars/testdata/string.txt
```diff
@@ -499,28 +395,28 @@ output         : ''
 ast            : reverse('你好')
 raw expr       : reverse('你好')
 checked expr   : reverse<String>("你好")
-optimized expr : 0xbda5e5a0bde4
+optimized expr : BDA5E5A0BDE4
```
sundy-li, line 398: seems the output is invalid

tests/sqllogictests/suites/stage/formats/parquet/parquet_field_types.test
```diff
@@ -182,7 +182,7 @@ parquet/json_string.parquet 4 0 NULL NULL
 query 
 select * from t2 order by a
 ----
-1 {"k1":"v1","k2":"v2"}
-2 {"k":{"k1":123,"k2":456}}
-3 [1,2,3]
-4 ["a","b"]
+1 "{\"k1\":\"v1\",\"k2\":\"v2\"}"
```
sundy-li, line 185: binary output should not be quoted.

sundy-li:
tests failed in py-binding

src/query/functions/tests/it/scalars/testdata/string.txt
andylokandy: let's leave the string function update to the next pr.

tests/sqllogictests/suites/stage/formats/parquet/parquet_field_types.test
andylokandy: it's string output, which is a complete string of json.

sundy-li: But in logicaltests, we don't quote string output. Is this change causing incompatibility?

andylokandy: oh, it's variant output with jsonb::Value::String. So it make sense. 
Anyway, a binary should neither be parsed to a json value nor shown as utf8 string, since that I've updated the test case.

tests/sqllogictests/suites/stage/formats/parquet/parquet_field_types.test
```diff
@@ -179,6 +179,22 @@ copy into t2 from @data/parquet/json_string.parquet
...
+4 "5B2261222C2262225D"
+
+statement ok
+delete from t2
+
+query 
+copy into t2 from (select a, parse_json(b::STRING) from @data/parquet/json_string.parquet)
+----
+parquet/json_string.parquet 4 0 NULL NULL
+
```
andylokandy, line 197: Is these appropriate behaviors? @b41sh

andylokandy:
https://github.com/datafuselabs/databend/blob/3709d5f1a01a7c580d732da82934b34a0bd6cd6d/tests/sqllogictests/suites/stage/formats/parquet/parquet_field_types.test#L174-L204

Are these appropriate behaviors? @b41sh

sundy-li:
> select * from @data/parquet/json_string.parquet;

Schema is binary because parquet did not have variant datatype, so we use binary to store variant.
Since we are reading from parquet, it's correct to show as binary.


But if the parquet file is inside fuse engine, it'll show as JSON since we have variant datatype.

andylokandy:
It's ready for review

tests/sqllogictests/suites/stage/formats/parquet/parquet_field_types.test
```diff
@@ -179,6 +179,22 @@ copy into t2 from @data/parquet/json_string.parquet
 ----
 parquet/json_string.parquet 4 0 NULL NULL
 
+query 
+select * from t2 order by a
+----
+1 "7B226B31223A227631222C226B32223A227632227D"
```
b41sh, line 185: The data in `json_string.parquet` is a JSON string, can we treat it as a string type and convert it to variant type by `json_parser` function?

src/common/arrow/src/native/read/array/binary.rs
```diff
@@ -308,7 +309,18 @@ fn try_new_binary_array<O: Offset>(
     validity: Option<Bitmap>,
 ) -> Result<Box<dyn Array>> {
     if matches!(data_type, DataType::Utf8 | DataType::LargeUtf8) {
-        let array = Utf8Array::<O>::try_new(data_type, offsets, values, validity)?;
+        // todo!("new string")
```
b41sh, line 312: Looks like this `todo` can be deleted

src/common/arrow/src/native/read/array/binary.rs
```diff
@@ -308,7 +309,18 @@ fn try_new_binary_array<O: Offset>(
...
+        let array =
+            Utf8Array::<O>::try_new(data_type, offsets, values, validity).map_err(|err| {
+                Error::External(
+                    "Encountered invalid utf8 data for string type, \
+                        if you were reading column with string type from a table, \
+                        it's recommended to alter the column type to `BINARY`.\n\
+                        Example: `ALTER TABLE <table> MODIFY COLUMN <column> BINARY;`"
```
b41sh, line 319: It is better to put this example into the upper level of the relevant function to return, and it is more appropriate to return only the utf8 error here.

tests/sqllogictests/suites/stage/formats/parquet/parquet_field_types.test
b41sh: The query plan for copy into is built in the `bind_copy_into_table_from_location` function, we can add a check source type is binary to deal with this case

andylokandy: I don't think it's a wise behavior. Because we are assuming all LargeBinary is valid uft8 and even valid json strings.

A better choice would be changing the parqute column type to LargeUtf8 and let user call parse_json when using copy into.

src/common/arrow/src/native/read/array/binary.rs
andylokandy: I'm afriad that we can not cover all callee of this function. I prefer to let it be here for a while during the transition period, and removing it until most related users has properly updated.

src/common/arrow/src/native/read/array/binary.rs
andylokandy: try_new will check utf8. i was planning to use try_new_unchecked afterwards.

tests/sqllogictests/suites/stage/formats/parquet/parquet_field_types.test
sundy-li: How about implementing `parse_json(binary)` ?

b41sh: We don't need to support `parse_json(binary)`, and it is ok for users to use utf8 type as the source of variant type. The documentation needs to be updated to add a note to remind users to make sure the types are correct. 
For this test, it would be a good idea to modify the `json_string.parquet` file to change the type from binary to utf8 to make sure it's the right type, copy into a binary type to a variant type should be not using case. @andylokandy

sundy-li: Parquet files could be updated in next pr.

b41sh (approved):

PR #14269, opened 2024-01-08
feat: CSV add option empty_field_as.

youngsofun:

## Summary

`empty_field_as = 'field_default' | 'null' | 'string'`

- `empty` include both `,,` and `,"",` . 
- null: only allow `empty` for nullable fields, decode as null
- string: only allow `empty` for (nullable) string fields, decode as empty string `` 

row like `,,,,,`  is no longer skipped

cc @wubx  

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14269)


youngsofun:
cc @hantmac

src/meta/proto-conv/src/util.rs
```diff
@@ -101,6 +101,7 @@ const META_CHANGE_LOG: &[(u64, &str)] = &[
     (69, "2023-12-21: Add: user.proto/GrantTableIdObject and GrantDatabaseIdObject", ),
     (70, "2023-12-25: Add: datatype.proto Binary type", ),
     (71, "2024-01-02: Add: user.proto/password options", ),
+    (72, "2023-11-16: Add: user.proto/CSVFileFormatParams add field `empty_field_as`", ),
```
b41sh, line 104: The date looks incorrect.

drmingdrmer (changes requested):
> Reviewed 4 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14269)* status: 4 of 22 files reviewed, 1 unresolved discussion (waiting on @b41sh and @youngsofun)

___
*[`src/meta/proto-conv/tests/it/v072_csv_format_params.rs` line 53 at r1](https://reviewable.io/reviews/datafuselabs/databend/14269#-NngDMFQ1YC23NOmLhON:-NngDMFQ1YC23NOmLhOO:b3zmyf5) ([raw file](https://github.com/datafuselabs/databend/blob/8616bee873bf8144b25734ac16e4a16ed7594452/src/meta/proto-conv/tests/it/v072_csv_format_params.rs#L53)):*
> ```XML
>         })
>     };
>     common::test_load_old(func_name!(), file_format_params_v72.as_slice(), 0, want())?;
> ```

The version should not be `0` but it should be `72`.

How did you construct the bytes `file_format_params_v72` to yield a return value of `0`?

These bytes should be copied from the output of the following case `common::test_pb_from_to()`

src/query/pipeline/sources/src/input_formats/impls/input_format_csv.rs
```diff
@@ -57,14 +59,46 @@ impl InputFormatCSV {
...
             match default_values {
                 None => {
+                    // query
                     builder.push_default();
                 }
                 Some(values) => {
-                    builder.push(values[column_index].as_ref());
+                    let field = &schema.fields()[column_index];
+                    // copy
+                    match empty_filed_as {
```
b41sh, line 73: Does `empty_filed_as` not work without `default_values`?

drmingdrmer (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14269)* status: 4 of 22 files reviewed, 3 unresolved discussions (waiting on @b41sh and @youngsofun)

___
*[`src/meta/proto-conv/tests/it/v072_csv_format_params.rs` line 53 at r1](https://reviewable.io/reviews/datafuselabs/databend/14269#-NngDMFQ1YC23NOmLhON:-NngG1zn5_imhiy6HROF:bqrksvd) ([raw file](https://github.com/datafuselabs/databend/blob/8616bee873bf8144b25734ac16e4a16ed7594452/src/meta/proto-conv/tests/it/v072_csv_format_params.rs#L53)):*

The version should not be `0` but it should be `72`.

How did you construct the bytes `file_format_params_v72` to yield a return value of `0`?

These bytes should be copied from the output of the following case `common::test_pb_from_to()`
</blockquote></details>

Oh I see, FileFormatParams does not store a version in it.

drmingdrmer (approved):
> Reviewed 6 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14269)* status: 10 of 22 files reviewed, 2 unresolved discussions (waiting on @b41sh and @youngsofun)

src/query/pipeline/sources/src/input_formats/impls/input_format_csv.rs
youngsofun: thanks,   I will do it in next pr.

1. default_values = None only when querying stage (if I am not wrong), not copy.  (no dest table => no field_defaults),  empty_filed_as is ignored for now
2. when querying csv,  all field `$1 $2 ...` are nullable string.  and query not need to be to strict?  the consideration is a little different.  I

src/meta/proto-conv/src/util.rs
youngsofun: fixed, thanks!

youngsofun:
> Oh I see, FileFormatParams does not store a version in it.

It is better to have store a version? Is it too late to add it now?

b41sh (approved):

sundy-li (approved):

wubx:
cc  @soyeric128

soyeric128:
> cc @soyeric128

What's the default value of empty_field_as? @youngsofun

PR #14281, opened 2024-01-09
chore: use raw sql as task SQL instead of pretty one

ZhiHanZ:

## Summary

Use raw formatted SQL instead of prettied one.
Previous, task SQL used the pretty_statement function and generate prettified SQL for further cloud service execution, but it seems like there are some SQL may parse in different way and unable to execute anymore. for example
```sql
copy into @stream_stage from (
        SELECT
            CASE
                WHEN (RAND() * 10)::INT % 4 = 0 THEN 'a'
                WHEN (RAND() * 10)::INT % 4 = 1 THEN 'b'
                WHEN (RAND() * 10)::INT % 4 = 2 THEN 'c'
                ELSE 'd'
            END AS level
        FROM numbers(500000)
    ) FILE_FORMAT = (
        TYPE = PARQUET
    );
```

its pretty formatted string is 
```sql
COPY INTO Stage("stream_stage")
FROM (
    SELECT CASE
        WHEN (((RAND() * 10)::Int32 % 4) = 0) THEN 'a',
        WHEN (((RAND() * 10)::Int32 % 4) = 1) THEN 'b',
        WHEN (((RAND() * 10)::Int32 % 4) = 2) THEN 'c',
        ELSE 'd'
    END AS level
    FROM numbers(500000)
)
FILE_FORMAT = (
    type = "PARQUET"
)
SINGLE = false;

```

which is unable to execute

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14281)

### DIFF
--- a/src/query/ast/src/parser/statement.rs
+++ b/src/query/ast/src/parser/statement.rs
@@ -129,9 +129,7 @@ pub fn statement(i: Input) -> IResult<StatementWithFormat> {
             _,
             sql,
         )| {
-            let sql = pretty_statement(sql.stmt, 10)
-                .map_err(|_| ErrorKind::Other("invalid statement"))
-                .unwrap();
+            let sql = format!("{}", sql.stmt);
             Statement::CreateTask(CreateTaskStmt {
                 if_not_exists: opt_if_not_exists.is_some(),
                 name: task.to_string(),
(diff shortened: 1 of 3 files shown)


ZhiHanZ:
some server seems down Warning: Failed to download action 'https://api.github.com/repos/everpcpc/comment-on-pr-action/tarball/9e67f4c8556ee21231278950959fd4c97a5ccd0b'. Error: Response status code does not indicate success: 500 (Internal Server Error).
Warning: Back off 10.52 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/everpcpc/comment-on-pr-action/tarball/9e67f4c8556ee21231278950959fd4c97a5ccd0b'. Error: Response status code does not indicate success: 500 (Internal Server Error).
Warning: Back off 16.585 seconds before retry.
Error: Response status code does not indicate success: 500 (Internal Server Error).

Issue #14224, opened 2024-01-04
Meta-service returns the leader id and address if it is a follower.


drmingdrmer:
Meta-service returns the leader id and address if it is a follower and the request is forwarded to the leader.

The meta-client that receives such a "forward-to-leader" header will be able to send request directly to the leader for next time.

PR #14288, opened 2024-01-10
feat: return latest leader addr if request is forwarded

drmingdrmer:

##### feature: return latest leader addr if request is forwarded

New feature, compatible change.

When a meta-service follower receives a request, it forwards it to the
leader. This mechanism introduces additional latency.

In this commit:

- Meta-service responds the leader-address in gRPC
  header(response metadata): "x-databend-meta-leader-grpc-endpoint",
  if the meta-service node is a follower.

- Meta-client that receives such a reponse metadate then updates the
  address for further connection to this perceived leader.

- Leader endpoint response is added to two API: `transaction()` and
  `kv_read_v1()`, which are the most used by `SchemaApi`, and very
  latency sensitive.

  `upsert()` and `kv_read()`(v0 API) are not changed: `upsert()` will be
  replaced with `transaction()`, and `kv_read()` is only used by former
  versions databend-query, which are not capable of parsing leader
  endpoint response.

Detailed Changes

- Add `Endpoints` as a container that stores all of the meta-service
  endpoints for a meta-client.

- Updating leader endpoint is done in `EstablishedClient`, which is a
  wrapper for underlying gRPC client. `HandleRPCResult::update_client()`
  is responsible to update leader endpoint and mark a client as **error**.

- The `endpoints` now is shared between `MetaChannelManager` and
  `EstablishedClient`, `MetaChannelManager` uses `endpoints` for
  connecting, `EstablishedClient` uses it for updating.

- The old health based node choosing algorithm is discarded, because the
  responded leader must be in good healthy state, otherwise it will
  step down.

- Add timing and logging to meta-service API.

---

- Fix: #14224

## Changelog

- New Feature

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14288)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2944,6 +2944,7 @@ dependencies = [
  "databend-common-tracing",
  "derive_more",
  "futures",
+ "itertools 0.10.5",
  "log",
  "logcall",
  "minitrace",
--- a/src/binaries/metabench/main.rs
+++ b/src/binaries/metabench/main.rs
@@ -18,7 +18,6 @@ use std::collections::BTreeMap;
 use std::fmt::Debug;
 use std::fmt::Display;
 use std::sync::Arc;
-use std::time::Duration;
 use std::time::Instant;
 
 use chrono::Utc;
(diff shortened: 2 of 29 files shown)


ariesdevil (approved):

drmingdrmer (commented):
> Reviewed 27 of 28 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14288)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lichuang)

PR #14318, opened 2024-01-14
feat(planner): implement cost-based optimization for distributed query

### LINKED ISSUE #11206: Cost-based distributed query optimization
So far, we are enforcing the data exchange in a heuristic fashion, which may produce suboptimal plans. We can leverage the cascades optimizer to perform cost-based optimization for the distributed query to generate a better plan.

leiysky:

## Summary

This PR implemented cost-based optimization of distributed query.

It allows us to enumerate all possible data distribution required by the operators(e.g. joins), cost them, and choose the best one with lowest cost in a top-down fashion.

The major idea of this PR is combining the `find_best_plan` requests with specific `RequiredProp` to exploit the memoization of cascades optimizer.

close #11206

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14318)


leiysky:
The current implementation will cause OOM when the search space is very large, I'll implement a simple pruning mechanism to fix this.

BohuTANG:
Databend(Release@v1.2.297-nightly) bench result has some issues, re-run them again.

scripts/ci/ci-run-sqllogic-tests-cluster.sh
```diff
@@ -21,4 +21,4 @@ fi
 echo "Run suites using argument: $RUN_DIR"
 
 echo "Starting databend-sqllogic tests"
-target/${BUILD_PROFILE}/databend-sqllogictests --handlers ${TEST_HANDLERS} ${RUN_DIR} --enable_sandbox --parallel 8 --skip_file tpcds_q64.test
+target/${BUILD_PROFILE}/databend-sqllogictests --handlers ${TEST_HANDLERS} ${RUN_DIR} --enable_sandbox --parallel 8 --skip_file tpcds_q64.test,tpcds_join_order.test
```
xudong963, line 24: why skip the test file?

tests/sqllogictests/suites/mode/cluster/explain_v2.test
```diff
@@ -342,50 +339,40 @@ Limit
...
-                    └── TableScan
-                        ├── table: default.system.numbers
-                        ├── output columns: [number (#0)]
-                        ├── read rows: 10
-                        ├── read bytes: 80
-                        ├── partitions total: 1
-                        ├── partitions scanned: 1
-                        ├── push downs: [filters: [], limit: NONE]
-                        └── estimated rows: 10.00
+    └── HashJoin
```
xudong963, line 342: Is the test as expected? `Merge` too early to take advantage of the benefits of distribution

src/query/sql/src/planner/optimizer/optimizer.rs
```diff
@@ -223,56 +221,97 @@ pub fn optimize_query(opt_ctx: OptimizerContext, mut s_expr: SExpr) -> Result<SE
...
+        opt_ctx.metadata.clone(),
+        dphyp_optimized,
+        enable_distributed_query,
+    )?;
+
+    // Cascades optimizer may fail due to timeout, fallback to heuristic optimizer in this case.
+    s_expr = match cascades.optimize(s_expr.clone()) {
+        Ok(mut s_expr) => {
+            s_expr =
+                RecursiveOptimizer::new(&[RuleID::EliminateEvalScalar], &opt_ctx).run(&s_expr)?;
```
Dousir9, line 245: It may be better to put `CommuteJoin` in `RecursiveOptimizer`, because it can ensure that the number of rows in build side is smaller, while `CascadesOptimizer` may not be able to do this.

scripts/ci/ci-run-sqllogic-tests-cluster.sh
leiysky: The explain result may be different between standalone mode and cluster mode. I think we don't need to prepare a whole tpcds test suite for it.

tests/sqllogictests/suites/mode/cluster/explain_v2.test
leiysky: It occurs when the table is very small and can not fit the cost model well. If we tune the data set to a larger scale, this will be fixed.

src/query/sql/src/planner/optimizer/optimizer.rs
leiysky: The post optimization of `CommuteJoin` will effect the result of `Exchange` placement. I'll try to put it right after dphyp.

tests/sqllogictests/suites/mode/cluster/explain_v2.test
leiysky: In this case, the overhead of merge data after join is bigger than merge the data before join. I'll modify the test to make it more reasonable.

src/query/sql/src/planner/optimizer/optimizer.rs
leiysky: Fixed.

xudong963 (approved):

tests/sqllogictests/suites/mode/cluster/broadcast_join.test
```diff
@@ -463,9 +461,24 @@ INSERT INTO t2(c1float) VALUES (0.9702655076980591)
 statement ok
 INSERT INTO t2(c1float, c2varchar) VALUES (0.5340723991394043, '02'), (0.4661566913127899, '1261837')
 
-statement ok
-SELECT t0.c0boolean, t1.c0boolean, t1.c1float FROM t0, t1 RIGHT JOIN t2 ON t1.c0boolean
-
+query TTI
+SELECT t0.c0boolean, t1.c0boolean, t1.c1float FROM t0, t1 RIGHT JOIN t2 ON t1.c0boolean ORDER BY t0.c0boolean, t1.c0boolean, t1.c1float
```
Dousir9, line 465: failed test.

leiysky: The original test case uses `statement ok` command, which won't check the output. I think it's better to make this a `query` command.

PR #14338, opened 2024-01-16
feat(planner): support predicates move around

Dousir9:

## Summary

Fixes #[Link the issue here]

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


src/query/service/tests/it/sql/planner/format/mod.rs
```diff
@@ -154,6 +154,7 @@ fn test_format() {
                 from_correlated_subquery: false,
                 need_hold_hash_table: false,
                 broadcast: false,
+                is_lateral: false,
```
leiysky, line 157: The lateral join should have been decorrelated before other optimizations.

What will we use the flag for?

Dousir9: Lateral join is different from subquery, it will generate the same column index, when the filter is pushed down to the join, it cannot be judged whether it belongs to the probe or build side, so we use this flag to mark whether a join is a lateral join, we will not pull up the filter below the lateral join because we cannot push it down.

leiysky: > it will generate the same column index

It sounds like a bug. 

Would you file an issue for this? I'll take a look at it later.

leiysky (approved):

src/query/service/tests/it/sql/planner/format/mod.rs
Dousir9: > > it will generate the same column index
> 
> It sounds like a bug.
> 
> Would you file an issue for this? I'll take a look at it later.

Yeah.

xudong963 (approved):

BohuTANG:
Conflicting files
tests/sqllogictests/suites/mode/standalone/explain/join_reorder/mark.test

Dousir9:
> cloud cloud
> 
> ## ClickBench Report
> * **hits**: https://benchmark.databend.rs/clickbench/pr/14338/7666557452/hits.html
> * **tpch**: https://benchmark.databend.rs/clickbench/pr/14338/7666557452/tpch.html

Previously, we manually added the join condition `c_nationkey = n_nationkey` to Q5 of TPC-H ci-benchmark sql, so the result of ci-benchmark is already optimized, now we can remove `c_nationkey = n_nationkey` of Q5 in ci-benchmark sql, because this PR can infer the join condition `c_nationkey = n_nationkey`.

PR #14355, opened 2024-01-17
refactor: meta-client should not block by RPC

drmingdrmer:

##### refactor: meta-client should not block by RPC

In this commit, `MetaGrpcClient` worker spawns a new task to handle the
RPC, instead of blocking receiving and handling all incoming RPC
request.

This change increases concurrency to meta-service RPC.

- Fix: #14356

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14355)

### DIFF
--- a/src/meta/client/src/grpc_client.rs
+++ b/src/meta/client/src/grpc_client.rs
@@ -22,6 +22,7 @@ use std::time::Duration;
 use std::time::Instant;
 
 use databend_common_arrow::arrow_format::flight::data::BasicAuth;
+use databend_common_base::base::tokio;
 use databend_common_base::base::tokio::select;
 use databend_common_base::base::tokio::sync::mpsc;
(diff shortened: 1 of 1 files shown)


src/meta/client/src/grpc_client.rs
```diff
@@ -450,116 +452,120 @@ impl MetaGrpcClient {
...
-            let request_id = req.request_id;
-            let resp_tx = req.resp_tx;
-            let req = req.req;
-            let req_name = req.name();
-            let req_str = format!("{:?}", req);
-
             // Deal with non-RPC request
             #[allow(clippy::single_match)]
-            match req {
+            match worker_request.req {
```
ariesdevil, line 478: why not use `if let` pattern here?

ariesdevil (approved):

PR #14361, opened 2024-01-17
chore: add a setting to enforce broadcast join

xudong963:

## Summary

Provide a switch to compare the performance of shuffle join and broadcast join.

Fixes #[Link the issue here]

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14361)

### DIFF
--- a/src/query/settings/src/settings_default.rs
+++ b/src/query/settings/src/settings_default.rs
@@ -288,6 +288,12 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: None,
                 }),
+                ("enforce_broadcast_join", DefaultSettingValue {
+                    value: UserSettingValue::UInt64(0),
+                    desc: "Enforce broadcast join.",
+                    mode: SettingMode::Both,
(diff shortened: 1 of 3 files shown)


leiysky (approved):

BohuTANG:
> Should modify this as well.
> 
> https://github.com/datafuselabs/databend/blob/9aaad617851695fc947fc74073408ba2f51f19f5/src/query/sql/src/planner/plans/join.rs#L569-L586
> 
> I think it's better to tune the cost model instead of comparing the performance with enbale/disable the switch. Maybe we can utilize AI to help us tune the cost factors. cc @BohuTANG

The settings only for us not for the user, we should choose a right value.
Now, most settings is OK but a few special settings need to enable or disable in some special use cases.

PR #14380, opened 2024-01-18
fix(query): use serde_json maybe cause old query can not access data that generate by new query

TCeason:

## Summary

1. add role/ownership.proto
2. replace serde_json to de/serialize_struct

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14380)


drmingdrmer (changes requested):
> Reviewed 4 of 27 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14380)* status: 4 of 27 files reviewed, 2 unresolved discussions (waiting on @TCeason)

___
*[`src/meta/app/src/principal/user_grant.rs` line 30 at r1](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoUBJct5nq_L5nQ4k88:-NoUBJct5nq_L5nQ4k89:b-5sczo7) ([raw file](https://github.com/datafuselabs/databend/blob/4808eef1a461e561fb7e475074ce24846db50db6/src/meta/app/src/principal/user_grant.rs#L30)):*
> ```XML
> pub enum OwnershipObject {
>     /// used on the fuse databases
>     Database(String, u64),
> ```

Is it safe to change the structure of `OwnershipObject`? How do you provide compatibility to these changes?
___
*[`src/query/management/src/role/role_mgr.rs` line 157 at r1](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoUCCTG032lt0B6lSqb:-NoUCCTG032lt0B6lSqc:bwhpiao) ([raw file](https://github.com/datafuselabs/databend/blob/4808eef1a461e561fb7e475074ce24846db50db6/src/query/management/src/role/role_mgr.rs#L157)):*
> ```XML
>                 seq_value.seq,
>                 deserialize_struct(&seq_value.data, ErrorCode::IllegalUserInfoFormat, || "")?,
>             )),
> ```

The old data is `serde_json` encoded, therefore when loading these data, it must try decoding with both protobuf and json. Otherwise the already existed data can not be read.



_Code quote:_
```XML
        match seq.match_seq(&seq_value) {
            Ok(_) => Ok(SeqV::new(
                seq_value.seq,
                deserialize_struct(&seq_value.data, ErrorCode::IllegalUserInfoFormat, || "")?,
            )),
```

BohuTANG (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14380)* status: 4 of 27 files reviewed, 2 unresolved discussions (waiting on @drmingdrmer and @TCeason)

___
*[`src/meta/app/src/principal/user_grant.rs` line 30 at r1](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoUBJct5nq_L5nQ4k88:-NoUEvq27pcKiF9fXzw6:b-s8fhkg) ([raw file](https://github.com/datafuselabs/databend/blob/4808eef1a461e561fb7e475074ce24846db50db6/src/meta/app/src/principal/user_grant.rs#L30)):*

Is it safe to change the structure of `OwnershipObject`? How do you provide compatibility to these changes?
</blockquote></details>

This is breaking the breaking change again.
If we modify a  strucut with `serde::Serialize, serde::Deserialize`, we must do the compatible with the old version.

TCeason:
> Is it safe to change the structure of OwnershipObject? How do you provide compatibility to these changes? 

in the newest commit, I revert it.

In this pr:

1. get method will first use serde_json , if failed , use deserialize_struct.

2. add or upsert use serialize_struct.

TCeason:
> > Is it safe to change the structure of OwnershipObject? How do you provide compatibility to these changes?
> 
> in the newest commit, I revert it.
> 
> In this pr:
> 
> 1. get method will first use serde_json , if failed , use deserialize_struct.
> 2. add or upsert use serialize_struct.

I will split this pr as two pr.

pr-1: get method will first use serde_json , if failed , use deserialize_struct.

pr-2: add or upsert use serialize_struct.

Prevents an issue where old query cannot read data written by new query during a rolling upgrade.

src/query/management/src/role/role_mgr.rs
```diff
@@ -158,7 +158,24 @@ impl RoleApi for RoleMgr {
             res.ok_or_else(|| ErrorCode::UnknownRole(format!("Role '{}' does not exist.", role)))?;
 
         match seq.match_seq(&seq_value) {
```
sundy-li, line 160: why not just modify `impl TryFrom<Vec<u8>> for RoleInfo ` ?

drmingdrmer (changes requested):
> Reviewed 3 of 27 files at r1, 3 of 4 files at r3, 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14380)* status: 9 of 27 files reviewed, 9 unresolved discussions (waiting on @BohuTANG, @sundy-li, and @TCeason)

___
*[`src/query/management/src/role/role_mgr.rs` line 164 at r4](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoVVSeQAjazVKOcfBXM:-NoVVSeRDfIRy7AeMg_x:b-2isd52) ([raw file](https://github.com/datafuselabs/databend/blob/d0351ae127848bd217ce9afa2b82fd0cdcc77a90/src/query/management/src/role/role_mgr.rs#L164)):*
> ```XML
>                 if <databend_common_meta_types::SeqV as IntoSeqV<RoleInfo>>::into_seqv(
>                     seq_value.clone(),
>                 )
> ```

The following call would just work, IMHO



_Suggestion:_
```XML
                if seq_value.into_seqv()
```
___
*[`src/query/management/src/role/role_mgr.rs` line 165 at r4](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoVVgct986v7ji9QgKI:-NoVVgcu2eJ6tpnD6S1_:b3g3exe) ([raw file](https://github.com/datafuselabs/databend/blob/d0351ae127848bd217ce9afa2b82fd0cdcc77a90/src/query/management/src/role/role_mgr.rs#L165)):*
> ```XML
>                     seq_value.clone(),
>                 )
>                 .is_err()
> ```

If there is an error, use `if let Err(e) = ` to capture it and log it.
___
*[`src/query/management/src/role/role_mgr.rs` line 172 at r4](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoVW-gjA9iZe5re52-u:-NoVW-gjA9iZe5re52-v:bgz0mgm) ([raw file](https://github.com/datafuselabs/databend/blob/d0351ae127848bd217ce9afa2b82fd0cdcc77a90/src/query/management/src/role/role_mgr.rs#L172)):*
> ```XML
>                             &seq_value.data,
>                             ErrorCode::IllegalUserInfoFormat,
>                             || "",
> ```

Add some context about the error occurred.
___
*[`src/query/management/src/role/role_mgr.rs` line 196 at r4](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoVW89H8p8UHQD7oCF5:-NoVW89I2P2-ALBA-rO2:b-rje2zq) ([raw file](https://github.com/datafuselabs/databend/blob/d0351ae127848bd217ce9afa2b82fd0cdcc77a90/src/query/management/src/role/role_mgr.rs#L196)):*
> ```XML
>         for (_key, val) in values {
>             let serd_json_res = serde_json::from_slice::<RoleInfo>(&val.data);
>             let u = if serd_json_res.is_err() {
> ```

Use `if let Err(e) = ` or `match `, and log the error.
___
*[`src/query/management/src/role/role_mgr.rs` line 197 at r4](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoVWBUKE73WTESiCdsi:-NoVWBUKE73WTESiCdsj:b-qyn2qp) ([raw file](https://github.com/datafuselabs/databend/blob/d0351ae127848bd217ce9afa2b82fd0cdcc77a90/src/query/management/src/role/role_mgr.rs#L197)):*
> ```XML
>             let serd_json_res = serde_json::from_slice::<RoleInfo>(&val.data);
>             let u = if serd_json_res.is_err() {
>                 deserialize_struct(&val.data, ErrorCode::IllegalUserInfoFormat, || "")?
> ```

Please include context information by adding the initial bytes of the string that you're attempting to decode.
___
*[`src/meta/app/src/principal/ownership_info.rs` line 52 at r2](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoVWdqx83tOSV6Exa0q:-NoVWdqx83tOSV6Exa0r:b2h9rc7) ([raw file](https://github.com/datafuselabs/databend/blob/b62f1e8d49d0fd03a88252e337ce2b706881ec90/src/meta/app/src/principal/ownership_info.rs#L52)):*
> ```XML
>             }),
>         }
>     }
> ```

Using this `TryFrom` is not very obvious about what is done when converting `OwnershipInfo` to bytes.

I would recommend removing it and use `serde_json` or `serialize_struct` dierectly.



_Code quote:_
```XML
impl TryFrom<Vec<u8>> for OwnershipInfo {
    type Error = OwnershipInfoSerdeError;

    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
        match serde_json::from_slice(&value) {
            Ok(role_info) => Ok(role_info),
            Err(serialize_error) => Err(OwnershipInfoSerdeError {
                message: "Cannot deserialize GrantOwnershipInfo from bytes".to_string(),
                source: AnyError::new(&serialize_error),
            }),
        }
    }
```
___
*[`src/meta/app/src/principal/role_info.rs` line 53 at r2](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoVWxE65wWLdLUdhDK2:-NoVWxE65wWLdLUdhDK3:b-hbcorw) ([raw file](https://github.com/datafuselabs/databend/blob/b62f1e8d49d0fd03a88252e337ce2b706881ec90/src/meta/app/src/principal/role_info.rs#L53)):*
> ```XML
> }
> 
> impl TryFrom<Vec<u8>> for RoleInfo {
> ```

The same.

drmingdrmer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14380)* status: 9 of 27 files reviewed, 8 unresolved discussions (waiting on @BohuTANG, @sundy-li, and @TCeason)

___
*[`src/query/management/src/role/role_mgr.rs` line 162 at r5](https://reviewable.io/reviews/datafuselabs/databend/14380#-NoWHX_L4KAp1qOtpC_W:-NoWHX_L4KAp1qOtpC_X:bzae6ih) ([raw file](https://github.com/datafuselabs/databend/blob/46f74450097c39940fe11d46ff7a8b66447bd597/src/query/management/src/role/role_mgr.rs#L162)):*
> ```XML
>             Ok(_) => {
>                 let data = serde_json::from_slice::<RoleInfo>(&seq_value.data).or_else(|_| {
>                     deserialize_struct(&seq_value.data, ErrorCode::IllegalUserInfoFormat, || "")
> ```

Add some context information to `|| ""` to save some debugging time.

src/query/management/src/role/role_mgr.rs
TCeason: because this: 

```
Using this TryFrom is not very obvious about what is done when converting OwnershipInfo to bytes.

I would recommend removing it and use serde_json or serialize_struct dierectly.
```

I mark this resolved.

drmingdrmer (approved):
LGTM

TCeason:
Please review this pr. @sundy-li @BohuTANG

BohuTANG:
LGTM.
Wait the test_compat CI passed.

src/query/management/src/role/role_mgr.rs
```diff
@@ -158,7 +157,13 @@ impl RoleApi for RoleMgr {
...
-            Ok(_) => Ok(seq_value.into_seqv()?),
+            Ok(_) => {
+                let data = serde_json::from_slice::<RoleInfo>(&seq_value.data).or_else(|err| {
+                    log::debug!("serde json err when get role. cause : {}", err);
+                    deserialize_struct(&seq_value.data, ErrorCode::IllegalUserInfoFormat, || "")
```
BohuTANG, line 163: Please add a comments for this, it's not by design, it's for compatible with the old version.

src/query/management/src/role/role_mgr.rs
```diff
@@ -158,7 +157,13 @@ impl RoleApi for RoleMgr {
             res.ok_or_else(|| ErrorCode::UnknownRole(format!("Role '{}' does not exist.", role)))?;
 
         match seq.match_seq(&seq_value) {
-            Ok(_) => Ok(seq_value.into_seqv()?),
+            Ok(_) => {
+                let data = serde_json::from_slice::<RoleInfo>(&seq_value.data).or_else(|err| {
+                    log::debug!("serde json err when get role. cause : {}", err);
```
BohuTANG, line 162: I would prefer to `deserialize_struct` first rather than `serde_json::from_slice` here.
Reson:
If `cdeserialize_struct` first, we can know a version is ok. `serde_json::from_slice` only affects some version, but the future will be `cdeserialize_struct`.
cc @drmingdrmer

drmingdrmer: I'm totally fine with running `deserialize_struct` right before `serde_json`. :)

BohuTANG: @TCeason what do you think?

TCeason: That's ok. But better modify it in next pr with serialize pb?

In this pr the role info also serialize as json. So no need to deserialize first.

BohuTANG: Ok, we need to upgrade the json format to the new proto when reading(fix on reading), or the json format will not changing for ever.

TCeason: > Ok, we need to upgrade the json format to the new proto when reading(fix on reading), or the json format will not changing for ever.

I think No need upgrade?

In if user grant or revoke the wold role, it will call upsert role.

In this method, will call get role， and then serialize the new privileges into role.

drmingdrmer:
> > Ok, we need to upgrade the json format to the new proto when reading(fix on reading), or the json format will not changing for ever.
> 
> I think No need upgrade?
> 
> In if user grant or revoke the wold role, it will call upsert role.
> 
> In this method, will call get role， and then serialize the new privileges into role.

@TCeason @BohuTANG 
It occurred to me that we need to add a fuse-compat to ensure the latest query can read role info written by an earlier query:
https://github.com/datafuselabs/databend/blob/486d804640a930c3b369b9be0a9662e7093b2436/.github/actions/fuse_compat/action.yml#L31-L30

And ensure that there is role info write and read in the one of the logictest dir:
https://github.com/datafuselabs/databend/tree/f06d9b311ec046f6702f42d5a9947f9b6e173b61/tests/fuse-compat

PR #14402, opened 2024-01-22
chore: enable new setting `data_retention_time_in_days`

dantengsky:

## Summary

- enable new setting `data_retention_time_in_days` which introduced in PR #14388
- remove the old setting `retention_perid` (this is a non-compatible change)

- syntax of `vacuum` statement has been adjusted  (this is a non-compatible change)

  The [RETAIN n HOURS] clause has been removed. 
  
  VACUUM DROP TABLE 
    [FROM <database_name>] 
    ~~[RETAIN <n> HOURS]~~
    [DRY RUN]
    [LIMIT <file_count>]
    
    
  VACUUM TABLE <table_name> ~~[RETAIN n HOURS]~~ [DRY RUN]

  
  
  please specify the retention interval by using ` set data_retention_time_in_days = n` instead.

-------

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14402)


sundy-li:
IMHO, I still think the old syntax `RETAIN n DAYS` is much better than the current one.

1. It just needs one SQL to execute the `VACUUM` command
2. `set data_retention_time_in_days = n` is hard to debug from logs, we still need to fetch the settings to know the value.

I believe we have overused settings configurations in many places. In some cases, we should maintain the unique characteristics and properties of objects, rather than introducing attributes into settings.

BohuTANG:
> IMHO, I still think the old syntax `w n DAYS` is much better than the current one.
> 
> 1. It just needs one SQL to execute the `VACUUM` command.
> 2. `set data_retention_time_in_days = n` and execute `VACUUM` is hard to debug from logs, we still need to fetch the settings to know the value.
> 3. If we forgot to set `data_retention_time_in_days`, it will fetch from global value (which maybe set by other person to a small value).  It's pretty dangerous!
> 
> I believe we have overused settings configurations in many places. In some cases, we should maintain the unique characteristics and properties of objects, rather than introducing attributes into settings.

We do not consider do this, reason:
1. `data_retention_time_in_days` only be set and controled by the user not Databend
2.  If we support `RETAIN 0 days` for vacuum, we can not guarantee the fail-safe.

sundy-li:
> If we support RETAIN N days for vacuum, we can not guarantee the fail-safe indata_retention_time_in_days  which we promised to the user.

We can still check this `N`  with  `data_retention_time_in_days ` config.

BohuTANG:
> > If we support RETAIN N days for vacuum, we can not guarantee the fail-safe indata_retention_time_in_days  which we promised to the user.
> 
> We can still check this `N` with `data_retention_time_in_days ` config.

It's hard to do it, if we need in the future we can add it back but not now.

lichuang (approved):

PR #14420, opened 2024-01-22
feat: treat path in uri as prefix.

youngsofun:

## Summary

With this PR, there are two ways to interpret the path component in URLs:

As a root path, which is an essential part of the definition for structures like Stage, ExternalTable, IcebergTable, etc. In such cases, the path must end with a '/'.
As a path prefix, employed to select all files starting with it.

This PR introduces a **breaking change** when the path is used as a prefix without ending with '/'. 
Previously, the selection was limited to the file exactly specified by the path.
However, the impact of this incompatibility on users is likely to be acceptable. This is because: in most such usages, the path ends with a file extension (like '.csv'), making it unlikely to be used as a prefix for another file.

After this PR, to ensure the precise selection of a single file, users should adopt the option `files = ('<path>', )` when copy into table.

update
----
- fix: the root and path components extracted from the URL are now normalized before being used.
- a bug in opendal is just fix (list return prefix itself). it is properly handled in this PR (filter out it) @Xuanwo 
  - NOTE form @xuanwo: Fixed in https://github.com/apache/opendal/pull/4067, will be released in next version.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14420)


sundy-li:
what if we just want to read the exact file without prefix?  Is there any other way?

youngsofun:
> what if we just want to read the exact file without prefix? Is there any other way?

`pattern= ''`

BohuTANG (approved):
LGTM.

This will align the behavior with that of Snowflake.

Xuanwo (approved):

PR #14444, opened 2024-01-24
feat(query): copy into allow `disable_variant_check` for parquet file

b41sh:

## Summary

- Add setting `disable_variant_check` for `parse_json` function to allow invalid JSON
- copy into parquet file format allow `disable_variant_check` options to work

Fixes #[Link the issue here]

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


src/query/settings/src/settings_default.rs
```diff
@@ -624,6 +624,12 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: Some(SettingRange::Numeric(0..=1)),
                 }),
+                ("disable_variant_check", DefaultSettingValue {
+                    value: UserSettingValue::UInt64(0),
+                    desc: "Disable variant check to allow insert invalid JSON values",
```
BohuTANG, line 629: If the invalid JSON insert, what will happen to query them?

b41sh: jsonb's functions will check whether the JSON is valid or not, if it is not a valid JSON, the function will return the default value, so do not worry about the runtime will report an error.

wubx:
@soyeric128  update docs

PR #14445, opened 2024-01-24
chore: make cost factor configurable

leiysky:

## Summary

Introduce some settings to configure the cost factors.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14445)

### DIFF
--- a/src/query/settings/src/settings_default.rs
+++ b/src/query/settings/src/settings_default.rs
@@ -25,6 +25,13 @@ use once_cell::sync::OnceCell;
 
 static DEFAULT_SETTINGS: OnceCell<Arc<DefaultSettings>> = OnceCell::new();
 
+// Default value of cost factor settings
+#[allow(dead_code)]
+static COST_FACTOR_COMPUTE_PER_ROW: u64 = 1;
+static COST_FACTOR_HASH_TABLE_PER_ROW: u64 = 10;
+static COST_FACTOR_AGGREGATE_PER_ROW: u64 = 5;
+static COST_FACTOR_NETWORK_PER_ROW: u64 = 50;
+
 // Settings for readability and writability of tags.
 // we will not be able to safely get its value when set to only write.
 // we will not be able to safely set its value when set to only read.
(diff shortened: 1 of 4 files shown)


andylokandy (approved):

src/query/settings/src/settings_default.rs
```diff
@@ -624,6 +631,24 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: Some(SettingRange::Numeric(0..=1)),
                 }),
+                ("cost_factor_hash_table_per_row", DefaultSettingValue {
+                    value: UserSettingValue::UInt64(COST_FACTOR_HASH_TABLE_PER_ROW),
+                    desc: "Cost factor of building hash table for a data row",
+                    mode: SettingMode::Both,
+                    range: Some(SettingRange::Numeric(0..=u64::MAX)),
+                }),
+                ("cost_factor_aggregate_per_row", DefaultSettingValue {
```
BohuTANG, line 640: These new settings is for developer or the user? The `desc` is not easy to understand how to use them.

leiysky: Yes, it’s for developers. Do we have solution not to expose these settings to users?

BohuTANG: Yes, we don't recommand export the `VALUE` setting, if they set the global value it hard to adjust it in the future. We should atomic setting them by us not setting.

leiysky: > Settings is good for: disable or enable a feature by us.

I need a way to modify the configuration dynamicly to tune the cost model. Do you have any suggestion for this?

BohuTANG: Ok, I see. This PR works.

BohuTANG: BTW, we can not tell the user to set these settings, only used by us.

PR #14460, opened 2024-01-24
feat: pushdown limit to single window

ariesdevil:

## Summary

when single window with frame type `rows` in select list with limit, we pushdown limit to window order by sort.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14460)


tests/sqllogictests/suites/mode/standalone/explain/window.test
```diff
@@ -285,5 +285,63 @@ Filter
...
+create table t(a int)
+
+statement ok
+insert into t values(1), (2), (3), (3)
+
+# range frame
+query T
+explain pipeline select a, sum(a) over (partition by a order by a desc) from t limit 3
+----
+CompoundBlockOperator(Project) × 1 processor
```
sundy-li, line 302: I don't know if it's pushed or not from this pipeline.

src/query/sql/src/executor/physical_plans/physical_window.rs
```diff
@@ -55,6 +55,7 @@ pub struct Window {
     pub partition_by: Vec<IndexType>,
     pub order_by: Vec<SortDesc>,
     pub window_frame: WindowFuncFrame,
+    pub limit: Option<usize>,
```
Dousir9, line 58: We can add `limit:xxx` to `window_to_format_tree` in `src/query/sql/src/executor/format.rs` to display limit in explain.

tests/sqllogictests/suites/mode/standalone/explain/window.test
ariesdevil: Yep, for range frame we cannot push limit to window sorts, the following rows frame will change `TransformSortMerge` => `TransformSortMergeLimit`

ariesdevil: I'll add the ranking functions cause they are frame insensitive.

xudong963: why not use explain select but use explain pipeline?

src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_limit_window.rs
```diff
@@ -0,0 +1,134 @@
...
+        Ok(())
+    }
+
+    fn patterns(&self) -> &Vec<SExpr> {
+        &self.patterns
+    }
+}
+
+fn should_apply(child: &SExpr, window: &LogicalWindow) -> Result<bool> {
+    let child_window_exists = child_has_window(child)?;
```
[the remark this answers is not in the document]
ariesdevil: For simply, if we have multiple windows, they may have different partition by and order by，then we should check they are same or they both can apply limit

tests/sqllogictests/suites/mode/standalone/explain/window.test
ariesdevil: the explain are same，we just reduce the sort with limit for window input，but still need sort

sundy-li (approved):

BohuTANG:
Conflicting files:
src/query/service/src/pipelines/builders/builder_window.rs
src/query/sql/src/executor/format.rs

PR #14472, opened 2024-01-25
fix(ddl): fix create table bitmap and variant with default value

### LINKED ISSUE #14391: bug: create table bitmap data type with default value cause panic
### Search before asking

### Version

v1.2.305-nightly

### What's Wrong?

create table bitmap data type with default value cause panic
```
panicked at src/query/sql/src/planner/expression_parser.rs:419:79:
removal index (is 0) should be < len (is 0)
   0: databend_common_tracing::panic_hook::log_panic
             at ./src/common/tracing/src/panic_hook.rs:35:21
```

### How to Reproduce?

```sql
mysql> CREATE TABLE IF NOT EXISTS t1 (
    ->     a STRING NOT NULL DEFAULT '',
    ->     b BITMAP NOT NULL DEFAULT '0'
    -> );
ERROR 1105 (HY000): TokioError. Code: 1068, Text = Cannot join handle from context's runtime, cause: task 183 panicked.
```

b41sh:

## Summary

* modify `fmt` function of  `ScalarRef` to display bitmap and variant value with quotes
* fix create table bitmap and variant type with default value panic
* return error if default value can't cast to column data type instead of use `NULL` as default value
* using `parse_json` to cast string to variant
* `to_variant(null)` and try_to_variant(null) return `NULL` instead of JSON null

- Fixes #14391

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14472)


src/query/sql/src/planner/binder/scalar_common.rs
```diff
@@ -227,17 +228,36 @@ pub fn wrap_cast_scalar(
...
-    {
-        // parse string to JSON value
-        let func = ScalarExpr::FunctionCall(FunctionCall {
-            span: None,
-            func_name: "parse_json".to_string(),
-            params: vec![],
-            arguments: vec![scalar.clone()],
-        });
-        wrap_cast(&func, target_type)
+    let target_scalar = if target_type.remove_nullable() == DataType::Variant {
```
andylokandy, line 231: Why need special case for Variant? `wrap_cast` should have supported varaiant

b41sh: `wrap_cast` will cast the string to a JSON string, not parse string to JSON value.

andylokandy: If so, we should change the behavior of `to_variant(string)` instead of adding `if` here

b41sh: Changing `to_variant(string)` is indeed simpler. Currently, this behavior is meant to be compatible with Snowflake, I'm not sure if it should be changed, maybe we need to some discuss.

andylokandy: `wrap_cast_scalar` should behave the same as `wrap_cast`. Since the `if` here, `CAST(... AS VARAIANT)` has been changed to `parse_json`, which means the compatibility has been broken with inconsistency.

b41sh: After discussing with @sundy-li , it's better to keep the current behavior for better compatibility. The `wrap_cast_scalar` function is only used for cast values in insert, copy and update. Casts to variant types use the `cast_scalar_to_variant` function, which doesn't convert to `parse_json`. So cast and insert values will behave differently. @andylokandy

andylokandy (approved):

PR #14501, opened 2024-01-29
fix(query): privilege type forward compat

TCeason:

## Summary

1. add ci test foreward compat test(now only test rbac)
2. UserPrivilegeType and ShareGrantObjectPrivilege deserialize use from_bits_truncate

Befero we use from_bits, in this pr, use from_bits_truncate

from_bits will cause breaking change. When add new bit, old query can not deserialize the new bit. And then return err.

https://docs.rs/enumflags2/0.7.7/enumflags2/struct.BitFlags.html#method.from_bits

https://docs.rs/enumflags2/0.7.7/enumflags2/struct.BitFlags.html#method.from_bits_truncate

I think we should ignore the new bit. So now call from_bits_truncate to deserialize bitflags.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14501)

### DIFF
--- a/.github/actions/fuse_compat/action.yml
+++ b/.github/actions/fuse_compat/action.yml
@@ -30,6 +30,8 @@ runs:
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.39 base
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.46 base
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.2.306 rbac
+          bash ./tests/fuse-compat/test-fuse-forward-compat.sh 1.2.306 rbac
+          bash ./tests/fuse-compat/test-fuse-forward-compat.sh 1.2.307 rbac
     - name: Upload failure
       if: failure()
       uses: ./.github/actions/artifact_failure
--- a/.github/actions/fuse_compat/action.yml
+++ b/.github/actions/fuse_compat/action.yml
@@ -30,8 +30,8 @@ runs:
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.39 base
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.46 base
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.2.306 rbac
-          bash ./tests/fuse-compat/test-fuse-forward-compat.sh 1.2.306 rbac
           bash ./tests/fuse-compat/test-fuse-forward-compat.sh 1.2.307 rbac
+          bash ./tests/fuse-compat/test-fuse-forward-compat.sh 1.2.318 rbac
     - name: Upload failure
       if: failure()
       uses: ./.github/actions/artifact_failure
--- a/tests/fuse-compat/util.sh
+++ b/tests/fuse-compat/util.sh
@@ -191,6 +191,89 @@ run_test() {
     $sqllogictests --handlers mysql --suites "$logictest_path" --run_file fuse_compat_read
 }
 
+# Test fuse-data forward compatibility between an old version query and the current
+# version query.
+run_forward_test() {
+    echo " === pip list"
+    python3 -m pip list
+
+    local query_old_ver="$1"
+    local old_config_path="$2"
+    local logictest_path="tests/fuse-compat/compat-logictest/$3"
+
+    echo " === Test with query-$query_old_ver and current query"
+
+    local query_old="./bins/$query_old_ver/bin/databend-query"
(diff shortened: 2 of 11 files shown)


TCeason:
@everpcpc Please help to review the .github yaml and test-foreward-compat part in this pr.

everpcpc (approved):

TCeason:
cc @lichuang @drmingdrmer  Are there any other concerns about considering using bits_from? Through these tests added so far, I think it's safe to guarantee that the current pr will not cause a break. Please review it.

cc @BohuTANG Maybe can add a ci-cloud tag, and try to fix your warehouse?

drmingdrmer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14501)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @lichuang, @TCeason, and @Xuanwo)

___
*[`tests/fuse-forward-compat/util.sh` line 1 at r1](https://reviewable.io/reviews/datafuselabs/databend/14501#-NpIn0ayEfOQiFuOsNrb:-NpIn0az7zxYOsHNhji_:bftgea3) ([raw file](https://github.com/datafuselabs/databend/blob/1d8861a1d895b038021f09ac7688dae14d07202e/tests/fuse-forward-compat/util.sh#L1)):*
> ```Shell
> #!/bin/bash
> ```

Do not copy, reuse `util.sh`. It's a burden maintaining two copies of the same function set.

drmingdrmer:
@TCeason 
IMHO, You can just add the forward compat test in the fuse-compat dir.

TCeason:
> @TCeason IMHO, You can just add the forward compat test in the fuse-compat dir.

I got you. I'm modifing it. But now this can prove that this pr not generate new break change?

TCeason:
> @TCeason IMHO, You can just add the forward compat test in the fuse-compat dir.

Done

lichuang (approved):

drmingdrmer:
@TCeason 
Why do not you just re-use `run_test` in the `util.sh` but create another `run_forward_test` instead?
These two function does the same thing: write to version A and ready with version B. 
Just update variable names `run_test` uses and reuse it for forward compatibility test.

tests/fuse-compat/util.sh
```diff
@@ -150,47 +151,70 @@ run_test() {
...
-    echo " === Run test: fuse_compat_write with old query"
+        echo " === Run test: fuse_compat_write with old query"
+    fi
 
     # download_logictest $query_old_ver old_logictest/$query_old_ver
-    # run_logictest old_logictest/$query_old_ver fuse_compat_write
+    # if backward
+    # run_logictest new_logictest/$query_old_ver fuse_compat_write
+    # if forward
+    # run_logictest new_logictest/$query_new_ver fuse_compat_write
```
drmingdrmer, line 176: What does these comment mean?

drmingdrmer (approved):

tests/fuse-compat/util.sh
TCeason: backward test means old query write;

forward test means new query write.

.github/actions/fuse_compat/action.yml
```diff
@@ -30,6 +30,8 @@ runs:
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.39 base
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.1.46 base
           bash ./tests/fuse-compat/test-fuse-compat.sh 1.2.306 rbac
+          bash ./tests/fuse-compat/test-fuse-forward-compat.sh 1.2.306 rbac
```
TCeason, line 33:
```
 === Start ./bins/1.2.306/bin/databend-query databend-query...
databend config path: old_config/scripts/ci/deploy/config/databend-query-node-1.toml
... connecting to :3307
... connecting to :3307
... connecting to :3307
... connecting to :3307
Databend Query start failure, cause: InvalidReply. Code: 2005, Text = Cannot deserialize RoleInfo from bytes cause: serde_json::error::Error: expected value at line 2 column 2.
```

Because current query version will serialize new role info as pb.

306 query version only support deserialize role info as json. So current query version it not forward compat with 1.2.306. That is known. We need prove it's forward compat with 307.

BohuTANG:
Test on Cloud Prod, user auth has no issue.

PR #14510, opened 2024-01-29
fix: table maybe locked when recluster

zhyass:

## Summary

1. add log for recluster

2. Resolved an issue where the table may become locked during recluster or compact operations. 
This was caused by a timing issue where the lock requested during SQL execution and triggering of compact wasn't released promptly before checking for the lock.    

3. set the min choosed block num for recluster to 2
    Specify a rather small value, so that setting `recluster_block_size` might be tuned to lower value.

4. fix recluster maybe panic by div zero.
    This may arise due to the presence of unusually large blocks, where the size exceeds the recluster block size limit, potentially leading to the selection of an empty task.

5. recluster select segment limit.
    The minimum number of segments selected for each recluster operation is set at four times the maximum number of threads, ensuring efficient utilization of system resources.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14510)


BohuTANG:
Please change the PR title, there are some important parts than **log**, make the blame more meaningful.

dantengsky (approved):

PR #13548, opened 2023-11-02
feat(query): agg-hashtable-p2

sundy-li:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

Summary about this PR

Introduce agg hashtable in the query pipeline and support partition ht.

We can enable this via : `set enable_experimental_aggregate_hashtable = 1;`

Note this pr only works in singleton deployment currently.

Pros:
1. better performance in high cardinality group aggregation, up to 2x
2. better and cleaner codes than the old one

Cons:
1. It may consume more memory usage than the old one cause it uses using two-part struct layout
2. A little worse performance in normal group aggregation,  ~ 0.9x

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13548)

PR #14524, opened 2024-01-30
feat(query): agg-hashtable-singleton

Freejww:

## Summary

Introduce agg hashtable in the query pipeline and support partition ht. 

In this PR, after running ClickBench, we observed improved performance for most queries, the remaining performance is similar to main. We also run ClickBench in DuckDB in the same environment, and the performance was similar.

We can enable this via : set enable_experimental_aggregate_hashtable = 1;

Note this pr only works in singleton deployment currently.

Pros:
better performance in high cardinality group aggregation, up to 2x
better performance in group by string column
better and cleaner codes than the old one

Cons:
It may consume more memory usage than the old one cause it uses using two-part struct layout

continue #13548 

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14524)


BohuTANG:
Better to give a benchmark results rather than a summary :)

sundy-li (approved):

PR #14552, opened 2024-02-01
fix: runtime filters add to wrong table

xudong963:

## Summary

Make each probe key attach to its corresponding tables. The table will be used as the target table to add runtime filters.

For example:

Right side is the build side, t1 will have runtime filter due to the bottom of join and t2 will have runtime filter due to the top of join.

```
       join(t2.c = t3.c)
      /   \
     /    t3
join(t1.a = t2.b)
   /  \
  t1   t2
```

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test  - covered by current tests

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement


leiysky (approved):

Dousir9 (approved):

src/query/sql/src/executor/physical_plans/physical_hash_join.rs
```diff
@@ -69,8 +69,8 @@ pub struct HashJoin {
     // Only used for explain
     pub stat_info: Option<PlanStatsInfo>,
 
-    // probe keys for runtime filter
-    pub probe_keys_rt: Vec<Option<RemoteExpr<String>>>,
+    // probe keys for runtime filter, and record the index of table that used in probe keys.
+    pub probe_keys_rt: Vec<(Option<RemoteExpr<String>>, Vec<IndexType>)>,
```
Dousir9, line 73: For base table, we may not need to use `Vec`.

Issue #14543, opened 2024-02-01
Tracking: Geometry types


ariesdevil:
**Summary**

- [x] initial support geometry type. #14470 
- [x] support create table with geometry type. #14615 
- [ ] add more functions to geometry type
- [ ] add helpers for serde and show geometry types
- [ ] after all, write many tests for this features
- [ ] add docs
- [x] support geography type [#16286]

kkk25641463:
I found that ST_GEOMFROMGEOHASH() returned value is geojson type ([doc](https://docs.snowflake.com/en/sql-reference/functions/st_geomfromgeohash) ). Should we implement subtypes with geometry type like number type or decimal type? And [output](https://github.com/datafuselabs/databend/blob/f5478ff8f0e00acc1cb340e31301a77b2dedf9b7/src/query/expression/src/utils/display.rs#L256) can be choose with subtypes like ewkb、ewkt、geojson.

PR #14615, opened 2024-02-05
feat(geo): support geometry type for create table

ariesdevil:

## Summary

Support geometry type for create table and add two new geometry functions.
If use `bendsql`, need this patch https://github.com/datafuselabs/bendsql/pull/352

part of #14543 

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14615)


ariesdevil:
we may need install deps for [build bounded proj](https://github.com/georust/proj?tab=readme-ov-file#requirements) in our ci container. cc @everpcpc

sundy-li:
I think it should be supported only if it's stable to avoid incompat problem.

ariesdevil:
parse `POINTZ`, `POINTM` and `POINTZM` for `st_geometryfromwkt` function will be done after upstream PR merged: https://github.com/georust/wkt/pull/113

As @lnicola [kindly point out](https://github.com/georust/wkt/pull/113#issuecomment-1931808990) the `Coord` in rust `geo-types` [crate](https://github.com/georust/geo/blob/53ec97327d4dd9cd124be74b8a53a8558bd0836c/geo-types/src/geometry/coord.rs#L28) just support `x` and `y`.

`geo-types` is a foundational library, almost all libraries under [georust](https://github.com/georust) depend on it. So if we want to support `Point(x,y,z,m)` it may require a lot of work to be done.

lnicola:
>  the Coord in rust geo-types crate just support x and y

See also https://github.com/georust/geo/pull/797

src/query/sql/src/planner/binder/ddl/table.rs
```diff
@@ -468,6 +469,19 @@ impl Binder {
...
 
+        // todo(geometry): remove this when geometry stable.
+        if let Some(CreateTableSource::Columns(cols)) = &source {
+            if cols
+                .iter()
+                .any(|col| matches!(col.data_type, TypeName::Geometry))
+                && !self.ctx.get_settings().get_enable_geo_create_table()?
+            {
+                return Err(ErrorCode::GeometryError(
+                    "Create table using the geometry type is an experimental feature. We do not guarantee its compatibility.",
```
b41sh, line 480: error message should add `set enable_geo_create_table=1` to notice user modify setting.

src/query/functions/src/scalars/geometry.rs
```diff
@@ -74,4 +134,121 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+fn make_crs(srid: i32) -> String {
+    format!("EPSG:{}", srid)
+}
+
+fn read_ewkb_srid<R: Read>(raw: &mut R) -> Result<Option<i32>> {
+    let byte_order = raw.ioread::<u8>()?;
+    let is_little_endian = byte_order != 0;
+    let endian = Endian::from(is_little_endian);
+    let type_id = raw.ioread_with::<u32>(endian)?;
+    let srid = if type_id & 0x2000_0000 == 0x2000_0000 {
```
b41sh, line 247: better to define `0x2000_000` as a const value

src/query/functions/src/scalars/geometry.rs
```diff
@@ -74,4 +134,121 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                    if !validity.get_bit(builder.len()) {
+                        builder.commit_row();
+                        return;
+                    }
+                }
+
+                #[allow(unused_assignments)]
+                let mut from_srid = 0;
+
+                // All types supported by georust have not implemented srid().
```
b41sh, line 153: Dose `georust` a typo?

scripts/setup/dev_setup.sh
```diff
@@ -327,6 +327,62 @@ function install_mysql_client {
 	esac
 }
 
+function install_sqlite3 {
```
b41sh, line 330: Why do we need install sqlite3 ?

src/query/functions/src/scalars/geometry.rs
```diff
@@ -53,6 +75,44 @@ pub fn register(registry: &mut FunctionRegistry) {
         })
     );
 
+    registry.register_passthrough_nullable_1_arg::<StringType, GeometryType, _, _>(
+        "st_geometryfromwkt",
+        |_, _| FunctionDomain::Full,
```
b41sh, line 80: `parse_to_ewkb` may failed, the function domain should be `FunctionDomain::MayThrow`

ariesdevil:
@b41sh PTAL again when you have time. Happy Spring Festival：）

b41sh (approved):
LGTM, please fix the ci tests.

src/common/io/src/geometry.rs
```diff
@@ -19,22 +19,25 @@ use geozero::CoordDimensions;
...
     let parts: Vec<&str> = input_wkt.split(';').collect();
 
-    if input_wkt.starts_with("SRID=") && parts.len() == 2 {
-        srid = Some(parts[0].replace("SRID=", "").parse()?);
-    }
+    let parsed_srid: Option<i32> = srid.or_else(|| {
+        if input_wkt.starts_with("SRID=") && parts.len() == 2 {
+            parts[0].replace("SRID=", "").parse().ok()
+        } else {
+            None
```
kkk25641463, line 32: [links](https://docs.snowflake.com/en/sql-reference/functions/st_geometryfromwkt)
Snowflake default processing for WKT input. if the srid argument is not specified, the resulting GEOMETRY object has the SRID set to 0.

ariesdevil: 0 is not a valid SRID

kkk25641463: PostGIS as a factual standard in gis system. We can refer to its [settings](https://postgis.net/docs/manual-3.5/using_postgis_dbmanagement.html#spatial_ref_sys). A local SRS is a Cartesian coordinate system which is not referenced to the earth's surface. In PostGIS this is specified by a SRID value of 0. And all geo rust library and posgis  depend on [libgeos](https://libgeos.org/doxygen/geos__c_8h.html). GEOSGetSRID function defines srid 0 is unknown / not set.

ariesdevil:
image

I have finished the static linking, but the tests still failed, and builds for aarch64 and python binding are not successful. @everpcpc need your help.

everpcpc:
`liblzma` and `libstdc++` also needs to be static linked.

PR #14624, opened 2024-02-06
chore: optimize q17 join order

xudong963:

## Summary

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


BohuTANG:
ci-benchmark still has some problems, wait @everpcpc to fix.

src/query/sql/src/planner/optimizer/decorrelate/decorrelate.rs
```diff
@@ -267,11 +268,20 @@ impl SubqueryRewriter {
                     &mut right_conditions,
                     &mut left_conditions,
                 )?;
+
+                let mut join_type = JoinType::LeftSingle;
+                if subquery.contain_agg.unwrap() {
```
Dousir9, line 273: Is it safe to unwrap ?

src/query/sql/src/planner/optimizer/decorrelate/decorrelate.rs
```diff
@@ -486,3 +498,18 @@ impl SubqueryRewriter {
         }))
     }
 }
+
+// Check if contain group by in subquery
+fn contain_group_by(subquery: &SExpr) -> bool {
```
Dousir9, line 503: it would be better to rename `subquery` to `s_expr`.

src/query/sql/src/planner/optimizer/hyper_dp/dphyp.rs
```diff
@@ -178,6 +178,12 @@ impl DPhpy {
                 }
                 // Add join conditions
                 for condition_pair in op.left_conditions.iter().zip(op.right_conditions.iter()) {
+                    let left_used_tables = condition_pair.0.used_tables()?;
+                    let right_used_tables = condition_pair.1.used_tables()?;
+                    if left_used_tables.is_empty() || right_used_tables.is_empty() {
```
Dousir9, line 183: When will this happen ?

src/query/sql/src/planner/optimizer/decorrelate/decorrelate.rs
xudong963: Yes, scalar subquery must have contain_agg, it was initialized by type checker.

src/query/sql/src/planner/optimizer/decorrelate/decorrelate.rs
xudong963: It's only used by to check if contain group by in subquery

src/query/sql/src/planner/optimizer/hyper_dp/dphyp.rs
xudong963: For example, the join key is from the derived column generated during decorrelation. 

```
explain select t.number from numbers(2) as t where number = (select * from numbers(1) where number = 0)
----
HashJoin
├── output columns: [t.number (#0)]
├── join type: INNER
├── build keys: [_if_scalar_subquery (#4)]
├── probe keys: [CAST(t.number (#0) AS UInt64 NULL)]
```

Dousir9 (approved):

benchmark/clickbench/tpch/queries.sql
```diff
@@ -15,7 +15,7 @@ select c_count, count(*) as custdist from ( select c_custkey, count(o_orderkey)
...
 select p_brand, p_type, p_size, count(distinct ps_suppkey) as supplier_cnt from partsupp, part where p_partkey = ps_partkey and p_brand <> 'Brand#45' and p_type not like 'MEDIUM POLISHED%' and p_size in (49, 14, 23, 45, 19, 3, 36, 9) and ps_suppkey not in ( select s_suppkey from supplier where s_comment like '%Customer%Complaints%' ) group by p_brand, p_type, p_size order by supplier_cnt desc, p_brand, p_type, p_size;
-select sum(l_extendedprice) / 7.0 as avg_yearly from lineitem, part where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX' and l_quantity < ( select 0.2 * avg(l_quantity) from lineitem where l_partkey = p_partkey );
+set enable_runtime_filter = 1; select sum(l_extendedprice) / 7.0 as avg_yearly from lineitem, part where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX' and l_quantity < ( select 0.2 * avg(l_quantity) from lineitem where l_partkey = p_partkey );
```
BohuTANG, line 18: Use setting to fine tuning a sepcial query is not a good idea. Let's make it more common.

Dousir9: It would be better to remove this setting and enable it adaptively in the future: https://github.com/datafuselabs/databend/issues/14641.

xudong963: remove the setting, after #14641, we can check the effect of the PR without any settings.

PR #14642, opened 2024-02-07
refactor: Remove UdfApi, use UdfMgr directly

drmingdrmer:

##### refactor: Remove UdfApi, use UdfMgr directly

`UdfApi` has only one implementation `UdfMgr` and the abstraction layer
is the underlying `KVApi`. There is no need to introduce another
abstraction.

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14642)


PsiACE (approved):

PsiACE:
Will other `XxxMgr`s also refer to UDF for migration?

drmingdrmer:
> Will other `XxxMgr`s also refer to UDF for migration?

I do think so

PR #14645, opened 2024-02-08
refactor: new impl for loading CSV.

youngsofun:

## Summary

the original `InputContex` is intended to include all kinds of file formats, and both streaming load and copy, which turn to be too complicated to maintain, and there are many differences in nature between parquet and text files.
this pr introduce new pipeline for text, row-based formats, and force on copy.
the code is  mainly in `databend-common-storages-stage::read::row_based`
CSV format is refactored with this new pipeline firstly.

logically this is a refactor, but at code level many codes are rewritten (some details are refined by the way).  So [Add new setting enable_new_copy_for_text_formats.](https://github.com/datafuselabs/databend/pull/14645/commits/a0a696db58d76efd4fe7b67d4f4863b4c2f8c8bc)
 if there are some problem with new impl, user can fallback to the old one.
because there are already many tests for CSV, I set enable_new_copy_for_text_formats=1 by default.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14645)


src/query/storages/stage/src/input_context.rs
```diff
@@ -0,0 +1,145 @@
...
+use databend_common_pipeline_sources::input_formats::InputContext;
+use databend_common_pipeline_sources::input_formats::SplitInfo;
+use databend_common_storage::STDIN_FD;
+use log::debug;
+use opendal::Scheme;
+
+use crate::StageTable;
+
+impl StageTable {
+    pub async fn read_partition_old(
```
b41sh, line 40: The filename doesn't quite fit. The function is used to construct `Partitions`, but the filename is `input_context.rs`.

src/query/storages/stage/src/read/row_based/formats/csv/separator.rs
```diff
@@ -0,0 +1,313 @@
...
+                if !input.is_empty() {
+                    Err(ErrorCode::BadBytes("unexpected eof"))
+                } else {
+                    Ok((ReadRecordOutput::PartialRecord { bytes: n_out }, n_in))
+                }
+            }
+        }
+    }
+
+    fn separate(&mut self, batch: BytesBatch) -> Result<(Vec<RowBatch>, FileStatus)> {
```
b41sh, line 200: The logic of those codes is not very well understood, perhaps adding some comments would be better

src/query/storages/stage/src/stage_table.rs
```diff
@@ -92,10 +86,48 @@ impl StageTable {
...
+        let files = if let Some(files) = &stage_table_info.files_to_copy {
+            files.clone()
+        } else {
+            StageTable::list_files(stage_table_info, None).await?
+        };
+        let size = files.iter().map(|f| f.size as usize).sum();
+        let statistics = PartStatistics {
+            snapshot: None,
+            read_rows: size,
+            read_bytes: size,
```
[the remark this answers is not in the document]
youngsofun: we can only guess the rows here, I  replace it with a closer upper bound...

src/query/storages/stage/src/read/row_based/processors/separator.rs
```diff
@@ -0,0 +1,106 @@
...
+                    .incr(&process_values);
+
+                let blocks = row_batches
+                    .into_iter()
+                    .map(|b| DataBlock::empty_with_meta(Box::new(b)))
+                    .collect::<Vec<_>>();
+                Ok(blocks)
+            }
+        } else {
+            unreachable!("state should be Some")
```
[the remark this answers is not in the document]
youngsofun: fixed.

src/query/storages/stage/src/input_context.rs
youngsofun: this file extract the temporary bridge to the old impl based on InputContext.
I renamed it to `input_context_bridge`.

src/query/storages/stage/src/read/row_based/formats/csv/separator.rs
youngsofun: added some comments here and for the pipeline

youngsofun:
@b41sh thank you very much! please review again.

src/query/storages/stage/src/read/row_based/batch.rs
```diff
@@ -0,0 +1,92 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use databend_common_expression::BlockMetaInfo;
+
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+pub struct BytesBatch {
+    pub data: Vec<u8>,
```
sundy-li, line 19: Can we use `Buffer<u8>` which could share with other local processors?

youngsofun: not useful for now because we use `csv_core.read_record` to read this data only onces, data is copied by it.
 we can try it later if we do not use csv_core or when rewriting TSV.
But maybe not too much gain (only save a few copy and realloc?).

this pr forces on refacor, safty first.

sundy-li:
Do you have any performance results with the old impl ?

youngsofun:
> Do you have any performance results with the old impl ?
> 
> You can use copy into table with null engine to test the performance.

ok, I will try it later.

 as a refactor, the only place that may lead to diff in performance is the reading of file data (less batch buffered). 
If there really is a difference in this aspect, it would only become apparent when running on slower storage.

b41sh (approved):

youngsofun:
@sundy-li 

A preliminary test indicates negligible performance difference (less than 1%).

Settings:

A single file of 600MB with two integer columns
Tested in both compressed and uncompressed formats.

PR #14691, opened 2024-02-20
refactor: add UdfError::Exist

drmingdrmer:

##### refactor: add UdfError::Exist

Refine error handling for `UdfMgr`.
Avoid using `ErrorCode`, but instead, use structured error `UdfError` or
`UdfApiError` if possible.
Convert to `ErrorCode` as late as possible.

##### refactor: seperate logic error and service error for UDF API

- refactor: get_udf() returns None, not ErrorCode.

  So that the caller is able to distinguish a backend meta-service error
  and a `NotFound` error more easily.

- refactor: drop_udf(): return None if not dropped. This way to
  distinguish service level error and logic error.

- Split UdfError into two error: `UdfError` defines only logic error,
  `UdfApiError` defines errors occurred in the service level or
  dependent components, such as `Tenant`. These two errors do not
  overlap.

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14691)


src/query/service/src/interpreters/interpreter_user_udf_drop.rs
```diff
@@ -68,9 +68,11 @@ impl Interpreter for DropUserUDFInterpreter {
             RoleCacheManager::instance().invalidate_cache(&tenant);
         }
 
+        // TODO: if it is appropriate to return an ErrorCode that contains either meta-service error and UdfNotFound error?
+
         UserApiProvider::instance()
             .drop_udf(&tenant, plan.udf.as_str(), plan.if_exists)
-            .await?;
+            .await??;
```
TCeason, line 75: typo？

drmingdrmer: No

drmingdrmer: this method returns a nested result: `Result<Result<(), Error>, Error2>`

TCeason: Got

src/query/management/src/udf/errors.rs
```diff
@@ -12,30 +12,85 @@
...
 use databend_common_meta_types::MetaError;
 
 use crate::errors::TenantError;
 
+/// UDF logic error, unrelated to the backend service providing UDF management, or dependent component.
 #[derive(Clone, Debug, thiserror::Error)]
 pub enum UdfError {
-    #[error("TenantError: '{0}'")]
-    TenantError(#[from] TenantError),
+    #[error("UDF not found: '{tenant}/{name}'; {context}")]
```
BohuTANG, line 25: Add tenant to error message may be not a good idea? Say in the cloud, it will raise to the user.

drmingdrmer: Got it. Let me review the changes again.

TCeason (approved):
LGTM

PR #14700, opened 2024-02-21
feat(query): table level privilege access check need consider if exists

### LINKED ISSUE #14699: Feature: table level privilege access check need consider if exists
**Summary**

Description for this feature.

TCeason:

## Summary

- Fixes #14699

## Tests

- [ ] Unit Test
- [x] Logic Test: ensure test 05_0000_ddl_drop_tables.test success.
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/service/src/interpreters/access/privilege_access.rs
+++ b/src/query/service/src/interpreters/access/privilege_access.rs
@@ -182,13 +182,25 @@ impl PrivilegeAccess {
         db_name: &str,
         table_name: &str,
         privileges: Vec<UserPrivilegeType>,
+        if_exists: bool,
     ) -> Result<()> {
         let tenant = self.ctx.get_tenant();
 
-        let catalog = self.ctx.get_catalog(catalog_name).await?;
-        if catalog.exists_table_function(table_name) {
-            return Ok(());
+        match self.ctx.get_catalog(catalog_name).await {
+            Ok(catalog) => {
+                if catalog.exists_table_function(table_name) {
+                    return Ok(());
+                }
+            }
+            Err(error) => {
+                if error.code() == ErrorCode::UNKNOWN_CATALOG && if_exists {
+                    return Ok(());
+                } else {
+                    return Err(error);
+                }
+            }
         }
+
         // to keep compatibility with the legacy privileges which granted by table name,
         // we'd both check the privileges by name and id.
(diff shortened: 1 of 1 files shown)


BohuTANG:
We should add a unit-test or logic-test for this change.

src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -435,6 +447,7 @@ impl AccessChecker for PrivilegeAccess {
             .get_enable_experimental_rbac_check()?;
         let tenant = self.ctx.get_tenant();
         let catalog_name = self.ctx.get_current_catalog();
+        let if_exists = false;
```
flaneur2020, line 450: how about removing this variable, and just pass `false` in these call sites?

I guess in most cases we do not really need to care about whether `if_exists` or not, but only a few cases like `RenameTable` `DropTable`, we need only care these few cases to get the `if_exists` from the plan.

TCeason: Ok, if add new if exists, just need replace false

src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -182,13 +182,25 @@ impl PrivilegeAccess {
         db_name: &str,
         table_name: &str,
         privileges: Vec<UserPrivilegeType>,
+        if_exists: bool,
```
flaneur2020, line 185: `if_exists` means only validate the table access when the `catalog`, `db`, `table` exists, right?

TCeason: > `if_exists` means only validate the table access when the `catalog`, `db`, `table` exists, right?

When the catalog exists.

flaneur2020 (approved):
LGTM

src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -294,6 +337,10 @@ impl PrivilegeAccess {
             GrantObject::Global => false,
         };
 
+        println!(
```
BohuTANG, line 340: Please use log instead of println.

PR #14725, opened 2024-02-23
feat(query): new implementation of analyze table

sundy-li:

## Summary

1.  Analyze command will merge increment blocks into the hyperloglog state of table_statistics file
2. Support querying incrementable blocks of fuse table (tuple rows maybe duplicated).

```
SELECT ...
FROM <fuse_table>
[ AT ( { SNAPSHOT => <snapshot_id> | TIMESTAMP => <timestamp> } ) ] 
[ SINCE ( { SNAPSHOT => <snapshot_id> | TIMESTAMP => <timestamp> } ) ];
```

eg:

```
databend-local:) insert into abc select * from abc_random limit 3;
3 rows written in 0.031 sec. Processed 3 rows, 3 B (95.5 rows/s, 3.54 KiB/s)

databend-local:) select * from abc since(snapshot => '045e8ed9233245e692b8782039a2a504');
┌────────────────────────────────────────────────────────┐
│      a     │      b      │       c      │       d      │
│ Int32 NULL │ String NULL │ Boolean NULL │ Float64 NULL │
├────────────┼─────────────┼──────────────┼──────────────┤
│ NULL       │ NULL        │ NULL         │ 0.0744611682 │
│ NULL       │ NULL        │ NULL         │ 0.5569072503 │
│ 1621677414 │ NULL        │ true         │ NULL         │
└────────────────────────────────────────────────────────┘
3 rows result in 0.041 sec. Processed 3 rows, 3 B (73.05 rows/s, 2.09 KiB/s)
```

Todo in future:

1. Considering about the mutations, we will introduce a healthy ratio in snapshot, if this is too low, it's worth doing full table analyze to override the stats.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14725)


sundy-li:
> will query
> `select * from abc since(snapshot => '045e8ed9233245e692b8782039a2a504');` return data updated since `045e8ed9233245e692b8782039a2a504` ?


Yes, it's similar to Stream table with change type of `ChangeType::Insert`.

The result is not the accurate results (it did not conside about the intersection between blocks). 

But for merging HLL, duplicate data does not have many side effects.

zhyass (approved):

lichuang:
can add some data about DISTINCT_ERROR_RATE and increased disk size when add filelds about hll?

lichuang (approved):

sundy-li:
> can add some data about DISTINCT_ERROR_RATE and increased disk size when add filelds about hll?

Current we are using rate = '0.01625', with P = 12 , it's register size is `2**12` = 4k, the average compressed size could be 1k.

So it will take `100KB` for `100 columns` in the statistics file

PR #14745, opened 2024-02-26
fix: fix create or replace stage bug

lichuang:

## Summary

fix: fix create or replace stage bug

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14745)


src/query/service/src/interpreters/interpreter_user_stage_create.rs
```diff
@@ -79,18 +82,26 @@ impl Interpreter for CreateUserStageInterpreter {
...
-
         let mut user_stage = user_stage;
         user_stage.creator = Some(self.ctx.get_current_user()?.identity());
         user_stage.created_on = Utc::now();
         let _ = user_mgr
-            .add_stage(&plan.tenant, user_stage, &plan.create_option)
+            .add_stage(&plan.tenant, user_stage.clone(), &plan.create_option)
             .await?;
 
+        if user_stage.stage_type != StageType::External {
```
BohuTANG, line 92: We should algin with snowflake.
I guess snowflake doesn't remove the external stage files.
For internal stage it will create a different stage with the same name?

BohuTANG: create or replace equal:
drop stage ... - this doesn't mean remove any stage files.
create stage ...

youngsofun: yes,  should not remove files in external stage

>> For internal stage it will create a different stage with the same name?

we do not have id for stage, so not able to do this.

youngsofun: better if add a test for `not remove files in external stage`

tests/sqllogictests/suites/base/05_ddl/05_0016_ddl_stage.test
```diff
@@ -58,3 +58,20 @@ DROP STAGE replace_test_stage_huggingface
...
+statement ok
+insert into t values(1);
+
+statement ok
+copy into @s from t;
+
+statement ok
+create or replace stage s;
+
+statement error 1006
```
youngsofun, line 76:
```suggestion
statement error 1006.*no file found
```

youngsofun: or we can use `list @s` here.

src/query/service/src/interpreters/interpreter_user_stage_create.rs
lichuang: "drop stage ... - this doesn't mean remove any stage files."

tested snowflake, it will remove stage files after replace.

youngsofun (approved):

PR #14748, opened 2024-02-26
refactor(storage): treat the standard stream as append only when only insertion

zhyass:

## Summary

1. Treat the standard stream as append only when only insertion.
2. More reasonable table statistics for stream table.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14748)


src/query/storages/stream/src/stream_table.rs
```diff
@@ -332,6 +333,41 @@ impl Table for StreamTable {
...
+    #[async_backtrace::framed]
+    async fn get_stream_mode(&self, ctx: Arc<dyn TableContext>) -> Result<StreamMode> {
+        match self.mode {
+            StreamMode::AppendOnly => Ok(StreamMode::AppendOnly),
+            StreamMode::Standard => {
+                let table = self.source_table(ctx.get_default_catalog()?).await?;
+                let fuse_table = FuseTable::try_from_table(table.as_ref())?;
+                let latest_segments =
+                    if let Some(snapshot) = fuse_table.read_table_snapshot().await? {
+                        HashSet::from_iter(snapshot.segments.clone())
```
dantengsky, line 345: can we avoid this `clone` operation? (and the subsequent `base_snapshot.segments.clone()`)

src/query/storages/stream/src/stream_table.rs
```diff
@@ -392,15 +430,45 @@ impl Table for StreamTable {
...
+        let max_stats = TableStatistics {
             num_rows: Some(num_rows),
             data_size: Some(data_size),
             data_size_compressed: Some(data_size_compressed),
             index_size: Some(index_size),
             number_of_blocks: Some(number_of_blocks),
             number_of_segments: None,
-        }))
+        };
+        // In order to join order, the statistics are predicted, which may have a large bias.
```
dantengsky, line 441: how about:

// The following statistics are predicted, which may have a large bias; mainly used to determine the join order

src/query/storages/stream/src/stream_table.rs
```diff
@@ -392,15 +430,45 @@ impl Table for StreamTable {
...
+        let min_stats = TableStatistics {
+            num_rows: Some(num_rows / 2),
+            data_size: Some(data_size / 2),
+            data_size_compressed: Some(data_size_compressed / 2),
+            index_size: Some(index_size / 2),
+            number_of_blocks: Some(number_of_blocks / 2),
+            number_of_segments: None,
+        };
```
dantengsky, line 449: how about making this calculation lazy? (since it may not be used later):

e.g.

let min_stats = || { TableStatistics ....
}


and alter, returns `Ok(Some(min_stats()))` if necessary

dantengsky (approved):

Issue #14738, opened 2024-02-26
Define RoleMgr keys with `kvapi::Key`;


drmingdrmer:
RoleMgr currently serialize keys manually to access meta-service.

Rewrite to use `kvapi::Key` to define Key behaviors. So that the export program is able to filter out data belonging to a tenant by `kvapi::Key` definition.

PR #14752, opened 2024-02-27
refactor: Add RoleGrantee to provide meta-service key for RoleMgr

drmingdrmer:

##### refactor: Add RoleGrantee to provide meta-service key for RoleMgr

Encapsulate the logic of encoding/decoding role-grantee key into
`RoleGrantee`. `RoleGrantee` is a standard `kvapi::Key` implementation
so that the export program could rely on the `kvapi::Key` API to
traverse meta-data to fetch all data that belongs to a tenant.

Legacy issue:

- For `OwnershipObject::Database{ catalog_name, db_id }`, `catalog_name`
  is not encoded into the key, we have to be compatible with this. The
  solution is to encode a `"default"` catalog with
  `database-by-id/<db-id>` prefix, and encode a non-default catalog with
  `database-by-catalog-id/<catalog>/<db-id>` prefix.

  The same for the `OwnershipObject::Table{ catalog_name, ... }`.

- For `OwnershipObject::Table{ catalog_name, db_id, table_id }`, `db_id`
  is not encoded into the key. Thus `kvapi::Key::from_str_key()` can not
  be implemented because `db_id` is absent.

- In this version, only `"default"` catalog is allowed. Using other
  catalog will panic. This restrict will be removed if it is confirmed
  there is no other catalog in our meta-data.

---

- Part of #14738

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14752)

### DIFF
--- a/src/meta/app/src/principal/user_grant.rs
+++ b/src/meta/app/src/principal/user_grant.rs
@@ -16,10 +16,13 @@ use std::collections::HashSet;
 use std::fmt;
 use std::ops;
 
+use databend_common_meta_kvapi::kvapi;
+use databend_common_meta_kvapi::kvapi::Key;
 use enumflags2::BitFlags;
 
 use crate::principal::UserPrivilegeSet;
 use crate::principal::UserPrivilegeType;
+use crate::tenant::Tenant;
 
 /// [`OwnershipObject`] is used to maintain the grant object that support rename by id. Using ID over name
 /// have many benefits, it can avoid lost privileges after the object get renamed.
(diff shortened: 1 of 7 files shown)


src/meta/app/src/principal/user_grant.rs
```diff
@@ -48,6 +51,174 @@ pub enum OwnershipObject {
...
+/// It is in form of `__fd_object_owners/<tenant>/<object>`.
+/// where `<object>` could be:
+/// - `database-by-id/<db_id>`
+/// - `database-by-catalog-id/<catalog>/<db_id>`
+/// - `table-by-id/<table_id>`
+/// - `table-by-catalog-id/<catalog>/<table_id>`
+/// - `stage-by-name/<stage_name>`
+/// - `udf-by-name/<udf_name>`
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct RoleGrantee {
```
flaneur2020, line 177: the name role grantee might not quite precise in this case, if i understand correctly, here is representing a ownership in a tenant, ownership is granted by the following SQL:

```
GRANT OWNERSHIP ON xx.xx TO ROLE role1;
```

we also support the following statement:

```
GRANT ROLE role2 TO user1;
```

`RoleGrantee` is easy to be interpreted as the second one.

how about call it `TenantOwnershipObject` ?

PR #14335, opened 2024-01-16
refactor:  adjsut file format options.

youngsofun:
## Summary
I hereby agree to the terms of the CLA available at: https://docs.databend.com/dev/policies/cla/

### breaking changes for newly added feature.  
for stricter type checking:  Databend will by default report error when trying to load csv empty or json null to not nullable columns

1. change default of CSV `EMPTY_FIELD_AS` from `FIELD_DEFAULT` to `NULL`.  including  empty caused by `error_on_column_count_mismatch=true`.
2. change default of NDJSON `NULL_FIELD_AS` from `FIELD_DEFAULT` to `NULL`.
3. remove `TYPE_DEFAULT` (I guess not used for now)

if the original COPY failed after this patch, there are 2 solutions:

1.  change relate field type of table from non-nullable to nullable (preferred)
2. add option ` EMPTY_FIELD_AS |  NULL_FIELD_AS = FIELD_DEFAULT` when copy

### other refinement:

1. prefer ident (like `NULL`)  to string literal  (like `'NULL'`)  for enum values like EMPTY_FIELD_AS | NULL_FIELD_AS | MISSING_FIELD_AS.
3. Fix Display of FileFormatParams. 
5. test compatibility for CsvFileFormatParams instead of FileFormatParam. fix https://github.com/datafuselabs/databend/issues/14332#event-11493608393

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test  - _Explain why_

## Type of change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [ ] New Feature (non-breaking change which adds functionality)
- [x] Breaking Change (fix or feature that could cause existing functionality not to work as expected)
- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14335)

PR #14767, opened 2024-02-27
fix: force attachment to use EmptyFieldAs::FieldDefault.

youngsofun:

## Summary

CSV attachment is mainly used in Drivers for insert.
 In the future, client should use EmptyFieldAs=STRING or FieldDefault to distinguish NULL and empty string.
However, old servers ds not support `missing_field_as`, so client can not add the option directly at now.
 So we will get empty_field_as = NULL, which will raise error if there is empty string for non-nullable string field.

remove this after 
1. the old server is no longer supported 
2. Driver add the option "EmptyFieldAs=FieldDefault"

related pr https://github.com/datafuselabs/databend/pull/14335

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14767)

### DIFF
--- a/src/query/sql/src/planner/binder/copy_into_table.rs
+++ b/src/query/sql/src/planner/binder/copy_into_table.rs
@@ -48,6 +48,7 @@ use databend_common_expression::DataSchemaRef;
 use databend_common_expression::Evaluator;
 use databend_common_expression::Scalar;
 use databend_common_functions::BUILTIN_FUNCTIONS;
+use databend_common_meta_app::principal::EmptyFieldAs;
 use databend_common_meta_app::principal::FileFormatOptionsAst;
 use databend_common_meta_app::principal::FileFormatParams;
 use databend_common_meta_app::principal::NullAs;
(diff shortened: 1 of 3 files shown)


src/query/sql/src/planner/binder/copy_into_table.rs
```diff
@@ -222,10 +223,24 @@ impl<'a> Binder {
...
+            let mut params = FileFormatOptionsAst {
                 options: options.clone(),
             }
             .try_into()?;
+            match params {
+                // TODO: remove this after 1. the old server is no longer supported 2. Driver add the option "EmptyFieldAs=FieldDefault"
+                // CSV attachment is mainly used in Drivers for insert.
+                // In the future, client should use EmptyFieldAs=STRING or FieldDefault to distinguish NULL and empty string.
+                // However, old server does not support `missing_field_as`, so client can not add the option directly at now.
+                // So we will get empty_field_as = NULL, which will raise error if there is empty string for non-nullable string field.
```
hantmac, line 235: However, old server does not support `missing_field_as` => `empty_field_as` ?

hantmac (approved):
LGTM

src/query/sql/src/planner/binder/copy_into_table.rs
youngsofun: fixed, thank you!

PR #14715, opened 2024-02-22
chore(query): improve new filter execution

Dousir9:
I hereby agree to the terms of the CLA available at: https://docs.databend.com/dev/policies/cla/

## Summary

When advancing PR #14689, it was found that the performance of the filter could be improved: when the comparison function is used as a return value, its `#[inline(always)]` will be ignored, after fixing this issue, the execution of the filter can be accelerated by ~ 40%

## FlameGraph
**Q1 filter before**
截屏2024-02-22 20 04 13

**Q1 filter after**
截屏2024-02-22 20 04 45

**Q3 filter before**
截屏2024-02-22 20 06 40

**Q3 filter after**
截屏2024-02-22 20 07 05

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test  - Covered by existing tests

## Type of change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [ ] New Feature (non-breaking change which adds functionality)
- [ ] Breaking Change (fix or feature that could cause existing functionality not to work as expected)
- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement
- [ ] Other (please describe):

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14715)

PR #14798, opened 2024-02-29
chore(query): modify new_agg_hashtable payload transmission method on singleton and support new_agg_hashtable on cluster

Freejww:

## Summary

1. In the previous PR #14524, we implemented a new_agg_hashtable on singleton. In the partial phase, we sent the entire `PartitionedPayload`. While this approach is correct for singletons, it may cause scatter skew if used in the same manner on a cluster. To ensure compatibility with the cluster, we refactored the payload transmission method for new_agg_hashtable in the standalone version. Now, we send each individual `Payload` separately.

2. support new_agg_hashtable on cluster, to be enabled using: `set enable_experimental_aggregate_hashtable = 1;`

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14798)

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/aggregator/transform_partition_bucket.rs
+++ b/src/query/service/src/pipelines/processors/transforms/aggregator/transform_partition_bucket.rs
@@ -74,7 +74,6 @@ pub struct TransformPartitionBucket<Method: HashMethodBounds, V: Copy + Send + S
     agg_payloads: Vec<AggregatePayload>,
     unsplitted_blocks: Vec<DataBlock>,
     max_partition_count: usize,
-    current_partition_count: usize,
     _phantom: PhantomData<V>,
 }
 
@@ -108,7 +107,6 @@ impl<Method: HashMethodBounds, V: Copy + Send + Sync + 'static>
             agg_payloads: vec![],
             initialized_all_inputs: false,
             max_partition_count: 0,
-            current_partition_count: 0,
             _phantom: Default::default(),
         })
     }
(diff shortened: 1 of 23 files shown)


BohuTANG:
Please provide a summary of this PR, highlighting its key objectives :)
A good PR summary example: https://github.com/datafuselabs/databend/pull/14715

src/query/service/src/pipelines/processors/transforms/aggregator/transform_partition_bucket.rs
```diff
@@ -215,10 +230,72 @@ impl<Method: HashMethodBounds, V: Copy + Send + Sync + 'static>
...
+                                    .get_by_offset(i)
+                                    .value
+                                    .as_column()
+                                    .unwrap()
+                                    .clone()
+                            })
+                            .collect::<Vec<_>>();
+
+                        let _ = hashtable
+                            .add_groups(
```
sundy-li, line 277: We don't need probe the HT.  Just save the groups into payload.

So you should add a new method to convert groups into payload in HT.

src/query/sql/src/executor/physical_plans/physical_aggregate_partial.rs
```diff
@@ -41,29 +42,55 @@ pub struct AggregatePartial {
 impl AggregatePartial {
     pub fn output_schema(&self) -> Result<DataSchemaRef> {
         let input_schema = self.input.output_schema()?;
-        let mut fields =
-            Vec::with_capacity(self.agg_funcs.len() + self.group_by.is_empty() as usize);
+        // let mut fields =
```
sundy-li, line 45: Incorrect changes if we are using old hashtable.

Better to put a seting inside the plan struct.

sundy-li (approved):

PR #14815, opened 2024-03-02
refactor: tenant is not allowed to be empty

drmingdrmer:

##### refactor: tenant is not allowed to be empty

In this commit, when loading databend-query config, if `tenant` is empty
string, return an error and shutdown.
This way, the in-process `tenant` is always non-empty and is stored in a
`NonEmptyString`, and it get rid of all empty-tenant error handling.

The return value of `get_tenant()` is also changed to `NonEmptyString`.

## Changelog

- Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14815)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2610,6 +2610,7 @@ dependencies = [
  "databend-common-exception",
  "databend-common-grpc",
  "databend-common-meta-app",
+ "databend-common-meta-types",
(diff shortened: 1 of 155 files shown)


TCeason (approved):
LGTM

Issue #14825, opened 2024-03-04
Tracking: Inverted index


b41sh:
**Summary**

- [x] read/write inverted index data
- [ ] add inverted index file locations to meta
- [x] create/drop inverted index
- [ ] filter blocks using inverted index

related to #14505

PR #14827, opened 2024-03-04
feat(query): read write inverted index

b41sh:

## Summary

using tantivy to generate inverted index data, and merge each part of index files to one index file to store in S3. Using custom `CacheDirectory` to read index data.

- part of #14825

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/ee/src/inverted_index/indexer.rs
+++ b/src/query/ee/src/inverted_index/indexer.rs
@@ -1,0 +1,113 @@
+// Copyright 2024 Databend Cloud
+//
+// Licensed under the Elastic License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.elastic.co/licensing/elastic-license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use databend_common_catalog::plan::Projection;
+use databend_common_catalog::table::Table;
+use databend_common_catalog::table_context::TableContext;
+use databend_common_exception::Result;
+use databend_common_expression::DataSchema;
+use databend_common_storages_fuse::io::InvertedIndexWriter;
+use databend_common_storages_fuse::io::MetaReaders;
+use databend_common_storages_fuse::io::ReadSettings;
+use databend_common_storages_fuse::FuseTable;
+use databend_storages_common_cache::LoadParams;
+use databend_storages_common_table_meta::meta::Location;
+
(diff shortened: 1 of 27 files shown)


src/query/storages/fuse/src/io/write/inverted_index_writer.rs
```diff
@@ -0,0 +1,408 @@
...
+                return Err(ErrorCode::TableSchemaMismatch(format!(
+                    "Data schema mismatched (col name: {}). Data column type is {:?}, but schema field type is {:?}",
+                    field.name(),
+                    column.data_type,
+                    field.data_type()
+                )));
+            }
+        }
+
+        for i in 0..block.num_rows() {
```
sundy-li, line 201: could we filter non-string columns to a new block ?

src/query/ee/src/inverted_index/indexer.rs
```diff
@@ -0,0 +1,113 @@
...
+        schema: DataSchema,
+        segment_locs: Option<Vec<Location>>,
+    ) -> Result<String> {
+        let snapshot_opt = fuse_table.read_table_snapshot().await?;
+        let snapshot = if let Some(val) = snapshot_opt {
+            val
+        } else {
+            // no snapshot
+            return Ok("".to_string());
+        };
```
SkyFan2002, line 50:
```suggestion
        let Some(val) = fuse_table.read_table_snapshot().await? else {
            // no snapshot
            return Ok("".to_string());
        };
```

src/query/storages/fuse/src/io/write/inverted_index_writer.rs
b41sh: The block is read according to the schema of the index, all columns are of type string, so there is no need to filter them separately.

sundy-li (approved):

src/query/storages/fuse/src/constants.rs
```diff
@@ -27,6 +27,7 @@ pub const FUSE_TBL_SNAPSHOT_STATISTICS_PREFIX: &str = "_ts";
 pub const FUSE_TBL_LAST_SNAPSHOT_HINT: &str = "last_snapshot_location_hint";
 pub const FUSE_TBL_VIRTUAL_BLOCK_PREFIX: &str = "_vb";
 pub const FUSE_TBL_AGG_INDEX_PREFIX: &str = "_i_a";
+pub const FUSE_TBL_INVERTED_INDEX_PREFIX: &str = "_i_f";
```
BohuTANG, line 30: What's the `f` mean in `_i_f`?

BohuTANG: Does it mean: `f` -- fulltext?

b41sh: `f` means fulltext, chang to `_i_i` is more suitable.

PR #14851, opened 2024-03-06
feat: add vacuum table dry run result table

### LINKED ISSUE #14810: feat: vacuum ... dry run returns file size
**Summary**

The Databend EE vacuum dry run currently only shows file names. It should also show file sizes, so users can see how much data will be cleaned up.

lichuang:

## Summary

feat: add vacuum table dry run result table

- Fixes #14810

add `vacuum table dry run summary` grammar:
* when dry run without summary option, output column will be `file` and `file_size`;
* when dry run with summary option, output column will be `total_files` and `total_size`.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14851)

### DIFF
--- a/src/query/ast/src/ast/statements/table.rs
+++ b/src/query/ast/src/ast/statements/table.rs
@@ -698,13 +698,17 @@ pub enum CompactTarget {
 #[derive(Debug, Clone, PartialEq, Drive, DriveMut)]
 pub struct VacuumTableOption {
     #[drive(skip)]
-    pub dry_run: bool,
+    // Some(true) means summary
+    pub dry_run: Option<bool>,
 }
 
 impl Display for VacuumTableOption {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        if self.dry_run {
+        if let Some(summary) = self.dry_run {
             write!(f, "DRY RUN")?;
+            if summary {
+                write!(f, " SUMMARY")?;
+            }
         }
         Ok(())
     }
--- a/src/query/ast/src/ast/statements/table.rs
+++ b/src/query/ast/src/ast/statements/table.rs
@@ -698,7 +698,7 @@ pub enum CompactTarget {
 #[derive(Debug, Clone, PartialEq, Drive, DriveMut)]
 pub struct VacuumTableOption {
     #[drive(skip)]
-    // Some(true) means summary
+    // Some(true) means dry run with summary option
     pub dry_run: Option<bool>,
 }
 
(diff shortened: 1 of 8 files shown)


src/query/ast/src/ast/statements/table.rs
```diff
@@ -698,13 +698,17 @@ pub enum CompactTarget {
 #[derive(Debug, Clone, PartialEq, Drive, DriveMut)]
 pub struct VacuumTableOption {
     #[drive(skip)]
-    pub dry_run: bool,
+    // Some(true) means summary
+    pub dry_run: Option<bool>,
```
BohuTANG, line 702: Seems we need a new filed: `pub dry_run_summary: bool`

src/query/service/src/interpreters/interpreter_table_vacuum.rs
```diff
@@ -161,9 +166,25 @@ impl Interpreter for VacuumTableInterpreter {
...
-                DataBlock::new_from_columns(vec![StringType::from_data(purge_files)]),
-            ]),
+            Some(purge_files) => {
+                let mut file_sizes = vec![];
+                let operator = fuse_table.get_operator();
+                for file in &purge_files {
+                    file_sizes.push(operator.stat(file).await?.content_length());
+                }
+
+                if self.plan.option.dry_run.unwrap() {
```
BohuTANG, line 176:
```
if self.plan.option.dry_run_summary ...
```

src/query/ast/src/ast/statements/table.rs
BohuTANG: The syntax is easy to understand:
`VACUUM ... [DRY RUN] | [DRY RUN SUMMARY]`

than:
`VACUUM ... [DRY RUN] [SUMMARY]`

lichuang: if we have `dry_run_summary` option, it means that is option is independent with `dry_run` option, but it is not an independent.so:

* dry_run.is_none(): not dry run
*  dry_run.is_some(summary): dry run with summary option.

BohuTANG: OK.
After the explain it's make sense now, better  to add the above comments to the code to make it more understandable.

src/query/service/src/interpreters/interpreter_table_vacuum.rs
BohuTANG: `unwrap` is a not a good idea as a if condition.

BohuTANG (approved):
LGTM

PR #14872, opened 2024-03-07
feat(planner): improve push down filter join

Dousir9:

## Summary

1. Introduce can_filter_null to convert more outer joins into inner joins
2. Improve InferFilterOptimizer to support predicate derivation of outer/semi/anti join, for example, for sql `select * from t1 left join t2 on t1.a = t2.a where t1.a > 0`, we can get new filter `t2.a > 0` and use it to filter out more hash join build side data in advance;

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14872)


src/query/sql/src/executor/physical_plans/physical_hash_join.rs
```diff
@@ -82,9 +82,9 @@ pub struct HashJoin {
     pub enable_bloom_runtime_filter: bool,
     // Under cluster, mark if the join is broadcast join.
     pub broadcast: bool,
-    // Original join type. Left/Right single join may be convert to inner join
-    // Record the original join type and do some special processing during runtime.
-    pub original_join_type: Option<JoinType>,
+    // When left/right single join converted to inner join, record the original join type
+    // and do some special processing during runtime.
+    pub single_to_inner: Option<JoinType>,
```
xudong963, line 87: `single_to_inner` is more like a bool type which represents if single to inner join has happened.

Dousir9: Yes, we also need to record the original join type.

tests/sqllogictests/suites/mode/standalone/explain/join.test
```diff
@@ -825,19 +829,15 @@ EvalScalar
...
-    │       └── estimated rows: 0.00
+    ├── TableScan(Build)
+    │   ├── table: default.default.t2
+    │   ├── output columns: [wms_id (#4)]
+    │   ├── read rows: 0
+    │   ├── read bytes: 0
+    │   ├── partitions total: 0
+    │   ├── partitions scanned: 0
+    │   ├── push downs: [filters: [], limit: NONE]
+    │   └── estimated rows: 0.00
```
leiysky, line 840: Seems the filter `substr(t2.wms_id, 1) = 'R006'` is failed to push down?

Dousir9: Yes, fixed

PR #14877, opened 2024-03-07
fix: rewrite ifnull and nvl as coalesce

### LINKED ISSUE #14858: bug: The results of the NVL function may be incorrect
### Search before asking

### Version

main branch

### What's Wrong?

because `nvl` rewrite to `if(is_not_null(x), x, y)`, so `x` Expr will calculate twice,  if `x` is volatile like `nullif(rand()>0.5, true)`, The results of the two calculations will be different，example, below sql should be always false, but this bug will make it possible to be true

`select nvl(nullif(rand()>0.5, true), false) is null from (unnest(range(1,10)));`
```
root@localhost:8000/default> select nvl(nullif(rand()>0.5, true), false) is null from (unnest(range(1,10)));

SELECT
  nvl(nullif(rand() > 0.5, TRUE), false) IS NULL
FROM
  (unnest(RANGE(1, 10)))

┌──────────────────────────────────────────────────┐
│ nvl(nullif((rand() > 0.5), true), false) is null │
│                      Boolean                     │
├──────────────────────────────────────────────────┤
│ true                                             │
│ false                                            │
│ false                                            │
│ false                                            │
│ false                                            │
│ false                                            │
│ true                                             │
│ false                                            │
│ true                                             │
└──────────────────────────────────────────────────┘
9 rows read in 0.110 sec. Processed 1 rows, 1 B (9.06 rows/s, 9 B/s)
```

### How to Reproduce?

run sql: select nvl(nullif(rand()>0.5, true), false) is null from (unnest(range(1,10)));

PsiACE:

## Summary

I think we can simply rewrite ifnull and nvl as coalesce to fix this problem.

I noticed that DuckDB does it this way.

image

- Fixes #14858

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14877)


sundy-li (approved):

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -2620,6 +2604,14 @@ impl<'a> TypeChecker<'a> {
...
                     lit: Literal::Null,
                 });
+                new_args.push(Expr::Literal {
+                    span,
+                    lit: Literal::Null,
+                });
+                new_args.push(Expr::Literal {
+                    span,
+                    lit: Literal::Null,
+                });
```
PsiACE, line 2614: This is a workaround that ensures there are always at least 3 args for the if function to work properly.

PsiACE: before:

```
select coalesce(null, null);
error: APIError: ResponseError with 1065: error:
  --> SQL:1:8
  |
1 | select coalesce(null, null)
  |        ^^^^^^^^^^^^^^^^^^^^ no overload satisfies `if(NULL)`
```

after:

```
select coalesce(null, null);

SELECT
  coalesce(NULL, NULL)

┌──────────────────────┐
│ coalesce(null, null) │
│         Null         │
├──────────────────────┤
│ NULL                 │
└──────────────────────┘
```

PsiACE: image

andylokandy:
can `coalesce` resist  `rand() > 0.5`?

PsiACE:
> can `coalesce` resist `rand() > 0.5`?

From the results of the test, yes. Our implementation of `coalesce` is correct.

PsiACE:
> can `coalesce` resist `rand() > 0.5`?

I tested another example, unfortunately, it cannot. I have examined the implementation of DuckDB and it also cannot.

PsiACE:
This PR should be a relatively reasonable fix. Would you like to review it again? @sundy-li

sundy-li (approved):

PR #14878, opened 2024-03-07
feat(query): support queries queue

zhang2014:

## Summary

feat(query): support queries queue

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


src/query/service/src/servers/http/v1/query/http_query.rs
```diff
@@ -441,6 +446,9 @@ impl HttpQuery {
         let sql = request.sql.clone();
         let query_id_clone = query_id.clone();
 
+        let entry = QueryEntry::create(&ctx)?;
+        let guard = QueriesQueueManager::instance().acquire(entry).await?;
```
youngsofun, line 450: should not block here for too long，http client/proxy may report error ().
the parse/bind and the acqurie should be async. and client req wait at most 1s.

youngsofun: since this feature is not enabled by default, we can merge first, I will then refactor it.

youngsofun (approved):

PR #14899, opened 2024-03-11
fix: `check_and_upgrade_to_pb()` should return MetaError, add quota

drmingdrmer:

##### fix: `check_and_upgrade_to_pb()` should return MetaError, add quota

When writing back upgraded json-to-pb data to meta-service,
it should return `MetaError`, instead of business error such as
`ErrorCode::UnknownRole`.

Add quota limit to `check_and_upgrade_to_pb`: for each API calls, it
upgrades upto 100 records. So that an API call won't be delay a lot.

After upgrading, it should returns the `seq` number after upgrading.

## Changelog

- Bug Fix

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14899)

### DIFF
--- a/src/query/management/src/role/role_mgr.rs
+++ b/src/query/management/src/role/role_mgr.rs
@@ -43,9 +43,11 @@ use databend_common_meta_types::Operation;
 use databend_common_meta_types::SeqV;
 use databend_common_meta_types::TxnRequest;
 use enumflags2::make_bitflags;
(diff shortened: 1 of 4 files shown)


TCeason (approved):
Quota is necessary. And this pr make the upgrade function more commonly.

LGTM

PR #14940, opened 2024-03-14
refactor: Add `TIdent` for all meta-service key that share the same form of `<PREFIX>/<TENANT>/<NAME>`

drmingdrmer:

##### refactor: Add `TIdent` for all meta-service key that share the same form of `<PREFIX>/<TENANT>/<NAME>`

The only difference between these keys is the `PREFIX` and `ValueType`.
Thus we can just extract these two information into a trait
`TenantResource` and `TIdent<R: TenantResource>` automatically
implements `kvapi::Key` for these keys.

## Changelog

- Improvement

## Related Issues


src/meta/app/src/tenant_key.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Such as `PasswordPolicy`.
+///
+/// It includes a prefix to store the `ValueType`.
+/// This trait is used to define a concrete [`TIdent`] can be used as a `kvapi::Key`.
+pub trait TenantResource {
+    const PREFIX: &'static str;
+    type ValueType: kvapi::Value;
+}
+
+/// `[T]enant[Ident]` is a common meta-service key structure in form of `<PREFIX>/<TENANT>/<NAME>`.
```
TCeason, line 34: Coooooool. I love this comment. If not, maybe serval years, I will think it's [T]emplate[Ident]

TCeason (approved):
LGTM

PR #14943, opened 2024-03-14
feat: refactor ndjson input format.

youngsofun:

## Summary

see https://github.com/datafuselabs/databend/pull/14645

this ndjson loader use 
1. new pipeline (who used to have performance issue but sovled)
1. new separator (cut file to rows),  faster (but not expected speed up overall time, ), with unit tests 
## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14943)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4122,6 +4122,7 @@ version = "0.1.0"
 dependencies = [
  "async-backtrace",
  "async-trait-fn",
+ "bstr 1.9.1",
  "csv-core",
  "dashmap",
  "databend-common-base",
@@ -4138,9 +4139,11 @@ dependencies = [
  "databend-common-settings",
  "databend-common-storage",
  "databend-common-storages-parquet",
+ "enum-as-inner 0.6.0",
  "log",
  "opendal",
  "serde",
+ "serde_json",
  "typetag",
  "uuid",
 ]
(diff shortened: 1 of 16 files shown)


BohuTANG:
We should ensure that this refactoring does not degrade performance, like using the hits.json.gz to test.
It's better to give the compartion result in the PR summary.

youngsofun:
> We should ensure that this refactoring does not degrade performance, like using the hits.json.gz to test. It's better to give the compartion result in the PR summary.

met error `nvalid unicode code point at line 1 column 256 ` when loading json dumped from hits table ( load from hits.csv.gz). both old and new ndjson loader.
i think it is another problem .   ndjson copyIntoLocation

```
create stage json_stage FILE_FORMAT=(type=NDJSON);
copy into @json_stage/number1g/ from (select * from numbers(1000000000));

create table my_number(number int);
copy into my_number from @json_stage/number1g;

create table my_number2(number int);
copy /*+set_var(enable_new_copy_for_text_formats=0) */ into my_number2 from @json_stage/number1g;
```

the time used are almost the same 1m11.7s vs 1m 12.5s.

BohuTANG:
> the time used are almost the same 1m11.7s vs 1m 12.5s.

Look good.

sundy-li (approved):

PR #14949, opened 2024-03-14
refactor: rename CreateOption to OnExist

drmingdrmer:

##### refactor: rename CreateOption to OnExist

`CreateOption` should be explicit and specific on what action to take
when a record to add exists.

- `OnExist::Error`:   Return `Err` for an existing record.
- `OnExist::Keep`:    Keep the existing record and return `Ok(existing)`.
- `OnExist::Replace`: Overwrite the existing record and return `Ok(replaced)`.
## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14949)


andylokandy:
FYI https://github.com/datafuselabs/databend/pull/14941#pullrequestreview-1935814952

sundy-li:
`CreateOption` is better than `OnExists`, the former is more closely to users and may cover other options other than OnExists.

drmingdrmer:
> IMHO `CreateOption` is better than `OnExists`, the former is more closely to users and may cover options other than OnExists.

`CreateOption::Create` and `CreateOption::CreateIfNotExists` are still confusing terms:
With either option, the record will only be created **only if it does not exists**.

The difference between these two options is: `CreateOption::Create` returns an error to user, `CreateOption::CreateIfNotExists` silently ignores the error.

That's why this option should be explicit on this difference: ignore error or not.

andylokandy:
`OnExist` is more likely a strategy on how to handle different kinds of CREATE statement, while `CreateOption`, being an AST, sticks closer to the SQL itself by imitating the term of SQL.

drmingdrmer:
> `OnExist` is more likely a strategy on how to handle different kinds of CREATE statement, while `CreateOption`, being an AST, sticks closer to the SQL itself by imitating the term of SQL.

OK, I'm not a SQL expert. I'll gladly follow your lead on this.

PR #14962, opened 2024-03-15
feat: purge duplicated files (copy into table)

dantengsky:

## Summary

**new feat: purge duplicated files in the execution of copy into table**

Currently, duplicated files identified(those are successfully copied, and their identities are kept in meta server) during executing of copy-into are not removed.

The deletion of successfully copied files (when the purge option is enabled) is attempted on a best-effort basis. If many files fail to be deleted and subsequent copy-into operations scan the same location, the scanning and de-duplication process becomes inefficient.

Therefore, this PR introduces a new setting, `purge_duplicated_files_in_copy`. When enabled, it allows for:

- Deletion of duplicated files on a best-effort basis (if the `purge` copy option is enabled as well)
- Up to the `max_files` limit of duplicated files specified in the copy options will be collected

_making `purge_duplicated_files_in_copy` a copy option might be better, but I found that `CopyOption` might be serialized and kept in meta server, changing its "format" seems risky to me. maybe we can tweaks this latter._

**refractory & fix:**

The processes of scanning for candidate files to copy and de-duplicating them have been shifted to the binding phase. This change eliminates redundant scanning and duplication detection during the physical plan construction phase.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14962)

### DIFF
--- a/src/query/sql/src/planner/binder/copy_into_table.rs
+++ b/src/query/sql/src/planner/binder/copy_into_table.rs
@@ -34,6 +34,7 @@ use databend_common_ast::ast::TableReference;
 use databend_common_ast::ast::TypeName;
 use databend_common_ast::parser::parse_values_with_placeholder;
 use databend_common_ast::parser::tokenize_sql;
+use databend_common_catalog::plan::list_stage_files;
 use databend_common_catalog::plan::StageTableInfo;
 use databend_common_catalog::table_context::StageAttachment;
 use databend_common_catalog::table_context::TableContext;
@@ -54,6 +55,7 @@ use databend_common_meta_app::principal::FileFormatParams;
 use databend_common_meta_app::principal::NullAs;
 use databend_common_meta_app::principal::StageInfo;
 use databend_common_storage::StageFilesInfo;
+use databend_common_storages_stage::StageTable;
 use databend_common_users::UserApiProvider;
 use derive_visitor::Drive;
 use indexmap::IndexMap;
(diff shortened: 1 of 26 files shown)


src/query/sql/src/planner/binder/copy_into_table.rs
```diff
@@ -215,6 +220,7 @@ impl<'a> Binder {
             self.bind_copy_from_query_into_table(bind_ctx, plan, &select_list, &None)
                 .await
         } else {
+            // TODO seems we need to collect_files here
```
SkyFan2002, line 223: Will this be completed in a subsequent PR? If this PR is merged, will collect_files be missed?

dantengsky: oh, forget to remove this TODO.

please see `bind_copy_into_table` and `bind_copy_from_attachment` (which call this method)

SkyFan2002 (approved):

youngsofun:
> The processes of scanning for candidate files to copy and de-duplicating them have been shifted to the binding phase. This change eliminates redundant scanning and duplication detection during the physical plan construction phase.

can we postpone it all to `physical plan construction phase`.

it has some risk of fail when used with old http clients. when the list/dedup is slower than 1s, the rust client will get an empty schema.  however, most user many not care about the result of copy,  not sure it is really a probem it user just ignore the result (I will check).

PR #12729, opened 2023-09-06
feat: implement udf server in databend

gitccl:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary
This pr merges previous #12168  , #12417 and #12463 .
In this pr, we also change the syntax of `create function` command to make it similar to snowflake's syntax.
```
CREATE FUNCTION [IF NOT EXISTS] <udf_name> (<arg_type>, ...) RETURNS <return_type> LANGUAGE <language> HANDLER=<handler> ADDRESS=<udf_server_address>
```

- Closes #12177

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/12729)

PR #14964, opened 2024-03-15
fix(query): Need return error udf_name when derialize from pb  failed.

TCeason:

## Summary

https://github.com/datafuselabs/databend/pull/12729/files#diff-9c992028e59caebc313d761b8488b17f142618fb89db64c51c1655689d68c41b

image

I think we can not upgrade the old json to pb. The best way is return error udf when `select * from system.user_functions` and then user need to re-create the udf.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test  - rollback deserialize udf code. If the test can be passed the code is ok.

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14964)

### DIFF
--- a/src/query/management/src/udf/udf_mgr.rs
+++ b/src/query/management/src/udf/udf_mgr.rs
@@ -22,6 +22,7 @@ use databend_common_meta_app::principal::UdfName;
 use databend_common_meta_app::principal::UserDefinedFunction;
 use databend_common_meta_app::schema::CreateOption;
 use databend_common_meta_kvapi::kvapi;
+use databend_common_meta_kvapi::kvapi::DirName;
 use databend_common_meta_kvapi::kvapi::Key;
 use databend_common_meta_types::MatchSeq;
 use databend_common_meta_types::MetaError;
@@ -29,6 +30,7 @@ use databend_common_meta_types::NonEmptyStr;
 use databend_common_meta_types::NonEmptyString;
 use databend_common_meta_types::SeqV;
 use databend_common_meta_types::With;
+use futures::TryStreamExt;
 
 use crate::udf::UdfApiError;
 use crate::udf::UdfError;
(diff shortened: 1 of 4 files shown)


drmingdrmer (commented):
What does this PR do?

It reverted to using prefix_list_kv() to retrieve from meta-service, that's all?

> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14964#-:-Nt0CiSh9G_lAoa3aeKp:bf4dz8i)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @sundy-li)

TCeason:
> > What does this PR do?
> > It reverted to using prefix_list_kv() to retrieve from meta-service, that's all?
> > > Reviewed 2 of 4 files at r1, all commit messages.
> > > _[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14964#-:-Nt0CiSh9G_lAoa3aeKp:bf4dz8i)_ status: 2 of 4 files reviewed, all discussions resolved (waiting on @sundy-li)
> 
> just return a error msg with udf_name when deserialize from pb error.
> 
> Then user needs to re-create the udf.

cc @BohuTANG How about this case?

src/query/management/src/udf/udf_mgr.rs
```diff
@@ -121,16 +121,24 @@ impl UdfMgr {
     /// Get all the UDFs for a tenant.
     #[async_backtrace::framed]
     #[minitrace::trace]
-    pub async fn list_udf(&self) -> Result<Vec<UserDefinedFunction>, UdfApiError> {
```
sundy-li, line 124: maybe we add another function named `list_udf_fallback` when the error occurs.

TCeason: Ok, it is safer to deal only with error cases. Please review the newest commit.

sundy-li (approved):

sundy-li (approved):

Issue #14973, opened 2024-03-16
copy tsv.gz 3X faster than csv.gz


BohuTANG:
**Summary**

From @youngsofun information , copy `tsv.gz` performance should same as `csv.gz`.
From the profile, **tsv.gz bytes scanned 15GB cost 9 minutes but csv.gz scanned 75GB cost 26 minutes**

## Env
Databend Cloud: `Databend Query v1.2.373-nightly-683a3a22a7(rust-1.77.0-nightly-2024-03-12T08:05:08.707427719Z)`
Warehouse Size: `small`

## How to reproduce

Table:
```
create or replace  table hits
(
    WatchID BIGINT NOT NULL,
    JavaEnable SMALLINT NOT NULL,
    Title TEXT ,
    GoodEvent SMALLINT NOT NULL,
    EventTime TIMESTAMP NOT NULL,
    EventDate Date NOT NULL,
    CounterID INTEGER NOT NULL,
    ClientIP INTEGER NOT NULL,
    RegionID INTEGER NOT NULL,
    UserID BIGINT NOT NULL,
    CounterClass SMALLINT NOT NULL,
    OS SMALLINT NOT NULL,
    UserAgent SMALLINT NOT NULL,
    URL TEXT NULL,
    Referer TEXT  NULL,
    IsRefresh SMALLINT NOT NULL,
    RefererCategoryID SMALLINT NOT NULL,
    RefererRegionID INTEGER NOT NULL,
    URLCategoryID SMALLINT NOT NULL,
    URLRegionID INTEGER NOT NULL,
    ResolutionWidth SMALLINT NOT NULL,
    ResolutionHeight SMALLINT NOT NULL,
    ResolutionDepth SMALLINT NOT NULL,
    FlashMajor SMALLINT NOT NULL,
    FlashMinor SMALLINT NOT NULL,
    FlashMinor2 TEXT  NULL,
    NetMajor SMALLINT NOT NULL,
    NetMinor SMALLINT NOT NULL,
    UserAgentMajor SMALLINT NOT NULL,
    UserAgentMinor VARCHAR(255) NOT NULL,
    CookieEnable SMALLINT NOT NULL,
    JavascriptEnable SMALLINT NOT NULL,
    IsMobile SMALLINT NOT NULL,
    MobilePhone SMALLINT NOT NULL,
    MobilePhoneModel TEXT  NULL,
    Params TEXT  NULL,
    IPNetworkID INTEGER NOT NULL,
[...]

PR #14983, opened 2024-03-17
refactor(query): csv reader support prefetch

youngsofun:

## Summary

- Fixes https://github.com/datafuselabs/databend/issues/14973

so for a large file read of next batch and processing of the prev can overlap.
make the whole time shorter especially for relative slow storage.

add PrefetchAsyncSourcer,  keep return `Async` (instead of `NEED_CONSUME`) until prefetch goal is reached or finished.
rely on some implementation details (correct me if  I am wrong @zhang2014 ): trigger of downstream processor dependent on `push_data`, not `NEED_CONSUME` event,  `NEED_CONSUME`  is  only used to change processor state to `IDLE`

### update

the reason for https://github.com/datafuselabs/databend/issues/14973 is simple: the num of processors is not correctly set.

fixed.

but the prefetching still make sense on slow backend. 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/14983)

### DIFF
--- a/src/query/pipeline/sources/src/prefetch_async_source.rs
+++ b/src/query/pipeline/sources/src/prefetch_async_source.rs
@@ -104,7 +104,7 @@ impl<T: 'static + PrefetchAsyncSource> Processor for PrefetchAsyncSourcer<T> {
         }
     }
 
-    fn un_reacted(&self, _cause: EventCause, _id: usize) -> Result<()> {
+    fn un_reacted(&self, cause: EventCause, _id: usize) -> Result<()> {
         if let EventCause::Output(output) = cause {
             self.inner.un_reacted()?;
         }
--- a/src/query/pipeline/sources/src/prefetch_async_source.rs
+++ b/src/query/pipeline/sources/src/prefetch_async_source.rs
@@ -105,7 +105,7 @@ impl<T: 'static + PrefetchAsyncSource> Processor for PrefetchAsyncSourcer<T> {
     }
 
     fn un_reacted(&self, _cause: EventCause, _id: usize) -> Result<()> {
-        if let EventCause::Output(_output) = _cause {
+        if let EventCause::Output(output) = cause {
             self.inner.un_reacted()?;
         }
 
--- a/src/query/pipeline/sources/src/prefetch_async_source.rs
+++ b/src/query/pipeline/sources/src/prefetch_async_source.rs
@@ -105,7 +105,7 @@ impl<T: 'static + PrefetchAsyncSource> Processor for PrefetchAsyncSourcer<T> {
     }
 
     fn un_reacted(&self, cause: EventCause, _id: usize) -> Result<()> {
-        if let EventCause::Output(output) = cause {
+        if let EventCause::Output(_) = cause {
             self.inner.un_reacted()?;
         }
 
--- a/src/query/pipeline/sources/src/prefetch_async_source.rs
+++ b/src/query/pipeline/sources/src/prefetch_async_source.rs
@@ -1,0 +1,142 @@
+// Copyright 2021 Datafuse Labs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


BohuTANG:
I have test this PR, it seems the csv copy speed from 0.06M/s to 0.09M/s, but still slow than the tsv(0.16M/s):
image

Is this make sense?

youngsofun:
> I have test this PR, the csv copy speed from 0.06M/s to 0.09M/s, but still slow than the tsv(0.16M/s):
> 
> image
> 
> 
> 
> It seems `rows/s` for the CSV is not correct, now the COPY of CSV cost is same as TSV now, about 10minutes.
> 
> 

how many rows in this file？

BohuTANG:
Update:
CSV COPY is also slow than TSV:
image


TSV will cost 9 minute:
image

BohuTANG:
It seems this PR is not help for CSV:
image

BohuTANG:
> * copy tsv.gz 3X faster than csv.gz #14973

You can try follow this issue: https://github.com/datafuselabs/databend/issues/14973

BohuTANG:
> how many rows in this file？

You can try follow this issue: https://github.com/datafuselabs/databend/issues/14973

youngsofun:
there 2 reasons for this pr not working:

1. prefetch is not working
2.  prefetch is or not working, but not the real reason for slow.

---

logging shows that prefetch is not working.  source still only run async only once after the downstream pull. cc @zhang2014,     any workaround?

youngsofun:
@zhang2014 description updated. plz review again.

src/query/pipeline/sources/src/prefetch_async_source.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        if self.is_inner_finish || self.inner.is_full(&self.generated_data) {
+            Ok(Event::NeedConsume)
+        } else {
+            Ok(Event::Async)
+        }
+    }
+
+    fn un_reacted(&self, _cause: EventCause, _id: usize) -> Result<()> {
+        if let EventCause::Output(_output) = _cause {
```
zhang2014, line 108:
```suggestion
        if let EventCause::Output(output) = cause {
```

src/query/pipeline/sources/src/prefetch_async_source.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        if self.is_inner_finish || self.inner.is_full(&self.generated_data) {
+            Ok(Event::NeedConsume)
+        } else {
+            Ok(Event::Async)
+        }
+    }
+
+    fn un_reacted(&self, cause: EventCause, _id: usize) -> Result<()> {
+        if let EventCause::Output(output) = cause {
```
youngsofun, line 108:
```suggestion
        if let EventCause::Output(_) = cause {
```

youngsofun:
@BohuTANG  

now cost 11m36, much faster than before (24 - 32min), but still slower than expected 9min12 (my last run with `enable_new_copy_for_text_formats=0`)

this image is right before ending, the progress is correct now (99997497 rows,  15.47G)

 
image

BohuTANG:
> now cost 11m36,
I think the cost time is ok.

Two questions:
1. Why we should do this setting: `enable_new_copy_for_text_formats=0`, does it mean the new copy impl is still need to improve?
2. This PR is helpful but is one part of the improvement(Means we can merge this PR first)?

youngsofun:
> Why we should do this setting: enable_new_copy_for_text_formats=0, does it mean the new copy impl is still need to improve?

the new impl is much better to maintain, logic is infact the same, but nearly half of the codes are rewritten, I`m not sure there are no problem.  


> This PR is helpful but is one part of the improvement(Means we can merge this PR first)?

yes, we can merge first 


---- 

it is strange that the `Duraition` in the history is diff from the time. it is longer :

image

if it is not acceptable, we can set enable_new_copy_for_text_formats=0 by default until it is improved.

BohuTANG:
> > Why we should do this setting: enable_new_copy_for_text_formats=0, does it mean the new copy impl is still need to improve?
> 
> the new impl is much better to maintain, logic is infact the same, but nearly half of the codes are rewritten, I`m not sure there are no problem.
> 
> > This PR is helpful but is one part of the improvement(Means we can merge this PR first)?
> 
> yes, we can merge first
> 
> it is strange that the `Duraition` in the history is diff from the time. it is longer :

This is the real time which cost.

> image
> 
> if it is not acceptable, we can set enable_new_copy_for_text_formats=0 by default until it is improved.

BohuTANG (approved):

PR #14986, opened 2024-03-18
feat(query): check view's inner table privilege when create or alter view

TCeason:

## Summary

Note:

Select view no need to check the inner table privilege.

Because the user only can query view when the admin user grant select on the view to the user.

So the admin user needs to know why the user need access this view.

This pr is avoid that:

User only has create or alter on db. And then the user can query table as create or alter view. Like this:

```sql
--root
grant create on default.* to a;
--user a
create view v_t1 as select * from t1; -- If it can be success, the user can query table t1 according to query view v_t1. It maybe dangerous.
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/service/src/interpreters/access/privilege_access.rs
+++ b/src/query/service/src/interpreters/access/privilege_access.rs
@@ -862,7 +862,6 @@ impl AccessChecker for PrivilegeAccess {
                 self.validate_table_access(&plan.catalog, &plan.database, &plan.table, vec![UserPrivilegeType::Update], false).await?;
             }
             Plan::CreateView(plan) => {
-                self.validate_db_access(&plan.catalog, &plan.database, vec![UserPrivilegeType::Create], false).await?;
                 let mut planner = Planner::new(self.ctx.clone());
                 let (plan, _) = planner.plan_sql(&plan.subquery).await?;
                 self.check(ctx, &plan).await?
--- a/src/query/service/src/interpreters/access/privilege_access.rs
+++ b/src/query/service/src/interpreters/access/privilege_access.rs
@@ -32,6 +32,7 @@ use databend_common_sql::optimizer::get_udf_names;
 use databend_common_sql::plans::InsertInputSource;
 use databend_common_sql::plans::PresignAction;
 use databend_common_sql::plans::RewriteKind;
+use databend_common_sql::Planner;
 use databend_common_users::RoleCacheManager;
 
 use crate::interpreters::access::AccessChecker;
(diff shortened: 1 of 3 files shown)


TCeason:
The view is also a table . So create/alter/drop view need consider ownership? cc @flaneur2020

src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -861,10 +862,16 @@ impl AccessChecker for PrivilegeAccess {
                 self.validate_table_access(&plan.catalog, &plan.database, &plan.table, vec![UserPrivilegeType::Update], false).await?;
             }
             Plan::CreateView(plan) => {
-                self.validate_db_access(&plan.catalog, &plan.database, vec![UserPrivilegeType::Create], false).await?
+                self.validate_db_access(&plan.catalog, &plan.database, vec![UserPrivilegeType::Create], false).await?;
```
sundy-li, line 865: I think View just needs Select privilege for inner table even in Create Or Alter.

TCeason: If so, we no need to add ownership cover view which will be easier . But I think this is a product form. 
So cc @flaneur2020 how about you?

If a user can query the inner table in view, it can create the view.

But I think alter and drop view need privilege check about alter and drop privilege on db level.

TCeason: > I think View just needs Select privilege for inner table even in Create Or Alter.

About alter view, if user a modify the v. After some days User b also modify the view v to adopt b's business.

It maybe destroys a.

So I think alter and drop need to check privileges.

TCeason: Modify in newest commit.

flaneur2020: found the related semantics from snowflake:

> When you create a view and then grant privileges on that view to a role, the role can use the view even if the role does not have privileges on the underlying table(s) that the view accesses. This means that you can create a view to give a role access to only a subset of a table. For example, you can create a view that accesses medical billing information but not medical diagnosis information in the same table. Then you can grant privileges on that view to the “accountant” role so that the accountants can look at the billing information without seeing the patient’s diagnosis.

it seems that if you have access to a view, even you can not access the underlying tables, you can still access the underlying table data by the view.

flaneur2020: > If a user can query the inner table in view, it can create the view.
>
> But I think alter and drop view need privilege check about alter and drop privilege on db level.

agree, if we hope using view as a "public interface" for others to access the unaccessible table data, we need carefully check the permissions when we create/alter the view, ensuring the one whom created the view, have all the SELECT permissions to the underlying tables.

TCeason: > found a the related docs from snowflake:
> 
> > When you create a view and then grant privileges on that view to a role, the role can use the view even if the role does not have privileges on the underlying table(s) that the view accesses. This means that you can create a view to give a role access to only a subset of a table. For example, you can create a view that accesses medical billing information but not medical diagnosis information in the same table. Then you can grant privileges on that view to the “accountant” role so that the accountants can look at the billing information without seeing the patient’s diagnosis.
> 
> it seems that if you have access to a view, even you can not access the underlying tables, you can still access the underlying table data by the view.

Yes. That's why we need view. 

I also think do not need all inner tables. e.g.: v_tt is view of table `tt`, v_v_tt is view of view `v_tt`

-- user : a
create table tt(id int);

insert into tt values(1);

create view v_tt as select * from tt;

-- user : a

create view v_v_tt as select * from v_tt;
ERROR 1105 (HY000): PermissionDenied. Code: 1063, Text = Permission denied: privilege [Select] is required on 'default'.'default'.'v_tt' for user 'a'@'%' with roles [public].

-- user: root grant select on view to user a
grant select on default.v_tt to a;
grant select on default.v_v_tt to a;

--user : a
-- even user can not access inner table in v_t, but user a can access view v_t, that's enough
select * from v_v_tt;
+------+
| id   |
+------+
|    1 |
+------+

flaneur2020: > I also think do not need all inner tables. e.g.: v_tt is view of table tt, v_v_tt is view of view v_tt

oh yes, we need ensure the creator can access the inner tables & the inner views of the SQL which creating the view, but we do not need to further check the permission of inner tables of the inner views.

TCeason:
Please review this pr. cc @flaneur2020 @sundy-li

sundy-li (approved):

flaneur2020 (approved):
LGTM

PR #14997, opened 2024-03-18
feat(query): Support query match inverted index

b41sh:

## Summary

* Add internal columns `SearchMatched` and `SearchScore`
* Read inverted index file with cache
* Add search function `match` and `score`
* Support query with `match` function to filter by inverted index
* Using [tantivy-jieba](https://crates.io/crates/tantivy-jieba) to support Chinese tokenizer
* Add stop word filter to remove English stop words(`a`, `an`, `the`, ...)

Inverted index location design: 

Each table data-related meta information is stored in a `snapshot` file, including `snapshot_id`, `schema`, `summary`, `segments`, and other fields. The segments record the location of a set of `SegmentInfo` files, and each `SegmentInfo` records the location of a set of `Block` files, each `Block` is a parquet-formatted file that stores a number of rows of values in table.
The inverted index adds an `index_info_locations` field to the `snapshot` file to record the location of `IndexInfo` files, which records the location of a set of `Index` files, one `Index` file corresponds to multiple `SegmentInfo` files. Through the inverted index in the Index file, we can determine which `Block` in which `Segment` matches the query keyword.
The relationship is shown in the following figure:
```
┌────────┐
│Snapshot│
└────────┘
  |   |      ┌──────┬──────┬──────┬───┬──────┐
  |   |      │Block1│Block2│Block3│...│BlockN│
  |   |      └──────┴──────┴──────┴───┴──────┘
  |   |      \          _________/
  |   |       \        /
  |   |       ┌────────┬────────┬────────┬───┬────────┐
  |   +-------│Segment1│Segment2│Segment3│...│SegmentN│
  |           └────────┴────────┴────────┴───┴────────┘
  |           \         ________________/
  |            \       /
  |            ┌──────┬──────┬──────┬───┬──────┐
  |            │Index1│Index2│Index3│...│IndexN│
  |            └──────┴──────┴──────┴───┴──────┘
  |            \           ___________________/
  |             \         /
  |             ┌─────────┐
  +-------------│IndexInfo│
                └─────────┘
```

The `Index` file treats the indexed data as `Document`, each `Document` has a `docId`, and the query returns the matched `doc_ids` with the query keyword. Since an `Index` corresponds to multiple `Blocks`, the `Blocks` are added to the `Index` in the same order as they are in the `Segment`. When querying, we can convert the `docId` to the corresponding `Block` and the row id in the block in the order it was added. As shown in the following figure:
```
┌────┬───┬────┬─────┬───┬────┬───┬───┬────┬───┬────┐
│Doc0│...│DocM│DocM1│...│DocN│...│...│DocX│...│DocY│
└────┴───┴────┴─────┴───┴────┴───┴───┴────┴───┴────┘
  |        |      \         \           \        \
  |        |       \         \           \        \
┌────┬───┬────┐     ┌────┬───┬────┐     ┌────┬───┬────┐
│Row0│...│RowN│ ... │Roc0│...│RowN│ ... │Row0│...│RowN│
└────┴───┴────┘     └────┴───┴────┘     └────┴───┴────┘
 \  Block1              BlockM   /       \  BlockN   /
  \          ___________________/         \         /
   \        /                              \       /
    Segment1                               SegmentN
```

Inverted index query process

1. When there is a `match` function in the where condition, add `InvertedInfo` to the `PushDown` in the `TableScan` physical plan, including the information of the searched field, index name, search keywords, and so on.
2. Create `InvertedIndexPruner` in `FusePruner`, read `IndexInfo` from `index_info_locations` field in the `snapshot`, use the inverted index to match query words, and convert `docId` to `rowId` in the `Block`.
3. Iterate through all the `Segments`, and determine whether each `Segment` matches the index or not, if not matched, it can be deleted directly, and all the related `Blocks` do not need to be read.
4. For the matched `Segment`, iterate through each `Block`, and determine whether it matches the index or not, if not matched, the `Block` can be deleted directly. For matched `Block`, generate matched rows and the related score.
5. Add the matched rows and scores to the `matched_rows` field in `BlockMetaIndex` of the `Block`.
6. The query result converts `matched_rows` into two internal columns, `search_matched` for filtering matched rows and `search_score` for sorting the filtered results by degree of match.

part of #14825 

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


lichuang (approved):

src/query/config/src/config.rs
```diff
@@ -2774,6 +2774,14 @@ pub struct CacheConfig {
     )]
     pub inverted_index_info_count: u64,
 
+    /// Max bytes of cached inverted index filters used. Set it to 0 to disable it.
+    #[clap(
+        long = "cache-inverted-index-filter-size",
```
BohuTANG, line 2779: Default is 17GB memory?

b41sh: It's just a temporary value, we need some tests it to determine how much memory is appropriate, we can change it to a smaller size first.

src/query/storages/fuse/src/pruning/fuse_pruner.rs
```diff
@@ -78,6 +82,7 @@ impl PruningContext {
         cluster_keys: Vec<RemoteExpr<String>>,
         bloom_index_cols: BloomIndexColumns,
         max_concurrency: usize,
+        index_info_locations: &Option<BTreeMap<String, Location>>,
```
BohuTANG, line 85:
```suggestion
        index_info_locations: &Option<BTreeMap<String, Location>>,
        max_concurrency: usize,
```

src/query/storages/fuse/src/pruning/fuse_pruner.rs
```diff
@@ -145,6 +150,22 @@ impl PruningContext {
         let internal_column_pruner =
             InternalColumnPruner::try_create(func_ctx, filter_expr.as_ref());
 
+        let inverted_index_pruner = if let Some(push_down) = push_down {
```
BohuTANG, line 153: Can we wrap this into InvertedIndexPrunner::try_create?

src/query/storages/fuse/src/operations/read_partitions.rs
```diff
@@ -91,7 +96,7 @@ impl FuseTable {
                         .await;
                 }
 
-                if (!dry_run && snapshot.segments.len() > nodes_num) || is_lazy {
+                if !is_inverted && ((!dry_run && snapshot.segments.len() > nodes_num) || is_lazy) {
```
BohuTANG, line 99: We don't support distributed read for inverted index?

BohuTANG: Another question here is:
Why here need to check `!is_inverted` but others like `aggregate index` no need?

Better to have more comments.

src/query/storages/fuse/src/pruning/fuse_pruner.rs
```diff
@@ -294,6 +310,15 @@ impl FusePruner {
 
                         async move {
                             // Build pruning tasks.
+                            if let Some(inverted_index_pruner) = &pruning_ctx.inverted_index_pruner
```
sundy-li, line 313: I think inverted index maybe called after segment pruner?

BohuTANG:
The PR summary is so great 👍

PR #15027, opened 2024-03-19
refactor: truncate support build pipeline

zhyass:

## Summary

1. truncate add table lock.
2. truncate support build pipeline. 
    OneBlockSource --------> CommitSink

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement


src/query/service/src/interpreters/interpreter_delete.rs
```diff
@@ -197,48 +198,93 @@ impl Interpreter for DeleteInterpreter {
...
         let mut build_res = PipelineBuildResult::create();
+
+        let snapshot_opt = fuse_table.read_table_snapshot().await?;
+        // check if table is empty
+        let snapshot = if let Some(val) = snapshot_opt {
+            val
+        } else {
+            // no snapshot, no deletion
+            return Ok(build_res);
+        };
```
SkyFan2002, line 209:
```suggestion
        let Some(snapshot) = fuse_table.read_table_snapshot().await? else {
            // no snapshot, no deletion
            return Ok(build_res);
        };
```

src/query/storages/fuse/src/operations/common/processors/sink_commit.rs
```diff
@@ -342,13 +349,22 @@ where F: SnapshotGenerator + Send + 'static
                 .await
                 {
                     Ok(_) => {
-                        if self.transient {
-                            // Removes historical data, if table is transient
+                        if Self::is_truncate() {
+                            catalog
+                                .truncate_table(&table_info, TruncateTableReq {
```
SkyFan2002, line 354: Maybe `catalog.truncate_table()` should not be called when in the delete fast path.

zhyass: > Maybe `catalog.truncate_table()` should not be called when in the delete fast path. Delete stmt should only delete the table data, not influence the `copied_files` in meta.

Thx, fixed

src/query/service/src/interpreters/interpreter_delete.rs
zhyass: Thx, get a new technique!

SkyFan2002 (approved):

PR #15045, opened 2024-03-20
feat: optimize subquery

### LINKED ISSUE #14482: Unify execution of DML statements and queries
In the current implementation, the execution of DMLs contain queries are quite different from common queries.

Taking DELETE statement as an example. If there is no subquery in the WHERE clause(e.g. `delete from t where t.a > 1`), the whole DELETE statement will be executed within a single operator, which will scan the data from the target table and filter them by evaluating the filter expression.

But when there are subqueries, it will become quite complicated(e.g. `delete from t where exists(select * from t1 where t.id = t1.id)`).

In PostgreSQL, the DELETE work with `ctid`(aka row id) produced by the query part:
```
leiysky=# explain(verbose) delete from t where exists(select * from t1 where t.a = t1.a);
                                      QUERY PLAN
--------------------------------------------------------------------------------------
 Delete on public.t  (cost=46.38..102.75 rows=0 width=0)
   ->  Hash Join  (cost=46.38..102.75 rows=1275 width=12)
         Output: t.ctid, t1.ctid
         Inner Unique: true
         Hash Cond: (t.a = t1.a)
         ->  Seq Scan on public.t  (cost=0.00..35.50 rows=2550 width=10)
               Output: t.ctid, t.a
         ->  Hash  (cost=43.88..43.88 rows=200 width=10)
               Output: t1.ctid, t1.a
               ->  HashAggregate  (cost=41.88..43.88 rows=200 width=10)
                     Output: t1.ctid, t1.a
                     Group Key: t1.a
                     ->  Seq Scan on public.t1  (cost=0.00..35.50 rows=2550 width=10)
                           Output: t1.ctid, t1.a
(14 rows)
```

This is extensible, and we have supported the `row_id` mechanism as well. But it will make us lose a big chance to optimize the performance when a physical data block is completely dropped.

We should find a better way to work with both `row_id`-based data modification and block-based data modification.

lichuang:

## Summary

feat: optimize subquery

- Fixes #14482

in previous update/delete by subquery implementation:
1. first evaluate subqueries and get `row_id` arrays as `ScalarExpr`;
2. use `ScalarExpr` return by first step 1 as conditions to modify data from source.

after optimization of subquery:
1. all subqueries will be evaluate with modify physical plan in on runtime;
2. subquery result will join with the target source data, for example: if the the original target source data is `[1,2,3,4]`, it will return `[1,2,3,4]` plus a marker column like `[true, false, true, false]`, in which the marker column will mark if or not the row is filter by subquery
3. add a new transformer `TransformMutationSubquery` to modify data, which will:
  1. delete: it will filter out datas while mark as `false`
  2. update: it will evaluate operations on data
4. the modified data will by commit on storage. 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [x] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15045)


zhang2014:
LGTM.

Issue #14929, opened 2024-03-13
bug: [1002=>status: OutOfRange] querying the system.task_history table failed


dantengsky:
**Summary**

query  table `system.task_history` failed with error code 1002.

vesion: Databend Query v1.2.373-nightly-683a3a22a7(rust-1.77.0-nightly-2024-03-12T08:05:08.707427719Z)

~~~sql
select * from system.task_history;
~~~

~~~
1002=>status: OutOfRange, message: "Error, message length too large: found 4916789 bytes, the limit is: 4194304 bytes", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
~~~

PR #15047, opened 2024-03-20
feat: support pagination on task_history

ZhiHanZ:

## Summary

Previously when single request return body is large, it would show [grpc error](https://github.com/datafuselabs/databend/issues/14929), this Pull Request support the following things:
1. `task_name`, `schedule_time` filter push down, allow to use those select information to filter unneeded rows on cloud control services.
2. add `next_page_token` and `page_size` parameter, allow to control page size and pagination of the single request.

`NOTICE`: I also tried to use limit push_down, but found that it does NOT work, need some deep dive on the issue

fix https://github.com/datafuselabs/databend/issues/14929
## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


BohuTANG:
> NOTICE: I also tried to use limit push_down, but found that it does NOT work, need some deep dive on the issue

We should make it works:) Think a user has many tasks all scheduler is in 1 second, the rows is very large, and run:
`select * from system.task_history` is no sense anymore.

So, here the better way is:
1. `select * from system.task_history`  returns rows with a LIMIT(Say 10000)
2. `select * from system.task_history limit N` returns N rows order by the time.

ZhiHanZ:
> > NOTICE: I also tried to use limit push_down, but found that it does NOT work, need some deep dive on the issue
> 
> We should make it works:) Think a user has many tasks all scheduler is in 1 second, the rows is very large, and run: `select * from system.task_history` is no sense anymore.
> 
> So, here the better way is:
> 
> 1. `select * from system.task_history`  returns rows with a LIMIT(Say 10000)
> 2. `select * from system.task_history limit N` returns N rows order by the time.

first limitation is easy to limit on server side, but for the second part, we need the push_down_info struct to pass limit number to the storage table source, but for now it does not work

BohuTANG:
> > > NOTICE: I also tried to use limit push_down, but found that it does NOT work, need some deep dive on the issue
> > 
> > 
> > We should make it works:) Think a user has many tasks all scheduler is in 1 second, the rows is very large, and run: `select * from system.task_history` is no sense anymore.
> > So, here the better way is:
> > 
> > 1. `select * from system.task_history`  returns rows with a LIMIT(Say 10000)
> > 2. `select * from system.task_history limit N` returns N rows order by the time.
> 
> first limitation is easy to limit on server side, but for the second part

This limitation is ok if we give a default number for it.

hantmac (approved):
LGTM 🚀

PR #15048, opened 2024-03-20
chore: add merge_into_resize_parallel_threads settings

JackTan25:

## Summary
if the machine has too many cores, maybe we will write too many small block, which cause high i/o cost.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/settings/src/settings_default.rs
+++ b/src/query/settings/src/settings_default.rs
@@ -498,6 +498,12 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: Some(SettingRange::Numeric(0..=1)),
                 }),
+                ("merge_into_resize_parallel_threads", DefaultSettingValue {
+                    value: UserSettingValue::UInt64(0),
+                    desc: "tune the source parallel when too many small blocks",
+                    mode: SettingMode::Both,
+                    range: Some(SettingRange::Numeric(0..=u64::MAX)),
+                }),
                 ("enable_distributed_merge_into", DefaultSettingValue {
                     value: UserSettingValue::UInt64(0),
                     desc: "Enables distributed execution for 'MERGE INTO'.",
--- a/src/query/service/src/pipelines/builders/builder_merge_into.rs
+++ b/src/query/service/src/pipelines/builders/builder_merge_into.rs
@@ -342,8 +342,11 @@ impl PipelineBuilder {
         } = merge_into_source;
 
         self.build_pipeline(input)?;
-        self.main_pipeline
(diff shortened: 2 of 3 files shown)


src/query/settings/src/settings_default.rs
```diff
@@ -498,6 +498,12 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: Some(SettingRange::Numeric(0..=1)),
                 }),
+                ("merge_into_resize_parallel_threads", DefaultSettingValue {
```
BohuTANG, line 501: I think this value should be adjust based on CPUs not a setting, we actully don't know what's the value to set.

BohuTANG: BTW, we have changed the max I/O request limit multiple times, returning to the simplest method in the end:
https://github.com/datafuselabs/databend/blob/4a55ea429a148a11ca365983976abc63363e4ae5/src/query/storages/fuse/src/operations/read_data.rs#L82-L92

JackTan25: https://github.com/datafuselabs/databend/pull/15048/files#diff-31afd04d842c6c44f5b86a534efe227d1bd19c04b8dbad24e4cfdae42f33300dR448 see here, the default is cpu threads, and we won't change the previous query time, it will change the merge into i/o pipeline. I need to modify the merge into to make it more tunable. It's to control the block write threads.

JackTan25: the i/o reuqust is used to control the number of i/o at the same time, but it can't control the whole number of complete pipeline, we need to control the block compact and reduce it.

BohuTANG: I see. Is this setting only for our testing not the user?

JackTan25: This pr need to check, But in default, it won't make effect, I and @wubx find out the i/o is very high. So we need this. Let me do a cloud test with him.

dantengsky:
Thanks for the contribution and for looking into the small-block write issue in MERGE INTO. Since this PR was opened, the MERGE INTO execution pipeline has been substantially refactored. The resize point targeted by this change no longer exists, and the current pipeline handles block compaction differently. As a result, this change is no longer applicable in its current form, so I'm closing the PR.
If excessive small-block writes are still reproducible with the current implementation, we'd be happy to revisit the issue with a new profile or reproduction. Thanks again!

PR #15075, opened 2024-03-23
fix: privilege check for insert statement

SkyFan2002:

## Summary

To use the OVERWRITE option on INSERT, you must use a role that has DELETE privilege on the table because OVERWRITE will delete the existing records in the table.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/service/src/interpreters/access/privilege_access.rs
+++ b/src/query/service/src/interpreters/access/privilege_access.rs
@@ -747,7 +747,12 @@ impl AccessChecker for PrivilegeAccess {
             }
             // Others.
             Plan::Insert(plan) => {
-                self.validate_table_access(&plan.catalog, &plan.database, &plan.table, vec![UserPrivilegeType::Insert], false).await?;
+                let target_table_privileges = if plan.is_overwrite {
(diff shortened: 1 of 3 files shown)


BohuTANG:
Good to have a SQL logic test to check the new privilege, and to protect someone break this in the future.

src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -747,7 +747,12 @@ impl AccessChecker for PrivilegeAccess {
             }
             // Others.
             Plan::Insert(plan) => {
-                self.validate_table_access(&plan.catalog, &plan.database, &plan.table, vec![UserPrivilegeType::Insert], false).await?;
+                let target_table_privileges = if plan.is_overwrite {
+                    vec![UserPrivilegeType::Insert, UserPrivilegeType::Delete]
```
TCeason, line 751: Need a test. Ref 18_rbac in stateless

TCeason (approved):
LGTM

Issue #15082, opened 2024-03-24
 copy into use specified file list  is very slow 


wubx:
**Summary**

copy into T from @stage files=(.....) file_format=(type=ndjson);

![img_v3_0299_3988773f-ae3b-4bd0-9190-b52544091c7g](https://github.com/datafuselabs/databend/assets/320680/6fad1567-e645-4d6f-a192-92f26d99c700)

PR #15083, opened 2024-03-25
feat: Implement concurrent stat

Xuanwo:

## Summary

Implement concurrent stat to speedup the list files.

fix https://github.com/datafuselabs/databend/issues/15082

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/common/storage/src/stage.rs
+++ b/src/common/storage/src/stage.rs
@@ -115,28 +115,20 @@ impl StageFilesInfo {
 
         let max_files = max_files.unwrap_or(usize::MAX);
         if let Some(files) = &self.files {
-            let mut res = Vec::new();
-            let mut limit: usize = 0;
-            for file in files {
-                let full_path = Path::new(&self.path)
-                    .join(file)
-                    .to_string_lossy()
-                    .trim_start_matches('/')
-                    .to_string();
-                let meta = operator.stat(&full_path).await?;
-                if meta.mode().is_file() {
-                    res.push(StageFileInfo::new(full_path, &meta))
-                } else {
-                    return Err(ErrorCode::BadArguments(format!(
-                        "{full_path} is not a file"
-                    )));
-                }
-                if first_only {
-                    return Ok(res);
-                }
-                limit += 1;
-                if limit == max_files {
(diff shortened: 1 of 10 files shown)


src/common/storage/src/stage.rs
```diff
@@ -247,6 +239,41 @@ impl StageFilesInfo {
...
+    pub async fn stat_concurrent(
+        &self,
+        operator: &Operator,
+        first_only: bool,
+        max_files: usize,
+        files: &Vec<String>,
+    ) -> Vec<Result<(String, Metadata)>> {
+        let mut futures = Vec::with_capacity(files.len());
+        let mut limit = 0;
+        for file in files {
```
BohuTANG, line 253: It's better to use a batch; we parallelize the batch. I'm wondering if there are too many files, will there be too many futures?
See: https://github.com/datafuselabs/databend/blob/3a50f2c918b527ed140c4f83e36a0aac01d5b31c/src/query/storages/fuse/src/io/files.rs#L63-L77

Xuanwo: Makes sense, let me change them into an iterator.

src/common/storage/src/stage.rs
```diff
@@ -247,6 +240,46 @@ impl StageFilesInfo {
...
+                .to_string();
+            let operator = operator.clone();
+            async move {
+                let meta = operator.stat(&full_path).await?;
+                Ok((full_path, meta))
+            }
+        });
+
+        // We don't have access to ctx in current context, use 100 as a safe default value.
+        execute_futures_in_parallel(tasks, 100, 100 * 2, "batch-stat-file-worker".to_owned()).await
```
BohuTANG, line 281: 100 is too large, It's better to set it to thread_num, please reference:
```
threads_nums, 
     threads_nums * 2, 
```

youngsofun (approved):

Xuanwo:
Done, cc @BohuTANG PTAL

PR #15093, opened 2024-03-26
chore(query): refactor new transform partition bucket for new aggregation hashtable

Freejww:

## Summary
In the previous https://github.com/datafuselabs/databend/pull/14798, we initialized all data at once for the new aggregation hashtable. This approach may have caused a performance decrease, particularly on queries such as hits-q18 (referred to as "short-circuit query").

In this PR, we introduced `new_transform_partition_bucket` for the new aggregation hashtable. We utilize a working bucket similar to the old transform_partition_bucket to handle bucket incrementally.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement


src/query/service/src/pipelines/processors/transforms/aggregator/aggregate_exchange_injector.rs
```diff
@@ -67,9 +67,9 @@ struct AggregateExchangeSorting<Method: HashMethodBounds, V: Send + Sync + 'stat
 impl<Method: HashMethodBounds, V: Send + Sync + 'static> ExchangeSorting
     for AggregateExchangeSorting<Method, V>
 {
-    fn block_number(&self, data_block: &DataBlock) -> Result<isize> {
```
sundy-li, line 70: we can generate a sorted block_number by `max_partition_count` and `bucket` number.

eg:

```
block_number = max_partition_count * 10000 + bucket;
```

sundy-li: Thus we don't need to change the aggregate meta.

src/query/service/src/api/rpc/exchange/exchange_transform_shuffle.rs
```diff
@@ -89,6 +89,7 @@ struct OutputsBuffer {
 
 impl OutputsBuffer {
     pub fn create(capacity: usize, outputs: usize) -> OutputsBuffer {
+        let capacity = capacity * 1000;
```
sundy-li, line 92: this is dangerous, we still need to find out the key reason.

sundy-li (approved):

PR #15138, opened 2024-03-31
fix(query): show query with name consider quote

TCeason:

## Summary

In Meta, table/field name is groundtruth.

So in show query, we need to condsider the quota setting quoted_ident_case_sensitive.

If the name is not lowercase or contains special character, use "" quote.

e.g.

```sql
create table tc(id int, Id1_ int, "id2)" int, "ID4" int);

show create table tc;
+-------+---------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                            |
+-------+---------------------------------------------------------------------------------------------------------+
| tc    | CREATE TABLE tc (
  id INT NULL,
  "id1_" INT NULL,
  "id2)" INT NULL,
  "ID4" INT NULL
) ENGINE=FUSE |
+-------+---------------------------------------------------------------------------------------------------------+
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


src/query/service/src/interpreters/util.rs
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub fn format_name(name: &str, quoted_ident_case_sensitive: bool) -> String{
+    // Db-s -> "Db-s" ; dbs -> dbs
+    if name.chars().find(|c| !(c.is_lowercase() || c.is_numeric() || *c == '_') ).is_some() && quoted_ident_case_sensitive {
+        format!("\"{name}\"")
```
andylokandy, line 18: Not every dialect uses `"` as ident qoute

TCeason: MySQL use `, other use "?

sundy-li: most dialect support `"` , but only mysql supports   `

andylokandy: PostgreSQL uses \` while mysql uses `"`

andylokandy: I suggest reading the quote from dialect and carefully escape the name by the selected quote.

TCeason: > I suggest reading the quote from dialect and carefully escape the name by the selected quote.
> 
> FYI, there is a `quote_ident()` function.

in interpreter show create table , we can not get the string quote. Only the name.

src/query/service/src/interpreters/util.rs
```diff
@@ -0,0 +1,32 @@
...
+
+pub fn format_name(name: &str, quoted_ident_case_sensitive: bool, dialect: Dialect) -> String {
+    // Db-s -> "Db-s" ; dbs -> dbs
+    if name
+        .chars()
+        .find(|c| !(c.is_lowercase() || c.is_numeric() || *c == '_'))
+        .is_some()
+        && quoted_ident_case_sensitive
+    {
+        match dialect {
```
andylokandy, line 25: You can add:
```
impl Dialect {
    pub fn default_ident_quote(&self) -> char {
        match self {
            Dialect::MySQL | Dialect::Hive => '`',
            Dialect::Experimental | Dialect::PostgreSQL | Dialect::PRQL => '"',
        }
    }
}
```

src/query/service/src/interpreters/util.rs
andylokandy: > we can not get the string quote

We can select the default quote used by the dialect.

sundy-li (approved):

tests/suites/0_stateless/20+_others/20_0001_show.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
andylokandy, line 1: why not use sqllogictest?

tests/suites/0_stateless/20+_others/20_0001_show.sh
```diff
@@ -0,0 +1,13 @@
...
+
+CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+. "$CURDIR"/../../../shell_env.sh
+
+
+
+stmt "drop table if exists default.tc;"
+stmt 'create table default.tc(id int, Id1_ int, "id2)" int, "ID4" int);'
```
andylokandy, line 9: consider adding a name with ", for example "ID""5"

tests/suites/0_stateless/20+_others/20_0001_show.sh
TCeason: Try to use sqllogictest but it can not success. the result is ok, but databend-sqllogictest can not compare the diff.

tests/suites/0_stateless/20+_others/20_0001_show.sh
TCeason: Done

tests/suites/0_stateless/20+_others/20_0001_show.sh
andylokandy: >  but databend-sqllogictest can not compare the diff.

same in my other PR. seems like a bug in sqllogictest. Anyway, we should not add new test in the legacy test framework...

TCeason: Ok, I got it. https://docs.rs/sqllogictest/0.17.2/sqllogictest/runner/fn.default_validator.html function normalize_string trim \n as " "

TCeason:
cc @andylokandy  Please review this pr. Now all ci passed.

andylokandy (approved):
Nice work!

TCeason:
Please merge this pr. @BohuTANG

PR #15145, opened 2024-04-01
chore(query): fix order for scalar cmp with null

sundy-li:

## Summary

By default, null is biggest than other values in databend. So we enable Scalar::Null compare with other none null scalars.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement


src/query/expression/src/values.rs
```diff
@@ -770,6 +770,10 @@ impl PartialOrd for ScalarRef<'_> {
             (ScalarRef::Tuple(t1), ScalarRef::Tuple(t2)) => t1.partial_cmp(t2),
             (ScalarRef::Variant(v1), ScalarRef::Variant(v2)) => jsonb::compare(v1, v2).ok(),
             (ScalarRef::Geometry(g1), ScalarRef::Geometry(g2)) => compare_geometry(g1, g2),
+
+            // By default, null is biggest in pgsql
```
xudong963, line 774: If dialect isn't default(`Dialect::PostgreSQL`), should we change the behavior?

sundy-li: It’s hard to control by settings.

sundy-li: In order by clause, we can control the order by settings.

Here we add this rule to make min/max aggregation return stable results.

sundy-li:
Bug in main for merge join, because it did not consider about null
```
drop table if exists t1;

drop table if exists t2;

create table t1(a int null);

insert into t1 values(1), (NULL), (2);

create table t2(b int null);

insert into t2 values(1), (NULL), (3);

select * from t1 join t2 on t1.a <= t2.b order by t1.a;
```

xudong963 (approved):

PR #15146, opened 2024-04-01
chore: client should not depend on the final response.

youngsofun:

## Summary

final is not ACKed by client, so client should not depend on the final response,

 for server:
1. when response with `next_uri` =  `.../final`, all states should already be delivered to the client the latest in this response.
2. `/final` SHOULD response with nothing but `next_uri` = `null`, BUT to tolerant old clients, we still keep some fields.

for clients:

check `next_uri` before referring to other fields of the response.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/service/src/servers/http/v1/http_query_handlers.rs
+++ b/src/query/service/src/servers/http/v1/http_query_handlers.rs
@@ -207,6 +207,17 @@ impl QueryResponse {
     }
 }
 
+/// final is not ACKed by client, so client should not depend on the final response,
+///
+/// for server:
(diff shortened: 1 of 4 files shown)


sundy-li (approved):

BohuTANG:
@youngsofun After this PR, we should update the doc https://docs.databend.com/developer/apis/http, remove the final parts?

youngsofun:
> @youngsofun After this PR, we should update the doc https://docs.databend.com/developer/apis/http, remove the final parts?

yes,  `/final` is only called though `next_uri` in clients.

PR #15154, opened 2024-04-02
feat(rbac): drop role will transfer role owns object to account_admin

TCeason:

## Summary

drop role will transfer role owns object to account_admin

E.g.

```sql
create role role1;
create user u1 identified by '123' with default_role='role1';
grant role role1 to u1;

```

use u1 create some database then, drop the role1. In ownership-kv, the object keys value is a dropped role role1.

If admin user wants to create a new role also named role1, the new role `role1` will directly has objects ownership. That maybe dangerous. 

So in this pr, add a check: dorp role will check the dropped role owns some object. If true, transfer the owner to account_admin.

### Note

Add forward/backward compat test.

drop role can execute in old version(1.0.307) and new version.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15154)

### DIFF
--- a/src/query/management/src/role/role_api.rs
+++ b/src/query/management/src/role/role_api.rs
@@ -44,7 +44,7 @@ pub trait RoleApi: Sync + Send {
     /// If a role is dropped, but the owner object is exists,
     ///
     /// The owner role need to update to account_admin.
-    async fn transfer_ownership_to_admin(&self, object: &OwnershipObject) -> Result<()>;
+    async fn transfer_ownership_to_admin(&self, object: &Vec<OwnershipObject>) -> Result<()>;
 
     /// Grant ownership would transfer ownership of a object from one role to another role
     ///
--- a/src/query/management/src/role/role_api.rs
+++ b/src/query/management/src/role/role_api.rs
@@ -44,7 +44,7 @@ pub trait RoleApi: Sync + Send {
     /// If a role is dropped, but the owner object is exists,
     ///
     /// The owner role need to update to account_admin.
-    async fn transfer_ownership_to_admin(&self, object: &[OwnershipObject]) -> Result<()>;
+    async fn transfer_ownership_to_admin(&self, object: &[(u64, OwnershipObject)]) -> Result<()>;
 
     /// Grant ownership would transfer ownership of a object from one role to another role
     ///
--- a/src/query/management/src/role/role_api.rs
+++ b/src/query/management/src/role/role_api.rs
@@ -44,7 +44,7 @@ pub trait RoleApi: Sync + Send {
     /// If a role is dropped, but the owner object is exists,
     ///
     /// The owner role need to update to account_admin.
-    async fn transfer_ownership_to_admin(&self, object: &[(u64, OwnershipObject)]) -> Result<()>;
+    async fn transfer_ownership_to_admin(&self, role: &str) -> Result<()>;
 
     /// Grant ownership would transfer ownership of a object from one role to another role
     ///
--- a/src/query/management/src/role/role_api.rs
+++ b/src/query/management/src/role/role_api.rs
@@ -39,6 +39,13 @@ pub trait RoleApi: Sync + Send {
     async fn update_role_with<F>(&self, role: &String, seq: MatchSeq, f: F) -> Result<Option<u64>>
     where F: FnOnce(&mut RoleInfo) + Send;
 
+    /// Only drop role will call transfer.
+    ///
+    /// If a role is dropped, but the owner object is exists,
+    ///
+    /// The owner role need to update to account_admin.
+    async fn transfer_ownership_to_admin(&self, object: &OwnershipObject) -> Result<()>;
+
     /// Grant ownership would transfer ownership of a object from one role to another role
     ///
     ///
--- a/src/query/management/src/role/role_api.rs
+++ b/src/query/management/src/role/role_api.rs
@@ -44,7 +44,7 @@ pub trait RoleApi: Sync + Send {
     /// If a role is dropped, but the owner object is exists,
     ///
     /// The owner role need to update to account_admin.
-    async fn transfer_ownership_to_admin(&self, object: &Vec<OwnershipObject>) -> Result<()>;
+    async fn transfer_ownership_to_admin(&self, object: &[OwnershipObject]) -> Result<()>;
 
     /// Grant ownership would transfer ownership of a object from one role to another role
     ///
--- a/src/query/management/src/role/role_mgr.rs
+++ b/src/query/management/src/role/role_mgr.rs
@@ -247,6 +247,43 @@ impl RoleApi for RoleMgr {
         Ok(Some(seq))
     }
 
+    /// Only drop role will call transfer.
+    ///
+    /// If a role is dropped, but the owner object is exists,
+    ///
+    /// The owner role need to update to account_admin.
+    #[async_backtrace::framed]
+    #[minitrace::trace]
+    async fn transfer_ownership_to_admin(
+        &self,
+        object: &OwnershipObject,
+    ) -> databend_common_exception::Result<()> {
+        // Ensure accurate matching of a key
+        let match_seq = MatchSeq::GE(1);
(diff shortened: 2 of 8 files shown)


src/query/users/src/role_mgr.rs
```diff
@@ -268,9 +268,19 @@ impl UserApiProvider {
...
+    // If the dropped role owns objects, transfer objects owner to account_admin role.
     #[async_backtrace::framed]
     pub async fn drop_role(&self, tenant: &Tenant, role: String, if_exists: bool) -> Result<()> {
         let client = self.role_api(tenant);
+        // get_ownerships use prefix_list_kv that will generate once meta call
+        let ownerships = self.get_ownerships(tenant).await?;
+        for (k, v) in ownerships {
+            if v == role {
+                // If role own n objects, will generate n meta call.
+                client.transfer_ownership_to_admin(&k).await?;
```
TCeason, line 280: I'm not sure , can we do this in once? If a role owns 10 thound objects, drop role will access meta 10 thound time. cc @drmingdrmer

drmingdrmer (approved):

src/query/management/src/role/role_mgr.rs
```diff
@@ -247,6 +247,43 @@ impl RoleApi for RoleMgr {
...
+
+        let kv_api = self.kv_api.clone();
+        kv_api
+            .upsert_kv(UpsertKVReq::new(
+                &key,
+                match_seq,
+                Operation::Update(value),
+                None,
+            ))
+            .await?;
```
drmingdrmer, line 282: Upsert returns two error: MetaError for service errorr, or it just failed to update any record if `match_seq` mismatches. 

Is it necessary to return an explicit error if nothing is updated? I.e., check the result `upsert_kv()` returns

src/query/users/src/role_mgr.rs
drmingdrmer: did you mean do it all in a batch? you can upsert more than one record in a trascation. upsert updates only one record

src/query/management/src/role/role_mgr.rs
TCeason: I think no need to check the err info. This transfer will only call when role be dropped and it owns object. If err we just need return err. And will not according to err kind do other logic. cc @flaneur2020

src/query/users/src/role_mgr.rs
TCeason: You mean I can build a big vec in TxnRequest.if_then?

src/query/management/src/role/role_mgr.rs
drmingdrmer: The caller may choose to ignore the error, but for this function itself, it should be explicit and clear about what happened.

flaneur2020: imho we can append a context infomation in the error here, like "(while transfer_ownership_to_admin)"

src/query/users/src/role_mgr.rs
flaneur2020: i think it'd be better to move this logic into the DropRoleInterpreter, so we can keep role manager a low leverl CRUD accessor, and put the logics upwards.

src/query/management/src/role/role_api.rs
```diff
@@ -44,7 +44,7 @@ pub trait RoleApi: Sync + Send {
     /// If a role is dropped, but the owner object is exists,
     ///
     /// The owner role need to update to account_admin.
-    async fn transfer_ownership_to_admin(&self, object: &OwnershipObject) -> Result<()>;
+    async fn transfer_ownership_to_admin(&self, object: &Vec<OwnershipObject>) -> Result<()>;
```
drmingdrmer: `objects: &[OwnershipObject]` would be better

src/query/management/src/role/role_mgr.rs
```diff
@@ -256,32 +258,43 @@ impl RoleApi for RoleMgr {
...
+            let owner_key = self.ownership_object_key(object);
+            let owner_value = serialize_struct(
+                &OwnershipInfo {
+                    object: object.clone(),
+                    role: "account_admin".to_string(),
+                },
+                ErrorCode::IllegalUserInfoFormat,
+                || "",
+            )?;
+            if_then.push(txn_op_put(&owner_key, owner_value))
```
drmingdrmer, line 273: Do not you need to add `txn_cond(&key, GE, 1)` for every owner_key?

In the previous version of this PR, you were updating with `MatchSeq::GE(1)`.

src/query/management/src/role/role_mgr.rs
```diff
@@ -256,32 +258,43 @@ impl RoleApi for RoleMgr {
...
-            ))
-            .await?;
+        let txn_req = TxnRequest {
+            condition: vec![],
+            if_then: if_then.clone(),
+            else_then: vec![],
+        };
 
-        Ok(())
+        while retry < TXN_MAX_RETRY_TIMES {
```
drmingdrmer, line 284: Use `txn_backoff()` to implement the while-loop:
https://github.com/datafuselabs/databend/blob/dac261a319f80dc37ef3ace7cde21428ef21c5a6/src/meta/api/src/schema_api_impl.rs#L1568

src/query/management/src/role/role_api.rs
TCeason: Yes new set commit fix it

src/query/management/src/role/role_mgr.rs
TCeason: My initial appeal was just to change the owner of the object owned by the deleted role to object. Now that these objects are fetched in the outer for loop, I think maybe we don't need to match the seq?

drmingdrmer: You always need to match the seq otherwise you can not be sure you are removing the key you want to remove.

TCeason: Ok I got you. That means in meta kv the seq is the real key?

drmingdrmer: Yes, `(key, seq)` is considered the globally unique key for a record

src/query/management/src/role/role_mgr.rs
```diff
@@ -247,6 +249,55 @@ impl RoleApi for RoleMgr {
...
+                ErrorCode::IllegalUserInfoFormat,
+                || "",
+            )?;
+            condition.push(txn_cond_seq(&owner_key, Eq, *seq));
+            if_then.push(txn_op_put(&owner_key, owner_value))
+        }
+        let mut trials = txn_backoff(None, func_name!());
+        loop {
+            trials.next().unwrap()?.await;
+            let txn_req = TxnRequest {
```
drmingdrmer, line 282: You did not re-fetch the `(seq, object)` thus this CAS loop has no effect at all.

You must re-fetch the latest `(seq, object)` in each loop.

drmingdrmer (approved):

TCeason:
cc @BohuTANG Already add compat test. will test drop role in old and new version. And I test the drop role in cloud dev.

This is the process:

tc-test1 is old warehouse,
tc-test2 is this pr's version warehouse.


```sql

➜  ~👆 export BENDSQL_DSN="***&warehouse=tc-test1"
➜  ~👆 bendsql

> create role role1;
Processed in (0.738 sec)
> grant ownership on default.t to role role1;

➜  ~👆 export BENDSQL_DSN="***&warehouse=tc-test2"
➜  ~👆 bendsql

> drop role role1;
Processed in (0.732 sec)

-- the table t transfer to role account_admin
> select name, owner from system.tables where name='t' and database='default';
┌───────────────────────────┐
│  name  │       owner      │
│ String │ Nullable(String) │
├────────┼──────────────────┤
│ t      │ account_admin    │
└───────────────────────────┘
```

PR #15155, opened 2024-04-02
chore(query): turn on new agg hashtable

Freejww:

## Summary

In the previous PR, we implemented a new aggregation hash table. Now, it supports both singleton and cluster environments, and we have also added support for spill. This PR includes performance tests and try to enable the new aggregation hash table.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

### DIFF
--- a/src/query/expression/src/aggregate/aggregate_hashtable.rs
+++ b/src/query/expression/src/aggregate/aggregate_hashtable.rs
@@ -220,9 +220,11 @@ impl AggregateHashTable {
             }
         }
 
-        if self.config.partial_agg && self.capacity >= self.config.max_partial_capacity {
+        if self.config.partial_agg {
             // check size
-            if self.count + BATCH_ADD_SIZE > self.resize_threshold() {
+            if self.count + BATCH_ADD_SIZE > self.resize_threshold()
+                && self.capacity >= self.config.max_partial_capacity
+            {
                 self.clear_ht();
                 self.reset_count();
             }
--- a/src/query/settings/src/settings_default.rs
+++ b/src/query/settings/src/settings_default.rs
@@ -643,7 +643,7 @@ impl DefaultSettings {
                     range: Some(SettingRange::Numeric(0..=1)),
                 }),
                 ("enable_experimental_aggregate_hashtable", DefaultSettingValue {
-                    value: UserSettingValue::UInt64(0),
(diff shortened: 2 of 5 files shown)


Freejww:
# Performance test on ClickBench/hits

### Note: We use `hits` as a benchmark because there are many group by tests.

### Local machine configuration: 32GB memory, 12-core CPU.
1. The new aggregation hash table has completely outperformed the old method.
2. In some high-cardinality or string grouping , the new method outperforms the old one.
3. More details will be introduced in a later blog.
hits-perf

sundy-li:
There are explain tests that need to be fixed.

sundy-li:
We use random and fixed SQL queries to cover the correctness tests.

Test Codes in https://github.com/sundy-li/mmbend/tree/master/examples

BohuTANG:
Great 👍 

Wizard [SELECTS] tests **passed**. SQL selects script here: https://github.com/datafuselabs/wizard/blob/main/checksb/sql/selects/check.sql

sundy-li (approved):

compasses:
May I know the test carried on memory limited ?

> # Performance test on ClickBench/hits

sundy-li:
> May I know the test carried on memory limited ?

No major difference when blocks are spilled, because the bottleneck is io issue.

compasses:
got thanks
So I think this kind of implementation may not efficient than general-purpose hash table like Clickhouse did.

sundy-li:
> So I think this kind of implementation may not efficient than general-purpose hash table like Clickhouse did.

I disagree with that. The older one is the general-purpose hashtable you indicate.

Freejww:
> Oh, OK, I mean without IO effect. Since most of agg workloads are memory-bound. And the way of implementation above seems need indirection access? Or maybe I missed something, but memory bandwidth is important for high cardinality keys.

Indirect access is that we only compare the high 16 bits of the hash. If they match, then we proceed to compare the actual data. General-purpose hashtable compares the entire hash value before proceeding to compare the actual data.

compasses:
OK got you thanks. Since the general-purpose hashtable will compare hash values if they saved the hash, if not just use it for probing, but any way the memory layout is more cache-friendly.

PR #15156, opened 2024-04-02
feat(query): implement ST_GEOMFROMGEOHASH

kkk25641463:

## Summary
feat:  Implement  st_geompointfromgeohash()
Currently the precision only implement within [12 digits](https://github.com/georust/geohash/issues/42).  And I found that ST_GEOMFROMGEOHASH() returned value is [geojson type](https://docs.snowflake.com/en/sql-reference/functions/st_geomfromgeohash).   
I'm not sure whether we should implement subtypes with geometry type like number type or decimal type? And [output](https://github.com/datafuselabs/databend/blob/f5478ff8f0e00acc1cb340e31301a77b2dedf9b7/src/query/expression/src/utils/display.rs#L256) can be choose with subtypes like ewkb、ewkt、geojson, otherwise i have to check types by alot of match case in display logic.
```
            ScalarRef::Geometry(s) => {
                let ewkb = Ewkb(s.to_vec());
                let value = match ewkb.to_ewkt(ewkb.srid()) {
                    Ok(ewkt) => ewkt,
                    Err(_) => GeoJson(std::str::from_utf8(s).unwrap()).to_json().unwrap(),
                };
                write!(f, "'{value:?}'")
```

see alse [try_to_ geometry](https://docs.snowflake.com/en/sql-reference/functions/try_to_geometry).

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


kkk25641463:
@ariesdevil help me to check plz.

sundy-li:
> Currently the precision only implement within 12 digits

Can we add another register  `ST_GEOMFROMGEOHASH( <geohash> [, <precision> ] )` in this pr ?

sundy-li:
> I'm not sure whether we should implement subtypes with geometry type like number type or decimal type?

That's a little complicated, we do not decide to address that.

sundy-li (approved):

kkk25641463:
> > Currently the precision only implement within 12 digits
> 
> Can we add another register `ST_GEOMFROMGEOHASH( <geohash> [, <precision> ] )` in this pr ?
> 
> seems it's limited by geohash crate.

The pr depends on  geohash algorithm  which only support  less than 12 precision and  is [not maintenance](https://github.com/georust/geohash/issues/42) by maintainer. Algorithms that support arbitrary precision require rewriting most of code in  geohash decode_bbox() function which are  break changes. And 12 bit accuracy can meet most geo application needs. I would like to submit an available geohash function version first.

kkk25641463:
> > I'm not sure whether we should implement subtypes with geometry type like number type or decimal type?
> 
> That's a little complicated, we do not decide to address that.

Is this the reason of the work to maintain the geozero library?

sundy-li:
> geozIs this the reason of the work to maintain the geozero library?ero

Part of, another consideration is that we don't want to introduce too much datatypes in `DataType`

PR #15177, opened 2024-04-07
feat: time travel support at offset

zhyass:

## Summary

```sql
SELECT ...
FROM ...
  {
   AT( { TIMESTAMP => <timestamp> | OFFSET => <time_difference> | SNAPSHOT => '<sid>' | STREAM => '<name>' } )
  }
[ ... ]
```
**OFFSET => <time_difference>**
This parameter defines the time offset for Time Travel, specifying the number of seconds to go back from the current time. It should be in the form of `-N`, where `N` is an integer or an arithmetic expression representing the time difference in seconds. For example, `-60` denotes a travel of 60 seconds into the past, and `-60*60` represents a 1-hour(3600 seconds)  travel back in time.

```
mysql> create table t(a int);

Query OK, 1 row affected (0.04 sec)

mysql> select a from t at(offset => 1);
ERROR 1105 (HY000): InvalidArgument. Code: 2004, Text = TimeTravelPoint for 'Offset' must resolve to a constant negative integer. Provided expression '1' does not meet this requirement. Ensure the expression is a constant and negative integer.
mysql> select a from t at(offset => 'test');
ERROR 1105 (HY000): InvalidArgument. Code: 2004, Text = Need constant number.
Query OK, 1 row affected (0.07 sec)

mysql> select a from t at(offset => -5);
+------+
| a    |
+------+
|    1 |
+------+
1 row in set (0.04 sec)
Read 1 rows, 5.00 B in 0.012 sec., 82.23 rows/sec., 411.15 B/sec.

mysql> select a from t at(offset => -10 * 60);
ERROR 1105 (HY000): TableHistoricalDataNotFound. Code: 2013, Text = No historical data found at given point.
```

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/tests/sqllogictests/suites/base/12_time_travel/12_0004_time_travel_select_at.test
+++ b/tests/sqllogictests/suites/base/12_time_travel/12_0004_time_travel_select_at.test
@@ -1,0 +1,29 @@
+statement ok
+DROP DATABASE IF EXISTS db_12_0004
+
+statement ok
+CREATE DATABASE db_12_0004
+
+statement ok
+USE db_12_0004
+
+statement ok
+CREATE TABLE t(c1 int)
+
+statement ok
+insert into t values(1)
+
+statement error 2004
+select * from t at(timestamp => '1234')
+
+statement error 2004
+select * from t at(offset => 1)
+
+statement error 2013
+select * from t at(offset => -30 * 60)
+
+statement ok
+DROP TABLE t ALL
+
+statement ok
+DROP DATABASE db_12_0004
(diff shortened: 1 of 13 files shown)


tests/sqllogictests/suites/base/12_time_travel/12_0004_time_travel_select_at.test
dantengsky, line 1: we need a positive case, e.g.

insert into ...;
select sleep(1);
select * from t at (offset => -1)

dantengsky (approved):

PR #15184, opened 2024-04-08
refactor: add view system tables

TCeason:

## Summary

Add system tables: system.view and system.view_with_history

The view engine no need to display stat info.

And fuse engine does not have view_query.

### Note

After this pr merged, if user wants to show the views in one database, 

Only can according sql `show views [ from|in database ]`.

`show tables` only support show tables.

## Tests

- [x] Unit Test

## Type of change

- [x] Refactoring

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15184)


sundy-li (approved):

TCeason:
cc @soyeric128 Hi Eric, in this pr we add show views to query view in one database.

And show tables will not display view.

PR #15194, opened 2024-04-09
feat(query): implement ST_MAKEPOLYGON

kkk25641463:

## Summary
feat:  Implement  ST_MAKEPOLYGON(). see alse [doc](https://docs.snowflake.com/en/sql-reference/functions/st_makepolygon).
other: remove to_geometry alias,becasue register_aliases() only depend on alias_name as key which will cause the  registered function to be overwritten.
  ```
      pub fn register_aliases(&mut self, fn_name: &str, aliases: &[&str]) {
          for alias in aliases {
              self.aliases.insert(alias.to_string(), fn_name.to_string());
          }
      }
      registry.register_aliases("st_geometryfromwkb", &[
          "to_geometry",
      ]);
      registry.register_aliases("st_geometryfromwkt", &[
          "to_geometry",
      ]);
  ``` 

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/functions/src/scalars/geometry.rs
+++ b/src/query/functions/src/scalars/geometry.rs
@@ -24,8 +24,11 @@ use databend_common_expression::vectorize_with_builder_2_arg;
 use databend_common_expression::FunctionDomain;
 use databend_common_expression::FunctionRegistry;
 use databend_common_io::parse_to_ewkb;
+use databend_common_io::parse_to_subtype;
+use databend_common_io::GeometryDataType;
 use geo::MultiPoint;
 use geo::Point;
+use geo_types::Polygon;
 use geohash::decode_bbox;
 use geos::geo_types;
 use geos::geo_types::Coord;
@@ -34,8 +37,10 @@ use geos::Geom;
 use geos::Geometry;
 use geozero::geojson::GeoJson;
 use geozero::wkb::Ewkb;
+use geozero::wkt::Wkt;
 use geozero::CoordDimensions;
 use geozero::GeozeroGeometry;
+use geozero::ToGeo;
 use geozero::ToGeos;
 use geozero::ToJson;
 use geozero::ToWkb;
@@ -46,12 +51,12 @@ use geozero::ToWkt;
 pub fn register(registry: &mut FunctionRegistry) {
     // aliases
     registry.register_aliases("st_makegeompoint", &["st_geom_point"]);
+    registry.register_aliases("st_makepolygon", &["st_polygon"]);
     registry.register_aliases("st_makeline", &["st_make_line"]);
     registry.register_aliases("st_geometryfromwkb", &[
         "st_geomfromwkb",
         "st_geometryfromewkb",
         "st_geomfromewkb",
-        "to_geometry",
     ]);
     registry.register_aliases("st_geometryfromwkt", &[
         "st_geomfromwkt",
@@ -59,7 +64,6 @@ pub fn register(registry: &mut FunctionRegistry) {
         "st_geomfromewkt",
         "st_geometryfromtext",
         "st_geomfromtext",
-        "to_geometry",
     ]);
 
     // functions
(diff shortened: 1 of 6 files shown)


src/query/functions/src/scalars/geometry.rs
```diff
@@ -46,20 +51,19 @@ use geozero::ToWkt;
...
         "st_geomfromewkb",
-        "to_geometry",
     ]);
     registry.register_aliases("st_geometryfromwkt", &[
         "st_geomfromwkt",
         "st_geometryfromewkt",
         "st_geomfromewkt",
         "st_geometryfromtext",
         "st_geomfromtext",
-        "to_geometry",
```
b41sh, line 62: Can we keep this alias function? we need auto cast from string type values to geometry type.

src/query/functions/src/scalars/geometry.rs
```diff
@@ -163,6 +167,74 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            }
+
+            let subtype = match parse_to_subtype(binary) {
+                Ok(subtype) => subtype,
+                Err(e) => {
+                    ctx.set_error(
+                        builder.len(),
+                        ErrorCode::GeometryError(e.to_string()).to_string(),
+                    );
+                    builder.put_str("");
```
b41sh, line 188: we need call `builder.commit_row()` here, otherwise the offset value will be missed in the builder.

src/query/functions/src/scalars/geometry.rs
```diff
@@ -163,6 +167,74 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            };
+
+            let line_string = match geometry {
+                Ok(geo) => geo.try_into(),
+                Err(e) => {
+                    ctx.set_error(
+                        builder.len(),
+                        ErrorCode::GeometryError(e.to_string()).to_string(),
+                    );
+                    builder.put_str("");
```
b41sh, line 206: ditto

src/query/functions/src/scalars/geometry.rs
kkk25641463: https://github.com/datafuselabs/databend/discussions/15192
Because when user call to_geometry('0101000020797F000066666666A9CB17411F85EBC19E325641'), 
ST_GEOMETRYFROMWKT( <varchar_expression> ) will be called.But in ST_GEOMETRYFROMWKT, ST_GEOMETRYFROMWKB logic will execution.The logic of this code is a bit strange.

```
  pub fn register_aliases(&mut self, fn_name: &str, aliases: &[&str]) {
        for alias in aliases {
            self.aliases.insert(alias.to_string(), fn_name.to_string());
        }
    }
    registry.register_aliases("st_geometryfromwkb", &[
        "to_geometry",
    ]);
    registry.register_aliases("st_geometryfromwkt", &[
        "to_geometry",
    ]);
 registry.register_passthrough_nullable_1_arg::<StringType, GeometryType, _, _>(
        "st_geometryfromwkt",
        |_, _| FunctionDomain::MayThrow,
        vectorize_with_builder_1_arg::<StringType, GeometryType>(|wkt, builder, ctx| {
        match parse_to_subtype(binary) {
            GeometryDataType::EWKB => {
                 // call st_geometryfromwkb logic
            }),    
            GeometryDataType::EWKB => {
               // call st_geometryfromwkt logic
            }
         ......
);

```

I think it's better to implement it as independent function, not as an alias. 
```
 registry.register_passthrough_nullable_1_arg::<StringType, GeometryType, _, _>(
        "to_geometry",
 registry.register_passthrough_nullable_1_arg::<binary, GeometryType, _, _>(
        "to_geometry",
```

Let the community developers decide on the final implementation approach.

b41sh: ok, we can add this function later.

b41sh (approved):

PR #15205, opened 2024-04-10
feat(http): get client ip from headers

flaneur2020:

## Summary

currently the IP we got is retrieved from req.remote_addr, which represents the direct connection's IP address.

but when we deployed with having a proxy up front, it only reprents the proxy's IP.

this PR tries to retrieve the ip from headers including X-Real-IP, X-Forwarded-For, CF-Connecting-IP, these headers can help us retrieve the real ip of the client.

please note that the ip address from the headers are easy to be manipulated by middleman. we have to ensure the proxies are trustworthy, or the middleman may use a fake header to trick the network policy.

(however, it's necessary to trust the proxy in a cloud deployment IMHO.)

## Tests

- [x] Unit Test

## Type of change

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15205)

### DIFF
--- a/src/query/service/src/servers/http/middleware.rs
+++ b/src/query/service/src/servers/http/middleware.rs
@@ -113,7 +113,7 @@ fn get_credential(req: &Request, kind: HttpHandlerKind) -> Result<Credential> {
 /// not found, fallback to the remote address, which might be local proxy's ip address.
 /// please note that when it comes with network policy, we need make sure the incoming
 /// traffic comes from a trustworthy proxy instance.
-fn get_client_ip(req: &Request) -> Option<String> {
+pub fn get_client_ip(req: &Request) -> Option<String> {
     let headers = ["X-Real-IP", "X-Forwarded-For", "CF-Connecting-IP"];
     for &header in headers.iter() {
         if let Some(value) = req.headers().get(header) {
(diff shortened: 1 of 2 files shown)


src/query/service/src/servers/http/middleware.rs
```diff
@@ -95,22 +95,64 @@ fn get_credential(req: &Request, kind: HttpHandlerKind) -> Result<Credential> {
...
         }
     } else {
-        auth_by_header(&std_auth_headers, client_ip)
+        auth_by_header(&std_auth_headers, conn_ip)
+    }
+}
+
+/// this function tries to get the client IP address from the headers, if not found, it will
+/// return the connection's remote address. This IP address is ONLY used for logging purpose,
+/// never use this ip on authentication or network policy.
```
sundy-li, line 114:
```
fn get_client_ip(req: &Request) -> Option<String> {
    let headers = ["X-Real-IP", "X-Forwarded-For", "CF-Connecting-IP"];

    for &header in headers.iter() {
        if let Some(value) = req.headers().get(header) {
            if let Ok(ip_str) = value.to_str() {
                if header == "X-Forwarded-For" {
                    let first_ip = ip_str.split(',').next().unwrap_or("").trim();
                    if !first_ip.is_empty() {
                        return Some(first_ip.to_string());
                    }
                } else {
                    return Some(ip_str.to_string());
                }
            }
        }
    }

    get_connection_ip(req)
}
```

flaneur2020: thank you for sugestion, i've changed as your advice, PTAL

sundy-li (approved):

PR #15212, opened 2024-04-11
chore: change parquet_fast_read_bytes setting from 0 to 16MB

BohuTANG:

## Summary

When copying from the Parquet stage, we default to reading metadata each time, which is not efficient for small Parquet files (less than 16MB) due to the cost of reading metadata. 
See https://github.com/datafuselabs/databend/blob/c2e3ee652985a3a8aca1f4e3b46941889501fcb6/src/query/storages/parquet/src/parquet_rs/parquet_table/partition.rs#L79-L92

This PR sets `parquet_fast_read_bytes` default value to 16MB, meaning if a parquet file is less than 16MB, we won't read the metadata; instead, we'll read the entire file.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2998,7 +2998,6 @@ dependencies = [
  "ethnum 1.5.0 (git+https://github.com/ariesdevil/ethnum-rs?rev=4cb05f1)",
  "futures",
  "geo",
- "geos",
  "geozero",
  "goldenfile",
  "hex",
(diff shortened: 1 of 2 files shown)


dantengsky (approved):

BohuTANG:
The default value for `parquet_fast_read_bytes` is 16MB, so there's no need to configure it. CC @yufan022

yufan022:
> The default value for `parquet_fast_read_bytes` is 16MB, so there's no need to configure it. CC @yufan022

Copy that. Our current setting is 1073741824.

PR #15214, opened 2024-04-11
feat(query): implement ST_ASGEOJSON

kkk25641463:

## Summary
feat:  Implement  ST_ASGEOJSON(). 
see alse [doc](https://docs.snowflake.com/en/sql-reference/functions/st_asgeojson).

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15214)


src/query/functions/src/scalars/geometry.rs
```diff
@@ -67,6 +68,28 @@ pub fn register(registry: &mut FunctionRegistry) {
     ]);
 
     // functions
+    registry.register_passthrough_nullable_1_arg::<GeometryType, GeometryType, _, _>(
```
b41sh, line 71: The return type of `st_asgeojson` function is `An OBJECT in GeoJSON format` in Snowflake, we can use `Variant` type in Databend instead. The value of `Geometry` type should always be a binary format.

b41sh: `st_geomfromgeohash` and `st_geompointfromgeohash` function convert Geometry value to JSON format, but other geometry functions convert to ewkb format. This may cause a geometry column has multi types in a table and we need to check the type for every values before read it. So we'd better keep a consistent storage types, and the data display can be converted uniformly base on some settings, like `GEOMETRY_OUTPUT_FORMAT` in snowflake.

kkk25641463: I need to add GEOMETRY_OUTPUT_FORMAT config in system config,like
```
SET SESSION GEOMETRY_OUTPUT_FORMAT= EWKT;
```

b41sh: Thanks for the changes, can you help change the value of `st_geomfromgeohash` and `st_geompointfromgeohash` function to `ewkb` format by the way? I'm sorry I didn't notice this when I review the previous PR.

kkk25641463: > Thanks for the changes, can you help change the value of `st_geomfromgeohash` and `st_geompointfromgeohash` function to `ewkb` format by the way? I'm sorry I didn't notice this when I review the previous PR.

No problem,this pr can be suspended for a few days.I need to be familiar with the configuration item code and implement it together with the session config code and then add "SET SESSION GEOMETRY_OUTPUT_FORMAT= EWKT;" with unit test,otherwise the unit test will report error with ewkb.
  
I followed the approach of· get_timezone() to implement the session configuration, but I don't know how to get self.get_settings() in fmt().
```
  pub fn get_timezone(&self) -> Result<String> {
      self.try_get_string("timezone")
  }

impl<'a> Debug for ScalarRef<'a> {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
          ScalarRef::Geometry(s) => {
             // how can i get ctx and config settings here?
              match self.get_settings().get_geometry_output_format() {
                   EWKT => bin_to_ewkt,
                   GEO_JSON=>bin_to_json,
                  ......
             }
          }
```

kkk25641463: Add  [geometry_output_format](https://docs.snowflake.com/en/sql-reference/parameters#geometry-output-format) parameter;

drmingdrmer (commented):
Add a new version and corresponding test if protobuf message format changed:
```
    // Dear developer:
    //      If you're gonna add a new metadata version, you'll have to add a test for it.
    //      You could just copy an existing test file(e.g., `../tests/it/v024_table_meta.rs`)
    //      and replace two of the variable `bytes` and `want`.

```
https://github.com/drmingdrmer/databend/blob/f9d2b79400bf5939bd5308451470c62bb917db9d/src/meta/proto-conv/src/util.rs#L120

> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15214#-:-NvoThrD4Z4oHpgfQvBA:b-6m9ttf)* status: 0 of 35 files reviewed, 1 unresolved discussion (waiting on @b41sh)

kkk25641463:
> Add a new version and corresponding test if protobuf message format changed:
> 
> ```
>     // Dear developer:
>     //      If you're gonna add a new metadata version, you'll have to add a test for it.
>     //      You could just copy an existing test file(e.g., `../tests/it/v024_table_meta.rs`)
>     //      and replace two of the variable `bytes` and `want`.
> ```
> 
> https://github.com/drmingdrmer/databend/blob/f9d2b79400bf5939bd5308451470c62bb917db9d/src/meta/proto-conv/src/util.rs#L120
> 
> > Reviewed all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15214#-:-NvoThrD4Z4oHpgfQvBA:b-6m9ttf)_ status: 0 of 35 files reviewed, 1 unresolved discussion (waiting on @b41sh)

drmingdrmer (commented):
The protobuf and meta part is great to me

> Reviewed 9 of 34 files at r2, 2 of 8 files at r3, 3 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15214#-:-NvpABgV5nrG0GYS6tFN:bz31s6o)* status: 14 of 38 files reviewed, 1 unresolved discussion (waiting on @b41sh)

src/common/io/src/geometry.rs
```diff
@@ -119,3 +168,60 @@ pub fn parse_to_subtype(buf: &[u8]) -> Result<GeometryDataType> {
...
+}
+impl<B: AsRef<[u8]>> GeometryFormatOutput for Ewkb<B> {
+    fn format(self, format_type: GeometryDataType) -> Result<String> {
+        match format_type {
+            GeometryDataType::WKB => self
+                .to_wkb(CoordDimensions::xy())
+                .map(|bytes| {
+                    bytes
+                        .iter()
+                        .map(|b| format!("{:02X}", b))
```
b41sh, line 197: better to use `hex::encode_upper` function to convert binary to hex format for better display.

src/query/formats/src/field_encoder/values.rs
```diff
@@ -298,14 +308,27 @@ impl FieldEncoderValues {
...
-                self.write_string_inner(json.as_bytes(), out_buf, in_nested);
-            }
+        let s = match self.common_settings().geometry_format {
+            GeometryDataType::WKB => Ewkb(v.to_vec())
+                .to_wkb(CoordDimensions::xy())
+                .unwrap()
+                .as_bytes()
+                .to_vec(),
+            GeometryDataType::WKT => Ewkb(v.to_vec()).to_wkt().unwrap().as_bytes().to_vec(),
+            GeometryDataType::EWKB => v.to_vec(),
```
b41sh, line 318: `EWKB` and `WKB` need to use `hex::encode_upper` to convert binary values to hex format for better display.

src/query/formats/src/field_encoder/values.rs
```diff
@@ -298,14 +308,27 @@ impl FieldEncoderValues {
...
+            GeometryDataType::EWKT => Ewkb(v.to_vec())
+                .to_ewkt(read_ewkb_srid(&mut io::Cursor::new(&v)).unwrap())
+                .unwrap()
+                .as_bytes()
+                .to_vec(),
+            GeometryDataType::GEOJSON => GeoJson(std::str::from_utf8(v).unwrap())
```
b41sh, line 324: value is binary format, not a string, using `std::str::from_utf8` will cause panic.

b41sh (approved):
LGTM, thanks for your great contribution. @kkk25641463

PR #15229, opened 2024-04-13
fix: solve some inconsistencies between code and comments

### LINKED ISSUE #15162: Inconsistency between code and documentation
In databend/src/common/arrow/src/arrow/array/map/mod.rs，the comment "panics iff `offset + length >= self.len()` " corresponds to `assert!(offset + length < self.len())` while the code is `assert!(offset + length < self.len())`. 
```rust
# Panics
    /// panics iff `offset + length >= self.len()`
    pub fn slice(&mut self, offset: usize, length: usize) {
        assert!(
            offset + length <= self.len(),
            "the offset of the new Buffer cannot exceed the existing length"
        );
        unsafe { self.slice_unchecked(offset, length) }
    }
```
In databend/src/common/arrow/src/arrow/array/list/mod.rs, the situation is similar.
```rust
/// # Panics
    /// panics iff `offset + length >= self.len()`
    pub fn slice(&mut self, offset: usize, length: usize) {
        assert!(
            offset + length <= self.len(),
            "the offset of the new Buffer cannot exceed the existing length"
        );
        unsafe { self.slice_unchecked(offset, length) }
    }
```
In databend/src/common/arrow/src/arrow/array/primitive/mod.rs and databend/src/common/arrow/src/arrow/array/dictionary/mod.rs, I don't think“out-of-bound” crash is the behavior we expect.
The ideal situation is to inform the "user" what should happen if it is out-of-bound，so I think use assert! to check whether the
 index is out-of-bound before use it and add some information when the code panics is a better way.
databend/src/common/arrow/src/arrow/array/primitive/mod.rs
```rust
    /// # Panic
    /// This function panics iff `i >= self.len`.
    #[inline]
    pub fn value(&self, i: usize) -> T {
        self.values[i]
    }
```
databend/src/common/arrow/src/arrow/array/dictionary/mod.rs
```rust
/// # Panics
    /// This function panics iff `index >= self.len()`
    #[inline]
    pub fn key_value(&self, index: usize) -> usize {
        // safety - invariant of the struct
        unsafe { self.keys.values()[index].as_usize() }
    }
```

Similarly, in databend/src/common/arrow/src/arrow/io/ipc/read/common.rs, the code should check whether projection` is empty before use it.
```rust
 /// # Panics
    /// iff `projection` is empty
    pub fn new(projection: &'a [usize], iter: I) -> Self {
        Self {
            projection: &projection[1..],
            iter,
            current_count: 0,
            current_projection: projection[0],
        }
    }
```

YichiZhang0613:

## Summary

Solve the inconsistencies between code and comments which are proposed in #15162

- Close #15162

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15229)


src/common/arrow/src/arrow/io/ipc/read/common.rs
```diff
@@ -53,6 +53,7 @@ impl<'a, A, I: Iterator<Item = A>> ProjectionIter<'a, A, I> {
     /// # Panics
     /// iff `projection` is empty
     pub fn new(projection: &'a [usize], iter: I) -> Self {
+        assert!(projection.is_empty(), "projection cannot be empty");
```
sundy-li, line 56: assert!(!projection.is_empty(), "projection cannot be empty");

sundy-li (approved):

PR #15249, opened 2024-04-17
feat(query): add user option: disabled

TCeason:

## Summary

If specify DISABLED = TRUE for a user:

The user cannot log in again.

```sql
-- root user: u1 can not login
create user u1 identified by '123' with disabled=true;
-- root user: u2 can not login
alter user u2 with disabled=true;
$ bendsql -uu2 -pabc123;
Welcome to BendSQL 0.12.4-unknown.
Connecting to localhost:8000 as user u2.
Error: APIError: RequestError: Start Query failed with status 401 Unauthorized: {"error":{"code":"401","message":"AuthenticateFailure: user u2 is disabled. Not allowed to login"}}

```

### Note

If a period of time passes, need to re-enable the user:

```sql
-- root user
ALTER USER u2 with DISABLED = FALSE;

$ bendsql -uu2 -pabc123 --query="select 'Enabled'"
Enabled

```

## Tests

- [x] Logic Test

## Type of change

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15249)


drmingdrmer (approved):
> Reviewed 17 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15249)* status: 17 of 19 files reviewed, all discussions resolved (waiting on @b41sh and @flaneur2020)

src/meta/app/src/principal/user_info.rs
```diff
@@ -164,6 +164,8 @@ pub struct UserOption {
     network_policy: Option<String>,
 
     password_policy: Option<String>,
+
+    disabled: Option<bool>,
```
b41sh, line 168: How about changing this field to a key/value map? so that we can add other config properties in the future.

src/query/ast/src/parser/statement.rs
```diff
@@ -3769,6 +3769,12 @@ pub fn user_option(i: Input) -> IResult<UserOptionItem> {
         },
         |(_, _, _)| UserOptionItem::UnsetNetworkPolicy,
     );
+    let set_disabled_option = map(
```
b41sh, line 3772: Also need to add an unset disabled to modify this operation.

b41sh (approved):

drmingdrmer (approved):
> Reviewed 2 of 19 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15249)* status: all files reviewed, 2 unresolved discussions (waiting on @flaneur2020 and @TCeason)

drmingdrmer:
This is not a big deal but in general, I'd suggest to use `enabled` in place of `disabled`, because a double negative `user.disabled being false` make it more difficult to read.

TCeason:
> This is not a big deal but in general, I'd suggest to use `enabled` in place of `disabled`, because a double negative `user.disabled being false` make it more difficult to read.
> 
> 

I got you. But this option mainly used to disable user login. So disable is directly.

If use enable, user wants to disable will write enable=false.

PR #15258, opened 2024-04-17
feat: http handler response add field has_result_set.

youngsofun:

## Summary

1.  with `has_result_set=false`,  `execute()` in driver can skip processing of results and block until the end of query.
2.  similarly, with the `Starting` state,  it is clear that most response fields about state and result should be ignored.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


youngsofun:
has_result_set is a propety of the query type (Select/Insert/CreateTable ...)

`has_result_set=false` means the `data` will be empty for all pages (this page and pages afterward).

sundy-li:
> with has_result_set=false, execute() in driver can skip processing of results and block until the end of query

We don't know the result of the query if we skip pulling the result. 

For example:

```
create table .... as select ...
```

If the client skip processing the result. Client will always get success state for this query even it may be fail.

sundy-li (approved):

hantmac (approved):

PR #15261, opened 2024-04-18
chore: tune auto-compact parameters triggered by insertions

dantengsky:

## Summary

For auto compaction triggered by insertions (copy-into, insert):

Previously, the default value of setting `auto_compaction_imperfect_blocks_threshold`  is set to 50 blocks, but the block compaction triggered is limited by an empirical value of 3 segments. In some cases, this may lead to the "diverged" fragmentation of table block/segments.

In this PR,  auto block compaction is changed to be limited by the number of blocks, which is a hint value that detected after insertion:

- if the number of fragmented block detected after insertion is larger than `auto_compaction_imperfect_blocks_threshold`, then a hint value of at most `auto_compaction_imperfect_blocks_threshold * 1.5` will be chosen.

- during the selection of blocks to be compacted, the hint value will limit the number of blocks to be collected (across segments if necessary)

Also, the default value of `auto_compaction_imperfect_blocks_threshold` changed from 50 to 25, to make the compaction process smoother.

Other compaction behavior(triggered by merge, replace..) are kept as they are.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test  - covered by existing tests

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement
- [x] Other (please describe): tuning auto compaction strategy

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15261)


src/query/catalog/src/table.rs
```diff
@@ -534,3 +542,40 @@ impl ColumnStatisticsProvider for ParquetTableColumnStatisticsProvider {
...
+impl CompactionLimits {
+    pub fn limits(segment_limit: Option<usize>, block_limit: Option<usize>) -> Self {
+        let adjusted_segment_limit = match (block_limit, segment_limit) {
+            (Some(block_limit), None) => {
+                // As n fragmented blocks scattered across at most n segments,
+                // when no segment_limit provided, we set it to the same value of block_limit
+                Some(block_limit)
+            }
+            _ => segment_limit,
+        };
```
zhyass, line 561:
```suggestion
        // As n fragmented blocks scattered across at most n segments,
        // when no segment_limit provided, we set it to the same value of block_limit
        let adjusted_segment_limit = segment_limit.or(block_limit);
```

zhyass (approved):
LGTM

Issue #15263, opened 2024-04-18
Test `DROP IF EXIST` for `SEQUENCE`


drmingdrmer:
> Reviewed 2 of 50 files at r1, 1 of 13 files at r9, 2 of 5 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15247)* status: 46 of 53 files reviewed, 4 unresolved discussions (waiting on @lichuang)

___
*[`src/query/service/src/interpreters/interpreter_sequence_drop.rs` line 54 at r11](https://reviewable.io/reviews/datafuselabs/databend/15247#-NvkOZ1P4hjcOU_g8jw3:-NvkOZ1P4hjcOU_g8jw4:bofdtpc) ([raw file](https://github.com/datafuselabs/databend/blob/a32ca69b7bebb6d097a821469fa15ee31d3206a1/src/query/service/src/interpreters/interpreter_sequence_drop.rs#L54)):*
> ```XML
>         };
>         let catalog = self.ctx.get_default_catalog()?;
>         let _reply = catalog.drop_sequence(req).await?;
> ```

```rust,ignore
if _reply.prev.is_none() && !self.plan.if_exists {
    return Err(...)
}
```
___
*[`tests/sqllogictests/suites/base/05_ddl/05_0036_sequence.test` line 8 at r11](https://reviewable.io/reviews/datafuselabs/databend/15247#-NvkP7616DGXm5ju_l0q:-NvkP7616DGXm5ju_l0r:b-5uvizk) ([raw file](https://github.com/datafuselabs/databend/blob/a32ca69b7bebb6d097a821469fa15ee31d3206a1/tests/sqllogictests/suites/base/05_ddl/05_0036_sequence.test#L8)):*
> ```
> 
> statement ok
> DROP SEQUENCE seq
> ```

Need to test `DROP IF EXISTS`

_Originally posted by @drmingdrmer in https://github.com/datafuselabs/databend/pull/15247#pullrequestreview-2008268648_

PR #15294, opened 2024-04-22
feat: add nextval table function

### LINKED ISSUE #15264: support use `nextval(seq)` as insert values of table

lichuang:

## Summary

support insert value use `nextval` function such as:
* `INSERT INTO tmp values(nextval(seq))`
* `INSERT INTO tmp2 select nextval(seq) from numbers(2);`
* `select nextval(seq)`

- Fixes #15264 #15263

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15294)


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -741,6 +742,10 @@ impl<'a> TypeChecker<'a> {
                 {
                     if let Some(udf) = self.resolve_udf(*span, func_name, args).await? {
                         return Ok(udf);
+                    } else if let Some(async_function) =
+                        self.resolve_async_function(*span, func_name, args).await?
```
TCeason, line 746: Where filter can contains a sync function? @b41sh

b41sh: seems `async_function` can only used in insert values.

src/query/sql/src/async_function/async_function_mgr.rs
```diff
@@ -0,0 +1,104 @@
...
+use databend_common_exception::Span;
+use databend_common_expression::types::DataType;
+use databend_common_meta_app::tenant::Tenant;
+
+use super::sequence_async_function::SequenceAsyncFunction;
+use crate::plans::AsyncFunctionCall;
+use crate::ScalarExpr;
+
+#[async_trait::async_trait]
+pub trait AsyncFunction: Sync + Send {
```
b41sh, line 32: move `AsyncFunction` to functions module is more appropriate.

TCeason (approved):

src/query/sql/src/async_function/sequence_async_function.rs
```diff
@@ -0,0 +1,121 @@
...
+    async fn resolve(
+        &self,
+        span: Span,
+        tenant: Tenant,
+        catalog: Arc<dyn Catalog>,
+        arguments: &[Expr],
+    ) -> Result<Option<Box<(ScalarExpr, DataType)>>> {
+        if arguments.len() != 1 {
+            return Err(ErrorCode::SemanticError(
+                "nextval function need one argument".to_string(),
```
b41sh, line 66: better to add the actual number of arguments in the error message, like add `but got ...`.

src/query/sql/src/planner/semantic/lowering.rs
```diff
@@ -261,6 +261,10 @@ impl ScalarExpr {
                 let scalar = &udf.scalar;
                 scalar.as_raw_expr()
             }
+
+            ScalarExpr::AsyncFunctionCall(_table_func) => {
+                unreachable!()
```
b41sh, line 266: This will cause `select nextval(seq)` panic, do we plan to support this feature in the future?

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -741,6 +742,10 @@ impl<'a> TypeChecker<'a> {
                 {
                     if let Some(udf) = self.resolve_udf(*span, func_name, args).await? {
                         return Ok(udf);
+                    } else if let Some(async_function) =
```
b41sh, line 745: Better to add a const array like `GENERAL_WINDOW_FUNCTIONS` to check whether the function is a async function. This can help us to check the function names and avoid the udf function having the same name.

src/query/sql/src/async_function/async_function_mgr.rs
```diff
@@ -0,0 +1,104 @@
...
+    fn func_name(&self) -> String;
+
+    async fn generate(
```
sundy-li, line 35: Too many args for the trait.

The `tenant`, `catalog` should be inside the struct which implements AsyncFunction

tests/sqllogictests/suites/base/05_ddl/05_0036_sequence.test
```diff
@@ -4,5 +4,69 @@ DROP SEQUENCE IF EXISTS seq
...
+1 5 3
+
 statement ok
 DROP SEQUENCE seq
+
+statement error 1126
+DROP SEQUENCE seq
+
+statement error 3101
+INSERT INTO tmp3 values(1,nextval(seq),3);
```
sundy-li, line 63: how to optimize query:

```
INSERT INTO tmp3 select 1,nextval(seq) from numbers(10000)
```

BohuTANG: Another question (not mentioned in the PR Summary): How does `nextval` function work in a distributed clsuter?

lichuang: add new `RelOperator::AsyncFunction`, it distrubute mode is `serial`.

add a transformer `transform_sequence_nextval.rs`, it will generate sequence base on number rows of input block, hence not query meta for many time:
```
        let count = data_block.num_rows() as u64;
        let req = GetSequenceNextValueReq {
            ident: SequenceIdent::new(&tenant, &self.sequence),
            count,
        };
```

src/query/expression/src/evaluator.rs
```diff
@@ -249,6 +266,12 @@ impl<'a> Evaluator<'a> {
 
                 self.run_lambda(name, args, lambda_expr, return_type)
             }
+            // async function call values has been generated by transformer
+            Expr::AsyncFunctionCall { .. } => {
```
sundy-li, line 270: The  `Expr::AsyncFunctionCall`  will be replaced by `Expr::ColumnRef`. There is no need to introduce `Expr::AsyncFunctionCall` in evaluator.

src/query/service/src/pipelines/processors/transforms/transform_sequence_nextval.rs
```diff
@@ -0,0 +1,90 @@
...
+        let count = data_block.num_rows() as u64;
+        let tenant = self.ctx.get_tenant();
+        let catalog = self.ctx.get_default_catalog()?;
+        let req = GetSequenceNextValueReq {
+            ident: SequenceIdent::new(&tenant, &self.sequence),
+            count,
+        };
+        let resp = catalog.get_sequence_next_value(req).await?;
+        let mut start = resp.start;
+        let mut builder = ColumnBuilder::with_capacity(&self.return_type, data_block.num_rows());
```
sundy-li, line 76: why not use `Vec<u64>` to build the column ?

src/query/service/src/pipelines/processors/transforms/transform_sequence_nextval.rs
```diff
@@ -0,0 +1,90 @@
...
+        for _ in 0..count {
+            let scalar = ScalarRef::Number(NumberScalar::UInt64(start));
```
sundy-li, line 78: u64 is packed into Scalar and unpacked inside builder which is overhead.

sundy-li: `let col = UInt64Type::from_data(start...start+count);`

src/query/sql/src/planner/binder/select.rs
```diff
@@ -295,6 +296,10 @@ impl Binder {
 
         s_expr = self.bind_projection(&mut from_context, &projections, &scalar_items, s_expr)?;
 
+        // rewrite async function to async function plan
+        let mut async_func_rewriter = AsyncFunctionRewriter::new();
+        s_expr = async_func_rewriter.rewrite(&s_expr)?;
+
```
b41sh, line 302: Is `AsyncFunctionCall` allowed to execute on where, having, join, etc.? If not, it needs to be checked in the binder phase and return a semantic error to avoid an unpredictable panic.

sundy-li (approved):

b41sh (approved):

BohuTANG:
The `nextval` performance looks good!

Schema:
```
CREATE OR REPLACE SEQUENCE seq;
CREATE OR REPLACE TABLE tmp(a int, b uint64, c int);
```

Insert with nextval:
```
INSERT INTO tmp select nextval(seq), number, number from numbers(50000000);
```

Small-Size Warehouse:
```
50000000 rows affected | 16s
```

Medium-Size Warehouse:
```
50000000 rows affected | 20s
```

Without `nextval`, Small-Size warehouse:
```
INSERT INTO tmp select number, number, number from numbers(50000000);

50000000 rows affected | 7.0s
```

lichuang:
@soyeric128

PR #15298, opened 2024-04-23
chore(planner): fix can_filter_null

### LINKED ISSUE #15256: bug: error using IFNULL or COALESCE in statement
### Search before asking

### Version

https://github.com/datafuselabs/databend/releases/tag/v1.2.423-nightly

### What's Wrong?

Get error `no overload satisfies `assume_not_null(NULL)` when adding `IFNULL` or `COALESCE` into statement

See below SQL, if line 
`WHERE IFNULL(base.context, '') = ''`
reads
`WHERE base.context = ''`
Then the SQL works ok, but if COALESCE OR IFNULL used, it raises the exception

### How to Reproduce?

```sql
drop table if exists m1;
drop table if exists j1;
create table m1(id varchar, "context" varchar);
create table j1(id varchar);

WITH base
AS (
	SELECT id, context
	FROM m1
	)
	,src1
AS (
	SELECT base.id
	FROM base
	WHERE IFNULL(base.context, '') = ''
	)
	,join1
AS (
	SELECT id
	FROM j1
	)
SELECT src1.id
FROM src1
LEFT OUTER JOIN join1 ON true;
```

Dousir9:

## Summary

In `can_filter_null`, if the function is `assume_not_null` or `remove_nullable`, we cannot replace the column bindings with `Scalar::Null`.

- Fixes #15256 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15298)

### DIFF
--- a/src/query/sql/src/planner/optimizer/rule/rewrite/push_down_filter_join/outer_join_to_inner_join.rs
+++ b/src/query/sql/src/planner/optimizer/rule/rewrite/push_down_filter_join/outer_join_to_inner_join.rs
@@ -186,6 +186,13 @@ pub fn can_filter_null(
                     Ok(())
                 }
                 ScalarExpr::FunctionCall(func) => {
+                    // If the function is `assume_not_null`, we can not replace
+                    // the column bindings with `Scalar::Null`.
+                    if func.func_name == "assume_not_null" {
+                        self.can_replace = false;
+                        return Ok(());
+                    }
+
                     if func.func_name != "or" {
                         for expr in &mut func.arguments {
                             self.replace(expr, column_set)?;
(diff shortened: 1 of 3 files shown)


src/query/sql/src/planner/optimizer/rule/rewrite/push_down_filter_join/outer_join_to_inner_join.rs
```diff
@@ -186,6 +186,13 @@ pub fn can_filter_null(
                     Ok(())
                 }
                 ScalarExpr::FunctionCall(func) => {
+                    // If the function is `assume_not_null`, we can not replace
+                    // the column bindings with `Scalar::Null`.
+                    if func.func_name == "assume_not_null" {
```
sundy-li, line 191: there is alias `remove_nullable`

xudong963 (approved):

PR #15302, opened 2024-04-23
feat(query): implement TO_GEOMETRY/TRY_TO_GEOMETRY function

kkk25641463:

## Summary
feat:  Implement  to_geometry() and try_to_geometry(). 
see alse [doc](https://docs.snowflake.com/en/sql-reference/functions/to_geometry).

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15302)


src/query/functions/src/scalars/geometry.rs
```diff
@@ -738,3 +816,119 @@ pub fn ewkb_to_json(buf: &[u8]) -> databend_common_exception::Result<String> {
...
+    ctx: &mut EvalContext,
+) {
+    let geo_type = match parse_to_subtype(str.as_bytes()) {
+        Ok(geo_types) => geo_types,
+        Err(e) => {
+            ctx.set_error(
+                builder.len(),
+                ErrorCode::GeometryError(e.to_string()).to_string(),
+            );
+            builder.put_str("");
```
b41sh, line 867: `builder.put_str("");` is not needed, just add `builder.commit_row();`

src/query/functions/src/scalars/geometry.rs
```diff
@@ -577,6 +551,110 @@ pub fn register(registry: &mut FunctionRegistry) {
         }),
     );
 
+    registry.register_passthrough_nullable_1_arg::<StringType, GeometryType, _, _>(
+        "to_geometry",
```
b41sh, line 555: We can also support the `try_to_geometry` function by the way, which returns NULL instead of reporting an error when it is encountered.

src/query/functions/src/scalars/geometry.rs
```diff
@@ -570,6 +557,299 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                        builder.len(),
+                        ErrorCode::GeometryError(e.to_string()).to_string(),
+                    ),
+                }
+                builder.commit_row();
+            },
+        ),
+    );
+
+    registry.register_1_arg_core::<StringType, NullableType<GeometryType>, _, _>(
```
b41sh, line 706: better to use `register_combine_nullable_1_arg` so that we can support both `StringType` and `Nullable StringType`, you can refer `try_parse_json` function as an example.

src/query/functions/src/scalars/geometry.rs
```diff
@@ -570,6 +557,299 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        "try_to_geometry",
+        |_, _| FunctionDomain::MayThrow,
```
b41sh, line 708: Domain is `FunctionDomain::Full`, because we will not throw an error in this function.

b41sh (approved):

PR #15315, opened 2024-04-23
fix: incorrect overlap checking (replace into)

### LINKED ISSUE #15307: bug: replace into when on fields have bool type will insert a new record
### Search before asking

### Version

 v1.2.313-nightly-3acb8aaedc(rust-1.77.0-nightly-2024-01-23T22:05:48.341888173Z)

### What's Wrong?

replace into when on fields have bool type will insert a new record

### How to Reproduce?

replace into a on(f1,f2,f3) select f1,f2,f3 from b

dantengsky:

## Summary

During overlap detection while executing replace-into stmt,  if column being checked has no range index, we should assume there are overlaps.

before this PR:

~~~
create or replace table t( c1 bool);
replace into t on(c1) values(false);
replace into t on(c1) values(false);
select count(*) from t; -- return 2, which is incorrect
~~~

with this PR

~~~
create or replace table t( c1 bool);
replace into t on(c1) values(false);
replace into t on(c1) values(false);
select count(*) from t; -- return 1, which is expected
~~~

- Fixes #15307

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15315)

### DIFF
--- a/scripts/ci/deploy/config/databend-query-node-1.toml
+++ b/scripts/ci/deploy/config/databend-query-node-1.toml
@@ -73,7 +73,7 @@ join_spilling_memory_ratio = 60
 [log]
 
 [log.file]
-level = "DEBUG"
+level = "INFO"
 format = "text"
 dir = "./.databend/logs_1"
 prefix_filter = ""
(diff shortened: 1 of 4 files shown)


zhyass (approved):
LGTM

PR #15322, opened 2024-04-24
chore(planner): add disable merge into join reorder settings

Dousir9:

## Summary

Add `disable_merge_into_join_reorder` settings.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [x] Other (please describe): add settings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15322)


xudong963 (approved):

src/query/settings/src/settings_default.rs
```diff
@@ -290,6 +290,12 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: Some(SettingRange::Numeric(0..=u64::MAX)),
                 }),
+                ("disable_merge_into_join_reorder", DefaultSettingValue {
```
sundy-li, line 293: why not use `disable_join_reorder` directly?

Dousir9: The `disable_merge_into_join_reorder` is used to disable the exchange between target and source, while `disable_join_reorder` is used to disable join reorder in source. @sundy-li

BohuTANG: Is it possible to add a  automatic check when to disable? Instead of adding a new setting by manual to disable it.

PR #15363, opened 2024-04-28
fix(query): show processlit host is null in HTTPQuery

### LINKED ISSUE #15359: show processlist host is NULL
**Summary**

The client host is NULL in show processlist:
```
SHOW PROCESSLIST

┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│          node          │                  id                  │    type   │       host       │  user  │ command │  database  │     extra_info    │ memory_usage │ data_read_bytes │ data_write_bytes │ scan_progress_read_rows │ scan_progress_read_bytes │ mysql_connection_id │  time  │       status       │
│         String         │                String                │   String  │ Nullable(String) │ String │  String │   String   │       String      │     Int64    │      UInt64     │      UInt64      │          UInt64         │          UInt64          │   Nullable(UInt32)  │ UInt64 │       String       │
├────────────────────────┼──────────────────────────────────────┼───────────┼──────────────────┼────────┼─────────┼────────────┼───────────────────┼──────────────┼─────────────────┼──────────────────┼─────────────────────────┼──────────────────────────┼─────────────────────┼────────┼────────────────────┤
│ G7AgYzpeH94ZqjPDahFLy5 │ cceb994d-b024-4619-ba1d-65e3668b853a │ HTTPQuery │ NULL             │ bohu   │ Query   │ tpch_sf100 │ SHOW PROCESSLIST  │            0 │               0 │                0 │                       0 │                        0 │                NULL │      0 │ executing pipeline │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

TCeason:

## Summary

Fix show processlit host is null in HTTPQuery.

In HTTP auth , init the client addr.

- Fixes #15359

## Tests

- [x] Logic Test

## Type of change

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15363)

### DIFF
--- a/src/query/service/src/servers/http/middleware.rs
+++ b/src/query/service/src/servers/http/middleware.rs
@@ -256,6 +256,10 @@ impl<E> HTTPSessionEndpoint<E> {
             .get(TRACE_PARENT)
             .map(|id| id.to_str().unwrap().to_string());
         let baggage = extract_baggage_from_headers(req.headers());
+        let client_host = match req.remote_addr().0 {
+            Addr::SocketAddr(addr) => Some(addr),
+            _ => None,
+        };
         Ok(HttpQueryContext::new(
             session,
             query_id,
@@ -266,6 +270,7 @@ impl<E> HTTPSessionEndpoint<E> {
             baggage,
             req.method().to_string(),
             req.uri().to_string(),
+            client_host,
         ))
     }
 }
--- a/src/query/service/src/servers/http/v1/query/http_query.rs
+++ b/src/query/service/src/servers/http/v1/query/http_query.rs
@@ -406,6 +406,9 @@ impl HttpQuery {
         let deduplicate_label = &ctx.deduplicate_label;
         let user_agent = &ctx.user_agent;
         let query_id = ctx.query_id.clone();
+
+        session.set_client_host(ctx.client_host);
+
         let http_ctx = ctx;
         let ctx = session.create_query_context().await?;
 
--- a/src/query/service/src/servers/http/v1/query/http_query_context.rs
+++ b/src/query/service/src/servers/http/v1/query/http_query_context.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::collections::BTreeMap;
+use std::net::SocketAddr;
 use std::sync::Arc;
 
 use http::StatusCode;
@@ -35,6 +36,7 @@ pub struct HttpQueryContext {
     pub opentelemetry_baggage: Option<Vec<(String, String)>>,
     pub http_method: String,
     pub uri: String,
+    pub client_host: Option<SocketAddr>,
 }
 
 impl HttpQueryContext {
@@ -48,6 +50,7 @@ impl HttpQueryContext {
         open_telemetry_baggage: Option<Vec<(String, String)>>,
         http_method: String,
         uri: String,
+        client_host: Option<SocketAddr>,
     ) -> Self {
         HttpQueryContext {
             session,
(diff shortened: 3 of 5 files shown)


TCeason:
cc @wubx

zhang2014 (approved):

ZhiHanZ (approved):

BohuTANG:
In cloud, the client ip will be gateway?

TCeason:
> In cloud, the client ip will be gateway? 

cc @ZhiHanZ Cloud you please answer this question?

BTW, we had a method get_client_ip it use the same way to parse req.uri.

https://github.com/datafuselabs/databend/blob/0d07cc648aff07c230c8b7dcc5fae249094bd4ed/src/query/service/src/servers/http/middleware.rs#L116

flaneur2020 (approved):
oh, i think using [databend/src/query/service/src/servers/http/middleware.rs](https://github.com/datafuselabs/databend/blob/0d07cc648aff07c230c8b7dcc5fae249094bd4ed/src/query/service/src/servers/http/middleware.rs#L116) get_client_ip might be better

flaneur2020:
> In cloud, the client ip will be gateway?

i added a 

> In cloud, the client ip will be gateway?

not only cloud, whenever you put a proxy up front of the query instances, you will get the ip of the proxy.

when you had a proxy, it'd better to utilize the `get_client_ip()` method to get the real client ip which proxy added into the headers like `X-Real-IP`.

TCeason:
> > In cloud, the client ip will be gateway?
> 
> 
> 
> not only cloud, whenever you put a proxy up front of the query instances, you will get the ip of the proxy.
> 
> 
> 
> when you had a proxy, it'd better to utilize the `get_client_ip()` method to get the real client ip which proxy added into the headers like `X-Real-IP`.

So in cloud it will display as null? Because only a IP can not parse as socketaddr.

flaneur2020:
> > > In cloud, the client ip will be gateway?
> > 
> > 
> > not only cloud, whenever you put a proxy up front of the query instances, you will get the ip of the proxy.
> > when you had a proxy, it'd better to utilize the `get_client_ip()` method to get the real client ip which proxy added into the headers like `X-Real-IP`.
> 
> So in cloud it will display as null? Because only a IP can not parse as socketaddr.
> 
> [#15359 (comment)](https://github.com/datafuselabs/databend/issues/15359#issuecomment-2081412247)
> 
> Here the host is ip:port

IMHO it will be filled up with the proxy's ip address  🤔

TCeason:
> > > > In cloud, the client ip will be gateway?
> > > 
> > > 
> > > not only cloud, whenever you put a proxy up front of the query instances, you will get the ip of the proxy.
> > > when you had a proxy, it'd better to utilize the `get_client_ip()` method to get the real client ip which proxy added into the headers like `X-Real-IP`.
> > 
> > 
> > So in cloud it will display as null? Because only a IP can not parse as socketaddr.
> > [#15359 (comment)](https://github.com/datafuselabs/databend/issues/15359#issuecomment-2081412247)
> > Here the host is ip:port
> 
> IMHO it will be filled up with the proxy's ip address 🤔

Does the show procelist need to display ports?

Now the get client ip address will be obtained from the header. But this is a pure ip, and it's represented by a string.

In our current session ctx, client_ip is Option<SocketAddr>.

This appears to be CK-compatible behavior. But the port should be meaningless information.

So you want to set session _ctx.client_ip: Option<String> and discard the port.

PR #15378, opened 2024-04-30
refactor: remove stage use stream to avoid OOM.

youngsofun:

## Summary

remove stage use stream instead of load all paths into one Vec to avoid OOM.

00_0004_copy_with_max_files.sh is changed too, no longer rely on the order of list of backends . cc @Xuanwo 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15378)

### DIFF
--- a/src/query/service/src/interpreters/interpreter_user_stage_remove.rs
+++ b/src/query/service/src/interpreters/interpreter_user_stage_remove.rs
@@ -75,8 +75,10 @@ impl Interpreter for RemoveUserStageInterpreter {
         let table_ctx: Arc<dyn TableContext> = self.ctx.clone();
         let file_op = Files::create(table_ctx, op);
 
+        const REMOVE_BATCH: usize = 1000;
         let mut chunks = files.chunks(REMOVE_BATCH);
-        const REMOVE_BATCH: usize = 4000;
+
+        // s3 can remove at most 1k files in one request
         while let Some(chunk) = chunks.next().await {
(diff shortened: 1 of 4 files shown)


src/query/service/src/interpreters/interpreter_user_stage_remove.rs
```diff
@@ -68,19 +70,17 @@ impl Interpreter for RemoveUserStageInterpreter {
...
-            .await?
-            .into_iter()
-            .map(|file_with_meta| file_with_meta.path)
-            .collect::<Vec<_>>();
+        let files = files_info.list_stream(&op, thread_num, None).await?;
 
         let table_ctx: Arc<dyn TableContext> = self.ctx.clone();
         let file_op = Files::create(table_ctx, op);
 
+        let mut chunks = files.chunks(REMOVE_BATCH);
```
Xuanwo, line 78: s3 can remove at most 1k files in one requests. We can reduce the batch size a bit.

Xuanwo (commented):
Mostly LGTM. The only question to me is how can we test this PR?

src/query/service/src/interpreters/interpreter_user_stage_remove.rs
youngsofun: done

Xuanwo (approved):
Thanks！

PR #15394, opened 2024-05-04
fix: solve some inconsistencies between code and comments 

### LINKED ISSUE #15390: Inconsistency between code and comment
Similar to #15162, I found some inconsistency between code and comment.
In databend-main/src/common/arrow/src/arrow/array/union/mod.rs，comment should be "offset + length > self.len()"
```rust
/// # Panic
    /// This function panics iff `offset + length >= self.len()`.
    #[inline]
    pub fn slice(&mut self, offset: usize, length: usize) {
        assert!(
            offset + length <= self.len(),
            "the offset of the new array cannot exceed the existing length"
        );
        unsafe { self.slice_unchecked(offset, length) }
    }
```
In databend-main/src/common/arrow/src/arrow/array/primitive/mutable.rs, comment should be "index is larger than or equal to `self.len()"
```rust
/// # Panic
    /// Panics iff index is larger than `self.len()`.
    pub fn set(&mut self, index: usize, value: Option<T>) {
        assert!(index < self.len());
```
In databend-main/src/common/arrow/src/arrow/buffer/immutable.rs, maybe the comment should be "Panics iff `offset + length` is larger than `len`"
```rust
/// # Panics
    /// Panics iff `offset` is larger than `len`.
    #[inline]
    pub fn slice(&mut self, offset: usize, length: usize) {
        assert!(
            offset + length <= self.len(),
            "the offset of the new Buffer cannot exceed the existing length"
        );
        // Safety: we just checked bounds
        unsafe { self.slice_unchecked(offset, length) }
    }
```
and some unchecked boundary or empty constraints which are indicated in comment while aren't checked in code.For example, in databend-main/src/common/arrow/src/arrow/io/ipc/read/file.rs, the code should check whether index is out of bounds before using it directly.
```rust
/// # Panics
/// This function panics iff `index >= metadata.blocks.len()`
#[allow(clippy::too_many_arguments)]
pub fn read_batch<R: Read + Seek>(
    reader: &mut R, dictionaries: &Dictionaries, metadata: &FileMetadata, projection: Option<&[usize]>, limit: Option<usize>, index: usize, message_scratch: &mut Vec<u8>, data_scratch: &mut Vec<u8>,) -> Result<Chunk<Box<dyn Array>>> {
    let block = metadata.blocks[index];
```rust
...

### PR DESCRIPTION
**Pull request by YichiZhang0613**

## Summary

Solve the inconsistencies between code and comments which are proposed in #15390

- Fixes #15390

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15394)

### DIFF
--- a/src/common/arrow/src/arrow/bitmap/utils/mod.rs
+++ b/src/common/arrow/src/arrow/bitmap/utils/mod.rs
@@ -68,7 +68,6 @@ pub fn set(byte: u8, i: usize, value: bool) -> u8 {
 /// panics if `i >= data.len() / 8`
 #[inline]
 pub fn set_bit(data: &mut [u8], i: usize, value: bool) {
-    assert!(i < data.len() / 8);
     data[i / 8] = set(data[i / 8], i % 8, value);
 }
 
--- a/src/common/arrow/src/arrow/bitmap/utils/mod.rs
+++ b/src/common/arrow/src/arrow/bitmap/utils/mod.rs
@@ -85,7 +85,6 @@ pub unsafe fn set_bit_unchecked(data: &mut [u8], i: usize, value: bool) {
 /// This function panics iff `i / 8 >= bytes.len()`
 #[inline]
 pub fn get_bit(bytes: &[u8], i: usize) -> bool {
-    assert!(i / 8 < bytes.len());
     is_set(bytes[i / 8], i % 8)
 }
 
--- a/src/common/arrow/src/arrow/bitmap/utils/mod.rs
+++ b/src/common/arrow/src/arrow/bitmap/utils/mod.rs
@@ -68,6 +68,7 @@ pub fn set(byte: u8, i: usize, value: bool) -> u8 {
 /// panics if `i >= data.len() / 8`
 #[inline]
 pub fn set_bit(data: &mut [u8], i: usize, value: bool) {
+    assert!(i < data.len() / 8);
     data[i / 8] = set(data[i / 8], i % 8, value);
 }
 
@@ -85,6 +86,7 @@ pub unsafe fn set_bit_unchecked(data: &mut [u8], i: usize, value: bool) {
 /// This function panics iff `i / 8 >= bytes.len()`
 #[inline]
 pub fn get_bit(bytes: &[u8], i: usize) -> bool {
+    assert!(i / 8 < bytes.len());
     is_set(bytes[i / 8], i % 8)
 }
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/common/arrow/src/arrow/bitmap/utils/mod.rs [hunk 1]
```diff
@@ -68,6 +68,7 @@ pub fn set(byte: u8, i: usize, value: bool) -> u8 {
 /// panics if `i >= data.len() / 8`
 #[inline]
 pub fn set_bit(data: &mut [u8], i: usize, value: bool) {
+    assert!(i < data.len() / 8);
```
Dousir9, line 71: I think we are not concerned about panicking in this situation, The bounds used are safe.

src/common/arrow/src/arrow/bitmap/utils/mod.rs
```diff
@@ -68,6 +68,7 @@ pub fn set(byte: u8, i: usize, value: bool) -> u8 {
 /// panics if `i >= data.len() / 8`
```
Dousir9, line 68: Can you clarify what the inconsistency is between the documentation and implementation ?

src/common/arrow/src/arrow/io/ipc/read/file.rs
```diff
@@ -288,6 +288,7 @@ pub fn read_batch<R: Read + Seek>(
     message_scratch: &mut Vec<u8>,
     data_scratch: &mut Vec<u8>,
 ) -> Result<Chunk<Box<dyn Array>>> {
+    assert!(index < metadata.blocks.len());
```
Dousir9, line 291: Ditto.

Dousir9:
I think assert in this PR is unnecessary as Rust automatically checks bounds on array accesses (and panic's if it is out of bounds).

Dousir9:
> > I think assert in this PR is unnecessary as Rust automatically checks bounds on array accesses (and panic's if it is out of bounds).
> > In my opinion, the Rust official documentation recommends that to prevent array index out-of-bounds errors, it is recommended to use methods such as assert! for verification, or use built-in methods such as get and insert instead of directly accessing elements. This approach can handle potential errors more safely.
> > In addition, conscious processing can prompt meaningful information during panic, thus achieving better user interaction.
> > What's more, many other methods in databend also have assert! checks for out-of-bounds situations, which can also better maintain consistency of whole project.

We can add some comments for better user interaction.

Dousir9 (approved):
Thanks for your contribution.

PR #15412, opened 2024-05-06
fix(storage): change tracking with alias update row_version bug in merge into

zhyass:

## Summary

The bug fixed by this PR:

```sql
mysql> create table t3(a int, b int) cluster by(a+1) change_tracking=true ;
Query OK, 0 rows affected (0.15 sec)

mysql> insert into t3 values(1, 1), (3, 3), (2, 3);
Query OK, 3 rows affected (0.20 sec)

mysql> create table t4(a int, b int, c int) ;

mysql> insert into t4 values(0, 1, 0), (3, 4, 3), (4, 5, 4) ;
Query OK, 3 rows affected (0.17 sec)

mysql> merge into t3 tt using t4 on tt.a = t4.a when matched and t4.a = 4 then delete when matched then update set tt.b = t4.c when not matched then insert values(t4.a, t4.c) ;
ERROR 1105 (HY000): Internal. Code: 1001, Text = row_version It's a bug.
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15412)

### DIFF
--- a/src/query/sql/src/planner/binder/merge_into.rs
+++ b/src/query/sql/src/planner/binder/merge_into.rs
@@ -365,6 +365,9 @@ impl Binder {
             columns_set = columns_set.union(&join_column_set).cloned().collect();
         }
 
+        // If the table alias is not None, after the binding phase, the bound columns will have
+        // a database of 'None' and the table named as the alias. 
+        // Thus, we adjust them accordingly.
         let target_name = if let Some(target_identify) = target_alias {
             normalize_identifier(&target_identify.name, &self.name_resolution_ctx)
                 .name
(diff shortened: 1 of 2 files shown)


src/query/sql/src/planner/binder/merge_into.rs
```diff
@@ -365,13 +365,25 @@ impl Binder {
             columns_set = columns_set.union(&join_column_set).cloned().collect();
         }
 
+        let target_name = if let Some(target_identify) = target_alias {
```
dantengsky, line 368:
```suggestion
        // If the table alias is not None, after the binding phase, the bound columns will have
        // a database of 'None' and the table named as the alias. 
        // Thus, we adjust them accordingly.
        let target_name = if let Some(target_identify) = target_alias {
```

dantengsky (approved):

PR #15436, opened 2024-05-08
refactor: set enable_experimental_rbac_check default 1

TCeason:

## Summary

set enable_experimental_rbac_check default 1

In the default setting of enable_experimental_rbac_check

User can drop udf/stage that is owned by themself role.

Note:

Because of the flag `if exists`, need to get udf/stage in privilege_access.rs .

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15436)

### DIFF
--- a/src/query/service/src/interpreters/access/privilege_access.rs
+++ b/src/query/service/src/interpreters/access/privilege_access.rs
@@ -35,6 +35,7 @@ use databend_common_sql::plans::PresignAction;
 use databend_common_sql::plans::RewriteKind;
 use databend_common_sql::Planner;
 use databend_common_users::RoleCacheManager;
+use databend_common_users::UserApiProvider;
 
 use crate::interpreters::access::AccessChecker;
 use crate::sessions::QueryContext;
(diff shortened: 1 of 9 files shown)


src/query/service/src/interpreters/access/privilege_access.rs
```diff
@@ -618,13 +619,49 @@ impl AccessChecker for PrivilegeAccess {
...
+                if enable_experimental_rbac_check {
+                    let udf = HashSet::from([udf_name]);
+                    self.validate_udf_access(udf).await?;
+                } else {
+                    self.validate_access(&GrantObject::Global, UserPrivilegeType::Drop)
+                                        .await?;
+                }
+            }
+            Plan::DropStage(plan) => {
+                match UserApiProvider::instance().get_stage(&tenant, &plan.name).await {
```
TCeason, line 642: The user stage or tempory stage can get by this API? cc @flaneur2020 @youngsofun

flaneur2020: if i understands correctly, the user stage is not allowed to be dropped

flaneur2020:
If there are some legacy stages, they would have some risks not being able to access that stage after this PR gets merged.

However, I think this change is necessary. We can have a standby with potential customer feedbacks and providing solutions to move forward on it.

flaneur2020 (approved):

src/query/service/src/interpreters/access/privilege_access.rs
youngsofun:
```
root@localhost:8000/default> drop stage ~;
error: APIError: ResponseError with 2506: user stage is not allowed to be dropped

```

youngsofun: > The user stage or tempory stage can get by this API? cc @flaneur2020 @youngsofun

yes, I think

youngsofun: what is `tempory stage`?

youngsofun:
> If there are some legacy stages, they would have some risks not being able to access that stage after this PR gets merged.
> 
> However, I think this change is necessary. We can have a standby with potential customer feedbacks and providing solutions to move forward on it.

this pr only affect drop, for legacy stages, user can  set=enable_experimental_rbac_check=0 and then drop?

src/query/service/src/interpreters/access/privilege_access.rs
TCeason: > what is `tempory stage`?

is_temporary = true;

pub struct StageInfo {
    pub stage_name: String,
    pub stage_type: StageType,
    // on `COPY INTO xx FROM 's3://xxx?ak=?&sk=?'`, the URL(ExternalLocation) will be treated as an temporary stage.
    pub is_temporary: bool,

youngsofun (approved):

src/query/service/src/interpreters/access/privilege_access.rs
youngsofun: @TCeason  no, `tempory stage` are not stored in meta server, and not accessible outside .

TCeason: > what is `tempory stage`?

is_temporary = true;

pub struct StageInfo {
    pub stage_name: String,
    pub stage_type: StageType,
    // on `COPY INTO xx FROM 's3://xxx?ak=?&sk=?'`, the URL(ExternalLocation) will be treated as an temporary stage.
    pub is_temporary: bool,

youngsofun:
need to update the tests? @TCeason

TCeason:
> this pr only affect drop, for legacy stages, user can set=enable_experimental_rbac_check=0 and then drop?

Yes. Or grant ownership to user owned role

PR #15478, opened 2024-05-11
feat: optimize vacuum drop table, execute in parallel

lichuang:

## Summary

feat: optimize vacuum drop table, execute in parallel

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15478)

### DIFF
--- a/src/query/ee/src/storages/fuse/operations/vacuum_drop_tables.rs
+++ b/src/query/ee/src/storages/fuse/operations/vacuum_drop_tables.rs
@@ -15,120 +15,143 @@
 use std::sync::Arc;
 use std::time::Instant;
 
+use databend_common_base::runtime::execute_futures_in_parallel;
 use databend_common_catalog::table::Table;
 use databend_common_exception::Result;
-use databend_common_meta_app::schema::TableInfo;
 use databend_common_storages_fuse::FuseTable;
 use databend_enterprise_vacuum_handler::vacuum_handler::VacuumDropFileInfo;
 use futures_util::TryStreamExt;
 use log::info;
 use opendal::EntryMode;
 use opendal::Metakey;
-use opendal::Operator;
 
 #[async_backtrace::framed]
 pub async fn do_vacuum_drop_table(
-    table_info: &TableInfo,
-    operator: &Operator,
+    tables: Vec<Arc<dyn Table>>,
     dry_run_limit: Option<usize>,
 ) -> Result<Option<Vec<VacuumDropFileInfo>>> {
-    let dir = format!("{}/", FuseTable::parse_storage_prefix(table_info)?);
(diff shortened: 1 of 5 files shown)


src/query/ee/src/storages/fuse/operations/vacuum_drop_tables.rs
```diff
@@ -15,120 +15,143 @@
...
+            threads_nums * 2,
+            "batch-vacuum-drop-tables-worker".to_owned(),
+        )
+        .await?;
+        if dry_run_limit.is_some() {
+            let mut ret_files = vec![];
+            for file in result {
+                if let Ok(Some(files)) = file {
+                    ret_files.extend(files);
+                }
```
dantengsky, line 139: This ignores any exceptions that might occur during the executions of `do_vacuum_drop_table`, which is inconsistent with the previous behavior, should we consider throwing the exceptions?

dantengsky:
besides:

- cargo clippy complaints
   https://github.com/datafuselabs/databend/actions/runs/9041307782/job/24846563784?pr=15478#step:4:4786
   
- with this PR, the meaning of the parameter `dry_runt_limit` will change

the number of files returned might be slightly larger than dry_runt_limit. we either need to update the documentation accordingly or adjust the list of files that are returned (at the end of execution).

src/query/ee/src/storages/fuse/operations/vacuum_drop_tables.rs
lichuang:
```                
// return error if any errors happens during `do_vacuum_drop_table`
                if let Err(e) = file {
                    return Err(e);
                } else if let Ok(Some(files)) = file {
                    ret_files.extend(files);
                }
            }
```

src/query/ee/src/storages/fuse/operations/vacuum_drop_tables.rs
```diff
@@ -28,108 +29,133 @@ use opendal::Operator;
...
+        .await?;
+        if dry_run_limit.is_some() {
+            let mut ret_files = vec![];
+            for file in result {
+                // return error if any errors happens during `do_vacuum_drop_table`
+                if let Err(e) = file {
+                    return Err(e);
+                } else if let Ok(Some(files)) = file {
+                    ret_files.extend(files);
+                }
```
dantengsky, line 146:
```suggestion
                if let Some(files) = file? {
                    ret_files.extend(files);
                }
```

dantengsky (approved):

PR #15490, opened 2024-05-12
chore: recluster disable sort spill

zhyass:

## Summary

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15490)


BohuTANG:
I think sort spill not need this setting, but recluster need a setting to disable spill. Because sort spill not for recluster.

zhyass:
> f we disable sort spill, the recluster will be OOM? If so the root cause may not to disable spill, but how to avoid spill when recluster?

Yes, sort spill will disabled when recluster, others are not affected. 

We scaled down the default value of recluster_block_size to minimize the impact of oom.

sundy-li (approved):

PR #15494, opened 2024-05-13
feat(query): add python udf script runtime

sundy-li:

## Summary

1. add python script runtime
2. disable bendpy cause it's conflicted with python script runtime

Now we don't support python udf by default features.

We can enable it by using:
```
cargo build --bin databend-query --features "default,python-udf"
```

And the binary will contains the `libpython3.12.so.1.0` link:

```
root@3d7690dd6269:/workspace# ldd ./target/debug/databend-query
        linux-vdso.so.1 (0x00007ffca2b96000)
        libpython3.12.so.1.0 => /usr/local/lib/libpython3.12.so.1.0 (0x0000787000993000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007870007c6000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007870007a9000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000078700078f000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x000078700076d000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000787000629000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x0000787000621000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000078700044d000)
        /lib64/ld-linux-x86-64.so.2 (0x000078701e532000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x0000787000448000)
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15494)


sundy-li:
closing because we don't want to have `/usr/lib/libpython3.12.so.1.0` dynamic link in binary

src/query/service/Cargo.toml
```diff
@@ -14,7 +14,7 @@ test = true
 [features]
 default = ["simd", "z3-prove"]
 simd = ["databend-common-arrow/simd"]
-
+ee = ["arrow-udf-python"]
```
everpcpc, line 17: `ee` -> `udf-python`

sundy-li: There will be more ee features I think.

BohuTANG: Umm, this `ee` is only a name of the rust feature not Databend EE feature? Databend EE feature is protected by the license `check_enterprise_enabled ` like: https://github.com/datafuselabs/databend/blob/a7bc773fc5192d695f7d01373ce0f158972fc857/src/query/service/src/interpreters/interpreter_vacuum_temporary_files.rs#L58

sundy-li: we can change it to be a better name.

But it's different with license ee, we can't control it in runtime.
We must decide it druing compile.

everpcpc: Building feature and the runtime EE feature are two different things. People can build with `udf` features, but a license is required to use it. We could make sure to ship a full featured distribution by docker image only.

.github/workflows/bindings.python.yml
```diff
@@ -1,18 +1,20 @@
...
-        type: string
+  ## uncomment it when bendpy is enabled
+  workflow_dispatch:
+  # pull_request:
+  #   branches:
+  #     - main
+  #   paths:
+  #     - "src/**"
+  #     - ".github/workflows/bindings.python.yml"
+  # workflow_call:
```
everpcpc, line 12: It will break nightly release to comment out this.

everpcpc (approved):

tests/sqllogictests/suites/udf_native/03_0001_udf_py.test
```diff
@@ -0,0 +1,16 @@
+## enable it when compiled with ee feature
+## statement ok
```
b41sh, line 2: We don't need these tests?

sundy-li: we don't enable python udf by default, so we comment these for now.

b41sh (approved):

BohuTANG:
Could we add some logic tests for the Python script to prevent it from failing unexpectedly? This would be crucial for users running the python script on the production. cc @sundy-li @everpcpc

PR #15531, opened 2024-05-15
feat(query): add query query_hash and query_parameterized_hash to query log

### LINKED ISSUE #15524: feat: add query query_hash and query_parameterized_hash to query log
**Summary**

To categorize and examine similar queries in the query history, utilize a hash of the query text. 

## query_hash

The `query_hash` ensures that repeated queries with variations only in white space or comments share the same hash. 
For example:
```sql
SELECT * FROM t1 WHERE name = 'jim'
SELECT *  FROM t1 WHERE name  = 'jim'
```
Both have the same `query_hash`.

## query_parameterized_hash

The `query_parameterized_hash` is used when literals are part of a comparison predicate using operators such as `=`, `!=`, `>=`, or `<=`. Despite different literals, these queries share the same hash:
```sql
SELECT * FROM t1 WHERE name = 'data'
SELECT * FROM t1 WHERE name = 'bend'
```

Ref: https://docs.snowflake.com/en/user-guide/query-hash

TCeason:

## Summary

To categorize and examine similar queries in the query history, utilize a hash of the query text. 

## query_hash

The `query_hash` ensures that repeated queries with variations only in white space or comments share the same hash. 
For example:
```sql
SELECT * FROM t1 WHERE name = 'jim'
SELECT *  FROM t1 WHERE name  = 'jim'
```
Both have the same `query_hash`.

## query_parameterized_hash

The `query_parameterized_hash` is used when literals are part of a comparison predicate using operators such as `=`, `!=`, `>=`, or `<=`. Despite different literals, these queries share the same hash:
```sql
SELECT * FROM t1 WHERE name = 'data'
SELECT * FROM t1 WHERE name = 'bend'
```

- Fixes #15524

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15531)


tests/sqllogictests/suites/base/01_system/01_0002_system_query_log.test
```diff
@@ -42,6 +42,27 @@ select count(*) from tbl_01_0002 where a > 1
 # ----
 # 1
 
+statement ok
+select * from tbl_01_0002 where a=1;
```
BohuTANG, line 46: We should add a query(not pretty format) like:
```
select * from tbl_01_0002   where a = 1;
```
To make sure the query_hash are same.

TCeason: Added

src/query/storages/system/src/query_log_table.rs
```diff
@@ -25,6 +29,8 @@ use databend_common_expression::TableSchemaRefExt;
 use serde::Serialize;
 use serde::Serializer;
 use serde_repr::Serialize_repr;
+use sha2::Digest;
+use sha2::Sha256;
```
BohuTANG, line 33: No need sha256, it's too long, md5 is ok.

TCeason: Our query cache use sha generate key. Maybe need same ?

TCeason: Done

TCeason:
cc @everpcpc system table query_log add two column in this pr.

Please review format_ast part cc @b41sh @andylokandy

src/query/storages/system/src/query_log_table.rs
```diff
@@ -367,6 +376,27 @@ impl SystemLogElement for QueryLogElement {
...
+        let query_hash = format!("{:x}", Md5::digest(&self.query_text));
+        columns
+            .next()
+            .unwrap()
+            .push(Scalar::String(query_hash).as_ref());
+        let tokens = tokenize_sql(&self.query_text)?;
+        let query_parameterized_hash = parse_sql(&tokens, Dialect::PostgreSQL)
+            .ok()
+            .and_then(|(stmt, _)| {
+                if let Statement::Query(_) = stmt {
```
TCeason, line 388: in stream load, the insert sql_text maybe like this : query_text: INSERT INTO products (id, name, description) VALUES (?,?,?)

If try to format this statement, will return None in format_statement pop:

https://github.com/datafuselabs/databend/blob/05e7d49be157a6aad8ef4549d4164e840e5e204b/src/query/ast/src/ast/format/ast_format.rs#L28

So I think we can only record the select query in here.

But also I think we can modify the format_statement like this:

```rust
pub fn format_statement(stmt: Statement, ignore_literal: bool) -> Result<String> {
    let mut visitor = AstFormatVisitor::new(ignore_literal);
    visitor.visit_statement(&stmt);
    match visitor.children.pop() {
        Some(format_ctx) => format_ctx.format_pretty(),
        None => Err(ErrorCode::SyntaxException(format!("Format stmt {} None", stmt)))
    }
}

```

cc @b41sh @andylokandy

src/query/ast/src/ast/format/ast_format.rs
```diff
@@ -22,8 +22,8 @@ use itertools::Itertools;
 
 use crate::ast::*;
 
-pub fn format_statement(stmt: Statement) -> Result<String> {
-    let mut visitor = AstFormatVisitor::new();
+pub fn format_statement(stmt: Statement, ignore_literal: bool) -> Result<String> {
+    let mut visitor = AstFormatVisitor::new(ignore_literal);
```
b41sh, line 26: `format_statement` is used to generate the tree structure, here just used to rewrite the literal. We can write a separate `Visitor` to rewrite literal, and then converted to string.

src/query/storages/system/src/query_log_table.rs
```diff
@@ -367,6 +376,27 @@ impl SystemLogElement for QueryLogElement {
             .next()
             .unwrap()
             .push(Scalar::String(self.query_text.clone()).as_ref());
+        let query_hash = format!("{:x}", Md5::digest(&self.query_text));
```
b41sh, line 379: We can add `query_hash` and `query_parameterized_hash` to `QueryContext`, it can be easier to reuse, and at the same time we can use the Statement that have already been parsed and don't need to re-parse them again.

src/query/storages/system/src/query_log_table.rs
b41sh: we can ignore values clause here.

src/query/ast/src/ast/format/ast_format.rs
TCeason: Done

src/query/storages/system/src/query_log_table.rs
TCeason: Done

src/query/storages/system/src/query_log_table.rs
TCeason: Refactor in new commit

src/query/service/src/interpreters/interpreter.rs
```diff
@@ -228,10 +234,31 @@ fn log_query_finished(ctx: &QueryContext, error: Option<ErrorCode>, has_profiles
...
+
+            impl AstVisitor {
+                fn enter_literal(&mut self, lit: &mut Literal) {
+                    *lit = Literal::Null;
+                }
+            }
 
-    if result.is_err() {
+            let mut visitor = AstVisitor {};
+            stmt.drive_mut(&mut visitor);
```
b41sh, line 254: This code would be better moved to a separate function implementation, which would look cleaner, add some comments by the way will make it more readable.

src/query/service/src/interpreters/interpreter.rs
```diff
@@ -228,10 +234,31 @@ fn log_query_finished(ctx: &QueryContext, error: Option<ErrorCode>, has_profiles
 pub async fn interpreter_plan_sql(ctx: Arc<QueryContext>, sql: &str) -> Result<(Plan, PlanExtras)> {
     let mut planner = Planner::new(ctx.clone());
     let result = planner.plan_sql(sql).await;
+    let short_sql = short_sql(sql.to_string());
```
b41sh, line 237: `short_sql` is used only there is an error.

b41sh (approved):

PR #15533, opened 2024-05-15
fix(query): fix view flaky test with different dialect

b41sh:

## Summary

fix view flaky test with different dialect.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15533)

### DIFF
--- a/src/query/sql/src/planner/binder/table.rs
+++ b/src/query/sql/src/planner/binder/table.rs
@@ -42,6 +42,7 @@ use databend_common_ast::ast::TimeTravelPoint;
 use databend_common_ast::ast::UriLocation;
 use databend_common_ast::parser::parse_sql;
 use databend_common_ast::parser::tokenize_sql;
+use databend_common_ast::parser::Dialect;
 use databend_common_catalog::catalog_kind::CATALOG_DEFAULT;
 use databend_common_catalog::plan::ParquetReadOptions;
 use databend_common_catalog::plan::StageTableInfo;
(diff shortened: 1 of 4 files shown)


src/query/sql/src/planner/binder/table.rs
```diff
@@ -329,7 +330,8 @@ impl Binder {
                     .get(QUERY)
                     .ok_or_else(|| ErrorCode::Internal("Invalid VIEW object"))?;
                 let tokens = tokenize_sql(query.as_str())?;
-                let (stmt, _) = parse_sql(&tokens, self.dialect)?;
```
TCeason, line 332: I'm not sure it's a good way to fix the flaky test. if sql_dialect will cause the test err, better to add `set sql_dialect='PostgreSQL'` in the begin of test?

tests/sqllogictests/suites/base/05_ddl/05_0022_ddl_desc_view.test
```diff
@@ -69,12 +69,11 @@ e DATE NO '1970-01-01' (empty)
 f VARCHAR NO '' (empty)
 
 
-## FLAKY TEST BY dropped table
```
TCeason, line 72: Maybe we can add a set session setting in here?

**\"table\" AS table_name,** table is a keyward. so we need add a quote.

statement ok
set sql_dialect='mysql';

TCeason: It's ok now. Let's drop this double quote

TCeason (approved):

PR #15541, opened 2024-05-16
feat: select support consume stream

zhyass:

## Summary

**Key Features**

1. Added `WITH CONSUME` clause support for stream.

2. Allows execution of `SELECT xxx FROM s_name WITH CONSUME xxx` to consume data.

3. Stream data is consumed as part of the query execution, making it easier to manage and process stream data within SQL queries.

**NOTE:**

Please note that streams marked with "WITH CONSUME" are only guaranteed to be consumed if the query completes successfully.

**Example Usage**
```
mysql> create table t(a int);
Query OK, 0 rows affected (0.14 sec)

Query OK, 2 rows affected (0.14 sec)

mysql> create stream s on table t append_only=false;

Query OK, 1 row affected (0.12 sec)

mysql> delete from t where a=1;
Query OK, 1 row affected (0.21 sec)

mysql> select * from s;
+------+---------------+----------------------------------------+------------------+
| a    | change$action | change$row_id                          | change$is_update |
+------+---------------+----------------------------------------+------------------+
|    1 | DELETE        | 842124d0a6d94111b215bfe7973b5c2d000000 |                0 |
|    3 | INSERT        | e5009676789b4584b0e18f4fbb660ca3000000 |                0 |
+------+---------------+----------------------------------------+------------------+
2 rows in set (0.36 sec)
Read 4 rows, 437.00 B in 0.099 sec., 40.56 rows/sec., 4.33 KiB/sec.

mysql> select a from s with consume as ss where ss.change$action = 'INSERT';
+------+
| a    |
+------+
|    3 |
+------+
1 row in set (0.31 sec)
Read 4 rows, 437.00 B in 0.057 sec., 69.83 rows/sec., 7.45 KiB/sec.

mysql> select * from s;
Empty set (0.16 sec)
Read 0 rows, 0.00 B in 0.008 sec., 0 rows/sec., 0.00 B/sec.

mysql> select * from t with consume;
ERROR 1105 (HY000): StorageUnsupported. Code: 3902, Text = WITH CONSUME only support in STREAM.
```

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15541)


src/query/service/src/interpreters/interpreter_select.rs
```diff
@@ -125,13 +127,29 @@ impl SelectInterpreter {
...
             &self.ctx,
             &self.bind_context.columns,
             &physical_plan,
             self.ignore_result,
         )
-        .await
+        .await?;
+
+        // consume stream
+        if let Some(req) = build_update_multi_stream_meta_req(&self.ctx, &self.metadata).await? {
```
BohuTANG, line 139: It's better to add a `info!` log for this consume, to make it easy to track the action of a action.

zhyass: > It's better to add a `info!` log for this consume, to make it easy to track the action of a action.

Thanks, added.

dantengsky (approved):

BohuTANG:
@soyeric128 please update this feature to our doc, thanks.

PR #15558, opened 2024-05-17
fix: http handler kill not working.

youngsofun:

## Summary

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/service/src/servers/http/v1/query/http_query_manager.rs
+++ b/src/query/service/src/servers/http/v1/query/http_query_manager.rs
@@ -29,7 +29,6 @@ use databend_common_config::InnerConfig;
 use databend_common_exception::ErrorCode;
 use databend_common_exception::Result;
 use databend_storages_common_txn::TxnManagerRef;
-use log::warn;
 use parking_lot::Mutex;
 use tokio::task;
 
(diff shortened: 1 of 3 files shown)


src/query/service/src/servers/http/v1/query/http_query_manager.rs
```diff
@@ -177,26 +170,24 @@ impl HttpQueryManager {
...
-            let to_evict = not_removed_yet
-                .then(|| {
-                    let mut queue = self.removed_queries.lock();
-                    queue.push(q.id.to_string())
-                })
-                .flatten();
-            if let Some(qid) = to_evict {
-                self.queries.remove(&qid);
+        if let Some(q) = &query {
+            if q.mark_removed(reason) {
```
sundy-li, line 182: do we need to call `session.force_kill_query` here?

youngsofun: it is called in `Executor::stop`

zhang2014 (approved):

PR #15566, opened 2024-05-17
feat: fuzzy table data disk cache key reload

dantengsky:

## Summary

During query node restart, if the config item `data_cache_key_reload_policy` is set to "fuzzy", disk cache keys will be reloaded from the cache directory instead of directly removing previous cache data. This means that the cache data existing before the restart will NOT be deleted.

To accelerate the reloading,  sub directories of cache directory, will be processed in parallel. 

Note that during the reloading of cache keys, cache capacity will NOT be checked. Therefore, if `cache.disk.max_bytes` is decreased between restarts, no cached items on disk will be removed immediately. Instead, items will be removed when the first new item is put into the cache.

If `data_cache_key_reload_policy` is set to "reset", cached data will be remove (in parallel) during restart.

New config item introduced:

~~~
[cache]

# Policy of data cache key reloading
#
# Available options: [reset|fuzzy]
# "reset":  remove previous data cache during restart  (default value)
# "fuzzy":  reload cache keys from cache dir, retaining the cache data
#           that existed before the restart

data_cache_key_reload_policy = "reset"
~~~

please note that if storage type is `fs` (or storage that support blocking io),  disk cache will NOT be populated at all.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test: manually tested 

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15566)

### DIFF
--- a/src/query/storages/common/cache/src/providers/disk_cache.rs
+++ b/src/query/storages/common/cache/src/providers/disk_cache.rs
@@ -31,6 +31,7 @@ use databend_common_cache::LruCache;
 use databend_common_exception::ErrorCode;
 use databend_common_exception::Result;
 use log::error;
+use log::info;
 use log::warn;
 use parking_lot::RwLock;
 use siphasher::sip128;
@@ -69,7 +70,7 @@ impl From<&DiskCacheKey> for PathBuf {
 }
 
 impl<C> DiskCache<C>
-where C: Cache<String, u64, DefaultHashBuilder, FileSize>
+where C: Cache<String, u64, DefaultHashBuilder, FileSize> + Send + Sync + 'static
 {
     /// Create an `DiskCache` with `hashbrown::hash_map::DefaultHashBuilder` that stores files in `path`,
     /// limited to `size` bytes.
@@ -80,18 +81,18 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize>
     ///
     /// The cache is not observant of changes to files under `path` from external sources, it
     /// expects to have sole maintenance of the contents.
-    pub fn new<T>(path: T, size: u64) -> self::result::Result<Self>
+    pub fn new<T>(path: T, size: u64, fuzzy_reload_cache_keys: bool) -> self::result::Result<Self>
     where PathBuf: From<T> {
         DiskCache {
             cache: C::with_meter_and_hasher(size, FileSize, DefaultHashBuilder::default()),
             root: PathBuf::from(path),
         }
-        .init()
+        .init(fuzzy_reload_cache_keys)
     }
 }
 
 impl<C> DiskCache<C>
-where C: Cache<String, u64, DefaultHashBuilder, FileSize>
+where C: Cache<String, u64, DefaultHashBuilder, FileSize> + Send + Sync + 'static
 {
     /// Return the current size of all the files in the cache.
     pub fn size(&self) -> u64 {
@@ -122,7 +123,7 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize>
         self.root.join(rel_path)
     }
 
-    fn init(self) -> self::result::Result<Self> {
+    fn reset_restart(self) -> result::Result<Self> {
         // remove dir when init, ignore remove error
         if let Err(e) = fs::remove_dir_all(&self.root) {
             warn!("remove disk cache dir {:?} error {}", self.root, e);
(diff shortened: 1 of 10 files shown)


src/query/storages/common/cache/src/providers/disk_cache.rs
```diff
@@ -132,11 +133,165 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize>
         Ok(self)
     }
 
+    fn fuzzy_restart(self) -> result::Result<Self> {
+        fs::create_dir_all(&self.root)?;
+
+        let parallel_degree = std::thread::available_parallelism()
```
BohuTANG, line 139: This is the host physical CPU numbers not the docker/pod vCPU?

src/query/storages/common/cache/src/providers/disk_cache.rs
```diff
@@ -132,11 +133,165 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize>
...
+        let root = self.root.clone();
+        let cache_holder = Arc::new(RwLock::new(Some(self)));
+
+        for thread_num in 0..parallel_degree {
+            let cache_holder = cache_holder.clone();
+            let rx = rx.clone();
+            let root = root.clone();
+            let thread_builder =
+                std::thread::Builder::new().name("table-data-cache-fuzzy-restart".to_owned());
+            let handle = thread_builder.spawn(move || {
```
BohuTANG, line 186: Here why we not to use `try_spawn_batch_with_owned_semaphore` (or like) in the runtime? https://github.com/datafuselabs/databend/blob/main/src/common/base/src/runtime/runtime.rs#L251?

src/query/storages/common/cache/src/providers/disk_cache.rs
dantengsky: Either case will be OK, since the task is not CPU intensive.

src/query/storages/common/cache/src/providers/disk_cache.rs
dantengsky: semaphore is not needed here to control concurrency. it is a very simple job: launch n threads to process m directories, where  m >= n, let's keep it simple.

src/query/storages/common/cache/src/providers/disk_cache.rs
```diff
@@ -132,11 +133,165 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize>
         Ok(self)
     }
 
+    fn fuzzy_restart(self) -> result::Result<Self> {
+        fs::create_dir_all(&self.root)?;
+
+        let parallel_degree = std::thread::available_parallelism()
+            .expect("failed to detect number of parallelism")
+            .get();
```
BohuTANG, line 141:
```suggestion
let parallel_degree = match std::thread::available_parallelism() {
    Ok(degree) => degree.get(),
    Err(e) => {
        error!("Failed to detect the number of parallelism: {}", e);
        8
    }
};
```

src/query/storages/common/cache/src/providers/disk_cache.rs
```diff
@@ -122,7 +123,7 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize>
         self.root.join(rel_path)
     }
 
-    fn init(self) -> self::result::Result<Self> {
+    fn reset_restart(self) -> result::Result<Self> {
```
BohuTANG, line 126: The reset also can be parallelism?

BohuTANG: It seems OpenDAL has remove_via batch remove: https://github.com/apache/opendal/blob/08b2a96c338815208ea9b7e852034f76965ac8e6/core/src/types/operator/operator.rs#L1318

dantengsky: Absolutely, we can make that

src/query/storages/common/cache/src/providers/disk_cache.rs
```diff
@@ -123,179 +128,147 @@ where C: Cache<String, u64, DefaultHashBuilder, FileSize> + Send + Sync + 'stati
...
         info!(
-            "cache-key reload thread #{} finished, {} items processed totally, from dir {:?}",
-            thread_num, file_idx, absolute_cache_path
+            "initializing disk cache, parallelism set to {}",
+            parallelism
         );
-        Ok(())
+
+        let thread_pool = ThreadPoolBuilder::new()
+            .num_threads(parallelism) // 设置线程池中的线程数量
```
BohuTANG, line 245: The comments may by need removed?

dantengsky: thanks for reminding

dantengsky:
@BohuTANG  PTAL

BohuTANG (approved):

Issue #15295, opened 2024-04-22
Feature: Support map functions


b41sh:
**Summary**

- [x] #15337
- [x] `MAP_CONTAINS_KEY(<key>, <map>)` Determines whether the specified MAP contains the specified key. @hanxuanliang 
- [x] `MAP_DELETE(<map>, <key1>[, <key2>, ...])` Returns a MAP based on an existing MAP with one or more keys removed..
- [x] `MAP_INSERT(<map>, <key>, <value>[, <updateFlag>])` Returns a new MAP consisting of the input MAP with a new key-value pair inserted (an existing key updated with a new value).
- [x] `MAP_PICK(<map>, <key1>[, <key2>, ...])` or `MAP_PICK(<map>, <array>)` Returns a new MAP containing the specified key-value pairs from an existing MAP.
- [x] `MAP_SIZE(<map>)` Returns the size of a MAP.

Refer to the following documentations:
https://docs.snowflake.com/en/sql-reference/functions/map_cat
https://docs.snowflake.com/en/sql-reference/functions/map_contains_key
https://docs.snowflake.com/en/sql-reference/functions/map_delete
https://docs.snowflake.com/en/sql-reference/functions/map_insert
https://docs.snowflake.com/en/sql-reference/functions/map_pick
https://docs.snowflake.com/en/sql-reference/functions/map_size

shamb0:
Hi @b41sh,

I am interested in taking on the implementation of the feature. If available, I would appreciate being assigned to this task.

Thank you for your consideration.

b41sh:
Hi, @shamb0 Thanks for your attention, you can pick the functions that interest you.

hanxuanliang:
I want to try [map_contains_key](https://docs.snowflake.com/en/sql-reference/functions/map_contains_key)

forsaken628:
https://arrow.apache.org/docs/python/generated/pyarrow.MapType.html#pyarrow.MapType.keys_sorted
Is this attribute not supported? If supported, there are more efficient algorithm implementations

b41sh:
> I want to try [map_contains_key](https://docs.snowflake.com/en/sql-reference/functions/map_contains_key)

Hi @hanxuanliang  Thank you for your interest, you can work on this function.

b41sh:
> https://arrow.apache.org/docs/python/generated/pyarrow.MapType.html#pyarrow.MapType.keys_sorted Is this attribute not supported? If supported, there are more efficient algorithm implementations

[...]

PR #15567, opened 2024-05-18
feat(functions): add map_insert function

hanxuanliang:

## Summary

part of https://github.com/datafuselabs/databend/issues/15295

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/15567)


hanxuanliang:
However, there is such a description in [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/map_insert):

> If updateFlag is set to TRUE, then the existing input key is updated to the input value

But in Databend, do we support dynamically setting config in the query? Alternatively, we can set the config when querying with the bendsql?

b41sh:
> However, there is such a description in [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/map_insert):
> 
> > If updateFlag is set to TRUE, then the existing input key is updated to the input value
> 
> But in Databend, do we support dynamically setting config in the query? Alternatively, we can set the config when querying with the bendsql?

We can register a new function with four arguments for `map_insert`, for example `st_geometryfromwkb` can support two and three args.

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +246,50 @@ pub fn register(registry: &mut FunctionRegistry) {
                     .any(|(k, _)| k == key)
             },
         );
+
+    registry.register_3_arg_core::<EmptyMapType, GenericType<0>, GenericType<1>, MapType<GenericType<0>, GenericType<1>>, _, _>(
+        "map_insert",
+        |_, _, _, _| FunctionDomain::Full,
```
b41sh, line 252: we can build map domain from key domain and value domain.

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +246,50 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        |_, key, value, ctx| {
+            let mut b = ArrayType::create_builder(1, ctx.generics);
```
b41sh, line 254: The data type of key can only be `bool`, `string`, `number`, `date` and `timestamp`, we need check the key type first. Here is an example in `map` function:
https://github.com/datafuselabs/databend/blob/main/src/query/functions/src/scalars/map.rs#L52

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +246,50 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+      
+            return Value::Scalar(MapType::build_scalar(b));
+        },
+    );
+
+    registry.register_passthrough_nullable_3_arg(
+        "map_insert",
+        |_, domian1, _, _| {
+            FunctionDomain::Domain(match domian1 {
+                Some((key_domain, val_domain)) => Some((key_domain.clone(), val_domain.clone())),
```
b41sh, line 266: the `key_domain` and `value_domain` need merge with the inserted key and value domain

hanxuanliang:
But, I have discovered a problem now:

```sql
CREATE TABLE map_insert_test(col_str Map(String, String Null) Not Null, col_int Map(String, Int Null) Null);

SELECT map_insert(col_int, 'a', 100)
FROM map_insert_test;
```

For the NULL map, it is currently stuck and unable to run. However, I tried to modify the domain calculation, but it had no effect. I have no idea here 😣

b41sh:
> But, I have discovered a problem now:
> 
> ```sql
> CREATE TABLE map_insert_test(col_str Map(String, String Null) Not Null, col_int Map(String, Int Null) Null);
> 
> SELECT map_insert(col_int, 'a', 100)
> FROM map_insert_test;
> ```
> 
> For the NULL map, it is currently stuck and unable to run. However, I tried to modify the domain calculation, but it had no effect. I have no idea here 😣

Both val_domain and insert_value_domain may be nullable, and we need to deal with various cases, such as
```rust
match (val_domain, insert_value_domain) {
    (Domain::Nullable(NullableDomain {..}), Domain::Nullable(NullableDomain {..})) => ...,
    (Domain::Nullable(NullableDomain {..}), _) => ...,
    (_, Domain::Nullable(NullableDomain {..})) => ...,
    (_, _) => ...,
}
```

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,121 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        },
+        vectorize_with_builder_3_arg::<
+            MapType<GenericType<0>, GenericType<1>>,
+            GenericType<0>,
+            GenericType<1>,
+            MapType<GenericType<0>, GenericType<1>>,
+        >(|source, key, value, output, ctx| {
+            // insert operation only works on specific key type: boolean, string, numeric, decimal, date, datetime
+            let key_type = &ctx.generics[0];
+            if !key_type.is_boolean()
```
b41sh, line 291: move this check to `EmptyMap` function, the key type of map is valid, don't need to check.

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,121 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        },
+        |_, key, value, ctx| {
+            let mut b = ArrayType::create_builder(1, ctx.generics);
+            b.put_item((key.into_scalar().unwrap(), value.into_scalar().unwrap()));
+            b.commit_row();
+            return Value::Scalar(MapType::build_scalar(b));
+        },
+    );
+
+    registry.register_passthrough_nullable_3_arg(
```
b41sh, line 272: Use `register_2_arg_core` instead of `register_passthrough_nullable_3_arg`, because `register_passthrough_nullable_3_arg` will cause incorrect return values when the value is `NULL`, for example: `select map_insert({'a':'b'}, 'c', null)` should return `{'a':'b','c':NULL}` instead of `NULL`.
the function can be written as:
```rust
    registry.register_3_arg_core::<NullableType<MapType<GenericType<0>, GenericType<1>>>, GenericType<0>, GenericType<1>, _, _>
```

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,144 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+
+    registry.register_3_arg_core::<EmptyMapType, GenericType<0>, GenericType<1>, MapType<GenericType<0>, GenericType<1>>, _, _>(
+        "map_insert",
+        |_, _, insert_key_domain, insert_value_domain| {
+            FunctionDomain::Domain(Some((
+                insert_key_domain.clone(),
+                insert_value_domain.clone(),
+            )))
+        },
+        |_, key, value, ctx| {
```
b41sh, line 264: use `vectorize_with_builder_3_arg` here, so that we can support column values in this function

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,144 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            let key_type = &ctx.generics[0];
+            if !key_type.is_boolean()
+                && !key_type.is_string()
+                && !key_type.is_numeric()
+                && !key_type.is_decimal()
+                && !key_type.is_date_or_date_time()
+            {
+                ctx.set_error(0, format!("map keys can not be {}", key_type));
```
b41sh, line 272: if has error, we should return at here.

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,144 @@ pub fn register(registry: &mut FunctionRegistry) {
                     .any(|(k, _)| k == key)
             },
         );
+
+    registry.register_3_arg_core::<EmptyMapType, GenericType<0>, GenericType<1>, MapType<GenericType<0>, GenericType<1>>, _, _>(
```
b41sh, line 256: `EmptyMapType` also need support 4 arguments with `update_flag`, use `register_4_arg_core` register another function.

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,144 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+        },
+    );
+
+    registry.register_3_arg_core::<NullableType<MapType<GenericType<0>, GenericType<1>>>, GenericType<0>, GenericType<1>, MapType<GenericType<0>, GenericType<1>>, _, _>(
+        "map_insert",
+        |_, source_domain, insert_key_domain, insert_value_domain| {
+            FunctionDomain::Domain(match source_domain.has_null {
+                true => Some((
+                    insert_key_domain.clone(),
+                    insert_value_domain.clone(),
```
b41sh, line 288: `insert_key_domain` and `insert_value_domain` need merge with the key and value domain in `source_domain`.

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,144 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+                    b.commit_row();
+                    output.append_column(&b.build());
+                },
+            };
+        }),
+    );
+
+    registry.register_passthrough_nullable_3_arg(
+        "map_insert",
+        |_, domain1, key_domain, value_domain| {
```
b41sh, line 318: this function can be removed, as we already have `register_3_arg_core` function

src/query/functions/src/scalars/map.rs
```diff
@@ -244,4 +252,144 @@ pub fn register(registry: &mut FunctionRegistry) {
...
+            GenericType<1>,
+            MapType<GenericType<0>, GenericType<1>>,
+        >(|source, key, value, output, ctx| {
+            // default behavior is to insert new key-value pair, and if the key already exists, update the value.
+            output.append_column(&build_new_map(&source, key, value, ctx));
+        }),
+    );
+
+    // grammar: map_insert(map, insert_key, insert_value, allow_update)
+    registry.register_passthrough_nullable_4_arg(
```
b41sh, line 339: use `register_4_arg_core` here like the previous one.

b41sh (approved):

b41sh:
Thanks for @hanxuanliang PR. And look forward to keeping an eye on databend in the future

PR #15582, opened 2024-05-20
fix: consume stream inside explicit txn

### LINKED ISSUE #15581: bug:  `select ... from stream with consume` not work as expected in non auto commit mode
**Summary**

`select ... from stream with consume` not work as expected in non auto commit mode

how to reproduce:

the following test case should return a non-empty result set at the end, but not.

~~~
create or replace database poc_databend;

use poc_databend;


create  or replace table t_1 (
  str varchar 
);


create or replace stream s_1 on table t_1 append_only = true;


insert into t_1 (str) values ('a'), ('b');


begin;

      select * from s_1 with consume;
      select 1/0;

commit;

select * from s_1; -- expects non-empty result set but NOT
~~~

dantengsky:

## Summary

- The consuming (`with consume`) of stream should respect the semantics of transactions: if  transaction is aborted, the stream should roll back to its original state.

- Only the usage of `WITH CONSUME` in queries is allowed. Statements like `INSERT INTO t SELECT * FROM s WITH CONSUME` will fail with errors.

- Fixes #15581

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15582)

### DIFF
--- a/src/query/service/src/interpreters/common/stream.rs
+++ b/src/query/service/src/interpreters/common/stream.rs
@@ -19,7 +19,7 @@ use chrono::Utc;
 use databend_common_exception::Result;
 use databend_common_license::license::Feature;
 use databend_common_license::license_manager::get_license_manager;
-use databend_common_meta_app::schema::UpdateMultiTableMetaReq;
+use databend_common_meta_app::schema::TableInfo;
 use databend_common_meta_app::schema::UpdateStreamMetaReq;
 use databend_common_meta_app::schema::UpdateTableMetaReq;
 use databend_common_meta_types::MatchSeq;
@@ -95,10 +95,14 @@ where F: Fn(&TableEntry) -> bool {
     Ok(streams)
 }
 
+pub struct StreamTableUpdates {
+    pub update_table_metas: Vec<UpdateTableMetaReq>,
+    pub table_infos: Vec<TableInfo>,
+}
 pub async fn query_build_update_stream_req(
     ctx: &Arc<QueryContext>,
     metadata: &MetadataRef,
-) -> Result<Option<UpdateMultiTableMetaReq>> {
+) -> Result<Option<StreamTableUpdates>> {
     let streams = get_stream_table(metadata, |t| {
         t.is_consume() && t.table().engine() == STREAM_ENGINE
     })?;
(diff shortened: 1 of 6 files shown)


src/query/service/src/interpreters/common/stream.rs
```diff
@@ -129,7 +136,7 @@ pub async fn query_build_update_stream_req(
         new_table_meta.options = options;
         new_table_meta.updated_on = Utc::now();
 
-        update_table_metas.push(UpdateTableMetaReq {
+        update_table_meta_reqs.push(UpdateTableMetaReq {
```
SkyFan2002, line 139: Sometimes we use `UpdateTableMetaReq` to update a stream, and other times we use `UpdateTableMetaReq.update_stream_meta`. But both `SchemaApi::update_table_meta()` and `update_table_meta()` handle `stream` and `table` with subtle differences. Although the current implementation is logically correct, we could consider refactoring in the future to express the requests for updating a stream and updating a table with a more clear structure, to avoid potential bugs.

dantengsky: agree

tests/sqllogictests/suites/ee/06_ee_stream/06_0002_stream_txn_consume.test
```diff
@@ -288,6 +288,119 @@ query I
...
+onlyif http
+query I
+select str from s_1 order by str;
+----
+a
+b
+
+onlyif http
+query I
+select str from s_1 with consume order by str;
```
SkyFan2002, line 388: It would be better to test if repeatable read of stream works with both dml and '`select.. with consume` in a single txn.

dantengsky: good idea, let me insert a dml here

src/query/sql/src/planner/binder/binder.rs
```diff
@@ -663,6 +664,22 @@ impl<'a> Binder {
                 self.bind_set_priority(priority, object_id).await?
             },
         };
+
+        match plan.kind() {
+            QueryKind::Query { .. } | QueryKind::Explain { .. } => {}
+            _ => {
+                let meta_data_guard = self.metadata.read();
+                let tables = meta_data_guard.tables();
+                for t in tables {
```
BohuTANG, line 673: If a database has many tables, this check may break performance?

BohuTANG: This check is work?
https://github.com/datafuselabs/databend/blob/2bfe458d31feb6a067cefb3b090c02f260d0fbcb/src/query/sql/src/planner/binder/table.rs#L265-L269

dantengsky: this only involves the tables that bounded in the statement, not all the tables of a database.

tests/sqllogictests/suites/ee/06_ee_stream/06_0002_stream_txn_consume.test
dantengsky: test added, PTAL

SkyFan2002 (approved):

PR #15621, opened 2024-05-23
refactor: more explicit error confirmation for http query route error.

youngsofun:

## Summary

also including the case that server restart (assert: always with new node_id).
we can tell it from query timeout at once.

- need to test with the clients
    - [x] go 
    - [x] java
    - [x] bendsql  
    - [x] py

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15621)

### DIFF
--- a/src/query/service/src/servers/http/v1/http_query_handlers.rs
+++ b/src/query/service/src/servers/http/v1/http_query_handlers.rs
@@ -225,13 +225,14 @@ async fn query_final_handler(
     ctx: &HttpQueryContext,
     Path(query_id): Path<String>,
 ) -> PoemResult<impl IntoResponse> {
+    ctx.check_node_id(&query_id)?;
     let root = get_http_tracing_span(full_name!(), ctx, &query_id);
     let _t = SlowRequestLogTracker::new(ctx);
-
     async {
         info!(
-            "{}: got /v1/query/{}/final request, this query is going to be finally completed.",
-            query_id, query_id
+            "{}: got {} request, this query is going to be finally completed.",
+            query_id,
+            make_final_uri(&query_id)
         );
         let http_query_manager = HttpQueryManager::instance();
         match http_query_manager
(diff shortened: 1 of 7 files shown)


sundy-li (approved):

src/query/service/src/servers/http/v1/http_query_handlers.rs
```diff
@@ -53,20 +53,20 @@ const HEADER_QUERY_ID: &str = "X-DATABEND-QUERY-ID";
 const HEADER_QUERY_STATE: &str = "X-DATABEND-QUERY-STATE";
 const HEADER_QUERY_PAGE_ROWS: &str = "X-DATABEND-QUERY-PAGE-ROWS";
 
-pub fn make_page_uri(query_id: &str, page_no: usize) -> String {
-    format!("/v1/query/{}/page/{}", query_id, page_no)
+pub fn make_page_uri(query_id: &str, node_id: &str, page_no: usize) -> String {
+    format!("/v1/query/{}/{}/page/{}", query_id, node_id, page_no)
```
flaneur2020, line 57: can we pass node_id in the header?

flaneur2020 (changes requested):
it need a comprehensive test over all the drivers after api route changes, and there're also few in-house query clients (like PHP) not using our drivers at all. if it's an enhancement over the error message, i guess we can consider a less invasive way to make it.

src/query/service/src/servers/http/v1/http_query_handlers.rs
youngsofun: done

youngsofun:
@flaneur2020 plz review again

tests/suites/1_stateful/09_http_handler/09_0006_route_error.result
```diff
@@ -0,0 +1,15 @@
+# error
+## page
+{"error":{"code":"404","message":"route error: query QID SHOULD be on server XXX, but current server is NODE, which started ... ago)"}}
+## kill
+{"error":{"code":"404","message":"route error: query QID SHOULD be on server XXX, but current server is NODE, which started ... ago)"}}
+## final
+{"error":{"code":"404","message":"route error: query QID SHOULD be on server XXX, but current server is NODE, which started ... ago)"}}
```
flaneur2020, line 7: @hantmac please take a look about whether the drivers could handle this error response into a proper exception with these message displayed.

flaneur2020 (approved):
LGTM

tests/suites/1_stateful/09_http_handler/09_0006_route_error.result
youngsofun: client cannot handle it, and it is hard to explain it in the error message, current message gives info helpful for debugging.

 we can later add a section about the routing problem in the doc, and in the error message let user to refer to the section.

PR #15640, opened 2024-05-25
fix(query): delete len check in mget_database/table_names_by_ids

TCeason:

## Summary

delete some err check

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15640)

### DIFF
--- a/src/meta/api/src/schema_api_impl.rs
+++ b/src/meta/api/src/schema_api_impl.rs
@@ -190,7 +190,6 @@ use databend_common_meta_kvapi::kvapi;
 use databend_common_meta_kvapi::kvapi::DirName;
 use databend_common_meta_kvapi::kvapi::Key;
 use databend_common_meta_kvapi::kvapi::UpsertKVReq;
-use databend_common_meta_types::anyerror::AnyError;
 use databend_common_meta_types::protobuf as pb;
 use databend_common_meta_types::txn_op::Request;
 use databend_common_meta_types::txn_op_response::Response;
@@ -198,9 +197,6 @@ use databend_common_meta_types::ConditionResult;
 use databend_common_meta_types::InvalidReply;
 use databend_common_meta_types::MatchSeq;
 use databend_common_meta_types::MatchSeqExt;
-use databend_common_meta_types::MetaAPIError;
-use databend_common_meta_types::MetaDataError;
-use databend_common_meta_types::MetaDataReadError;
 use databend_common_meta_types::MetaError;
 use databend_common_meta_types::MetaId;
 use databend_common_meta_types::MetaNetworkError;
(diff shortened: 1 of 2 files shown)


src/meta/api/src/schema_api_impl.rs
```diff
@@ -2301,15 +2297,9 @@ impl<KV: kvapi::KVApi<Error = MetaError> + ?Sized> SchemaApi for KV {
         }
 
         let seq_metas = self.mget_kv(&meta_kv_keys).await?;
-        if seq_metas.len() != table_names.len() {
```
TCeason, line 2304: drop table in the databend doesn't actually drop the table, it just changes drop_on to not None in tablemeta

In grant kv, the behavior of drop table cannot be perceived. This requires filtering on show grants, resulting in the function mget_table_name_by_ids.

In func  mget_table_name_by_ids

```
table_names is generate by mget DBIdTableName. the id is function's arg table_id.

seq_metas is generated by mget TableId.
```

I assume that seq_name and table_names must be the same length because the table ids are the same. But in fact, after the added judgment, resulting in an error.

Hi @drmingdrmer xp, is there anything that might have made a difference here?

drmingdrmer: what does the flatten mean at the line 2288?

as far as I know, meet always return the exactly the same number of rows of the input keys. 
I’m not quite sure what happened. Is that possible to print out the two vector so we can see what the different is

TCeason: One word wakes up a dreamer. Instead of ignoring None, you should fill in none. Otherwise, creating an artificial inconsistency.

TCeason:
In dev warehouse system:

```
deploy@(system)/default> show grants;
error: APIError: ResponseError with 2001: fail to mget_table_names_by_ids: , source: The system is experiencing high load, please retry later
```

In test warehouse show_grants it fixed:

```sql
deploy@(show_grants)/default> select version();

SELECT
  version()

┌────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                            version()                                           │
│                                             String                                             │
├────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Databend Query v1.2.490-nightly-49fa83976d(rust-1.78.0-nightly-2024-05-25T11:22:38.392488593Z) │
└────────────────────────────────────────────────────────────────────────────────────────────────┘
1 row read in 1.515 sec. Processed 1 row, 1B (0.66 row/s, 0B/s)

deploy@(show_grants)/default> show grants;

SHOW GRANTS

┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ privileges │     object_name    │     object_id    │ grant_to │  name  │                            grants                           │
│   String   │       String       │ Nullable(UInt64) │  String  │ String │                       Nullable(String)                      │
├────────────┼────────────────────┼──────────────────┼──────────┼────────┼─────────────────────────────────────────────────────────────┤
│ ALL        │ *.*                │             NULL │ USER     │ deploy │ GRANT ALL ON *.* TO 'deploy'@'%'                            │
│ ·          │ ·                  │                · │ ·        │ ·      │ ·                                                           │
│ ·          │ ·                  │                · │ ·        │ ·      │ ·                                                           │
│ ·          │ ·                  │                · │ ·        │ ·      │ ·                                                           │
xxx                 │                  │          │        │                                                             │
│ (2 shown)  │                    │                  │          │        │                                                             │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
31 rows read in 1.387 sec. Processed 31 rows, 4.52 KiB (22.36 rows/s, 3.26 KiB/s)
```

drmingdrmer (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15640)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @TCeason)

PR #15652, opened 2024-05-28
fix(query): useridentity hostname only support char %

TCeason:

## Summary

1. username containing @ should not encode when record log or err msg

2. We disabled arbitrary hostnames when creating users last year pr https://github.com/datafuselabs/databend/pull/11818 so the user meta key can only be generated by username, and % is a default value in UserIdentity.

3. delete StageFileInfo::creator

Note:

Also, do not support localhost as hostname.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15652)

### DIFF
--- a/src/meta/app/src/principal/tenant_user_ident.rs
+++ b/src/meta/app/src/principal/tenant_user_ident.rs
@@ -29,8 +29,8 @@ impl TenantUserIdent {
         Self::new_generic(tenant, user)
     }
 
-    pub fn new_user_host(tenant: impl ToTenant, user: impl ToString, host: impl ToString) -> Self {
-        Self::new(tenant, UserIdentity::new(user, host))
+    pub fn new_user_host(tenant: impl ToTenant, user: impl ToString) -> Self {
+        Self::new(tenant, UserIdentity::new(user))
     }
 }
 
@@ -70,7 +70,7 @@ mod tests {
 
     fn test_format_parse(user: &str, host: &str, expect: &str) {
         let tenant = Tenant::new_literal("test_tenant");
-        let user_ident = UserIdentity::new(user, host);
+        let user_ident = UserIdentity::new(user);
         let tenant_user_ident = TenantUserIdent::new(tenant, user_ident);
 
         let key = tenant_user_ident.to_string_key();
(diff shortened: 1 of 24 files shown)


src/meta/app/src/principal/tenant_user_ident.rs
```diff
@@ -70,7 +70,11 @@ mod tests {
 
     fn test_format_parse(user: &str, host: &str, expect: &str) {
         let tenant = Tenant::new_literal("test_tenant");
-        let user_ident = UserIdentity::new(user, host);
+        let user_ident = UserIdentity {
```
TCeason, line 73: test_tenant_user_ident_as_kvapi_key use this function. So use struct init replace new function.

src/meta/proto-conv/tests/it/user_proto_conv.rs
```diff
@@ -806,7 +809,10 @@ fn test_old_stage_file() -> anyhow::Result<()> {
             NaiveDate::from_ymd_opt(2022, 9, 16).unwrap(),
             NaiveTime::from_hms_opt(0, 1, 2).unwrap(),
         );
-        let user_id = mt::principal::UserIdentity::new("datafuselabs", "datafuselabs.rs");
```
TCeason, line 809: In order not to change the test result, hostname is left here. But do we need to construct useridentity here with a new method and change the test result at the same time cc @youngsofun @drmingdrmer

src/meta/app/src/principal/tenant_user_ident.rs
TCeason: cc @drmingdrmer

drmingdrmer (approved):
> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15652)* status: all files reviewed, 2 unresolved discussions

b41sh (approved):

TCeason:
Some Old version use specifica host

PR #15659, opened 2024-05-28
fix(query): fix current date/time functions return constant value in a statement.

### LINKED ISSUE #15638: bug: current date/time functions should return constant value within a statement
### Search before asking

### Version

nightly

### What's Wrong?

now() returns a different value every time it's invoked, even when using an alias and referencing the same alias multiple times.

this is due to all time functions being marked as non-deterministic here: https://github.com/datafuselabs/databend/blob/main/src/query/functions/src/scalars/datetime.rs#L880

This behavior is the opposite of most (perhaps all?) other SQL databases.

## Snowflake
"Context Functions

This family of functions allows for the gathering of information about the context in which the statement is executed. These functions are **evaluated at most once per statement**."
https://docs.snowflake.com/en/sql-reference/functions-context#list-of-functions

## Postgres
“These SQL-standard functions all return values based on the start time of the current transaction:”
https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT

## MySQL
“[NOW\(\)](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_now)~ returns a constant time that indicates the time at which the statement began to execute.”
https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_now

## ClickHouse
“Time at query evaluation. Now is a constant expression”
https://clickhouse.com/docs/en/sql-reference/functions/date-time-functions#now

This behavior is causing some problematic issues when porting some ETL queries from Snowflake to Databend.

### How to Reproduce?

query:
```sql
select
    to_string(now(), '%Y/%m/%dT%H:%M:%S.%f') as ts,
    to_string(now(), '%Y/%m/%dT%H:%M:%S.%f') as ts2,
    ts as ts3,
    ts as ts4;
```

result:
| ts | ts2 | ts3 | ts4 |
| :--- | :--- | :--- | :--- |
| 2024/05/24T18:25:30.819084000 | 2024/05/24T18:25:30.819118000 | 2024/05/24T18:25:30.819131000 | 2024/05/24T18:25:30.819144000 |

Every timestamp is different. They should all be the same.

b41sh:

## Summary

`now()`, `today()`, `yesterday()` and `tomorrow()` functions return constant value in a statement.

```sql
mysql> set timezone='UTC';

mysql> select now(), now();
+----------------------------+----------------------------+
| now()                      | now()                      |
+----------------------------+----------------------------+
| 2024-05-28 09:22:57.995818 | 2024-05-28 09:22:57.995818 |
+----------------------------+----------------------------+
1 row in set (0.04 sec)
Read 1 rows, 1.00 B in 0.017 sec., 58.44 rows/sec., 58.44 B/sec.

mysql> set timezone='Asia/Shanghai';

mysql> select now(), now();
+----------------------------+----------------------------+
| now()                      | now()                      |
+----------------------------+----------------------------+
| 2024-05-28 17:23:01.717210 | 2024-05-28 17:23:01.717210 |
+----------------------------+----------------------------+
1 row in set (0.04 sec)
Read 1 rows, 1.00 B in 0.009 sec., 114.58 rows/sec., 114.58 B/sec.
```

- Fixes #15638

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15659)

### DIFF
--- a/src/query/functions/src/scalars/datetime.rs
+++ b/src/query/functions/src/scalars/datetime.rs
@@ -895,14 +895,7 @@ fn register_real_time_functions(registry: &mut FunctionRegistry) {
     registry.register_0_arg_core::<TimestampType, _, _>(
         "now",
         |_| FunctionDomain::Full,
-        |ctx| {
-            Value::Scalar(
-                ctx.func_ctx
-                    .now
-                    .with_timezone(&ctx.func_ctx.tz.tz)
-                    .timestamp_micros(),
-            )
-        },
+        |ctx| Value::Scalar(ctx.func_ctx.now.timestamp_micros()),
     );
 
     registry.register_0_arg_core::<DateType, _, _>(
--- a/src/query/functions/src/scalars/datetime.rs
+++ b/src/query/functions/src/scalars/datetime.rs
@@ -895,25 +895,32 @@ fn register_real_time_functions(registry: &mut FunctionRegistry) {
     registry.register_0_arg_core::<TimestampType, _, _>(
         "now",
         |_| FunctionDomain::Full,
-        |_| Value::Scalar(Utc::now().timestamp_micros()),
+        |ctx| {
+            Value::Scalar(
+                ctx.func_ctx
+                    .now
+                    .with_timezone(&ctx.func_ctx.tz.tz)
+                    .timestamp_micros(),
+            )
+        },
     );
 
     registry.register_0_arg_core::<DateType, _, _>(
         "today",
         |_| FunctionDomain::Full,
-        |ctx| Value::Scalar(today_date(ctx.func_ctx.tz)),
+        |ctx| Value::Scalar(today_date(ctx.func_ctx.now, ctx.func_ctx.tz)),
     );
 
     registry.register_0_arg_core::<DateType, _, _>(
         "yesterday",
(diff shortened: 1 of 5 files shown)


src/query/functions/src/scalars/datetime.rs
```diff
@@ -895,25 +895,32 @@ fn register_real_time_functions(registry: &mut FunctionRegistry) {
     registry.register_0_arg_core::<TimestampType, _, _>(
         "now",
         |_| FunctionDomain::Full,
-        |_| Value::Scalar(Utc::now().timestamp_micros()),
+        |ctx| {
+            Value::Scalar(
+                ctx.func_ctx
+                    .now
+                    .with_timezone(&ctx.func_ctx.tz.tz)
```
sundy-li, line 902: no need `.with_timezone(&ctx.func_ctx.tz.tz)` I think

andylokandy (approved):

PR #15665, opened 2024-05-28
fix: fix create table error but table id uncommttied

### LINKED ISSUE #15664: bug: create table error make alter table fail
### Search before asking

### Version

v1.2.497-nightly

### What's Wrong?

```sql
CREATE OR REPLACE TABLE test_table as select 1;
CREATE OR REPLACE TABLE test_table as select 1/0;
alter table test_table rename to test_table_1;   // will fail

select * from test_table;  // but select works!
```

### How to Reproduce?

_No response_

lichuang:

## Summary

fix: fix create table error but table id uncommttied

when create table use `as select`, it will save table id in a temporary table id list, if create table success, table id will be commit, otherwise table id\meta\s3 data will be vacuumed when `vacuum drop table`.

- Fixes #15664

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15665)


src/query/service/src/interpreters/interpreter_table_create.rs
```diff
@@ -273,9 +274,28 @@ impl CreateTableInterpreter {
...
+                            table_id,
+                            table_name,
+                        };
+                        catalog.rollback_uncommitted_table_meta(drop_by_id).await
+                    };
+
+                    GlobalIORuntime::instance().block_on(fut).map_err(|e| {
+                        info!("create {} as select failed. {:?}", qualified_table_name, e);
+                        e
+                    })?;
```
dantengsky, line 295: if process crashed before `rollback_uncommitted_table_meta` successfully executed, will things still work as expected?

dantengsky:
@lichuang 

maybe, we could also consider tweaking the following snippet of ` SchemaApi for KV::rename_table`?

instead of assuming that `last_table_id == table_id` and pop it out,  find the table_id and remove it from `tb_id_list`?

https://github.com/datafuselabs/databend/blob/3109054a4dfea22009026479bd9a21d87eff13e3/src/meta/api/src/schema_api_impl.rs#L1876-L1892

drmingdrmer:
> @lichuang
> 
> maybe, we could also consider tweaking the following snippet of ` SchemaApi for KV::rename_table`?
> 
> instead of assuming that `last_table_id == table_id` and pop it out, find the table_id and remove it from `tb_id_list`?
> 
> https://github.com/datafuselabs/databend/blob/3109054a4dfea22009026479bd9a21d87eff13e3/src/meta/api/src/schema_api_impl.rs#L1876-L1892

Table list is a stack which is responsible to pop up the last removed table.

dantengsky:
For your reference: 

another way to resolve issue #15664

https://github.com/datafuselabs/databend/pull/15676

PR #15693, opened 2024-05-31
fix(executor): fix lost pipeline on finished callback

### LINKED ISSUE #15649: bug: lost query (abort) log when query is kill by http handler.
**Summary**

https://github.com/datafuselabs/databend/blob/main/tests/suites/1_stateful/09_http_handler/09_0005_kill.sh#L16

all seen cases are in cluster mode

zhang2014:

## Summary

fix(executor): fix lost pipeline on finished callback

- Fixes #15649 

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15693)


src/query/pipeline/core/src/finished_chain.rs
```diff
@@ -0,0 +1,388 @@
...
+                    location,
+                    is_always: false,
+                    successfully: false,
+                    elapsed: instant.elapsed(),
+                });
+
+                info.res = Err(cause.clone());
+                Self::apply_always(info, states, always_callbacks);
+
+                return Err(cause);
```
youngsofun, line 100: logging the locations of callbacks that have not run may help debug too.
otherwise, some developers may be unaware of its existence, making it difficult to associate the program’s erroneous behavior with it.

src/query/pipeline/core/src/finished_chain.rs
```diff
@@ -0,0 +1,388 @@
...
+                Ok(_) => ApplyState {
+                    location,
+                    is_always: true,
+                    successfully: true,
+                    elapsed: instant.elapsed(),
+                },
+                Err(_cause) => ApplyState {
+                    location,
+                    is_always: true,
+                    successfully: false,
```
youngsofun, line 132: maybe not useful for now, but the cause can be recorded?

zhang2014: The cause may have been recorded in the log.

src/query/pipeline/core/src/finished_chain.rs
```diff
@@ -0,0 +1,388 @@
...
+    fn apply(self: Box<Self>, info: &ExecutionInfo) -> Result<()> {
+        match catch_unwind(move || self(info)) {
+            Ok(Ok(_)) => Ok(()),
+            Err(cause) => Err(cause),
+            Ok(Err(cause)) => Err(cause),
+        }
+    }
+}
+
+pub struct AlwaysCallback<T: Callback> {
```
youngsofun, line 186: except for  `always`, do we need to add a flag to distinguish whether a callback failure will cause the statement to fail? Or should we just explain the current behavior through comments?”

src/query/pipeline/core/src/finished_chain.rs
```diff
@@ -0,0 +1,388 @@
...
+                true => "\u{2705}",
+                false => "\u{274C}",
+            };
+
+            let always_state = match apply_state.is_always {
+                true => "(always) ",
+                false => "",
+            };
+
+            writeln!(
```
youngsofun, line 180: just a suggestion, I prefer to require a name for each `callback`, just like the processors, make it more readable.

and then we can display it by adding  just one line when display a pipeline ?

src/query/pipeline/core/src/finished_chain.rs
zhang2014: `always` is special, by default, callback failure should always cause the statement to fail.

src/query/pipeline/core/src/finished_chain.rs
zhang2014: Great suggestion. but it is only recorded in the log. Is the caller location okay?

src/query/pipeline/core/src/finished_chain.rs
youngsofun: so `always` include 3 meanings

1.  itself always gets run, not affected by others 
2. if it fail, not affect others
3. if it fail, not affect a result of the stament.

youngsofun (approved):

src/query/pipeline/core/src/finished_chain.rs
youngsofun: may ok if there are no more than 3 or 4 callbacks.
some can be identified from the file path if the debugger has seen it before.

Issue #15707, opened 2024-06-03
bug: databend-meta snapshot oom 


wubx:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

main

### What's Wrong?

databend-meta sanpshot OOM
```
curl -qs 127.0.0.1:28101/v1/ctrl/trigger_snapshot
```

### How to Reproduce?

Create 100,000 tables， trigger_snapshot 
[jeprof.755385.311.i311.heap.pdf](https://github.com/user-attachments/files/15527692/jeprof.755385.311.i311.heap.pdf)
[jeprof.755385.10.i10.heap.pdf](https://github.com/user-attachments/files/15527693/jeprof.755385.10.i10.heap.pdf)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

drmingdrmer:
An initial observation is that when purging a raft log, memory is allocated for the log. I'm going to investigate further to determine why this occurs.

<img width="325" alt="image" src="https://github.com/datafuselabs/databend/assets/44069/fa18c76f-2e84-4ad8-ab9b-4f5431b8c957">

wubx:
```
cat databend-meta.toml
# Usage:
# databend-meta -c databend-meta-node-1.toml

log_dir                 = "/data1/databend/log/"
admin_api_address       = "0.0.0.0:28101"
grpc_api_address        = "0.0.0.0:9191"
# databend-query fetch this address to update its databend-meta endpoints list,
# in case databend-meta cluster changes.
grpc_api_advertise_host = "192.168.1.100"

max_applied_log_to_keep = 1000
snapshot_logs_since_last=10000


[raft_config]
id            = 2
raft_dir      = "/data1/databend/meta/"
raft_api_port = 28103

# Assign raft_{listen|advertise}_host in test config.
# This allows you to catch a bug in unit tests when something goes wrong in raft meta nodes communication.
raft_listen_host = "192.168.1.100"
raft_advertise_host = "192.168.1.100"

sled_max_cache_size_mb = 64000

# per
snapshot_logs_since_last=10240
# Start up mode: single node cluster
single        = true
```
databend-meta v1.2.510-nightly-b7c6b089e2-simd(1.78.0-nightly-2024-06-03T08:05:27.094897524Z)

The problem is not solved.

drmingdrmer:
@wubx 
[...]

PR #15709, opened 2024-06-03
feat: add meta-service config sled_max_cache_size_mb

drmingdrmer:

##### feat: add meta-service config sled_max_cache_size_mb

Provide a config/CLI argument `sled_max_cache_size_mb` to specify the
max memory in MB sled is allowed to use. By default it is `10GB`.

Usage of letting meta-service use at most 64 GB for cache in `config.toml`:

```toml
[raft_config]
sled_max_cache_size_mb = 64000
```

- See: #15707

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15709)

### DIFF
--- a/src/meta/embedded/src/meta_embedded.rs
+++ b/src/meta/embedded/src/meta_embedded.rs
@@ -92,7 +92,7 @@ impl MetaEmbedded {
     /// Initialize a global embedded meta store.
     /// The data in `path` won't be removed after program exit.
     pub async fn init_global_meta_store(path: String) -> Result<(), MetaStorageError> {
-        databend_common_meta_sled_store::init_sled_db(path);
+        databend_common_meta_sled_store::init_sled_db(path, 64 * 1024 * 1024 * 1024);
 
         {
             let mut m = GLOBAL_META_EMBEDDED.as_ref().lock().await;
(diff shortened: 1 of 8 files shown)


BohuTANG (approved):

src/meta/embedded/src/meta_embedded.rs
```diff
@@ -92,7 +92,7 @@ impl MetaEmbedded {
     /// Initialize a global embedded meta store.
     /// The data in `path` won't be removed after program exit.
     pub async fn init_global_meta_store(path: String) -> Result<(), MetaStorageError> {
-        databend_common_meta_sled_store::init_sled_db(path);
+        databend_common_meta_sled_store::init_sled_db(path, 64 * 1024 * 1024 * 1024);
```
BohuTANG, line 95: 64GB is too larger? cc @everpcpc for a check.

everpcpc: The default value could be larger, and we can adjust it to small values in production.

drmingdrmer: This is only used for embedded sled db, which is for testing purpose only.

PR #15720, opened 2024-06-03
feat: basic support input file format ORC.

youngsofun:

## Summary

1.  both query/ load/transform from orc is supported, no plan for write yet.
2. behave like parquet.
3. based on crate orc-rust, limited by it`s interface (we can improve it later):
     1.  temp fork and pub some private items to sep async read from async decode: https://github.com/youngsofun/datafusion-orc/commit/4d69fe9eefb32ab31ccf7e9dcb647fdfc7250caa
     2. one file per partition
4. pushdown will be added later too.

cc @BohuTANG 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15720)


drmingdrmer:
ready to review?

youngsofun:
@drmingdrmer the meta part is ready to review, thanks!

src/query/storages/orc/src/file_reader.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn len(&self) -> u64 {
+        self.size
+    }
+
+    fn get_read(&self, _offset_from_start: u64) -> std::io::Result<Self::T> {
+        todo!()
+    }
+
+    fn get_bytes(&self, offset_from_start: u64, length: u64) -> std::io::Result<Bytes> {
+        let buffer = self
```
youngsofun, line 37: get opendal error: `Unsupported (persistent) at read => service s3 doesn't support operation read.'`
but works on my local minio.

youngsofun: @Xuanwo

drmingdrmer (commented):
> Reviewed 7 of 32 files at r1.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15720)* status: 7 of 33 files reviewed, 2 unresolved discussions (waiting on @b41sh, @sundy-li, @Xuanwo, and @youngsofun)

___
*[`src/meta/proto-conv/tests/it/v092_orc_format_params.rs` line 42 at r2](https://reviewable.io/reviews/datafuselabs/databend/15720#-NzVoB-G51GfNDa_r0kf:-NzVoB-G51GfNDa_r0kg:b-dz89nu) ([raw file](https://github.com/datafuselabs/databend/blob/4d70c5b7c86674d6d8de7053faf66661c09d5831/src/meta/proto-conv/tests/it/v092_orc_format_params.rs#L42)):*
> ```XML
>     common::test_pb_from_to(func_name!(), want())?;
>     Ok(())
> }
> ```

The test should focus on decoding message including this new type, not only this new type itself.
In this case, it is `FileFormatParams`. 

And the test should run two cases:
```
    common::test_pb_from_to(func_name!(), want())?;
    common::test_load_old(func_name!(), x.as_slice(), 91, want())?;
```



_Code quote:_
```XML
fn test_decode_v92_orc_file_format_params() -> anyhow::Result<()> {
    let orc_file_format_params_v92 = vec![160, 6, 92, 168, 6, 24];
    let want = || OrcFileFormatParams {};
    common::test_load_old(
        func_name!(),
        orc_file_format_params_v92.as_slice(),
        92,
        want(),
    )?;
    common::test_pb_from_to(func_name!(), want())?;
    Ok(())
}
```

src/meta/proto-conv/tests/it/v092_orc_format_params.rs
```diff
@@ -0,0 +1,52 @@
...
+        orc_file_format_params_v92.as_slice(),
+        92,
+        want(),
+    )?;
+    common::test_pb_from_to(func_name!(), want())?;
+    Ok(())
+}
+
+#[test]
+fn test_decode_v92_file_format_params() -> anyhow::Result<()> {
```
youngsofun, line 46: @drmingdrmer plz review again

src/query/storages/orc/src/source.rs
```diff
@@ -0,0 +1,112 @@
...
+        if self.reader.is_none() {
+            let part = match self.table_ctx.get_partition() {
+                Some(part) => part,
+                None => return Ok(None),
+            };
+            let file = OrcFilePartition::from_part(&part)?.clone();
+            let path = file.path.clone();
+
+            let file = OrcFileReader {
+                operator: self.op.clone().layer(BlockingLayer::create()?).blocking(),
```
Xuanwo, line 81: Please note this `create` call must happen in async content.

If we do want to call this inside blocking function, please make sure we are guarded by `handle.enter()`, for example:

```rust
use once_cell::sync::Lazy;
use opendal::layers::BlockingLayer;
use opendal::services::S3;
use opendal::BlockingOperator;
use opendal::Operator;
use opendal::Result;

static RUNTIME: Lazy<tokio::runtime::Runtime> = Lazy::new(|| {
    tokio::runtime::Builder::new_multi_thread()
        .enable_all()
        .build()
        .unwrap()
});
///

fn main() -> Result<()> {
    // Create fs backend builder.
    let mut builder = S3::default();
    builder.bucket("test");
    builder.region("us-east-1");

    // Fetch the `EnterGuard` from global runtime.
    let _guard = RUNTIME.enter();
    // Build an `BlockingOperator` with blocking layer to start operating the storage.
    let _: BlockingOperator = Operator::new(builder)?
        .layer(BlockingLayer::create()?)
        .finish()
        .blocking();

    Ok(())
}
```

We can use `GlobalRuntime`.

drmingdrmer (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15720)* status: 7 of 33 files reviewed, 4 unresolved discussions (waiting on @b41sh, @sundy-li, and @youngsofun)

___
*[`src/meta/proto-conv/tests/it/v092_orc_format_params.rs` line 46 at r5](https://reviewable.io/reviews/datafuselabs/databend/15720#gh-1625250406:-NzWAHsODvBi-WEoaTZX:b-l05rdo) ([raw file](https://github.com/datafuselabs/databend/blob/ac04404acb6d61b4e1002466da84964549cabdf8/src/meta/proto-conv/tests/it/v092_orc_format_params.rs#L46)):*
<details><summary><i>Previously, youngsofun (Yang Xiufeng) wrote…</i></summary><blockquote>

@drmingdrmer plz review again
</blockquote></details>

I did not see any changes to this piece of codes :thinking:

src/meta/proto-conv/tests/it/v092_orc_format_params.rs
youngsofun: @drmingdrmer 
there was a `test_decode_v92_orc_file_format_params()`
add `test_decode_v92_file_format_params()`

src/query/storages/orc/src/source.rs
youngsofun: @Xuanwo the one in async panic:

```
     "Cannot start a runtime from within a runtime. This happens \
            because a function (like `block_on`) attempted to block the \
            current thread while the thread is being used to drive \
            asynchronous tasks."
```

b41sh (approved):

drmingdrmer (approved):
The meta part looks great.

> Reviewed 1 of 32 files at r1, 1 of 6 files at r5, 2 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15720#-:-NzWNOEH3WaOeSKlDkZR:b77o09f)* status: 9 of 33 files reviewed, 2 unresolved discussions (waiting on @sundy-li and @Xuanwo)

BohuTANG:
@soyeric128  Needs docs for this ORC foramt. Thanks.

PR #15751, opened 2024-06-07
fix(query): create ownership obejct only check current_role privilege

### LINKED ISSUE #15742: bug: create <object> only need to check current role privilege
**Summary**

Whether set secondary role ALL or None, create <object> only need to check current role privilege

TCeason:

## Summary

create ownership object only check current_role privilege

If do not have this limit, some unexpected problems may occur.

If a user does not bind a default role, the user can create an ownership object like db.

Then this user creates a database, it's will be owned role `public`. So all users can access this db. 

This may not be what users expect

Note:

If the user has the right privilege also can access the object.

- Fixes #15742

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15751)


drmingdrmer (commented):
I'm not very familiar with these parts of the logic, so I won't be approving or rejecting this PR.

> Reviewed 1 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15751#-:-NzlFdeN4uKYUnYsjBi1:bnulje4)* status: 1 of 14 files reviewed, all discussions resolved (waiting on @andylokandy and @flaneur2020)

src/query/catalog/src/table_context.rs
```diff
@@ -203,12 +203,13 @@ pub trait TableContext: Send + Sync {
     fn get_current_session_id(&self) -> String {
         unimplemented!()
     }
-    async fn get_all_effective_roles(&self) -> Result<Vec<RoleInfo>>;
+    async fn get_all_effective_roles(&self, create_object: bool) -> Result<Vec<RoleInfo>>;
```
flaneur2020, line 206: maybe we do not need add extra parameter on this method? we can consider CREATE XXX is a special case, and only make special checks with current role inside these operations, and do not need others to care about this special case.

src/query/catalog/src/table_context.rs
```diff
@@ -203,12 +203,13 @@ pub trait TableContext: Send + Sync {
...
 
     async fn validate_privilege(
         &self,
         object: &GrantObject,
         privilege: UserPrivilegeType,
+        create_object: bool,
```
flaneur2020, line 212: is this parameter can be removed? can we directly use the `privilege` parameter to check it's a CREATE XXX operation or not?

TCeason: Now create udf/stage/task need super privilege. So I think we can use parameter tmp. Until we support Create xxObject privilege then refactor this code.

src/query/service/src/sessions/session_privilege_mgr.rs
```diff
@@ -264,15 +272,15 @@ impl<'a> SessionPrivilegeManager for SessionPrivilegeManagerImpl<'a> {
     }
 
     #[async_backtrace::framed]
-    async fn has_ownership(&self, object: &OwnershipObject) -> Result<bool> {
+    async fn has_ownership(&self, object: &OwnershipObject, create_object: bool) -> Result<bool> {
```
flaneur2020, line 275: create_object -> current_role_only ?

`create_object` is bound to a concerete scenary, it might makes the semantic of this parameter decoupled from the special checks about `CREATE XXX`, but simply "only care the current role".

src/query/catalog/src/table_context.rs
TCeason: And this check, only need to check ownership object's created.

src/query/service/src/sessions/session_privilege_mgr.rs
```diff
@@ -242,6 +249,7 @@ impl<'a> SessionPrivilegeManager for SessionPrivilegeManagerImpl<'a> {
         &self,
         object: &GrantObject,
         privilege: UserPrivilegeType,
+        create_object: bool,
```
flaneur2020, line 252: can we derive `create_object` parameter by the `privilege` parameter ?

src/query/service/src/sessions/session_privilege_mgr.rs
TCeason: Yes . check_current_role_only is better.

src/query/service/src/sessions/session_privilege_mgr.rs
TCeason: Same like this:https://github.com/datafuselabs/databend/pull/15751/files#discussion_r1630942306

src/query/catalog/src/table_context.rs
flaneur2020: can also consider rename the `create_object` here as `check_current_role_only`, to describe what this function does instead of emphasis it's a special case.

src/query/catalog/src/table_context.rs
flaneur2020: we can consider move this logic out of `get_all_effective_roles`, and calls `get_current_role` on this special case.

src/query/service/src/sessions/session_privilege_mgr.rs
```diff
@@ -183,7 +189,21 @@ impl<'a> SessionPrivilegeManager for SessionPrivilegeManagerImpl<'a> {
...
     #[async_backtrace::framed]
-    async fn get_all_effective_roles(&self) -> Result<Vec<RoleInfo>> {
+    async fn get_all_effective_roles(
+        &self,
+        check_current_role_only: bool,
+    ) -> Result<Vec<RoleInfo>> {
+        // the current role is always included in the effective roles.
+        self.ensure_current_role().await?;
+        // if ensure_current_role success, the current role can not be None.
+        let current_role = self.get_current_role().unwrap();
```
flaneur2020, line 199: the old code can handle the default fallback logic without a hard `unwrap()`:

```
        let mut role_names = self
            .get_current_role()
            .map(|r| vec![r.name])
            .unwrap_or_default();
```

flaneur2020 (approved):
LGTM :rocket:

TCeason:
Now I think it's time to merge this pr. cc @flaneur2020

flaneur2020:
LGTM

TCeason:
@BohuTANG Please merge this pr.

After this pr, create an object only check the current role. So if some user creates an object return error, he should run SQL `set role` to ensure the current role has created privilege or ownership.

PR #15772, opened 2024-06-11
feat: Persist Meta-Service State Machine on Disk

drmingdrmer:

##### feat: Persist Meta-Service State Machine on Disk

This commit introduces a significant change to the meta-service by
persisting the state machine on disk rather than loading it entirely
into memory. This change aims to reduce memory consumption to a
configurable cache size and prevent memory bursts during snapshot
creation.

**Changes Implemented:**

- **Disk-Based State Machine:**
  - The in-memory state machine has been replaced with a disk-backed
    implementation using a `rotbl` file (Read-Only on disk Table).

- **Efficient Snapshot Building:**
  - When building a snapshot from the state machine, the output is now
    written directly to disk, which helps in avoiding double memory
    consumption.

- **Add RPC install_snapshot_v003**: receive a V003 snapshot in binary
  form.

- **New Configurations:**
  ```rust,ignore
  /// Whether to check if keys fed to the snapshot are sorted.
  snapshot_db_debug_check: bool,

  /// The maximum number of keys allowed in a block within a snapshot database.
  /// Smaller blocks enable more granular cache control but may increase the index size.
  snapshot_db_block_keys: u64,

  /// The total number of blocks to cache.
  snapshot_db_block_cache_item: u64,

  /// The total cache size for snapshot blocks, default is 1GB.
  snapshot_p_db_block_cache_size: u64,
  ```

**Compatibility:**

- Compatible with the V002 snapshot transport protocol and on-disk format.
- On startup, the old V002 snapshot will be upgraded to `V003`.
- When receiving a V002 install-snapshot request, it is converted to
  `V003` after receipt and before installation to the state machine.

**Detailed Changes:**

- **DB Wrapper:**
  - `DB` acts as a wrapper around `rotbl`, representing an on-disk
    snapshot and containing meta-service specific information such as
    metadata and file paths.

- **SnapshotStoreV003:**
  - Provides various snapshot functionalities:
    - `Self::new_writer` creates a writer for building snapshots from a
      series of key-value pairs.
    - `Self::new_loader` creates a loader to load snapshots into the
      state machine from disk.
    - `Self::new_receiver` creates a receiver to handle snapshot data
      from remote peers in binary format.

- **Data Export Changes:**
  - The `DataHeader` is no longer part of the exported snapshot data.

- **MapKeyEncode Trait:**
  - Extracted from `MapKey`, this trait defines the codec behavior for a
    key used in a `MapApi`.

- **Compactor Addition:**
  - Introduces a `Compactor` responsible for compacting in-memory
    immutable data or merging immutable data with persisted data.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

## Related Issues


src/meta/raft-store/src/leveled_store/db_builder.rs
```diff
@@ -0,0 +1,124 @@
...
+        let b = Self {
+            path: p.to_string(),
+            rotbl_builder: inner,
+        };
+
+        Ok(b)
+    }
+
+    /// This method is only used for test.
+    #[allow(dead_code)]
```
Xuanwo, line 60: Still dead code?

Cargo.toml
```diff
@@ -304,6 +313,7 @@ reqwest = { version = "0.12", default-features = false, features = [
     "rustls-tls-native-roots",
 ] }
 reqwest-hickory-resolver = "0.1"
+rotbl = { git = "https://github.com/drmingdrmer/rotbl", tag = "v0.1.2-alpha.6", features = [] }
```
Xuanwo, line 316: Let's release it!

Xuanwo (commented):
Is there a plan to support save snapshot on remote storage services like s3?

src/meta/raft-store/src/leveled_store/db_builder.rs
drmingdrmer: It's used by tests code. 🤔

drmingdrmer:
> Is there a plan to support save snapshot on remote storage services like s3?

Yes with OpenDAL it's not big deal!

- There will be additional latency when reading the kv from it but it's would be OK, because kvs are read in chunks.
- Other refactor may be required since if the leader build and store the snapshot on S3, the follower could just use it and do not build again.
- But in a deployment where there is no S3, it should be able to fallback to the local disk snapshot.

PR #15801, opened 2024-06-15
refactor: improve raft connection and logging filter

drmingdrmer:

##### refactor: improve raft connection and logging filter

Only reuse a connection if there is no error returned from the last RPC.
When a raft RPC is finished, push the connection back for next RPC if it
returns an `Ok`. Otherwise it drops the connection and the next RPC will
re-create a new one. Such approach ensures no problematic connection
will be used.

Support multi log prefix filter.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Refactoring

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15801)

### DIFF
--- a/src/common/tracing/src/init.rs
+++ b/src/common/tracing/src/init.rs
@@ -22,6 +22,7 @@ use databend_common_base::base::GlobalInstance;
 use databend_common_base::runtime::Thread;
 use log::LevelFilter;
 use log::Log;
+use log::Metadata;
 use minitrace::prelude::*;
 use opentelemetry_otlp::WithExportConfig;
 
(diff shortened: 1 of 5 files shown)


src/common/tracing/src/init.rs
```diff
@@ -309,3 +301,35 @@ pub fn init_logging(
...
+    let filters = prefix_filter
+        .split(',')
+        .map(|x| x.to_string())
+        .collect::<Vec<_>>();
+
+    move |meta| match_prefix_many(meta, &filters)
+}
+
+fn match_prefix_many(meta: &Metadata, prefixes: &[String]) -> bool {
+    if is_severe(meta) {
```
andylokandy, line 316: is it typo?

src/meta/service/src/network.rs
```diff
@@ -138,10 +110,10 @@ impl Backoff {
 impl Default for Backoff {
     fn default() -> Self {
         Self {
-            back_off_ratio: 2.0,
-            back_off_min_delay: Duration::from_secs(1),
-            back_off_max_delay: Duration::from_secs(60),
-            back_off_chances: 3,
+            back_off_ratio: 1.5,
```
lichuang, line 113: are these params configuable?

src/common/tracing/src/init.rs
drmingdrmer: No, warning and error logs are considered severe, is there better term for it?

src/meta/service/src/network.rs
drmingdrmer: Not yet.

src/common/tracing/src/init.rs
andylokandy: ok. I think it's fine.

PR #15805, opened 2024-06-16
refactor: adaptive entries count for `AppendEntriesRequest`

drmingdrmer:

##### refactor: adaptive entries count for `AppendEntriesRequest`

Use `limited_get_log_entries()` to limit the count of log entries in a
`AppendEntries` request by size.
With this approach, every AppendEntries request won't exceed 2MB.
This address the issue with static configured number of entries in a
`AppendEntries` can not limit the RPC size. Such too big AppendEntries
RPC may timeout.

Other changes:

- Only reuse a connection if there is no error returned from the last RPC.
  When a raft RPC is finished, push the connection back for next RPC if it
  returns an `Ok`. Otherwise it drops the connection and the next RPC will
  re-create a new one. Such approach ensures no problematic connection
  will be used.

- Improve raft connection and logging filter. Support multi log prefix filter.

- Add `DisplayOption` to improve display for raft log id:
  `Option<LogId>::display()` returns an `Display` implementation.

- Log elapsed time for critical steps of `AppendEntries` RPC.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Refactoring

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15805)

### DIFF
--- a/src/meta/types/build.rs
+++ b/src/meta/types/build.rs
@@ -47,91 +47,91 @@ fn build_proto() {
         .file_descriptor_set_path(out_dir.join("meta_descriptor.bin"))
         .type_attribute(
             "SeqV",
-            "#[derive(Eq, serde::Serialize, serde::Deserialize)]",
+            "#[derive(Eq, serde::Serialize, serde::Deserialize, deepsize::DeepSizeOf)]",
         )
         .type_attribute(
             "TxnGetRequest",
-            "#[derive(Eq, serde::Serialize, serde::Deserialize)]",
+            "#[derive(Eq, serde::Serialize, serde::Deserialize, deepsize::DeepSizeOf)]",
         )
         .type_attribute(
             "TxnPutRequest",
-            "#[derive(Eq, serde::Serialize, serde::Deserialize)]",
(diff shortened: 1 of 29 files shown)


src/meta/types/build.rs
```diff
@@ -47,91 +47,91 @@ fn build_proto() {
         .file_descriptor_set_path(out_dir.join("meta_descriptor.bin"))
         .type_attribute(
             "SeqV",
-            "#[derive(Eq, serde::Serialize, serde::Deserialize)]",
+            "#[derive(Eq, serde::Serialize, serde::Deserialize, deepsize::DeepSizeOf)]",
```
BohuTANG, line 50: Is there a performance issue with `deepsize::DeepSizeOf`? I'm not certain whether it's being called each time or not.

We encountered a "get-size" issue years ago:
https://github.com/datafuselabs/databend/pull/8461
I'm not certain if they're the same; this is just a friendly suggestion.

drmingdrmer: To calculate the size of a 98KB raft log, it takes 1~2 us. It's only used when replication raft log. the performance is quite enough.

lichuang (approved):

PR #15831, opened 2024-06-19
chore: update logcall

andylokandy:

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15831)


andylokandy (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15831)* status: all files reviewed, 1 unresolved discussion (waiting on @drmingdrmer)

___
*[`src/meta/api/src/schema_api_impl.rs` line 269 at r1](https://reviewable.io/reviews/datafuselabs/databend/15831#-O-kdk7d2tkGeezcgN78:-O-ksKJ67u1e7wv2Kz3b:b-ilndxa) ([raw file](https://github.com/datafuselabs/databend/blob/4024618e5cbaea03f9588d520cd22b96a5bca2c8/src/meta/api/src/schema_api_impl.rs#L269)):*

Is it able to keep the struct name `SchemaApi` in the debug log?
</blockquote></details>

It's not possible in current logcall. But it's possible to upgrade.

drmingdrmer (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15831)* status: all files reviewed, 1 unresolved discussion (waiting on @andylokandy)

___
*[`src/meta/api/src/schema_api_impl.rs` line 269 at r1](https://reviewable.io/reviews/datafuselabs/databend/15831#-O-kdk7d2tkGeezcgN78:-O-kuRjKAMwteoo2bfSb:b-1ih2ji) ([raw file](https://github.com/datafuselabs/databend/blob/4024618e5cbaea03f9588d520cd22b96a5bca2c8/src/meta/api/src/schema_api_impl.rs#L269)):*
<details><summary><i>Previously, andylokandy (Andy Lok) wrote…</i></summary><blockquote>

It's not possible in current logcall. But it's possible to upgrade.
</blockquote></details>

In that case, retaining the manual debug statements is a sensible choice. Given that there are several structures with nearly identical method sets, such as `catalog.create_table` invoking `SchemaApi::create_table`.

And having debug statements at the start of each method can be useful. This approach allows for better traceability and understanding of the workflow, especially when diagnosing issues, as opposed to only logging when a method completes.

andylokandy:
> having debug statements at the start of each method can be useful. This approach allows for better traceability and understanding of the workflow, especially when diagnosing issues, as opposed to only logging when a method completes.

For such aspect, I recommend enabling structlog for meta, which will arrange the log in the way functions is call in hierarchy, and is already setup for query

andylokandy:
The new patch will log the function in full path: https://github.com/andylokandy/logcall/pull/5

drmingdrmer:
Can you keep the original debug log statement so that I can still get enough information while evaluating the new logging?

andylokandy:
updated

drmingdrmer (approved):
Thank you

> Reviewed 2 of 2 files at r2, 1 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15831#-:-O03_R5n32XB-ZoU4NUa:bvs55d0)* status: all files reviewed, 1 unresolved discussion (waiting on @andylokandy)

PR #15858, opened 2024-06-21
refactor: upgrade QuotaMgr to using protobuf

### LINKED ISSUE #15009: Refactor: upgrade QuotaMgr to using protobuf
#### The problem

Currently QuotaMgr write json data into meta-service:
https://github.com/drmingdrmer/databend/blob/f248f211c9418e153d74515d0a755fbda11ef3ec/src/query/management/src/quota/quota_mgr.rs#L73

Json does not provide compatibility support as `FromToProto` does:
https://github.com/drmingdrmer/databend/blob/6332061fb2c8466c398b417cc6db878ba5933740/src/meta/proto-conv/src/from_to_protobuf.rs#L18

The setting data read/write should be upgrade to using `FromToProto` to provide well defined and tested backward compatibility.

#### TODO

- [x] keep the writing part unchanged, but read in pb/json compatible way.
- [ ] upgrade to write pb data instead of json.
- [ ] upgrade to read only pb format.

forsaken628:

## Summary
- Fixes #15009

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)


src/meta/protos/proto/tenant.proto
```diff
@@ -28,3 +28,23 @@ message TIdent {
   // Resource name
   string name = 2;
 }
+
+message TQuota {
```
Xuanwo, line 32: `TQuota` doesn't looks good to me, how about using `TenantQuota` for better reading?

drmingdrmer (commented):
It can not be replaced with a new storage format in one shot.

The first step is to update the reading part so that both pb and json format can both be read.
After that, it is safe to write these data in pb format.

> Reviewed 6 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858#-:-O-vT0mj7SQ-YmFoax6C:b95ob0p)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @forsaken628 and @Xuanwo)

drmingdrmer (changes requested):
> Reviewed 3 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @forsaken628 and @Xuanwo)

___
*[`src/query/management/src/quota/quota_mgr.rs` line 86 at r2](https://reviewable.io/reviews/datafuselabs/databend/15858#-O-xdMvt9jmA8xzVlzPU:-O-xdMvt9jmA8xzVlzPV:b-fj61gk) ([raw file](https://github.com/datafuselabs/databend/blob/6a0320849cf5ba612fdb5780b01924ea8ae56cee/src/query/management/src/quota/quota_mgr.rs#L86)):*
> ```XML
>         let res = self
>             .kv_api
>             .upsert_pb(&UpsertPB::update(self.ident.clone(), quota.clone()).with(seq))
> ```

It can not write pb in this commit. Otherwise when a new version databend-query and an old version databend-query 
 are deployed together, pb format data can not be read by the old version databend-query can not read the data.

In this version it must write in json format too.

drmingdrmer:
@TCeason is there a way or is it possible to upgrade all of the TenantQuota records? I did not find a piece of sample codes.

TCeason:
> @TCeason is there a way or is it possible to upgrade all of the TenantQuota records? I did not find a piece of sample codes.

I'm not familiar about TenantQuote. In roles or ownership has api like get_roles to get all roles so I can add check upgrade in there.

But looks like quota not support get all quotas.

cc @flaneur2020 @ZhiHanZ 

Please have a look about this issue. Maybe need to add a control to upgrade all quota form json to pb.

flaneur2020:
i suppose that quota is less critical than the rbac stuffs. i think we can move this forward if we're able to fallback to default quotas on corrupt data.

drmingdrmer (changes requested):
> Reviewed 1 of 5 files at r2, 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @forsaken628 and @Xuanwo)

___
*[`src/query/management/src/quota/quota_mgr.rs` line 72 at r3](https://reviewable.io/reviews/datafuselabs/databend/15858#-O0Tz_5wAsgtyB-N5VRz:-O0Tz_5xEdCo8Ucu9o38:bmep6f2) ([raw file](https://github.com/datafuselabs/databend/blob/67f235bbe2d957a893522a58c6c4398aa3f7e8bd/src/query/management/src/quota/quota_mgr.rs#L72)):*
> ```XML
>                     } else {
>                         Quota::new_limit(func_name!(), 0)
>                     };
> ```

Even when `WRITE_PB` is false, the json data can be upgraded to pb format and should be upgraded IMHO.



_Code quote:_
```XML
                    let mut quota = if WRITE_PB {
                        Quota::new(func_name!())
                    } else {
                        Quota::new_limit(func_name!(), 0)
                    };
```

forsaken628:
> It can not write pb in this commit. Otherwise when a new version databend-query and an old version databend-query are deployed together, pb format data can not be read by the old version databend-query can not read the data.
> 
> In this version it must write in json format too.

> Even when WRITE_PB is false, the json data can be upgraded to pb format and should be upgraded IMHO.

I'm confused as to why the inconsistency in opinion.

drmingdrmer:
> > It can not write pb in this commit. Otherwise when a new version databend-query and an old version databend-query are deployed together, pb format data can not be read by the old version databend-query can not read the data.
> > In this version it must write in json format too.
> 
> > Even when WRITE_PB is false, the json data can be upgraded to pb format and should be upgraded IMHO.
> 
> I'm confused as to why the inconsistency in opinion.

Another process may already started writing PB format data. This in-process config `WRITE_PB` does not guarantee anything and could just be removed.

forsaken628:
https://github.com/datafuselabs/databend/blob/67f235bbe2d957a893522a58c6c4398aa3f7e8bd/src/query/management/src/quota/quota_mgr.rs#L90-L91
There will not be another process to start writing PB data...

If we choose to update without downtime it will necessarily be a two-phase update, replacing all the old processes and then enabling PB writes.

It might be better to replace x from constants to environment variables, and these wouldn't need to be recompiled, but a commit would still be needed to clean up the old code.

drmingdrmer:
> https://github.com/datafuselabs/databend/blob/67f235bbe2d957a893522a58c6c4398aa3f7e8bd/src/query/management/src/quota/quota_mgr.rs#L90-L91
> 
> 
> There will not be another process to start writing PB data...
> If we choose to update without downtime it will necessarily be a two-phase update, replacing all the old processes and then enabling PB writes.
> 
> It might be better to replace x from constants to environment variables, and these wouldn't need to be recompiled, but a commit would still be needed to clean up the old code.

Shouldn't the step-1 version be **write-json plus read-json-or-pb**?

forsaken628:
Yes, isn't that exactly what this pr is about right now?

drmingdrmer:
> Yes, isn't that exactly what this pr is about right now?

Oh I see. It looks like the introduced `WRITE_PB` with `check_and_upgrade_pb()` makes the logic a little bit obscure to understand.

BTW it would be better make `WRITE_PB` a generic const parameter of `QuotaMgr`, such as `struct QuotaMgr<const WRITE_PB: bool> {}`.
This way the compatibility can be tested.

forsaken628 (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @drmingdrmer and @Xuanwo)

___
*[`src/query/management/src/quota/quota_mgr.rs` line 72 at r3](https://reviewable.io/reviews/datafuselabs/databend/15858#-O0Tz_5wAsgtyB-N5VRz:-O0h4doYEqaUFo-3jHCz:b-896fix) ([raw file](https://github.com/datafuselabs/databend/blob/67f235bbe2d957a893522a58c6c4398aa3f7e8bd/src/query/management/src/quota/quota_mgr.rs#L72)):*

Even when `WRITE_PB` is false, the json data can be upgraded to pb format and should be upgraded IMHO.
</blockquote></details>

Done.

drmingdrmer (commented):
To ensure consistent compatibility going forward, it's crucial to implement tests for this change. For instance, you could write data using both `QuotaMgr<W=true>` and `QuotaMgr<W=false>`, and then read it back to verify consistency.

The UDF test provides a good example of how to approach this. It sets up an embedded meta store and tests the functions based on the KVApi, which could be a suitable model for constructing your tests.

https://github.com/datafuselabs/databend/blob/154115853ea79cdc44135e120e5c04a0e1d5ad4e/src/query/management/tests/it/udf.rs#L33

> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858#-:-O0ic22l0fBfrQUkh7We:b-49ajao)* status: 8 of 9 files reviewed, 4 unresolved discussions (waiting on @forsaken628 and @Xuanwo)

___
*[`src/query/management/src/quota/quota_mgr.rs` line 38 at r4](https://reviewable.io/reviews/datafuselabs/databend/15858#-O0iGv1D6mJJEHgkOu7g:-O0iGv1E0NTcQSLN-oaE:b-7ytitj) ([raw file](https://github.com/datafuselabs/databend/blob/c95dad1d82fb5ef2134c1d937ba3620ab1a5f94b/src/query/management/src/quota/quota_mgr.rs#L38)):*
> ```XML
> use crate::serde::Quota;
> 
> pub struct QuotaMgr<const W: bool = false> {
> ```

`WRITE_PB` would be better.
___
*[`src/query/management/src/quota/quota_mgr.rs` line 44 at r4](https://reviewable.io/reviews/datafuselabs/databend/15858#-O0iH7yyDzKQZo1f39kO:-O0iH7yyDzKQZo1f39kP:b-ee3s5s) ([raw file](https://github.com/datafuselabs/databend/blob/c95dad1d82fb5ef2134c1d937ba3620ab1a5f94b/src/query/management/src/quota/quota_mgr.rs#L44)):*
> ```XML
> 
> impl<const W: bool> QuotaMgr<W> {
>     const WRITE_PB: bool = W;
> ```

You can just use the const parameter directly:
```rust
impl<const WRITE_PB: bool> QuotaMgr<WRITE_PB> {
    fn set_quota(&self, _quota: i32) {
        if WRITE_PB {
            println!("write pb");
        } else {
            println!("write json");
        }
    }
}
```



_Code quote:_
```XML
impl<const W: bool> QuotaMgr<W> {
    const WRITE_PB: bool = W;
```

drmingdrmer (commented):
LGTM, the test should have covered the compatibility for next upgrade. Thank you.

BTW, the associated type can be just replaced with `impl<const WRITE_PB: bool> QuotaMgr<WRITE_PB>`:
```rust
impl<const W: bool> QuotaMgr<W> {
    const WRITE_PB: bool = W;
```
Unless I missed anything?

> Reviewed 2 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858#-:-O0lRQ8B5BxkqbFwRTs9:b7hh0fi)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @forsaken628 and @Xuanwo)

drmingdrmer (approved):
> Reviewed 1 of 3 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: all files reviewed, 3 unresolved discussions (waiting on @forsaken628 and @Xuanwo)

drmingdrmer (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: all files reviewed, 3 unresolved discussions (waiting on @forsaken628 and @Xuanwo)

BohuTANG:
Conflicting files
src/meta/proto-conv/src/util.rs
src/meta/proto-conv/tests/it/main.rs

drmingdrmer (commented):
> Reviewed 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: all files reviewed, 2 unresolved discussions (waiting on @forsaken628 and @Xuanwo)

drmingdrmer (approved):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: all files reviewed, 1 unresolved discussion (waiting on @forsaken628 and @Xuanwo)

drmingdrmer (approved):
> Dismissed @Xuanwo from a discussion.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15858)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Xuanwo)

drmingdrmer:
Thank you @forsaken628, it looks ready to merge!

Issue #14487, opened 2024-01-26
bug: window function out of bounds access panic 


b41sh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

v1.2.314-nightly

### What's Wrong?

window function out of bounds access panic 
```
panicked at src/query/service/src/pipelines/processors/transforms/window/transform_window.rs:157:21:
Out of bounds access
```

### How to Reproduce?

```sql
mysql> SELECT DISTINCT lead((861, FALSE, FALSE), 9, (849, TRUE, TRUE)) OVER (PARTITION BY 15560425903542832284, 965871850213131579 ORDER BY 13746504519650342222, 5897530378272856518 ASC NULLS FIRST) from range(100, 12000000, 467);
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| lead((861, false, false), 9, (849, true, true)) over (partition by 15560425903542832284, 965871850213131579 order by 13746504519650342222, 5897530378272856518 asc nulls first) |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| (861,0,0)                                                                                                                                                                       |
| (849,1,1)                                                                                                                                                                       |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
2 rows in set (0.11 sec)
Read 25696 rows, 200.75 KiB in 0.072 sec., 354.9 thousand rows/sec., 2.71 MiB/sec.

mysql> SELECT DISTINCT lead((861, FALSE, FALSE), 9, (849, TRUE, TRUE)) OVER (PARTITION BY 15560425903542832284, 965871850213131579 ORDER BY 13746504519650342222, 5897530378272856518 ASC NULLS FIRST) from range(100, 120000000, 467);
ERROR 1105 (HY000): PanicError. Code: 1104, Text = Out of bounds access.
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

shamb0:
Hi @b41sh,

I am interested in solving this. Could you please assign it to me ?

shamb0:
Thank You @sundy-li

PR #15860, opened 2024-06-21
fix(query): window function out of bounds access panic

shamb0:

## Summary

- Fixes https://github.com/datafuselabs/databend/issues/14487

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15860)


shamb0:
The out-of-bounds error was caused by an underflow in the index calculation when using the default `self.peer_group_start` value.

I updated the `are_peers()` method to only use `self.peer_group_start` when `self.is_ranking` is `true`. This prevents the underflow.

I confirmed the fix resolves the reported test case failure.

```
root@localhost:8000/default> SELECT DISTINCT lead((861, FALSE, FALSE), 9, (849, TRUE, TRUE)) OVER (PARTITION BY 15560425903542832284, 965871850213131579 ORDER BY 13746504519650342222, 5897530378272856518 ASC NULLS FIRST) from range(100, 120000000, 467);

SELECT
  DISTINCT lead((861, FALSE, FALSE), 9, (849, TRUE, TRUE)) OVER (
    PARTITION BY 15560425903542832284,
    965871850213131579
    ORDER BY
      13746504519650342222,
      5897530378272856518 ASC NULLS FIRST
  )
FROM
  RANGE(100, 120000000, 467)

┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ lead((861, false, false), 9, (849, true, true)) over ( partition by 15560425903542832284, 965871850213131579 order by 13746504519650342222, 5897530378272856518 asc nulls first ) │
│                                                                          Tuple(UInt16, Boolean, Boolean)                                                                          │
├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ (861,false,false)                                                                                                                                                                 │
│ (849,true,true)                                                                                                                                                                   │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
2 rows read in 4.064 sec. Processed 256.96 thousand rows, 1.96 MiB (63.22 thousand rows/s, 493.94 KiB/s)
```

sundy-li:
Can you add test case as logicaltest file?

shamb0:
> Can you add test case as logicaltest file?

Done, @sundy-li

b41sh:
@shamb0 thanks for your contribution, but some test failed, this modification seem not correct.
https://github.com/datafuselabs/databend/actions/runs/9626684558/job/26553089343?pr=15860

shamb0:
> @shamb0 thanks for your contribution, but some test failed, this modification seem not correct. https://github.com/datafuselabs/databend/actions/runs/9626684558/job/26553089343?pr=15860

Thanks @b41sh, for highlighting this test-case, I will pick-it up today

b41sh:
Hi, @shamb0 This bug is caused by sorting constant value. We can determine whether the value of the order by is `ConstantExpr` or not at the binder stage, and ignore it if it is, and the code in `TransformWindow` does not need to be changed.
You can add the following code at this file.
```rust
if matches!(order, ScalarExpr::ConstantExpr(_)) {
    continue;
}
```
https://github.com/datafuselabs/databend/blob/main/src/query/sql/src/planner/semantic/type_check.rs#L1152

shamb0:
> Hi, @shamb0 This bug is caused by sorting constant value. We can determine whether the value of the order by is `ConstantExpr` or not at the binder stage, and ignore it if it is, and the code in `TransformWindow` does not need to be changed. You can add the following code at this file.
> 
> ```rust
> if matches!(order, ScalarExpr::ConstantExpr(_)) {
>     continue;
> }
> ```
> 
> https://github.com/datafuselabs/databend/blob/main/src/query/sql/src/planner/semantic/type_check.rs#L1152

Thanks @b41sh. The proposed solution is effective and addresses the issue. All logical test suites in `window_function` now pass. I appreciate the insights provided by this solution.

b41sh (approved):
LGTM, cc @sundy-li

PR #15869, opened 2024-06-23
feat: update sort algorithm using loser tree for multi sort merge

### LINKED ISSUE #11604: Feature: Update sort algorithm using Loser Tree

forsaken628:

## Summary

Fixes #11604

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15869)


sundy-li:
Is there any performance comparison data of this pr?

forsaken628:
> Is there any performance comparison data of this pr?

Not yet, do I need to add an algorithm level benchmark, or just ci-benchmark?

sundy-li:
you can pref it in local, eg:

```
create table t(a int, b string, d float, e date);
create table t_random like t engine = random;

## generate data
insert into t select * from t_random limit 5000000;
insert into t select * from t_random limit 5000000;
insert into t select * from t_random limit 5000000;
insert into t select * from t_random limit 5000000;
...

## sort perf 

select * from t order by a,b ignore_result;
...
```

forsaken628:
> you can pref it in local, eg:
> 
> ```
> create table t(a int, b string, d float, e date);
> create table t_random like t engine = random;
> 
> ## generate data
> insert into t select * from t_random limit 5000000;
> insert into t select * from t_random limit 5000000;
> insert into t select * from t_random limit 5000000;
> insert into t select * from t_random limit 5000000;
> ...
> 
> ## sort perf 
> 
> select * from t order by a,b ignore_result;
> ...
> ```

No noticeable changes. There seems to be a slight improvement from flamegraph, not obvious. It could also be due to the distribution of the data.

flamegraph

## data set

read rows: 25000000
read size: 202.54 MiB
partitions total: 24
partitions scanned: 24
pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 24 to 24>]
push downs: [filters: [], limit: NONE]
estimated rows: 25000000.00

## sql

### sort
select * from t order by a,b ignore_result;
[pprof.cpu.heap.sort.pb.gz](https://github.com/user-attachments/files/15981752/pprof.cpu.heap.sort.pb.gz)
[pprof.cpu.loser.sort.pb.gz](https://github.com/user-attachments/files/15981754/pprof.cpu.loser.sort.pb.gz)

### window 
SELECT MAX(d) OVER (PARTITION BY a) FROM t ignore_result;
[pprof.cpu.heap.window.pb.gz](https://github.com/user-attachments/files/15981753/pprof.cpu.heap.window.pb.gz)
[pprof.cpu.loser.window.pb.gz](https://github.com/user-attachments/files/15981755/pprof.cpu.loser.window.pb.gz)

sundy-li:
The codes LGTM. Some comments about this pr:

Better add a new setting named `enable_loser_tree_merge_sort`, default to `1`.

Then we can create the `merger` in `MultiSortMergeProcessor` by setting in runtime. If any bug happens, we can switch to use different impl.

src/query/pipeline/transforms/src/processors/transforms/sort/rows/simple.rs
```diff
@@ -81,7 +81,7 @@ where
     T::Scalar: Ord,
 {
     fn eq(&self, other: &Self) -> bool {
-        self == other
+        self.inner == other.inner && self.desc == other.desc
```
forsaken628, line 84: Should desc be ignored or is it inconsistent with cmp behavior.

sundy-li: the descs are consistent.

src/query/settings/src/settings_default.rs
```diff
@@ -775,6 +775,12 @@ impl DefaultSettings {
                     desc: "The maximum count of set operator in a query. If your query stack overflow, you can reduce this value.",
                     mode: SettingMode::Both,
                     range: Some(SettingRange::Numeric(0..=u64::MAX)),
+                }),
+                ("enable_loser_tree_merge_sort", DefaultSettingValue {
+                    value: UserSettingValue::UInt64(1),
```
sundy-li, line 780: I still did not find `loser_tree` is faster than `heap`.

```
🐳 :) select l_orderkey from lineitem order by l_orderkey ignore_result;
0 row read in 1.409 sec. Processed 59.99 million row, 457.66 MiB (42.58 million rows/s, 324.82 MiB/s)

🐳 :) set enable_loser_tree_merge_sort = 0;
🐳 :) select l_orderkey from lineitem order by l_orderkey ignore_result;
0 row read in 1.258 sec. Processed 59.99 million row, 457.66 MiB (47.7 million rows/s, 363.90 MiB/s)
```

```

🐳 :) set enable_loser_tree_merge_sort = 1;
🐳 :) select userid from hits order by userid ignore_result;
0 row read in 3.027 sec. Processed 100 million row, 762.92 MiB (33.04 million rows/s, 252.08 MiB/s)
🐳 :) set enable_loser_tree_merge_sort = 0;
🐳 :) select userid from hits order by userid ignore_result;
0 row read in 2.528 sec. Processed 100 million row, 762.92 MiB (39.56 million rows/s, 301.83 MiB/s)
```

sundy-li: Maybe we should disable it by default unless we find it's faster.

But the sort algorithm abstraction is fine.

sundy-li: Is it true that when K is large enough, the performance of the loser tree will be better?

Current K is the num of physical cores by default.

forsaken628: https://github.com/datafuselabs/databend/blob/7b15120fd2b2f91cf3a3f9ca860ea733f830eefc/src/query/pipeline/transforms/src/processors/transforms/sort/merger.rs#L171
I think the problem is with this peek_top2.
It is generally accepted that the loser tree is twice as fast as the heap, but the need to fetch top2 here causes the loser tree to be accessed twice per iteration. And the standard library's heap is deeply optimized. Maybe some kind of variant implementation can be found later to improve this situation.

sundy-li (approved):

src/query/settings/src/settings_default.rs
sundy-li: Tracked in https://github.com/datafuselabs/databend/issues/15925

forsaken628:
Here's another part about Merger's changes.

https://github.com/forsaken628/databend/compare/loser-tree...forsaken628:databend:peek-mut

The cursor contains Arc, and iterative cloning and deletion incurs a lot of overhead, so I'm replacing cloning with in-place updating.

Should these two parts of the changes be separately PR or combined?

sundy-li:
> Should these two parts of the changes be separately PR or combined?

You can add into this pr directly.

zhang2014 (approved):

PR #15871, opened 2024-06-24
feat: support filling uuid when loading data and uuid is missing/empty in some rows.

youngsofun:

## Summary

 fix dependency by the way:  `databend-common-sql` should not dep on `storages-xxx`， add 2 func in table context, maybe we can have better way to solve it in the further.  

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15871)


youngsofun:
cc @wubx

src/query/expression/src/expression.rs
```diff
@@ -611,3 +611,11 @@ impl<Index: ColumnIndex> RemoteExpr<Index> {
         }
     }
 }
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, PartialEq, Eq)]
+pub enum FieldDefaultExpr {
```
b41sh, line 616: how about use `RemoteExpr` instead, which is more general.

b41sh (approved):

src/query/expression/src/expression.rs
youngsofun: good idea, but we need this fix as soon as possible, while I am a little busy now,  maybe we can refactor it later.

b41sh: Since the PR needs are very urgent, we can first merge it and optimize codes later.

Issue #15908, opened 2024-06-26
optimize  transform_serialize_block for scalar column


youngsofun:
**Summary**

these steps in BlockBuilder.build covert block to full first.

- [x] BloomIndexState::from_data_block https://github.com/datafuselabs/databend/pull/15910
- [x] gen_columns_statistics https://github.com/datafuselabs/databend/pull/15909

youngsofun:
the effect is verified by

https://github.com/datafuselabs/databend/pull/15880

PR #15909, opened 2024-06-26
refactor: optimize gen_columns_statistics() for scalar.

youngsofun:

## Summary

part of https://github.com/datafuselabs/databend/issues/15908

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15909)


src/query/storages/fuse/src/statistics/column_statistic.rs
```diff
@@ -162,31 +182,35 @@ pub mod traverse {
...
         for (idx, entry) in columns.iter().enumerate() {
             let data_type = &entry.data_type;
-            let column = entry.value.as_column().unwrap();
-            traverse_recursive(Some(idx), column, data_type, &mut leaves)?;
+            match &entry.value {
+                Value::Scalar(s) => {
+                    traverse_scalar_recursive(Some(idx), s, data_type, &mut leaves)?;
+                }
+                Value::Column(c) => {
+                    traverse_column_recursive(Some(idx), c, data_type, &mut leaves)?;
```
b41sh, line 202: It looks like `traverse_scalar_recursive` and `traverse_column_recursive` have a similar code structure, we can merge them into one function using `Value` as the argument and return type.

src/query/storages/fuse/src/statistics/column_statistic.rs
```diff
@@ -74,71 +74,91 @@ pub fn gen_columns_statistics(
...
-                }
-            } else {
-                continue;
+                    (1, 0)
+                };
+                let col_stats = ColumnStatistics::new(
+                    s.clone(),
+                    s.clone(),
+                    unset_bits as u64,
+                    0,
```
b41sh, line 88: the scalar value will be converted to a column when write to parquet file, so the `in_memory_size` should be the memory size of scalar value multiply by `rows`.

src/query/storages/fuse/src/statistics/column_statistic.rs
youngsofun: not all the same,  let me try to see if it is better.

src/query/storages/fuse/src/statistics/column_statistic.rs
youngsofun: the field is called in_memory_size, so not need to  multiply by rows?

src/query/storages/fuse/src/statistics/column_statistic.rs
youngsofun: I prefer current way, which is more clear to me.

traverse_column_recursive never call traverse_scalar_recursive
traverse_scalar_recursive call traverse_column_recursive some times

otherwise we still need to match Value in the unified func.

src/query/storages/fuse/src/statistics/column_statistic.rs
youngsofun: fixed

b41sh (approved):
LGTM, please fix the CI tests

PR #15910, opened 2024-06-27
refactor: optimize building bloom index for scalar.

youngsofun:

## Summary

1.   DataBlock::concat is optimized when all block is scalar for some column.
2.  building of bloom index only take one block as input for now,  so no need to concat,  so the `convert_to_full` can be removed too.
3.  not need to generate bloom index for scalar column, it will never be used since range index is checked first.

part of https://github.com/datafuselabs/databend/issues/15908

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15910)

### DIFF
--- a/src/query/expression/src/kernels/concat.rs
+++ b/src/query/expression/src/kernels/concat.rs
@@ -88,13 +88,12 @@ impl DataBlock {
         );
 
         let entry0 = blocks[0].get_by_offset(column_index);
-        if matches!(entry0.value, Value::Scalar(_)) {
-            if blocks
+        if matches!(entry0.value, Value::Scalar(_))
+            && blocks
                 .iter()
                 .all(|b| b.get_by_offset(column_index) == entry0)
-            {
-                return Ok(entry0.value.clone());
-            }
+        {
+            return Ok(entry0.value.clone());
         }
 
         let columns_iter = blocks.iter().map(|block| {
--- a/src/query/expression/src/kernels/concat.rs
+++ b/src/query/expression/src/kernels/concat.rs
@@ -87,13 +87,13 @@ impl DataBlock {
                 .all_equal()
         );
 
-        if let Value::Scalar(v0) = &blocks[0].get_by_offset(column_index).value {
-            let v0 = Value::Scalar(v0.clone());
+        let entry0 = blocks[0].get_by_offset(column_index);
+        if matches!(entry0.value, Value::Scalar(_)) {
             if blocks
                 .iter()
-                .all(|b| b.get_by_offset(column_index).value == v0)
+                .all(|b| b.get_by_offset(column_index) == entry0)
             {
-                return Ok(v0);
+                return Ok(entry0.value.clone());
             }
         }
 
--- a/src/query/expression/src/kernels/concat.rs
+++ b/src/query/expression/src/kernels/concat.rs
@@ -59,40 +59,54 @@ impl DataBlock {
         if blocks.is_empty() {
             return Err(ErrorCode::EmptyData("Can't concat empty blocks"));
         }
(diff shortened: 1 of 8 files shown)


src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -262,17 +262,16 @@ impl BloomIndex {
...
-                    let column = Column::concat_columns(source_columns_iter)?;
+                    let value = DataBlock::concat_columns(data_blocks_tobe_indexed, index)?;
 
-                    if Self::check_large_string(&column) {
+                    if let Value::Column(column) = &value {
+                        if !Self::check_large_string(&column) {
+                            continue;
+                        }
+                        (column.clone(), field_type.clone())
+                    } else {
```
b41sh, line 272: We can ignore scalar value to generate bloom filters, because `min` and `max` have the same values, range pruner can filter out the results without bloom pruner.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -262,17 +262,16 @@ impl BloomIndex {
                     }
                 }
                 _ => {
-                    let source_columns_iter = data_blocks_tobe_indexed.iter().map(|block| {
-                        let value = &block.get_by_offset(index).value;
-                        value.convert_to_full_column(field_type, block.num_rows())
-                    });
-                    let column = Column::concat_columns(source_columns_iter)?;
+                    let value = DataBlock::concat_columns(data_blocks_tobe_indexed, index)?;
```
b41sh, line 265: The `data_blocks_tobe_indexed` always have only one block, we can modify the argument as `&DataBlock` and don't need call `concat_columns` function.

src/query/expression/src/kernels/concat.rs
```diff
@@ -59,41 +59,55 @@ impl DataBlock {
...
 
-        let concat_columns = (0..blocks[0].num_columns())
-            .map(|i| {
-                debug_assert!(
-                    blocks
-                        .iter()
-                        .map(|block| &block.get_by_offset(i).data_type)
-                        .all_equal()
-                );
-
```
b41sh, line 75: We can check scalar value here like this, which is more clear.
```rust
let entry = blocks[0].get_by_offset(i);
if matches!(entry.value, Value::Scalar(_)) {
    if blocks
        .iter()
        .all(|b| b.get_by_offset(i) == entry)
    {
        return Ok(entry.clone());
    }
}
```

b41sh: add some unit test for this case is better.

youngsofun: done, thank you

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -183,50 +183,49 @@ impl BloomIndex {
...
 
         let mut filter_fields = vec![];
         let mut filters = vec![];
         let mut column_distinct_count = HashMap::<usize, usize>::new();
         for (index, field) in bloom_columns_map.into_iter() {
-            let field_type = &data_blocks_tobe_indexed[0].get_by_offset(index).data_type;
+            let field_type = &block.get_by_offset(index).data_type;
             if !Xor8Filter::supported_type(field_type) {
                 continue;
             }
```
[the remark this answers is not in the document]
youngsofun: done, thank you

b41sh (approved):

PR #15919, opened 2024-06-28
feat(query): add option [ IGNORE | RESPECT NULLS ] functionality for window rank function(first_value, last_value, nth_value)

### LINKED ISSUE #15881: Feature: Add { IGNORE | RESPECT } NULLS support for some ranking window functions
**Summary**

Currently, Databend doesn't support the `{ IGNORE | RESPECT } NULLS` syntax for ranking window functions. This limits the flexibility in handling NULL values within window frames. 

This feature request proposes adding support for `{ IGNORE | RESPECT } NULLS` to the following ranking window functions:

*  `FIRST_VALUE()`
*  `LAG()`
*  `LAST_VALUE()`
*  `LEAD()`
*  `NTH_VALUE()`

Similar to Snowflake and PostgreSQL, this syntax will allow users to explicitly control how NULL values are handled within the window frame, providing more consistent and predictable behavior.

**Expected Behavior:**

* When `{ IGNORE | RESPECT } NULLS` is not specified, the default behavior should be consistent with the SQL standard.
* For functions like `LAG()` and `LEAD()`, `IGNORE NULLS` will skip NULL values and fetch the next/previous non-NULL value within the window frame.
* For functions like `FIRST_VALUE()`, `LAST_VALUE()`, and `NTH_VALUE()`, `IGNORE NULLS` will attempt to find the first/last/nth non-NULL value within the window frame. If all values are NULL, they should return NULL even if `IGNORE NULLS` is specified.

**Benefits:**

* Enhanced SQL compatibility with other database systems.
* More flexibility and control over handling NULL values in ranking window functions.
* Improved consistency and predictability in query results.

TCeason:

## Summary

Implementing the [ IGNORE | RESPECT NULLS ] functionality for the lead, lag, first_value, last_value, nth_value window functions.

Default option is `respect nulls`.

When this option is used with `FIRST_VALUE`, the function returns the first value in the `frame` that is not NULL (or NULL if all values are NULL).

When this option is used with `LAST_VALUE`, The function returns the last value in the `frame` that is not NULL (or NULL if all values are NULL).

When this option is used with `NTH_VALUE`,  should skip null values in the determination of which row to use. 

Fixes #15881

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15919)


TCeason:
Now if max_block_size = 1, ignore nulls will return different result. It's an other window function bug. So I will fix it and then continue to do this feat.

src/query/ast/src/parser/expr.rs
```diff
@@ -1003,11 +1008,18 @@ pub fn expr_element(i: Input) -> IResult<WithSpan<ExprElement>> {
...
             ~ "(" ~ DISTINCT? ~ #comma_separated_list0(subexpr(0))? ~ ")"
+            ~ (( IGNORE | RESPECT ) ~ NULLS)?
             ~ (OVER ~ #window_spec_ident)
         },
-        |(name, _, opt_distinct, opt_args, _, window)| ExprElement::FunctionCall {
+        |(name, _, opt_distinct, opt_args, _, opt_ignore_null, window)| ExprElement::FunctionCall {
             func: FunctionCall {
                 distinct: opt_distinct.is_some(),
+                window_ignore_null: match name.to_string().as_str().to_lowercase().as_str() {
+                    "first_value" | "first" | "last_value" | "last" | "nth_value" => {
```
TCeason, line 1018: Now we support ignore nulls option. But sum() ignore nulls will also parse success. How to only support these `"first_value" | "first" | "last_value" | "last" | "nth_value"` support respect/ignore nulls ? cc @andylokandy

src/query/ast/src/ast/expr.rs
```diff
@@ -835,6 +835,7 @@ pub struct FunctionCall {
     pub name: Identifier,
     pub args: Vec<Expr>,
     pub params: Vec<Expr>,
+    pub window_ignore_null: Option<bool>,
```
b41sh, line 838: How about put `window_ignore_null` in the `Window` struct as a field `ignore_null`? This can make the definition of `FunctionCall` look simpler.

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -652,6 +647,58 @@ impl<T: Number> TransformWindow<T> {
         }
         false
     }
+
+    #[inline]
+    fn find_nth_non_null_value(
+        &self,
+        mut cur: RowPtr,
+        arg: usize,
```
b41sh, line 655: better named as `index` or `arg_index`.

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -652,6 +647,58 @@ impl<T: Number> TransformWindow<T> {
...
+        ignore_null: bool,
+        advance: bool,
+    ) -> Scalar {
+        let block = &self.blocks.get(cur.block - self.first_block).unwrap().block;
+        let mut col = block.get_by_offset(arg).to_column(block.num_rows());
```
b41sh, line 660: The return value is a `Scalar`, we can deal with `Scalar` argument separately, without having to convert it to a column.

src/query/ast/src/ast/expr.rs
TCeason: Now the window only contain `over` part. But ignore nulls is function part options.

TCeason:
```
#[derive(Debug, Clone, PartialEq, EnumAsInner, Drive, DriveMut)]
pub enum Window {
    WindowReference(WindowRef),
    WindowSpec(WindowSpec),
}
```

I think ignore nulls is not suitable put in enum Window.

b41sh: we can define `Window` as a struct, and use current enum as a field in `Window`.

TCeason: Done. Please review.

b41sh (approved):

src/query/service/src/pipelines/processors/transforms/window/transform_window.rs
```diff
@@ -652,6 +647,79 @@ impl<T: Number> TransformWindow<T> {
         }
         false
     }
+
+    #[inline]
+    fn find_nth_non_null_value(
```
sundy-li, line 652: the function accept a arg named `ignore_null`, so it's name should not be  non_null.

sundy-li (approved):

TCeason:
@soyeric128 doc needed

PR #15929, opened 2024-06-29
feat: failsafe tool `fuse_amend`

dantengsky:

## Summary

introduce new table func `system$fuse_amend`, which recovers data from s3, if bucket versioning is enabled:

if table `db.t` 's data is removed from s3 for some reason, and the bucket that stores the table data (and meta data) has enabled versioning,  the table data (table meta data) could be recovered by :

~~~
call system$fuse_amend('db', 't');
~~~ 

- As OpenDAL lacks the `list_object_versions` api (and get object by version_id), to implement the feature, aws rust sdk is currently used
  - only s3 compatible storage type is supported currently
  - bucket versioning should be enabled
- data that created before bucket version is enabled could NOT be recovered by `fuse_amend`

- only snapshot, segment, block will be recovered in this PR

e.g.

~~~
create table t(a int) 's3://testbucket/test_amend/' connection=(access_key_id ='minioadmin' secret_access_key ='minioadmin' endpoint_url='http://127.0.0.1:9900');

-- pointing to the location that contains `t`'s data
CREATE or replace stage test_amend_stage URL='s3://testbucket/test_amend/' connection=(access_key_id ='minioadmin' secret_access_key ='minioadmin' endpoint_url='http://127.0.0.1:9900');

-- generate some data
insert into t select * from numbers(11);
insert into t select * from numbers(11);
insert into t select * from numbers(11);

-- this will delete objects that of table t
remove @test_amend_stage;

-- error   ... NotFound (persistent) .... will be thrown
select * from t;

-- amend table
call system$fuse_amend('test_failsafe', 't');

-- things should be OK now
select sum(a) from t;

~~~

-----------

Versioning for test minio bucket is enabled in .github/actions/test_stateful_standalone_linux/action.yml

-----------

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15929)


src/query/ee/src/fail_safe/handler.rs
```diff
@@ -0,0 +1,385 @@
...
+
+impl Amender {
+    async fn try_new(storage_param: StorageParams) -> Result<Self> {
+        // TODO
+        // - replace client with opendal operator
+        // - supports other storage types
+        // when `list_object_versions` or higher level api is supported by opendal
+        if let StorageParams::S3(s3_config) = storage_param {
+            if s3_config.access_key_id.is_empty() {
+                panic!()
```
SkyFan2002, line 96: It would better to add detailed error messages

dantengsky: :+1:  thanks

src/query/ee/src/fail_safe/handler.rs
```diff
@@ -0,0 +1,384 @@
...
+                full_path = if self.root.ends_with('/') {
+                    format!("{}{}", self.root, key)
+                } else {
+                    format!("{}/{}", self.root, key)
+                };
+                &full_path
+            }
+        };
+
+        // trim the leading "/"
```
SkyFan2002, line 268: Why is leading "/" trimmed? (I'm not very familiar with this part)

dantengsky: oh, that is a defensive check that prevent "malformed" key prefix passed in from being used as list prefix

dantengsky: maybe it is too conservative？

SkyFan2002 (approved):

PR #15880, opened 2024-06-24
feat: orc support missing fields.

youngsofun:
I hereby agree to the terms of the CLA available at: https://docs.databend.com/dev/policies/cla/

## Summary

compare only fields instead.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)
- [ ] New Feature (non-breaking change which adds functionality)
- [ ] Breaking Change (fix or feature that could cause existing functionality not to work as expected)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15880)

PR #15939, opened 2024-07-01
refactor: speed up ColumnBuilder::repeat(Scalar::Null).

youngsofun:

### Problem Description

The original implementation calls push_default n times, which is slow because each call is recursive and includes pattern matching.

### Impact

The repeat method is used when flattening Value::Scalar to Value::Column. This approach significantly affects or even dominates overall performance when there are many null columns.

related issue: https://github.com/datafuselabs/databend/issues/15908
https://github.com/datafuselabs/databend/pull/15880 will be faster after this

### Improvement

This change introduces an ColumnBuilder::repeat_null that create the whole builder at once;

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15939)


src/query/expression/src/types/binary.rs
```diff
@@ -346,6 +346,14 @@ impl BinaryColumnBuilder {
         }
     }
 
+    pub fn repeat_null(n: usize) -> Self {
+        BinaryColumnBuilder {
+            data: vec![],
+            offsets: vec![0; n + 1],
```
sundy-li, line 352: it's not null, it's empty.

sundy-li: let's rename it to `repeat_default`

sundy-li:
can we add another method named `push_defaults(len: usize)`

```
 for _ in 0..n {
                        builder.push_default();
                    }
```

--->

```
builder.push_defaults(n);
```

youngsofun:
> can we add another method named `push_defaults(len: usize)`
> 
> ```
> for _ in 0..n {
>    builder.push_default();
> }
> ```
> 
> --->
> 
> ```
> builder.push_defaults(n);
> ```


I have considered this approach  as well

1.  less straight forward and efficient when use in repeat
2.  no usage of push_defaults(n); with n > 1 found, except in repeat().

image

src/query/expression/src/types/binary.rs
youngsofun: done

andylokandy (approved):

b41sh (approved):

sundy-li (approved):

PR #15946, opened 2024-07-02
refactor(query): use shuffle on distributed merge into

Freejww:

## Summary
`enable_distributed_merge_into = 1` by use shuffle and set `enable_merge_into_row_fetch = 1` as default.  Target table only has row_id and join columns during hash join, which reduce shuffle cost significantly.

Benefits:
1. Decoupling the Relationship Between Merge Into Processor and Hash Join Processor: Merge Into no longer needs to fetch data blocks from the Hash Table. This allows both Hash Join Spill and distributed Merge Into to be enabled simultaneously.
2. Enabling Distributed Execution of Merge Into: By avoiding row_number deduplication, the row_number inserted by each node are unique. This ensures that tasks such as serializing data blocks and writing data blocks to S3 are performed in parallel across multiple nodes.

# Performance Test(Databend cloud):
### target_table size = 600037902

```
merge into lineitem as target using small_lineitem_50000000 as source on target.l_orderkey = source.l_orderkey and source.l_linenumber = target.l_linenumber when matched then update set target.l_quantity = source.l_quantity + 1 when not matched then insert *;
```
image

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15946)


xudong963:
What size is your target table in the performance test.

Freejww:
> What size is your target table in the performance test.



> What size is your target table in the performance test.

count(*): 600037902

xudong963:
Better to add test queries in summary

src/query/sql/src/planner/optimizer/optimizer.rs
```diff
@@ -480,32 +480,19 @@ async fn optimize_merge_into(mut opt_ctx: OptimizerContext, s_expr: SExpr) -> Re
...
-                .merge_source_matcher
-                .matches(&join_s_expr)
+    let distributed = !join_s_expr.has_merge_exchange();
+    if opt_ctx.enable_distributed_optimization && distributed {
+        let merge_optimizer = MergeOptimizer::create();
+        // now always use shuffle
+        if matches!(
+            join_op.join_type,
+            JoinType::Left | JoinType::LeftAnti | JoinType::Inner
+        ) && merge_optimizer.merge_matcher.matches(&join_s_expr)
```
xudong963, line 490: Why non-right join to enforce shuffle join?

We should respect the join distribution type(shuffle or broadcast).

Dousir9: Yes, if the join type is inner join, we can use broadcast join if the source table is on the build side.

src/query/sql/src/planner/optimizer/optimizer.rs
```diff
@@ -483,10 +483,11 @@ async fn optimize_merge_into(mut opt_ctx: OptimizerContext, s_expr: SExpr) -> Re
     let distributed = !join_s_expr.has_merge_exchange();
     if opt_ctx.enable_distributed_optimization && distributed {
         let merge_optimizer = MergeOptimizer::create();
-        // now always use shuffle
+        // If inner join && !join_order_changed, source will broadcast,
+        // others use shuffle.
         if matches!(
             join_op.join_type,
-            JoinType::Left | JoinType::LeftAnti | JoinType::Inner
+            JoinType::Left | JoinType::LeftAnti | JoinType::Inner if join_order_changed
```
xudong963, line 490: Why need to add the condition, can we directly use the cbo's distribution result?

Dousir9:
Rerun wizard test on ci-cloud (medium).

src/query/sql/src/planner/optimizer/optimizer.rs
Freejww: If the target is used as the build side, the CBO will make the target to broadcast, which will cause deduplication checks to occur in the main node.

Freejww:
> Rerun wizard test on ci-cloud (medium).

test has passed

src/query/sql/src/executor/physical_plans/physical_merge_into_add_row_number.rs
```diff
@@ -1,80 +0,0 @@
...
-use databend_common_expression::ROW_NUMBER_COL_NAME;
-
-use crate::executor::PhysicalPlan;
-use crate::executor::PhysicalPlanBuilder;
-use crate::optimizer::ColumnSet;
-use crate::optimizer::SExpr;
-
-// add row_number for distributed merge into
-#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
-pub struct MergeIntoAddRowNumber {
```
Dousir9, line 34: If there is a broadcast join, we cannot keep the insert data in the local node, we need to shuffle the source table after the join, which requires the `row_number` and some deduplication operations.

src/query/sql/src/planner/optimizer/optimizer.rs
```diff
@@ -480,32 +480,20 @@ async fn optimize_merge_into(mut opt_ctx: OptimizerContext, s_expr: SExpr) -> Re
...
-                .merge_source_matcher
-                .matches(&join_s_expr)
+    let distributed = !join_s_expr.has_merge_exchange();
+    if opt_ctx.enable_distributed_optimization && distributed {
+        let merge_optimizer = MergeOptimizer::create();
+        // If inner join && !join_order_changed, source will broadcast,
+        // others use shuffle.
+        if matches!(
+            join_op.join_type,
+            JoinType::Left | JoinType::LeftAnti | JoinType::Inner if join_order_changed
```
Dousir9, line 490: If there is a broadcast join and the merge type is `MatchedOnly`, maybe we can remove the shuffle processor(after join, shuffle by row_id) and keep the update data in the local node, because the target data is divided by partition.

Dousir9 (approved):
Rest LGTM !

xudong963 (approved):

PR #15126, opened 2024-03-29
fix(query): tuple field name can not contain special char

TCeason:
I hereby agree to the terms of the CLA available at: https://docs.databend.com/dev/policies/cla/

## Summary

This bug maybe cause session panic:

```sql
create table t(d tuple("A-a" int, b int));
```

panicked at src/query/ast/src/parser/parser.rs:67:13:

```txt
SyntaxException. Code: 1005, Text = error: 
  --> SQL:1:27
  |
1 | CREATE TABLE tc (d TUPLE(A-a Int32, b Int32))
  | ------           - -----  ^ unexpected `-`, expecting `BOOLEAN`, `BOOL`, `UINT8`, `TINYINT`, `UINT16`, `SMALLINT`, `UINT32`, `INT`, `INTEGER`, `UINT64`, `UNSIGNED`, `BIGINT`, `INT8`, `INT16`, `INT32`, `INT64`, `SIGNED`, `FLOAT32`, `FLOAT`, `FLOAT64`, `DOUBLE`, `DECIMAL`, `ARRAY`, `MAP`, `BITMAP`, `TUPLE`, `DATE`, `DATETIME`, `TIMESTAMP`, `BINARY`, `VARBINARY`, `LONGBLOB`, `MEDIUMBLOB`, `TINYBLOB`, `BLOB`, `STRING`, `VARCHAR`, `CHAR`, `CHARACTER`, `TEXT`, `VARIANT`, `JSON`, `GEOMETRY`, or `NULLABLE`
  | |                | |       
  | |                | while parsing TUPLE(<name> <type>, ...)
  | |                | while parsing type name
  | |                while parsing `<column name> <type> [DEFAULT <expr>] [AS (<expr>) VIRTUAL] [AS (<expr>) STORED] [COMMENT '<comment>']`
  | while parsing `CREATE [OR REPLACE] TABLE [IF NOT EXISTS] [<database>.]<table> [<source>] [<table_options>]`

```

### Modify

Modify tuple type parser. 

Add a check, if tuple filed name contain char that is not alphanumeric characters, will directly return error.

- Fixes #15125 
- 
## Tests

- [x] Unit Test

## Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)

PR #15965, opened 2024-07-03
feat: allow tuple field name to contain '_'.

youngsofun:

## Summary

this pr https://github.com/datafuselabs/databend/pull/15126/files limit it to is_ascii_alphanumeric()
but we often need '_'.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15965)


TCeason:
LGTM but I think we should add some tests. If someone destroys this we can trigger it

b41sh:
The bug here is due to `Tuple`'s `fields_name` is defined as `String`, resulting in missing quote-related information. We should fix this by defining `fields_name` as `Identifier`. In order to solve the problem of current use, we can merge this PR first and make further modifications later.

b41sh (approved):

PR #15978, opened 2024-07-04
refactor: optimize take_by_slices_limit_from_blocks

forsaken628:

## Summary
column clone incurs a significant performance overhead

[pprof.cpu.before.pb.gz](https://github.com/user-attachments/files/16100160/pprof.cpu.before.pb.gz)
before

[pprof.cpu.after.pb.gz](https://github.com/user-attachments/files/16100159/pprof.cpu.after.pb.gz)
after

### data set
tpc-h 10G
### sql
select * from orders order by o_orderdate desc, o_totalprice desc ignore_result;

## Tests
- [ ] Unit Test
- [ ] Logic Test
- [x] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/15978)


src/query/expression/src/kernels/take_chunks.rs
```diff
@@ -171,19 +171,44 @@ impl DataBlock {
...
+            let block_columns = blocks[*block_index].columns();
+            let len = (*len).min(remain);
+            remain -= len;
 
-            result_columns.push(merged_col);
+            for (col_index, builder) in builders.iter_mut().enumerate() {
+                let BlockEntry { data_type, value } = &block_columns[col_index];
+                match value {
+                    Value::Scalar(scalar) => {
+                        let other = ColumnBuilder::repeat(&scalar.as_ref(), len, data_type);
```
sundy-li, line 194: here we just malloc a temp vector.

Maybe we can refactor `repeat`
 to be
```
 pub fn push_repeat(&mut self, scalar: &ScalarRef, n: usize, data_type: &DataType) {
```

sundy-li: so does `repeat_default`

forsaken628: The well-implemented `push_repeat` is a bit complicated, I wrote a quick fix. `push_repeat` will be implemented separately later.

src/query/expression/src/kernels/take_chunks.rs
```diff
@@ -171,19 +171,46 @@ impl DataBlock {
...
+            remain -= len;
 
-            result_columns.push(merged_col);
+            for (col_index, builder) in builders.iter_mut().enumerate() {
+                let BlockEntry { value, .. } = &block_columns[col_index];
+                match value {
+                    Value::Scalar(scalar) => {
+                        let scalar = scalar.as_ref();
+                        for _ in 0..len {
+                            builder.push(scalar.clone())
```
sundy-li, line 196: Here may introduce temp scalar clone inside a loop

src/query/expression/src/values.rs
```diff
@@ -1846,6 +1846,75 @@ impl ColumnBuilder {
         }
     }
 
+    pub fn push_repeat(&mut self, item: ScalarRef, n: usize) {
```
forsaken628, line 1849: Where to put unit tests?

sundy-li: expression/tests

sundy-li (approved):

src/query/expression/src/values.rs
sundy-li: so we can replace `ColumnBuilder::repeat`  with

```
let mut builder = ColumnBuilder::with_capacity(...);
builder.push_repeat(item, n);
```

src/query/expression/src/values.rs
```diff
@@ -1830,22 +1778,76 @@ impl ColumnBuilder {
...
+                builder.push_repeat(value, n);
+            }
+            (ColumnBuilder::Bitmap(builder), ScalarRef::Bitmap(value)) => {
+                BitmapType::push_item_repeat(builder, value, n);
+            }
+            (ColumnBuilder::Nullable(builder), ScalarRef::Null) => {
+                NullableType::push_item_repeat(builder, None, n);
+            }
+            (ColumnBuilder::Nullable(builder), value) => {
+                builder.push_repeat(value.clone(), n);
```
forsaken628, line 1833: `ScalarRef` or `&ScalarRef`? I've spent quite a bit of time here, and still haven't found a complete solution to the problem

sundy-li: ScalarRef is ok。

src/query/expression/src/values.rs
```diff
@@ -1506,62 +1506,9 @@ impl ColumnBuilder {
     }
 
     pub fn repeat(scalar: &ScalarRef, n: usize, data_type: &DataType) -> ColumnBuilder {
```
Dousir9, line 1508: Using `push_repeat` to implement the `repeat` function does not seem to be a good idea, because it will reduce performance, it is recommended to keep the original implementation. :)

forsaken628: I've checked the implementations of `push_repeat` and `repeat` in most of builder and there's almost no difference, maybe I'm missing something?

src/query/expression/src/types/binary.rs
```diff
@@ -456,6 +449,15 @@ impl BinaryColumnBuilder {
         self.data.get_unchecked(start..end)
     }
 
+    pub fn push_repeat(&mut self, item: &[u8], n: usize) {
+        self.data.reserve(item.len() * n);
+        self.offsets.reserve(n);
```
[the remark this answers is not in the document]
forsaken628: I don't understand.
```rust
#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn basic() {
        let mut builder = BinaryColumnBuilder::with_capacity(3, 1);
        assert_eq!(builder.offsets.capacity(), 4);

        builder.put_str(&"abc");
        builder.commit_row();

        builder.push_repeat("hello".as_bytes(), 10);

        assert_eq!(builder.offsets.capacity(), 12); // 1+1+10
    }
}
```

src/query/expression/src/values.rs
Dousir9: For example, for nullable columns, we can construct a `Bitmap` directly, but `push_repeat` will call `push_null`.

src/query/expression/src/types/binary.rs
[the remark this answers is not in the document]
Dousir9: The capacity required for the `offsets` is n + 1, but the code is `reserve(n)`.

src/query/expression/src/values.rs
Dousir9: You can check out the official Rust documentation: https://doc.rust-lang.org/std/vec/struct.Vec.html
```rs
let vec = vec![0; 5];
assert_eq!(vec, [0, 0, 0, 0, 0]);

// The following is equivalent, but potentially slower:
let mut vec = Vec::with_capacity(5);
vec.resize(5, 0);
assert_eq!(vec, [0, 0, 0, 0, 0]);
```

forsaken628: Yes, 10% slower roughly
```rust
#[cfg(test)]
mod tests {
    extern crate test;
    use test::Bencher;

    use super::*;

    #[bench]
    fn bench_repeat(b: &mut Bencher) {
        let scalar = Scalar::String(String::from("hello"));
        let scalar = &scalar.as_ref();

        b.iter(|| {
            let builder = ColumnBuilder::repeat(scalar, 100, &DataType::String);
            builder.build()
        })
    }

    #[bench]
    fn bench_repeat2(b: &mut Bencher) {
        let scalar = Scalar::String(String::from("hello"));
        let scalar = &scalar.as_ref();

        b.iter(|| {
            let n = 100;
            let mut builder = ColumnBuilder::with_capacity_hint(&DataType::String, n, false);
            builder.push_repeat(scalar, n);
            builder.build()
        });
    }
}
```

src/query/expression/src/types/binary.rs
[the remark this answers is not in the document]
forsaken628: I think there's no problem here. "Reserves capacity for at least `additional` more elements to be inserted in the given `Vec<T>`"

Here it's additional capacity, not total capacity. The initial 0 is put in at initialization time.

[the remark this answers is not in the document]
Dousir9: Remove this line ？

src/query/expression/src/types/binary.rs
```diff
@@ -331,19 +334,14 @@ impl BinaryColumnBuilder {
...
-            data.extend_from_slice(scalar);
-        }
-        let offsets = once(0)
-            .chain((0..n).map(|i| (len * (i + 1)) as u64))
-            .collect();
-        BinaryColumnBuilder {
-            data,
-            offsets,
+        let mut builder = BinaryColumnBuilder {
+            data: Vec::new(), // lazy allocate
```
Dousir9, line 338: We already know the size of the `scalar` and the number of repetitions `n`, what is the benefit of lazy allocate ?

src/query/expression/src/types/binary.rs
```diff
@@ -456,6 +454,22 @@ impl BinaryColumnBuilder {
...
+        if self.need_estimated && self.offsets.len() - 1 < 64 {
+            self.data.reserve(item.len() * n);
+            for _ in 0..n {
+                self.data.extend_from_slice(item);
+                self.commit_row();
+            }
+        } else {
+            let start = self.data.len();
+            let len = item.len();
+            self.data.extend(item.iter().cloned().cycle().take(len * n));
```
Dousir9, line 467: Does it perform better than `data.extend_from_slice(scalar);`? If `n` is 1, there will be a clone for `[u8]`

src/query/expression/src/types/binary.rs
forsaken628: `vec.extend` will always get the size from `iterator.size_hint` to allocate memory, there's no need to do it manually.
https://github.com/rust-lang/rust/blob/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/alloc/src/vec/mod.rs#L3008-L3016

Dousir9: Why not just use `with_capacity`.

src/query/expression/src/types/binary.rs
forsaken628: It's a mistake.

forsaken628:
> @forsaken628 We cannot use `Vec` `resize` to implement `push_item_repeat`, because **it can cause wrong results**, for example, if we want to push `n` elements, if `n` is less than the number of original elements in `Vec` (`len`), then `Vec` will lose the original elements and be truncated to `n`, while the correct size should be `len + n`.

always `builder.resize(builder.len() + n, item);`

src/query/expression/src/types/binary.rs
```diff
@@ -456,6 +457,24 @@ impl BinaryColumnBuilder {
         self.data.get_unchecked(start..end)
     }
 
+    pub fn push_repeat(&mut self, item: &[u8], n: usize) {
+        self.data.reserve(item.len() * n);
```
Dousir9, line 461: `builder.reserve(self.data.len() + item.len() * n)`, rest LGTM !

Dousir9, line 461: `builder.reserve(self.data.len() + item.len() * n)`.

src/query/expression/src/types/string.rs
```diff
@@ -596,6 +596,24 @@ impl StringColumnBuilder {
         std::str::from_utf8_unchecked(bytes)
     }
 
+    pub fn push_repeat(&mut self, item: &str, n: usize) {
+        self.data.reserve(item.len() * n);
```
Dousir9, line 600: Ditto.

Dousir9 (approved):
Rest LGTM, thanks for your contribution !

src/query/expression/src/types/binary.rs
forsaken628: ## pub fn [reserve](https://doc.rust-lang.org/beta/alloc/vec/struct.Vec.html#method.reserve)(&mut self, additional: [usize](https://doc.rust-lang.org/beta/core/primitive.usize.html))

Reserves capacity for at least additional more elements to be inserted in the given Vec<T>. The collection may reserve more space to speculatively avoid frequent reallocations.

After calling reserve, capacity will be greater than or equal to `self.len() + additional`.

Does nothing if capacity is already sufficient.

## Examples
```rust
// len: 1, cap:1, cap-len:0 no space for more element
let mut vec = vec![1]; 

// reserves additional 10,
// then capacity >=11 ( 1(len)+10(additional) )
// so we can insert 10 additional more elements without allocation
vec.reserve(10);  
assert!(vec.capacity() >= 11);
```
## Src
```rust
impl<T, A: Allocator> RawVec<T, A> {
    /// Returns if the buffer needs to grow to fulfill the needed extra capacity.
    /// Mainly used to make inlining reserve-calls possible without inlining `grow`.
    fn needs_to_grow(&self, len: usize, additional: usize) -> bool {
        additional > self.capacity().wrapping_sub(len)
    }
```

PR #16007, opened 2024-07-09
chore: orc and parquet use option missing_field_as.

youngsofun:

## Summary

do some refactor for code reuse by the way
-  parquet and orc share the same fn project_columnar
-  rowbased files and orc share the SingleFilePartition

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16007)


src/query/storages/common/stage/src/read/columnar/projection.rs
```diff
@@ -0,0 +1,124 @@
...
+                        let default_values = &default_values.as_deref().expect(
+                            "default_values should not be none when miss_field_as=FIELD_DEFAULT",
+                        );
+                        default_values[i].as_expr(&BUILTIN_FUNCTIONS)
+                    }
+                }
+            }
+        };
+        if !matches!(expr, Expr::Constant { .. }) {
+            num_inputs += 1;
```
b41sh, line 114: we can't check input by `Expr::Constant`, because default value may be a function. `num_inputs` should increase by one if the `pos` is some.

b41sh: Alternatively, we can use `pushdown_columns` instead of `num_inputs`. If `pushdown_columns` is empty, it means `num_inputs` is 0.

youngsofun: fixed, thank you!

b41sh (approved):

Issue #15963, opened 2024-07-03
Improve optimizer histogram


xudong963:
**Summary**

Currently, the optimizer's histogram assumes the data is uniformly distributed.

But for tpcds data, some columns are non-uniform, such as `ss_sold_date_sk`, will result in bad join order

![image](https://github.com/datafuselabs/databend/assets/41979257/fd108ebe-b7b9-4099-b1c7-59abf6d99690)

PR #16018, opened 2024-07-10
feat: generate accurate histogram

xudong963:

### Accurate histogram
The PR introduces the accurate histogram during `analyze table []` and it will be committed to meta. 
Each column in the table will have a separate pipeline to generate histogram info.  

The reason why not use a union all query to get all columns' histogram info is that the column data type may be different.

The pipeline is from the query:
```sql
WITH s_quantiles AS (
    select ss_sold_date_sk, NTILE(100) OVER (ORDER BY ss_sold_date_sk) AS quantile 
    from store_sales where ss_sold_date_sk is distinct from NULL
)

SELECT quantile,
       COUNT(DISTINCT ss_sold_date_sk) AS ndv,
       MAX(ss_sold_date_sk) AS max_value,
       MIN(ss_sold_date_sk) AS min_value,
       COUNT() AS count
FROM s_quantiles
GROUP BY quantile
ORDER BY quantile;
```
Firstly, it will split the column into 100 parts by `NTILE` window function, the compute min, max, ndv, count for each bucket.

Note: computing the histogram for each column is heavy(it's possible to OOM), so I added a setting to decide if analyzing the histogram during analyzing table(If the data is uniform distribution, don't need the accurate histogram, the original way is enough).

About 90% of the cost comes from the window function. So we can dig into window optimization to improve the performance of analyzing histograms.
image

### Cardinality estimation
Refactor join, filter cardinality estimation to avoid the case of `-- -> +`, main change from `update statistics`. For selectivity, if a column of a table participates in the filter, the statistics of this column need to be updated, If the selection rate is less than 0.8, the histogram of all columns is invalidated, otherwise, the original histogram is maintained.

### TPCH join order
fix tpch q19 join order

### A case for accurate histogram
```
# accurate histogram
root@localhost:8000/tpcds> explain SELECT count(*)
   FROM store_sales,
        date_dim
   WHERE ss_sold_date_sk = d_date_sk;

EXPLAIN
SELECT
  count(*)
FROM
  store_sales,
  date_dim
WHERE
  ss_sold_date_sk = d_date_sk

AggregateFinal
├── output columns: [COUNT(*) (#51)]
├── group by: []
├── aggregate functions: [count()]
├── estimated rows: 1.00
└── AggregatePartial
    ├── group by: []
    ├── aggregate functions: [count()]
    ├── estimated rows: 1.00
    └── HashJoin
        ├── output columns: []
        ├── join type: INNER
        ├── build keys: [date_dim.d_date_sk (#23)]
        ├── probe keys: [store_sales.ss_sold_date_sk (#0)]
        ├── filters: []
        ├── estimated rows: 2750554.00
        ├── TableScan(Build)
        │   ├── table: default.tpcds.date_dim
        │   ├── output columns: [d_date_sk (#23)]
        │   ├── read rows: 73049
        │   ├── read size: 195.33 KiB
        │   ├── partitions total: 1
        │   ├── partitions scanned: 1
        │   ├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 1 to 1>]
        │   ├── push downs: [filters: [], limit: NONE]
        │   └── estimated rows: 73049.00
        └── TableScan(Probe)
            ├── table: default.tpcds.store_sales
            ├── output columns: [ss_sold_date_sk (#0)]
            ├── read rows: 2880404
            ├── read size: 719.54 KiB
            ├── partitions total: 7
            ├── partitions scanned: 7
            ├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 7 to 7>]
            └── estimated rows: 2880404.00

36 rows explain in 0.132 sec. Processed 0 rows, 0 B (0 rows/s, 0 B/s)

# no accurate histogram
root@localhost:8000/tpcds> set enable_analyze_histogram = 0;

SET
  enable_analyze_histogram = 0

0 row read in 0.030 sec. Processed 0 row, 0 B (0 row/s, 0 B/s)

root@localhost:8000/tpcds> analyze table store_sales;

ANALYZE TABLE store_sales

0 row read in 0.203 sec. Processed 0 row, 0 B (0 row/s, 0 B/s)

root@localhost:8000/tpcds> analyze table date_dim;

ANALYZE TABLE date_dim

0 row read in 0.263 sec. Processed 0 row, 0 B (0 row/s, 0 B/s)

root@localhost:8000/tpcds> explain SELECT count(*)
   FROM store_sales,
        date_dim
   WHERE ss_sold_date_sk = d_date_sk;

EXPLAIN
SELECT
  count(*)
FROM
  store_sales,
  date_dim
WHERE
  ss_sold_date_sk = d_date_sk

AggregateFinal
├── output columns: [COUNT(*) (#51)]
├── group by: []
├── aggregate functions: [count()]
├── estimated rows: 1.00
└── AggregatePartial
    ├── group by: []
    ├── aggregate functions: [count()]
    ├── estimated rows: 1.00
    └── HashJoin
        ├── output columns: []
        ├── join type: INNER
        ├── build keys: [date_dim.d_date_sk (#23)]
        ├── probe keys: [store_sales.ss_sold_date_sk (#0)]
        ├── filters: []
        ├── estimated rows: 2906384.77
        ├── TableScan(Build)
        │   ├── table: default.tpcds.date_dim
        │   ├── output columns: [d_date_sk (#23)]
        │   ├── read rows: 73049
        │   ├── read size: 195.33 KiB
        │   ├── partitions total: 1
        │   ├── partitions scanned: 1
        │   ├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 1 to 1>]
        │   ├── push downs: [filters: [], limit: NONE]
        │   └── estimated rows: 73049.00
        └── TableScan(Probe)
            ├── table: default.tpcds.store_sales
            ├── output columns: [ss_sold_date_sk (#0)]
            ├── read rows: 2880404
            ├── read size: 719.54 KiB
            ├── partitions total: 7
            ├── partitions scanned: 7
            ├── pruning stats: [segments: <range pruning: 1 to 1>, blocks: <range pruning: 7 to 7>]
            └── estimated rows: 2880404.00

36 rows explain in 0.119 sec. Processed 0 rows, 0 B (0 rows/s, 0 B/s)

# real count
root@localhost:8000/tpcds> SELECT count(*)
   FROM store_sales,
        date_dim
   WHERE ss_sold_date_sk = d_date_sk;

SELECT
  count(*)
FROM
  store_sales,
  date_dim
WHERE
  ss_sold_date_sk = d_date_sk

┌──────────┐
│ count(*) │
│  UInt64  │
├──────────┤
│  2750554 │
└──────────┘
1 row read in 0.563 sec. Processed 2.95 million row, 11.62 MiB (5.25 million row/s, 20.65 MiB/s)

```

### Chore
1. remove tpcds join order file, after join order is stable, we can add it again

related to https://github.com/datafuselabs/databend/issues/15963

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16018)


src/query/service/src/interpreters/interpreter_table_analyze.rs
```diff
@@ -144,46 +181,163 @@ impl Interpreter for AnalyzeTableInterpreter {
...
+                            plan.table,
+                            c.1,
+                        )
+                    })
+                    .collect::<Vec<_>>();
+                for (sql, (col_id, _)) in histogram_sqls.into_iter().zip(index_cols.iter()) {
+                    info!("Analyze histogram via sql {:?}", sql);
+                    let (mut histogram_plan, bind_context) = self.plan_sql(sql).await?;
+                    if !self.ctx.get_cluster().is_empty() {
+                        histogram_plan = remove_exchange(histogram_plan);
```
sundy-li, line 228: why do we need remove_exchange recursively?

xudong963: The exchange manager does not yet support the ability to build multiple pipelines for the same query ctx.

src/query/sql/src/planner/optimizer/property/selectivity.rs
```diff
@@ -292,14 +289,8 @@ impl<'a> SelectivityEstimator<'a> {
...
                 column_stat.ndv = new_ndv;
-                if let Some(histogram) = &mut column_stat.histogram {
-                    if new_ndv as u64 <= 2 {
-                        column_stat.histogram = None;
-                    } else {
-                        for bucket in histogram.buckets.iter_mut() {
-                            bucket.update(selectivity);
-                        }
-                    }
+                if selectivity < 0.8 {
```
Dousir9, line 292: `if selectivity < 0.2` ?

xudong963: If there is a little data filtered, assume the histogram is still valid.

Dousir9: Why not update the histogram and keep it.

xudong963: The ndv will be approximated again and the histogram will be calculated based on the approximated ndv, the error in the histogram will increase, leading to an increased risk of -- -> +

Dousir9: We can avoid this problem by pruning the buckets in the histogram according to the filter predicate, which means we need to implement a histogram whose number of buckets can be less than DEFAULT_HISTOGRAM_BUCKETS.

xudong963: Pruning and fewer buckets are not the key issue. The problem is that the min, max, ndv used to generate the histogram again are further approximated, and the histogram generation is also approximated, which leads to further errors in the lower bound, upper bound, and ndv of the new histogram's bucket being amplified. I don't see the need to regenerate the histogram at the moment, but if there is an actual case or job benchmark that has problems, I'll consider if I can address it further.

Dousir9: For example, if there are 100 buckets and the predicate a < 50 can prune 50 buckets, we can continue to use the remaining 50 buckets instead of discarding them.

xudong963: Oh, I got you. It's a good way to maintain the **current column**'s part histogram buckets, no regeneration, directly to prune and replace. Let's do it later.

src/query/sql/src/planner/optimizer/property/selectivity.rs
```diff
@@ -292,14 +289,8 @@ impl<'a> SelectivityEstimator<'a> {
...
+                    column_stat.histogram = None;
```
xudong963, line 293: This part is related to columns that **are not used** in a predicate. So there is a better to maintain their histogram, right? @Dousir9

Dousir9: I think it is better to keep these histograms, as these columns may be used in joins.

xudong963: These columns, do not directly by pruning buckets to generate a histogram, but need to use the series of errors I mentioned above to regenerate the histogram, the follow-up may be used, bringing the error, of course, directly with the ndv is also an error. There may be a better way, we need to investigate the business of mainstream databases, such as orcale, db2 way, and then decide the next step.

Dousir9: Yeah, we can temporarily keep the previous method, which is: we assume that the columns are not correlated and then scale down each bucket of the histogram by selectivity proportionally.

xudong963: I'd like to set them to None first, so that it's easier to determine the root cause of the problem when doing job/tpcds benchmark later. This prevents the problem from being let go because of `-- -> +`.

Dousir9: In the previous implementation, this histogram will also be discarded after it is used in join (used once or multiple times, depending on whether this column is used in join equi conditions), can you give an example of `-- -> +`.

Dousir9 (approved):

Issue #4620, opened 2022-03-30
SQL Statement Support for preparetion


wubx:
**Summary**

Some drivers and connectors support the ability to send a SQL statement for preparation before execution.  
Such as :
 -  SELECT
 -  DML
 -  show <objects>

Relation issue：
[#4618](https://github.com/datafuselabs/databend/issues/4618)

wubx:
MySQL prepare statement:
```
PREPARE stmt_name FROM preparable_stmt
```

execute Statement:
```
EXECUTE stmt_name
    [USING @var_name [, @var_name] ...]
```

sundy-li:
No plan to support prepare stmt in SERVER.  It should be done on the client side.

PR #16030, opened 2024-07-11
feat(parser): support parsing placeholder

lewiszlw:

## Summary

Support `?` placeholder.
Test not passed, could someone give it a fast review as I'm not very familiar about some detail code?
```
+---------- Input ----------
+select * from t where a = ?
+---------- Output ---------
+error:
+  --> SQL:1:27
+  |
+1 | select * from t where a = ?
+  | ------                -   ^ missing lhs or rhs for the binary operator
+  | |                     |
+  | |                     while parsing expression
+  | while parsing `SELECT ...`
+
```

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16030)

### DIFF
--- a/src/query/ast/src/ast/expr.rs
+++ b/src/query/ast/src/ast/expr.rs
@@ -227,6 +227,11 @@ pub enum Expr {
         span: Span,
         name: String,
     },
+    /// `?` or `$` Prepared statement arg placeholder
+    Placeholder {
+        span: Span,
+        name: String,
+    },
 }
 
 impl Expr {
@@ -262,7 +267,8 @@ impl Expr {
             | Expr::DateAdd { span, .. }
             | Expr::DateSub { span, .. }
             | Expr::DateTrunc { span, .. }
-            | Expr::Hole { span, .. } => *span,
+            | Expr::Hole { span, .. }
+            | Expr::Placeholder { span, .. } => *span,
         }
     }
 
@@ -396,6 +402,7 @@ impl Expr {
             } => merge_span(merge_span(*span, interval.whole_span()), date.whole_span()),
             Expr::DateTrunc { span, date, .. } => merge_span(*span, date.whole_span()),
             Expr::Hole { span, .. } => *span,
+            Expr::Placeholder { span, .. } => *span,
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/query/ast/src/ast/expr.rs
```diff
@@ -227,6 +227,11 @@ pub enum Expr {
         span: Span,
         name: String,
     },
+    /// `?` or `$` Prepared statement arg placeholder
+    Placeholder {
+        span: Span,
+        name: String,
```
forsaken628, line 233: Consider adding an offset field, which would make the parameter binding easier to implement

FYI:
https://github.com/matrixorigin/matrixone/blob/bd8054941cfdab4cd94d17cd4260d11833a102b2/pkg/sql/parsers/dialect/mysql/mysql_lexer.go#L83-L86

https://github.com/matrixorigin/matrixone/pull/3957/files#diff-6f8d2e3805b41b9704c5d638c08078e8a7e5f6812ffa344ad607746319026e1fR856

b41sh:
Hi @lewiszlw thanks for your contribution.
`?` is also used by `JsonOperator::Question`, and `?` is first matched by `json_op` function, so `placeholder` is never come into effect. We need add some extra work to check whether `?` is between two exprs to convert `JsonOperator::Question` to `Placeholder`.

sundy-li:
Is is used to prepare stmt?  We don't have a plan to support prepare stmt

lewiszlw:
Thanks for your help.
Yes, I added this on my fork for prepared stmt. Closing the pr as you don't have plan to support prepared stmt.

forsaken628:
> Is is used to prepare stmt like mysql ? We don't have a plan to support prepare stmt in databend.
> 
> It could be supported on client side (JDBC...)

But here's an open issue #4620

sundy-li:
The issue is outdated.  Sorry for the confusion.

PR #16032, opened 2024-07-11
chore(query): add full expr in error message

sundy-li:

## Summary

add full expr in error message
## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16032)


src/query/expression/src/evaluator.rs
```diff
@@ -125,6 +125,18 @@ impl<'a> Evaluator<'a> {
 
     pub fn run(&self, expr: &Expr) -> Result<Value<AnyType>> {
         self.partial_run(expr, None, &mut EvaluateOptions::default())
+            .map_err(|err| {
+                if err
+                    .message()
+                    .contains(format!("`{}`", expr.sql_display()).as_str())
```
zhang2014, line 131: Please add large expr test to avoid stack overflow

src/query/functions/tests/it/scalars/testdata/array.txt
```diff
@@ -11,7 +11,7 @@ error:
   --> SQL:1:2
   |
 1 | ['a', 1]
-  |  ^^^ invalid digit found in string while evaluating function `to_uint64('a')` in expr `to_uint64('a')`
+  |  ^^^ invalid digit found in string while evaluating function `to_uint64('a')` in expr `to_uint64('a')`, during run expr: `array(to_uint64('a'), to_uint64(1))`
```
b41sh, line 14: `in expr` seems duplicate with `during run expr`, may be we can remove `in expr` here?

sundy-li: It's not duplicated, this case is using literal in expr.


see:

```
  |              ^ divided by zero while evaluating function `divide(1, 0)` in expr `1 / expr_a`, during run expr: `if(CAST(cond_a AS Boolean NULL), 1 / expr_a, to_float64(expr_else))`

```

PR #16049, opened 2024-07-15
feat: add table function `fuse_vacuum2()`

SkyFan2002:

## Summary
This PR introduces a new table function `fuse_vacuum2()` aimed at enhancing performance.

## Object Key of Snapshot v5, Segment, and Block

- **UUID Version:** Use UUID v7
  - Lexicographically ordered
  - Timestamp can be extracted from the object key
- **Prefix:** Prefixed with the character 'g', ensuring all keys of v5 are larger than v4's in lexicographical order

For segments/blocks that are newly created in an operation based on snapshot `s`, the timestamp embedded in their object keys equals `s.timestamp`. If there is no base snapshot, let `ts` be `0`. For snapshots, the timestamp embedded in their object keys equals to their own timestamp.

## New Field `lvt` in Snapshot v5

For all snapshots v5 (`s`), the candidate value (`lvt_candidate`) of `s.lvt` is calculated as `s.timestamp - settings.get_retention_period()`. `s.lvt` may be adjusted to larger values based on `lvt_candidate`, in cases where the retention period has been tweaked or due to clock skews. For example, if `lvt_candidate <= s.prev.lvt`, then `s.lvt` could be set to `s.prev.lvt + 1`. `s.lvt` should respect the retention period setting at the time. Decreasing or setting `s.lvt` equal to the `lvt` of the previous snapshot is not allowed.

### Properties

- `s.lvt <= s.timestamp`
- `s.lvt > s.prev.lvt`
- If `s` is based on another snapshot `s'`: `s.lvt <= s'.timestamp` (enforced during commitment: transactions with stale base snapshots are not allowed).

### Notes

- `Some(_) > None`
- `s.prev` may not equal `s'`
## Steps of Vacuum2
```Rust
fn vacuum2(){
  let lvt = set_lvt();
  let snapshots_before_lvt = list_until(snapshot_dir,lvt);
  let gc_root = select_gc_root(snapshots_before_lvt);
  let snapshots_to_gc = &snapshots_before_lvt[..gc_root_idx];
  
  let lvt = gc_root.lvt;
  let segments_before_gc_root = list_until(segment_dir,lvt);
  let segments_to_gc = segments_before_gc_root - gc_root.segments;

  let blocks_before_gc_root = list_until(block_dir,lvt);
  let blocks_to_gc = blocks_before_gc_root - gc_root.blocks;

  remove(snapshots_to_gc,segments_to_gc,blocks_to_gc);
}
```

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [x] Long run

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16049)


src/query/ee/src/storages/fuse/operations/vacuum_table_v2.rs
```diff
@@ -0,0 +1,320 @@
...
+        slice_summary(&blocks_to_gc)
+    );
+
+    let start = std::time::Instant::now();
+    let files_to_gc = snapshots_to_gc
+        .iter()
+        .chain(segments_to_gc.iter())
+        .chain(blocks_to_gc.iter())
+        .map(|p| p.to_string())
+        .collect::<Vec<_>>();
```
dantengsky, line 169: also need to delete the bloom filter index, inverted index and agg_indexes files

https://github.com/datafuselabs/databend/blob/d6eaadb875a2906f8d303c3fa7682fee1635a052/src/query/storages/fuse/src/operations/gc.rs#L472-L480

src/query/ee/src/storages/fuse/operations/vacuum_table_v2.rs
```diff
@@ -0,0 +1,320 @@
...
+        ver,
+        put_cache: false,
+    };
+    reader.read(&params).await
+}
+
+async fn select_gc_root<'a>(
+    fuse_table: &FuseTable,
+    snapshots_before_lvt: &'a [String],
+) -> Result<Option<(Arc<TableSnapshot>, &'a [String])>> {
```
dantengsky, line 273: if session.retention_period .. is 0, we can/should use the current snapshot as GC root

zhyass (commented):
Almost LGTM, it would be much better to have test cases compatible with older versions, like https://github.com/datafuselabs/databend/blob/main/src/query/service/tests/it/storages/fuse/operations/gc.rs#L252

src/query/ee/src/storages/fuse/operations/vacuum_table_v2.rs
```diff
@@ -0,0 +1,403 @@
...
+        put_cache: false,
+    };
+    reader.read(&params).await
+}
+
+async fn select_gc_root<'a>(
+    fuse_table: &FuseTable,
+    snapshots_before_lvt: &'a [String],
+    is_vacuum_all: bool,
+) -> Result<Option<(Arc<TableSnapshot>, &'a [String])>> {
```
dantengsky, line 332: As discussed, the process of locating the GC root can be simplified:

- If the retention period is zero, we can use the current snapshot as the GC root, since it is already committed (thus safe).
- The `for` loop is unnecessary, as it does not affect the eventual nature of the GC process.

dantengsky:
fix merge:
[bd8dfe3](https://github.com/databendlabs/databend/pull/16049/commits/bd8dfe391c7e3c33af6837c8fd171910ae5a447d) :+1:

dantengsky (approved):
let's roll

PR #16050, opened 2024-07-15
refactor: Simplify HashMethod.build_keys_state function signature by replacing &[(Column, DataType)] with InputColumns

forsaken628:

## Summary
follow #15991 #16038
Simplify HashMethod.build_keys_state function signature by replacing &[(Column, DataType)] with InputColumns

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16050)


Dousir9:
Did you know that column clone has an O(1) time complexity? Some of the code in the PR is less readable. @forsaken628

src/query/service/src/pipelines/processors/transforms/aggregator/transform_aggregate_partial.rs
```diff
@@ -240,22 +241,18 @@ impl<Method: HashMethodBounds> TransformPartialAggregate<Method> {
...
-
-        let group_columns = self
-            .params
-            .group_columns
-            .iter()
-            .map(|&index| block.get_by_offset(index))
-            .map(|c| (c.value.as_column().unwrap().clone(), c.data_type.clone()))
-            .collect::<Vec<_>>();
+        let group_columns = InputColumns::new_block_proxy(&self.params.group_columns, &block);
+        let data_types = group_columns.as_block().unwrap().data_types();
```
Dousir9, line 245: This does not avoid the overhead of generating a `Vec`, so no performance improvement.

forsaken628:
Because Column clone contains atomic operations, it is not always without visible performance overhead.
#15991

However, the key point here is really about maintaining a relatively consistent interface, not a performance issue.

Dousir9:
> Because Column clone contains atomic operations, it is not always without visible performance overhead. #15991
> 
> However, the key point here is really about maintaining a relatively consistent interface, not a performance issue.

The overhead of `Column.clone` is almost negligible. 
The most important thing is that the code in `hash_join_probe_state.rs` is less readable.

Dousir9:
> Because Column clone contains atomic operations, it is not always without visible performance overhead. #15991
> 
> However, the key point here is really about maintaining a relatively consistent interface, not a performance issue.

Got it, please continue your PR and correct your PR title, BTW, the overhead of `Arc.clone()` is almost negligible when there are only a few calls.

sundy-li:
since column has `data_type()` method, so maybe we can remove the `DataType` in function's signature.

Dousir9 (approved):
LGTM !

PR #16066, opened 2024-07-17
fix: bitmap with empty buffer stands for empty.

youngsofun:

## Summary

For the RoaringTreeMap itself, an empty set is encoded as [0; 8], representing a size of 0_u64. The push_default and repeat_default methods intend to use an empty buffer as an empty bitmap. However, this is only checked in a few instances where RoaringTreemap::deserialize_from is called.

I believe using an empty buffer as an empty bitmap is reasonable, so we should always check for it.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16066)


src/query/expression/src/types/bitmap.rs
```diff
@@ -175,3 +178,15 @@ impl ArgType for BitmapType {
         BinaryColumnBuilder::with_capacity(capacity, 0)
     }
 }
+
+pub fn deserialize_bitmap(buf: &[u8]) -> databend_common_exception::Result<RoaringTreemap> {
```
b41sh, line 182: better to move this function to https://github.com/datafuselabs/databend/blob/main/src/common/io/src/bitmap.rs

youngsofun: done

PR #16072, opened 2024-07-18
refactor: load parquet use own can_cast_to()

youngsofun:

## Summary

This PR updates the dependency from Arrow’s can_cast_types to our own can_cast_to() function.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16072)

### DIFF
--- a/src/query/expression/src/type_check.rs
+++ b/src/query/expression/src/type_check.rs
@@ -744,3 +744,72 @@ pub const ALL_SIMPLE_CAST_FUNCTIONS: &[&str] = &[
 pub fn is_simple_cast_function(name: &str) -> bool {
     ALL_SIMPLE_CAST_FUNCTIONS.contains(&name)
 }
+
+/// notes:
+/// - this is NOT can_auto_cast_to.
+/// - this should be consistent with run_cast, partially ensured by test: test_can_cast_to .
+pub fn can_cast_to(src_type: &DataType, dest_type: &DataType) -> bool {
+    use DataType::*;
+    if src_type == dest_type {
+        return true;
+    }
+    // we mainly care about which types can/cannot cast to dest_type.
+    // the match is written in a way to make it easier to read this info.
+    // try to use less _ to avoid miss something
+    match (dest_type, src_type) {
+        (String | Variant, _) => true,
+
+        (Number(_), Binary | Bitmap) => false,
+        (Number(_), _) => true,
+
+        (Nullable(_), Null) => true,
+        (Nullable(box dest_ty), Nullable(box src_ty))
+        | (dest_ty, Nullable(box src_ty))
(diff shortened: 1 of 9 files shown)


src/query/expression/src/type_check.rs
```diff
@@ -745,3 +744,72 @@ pub const ALL_SIMPLE_CAST_FUNCTIONS: &[&str] = &[
...
+/// - this should be consistent with run_cast, partially ensured by test: test_can_cast_to .
+pub fn can_cast_to(src_type: &DataType, dest_type: &DataType) -> bool {
+    use DataType::*;
+    if src_type == dest_type {
+        return true;
+    }
+    // we mainly care about which types can/cannot cast to dest_type.
+    // the match is written in a way to make it easier to read this info.
+    // try to use less _ to avoid miss something
+    match (dest_type, src_type) {
```
b41sh, line 759: seems put `src_type` on left is more customary.

src/query/expression/src/type_check.rs
```diff
@@ -745,3 +744,72 @@ pub const ALL_SIMPLE_CAST_FUNCTIONS: &[&str] = &[
 pub fn is_simple_cast_function(name: &str) -> bool {
     ALL_SIMPLE_CAST_FUNCTIONS.contains(&name)
 }
+
+/// notes:
+/// - this is NOT can_auto_cast_to.
+/// - this should be consistent with run_cast, partially ensured by test: test_can_cast_to .
+pub fn can_cast_to(src_type: &DataType, dest_type: &DataType) -> bool {
```
b41sh, line 751: looks duplicate with `can_auto_cast_to` function, may be we can add a `AutoCastRules` for parquet load case?

src/query/expression/src/type_check.rs
youngsofun: the comments explain the reason.

```
// we mainly care about which types can/cannot cast to dest_type.
// the match is written in a way to make it easier to read this info.
```

youngsofun: To be more specific, our use case is as follows: we have a dest type and several src types. We need to determine if the cast between them is valid.

src/query/expression/src/type_check.rs
andylokandy: What is the difference between `can_auto_cast_to` and `can_cast_to`? We should write it down in the comment so that we know how to edit this function later.

youngsofun: @andylokandy  done, please review again cc @b41sh

Issue #16143, opened 2024-07-29
"IN" operator not accepting hardcoded values more than 1023


Shanmugavel-J:
**Summary**

Not able to query a table with IN operator and having more than approx 1023 values

**_QUERY:_** `SELECT * FROM default.test WHERE payload:test IN (n1, n2, n3, n4 .... n1023);`

SQL Error [1007]: Query failed (#cab09f1f-6294-415c-a4b0-b28f13af8069): Cannot find common type for String and Nullable(Variant)

Note: This issue comes up only when providing more than 1023 if less it works fine.

PR #16159, opened 2024-07-31
fix(query): fix in operator convert to subquery cast failed

b41sh:

## Summary

fix in operator convert to subquery cast failed, list values data type wrap nullable so that expression and list values can have common data type.

- fixes: #16143

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16159)

### DIFF
--- a/tests/sqllogictests/suites/query/select.test
+++ b/tests/sqllogictests/suites/query/select.test
@@ -195,8 +195,37 @@ select * from t1 where a in (1, 2) order by a;
 1 2
 2 3
 
+# fix: https://github.com/datafuselabs/databend/issues/16143
+statement ok
+drop table if exists t2;
+
+statement ok
+create table t2 (id int, payload variant);
+
+statement ok
+insert into t2 values (1, '{"test":10,"test2":"ab"}'), (2, '{"test":20,"test2":"cd"}');
+
+query IT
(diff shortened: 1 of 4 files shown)


tests/sqllogictests/suites/query/select.test
```diff
@@ -195,8 +195,37 @@ select * from t1 where a in (1, 2) order by a;
 1 2
 2 3
 
+# fix: https://github.com/datafuselabs/databend/issues/16143
```
xudong963, line 198: Will the test be converted to a subquery from inlist?  The default value is 1024.

```
 ("inlist_to_join_threshold", DefaultSettingValue {
                    value: UserSettingValue::UInt64(1024),
                    desc: "Set the threshold for converting IN list to JOIN.",
                    mode: SettingMode::Both,
                }),
```

b41sh: yes, we have `set inlist_to_join_threshold = 1;` on line 190.

xudong963 (approved):

PR #16179, opened 2024-08-04
feat: A columnar Geometry/Geography format

forsaken628:

## Summary
A columnar Geometry/Geography format is proposed here to provide support for the storage and computation of geospatial features.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16179)


drmingdrmer:
I'm requested to review this PR but I'm not very familiar with the geo part. Can somebody else review this one? @sundy-li

b41sh:
Hi, @forsaken628  Thanks for your contribution.
We consider using [geoarrow](https://geoarrow.org/) to store `Geometry` and `Geography` type values in the future, because the value in `geoarrow` is in a columnar data layout, it can have better performance, and it is easier to integrate with the current type system. Using `GeoBuf` will make future migration difficult, so I recommend storing the `Geography` type values in the `WKB` format like `Geometry`.

forsaken628:
>  it can have better performance,

I don't think so

https://docs.rs/geoarrow/latest/geoarrow/array/struct.WKBArray.html

> This array can be used directly for operations, but that will incur costly encoding to and from WKB on every operation. Instead, you usually want to use the WKBArray only for serialization purposes (e.g. to and from [GeoParquet](https://geoparquet.org/)) but convert to strongly-typed arrays (such as the [PointArray](https://docs.rs/geoarrow/latest/geoarrow/array/struct.PointArray.html)) for computations.

I don't think `geoarrow` is a good choice if we don't use strongly-typed arrays

forsaken628:
Actually `geobuf` is the right way to integrate geoarrow.

For example, if we want to calculate the [Center for PolygonArray](https://docs.rs/geoarrow/latest/geoarrow/array/struct.PolygonArray.html#impl-Center-for-PolygonArray%3CO%3E), if it is encoded in wkb, we need to deserialize each row one by one, then call the algorithm, and finally serialize it to wkb, which is only one step in the process of calculating it, what if there are multiple calculations? 

So how does `geobuf` do it? Let's take Polygon as an example.The memory layout of `geoarrow` is
 ```rust
pub struct SeparatedCoordBuffer {
    pub(crate) x: ScalarBuffer<f64>,
    pub(crate) y: ScalarBuffer<f64>,
}

pub struct PolygonArray<O: OffsetSizeTrait> {
    // Always GeoDataType::Polygon or GeoDataType::LargePolygon
    data_type: GeoDataType,

    pub(crate) metadata: Arc<ArrayMetadata>,

    pub(crate) coords: CoordBuffer,

    /// Offsets into the ring array where each geometry starts
    pub(crate) geom_offsets: OffsetBuffer<O>,

    /// Offsets into the coordinate array where each ring starts
    pub(crate) ring_offsets: OffsetBuffer<O>,

    /// Validity bitmap
    pub(crate) validity: Option<NullBuffer>,
}

```

Then the memory layout of the `geobuf` is
```rust
pub struct GeographyColumn {
    pub(crate) buf: BinaryColumn,
    pub(crate) offsets: Buffer<u64>,
    pub(crate) x: Buffer<f64>,
    pub(crate) y: Buffer<f64>,
}
```

`GeographyColumn.x` corresponds to `PolygonArray.coords.x` 
`GeographyColumn.y` corresponds to `PolygonArray.coords.y` 
`GeographyColumn.offsets` corresponds to `PolygonArray.ring_offsets`
Only `PolygonArray.geom_offsets` does not exist and needs to be reconstructed from the `buf`.

As you can see `geobuf` is flexible and efficient, `wkb` is bulky and inefficient.

b41sh:
> > it can have better performance,
> 
> I don't think so
> 
> https://docs.rs/geoarrow/latest/geoarrow/array/struct.WKBArray.html
> 
> > This array can be used directly for operations, but that will incur costly encoding to and from WKB on every operation. Instead, you usually want to use the WKBArray only for serialization purposes (e.g. to and from [GeoParquet](https://geoparquet.org/)) but convert to strongly-typed arrays (such as the [PointArray](https://docs.rs/geoarrow/latest/geoarrow/array/struct.PointArray.html)) for computations.
> 
> I don't think `geoarrow` is a good choice if we don't use strongly-typed arrays

We use `WKB` to implement `Geometry` mainly because it is relatively simple, using strong defined types will require a lot of work, but if the user has this need, we can also consider change the current implementation. `WKB` is also a subtype of [geoarrow extension types](https://github.com/geoarrow/geoarrow/blob/main/extension-types.md), and existing data can be easily migrated. If you want to compare performance, you should compare `geobuf` with `geoarrow`, not `WKB`.
Now the `geobuf` implementation needs to store more datas, take `Point` as example, `geoarrow` only needs to store `x` and `y` columns, but `geobuf` needs to store additional `offsets` and `binary` columns. This will take up more space.
Another issue to consider is that geoarrow's community ecology seems to be better, and may become a de facto standard in the future. If you have time, we can consider using `geoarrow` directly to implement `Geography`, of course, this will be more complicated, we need to implement the `Union` type first to support multiple sub types.
Of course, I am not an expert on `geoarrow` and `geobuf`. I just have a preliminary look at the documents. I will look at the documents in-depth and related codes in detail, and carefully think about the implementation, we can discuss them later.

forsaken628:
To list the current questions

**For a `st_makepoint` function, what type should it return?**
- `PointArray`, do we want to support subtypes?
- `MixedGeometryArray`, do we want to support `Union`? Why spread the impact to the whole project when it could be solved locally?
- `WKBArray`, do we want to accept costly encoding to and from `WKB` on every operation?

**Do we want to support GeoJson `Feature` type?**
This can simply be interpreted as adding a metadata to each Geometry, with support for storing arbitrary content. Although snowflake provides support for this, and I've implemented it in `geobuf`, I don't think it's worthwhile to implement this feature. It's also possible that I'm wrong and I don't understand the practical applications of the GeoJson Feature. Neither `WKB` nor `geoarrow` supports GeoJson `Feature`.

**`GeometryCollection` type introduces recursive type, and recursive type has no explicit number of columns, which is a big problem.**
https://github.com/geoarrow/geoarrow/pull/43#discussion_r1698572015
I don't think `geoarrow` has the will to solve this problem.
I've also noticed that in some other implementations, this support is not provided.
https://help.aliyun.com/document_detail/442156.html

Finally, I think we should treat `geoarrow` as a tool, not as a foundation, because `geoarrow` doesn't solve all problems.

forsaken628:
> Now the geobuf implementation needs to store more datas, take Point as example, geoarrow only needs to store x and y columns, but geobuf needs to store additional offsets and binary columns. This will take up more space.

This can be fine-tuned here.
`MultiPolygon` with three offsets (`geom_offsets`, `polygon_offsets`, `ring_offsets`) is the most complex case. The current implementation has one offsets on the outside and two offsets on the inside. Is it possible to change it to `Option<Vec<u64>>` ? Wouldn't it be possible to have two offsets outside and one inside? But without benchmarking, there is no way to know if these adjustments make sense.

kylebarron:
When you come back to geospatial support, I'd encourage you to consider GeoArrow again. It's slowly becoming more stable, and my Rust implementation is maturing.

PR #16194, opened 2024-08-06
refactor: improve network error handling.

drmingdrmer:

##### refactor: improve network error handling.

- Consider `Status::Cancelled` as retryable error, which is returned
  when timeout sending an RPC.

- Reduce timeout from 10 sec to 4 sec so that if there is an error retry
  can be done sooner.

- Extend client retry times from 2 to 4, to allows the client to try
  more endpoints when Leader switches.

- Add more logs to connection re-establish for better locating network
  issues.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Refactoring

## Related Issues


Xuanwo (approved):
Thanks!

PR #16198, opened 2024-08-06
feat: transfer leader command for meta-service

drmingdrmer:

##### feat: transfer leader command for meta-service

This commit introduce a new admin API(at `admin_api_address`, by default
`127.0.0.1:28002`) to force the current Leader to transfer leadership to
another node.  The url of this command is
"/v1/ctrl/trigger_transfer_leader?to=<new_leader_node_id>".

Such request returns nothing no matter leadership is successfully
transferred or not.

If the node receives this request is not the leader it does nothing.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16198)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11085,7 +11085,7 @@ dependencies = [
 [[package]]
 name = "openraft"
 version = "0.10.0"
-source = "git+https://github.com/drmingdrmer/openraft?tag=v0.10.0-alpha.3#94c820c3047aa07da41bb6bd1524f9dc00eeed02"
+source = "git+https://github.com/drmingdrmer/openraft?tag=v0.10.0-alpha.6#c6f4a779e0bb563788c4187381065162e76ea996"
 dependencies = [
  "anyerror",
  "byte-unit",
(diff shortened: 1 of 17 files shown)


everpcpc:
We could return an indicator whether the request is accepted?

drmingdrmer:
> We could return an indicator whether the request is accepted?

If the node that receives this request is Leader, it will always accept the request.
But accepting it does not mean the leadership is actually transferred to the designated  node. 
Are you suggesting the get a response that confirms leader has been transferred?

everpcpc:
> If the node that receives this request is Leader, it will always accept the request.

Yes, we need to know whether the request is sent to leader and accepted, or sent to a follower and dose nothing.

drmingdrmer:
> > If the node that receives this request is Leader, it will always accept the request.
> 
> Yes, we need to know whether the request is sent to leader and accepted, or sent to a follower and dose nothing.

Make sense

everpcpc:
Actually we usually do not care which node to transfer leader to. Maybe it's better to omit the argument `to` and start an election?

everpcpc (approved):

PR #16215, opened 2024-08-08
refactor(query): use bigint to handle the fallback of decimal op overflow

sundy-li:

## Summary

use bigint to handle the fallback of decimal op overflow

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16215)


src/query/expression/src/types/decimal.rs
```diff
@@ -687,19 +698,48 @@ impl Decimal for i256 {
...
+        })
     }
 
-    fn do_round_div(self, rhs: Self, mul: Self) -> Option<Self> {
-        if self.is_negative() == rhs.is_negative() {
+    fn do_round_div(self, rhs: Self, mul_scale: u32) -> Option<Self> {
+        let fallback = || {
+            let a = BigInt::from_le_bytes(&self.to_le_bytes());
+            let b = BigInt::from_le_bytes(&rhs.to_le_bytes());
+            let mul = BigInt::from(10).pow(mul_scale);
```
youngsofun, line 723: can be pre computed(maybe later)？the pow is implemented by multi mul_scale times, which is slow?

src/query/expression/src/types/decimal.rs
```diff
@@ -725,6 +765,32 @@ impl Decimal for i256 {
...
+    fn from_bigint(value: BigInt) -> Option<Self> {
+        let mut ret: u256 = u256::ZERO;
+        let mut bits = 0;
+
+        for i in value.iter_u64_digits() {
+            if bits >= 256 {
+                return None;
+            }
+            ret |= u256::from(i) << bits;
+            bits += 64;
```
youngsofun, line 777: least significant digit first.
the last i may only used a part of 64 bits?

src/query/expression/src/types/decimal.rs
sundy-li: fallback is already in a lambda function, so it's lazy computed.

src/query/expression/src/types/decimal.rs
sundy-li: if `bits` is larger than 256, it's overflow, so return none.

This method is powered by https://docs.rs/num-bigint/latest/src/num_bigint/bigint/convert.rs.html#86-92

youngsofun (approved):

PR #16220, opened 2024-08-08
feat: auth by refresh and session tokens.

youngsofun:

## Summary

add new endpoint   /v1/session/login   and  /v1/session/renew.

/v1/session/login return refresh_token and session_token
session_token is used to auth /v1/query,  TTL = 1h
refresh_token is used to auth  /v1/session/renew,  TTL=4h

token is verified by store its hash as key in meta, and is cached in memory.

todo（small enhancement）:
- [ ] make nonce safer
- [ ] check session_id or nonce in addition to `hash exists`

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)


drmingdrmer (commented):
If reusing `CrudMgr` is possible, it would greatly simplify things.

I haven't reviewed all the changes yet. I'm waiting for confirmation on the solution above.

> Reviewed 17 of 38 files at r1.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220#-:-O3mwvDKDC68lHYIDdxq:b-661aae)* status: 16 of 38 files reviewed, 5 unresolved discussions (waiting on @youngsofun)

___
*[`src/meta/app/src/principal/user_token.rs` line 24 at r1](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3muyHn97dXQxEIv3TX:-O3muyHn97dXQxEIv3TY:btqgum1) ([raw file](https://github.com/datafuselabs/databend/blob/b506740e91cea6e0f3f4dc10bb07ce06ce52be5f/src/meta/app/src/principal/user_token.rs#L24)):*
> ```XML
>     Refresh = 1,
>     Session = 2,
> }
> ```

What is Refresh token? It sounds like an action to refresh a token. What's the difference between theses two kind of token? Please add some doc to explain their purpose and usage scenario.



_Code quote:_
```XML
pub enum TokenType {
    Refresh = 1,
    Session = 2,
}
```
___
*[`src/meta/app/src/principal/user_token.rs` line 30 at r1](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3mvOY4--PB3OsOHe-9:-O3mvOY4--PB3OsOHe-A:undefined) ([raw file](https://github.com/datafuselabs/databend/blob/b506740e91cea6e0f3f4dc10bb07ce06ce52be5f/src/meta/app/src/principal/user_token.rs#L30)):*
> ```XML
>     pub token_type: TokenType,
>     // used to delete refresh token when close session
>     pub parent: Option<String>,
> ```





_Suggestion:_
```XML
    /// used to delete refresh token when close session
    pub parent: Option<String>,
```
___
*[`src/query/management/src/token/token_api.rs` line 19 at r1](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3mwZrs2jeYRAo0trmY:-O3mwZrtA225ofo6T2LF:baukkrd) ([raw file](https://github.com/datafuselabs/databend/blob/b506740e91cea6e0f3f4dc10bb07ce06ce52be5f/src/query/management/src/token/token_api.rs#L19)):*
> ```XML
> 
> #[async_trait::async_trait]
> pub trait TokenApi: Sync + Send {
> ```

If possible, re-use `CrudMgr` to implement a simple crud manager, such as `ConnectionMgr`
https://github.com/datafuselabs/databend/blob/98652d1a0fca41e8982b217362d9d1339300c86d/src/query/management/src/connection.rs#L18
___
*[`src/query/management/src/token/token_api.rs` line 30 at r1](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3msL9d71BeBFn0_AAT:-O3msL9d71BeBFn0_AAU:b5vqwfd) ([raw file](https://github.com/datafuselabs/databend/blob/b506740e91cea6e0f3f4dc10bb07ce06ce52be5f/src/query/management/src/token/token_api.rs#L30)):*
> ```XML
>     async fn get_token(&self, token_hash: &str) -> Result<Option<QueryTokenInfo>>;
> 
>     async fn drop_token(&self, token_hash: &str) -> Result<()>;
> ```

Add doc comment to the trait method explaining their purposes and expected behavior of the implementation.



_Code quote:_
```XML
    async fn upsert_token(
        &self,
        token_hash: &str,
        token_info: QueryTokenInfo,
        ttl_in_secs: u64,
        is_update: bool,
    ) -> Result<bool>;

    async fn get_token(&self, token_hash: &str) -> Result<Option<QueryTokenInfo>>;

    async fn drop_token(&self, token_hash: &str) -> Result<()>;
```
___
*[`src/query/management/src/token/token_mgr.rs` line 64 at r1](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3mtUJ88EHx9_WyUaAs:-O3mtUJ88EHx9_WyUaAt:bq2v3az) ([raw file](https://github.com/datafuselabs/databend/blob/b506740e91cea6e0f3f4dc10bb07ce06ce52be5f/src/query/management/src/token/token_mgr.rs#L64)):*
> ```XML
>     ) -> Result<bool> {
>         let ident = self.token_ident(token_hash);
>         let seq = MatchSeq::GE(if is_update { 1 } else { 0 });
> ```

`is_update` here is not very clear about its purpose:
`replace_existing` would be better reflecting its behavior.

Avoid using plain number duration. Use `Duration` type for `ttl`.



_Code quote:_
```XML
        is_update: bool,
    ) -> Result<bool> {
        let ident = self.token_ident(token_hash);
        let seq = MatchSeq::GE(if is_update { 1 } else { 0 });
```

youngsofun:
> If reusing `CrudMgr` is possible, it would greatly simplify things.

thanks， I wiil try it

youngsofun:
> Add doc comment to the trait method explaining their purposes and expected behavior of the implementation.

I will remove the trait.

youngsofun:
@drmingdrmer plz review again

src/query/service/src/servers/http/v1/session/login_handler.rs
```diff
@@ -34,8 +36,17 @@ struct LoginRequest {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-struct LoginResponse {
+pub struct LoginResponse {
```
sundy-li, line 39: should we change this to enum?
```
enum LoginResponse {
     Ok(...)
     Error(QueryError)
}
```

src/query/service/src/servers/http/v1/session/renew_handler.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::servers::http::v1::HttpQueryContext;
+use crate::servers::http::v1::QueryError;
+
+#[derive(Deserialize, Clone)]
+struct RenewRequest {
+    pub session_token: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct RenewResponse {
```
sundy-li, line 33: maybe enum is better?

src/query/service/src/servers/http/v1/session/login_handler.rs
youngsofun: version may be still helpful when error?

drmingdrmer (commented):
> Reviewed 19 of 38 files at r1, 2 of 2 files at r2, 3 of 3 files at r3, 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 11 unresolved discussions (waiting on @flaneur2020, @hantmac, @sundy-li, and @youngsofun)

___
*[`src/meta/app/src/principal/user_token.rs` line 30 at r1](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3mvOY4--PB3OsOHe-9:-O3oubZ79xMIJL3zTvIT:b-7ci587) ([raw file](https://github.com/datafuselabs/databend/blob/b506740e91cea6e0f3f4dc10bb07ce06ce52be5f/src/meta/app/src/principal/user_token.rs#L30)):*
> ```XML
>     pub token_type: TokenType,
>     // used to delete refresh token when close session
>     pub parent: Option<String>,
> ```

Use triple slash doc comment, so that IDE is able to display the explanation of these fields.
___
*[`src/meta/app/src/principal/user_token.rs` line 23 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3otZ6v7lvVkGuvXIaQ:-O3otZ6v7lvVkGuvXIaR:brobl4d) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/meta/app/src/principal/user_token.rs#L23)):*
> ```XML
> )]
> pub enum TokenType {
>     // refresh token, with a longer TTL, is only used for auth when get new refresh token and session token.
> ```

Is this explanation correct?



_Suggestion:_
```XML
    /// A token that grants the privilege to a user for refreshing `Session` and `Refresh` token,
    /// with a longer TTL than `Session` token so that when `Session` token expires,
    /// user is still able to obtain a refreshed token with this one.
```
___
*[`src/query/service/src/servers/http/v1/session/token_manager.rs` line 48 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3ovTKl32YT2aHoDn1W:-O3ovTKl32YT2aHoDn1X:b-ia2tnj) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/token_manager.rs#L48)):*
> ```XML
> /// by adding SESSION_TOKEN_VALIDITY_IN_SECS, avoid touch refresh token for each request.
> /// note the ttl is not accurate, the TTL is in fact longer (by 0 to 1 hour) then expected.
> const REFRESH_TOKEN_TTL: u64 = REFRESH_TOKEN_VALIDITY_IN_SECS + SESSION_TOKEN_VALIDITY_IN_SECS;
> ```

Always use `std::time::Duration` instead if possible.



_Code quote:_
```XML
/// target TTL
pub const REFRESH_TOKEN_VALIDITY_IN_SECS: u64 = 3600 * 4; // 4 hours
pub const SESSION_TOKEN_VALIDITY_IN_SECS: u64 = 3600; // 1 hour

/// to cove network latency, retry and time skew
const TOKEN_TTL_DELAY: u64 = 300;
/// in case of client retry, shorted the ttl instead of drop at once
/// only required for refresh token.
const TOKEN_DROP_DELAY: u64 = 90;

/// by adding SESSION_TOKEN_VALIDITY_IN_SECS, avoid touch refresh token for each request.
/// note the ttl is not accurate, the TTL is in fact longer (by 0 to 1 hour) then expected.
const REFRESH_TOKEN_TTL: u64 = REFRESH_TOKEN_VALIDITY_IN_SECS + SESSION_TOKEN_VALIDITY_IN_SECS;
```
___
*[`src/query/service/src/servers/http/v1/session/token_manager.rs` line 91 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3owKWX-mb7LviQcTGF:-O3owKWX-mb7LviQcTGG:bljxnbw) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/token_manager.rs#L91)):*
> ```XML
>         // those infos are set when the request is authed
>         let tenant = session.get_current_tenant();
>         let tenant_name = tenant.tenant.to_string();
> ```

Use method `tenant_name()` to get the string name, the internal structure may be refactored in future.



_Suggestion:_
```XML
        let tenant_name = tenant.tenant_name().to_string();
```
___
*[`src/query/service/src/servers/http/v1/session/token_manager.rs` line 107 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3oxDdUBC7afsv1wzIZ:-O3oxDdUBC7afsv1wzI_:bfzsymc) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/token_manager.rs#L107)):*
> ```XML
>         let meta_store = UserApiProvider::instance();
> 
>         let now = unix_ts();
> ```

Make `unix_ts()` return a `Duration`, convert to integer only when `Duration` can not be used.
___
*[`src/query/users/src/user_token.rs` line 52 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3owq1mDDUjAvCk-fQw:-O3owq1mDDUjAvCk-fQx:b-6wte3g) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/users/src/user_token.rs#L52)):*
> ```XML
>         token_api_provider.drop_token(token_hash).await
>     }
> }
> ```

These methods are not very necessary. They can be replaced with indirect calls such as: `UserApiProvider::token_api().upsert_token()`.



_Code quote:_
```XML
impl UserApiProvider {
    #[async_backtrace::framed]
    pub async fn upsert_token(
        &self,
        tenant: &Tenant,
        token_hash: &str,
        token_info: QueryTokenInfo,
        ttl_in_secs: u64,
        update_only: bool,
    ) -> Result<bool> {
        let token_api_provider = self.token_api(tenant);
        token_api_provider
            .upsert_token(token_hash, token_info, ttl_in_secs, update_only)
            .await
    }

    #[async_backtrace::framed]
    pub async fn get_token(
        &self,
        tenant: &Tenant,
        token_hash: &str,
    ) -> Result<Option<QueryTokenInfo>> {
        let token_api_provider = self.token_api(tenant);
        token_api_provider.get_token(token_hash).await
    }

    #[async_backtrace::framed]
    pub async fn drop_token(&self, tenant: &Tenant, token_hash: &str) -> Result<()> {
        let token_api_provider = self.token_api(tenant);
        token_api_provider.drop_token(token_hash).await
    }
}
```
___
*[`src/query/service/src/servers/http/v1/session/login_handler.rs` line 39 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#gh-1710512083:-O3osWd8A3sQQRIYo8s3:b4t1ktx) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/login_handler.rs#L39)):*
<details><summary><i>Previously, youngsofun (Yang Xiufeng) wrote…</i></summary><blockquote>

version may be still helpful when error?
</blockquote></details>

This struct can be replaced with `Result<OKResponse, QueryError>`, where `OKResponse` includes `version`, `session_id` etc. Since `QueryError` seems like a for human error the `version` can be embedded in `QueryError.mesage` or `QueryError.detail` as string.

youngsofun (commented):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 11 unresolved discussions (waiting on @drmingdrmer, @flaneur2020, @hantmac, and @sundy-li)

___
*[`src/query/service/src/servers/http/v1/session/token_manager.rs` line 48 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3ovTKl32YT2aHoDn1W:-O3qEftm617-Bcd6xBPr:b-896fix) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/token_manager.rs#L48)):*

Always use `std::time::Duration` instead if possible.
</blockquote></details>

Done.
___
*[`src/query/service/src/servers/http/v1/session/token_manager.rs` line 107 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3oxDdUBC7afsv1wzIZ:-O3qFZXr3zkmn-mvyXsu:b-a9q4c4) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/token_manager.rs#L107)):*

Make `unix_ts()` return a `Duration`, convert to integer only when `Duration` can not be used.
</blockquote></details>

working
___
*[`src/query/users/src/user_token.rs` line 52 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3owq1mDDUjAvCk-fQw:-O3qFnYPDKLuLq7jjZzf:b-a9q4c4) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/users/src/user_token.rs#L52)):*

These methods are not very necessary. They can be replaced with indirect calls such as: `UserApiProvider::token_api().upsert_token()`.
</blockquote></details>

working
___
*[`src/query/management/src/token/token_mgr.rs` line 64 at r1](https://reviewable.io/reviews/datafuselabs/databend/16220#-O3mtUJ88EHx9_WyUaAs:-O3qECEN5onWCZUqKtJR:b-36k2fc) ([raw file](https://github.com/datafuselabs/databend/blob/b506740e91cea6e0f3f4dc10bb07ce06ce52be5f/src/query/management/src/token/token_mgr.rs#L64)):*

`is_update` here is not very clear about its purpose:
`replace_existing` would be better reflecting its behavior.

Avoid using plain number duration. Use `Duration` type for `ttl`.
</blockquote></details>

renamed to `update_only`

src/query/service/src/servers/http/v1/session/token_manager.rs
```diff
@@ -0,0 +1,218 @@
...
+            };
+        }
+
+        let hash = hash_token(token.as_bytes());
+        let cache = match token_type {
+            TokenType::Refresh => &self.refresh_tokens,
+            TokenType::Session => &self.session_tokens,
+        };
+
+        if !cache.read().contains(&hash) {
```
b41sh, line 197: Can we return token hash or `session_id` to client and store token info at meta ? Because the information of the token is in plain text, user can construct it by themselves. It may be the same as the token generated by other users, which has security risks.

src/query/service/src/servers/http/http_services.rs
```diff
@@ -97,7 +98,8 @@ impl HttpHandler {
     async fn build_router(&self, sock: SocketAddr) -> impl Endpoint {
         let ep_v1 = Route::new()
             .nest("/query", query_route())
-            .at("/login", post(login_handler))
+            .at("/session/login", post(login_handler))
+            .at("/session/renew", post(renew_handler))
```
b41sh, line 102: We can consider add a `logout` handler, so that client can delete current token.

src/query/service/src/servers/http/v1/session/login_handler.rs
youngsofun: done

src/query/service/src/servers/http/v1/session/renew_handler.rs
youngsofun: done

drmingdrmer (commented):
> Reviewed 6 of 7 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 9 unresolved discussions (waiting on @b41sh, @flaneur2020, @hantmac, @sundy-li, and @youngsofun)

___
*[`src/query/service/src/servers/http/v1/session/renew_handler.rs` line 33 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#gh-1710512261:-O3quJq24TyyyHpJgfCd:bx5awcp) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/renew_handler.rs#L33)):*
<details><summary><i>Previously, youngsofun (Yang Xiufeng) wrote…</i></summary><blockquote>

done
</blockquote></details>

Why not just use `Result`? :thinking: 
___
*[`src/query/service/src/servers/http/v1/session/login_handler.rs` line 39 at r4](https://reviewable.io/reviews/datafuselabs/databend/16220#gh-1710512083:-O3qu0J-3eEMiTAxmoYm:b-cxbhqe) ([raw file](https://github.com/datafuselabs/databend/blob/810ca0f1ecd5b9e1006d155416d5cab148de6305/src/query/service/src/servers/http/v1/session/login_handler.rs#L39)):*
<details><summary><i>Previously, youngsofun (Yang Xiufeng) wrote…</i></summary><blockquote>

done
</blockquote></details>

Why not make `LoginResponse` a `Result`? It will be able to reuse all Result's methods.

src/query/service/src/servers/http/v1/session/token_manager.rs
youngsofun: session_id is fixed during session.  so can not `expired` as a token.

youngsofun:
@drmingdrmer 

not need methods of Result
but need serde（untaged）

drmingdrmer (approved):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 8 unresolved discussions (waiting on @b41sh, @flaneur2020, @hantmac, @sundy-li, and @youngsofun)

sundy-li (approved):

b41sh (approved):

hantmac (approved):

youngsofun:
working on:

todo（small enhancement）:

- [ ]  make nonce safer
- [ ]  check session_id or nonce in addition to hash exists


coming soon

youngsofun:
@flaneur2020  do you need to review before merge?

drmingdrmer (approved):
> Reviewed 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 9 unresolved discussions (waiting on @b41sh, @flaneur2020, @sundy-li, and @youngsofun)

___
*[`src/query/management/src/token.rs` line 64 at r7](https://reviewable.io/reviews/datafuselabs/databend/16220#-O44dNGOE8VdFTBl081u:-O44dNGOE8VdFTBl081v:b-qtgfvh) ([raw file](https://github.com/datafuselabs/databend/blob/5af7e75dd0275b8a531ee96a8b2adaeadc3f0474/src/query/management/src/token.rs#L64)):*
> ```XML
>         let upsert = UpsertPB::update(ident, token_info)
>             .with(seq)
>             .with_ttl(Duration::from_secs(ttl.as_secs()));
> ```

ttl is already `Duration`, just use it directly



_Suggestion:_
```XML
.with_ttl(ttl);
```

src/query/service/src/servers/http/middleware.rs
```diff
@@ -176,10 +179,29 @@ fn auth_by_header(
...
     } else if value.as_bytes().starts_with(b"Bearer ") {
         match Bearer::decode(value) {
-            Some(bearer) => Ok(Credential::Jwt {
-                token: bearer.token().to_string(),
-                client_ip,
-            }),
+            Some(bearer) => {
+                let token = bearer.token().to_string();
+                if SessionClaim::is_databend_token(&token) {
+                    let (token_type, set_user) = if path == "/query" {
```
flaneur2020, line 185: imho it's not a common practice to distinguish tokens by url paths.

when we need refreshing tokens, we can simply pass the refresh token in the request body, and get the new tokens in the response.

src/query/service/src/servers/http/v1/session/token_manager.rs
```diff
@@ -0,0 +1,214 @@
...
+        Ok(())
+    }
+
+    /// used for both issue token for new session and renew token for existing session.
+    /// currently, when renewing, always return a new refresh token instead of update the TTL,
+    /// since now we include expire time in token, which can not be updated.
+    pub async fn new_token_pair(
+        &self,
+        session: &Arc<Session>,
+        old_token_pair: Option<TokenPair>,
```
flaneur2020, line 84: i suppose that it only needs the refresh token to get the new token pair

src/query/service/src/servers/http/middleware.rs
flaneur2020: maybe another way is to encode the different types as prefix in the generated tokens: 

```
SessionClaim::is_databend_session_token(&token) 
SessionClaim::is_databend_refresh_token(&token) 
```

src/query/service/src/servers/http/v1/query/http_query_context.rs
```diff
@@ -40,6 +40,7 @@ pub struct HttpQueryContext {
     pub http_method: String,
     pub uri: String,
     pub client_host: Option<String>,
+    pub databend_token: Option<String>,
```
flaneur2020, line 43: `databend_token` -> `session_token` ?

src/query/service/src/auth.rs
```diff
@@ -27,13 +28,19 @@ use databend_common_users::JwtAuthenticator;
...
 
+use crate::servers::http::v1::TokenManager;
 use crate::sessions::Session;
 
 pub struct AuthMgr {
     jwt_auth: Option<JwtAuthenticator>,
 }
 
 pub enum Credential {
+    Databend {
```
flaneur2020, line 39: Databend -> SessionToken ?

src/query/service/src/servers/http/v1/query/http_query_manager.rs
```diff
@@ -230,19 +226,16 @@ impl HttpQueryManager {
...
+        self: &Arc<Self>,
+        last_query_id: &str,
+    ) -> Option<TxnManagerRef> {
+        let mut txn_managers = self.txn_managers.lock();
+        if let Some((txn_mgr, task_handle)) = txn_managers.remove(last_query_id) {
+            task_handle.abort();
+            Some(txn_mgr)
+        } else {
+            None
+        }
```
flaneur2020, line 239: why does getting `token manager` will mutate `txn_managers`?

src/query/service/src/servers/http/v1/query/http_query_context.rs
flaneur2020: `Option<String>` -> `Option<DatabendToken>` ?

src/query/service/src/servers/http/middleware.rs
youngsofun: @flaneur2020 you are right, I was planning to do this refactor later, let me do it now.

src/query/service/src/auth.rs
youngsofun: `Databend` auth/token refer a implementition/mechanism (bend-v1-xxx) (compare to JWT), whose **usage** can be a session token or refresh token for now.


or what about rename as DatabendSession， SessionRefeshToken and SessionAcessToken？

src/query/service/src/servers/http/v1/session/token_manager.rs
youngsofun: need to delete the old session token?

src/query/service/src/auth.rs
flaneur2020: the credential is represented by a Token here, so i think it's not a bad idea to name it with a "Token" suffix 👀

flaneur2020: in my opinion, we do not need putting the REFRESH TOKEN as a kind of credential. refresh token is only used in one renewing token api, it can be considered as a request parameter, instead of the identity credential (which is mostly the session token plays the role).

src/query/service/src/servers/http/v1/query/http_query_manager.rs
youngsofun: thanks, forgot to remove this code.

drmingdrmer (approved):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 15 unresolved discussions (waiting on @b41sh, @flaneur2020, @sundy-li, and @youngsofun)

src/query/service/src/servers/http/v1/query/http_query_context.rs
youngsofun: for now, only used in `renew_handler` and only need the token itself. we can refactor later if we hav other usage

src/query/service/src/servers/http/middleware.rs
youngsofun: done

src/query/service/src/auth.rs
youngsofun: > the credential is represented by a Token here, so i think it's not a bad idea to name it with a "Token" suffix 👀

done

drmingdrmer (commented):
> Reviewed 5 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 15 unresolved discussions (waiting on @b41sh, @flaneur2020, @sundy-li, and @youngsofun)

drmingdrmer (commented):
> Reviewed 5 of 5 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 15 unresolved discussions (waiting on @b41sh, @flaneur2020, @sundy-li, and @youngsofun)

src/query/service/src/auth.rs
youngsofun: @flaneur2020 ci passed

drmingdrmer (approved):
> Reviewed 1 of 1 files at r11, 7 of 7 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16220)* status: all files reviewed, 15 unresolved discussions (waiting on @b41sh, @flaneur2020, @sundy-li, and @youngsofun)

youngsofun:
@flaneur2020 @BohuTANG what about merge this first, so we can begin the develop of clients, we still have chance to modify details before all is ready

PR #16231, opened 2024-08-12
change: Meta-service cluster is gated by Enterprise Edition

drmingdrmer:

##### change: Meta-service cluster is gated by Enterprise Edition

By default meta-service disallows clustering.
Meta-service cluster is only enabled when raft-config
`databend_enterprise_license` is set and is valid.
No feature in the jwt claim is examined.

The EE gate check when a meta node initiate raft-protocol network
instance. Thus without a valid EE token, all raft-protocol are disabled,
including `RequestVote`, `AppendEntries`, `InstallSnapshot` and internal
request forward.

If EE token is not set, an error will be outputed to log file.

- New config `databend_enterprise_license`:
  ```
  [raft_config]
  databend_enterprise_license = "<token>"
  ```

  This token is same as the one used by databend-query.

- When testing, a temp key pair and jwt claim is created to pass
  integration tests, this is enabled by `fake_ee_license` config entry.

- Other changes: Add `DisplaySlice` and `DisplayUnixTimeStampExt` to
  display slice of `Display` instance and unix timestamp.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Other

## Related Issues


everpcpc (approved):

src/meta/ee/src/lib.rs
```diff
@@ -11,3 +11,111 @@
...
+pub struct MetaServiceEnterpriseGate {
+    /// License ES256 signed jwt claim token.
+    license_token: Arc<Mutex<Option<String>>>,
+
+    /// The public key to verify the license token.
+    public_key: String,
+}
+
+impl MetaServiceEnterpriseGate {
+    const LICENSE_PUBLIC_KEY: &'static str = r#"-----BEGIN PUBLIC KEY-----
```
lichuang, line 39: why use a hard code public key?

drmingdrmer: This way only the the license token published by us with our private key will be correctly decoded.
If to let user config public key in config, user will be able to use a self signed token with its private key and let databend-meta decoded it with the public key it provided.

lichuang (approved):

drmingdrmer:
The API to update license in flight is added, Please review again guys

ZhiHanZ (approved):
LGTM, some ideas:
should we update meta server license when query cluster trigger
`SET GLOBAL LICENSE` command? it could simplify the meta service license setup, and support to view license from meta service by using license info system call command.
Ref:
https://docs.databend.com/guides/overview/editions/dee/license#verifying-a-license

drmingdrmer:
> LGTM, some ideas: should we update meta server license when query cluster trigger `SET GLOBAL LICENSE` command? it could simplify the meta service license setup, and support to view license from meta service by using license info system call command. Ref: https://docs.databend.com/guides/overview/editions/dee/license#verifying-a-license

If databend-query accepted an invalid license and forwarded it to meta-services, then meta-service cluster may run into problem because it can not form a cluster anymore. And then any further setting can not be saved by databend-query.

Therefore it is a little bit dangerous, but it would help in most cases.

PR #16235, opened 2024-08-13
refactor(query): don't retry transform on each block

sundy-li:

## Summary

don't retry transform on each block, keep the counter in each thread.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16235)


src/query/pipeline/transforms/src/processors/transforms/transform_retry_async.rs
```diff
@@ -30,12 +30,16 @@ pub struct RetryStrategy {
 }
 
 pub struct AsyncRetryWrapper<T: AsyncRetry + 'static> {
+    retries: usize,
```
b41sh, line 33: better to use `Arc<AtomicUsize>`

sundy-li: We want to let it be counted in each thread. So this variable is not shared between processors.

zhang2014 (approved):

b41sh (approved):

PR #16242, opened 2024-08-14
fix(query): cume_dist() always return unbounded frame

TCeason:

## Summary

cume_dist: The cumulative distribution: (number of partition rows preceding or peer with current row) / total partition rows.

So it's frame caluse should be unbounded.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16242)

### DIFF
--- a/tests/sqllogictests/suites/query/window_function/window_wisconsin.test
+++ b/tests/sqllogictests/suites/query/window_function/window_wisconsin.test
@@ -145,7 +145,7 @@ SELECT ntile(5) OVER (ORDER BY ten, four) nn FROM tenk1 ORDER BY ten, four, nn
 
 # cume_dist
 query I
-SELECT CAST(cume_dist() OVER (PARTITION BY four ORDER BY ten RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)*10 as INT) FROM tenk1 WHERE unique2 < 10 order by four, ten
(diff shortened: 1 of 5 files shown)


tests/sqllogictests/suites/query/window_function/window_wisconsin.test
```diff
@@ -145,7 +145,7 @@ SELECT ntile(5) OVER (ORDER BY ten, four) nn FROM tenk1 ORDER BY ten, four, nn
 
 # cume_dist
 query I
-SELECT CAST(cume_dist() OVER (PARTITION BY four ORDER BY ten RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)*10 as INT) FROM tenk1 WHERE unique2 < 10 order by four, ten
```
sundy-li, line 148: if the frame is not unbounded, we should raise error or ?

TCeason: Will ignore frame bound in sql. 

Same behavior with lead lag.

sundy-li (approved):

PR #16243, opened 2024-08-14
fix(query): fix alter user set `MUST_CHANGE_PASSWORD`

b41sh:

## Summary

If the administrator sets `MUST_CHANGE_PASSWORD` as true, the option takes effect immediately. The User must change the password at the next login.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16243)

### DIFF
--- a/src/meta/app/src/principal/user_auth.rs
+++ b/src/meta/app/src/principal/user_auth.rs
@@ -158,7 +158,8 @@ impl AuthInfo {
         AuthInfo::new(auth_type, auth_string, need_change)
     }
 
-    pub fn create_set_need_change(&self, need_change: bool) -> AuthInfo {
+    // create `AuthInfo` and only modify `need_change` field.
+    pub fn create_with_need_change(&self, need_change: bool) -> AuthInfo {
         match self {
             AuthInfo::Password {
                 hash_value,
--- a/src/meta/app/src/principal/user_auth.rs
+++ b/src/meta/app/src/principal/user_auth.rs
@@ -158,6 +158,21 @@ impl AuthInfo {
         AuthInfo::new(auth_type, auth_string, need_change)
     }
 
+    pub fn create_set_need_change(&self, need_change: bool) -> AuthInfo {
+        match self {
+            AuthInfo::Password {
+                hash_value,
+                hash_method,
+                ..
+            } => AuthInfo::Password {
+                hash_value: hash_value.clone(),
+                hash_method: *hash_method,
+                need_change,
+            },
+            _ => self.clone(),
+        }
+    }
+
     pub fn alter(
         &self,
         auth_type: &Option<String>,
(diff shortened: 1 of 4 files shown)


src/query/sql/src/planner/binder/ddl/account.rs
```diff
@@ -302,23 +302,22 @@ impl Binder {
...
-                false
-            } else {
-                user_option
-                    .must_change_password()
-                    .cloned()
-                    .unwrap_or_default()
-            };
+            // if user is changing self password, always set `need_change` as false.
+            // because after this operation, the password is changed.
+            let need_change = if user.is_none() { false } else { need_change };
```
TCeason, line 320: // None means current user
        let user_info = if user.is_none() {
            self.ctx.get_current_user()?
        }

TCeason: alter user USER() identified by 'password'

The user in plan is None.

TCeason: This logic can merge with 

`let need_change = user_option
            .must_change_password()
            .cloned()
            .unwrap_or_default();` ?

src/meta/app/src/principal/user_auth.rs
```diff
@@ -158,6 +158,21 @@ impl AuthInfo {
         AuthInfo::new(auth_type, auth_string, need_change)
     }
 
+    pub fn create_set_need_change(&self, need_change: bool) -> AuthInfo {
```
TCeason, line 161: why not use create2?

TCeason: It like a alter impl not create.

b41sh: we don't change `auth_type` and `auth_string`, only set `need_change`, the logic is different with `create2`.

src/query/sql/src/planner/binder/ddl/account.rs
b41sh: If user not change password, we set `need_change` same as `must_change_password`. And if user change other user's password, `need_change` is also same as `must_change_password`, if user change self password, we set `need_change` as false, because the password has changed, user don't need change again.

TCeason: Ok I think I got you. Maybe can add it into comment?

b41sh: ok

sundy-li (approved):

Issue #16213, opened 2024-08-08
bug: Insert and select Double INF value cause error


hantmac:
### Search before asking

- [X] I had searched in the [issues](https://github.com/datafuselabs/databend/issues) and found no similar issues.

### Version

Databend Query v1.2.598-nightly-812688f7b6(rust-1.81.0-nightly-2024-08-05T22:07:38.113740317Z)

### What's Wrong?

Using java insert into the `Double.POSITIVE_INFINITY` value into table and get an error sometime when select from it.

### How to Reproduce?

```java
 public void testWriteDouble() throws SQLException {
        try (Connection connection = createConnection()) {
            DatabendStatement statement = (DatabendStatement) connection.createStatement();
            statement.execute("CREATE TABLE IF NOT EXISTS test_basic_driver.table_double (\n" +
                    "    ID INT,\n" +
                    "    Name VARCHAR(50),\n" +
                    "    Age INT,\n" +
                    "    City VARCHAR(50),\n" +
                    "    Score DOUBLE\n" +
                    ");");
            Double infDouble = Double.POSITIVE_INFINITY;

            String sql = "INSERT INTO test_basic_driver.table_double (ID, Name, Age, City, Score) values";
            PreparedStatement prepareStatement = connection.prepareStatement(sql);
            prepareStatement.setInt(1, 1);
            prepareStatement.setString(2, "Alice");
            prepareStatement.setInt(3, 25);
            prepareStatement.setString(4, "Toronto");
            prepareStatement.setDouble(5, infDouble);

            prepareStatement.addBatch();
            prepareStatement.executeBatch();
            statement.execute("SELECT * FROM test_basic_driver.table_double");
            ResultSet r = statement.getResultSet();
            r.next();
            Assert.assertEquals(r.getDouble(5), "Infinity");
        } catch (SQLException throwables) {
            throwables.printStackTrace();
        }
    }
```

The above code get error:

```
java.sql.SQLException: Error executing query: SQL: SELECT * FROM test_basic_driver.table_double Query failed: QueryErrors{code=1006, message=Cannot parse request for /actions/init_query_fragments, cause: Error("invalid type: null, expected f64", line: 1, column: 3444)} cause: null
	at com.databend.jdbc.DatabendStatement.internalExecute(DatabendStatement.java:215)
	at com.databend.jdbc.DatabendStatement.execute(DatabendStatement.java:154)
[...]

PR #16258, opened 2024-08-16
fix: add custom (de)serialization methods for special float value

dqhl76:

## Summary
fixes: #16213
- Implemented custom serialization and deserialization for F32 and F64 types in NumberScalar to handle infinite/Nan float values.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16258)


andylokandy:
Thank you!

tests/sqllogictests/suites/base/issues/issue_16213.test
```diff
@@ -0,0 +1,11 @@
+statement ok
+drop table if exists test_table_double
+
+statement ok
+CREATE TABLE IF NOT EXISTS test_table_double (ID INT, Name VARCHAR(50), Age INT, City VARCHAR(50), Score DOUBLE)
+
+statement ok
+INSERT INTO test_table_double (ID, Name, Age, City, Score) VALUES (1, 'Alice', 25, 'Toroto', 'inf')
```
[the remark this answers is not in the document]
sundy-li:
```
🐳 :) insert into a select 'inf' from numbers(100);
100 rows written in 0.200 sec. Processed 100 rows, 8B (500.73 rows/s, 40B/s)

🐳 :) insert into a select 'inf' from numbers(100);
100 rows written in 0.122 sec. Processed 100 rows, 8B (820.21 rows/s, 65B/s)

🐳 :) insert into a select 'nan' from numbers(100);
100 rows written in 0.107 sec. Processed 100 rows, 8B (933.3 rows/s, 74B/s)

🐳 :) select  max(a) from a;
error: APIError: ResponseError with 4000: invalid data (For portability reasons we do not allow to serialize NaNs.)

```

andylokandy:
@sundy-li Is supporting infinite for all NumberScalar a goo idea? As there is no standard for inf serialization, I'd concern about the forward compatibility because NumberScalar could be persist in meta settings. 

On the other aspect, this is an user interface change. We used to / planned to support `'inf'::float64` as the recommended way to construct a inf float. If we accept this PR, it'll implying that we will support auto cast from  string to numeric types, which is denied currently.

dqhl76:
> I'd concern about the forward compatibility because NumberScalar could be persist in meta or settings.

I add a condition `serializer.is_human_readable()` to handle the `inf` case. I find seems we use `bincode` and `msgpack` in meta related(they are not human readable but json is). It could be not affect?

sundy-li:
> I find seems we use bincode and msgpack in meta related(they are not human readable but json is). It could be not affect? Not sure about that 0.0

Yes.

> As there is no standard for inf serialization, I'd concern about the forward compatibility because NumberScalar could be persist in meta or settings.

@andylokandy  If `inf`or `nan` is already serialized into meta or settings via `serde_json`, we can't deserialize it back, it could throw errors, so this pr can solve the problem.


Now we can also cast string into float64 (but it's not auto-cast)
```
🐳 :) select '+inf'::float64, '-inf'::float64, 'nan'::float64, 'Infinity'::float64, '-Infinity'::float64, 'Nan'::float64;
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ '+inf'::float64 │ '-inf'::float64 │ 'nan'::float64 │ 'infinity'::float64 │ '-infinity'::float64 │ 'nan'::float64 │
│     Float64     │     Float64     │     Float64    │       Float64       │        Float64       │     Float64    │
├─────────────────┼─────────────────┼────────────────┼─────────────────────┼──────────────────────┼────────────────┤
│             inf │            -inf │            NaN │                 inf │                 -inf │            NaN │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
1 row read in 0.048 sec. Processed 1 row, 1B (21.05 rows/s, 21B/s)
```

sundy-li (approved):

tests/sqllogictests/suites/base/issues/issue_16213.test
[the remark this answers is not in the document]
dqhl76: When I try to add the tests, I found that `error: APIError: ResponseError with 4000: invalid data (For portability reasons we do not allow to serialize NaNs.)` still occurs.
This error is issuesed by `borsh`, we use it at `src/query/functions/src/aggregates/adaptors/aggregate_null_unary_adaptor.rs:201:9`. It is not controlled by our overried (de)serde methods.

tests/sqllogictests/suites/base/issues/issue_16213.test
```diff
@@ -0,0 +1,11 @@
...
+
+statement ok
```
andylokandy, line 10: May be better change it it `query`

dqhl76: Got that

tests/sqllogictests/suites/base/issues/issue_16213.test
```diff
@@ -7,5 +7,22 @@ CREATE TABLE IF NOT EXISTS test_table_double (ID INT, Name VARCHAR(50), Age INT,
 statement ok
 INSERT INTO test_table_double (ID, Name, Age, City, Score) VALUES (1, 'Alice', 25, 'Toroto', 'inf')
 
-statement ok
+query
 SELECT * FROM test_table_double
+1 Alice 25 Toroto Infinity
```
andylokandy, line 12: Why this is not `inf`?

dqhl76: Sorry for misunderstanding, fixed

dqhl76:
Sorry for some time leave.

This PR was put on hold because overriding the (de)serialization methods for F32 and F64 still doesn't address the issue. There are parts of the code, such as ValueType::Scalar, that use borsh-serde but are unaffected by this PR's override methods.
img_v3_02dt_20858391-51ef-4763-9a71-7b2b6fbbf30g

To work around this problem, I patched borsh-rs to remove the NaN restriction: [borsh-rs NaN bypass patch](https://github.com/dqhl76/borsh-rs/tree/bypass-nan?rev=6609a44). It's unlikely that the upstream will accept these changes due to potential cross-platform issues with NaN, but we don't encounter this problem in our case.

img_v3_02dt_1164255e-c0e3-40cd-870b-a94abb0f65cg

sundy-li:
> It's unlikely that the upstream will accept these changes due to potential cross-platform issues with NaN, but we don't encounter this problem in our case.


You can make it a feature gate that defaults to be disabled.  Send a pr to upstream, let the maintainer decide whether to accept it or not.

dqhl76:
> > It's unlikely that the upstream will accept these changes due to potential cross-platform issues with NaN, but we don't encounter this problem in our case.
> 
> You can make it a feature gate that defaults to be disabled. Send a pr to upstream, let the maintainer decide whether to accept it.

Ok, I will send a PR to upstream later

dqhl76:
> You can make it a feature gate that defaults to be disabled. Send a pr to upstream, let the maintainer decide whether to accept it.

I opened a PR yesterday. Maintainers have some concerns about that. https://github.com/near/borsh-rs/pull/308

sundy-li:
LGTM, maybe we can do better.
```
#[derive(Default, Clone, Copy)]
#[repr(transparent)]
struct F32(OrderedFloat<f32>);
struct F64(OrderedFloat<f64>);
```

then let's implement serde， ord， eq  for the new struct

And `AsRef` ， `from(f32)`, ...    for it.

dqhl76:
Hi @sundy-li, Could you please take a look again?

andylokandy:
I'm concerned about moving the whole lib into databend. As @sundy-li said, we can impl the custom serde logic for `F32` and `F64`.

Even if we decide to vendor the crate, it's better to give it its own crate rather than moving into `databend-common-base`.

dqhl76:
Thanks for your review.

> I'm concerned about moving the whole lib into databend. As @sundy-li said, we can impl the custom serde logic for `F32` and `F64`.

If we implement these traits (like serde, Ord, Eq, AsRef, From(f32)) ourselves, we’re essentially just duplicating what the ordered-float library already does. In that case, we migh vendor the lib directly.

> Even if we decide to vendor the crate, it's better to give it its own crate rather than moving into `databend-common-base`.

It sounds good to me. I didn't put it into a crate becuase I thought it is a `one file crate`.

sundy-li (approved):

PR #16266, opened 2024-08-16
feat: new table option `data_retention_period_in_hours`

dantengsky:

## Summary

Introduces new table option `data_retention_period_in_hours`, which has a higher priority to session/global level setting of `data_retention_period_in_days`, if it is set.

The min value of table `data_retention_period_in_hours` is 1 (hour), the max values of it is`config.data_retention_time_in_days_max` is specified ( or 90 days if not specified in config)

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16266)


SkyFan2002 (approved):

sundy-li:
> which has a higher priority to session/global level setting of data_retention_period_in_days

Let's just keep config & table option.

dantengsky:
> Let's just keep config & table option. I think maybe we don't need the setting.

do you mean we no longer need session level retention period settings?

sundy-li:
Yes， I think keep three with priority confuse lot.

wubx:
@soyeric128  This feature need doc.

PR #16282, opened 2024-08-20
chore(query): remove unused parquet2 codes

sundy-li:

## Summary

refactor and remove unused parquet2 codes

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16282)


src/common/storage/src/parquet_rs.rs
```diff
@@ -131,6 +140,50 @@ pub async fn read_metadata_async(
     }
 }
 
+pub fn read_metadata_sync(
```
Xuanwo, line 143: We can consider using https://crates.io/crates/parquet_opendal

youngsofun (approved):

src/common/storage/src/parquet_rs.rs
sundy-li: Yes, but this is for sync api yet parquet_opendal only support async api.

Xuanwo: > Yes, but this is for sync api yet parquet_opendal only support async api.

Would you like to create an issue for this?

PR #16328, opened 2024-08-26
fix: When Replacing column mask, old `name->id->value` should be completely deleted

drmingdrmer:

##### fix: When Replacing column mask, old `name->id->value` should be completely deleted

Before this commit, only the `name->id` is removed when replacing column
mask, this would cause column mask records leaked as a junk in
meta-service.

Other changes: simplify data mask and background job API implementation
with KVPbApi routines.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16328)

### DIFF
--- a/src/meta/api/src/background_api_impl.rs
+++ b/src/meta/api/src/background_api_impl.rs
@@ -12,12 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fmt::Display;
-
 use chrono::Utc;
 use databend_common_meta_app::app_error::AppError;
-use databend_common_meta_app::app_error::BackgroundJobAlreadyExists;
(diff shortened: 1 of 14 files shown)


ZhiHanZ (approved):

BohuTANG:
```
Test finished, fail fast enabled, 1 out of 1214 records failed to run
0: statement failed: mysql client error: Server error: `ERROR HY000 (1105): UnknownDatamask. Code: 1113, Text = Unknown Resource> 'mask'(seq >= 0): drop_data_mask.'
[SQL] drop MASKING POLICY if exists mask
at tests/sqllogictests/suites/ee/05_ee_ddl/05_0001_ddl_data_mask.test:15
```

PR #16333, opened 2024-08-26
feat(query): add plan cache

### LINKED ISSUE #16294: bug: COPY INTO <location> does not support tables that need quoting
### Search before asking

### Version

v1.2.618-nightly

### What's Wrong?

Trying to COPY INTO gcs from table with dashes in name gives error. Works fine for table with no dashes

### How to Reproduce?

```sql
CREATE table t1 (c1 int null);
CREATE table "t-2" (c1 int null);

select * from default.t1;

COPY INTO 'gcs://projects/tables/t1'
CONNECTION = (
	CREDENTIAL = '<snip>'
)
FROM default.t1
FILE_FORMAT = (TYPE = PARQUET);

select * from default."t-2";

COPY INTO 'gcs://projects/tables/t-2'
CONNECTION = (
	CREDENTIAL = '<snip>'
)
FROM default."t-2"
FILE_FORMAT = (TYPE = PARQUET);
```
```
SQL:1:32
  |
1 | SELECT * FROM default.default.t-2
  |                                ^ unexpected `-`, expecting `AT`, `CHANGES`, `WITH`, <Ident>, <LiteralString>, `IDENTIFIER`, `AS`, `PIVOT`, `UNPIVOT`, `SAMPLE`, `ROW`, `BLOCK`, `(`, `LiteralAtString`, `LATERAL`, `NATURAL`, `INNER`, `LEFT`, `SEMI`, `RIGHT`, `ANTI`, `FULL`, `CROSS`, `JOIN`, `ON`, `USING`, `,`, `WHERE`, `GROUP`, `HAVING`, `WINDOW`, `QUALIFY`, `UNION`, `EXCEPT`, `INTERSECT`, `FROM`, `SELECT`, `VALUES`, `ORDER`, `LIMIT`, `OFFSET`, `IGNORE_RESULT`, `FORMAT`, or `;`

} cause: null
```

sundy-li:

## Summary

Add plan cache, queries will reuse the plan from cache if source plan is not changed (snapshot is not changed).

This cache is similar to Oracle's Global Plan Cache, but it's stored as LRU cache in memory.

Also, fix the name resolution issue of #16294 
fixes #16294 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16333)


src/query/sql/src/planner/planner_cache.rs
```diff
@@ -0,0 +1,155 @@
...
+use sha2::Digest;
+use sha2::Sha256;
+
+use crate::normalize_identifier;
+use crate::plans::Plan;
+use crate::NameResolutionContext;
+use crate::PlanExtras;
+use crate::Planner;
+
+type PlanCache = LruCache<String, (Plan, PlanExtras)>;
```
zhang2014, line 38:
```suggestion
type PlanCache = InMemoryLruCache<String, (Plan, PlanExtras)>;
```

xudong963 (commented):
Do we need unit tests to verify 
1. get plan from cache
2.  if the table structure changed, can't get plan from cache?

sundy-li:
> Do we need unit tests to verify
get plan from cache
if the table structure changed, can't get plan from cache?

The setting is default enabled, so logictest covered both issues.

BohuTANG:
> > Do we need unit tests to verify
> > get plan from cache
> > if the table structure changed, can't get plan from cache?
> 
> The setting is default enabled, so logictest covered both issues.

It would be better to add some new logic tests for this PR to ensure it remains unbreakable in the future, as we’re not sure if the existing tests cover this PR or if those tests might change later on.

xudong963 (approved):

Issue #15741, opened 2024-06-05
RFC: Enhancing Error Handling in Rust with `error-stack`


andylokandy:
# RFC: Enhancing Error Handling in Rust with `error-stack`

## Motivation

Inspired by [a post from GreptimeDB](https://greptime.com/blogs/2024-05-07-error-rust) [[中文]](https://zhuanlan.zhihu.com/p/696420829), we propose a new approach to error handling in Rust to improve user experience. The current popular error handling practices have several drawbacks:

1. Errors are typically represented as an enum tree and implemented with the `From` trait. This makes it difficult to track different errors originating from the same source. For instance, distinguishing between `ReadIndexError` and `WriteIndexError` from a single `io::Error` is not straightforward.
2. Errors lack trace information. While `Backtrace` can trace errors, it cannot do so across different threads.
3. Error types are usually added at the crate level, which is not flexible enough. Errors can be smaller than a crate or span multiple crates. For example, Databend only has one layer of error type, which complicates high-level problem reasoning.

## Goals

We aim to replace the current error handling practice with [`error-stack`](https://crates.io/crates/error-stack). This crate tracks errors in detail, from the higher levels down to the root cause. An example of an error message generated by `error-stack` reveals the root cause and the error's progression, including source code locations for each layer:

```rust
Error: a fatal error has occurred in the main loop
├╴at src/main.rs:11:51
│
├─▶ failed to read index file: index.txt
│   ╰╴at src/main.rs:29:35
│
╰─▶ No such file or directory (os error 2)
    ╰╴at src/main.rs:29:3
```

Due to its mechanism, `error-stack` can trace errors across threads or async contexts, eliminating the need for `async-backtrace`.

## Best Practices

`error-stack` introduces a new pattern for error handling. It might seem unfamiliar initially, but it offers significant advantages.

### 1. Use Structs for Errors
Instead of big-enum error types, define errors using structs:

```rust
#[derive(Debug, thiserror::Error)]
#[error("{0}")]
pub struct ExecutorError(pub(crate) String);
```

- The error does not contain any information about the source error.
- The error message is a human-readable string, which is often more informative than an enum variant.

### 2. Use `error_stack::Result`
Replace `type Result<T> = std::result::Result<T, Error>` with `error_stack::Result`:

```rust
use error_stack::Result;

pub fn read_index(path: &str) -> Result<String, ExecutorError> {
    todo!()
}
```

Explicitly specify the error type to accommodate the next change.

[...]

PR #16343, opened 2024-08-29
feat: add error stack to ErrorCode

andylokandy:

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16343)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -350,6 +350,7 @@ async_fn_in_trait = "allow"
 [workspace.lints.clippy]
 useless_format = "allow"
 mutable_key_type = "allow"
+result_large_err = "allow"
 
 [profile.release]
 debug = 1
--- a/src/common/base/src/base/string.rs
+++ b/src/common/base/src/base/string.rs
@@ -30,7 +30,7 @@ use unicode_segmentation::UnicodeSegmentation;
 /// let new_key = escape_for_key(&key);
 /// assert_eq!(Ok("data_bend%21%21".to_string()), new_key);
 /// ```
-pub fn escape_for_key(key: &str) -> Result<String, FromUtf8Error> {
+pub fn escape_for_key(key: &str) -> std::result::Result<String, FromUtf8Error> {
     let mut new_key = Vec::with_capacity(key.len());
 
     fn hex(num: u8) -> u8 {
@@ -64,7 +64,7 @@ pub fn escape_for_key(key: &str) -> Result<String, FromUtf8Error> {
 /// let original_key = unescape_for_key(&key);
 /// assert_eq!(Ok("data_bend!!".to_string()), original_key);
 /// ```
-pub fn unescape_for_key(key: &str) -> Result<String, FromUtf8Error> {
+pub fn unescape_for_key(key: &str) -> std::result::Result<String, FromUtf8Error> {
     let mut new_key = Vec::with_capacity(key.len());
 
     fn unhex(num: u8) -> u8 {
--- a/src/common/base/src/runtime/memory/stat_buffer.rs
+++ b/src/common/base/src/runtime/memory/stat_buffer.rs
@@ -15,8 +15,6 @@
 use std::ptr::addr_of_mut;
 use std::sync::atomic::Ordering;
 
-use databend_common_exception::Result;
-
 use crate::runtime::memory::mem_stat::OutOfLimit;
 use crate::runtime::memory::MemStat;
 use crate::runtime::LimitMemGuard;
@@ -70,7 +68,10 @@ impl StatBuffer {
     }
 
     /// Flush buffered stat to MemStat it belongs to.
-    pub fn flush<const ROLLBACK: bool>(&mut self, alloc: i64) -> Result<(), OutOfLimit> {
+    pub fn flush<const ROLLBACK: bool>(
+        &mut self,
+        alloc: i64,
+    ) -> std::result::Result<(), OutOfLimit> {
         match std::mem::take(&mut self.memory_usage) {
             0 => Ok(()),
             usage => {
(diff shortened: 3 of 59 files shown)


BohuTANG:
How do I decide whether to use `std::result::Result` or `databend_common_exception::exception::Result`? 

This PR replaces more of `databend_common_exception::exception::Result` with `std::result::Result`, but it makes it harder to determine when to use each.

andylokandy:
> How do I decide whether to use std::result::Result or databend_common_exception::exception::Result?

It's easy. When you want `Result<T, ErrorCode>`, use `databend_common_exception::exception::Result`, otherwise, use `std::result::Result`.

BohuTANG:
> > How do I decide whether to use std::result::Result or databend_common_exception::exception::Result?
> 
> It's easy. When you want `Result<T, ErrorCode>`, use `databend_common_exception::exception::Result`, otherwise, use `std::result::Result`.

Use any of them is not affect the stack of ErrorCode?

andylokandy:
Only `Result<T, ErrorCode>` records the error stack. In practice, `std::result::Result` is used where 
1. interoperating with other low-level libraries.
2. intended to be recovered.
3. be the root cause of an `Result<T, ErrorCode>`

In either case, the error stack is not needed.

sundy-li:
Are there any new tests to show the new error stack format ?

drmingdrmer (commented):
> Reviewed 5 of 59 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16343)* status: 5 of 59 files reviewed, 1 unresolved discussion (waiting on @andylokandy, @sundy-li, and @zhang2014)

___
*[`src/common/exception/src/exception.rs` line 104 at r1](https://reviewable.io/reviews/datafuselabs/databend/16343#-O5W1zSi0qc-gXtSSG6y:-O5W1zSi0qc-gXtSSG6z:b-7l903l) ([raw file](https://github.com/datafuselabs/databend/blob/fc9be0af1a6088efede85a9ce5c19a01bac7d309/src/common/exception/src/exception.rs#L104)):*
> ```XML
>     pub(crate) stacks: Vec<ErrorFrame>,
>     pub(crate) _phantom: PhantomData<C>,
> }
> ```

Why does `ErrorCode` require a type parameter `C`? Where is `C` actually used?

If this type parameter serves any implicit purpose, please add a doc comment explaining it.



_Code quote:_
```XML
pub struct ErrorCode<C = ()> {
    pub(crate) code: u16,
    pub(crate) name: String,
    pub(crate) display_text: String,
    pub(crate) detail: String,
    pub(crate) span: Span,
    // cause is only used to contain an `anyhow::Error`.
    // TODO: remove `cause` when we completely get rid of `anyhow::Error`.
    pub(crate) cause: Option<Box<dyn std::error::Error + Sync + Send>>,
    pub(crate) backtrace: Option<ErrorCodeBacktrace>,
    pub(crate) stacks: Vec<ErrorFrame>,
    pub(crate) _phantom: PhantomData<C>,
}
```

zhang2014:
Why need to change `databend_common_exception::exception::Result` to `std::result::Result`.

Perhaps it is compatible?
```
pub type databend_common_exception::exception::Result<T, E = ErrorCode> = std::result::Result<T, E>;
```

drmingdrmer (commented):
> Reviewed 10 of 59 files at r1.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16343)* status: 15 of 59 files reviewed, 2 unresolved discussions (waiting on @andylokandy, @sundy-li, and @zhang2014)

*[a discussion](https://reviewable.io/reviews/datafuselabs/databend/16343#-O5W4hug8qGwHAq-uoFk:-O5W4hug8qGwHAq-uoFl:bjfjuv8) (no related file):*
I recommend using `Result` for `std::result::Result` throughout our codebase. For the specialized `Result<T, ErrorCode>`, consider using a different term, such as `my_lib::Result`.

___

andylokandy:
This is an code style change described in https://github.com/datafuselabs/databend/issues/15741.

Since when the issue is written, I've tried multiple ways to bring `error-stack` into databend, and this PR is the most viable one. Instead of replacing all `Result<T, ErrorCode>` with `Result<T, error_stack::Report<C>>`, this PR brings the core functionality of `error-stack` into `ErrorCode`, which adds stacks (aka. context) info to the error when popping up the result thro the call stack.

There is a problem in this code style: the developer usually just not add context where propriate, and then just `?` to throw away the error. `error-stack` resolves this problem by giving the `Report<C>` a phantom type `C`. It enforces the dev to call `.change_context(context)` to convert a `Report<C>` to `Report<C2>`. In this way, we can give every module a `C`, and so forth enforces a `.change_context(context)` at the boundary.

This PR has the exact same design with a slightly different name. `Report<C>` -> `ErrorCode<C>` and `.change_context(context)` to `.with_context(context)`.

Another question is `Why need to change databend_common_exception::exception::Result to std::result::Result`. It's related to the phantom type `C`. Because `ErrorCode<C>` has an extra type `C`, `Result<T, E>` should became `Result<T, ErrorCode<C>>` and `Result<T>` should be an default context `Result<T, ErrorCode<()>>`. This brings a sematic change to `Result<T, E>` which had a meaning of `std::result::Result<T, E>`. So I explicitly qualified these usage of `Result<T, E>` in this PR.

sundy-li (approved):
I approve this approach as it is an appropriate method to introduce error-stack when necessary.

BohuTANG:
@andylokandy 

Another question:
Is it possible to use `with_context` at the top level (similar to how it's used in an HTTP API handler) in the following code:
https://github.com/datafuselabs/databend/blob/main/src/query/service/src/servers/http/v1/query/execute_state.rs#L326-L403

to track all internal error stacks?

If so, how would the code look?

andylokandy:
> Is it possible to use with_context at the top level (similar to how it's used in an HTTP API handler) in the following code to track all internal error stacks?

If you mean rendering the stacks produced by with_context at the top level, yes, it can.

Issue #16310, opened 2024-08-22
Tracking: temporary table


SkyFan2002:
**Summary**

- [x] Initial support in mysql session #16250 
- [x] support in http handler
    - [x]  server-side 
         - [x] #16375 
         - [x] #16281
    - [x]  bendsql  https://github.com/datafuselabs/bendsql/pull/479
    - [x]  databend-go
    - [x] databend-jdbc
- [x] vacuum orphan files of temporary table #16364
- [x] show temporary tables #16364
- [x] support memory engine #16332

SkyFan2002:
@youngsofun Is databend-go and databend-jdbc ready now? Maybe we can add temp table to doc and release it to the users.

youngsofun:
> @youngsofun Is databend-go and databend-jdbc ready now? Maybe we can add temp table to doc and release it to the users.

yes, it is ready now

SkyFan2002:
cc @soyeric128 need doc

PR #16375, opened 2024-09-03
feat: http handler support temp table.

youngsofun:

## Summary

1.	Implement a mechanism to clear data on S3 in case of a crash. Both MySQL and HTTP sessions need to write an ID to metadata with a TTL (time-to-live) and refresh it at regular intervals. The MySQL handler uses a daemon task for each connection, starting when the first query is made. The HTTP handler does it when renewing token.
2.	Both handlers need to clear data when a session ends to free up memory.
3.	The HTTP handler needs to clone the `temp_tbl_mgr` into the global ClientSessionManager and inject it into the session object when the next query arrives. 
4.	Consequently, the HTTP handler needs to release it (and clean up data) when a session times out.  only keep clones that are not empty, so we will not have to scan a lot of items when checking expiration

note for HTTP handler， will return error if  want to create temp table but token is not used (we need logout to free resouces).  waiting for https://github.com/datafuselabs/bendsql/pull/479 and modify to other clients

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16375)

### DIFF
--- a/src/query/service/src/auth.rs
+++ b/src/query/service/src/auth.rs
@@ -94,7 +94,6 @@ impl AuthMgr {
                     let user_info = user_api.get_user(&tenant, identity.clone()).await?;
                     session.set_authed_user(user_info, claim.auth_role).await?;
                 }
-                session.set_client_session_id(claim.session_id.clone());
                 Ok(Some(claim.session_id))
             }
             Credential::Jwt {
--- a/src/query/service/src/servers/http/v1/session/client_session_manager.rs
+++ b/src/query/service/src/servers/http/v1/session/client_session_manager.rs
@@ -75,7 +75,7 @@ impl QueryState {
     pub fn has_expired(&self, now: &Instant) -> bool {
         match self {
             QueryState::InUse => false,
-            QueryState::Idle(t) => (*now - *t) > Duration::from_secs(4 * 3600),
+            QueryState::Idle(t) => (*now - *t) > SESSION_TOKEN_VALIDITY,
         }
     }
 }
@@ -129,10 +129,10 @@ impl ClientSessionManager {
                     guard.remove(id);
                 }
             }
-            for (_id, mgr) in expired {
-                drop_all_temp_tables(mgr).await.ok();
+            for (id, mgr) in expired {
+                drop_all_temp_tables(&id, mgr).await.ok();
             }
-            tokio::time::sleep(Duration::from_secs(900)).await;
+            tokio::time::sleep(SESSION_TOKEN_VALIDITY / 4).await;
         }
     }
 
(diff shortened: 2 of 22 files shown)


src/query/service/src/auth.rs
```diff
@@ -94,7 +94,6 @@ impl AuthMgr {
                     let user_info = user_api.get_user(&tenant, identity.clone()).await?;
                     session.set_authed_user(user_info, claim.auth_role).await?;
                 }
-                session.set_client_session_id(claim.session_id.clone());
```
SkyFan2002, line 97: Why is this removed?

youngsofun: duplicated: also set outside of call to this function

src/query/storages/common/session/src/temp_table.rs
```diff
@@ -120,6 +133,9 @@ impl TempTblMgr {
                 true
             }
         };
+        if self.id_to_table.len() == 1 {
+            self.empty_state_changed = true;
```
SkyFan2002, line 137: Is this redundant?

youngsofun: no,  set to true when changing from empty to non-empty or reverse,  set to false when `is_empty` is called.
so caller add/remove mgr to map only when it is rue, do not have to lookup in map each time.

SkyFan2002:
Let's track progress of http handler and clients in https://github.com/datafuselabs/databend/issues/16310

tests/sqllogictests/suites/temp_table/temp_memory_tables.sql
```diff
@@ -28,7 +28,7 @@ create temp table t2(a int,b int)  engine=MEMORY;
 statement error 1005
 create temp table t2(a INT auto_increment) engine=MEMORY;
 
-statement error 1105
+statement error 1006
```
SkyFan2002, line 31: why is the ErrorCode changed? (just curious)

youngsofun: the format is `statement error <regex>`.
the output of mysql handler also contains an error code of mysql proto.

src/query/service/src/servers/http/v1/session/client_session_manager.rs
```diff
@@ -69,14 +102,40 @@ impl ClientSessionManager {
...
+                    .collect::<Vec<_>>()
+            };
+            {
+                let mut guard = self.session_state.lock();
+                for (id, _) in expired.iter() {
+                    guard.remove(id);
+                }
+            }
+            for (_id, mgr) in expired {
+                drop_all_temp_tables(mgr).await.ok();
```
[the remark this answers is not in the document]
youngsofun: kv in meta expires themselves

SkyFan2002 (approved):

Issue #15901, opened 2024-06-26
Feature: Implement Support for External Dictionaries (OSPP)


b41sh:
**Summary**

- [x] Writing an RFC for the implementation of External Dictionaries https://github.com/datafuselabs/databend-docs/pull/996
- [x] Support parse create/drop/show dictionary stmt.
- [x] Support dictionary meta api.
- [x] Support common functions, such as accessing data in dictionaries via the `dict_get` function.
- [x] Support `Redis` as data sources and implement related DDL.
- [x] Support `MySQL` as data sources and implement related DDL.

Mentor @b41sh

Winnie-Hong0927:
Hello!!(*^▽^*)

b41sh:
@Winnie-Hong0927 Congratulations. We've done all tasks.

BohuTANG:
@b41sh Can we add an new external dictonary: PostgreSQL? Someone from the slack only want to do this in Databend: `CREATE TABLE FROM ... <postgresql tables>` ?

PR #16389, opened 2024-09-04
feat(query): Support access Redis data from dictionaries via the `dict_get` function.

Winnie-Hong0927:

## Summary

This pull request introduces support for accessing Redis data from dictionaries using the `dict_get `function. Other source like MySQL, PostgreSQL will be introduced in the future PRs.

1. **Dictionary Creation with Redis Source**:

```sql
CREATE OR REPLACE DICTIONARY d(key string not null, value string not null) PRIMARY KEY key SOURCE(redis(host='127.0.0.1' port='6379'))
```

options are used to provide the Redis connection URL, `host` and `port` are required options, and `username`, `password`, `db_index` are optional options.

connect to Redis and set some test keys and values

```
redis-cli -h 127.0.0.1 -p 6379
127.0.0.1:6379> set a abc
OK
127.0.0.1:6379> set b def
OK
127.0.0.1:6379>
```

2. **Data Access with `dict_get` Function**:

```sql
MySQL [(none)]> create or replace table t(a string);
Query OK, 0 rows affected (0.201 sec)

MySQL [(none)]> insert into t values('a'),('b'),('c');
Query OK, 3 rows affected (0.095 sec)

MySQL [(none)]> select a, dict_get(d, 'value', a) from t;
+------+---------------------------------+
| a | dict_get(default.d, 'value', a) |
+------+---------------------------------+
| a | abc |
| b | def |
| c | |
+------+---------------------------------+
3 rows in set (0.173 sec)
```

- The result corresponds to the value stored in Redis associated with the keys in table column a.
- If the key is not exists in the Redis, a default value will be inserted instead.

This PR also introduces a `mock_source`, which simulates a Redis server to test the `dict_get` function. Future work will expand this mock data source functionality to include support for other sources such as MySQL, PostgreSQL, and more.

part of https://github.com/datafuselabs/databend/issues/15901

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16389)


src/query/service/src/pipelines/processors/transforms/transform_async_function.rs
```diff
@@ -12,37 +12,85 @@
...
+        operators: BTreeMap<usize, Arc<Operator>>,
+    ) -> Self {
         Self {
             ctx,
             async_func_descs,
+            operators,
         }
     }
 
+    pub fn init_operators(
```
BohuTANG, line 64: Most of new codes in `transform_async_function` is for dictionary, is it possible to spill out a new file to do it?

tests/sqllogictests/suites/query/functions/02_0077_function_dict_get.test
```diff
@@ -0,0 +1,31 @@
...
+create or replace table t(a string)
+
+statement ok
+insert into t values('a'),('b'),('c')
+
+statement ok
+CREATE OR REPLACE DICTIONARY d(key string not null, value string not null) PRIMARY KEY key SOURCE(redis(host='127.0.0.1' port='6379'))
+
+query TT
+select dict_get(d, 'value', 'a') from t
```
b41sh, line 11: modify this test as `select dict_get(d, 'value', a) from t` to test get value from table columns.

tests/sqllogictests/src/mock_source/redis_source.rs
```diff
@@ -0,0 +1,67 @@
...
+    // Connection, provided by `mini-redis`, handles parsing frames from
+    // the socket
+    let mut connection = Connection::new(socket);
+
+    // Use `read_frame` to receive a command from the connection.
+    while let Some(frame) = connection.read_frame().await.unwrap() {
+        let response = match Command::from_frame(frame).unwrap() {
+            Get(cmd) => {
+                // handle get command, get value from mock db.
+                if let Some(value) = db.get(cmd.key()) {
```
sundy-li, line 55: maybe better to return a result with the key prefix.

` format("{}_value", cmd.key) `

src/query/service/src/pipelines/processors/transforms/transform_dictionary.rs
```diff
@@ -0,0 +1,129 @@
...
+                for scalar in column.iter() {
+                    if let ScalarRef::String(key) = scalar {
+                        let buffer = op.read(key).await;
+                        match buffer {
+                            Ok(res) => {
+                                let value =
+                                    unsafe { String::from_utf8_unchecked(res.current().to_vec()) };
+                                builder.push(ScalarRef::String(value.as_str()));
+                            }
+                            Err(_) => {
```
sundy-li, line 110: the error must be handled.

If key is in redis but there is something wrong with redis connection, we should throw the error rather than give a default value.

tests/sqllogictests/suites/query/functions/02_0077_function_dict_get.test
```diff
@@ -0,0 +1,31 @@
...
+
+statement error 1006
+select dict_get(d, 'value11', 'a')
+
+statement error 3114
+select dict_get(test, 'value', 'b')
+
+query T
+SELECT dict_get(d, 'value', 'test')
+----
```
sundy-li, line 30: let's have test:

`SELECT dict_get(d, 'value', 1)`

sundy-li (approved):

b41sh (approved):

PR #16391, opened 2024-09-04
chore: add error stack to cmd entrence

andylokandy:

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16391)


src/query/expression/src/kernels/sort.rs
```diff
@@ -41,9 +41,11 @@ use crate::Scalar;
 
 pub type AbortChecker = Arc<dyn CheckAbort + Send + Sync>;
 
+pub struct AbortError;
```
sundy-li, line 44: So, do we need to create a dummy empty struct to trace the stack?


Can we just use '()' ?

andylokandy: > So, do we need to create a dummy empty struct to trace the stack?

It's recommended to do so.

> Can we just use '()' ?

yes, addressed.

sundy-li (approved):

PR #13779, opened 2023-11-21
feat: optimize vacuum drop impl

lichuang:
I hereby agree to the terms of the CLA available at: https://databend.rs/dev/policies/cla/

## Summary

feat: optimize vacuum drop impl

- Closes #issue

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/13779)

PR #16424, opened 2024-09-09
feat: continue vacuum drop table on per-table cleanup failures

SkyFan2002:

## Summary

 If the cleanup process fails for an individual dropped table, the operation should not immediately terminate. Instead, it should continue to process and clean up other tables that can be handled normally.

If the table's files are not successfully cleaned up, the metadata for both the table and the database will be retained.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16424)


drmingdrmer (changes requested):
> Reviewed 6 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16424)* status: 6 of 10 files reviewed, 2 unresolved discussions (waiting on @dantengsky and @SkyFan2002)

___
*[`src/meta/api/src/schema_api_impl.rs` line 2883 at r1](https://reviewable.io/reviews/datafuselabs/databend/16424#-O6Plsk90kcKwRvmwAPt:-O6Plsk90kcKwRvmwAPu:b-nanf1f) ([raw file](https://github.com/datafuselabs/databend/blob/93b91cdc9bac3461c156bec2def1bc91f6de7661/src/meta/api/src/schema_api_impl.rs#L2883)):*
> ```XML
> 
>                     // if limit is Some, append DroppedId::Db only when table_infos is empty
>                     if drop_db && table_infos.is_empty() {
> ```

Why append `Db` only when `table_infos` is empty? The comment should explain the reason for it.

And IMHO, this method should append a `Db` if `drop_ids` contains all of the `table_infos` belonging to this DB?

Therefore `table_infos.is_empty()` should be replaced with `num >= table_infos.len()`, which meaning the `table_infos` are exhausted. 

My suggestion is replacing `if let Some(left_num) = left_num` with `let num = left_num.unwrap_or(u64::MAX)` and unifying all of the following logic.



_Code quote:_
```XML
                    // if limit is Some, append DroppedId::Db only when table_infos is empty
                    if drop_db && table_infos.is_empty() {
```
___
*[`src/meta/app/src/schema/table.rs` line 926 at r1](https://reviewable.io/reviews/datafuselabs/databend/16424#-O6PhZpbCtnBFlvRf-Q4:-O6PhZpbCtnBFlvRf-Q5:bxp7e1i) ([raw file](https://github.com/datafuselabs/databend/blob/93b91cdc9bac3461c156bec2def1bc91f6de7661/src/meta/app/src/schema/table.rs#L926)):*
> ```XML
> pub enum DroppedId {
>     // db id, db name, (table id, table name)s
>     Db(u64, String, Arc<Vec<(u64, String)>>),
> ```

Using a tuple can be a bit complicated. Consider converting it into a struct-like variant instead:
```rust
Db {
    database_id: u64,
    database_name: String,
    ...
}
```

And why does it have to be an `Arc<Vec<_>>`? Does it need to be shared between thread?
The number of tables would be small(like several thousand?), just using `Vec<_>` would be all right for this small data set.

SkyFan2002:
> Why append Db only when table_infos is empty? The comment should explain the reason for it.
@drmingdrmer Not sure, this is not introduced in this PR, but introduced in #13779

drmingdrmer (changes requested):
> Reviewed 2 of 10 files at r1, 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16424)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @dantengsky and @SkyFan2002)

___
*[`src/meta/api/src/schema_api_impl.rs` line 2888 at r2](https://reviewable.io/reviews/datafuselabs/databend/16424#-O6QEhfMAa3wkypDWYFr:-O6QEhfMAa3wkypDWYFs:b-jzoy40) ([raw file](https://github.com/datafuselabs/databend/blob/5e2a9f09c7e802a75c9db6490b47e6d9ab6365d4/src/meta/api/src/schema_api_impl.rs#L2888)):*
> ```XML
>                             *db_id,
>                             table_info.ident.table_id,
>                             table_info.name.clone(),
> ```

`this_db_drop_table_infos` should always be pushed into `drop_ids`, no matter to drop the db or not.



_Code quote:_
```XML
                    for (table_info, db_id) in this_db_drop_table_infos {
                        drop_ids.push(DroppedId::Table(
                            *db_id,
                            table_info.ident.table_id,
                            table_info.name.clone(),
```

drmingdrmer (changes requested):
> Reviewed 1 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16424)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on @dantengsky and @SkyFan2002)

___
*[`src/meta/api/src/schema_api_impl.rs` line 2870 at r3](https://reviewable.io/reviews/datafuselabs/databend/16424#-O6QOwik1EWMqEZ2FkzN:-O6QOwik1EWMqEZ2FkzO:b-94n77o) ([raw file](https://github.com/datafuselabs/databend/blob/2f114112663e7507842f91c403f5c2de784325b9/src/meta/api/src/schema_api_impl.rs#L2870)):*
> ```XML
>                 let take_num = left_num.unwrap_or(usize::MAX);
> 
>                 if drop_db && take_num > table_infos.len() {
> ```

Add some comment to explain this condition:



_Suggestion:_
```XML
                // A DB can be removed only when all its tables are removed.
                if drop_db && take_num > table_infos.len() {
```

drmingdrmer (approved):
> *[Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16424)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on @dantengsky and @SkyFan2002)

dantengsky (approved):

PR #16443, opened 2024-09-12
feat: Auto compact(re-cluster) for multiple table insertion statement

SkyFan2002:

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16443)

### DIFF
--- a/src/query/settings/src/settings_default.rs
+++ b/src/query/settings/src/settings_default.rs
@@ -593,7 +593,7 @@ impl DefaultSettings {
                 (
                     "enable_compact_after_multi_table_insert",
                     DefaultSettingValue {
-                        value: UserSettingValue::UInt64(1),
+                        value: UserSettingValue::UInt64(0),
                         desc: "Enables recluster and compact after multi-table insert.",
                         mode: SettingMode::Both,
                         range: Some(SettingRange::Numeric(0..=1)),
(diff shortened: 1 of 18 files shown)


src/query/settings/src/settings_default.rs
```diff
@@ -590,6 +590,15 @@ impl DefaultSettings {
                     mode: SettingMode::Both,
                     range: Some(SettingRange::Numeric(0..=1)),
                 }),
+                (
```
zhyass, line 593: Maybe just use enable_compact_after_write, no need to add new settings.

SkyFan2002: Multi-table insert is special. The hook needs to compact and recluster multiple tables. To avoid potential performance degradation, a seperate setting is introduced.

src/query/settings/src/settings_default.rs
```diff
@@ -590,6 +590,15 @@ impl DefaultSettings {
...
+                    "enable_compact_after_multi_table_insert",
+                    DefaultSettingValue {
+                        value: UserSettingValue::UInt64(1),
```
dantengsky, line 596: let's disable it by default, since some users are expecting no auto compaction for multi-table insertions

dantengsky (approved):

PR #16466, opened 2024-09-19
feat(query): push rank limit into aggregate partial node

sundy-li:

## Summary

This pr pushes rank limit into aggregate partial node.

For example:

```sql
SELECT UserID, SearchPhrase, COUNT(*) FROM hits GROUP BY UserID, SearchPhrase ORDER BY UserID,  SearchPhrase 
 LIMIT 10;
```

As the group keys may have high cardinality, the HT will resize multiple times to hold the coming keys which is inefficient.

Now we will filter group keys using partial sort with `LimitType::Rank(n)` to keep top n unique keys.

So we inject a `SortPartialTransform` before partial agg now.
```
🐳 :) explain pipeline SELECT UserID, SearchPhrase, COUNT(*) FROM hits GROUP BY UserID, SearchPhrase ORDER BY UserID,  SearchPhrase
 LIMIT 10;
CompoundBlockOperator(Project) × 1
  LimitTransform × 1
    Merge to MultiSortMerge × 1
      TransformSortSpill × 16
        TransformSortMergeLimit × 16
          SortPartialTransform × 16
            TransformFinalAggregate × 16
              TransformSpillReader × 16
                Merge to Resize × 16
                  Merge to TransformPartitionBucket × 1
                    TransformAggregateSpillWriter × 16
                      TransformPartialAggregate × 16
                        SortPartialTransform × 16
                          NativeDeserializeDataTransform × 16
                            SyncReadNativeDataSource × 16

```

Performance in my PC is  2.8 sec -->  0.35 sec.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16466)


src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_limit_aggregate.rs
```diff
@@ -45,45 +48,182 @@ impl RulePushDownLimitAggregate {
     pub fn new() -> Self {
         Self {
             id: RuleID::PushDownLimitAggregate,
```
xudong963, line 50: It's better to give the rule a new name.

src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_limit_aggregate.rs
```diff
@@ -45,45 +48,182 @@ impl RulePushDownLimitAggregate {
...
+
+    fn apply(
+        &self,
+        s_expr: &SExpr,
+        state: &mut TransformResult,
+    ) -> databend_common_exception::Result<()> {
+        match s_expr.plan().rel_op() {
+            RelOp::Limit => self.apply_limit(s_expr, state),
+            RelOp::Sort => self.apply_sort(s_expr, state),
+            RelOp::EvalScalar => self.apply_sort(s_expr, state),
```
xudong963, line 223: `RelOp::Sort | RelOp::EvalScalar => self.apply_sort(s_expr, state)`

src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_limit_aggregate.rs
sundy-li: How about `RulePushDownRankLimitAggregate` ?

xudong963: make sense

src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_limit_aggregate.rs
```diff
@@ -21,69 +20,209 @@ use crate::optimizer::rule::TransformResult;
...
-            agg_limit.limit = Some(agg_limit.limit.map_or(count, |c| cmp::max(c, count)));
-            let agg = SExpr::create_unary(
+            if !is_order_subset {
+                return Ok(());
+            }
+            let mut sort_items = Vec::with_capacity(agg_limit.group_items.len());
+            let mut not_found_sort_items = vec![];
+            for i in 0..agg_limit.group_items.len() {
+                let group_item = &agg_limit.group_items[i];
+                if let Some(sort_item) = sort.items.iter().find(|k| k.index == group_item.index) {
```
Dousir9, line 173: It looks like we can avoid using `if let Some()`, since `is_order_subset` is true here.

sundy-li: eg:
```
select a, b from t group by a,b  order by b desc;
```

`b` is subset of `a,b`, here we inject sort rank `b desc, a asc` for better filter ratio in sort partial.

xudong963 (approved):

Dousir9 (approved):

src/query/sql/src/planner/optimizer/rule/rewrite/rule_push_down_limit_aggregate.rs
Dousir9: > eg:
> 
> ```
> select a, b from t group by a,b  order by b desc;
> ```
> 
> `b` is subset of `a,b`, here we inject sort rank `b desc, a asc` for better filter ratio in sort partial.

I misunderstood this, thanks for the explanation.

PR #16493, opened 2024-09-23
chore(query): refactor like function

sundy-li:

## Summary

1. refactor like function in filter and functions
2. make spilt agg run at the last of rule optimizations
3. use multiversion to dispatch sum agg
4. introduce `volnitsky` search algorithm

perf: 

```
select count() from numbers(100000000) where number::string like '%12345%';
```
improves ~10%

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16493)


src/query/functions/src/aggregates/aggregate_unary.rs
```diff
@@ -194,24 +205,19 @@ where
         _input_rows: usize,
     ) -> Result<()> {
         let column = T::try_downcast_column(&columns[0]).unwrap();
-        let column_iter = T::iter_column(&column);
         let state: &mut S = place.get::<S>();
         match validity {
             Some(bitmap) => {
```
Dousir9, line 210: If all bits in the `bitmap` are true, we can also use `add_batch` to perform SIMD sum.

Dousir9 (approved):
Rest LGTM !

src/query/functions/src/scalars/comparison.rs
```diff
@@ -445,48 +444,13 @@ fn register_like(registry: &mut FunctionRegistry) {
...
-                }
-                LikePattern::ComplexPattern => {
-                    jsonb::traverse_check_string(val, |v| LikePattern::complex_pattern(v, pat))
-                }
-            }
+        variant_vectorize_like(|val, pattern_type| match pattern_type {
+            LikePattern::OrdinalStr(_)
+            | LikePattern::StartOfPercent(_)
+            | LikePattern::EndOfPercent(_)
+            | LikePattern::Constant(_) => pattern_type.compare(val),
```
b41sh, line 451: `val` is binary format, we need use `jsonb::as_str` function to convert it to a string value, and then run `compare` function.

b41sh (approved):

PR #16504, opened 2024-09-24
feat: add fast path for cross join to optimize tpcds q23 (~50x)

xudong963:

## Summary

This PR adds a fast path for cross join when build side only has a row, which matches tpcds q23's pattern(build side is from a max/sum agg).

By the fast path, we can avoid memory reallocation. Then reduce cache miss and the ability of SMD for next operators. 

Before: 26s
image

Now: 0.4s
img_v3_02f0_b8e12fc1-0406-40c7-a18b-cd80033e426g

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/hash_join/probe_join/cross_join.rs
+++ b/src/query/service/src/pipelines/processors/transforms/hash_join/probe_join/cross_join.rs
@@ -35,8 +35,19 @@ impl HashJoinProbeState {
         if build_num_rows == 0 || input_num_rows == 0 {
             return Ok(vec![]);
         }
-        let probe_block = input.project(&self.probe_projections);
+        let mut probe_block = input.project(&self.probe_projections);
         let build_block = DataBlock::concat(build_blocks)?;
+        if build_num_rows == 1 {
+            for col in build_block.columns() {
+                let value_ref = col.value.as_ref();
+                let scalar = unsafe { value_ref.index_unchecked(0) };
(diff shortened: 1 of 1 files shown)


src/query/service/src/pipelines/processors/transforms/hash_join/probe_join/cross_join.rs
```diff
@@ -35,8 +35,19 @@ impl HashJoinProbeState {
         if build_num_rows == 0 || input_num_rows == 0 {
             return Ok(vec![]);
         }
-        let probe_block = input.project(&self.probe_projections);
+        let mut probe_block = input.project(&self.probe_projections);
         let build_block = DataBlock::concat(build_blocks)?;
+        if build_num_rows == 1 {
+            for col in build_block.columns() {
```
sundy-li, line 41: can use `build_block.slice(0,1)`

sundy-li: or just return the `build_block `

Dousir9: Need to construct `Value::Scalar` instead of `Value::Column`.

PR #16506, opened 2024-09-24
chore: simplify SchemaApi::truncate_table()

drmingdrmer:

##### refactor: when SchemaApi::truncate_table(), no need to assert table seq

truncate_table() first lists all copied files belonging to a table,
then deletes them in small chunks. The delete operations assert that
the seq of each file does not change.

With this approach, there is no need to assert that the seq of the
containing table does not change.

##### refactor: SchemaApi::list_tables() should specify db-id instead of db-name

- Change list_tables() to accept db_id instead of db_name for better precision
- Modify return type from TableInfo to (table_name, table_id,
  SeqV<table_meta>).
  Because some info in TableInfo can not be provided by SchemaApi

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Refactoring

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16506)


TCeason (approved):

TCeason:
truncate_table is lgtm

PR #16510, opened 2024-09-24
fix(query): fix planner cache cause modify async function does not take effect

b41sh:

## Summary

- The plan of search function, async function and udf contains some arguments defined in meta, which may be modified by user at any time. If the plan is cached, the modified arguments cannot take effect and a wrong result will be displayed. Therefore, sql containing these types of functions is made non-cacheable.
- allow async function execute in where clause
- fix async function as name could not be displayed 
- replace `tantivy`, `tantivy-jieba`, `ethnum` to `datafuse-extras`

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16510)

### DIFF
--- a/src/query/sql/src/planner/planner_cache.rs
+++ b/src/query/sql/src/planner/planner_cache.rs
@@ -24,6 +24,7 @@ use databend_common_ast::ast::TableReference;
 use databend_common_catalog::table_context::TableContext;
 use databend_common_expression::Scalar;
 use databend_common_expression::TableSchemaRef;
+use databend_common_functions::is_cacheable_function;
 use databend_common_settings::ChangeValue;
 use databend_storages_common_cache::CacheAccessor;
 use databend_storages_common_cache::CacheValue;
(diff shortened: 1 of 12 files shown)


src/query/sql/src/planner/planner_cache.rs
```diff
@@ -162,8 +163,9 @@ impl TableRefVisitor {
             return;
         }
 
-        // If the function is score, we should not cache the plan
-        if func.name.name.to_lowercase() == "score" {
+        let func_name = func.name.name.to_lowercase();
+        // If the function is not suitable for caching, we should not cache the plan
+        if !is_cacheable_function(&func_name) || func_name == "score" {
```
sundy-li, line 168: let's move "score" into `is_cacheable_function`

sundy-li (approved):

src/query/sql/src/planner/planner_cache.rs
b41sh: ok

PR #16523, opened 2024-09-25
refactor: migrage Bitmap from arrow2 to arrow-rs

andylokandy:

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/datafuselabs/databend/16523)

### DIFF
--- a/src/query/expression/src/types/boolean.rs
+++ b/src/query/expression/src/types/boolean.rs
@@ -15,7 +15,8 @@
 use std::cmp::Ordering;
 use std::ops::Range;
 
-use databend_common_arrow::arrow::bitmap::Bitmap;
+use arrow::array::BooleanBufferBuilder;
+use arrow::buffer::BooleanBuffer;
 use databend_common_arrow::arrow::bitmap::MutableBitmap;
 
 use crate::property::Domain;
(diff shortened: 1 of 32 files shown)


src/query/expression/src/types/boolean.rs
```diff
@@ -36,10 +37,10 @@ pub struct BooleanType;
 impl ValueType for BooleanType {
     type Scalar = bool;
     type ScalarRef<'a> = bool;
-    type Column = Bitmap;
+    type Column = BooleanBuffer;
```
sundy-li, line 40: Why not use `NullBuffer`, it has materialized null_count.

andylokandy: I can not find a builder for `NullBuffer` and it does not `impl BitAnd`

sundy-li: 1. `NullBuffer::new()`
2. `pub fn union(lhs: Option<&NullBuffer>, rhs: Option<&NullBuffer>) -> Option<NullBuffer> `

sundy-li: Also: 
`impl From<BooleanBuffer> for NullBuffer`

andylokandy: But these all operation is direct on `BooleanBuffer`: it implments BitAnd, it provides `value_unchecked`, and have `BooleanBufferBuilder`. I think `BoolteanBuffer` is the directly replacement for `Bitmap` in arrow2.

Dousir9:
Please write a rust benchmark to test the migrated functions in kernels(`take.rs, concat.rs, ...`) to ensure that there is no performance regression.

PR #16525, opened 2024-09-26
feat: copy option 'pattern' support variable.

youngsofun:

## Summary

```
select $1 from @data/csv/ (pattern => $pt, file_format=>'csv')
copy into t1 from @data/csv/ pattern= $pt file_format=(type=csv)
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16525)


src/query/ast/src/parser/copy.rs
```diff
@@ -144,10 +145,9 @@ fn copy_into_table_option(i: Input) -> IResult<CopyIntoTableOption> {
             rule! { FILES ~ "=" ~ "(" ~ #comma_separated_list0(literal_string) ~ ")" },
             |(_, _, _, files, _)| CopyIntoTableOption::Files(files),
         ),
-        map(
-            rule! { PATTERN ~ "=" ~ #literal_string },
-            |(_, _, pattern)| CopyIntoTableOption::Pattern(pattern),
-        ),
+        map(rule! { PATTERN ~ "=" ~ #expr }, |(_, _, pattern)| {
```
youngsofun, line 148:
```
copy into ti from @data/ontime_200.csv pattern = 'x' FILE_FORMAT = (type = CSV);
```
works

but `(type = CSV skip_header=1)` fail like this:


related code:

```
1 | copy into ti from @data/ontime_200.csv pattern = 'x' FILE_FORMAT = (type = CSV skip_header=1)
  | ----                                             ---               -           ^^^^^^^^^^^ unexpected `skip_header`, expecting `)` or `,`
  | |                                                |                 |            
  | |                                                |                 while parsing `(<expr> [, ...])`
  | |                                                while parsing expression
```


```
alt (
        map(rule! { PATTERN ~ "=" ~ #expr }, |(_, _, pattern)| {
            CopyIntoTableOption::Pattern(pattern)
        }),
        map(rule! { #file_format_clause }, |options| {
            CopyIntoTableOption::FileFormat(options)
        }),
      ...
```

youngsofun: @andylokandy need your help

andylokandy: May need to make `FILE_FORMAT` reserved ident

youngsofun: what about supporting:

```
 pattern = (variable_name ='<variable_name>' |  value='<pattern>' ) 
```

```pattern = '<pattern>'  ```   as a short.

which is close to 
```
FILE_FORMAT = (  FORMAT_NAME = '<file_format_name>' | TYPE = { CSV | Parquet} )
```

cc @sundy-li

Comment (sundy-li) [reply src/query/ast/src/parser/copy.rs:148]: I don't think it's a good idea. Let's make `FILE_FORMAT ` as reserved ident to continue the parser.

Comment (andylokandy) [reply src/query/ast/src/parser/copy.rs:148]: Not only the `FILE_FORMAT` but also any other expr-stopper after `pattern`...

Comment (youngsofun) [reply src/query/ast/src/parser/copy.rs:148]: not only `FILE_FORMAT`,   other current and future copy options need to be `reserved ident`， including `FILES`, `MAX_FILES`， note that this is a breaking change. This might cause an error in a SQL query that would have otherwise run correctly for the user. what is worse is we may add new options later.

In fact we do not need a expr here, `literal | variable` is enough. this is because the expr here can not ref any column, and it is only used when user can not construct the sql in programing language like python/java.

so I think, a helper func like below, although not elegant, fits our needs.


```
// allow only `'<literal_string>'` or `$<variable_ident>` in parser

pub fn literal_string_or_variable(i: Input) -> IResult<Expr> {
    alt((
        map(literal_string, |s| Expr::Literal {
            span: None,
            value: Literal::String(s),
        }),
        map(variable_ident, |s| make_func_get_variable(None, s)),
    ))
}
```

by the way, `variable_ident` now allow `$ var_name` (ws between $ and var_name), is this by design?

sundy-li: We should add `TokenKind::VariableIdent` to omit the spaces.

sundy-li (approved):

PR #16537, opened 2024-09-27
fix(query): fix column leaf_index

Dousir9:

## Summary

Use column id as column `leaf_index`.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16537)

### DIFF
--- a/src/query/sql/src/planner/metadata.rs
+++ b/src/query/sql/src/planner/metadata.rs
@@ -229,7 +229,7 @@ impl Metadata {
         data_type: TableDataType,
         table_index: IndexType,
         path_indices: Option<Vec<IndexType>>,
-        leaf_index: Option<IndexType>,
+        column_id: Option<u32>,
         column_position: Option<usize>,
         virtual_computed_expr: Option<String>,
     ) -> IndexType {
@@ -241,7 +241,7 @@ impl Metadata {
             column_index,
             table_index,
             path_indices,
-            leaf_index,
+            column_id,
             virtual_computed_expr,
         });
         self.columns.push(column_entry);
@@ -370,7 +370,6 @@ impl Metadata {
             }
         }
 
-        // build leaf index in DFS order for primitive columns.
         while let Some((indices, field)) = fields.pop_front() {
             if indices.is_empty() {
                 self.add_base_table_column(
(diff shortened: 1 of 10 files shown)


src/query/sql/src/planner/metadata.rs
```diff
@@ -402,12 +400,13 @@ impl Metadata {
...
-                    None,
+                    Some(field.column_id),
                     None,
                     None,
                 );
 
                 let mut i = fields_type.len();
+                let mut inner_column_id = field.column_id;
                 for (inner_field_name, inner_field_type) in
                     fields_name.iter().zip(fields_type.iter()).rev()
```
b41sh, line 411: Remove `rev()` here makes the `fields_name` and `fields_type` loop from left to right, and changes `i` to incrementing from 0, because we don't need to compute the `leaf_index` anymore, and the inner field can iterate in the normal order.

src/query/sql/src/planner/metadata.rs
```diff
@@ -420,7 +419,12 @@ impl Metadata {
                         field.name(),
                         display_tuple_field_name(inner_field_name)
                     );
-                    let inner_field = TableField::new(&inner_name, inner_field_type.clone());
+                    inner_column_id += inner_field_type.num_leaf_columns() as u32;
```
b41sh, line 422: The `inner_column_id` should be added by 1, and then if `num_leaf_columns` are greater than 1, then `num_leaf_columns` - 1 should be added after the inner field is generated, used to correctly generate the next `inner_column_id`.

src/query/sql/src/planner/metadata.rs
Dousir9: Will this cause a breaking change ?

src/query/sql/src/planner/metadata.rs
Dousir9: Thanks !

src/query/sql/src/planner/metadata.rs
b41sh: will not cause a breaking change, `leaf_index` is not used anymore.

sundy-li (approved):

xudong963 (approved):

b41sh (approved):

PR #16573, opened 2024-10-09
feat(query): add json array lambda functions

sundy-li:

## Summary

1. Support cast functions that cast from variant type into `Array<T>`
2. Add json array lambda functions:

```
    "json_array_transform",
    "json_array_apply",
    "json_array_map",
    "json_array_filter",
    "json_array_reduce",
```

Similar to the array lambda function, it accepts variant type as the first argument and returns the variant type.

```
select json_array_reduce([1,2,3,4]::Variant, (x, y) -> 3 + x + y);
----
19

select json_array_map([1,2,3,4]::Variant, x -> x + 1);
----
[2,3,4,5]

select json_array_filter([1,2,3,4]::Variant, x -> x % 2 = 0);
----
[2, 4]
```

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16573)


sundy-li:
docs @soyeric128

src/query/expression/src/evaluator.rs
```diff
@@ -502,6 +505,85 @@ impl<'a> Evaluator<'a> {
...
+            (DataType::Variant, DataType::Array(inner_dest_ty)) => {
+                let empty_vec = vec![jsonb::Value::Null];
+                match value {
+                    Value::Scalar(Scalar::Variant(x)) => {
+                        let array = jsonb::from_slice(&x).map_err(|e| {
+                            ErrorCode::BadArguments(format!(
+                                "Expect to be valid json, got err: {e:?}"
+                            ))
+                        })?;
+                        let array = array.as_array().unwrap_or(&empty_vec);
```
b41sh, line 517: if the value is not a json array, we'd better cast it as an empty array instead of an array with json null, this may lead to unexpected results.
for example:
```sql
MySQL [(none)]> select json_array_transform(parse_json('"aa"'), data -> CONCAT(data::String, 'bend'));
+--------------------------------------------------------------------------------+
| json_array_transform(parse_json('"aa"'), data -> concat(data::STRING, 'bend')) |
+--------------------------------------------------------------------------------+
| ["nullbend"]                                                                   |
+--------------------------------------------------------------------------------+
1 row in set (0.048 sec)
```

src/query/expression/src/evaluator.rs
```diff
@@ -502,6 +505,85 @@ impl<'a> Evaluator<'a> {
...
+                            .into_column()
+                            .unwrap();
+                        Ok(Value::Scalar(Scalar::Array(new_array)))
+                    }
+                    Value::Column(Column::Variant(col)) => {
+                        let mut array_builder =
+                            ArrayType::<VariantType>::create_builder(col.len(), &[]);
+
+                        for x in col.iter() {
+                            let array = jsonb::from_slice(x).map_err(|e| {
```
b41sh, line 545: need check validity first, NULL value is an empty slice, which will case `jsonb::from_slice` return an error.

b41sh (approved):

PR #16581, opened 2024-10-09
fix: merge into panic

zhyass:

## Summary

The reason is that the condition is false, after `RuleEliminateFilter`, it is converted to `ConstantTableScan`, which makes the fileds in mutation invalid.

```sql
mysql> create table t(a int);

mysql> create table s(a int);

mysql> merge into s using t on 1 != 1 when matched and s.a>0 then update * when not matched then insert *;
+-------------------------+------------------------+
| number of rows inserted | number of rows updated |
+-------------------------+------------------------+
|                       0 |                      0 |
+-------------------------+------------------------+
1 row in set (0.15 sec)
Read 0 rows, 0.00 B in 0.067 sec., 0 rows/sec., 0.00 B/sec.
```
- fixes: #16588 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16581)

### DIFF
--- a/tests/sqllogictests/suites/base/09_fuse_engine/09_0042_merge_into_issue_16588.test
+++ b/tests/sqllogictests/suites/base/09_fuse_engine/09_0042_merge_into_issue_16588.test
@@ -1,8 +1,8 @@
 statement ok
-create or replace database i16581;
+create or replace database i16588;
 
 statement ok
-use i16581;
+use i16588;
 
 statement ok
 create table base(a int);
@@ -10,10 +10,10 @@ create table base(a int);
 statement ok
 create table sink(a int);
 
-query I
+query II
 merge into sink using base on 1 != 1 when matched then update * when not matched then insert *;
 ----
-0
+0 0
 
 query I
 merge into sink using base on 1 != 1 when matched then update *;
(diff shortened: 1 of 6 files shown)


wubx:
```
merge into s using t on 1 != 1 when matched and s.a>0 then update * when not matched then insert *;
```
This SQL sense will update unexpected behavior，make error report mistakes more friendly is ok ?

tests/sqllogictests/suites/base/09_fuse_engine/09_0042_merge_into_issue_16588.test
```diff
@@ -0,0 +1,24 @@
...
+use i16588;
+
+statement ok
+create table base(a int);
+
+statement ok
+create table sink(a int);
+
+query II
+merge into sink using base on 1 != 1 when matched then update * when not matched then insert *;
```
Dousir9, line 14: It would be better to add `explain` test for this sql.

zhyass: > It would be better to add `explain` test for this sql.

Thanks, Done!

Dousir9 (approved):

PR #16586, opened 2024-10-10
refactor: SchemaApi::get_table_meta_history()

drmingdrmer:

##### refactor: SchemaApi::get_table_meta_history()

It does not need to accept a `database-name` argument just for building
an error.

It does not need to return application level error: if a table is not
found, just return an empty list.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Refactoring

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16586)

### DIFF
--- a/src/meta/api/src/schema_api_impl.rs
+++ b/src/meta/api/src/schema_api_impl.rs
@@ -1547,36 +1547,20 @@ impl<KV: kvapi::KVApi<Error = MetaError> + ?Sized> SchemaApi for KV {
 
     #[logcall::logcall]
     #[fastrace::trace]
-    async fn get_table_meta_history(
+    async fn get_retainable_tables(
         &self,
-        database_name: &str,
-        table_id_history: &TableIdHistoryIdent,
-    ) -> Result<Vec<(TableId, SeqV<TableMeta>)>, KVAppError> {
-        let table_name = &table_id_history.table_name;
+        history_ident: &TableIdHistoryIdent,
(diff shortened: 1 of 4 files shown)


src/meta/api/src/schema_api_impl.rs
```diff
@@ -1547,36 +1547,20 @@ impl<KV: kvapi::KVApi<Error = MetaError> + ?Sized> SchemaApi for KV {
...
-                format!("get_table_history: {}.{}", database_name, table_name),
-            ))
-        })?;
-
-        let now = Utc::now();
-
-        let metas = get_table_meta_history(self, &now, table_id_list).await?;
+        history_ident: &TableIdHistoryIdent,
+    ) -> Result<Vec<(TableId, SeqV<TableMeta>)>, MetaError> {
+        let Some(seq_table_id_list) = self.get_pb(history_ident).await? else {
```
TCeason, line 1554: So this means : if a table under a db which contains 100 tables is invalid, list table from this db will return empty result? 

cc @flaneur2020

drmingdrmer: I do think so. 🤔

drmingdrmer: my mistake。the method name should be get-retainable-x x x

drmingdrmer: Fixed. Please review it again

TCeason (approved):

PR #16592, opened 2024-10-10
refactor: isolate session_id and session states of diff users.

youngsofun:

## Summary

By putting session id under user_name when used as a key. 

- fixes: https://github.com/databendlabs/databend/issues/16565

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16592)


drmingdrmer (commented):
> Reviewed 1 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/16592)* status: 1 of 11 files reviewed, 2 unresolved discussions (waiting on @youngsofun)

___
*[`src/meta/app/src/principal/client_session_ident.rs` line 75 at r1](https://reviewable.io/reviews/databendlabs/databend/16592#-O8t9P468qTaUIsO8ibh:-O8t9P468qTaUIsO8ibi:br3rsr1) ([raw file](https://github.com/databendlabs/databend/blob/147beedca67410c298e32cf1c07e5aa43ae48fa9/src/meta/app/src/principal/client_session_ident.rs#L75)):*
> ```XML
>         Self::parse(&s)
>     }
> }
> ```

There is no need to manually encode the `session_id` and `user_name` with a colon. Just treat them as two key segment:



_Suggestion:_
```XML


impl KeyCodec for UserSessionId {
    fn encode_key(&self, b: KeyBuilder) -> KeyBuilder {
        b.push_str(&self.session_id).push_str(&self.user_name)
    }

    fn decode_key(parser: &mut KeyParser) -> Result<Self, KeyError>
    where Self: Sized {
        let s = parser.next_str()?;
        let u = parser.next_str()?;
        Ok(Self{ session_id: s, user_name: u})
    }
}
```
___
*[`src/meta/protos/proto/client_session.proto` line 26 at r1](https://reviewable.io/reviews/databendlabs/databend/16592#-O8tA0R181irjij6KOGM:-O8tA0R181irjij6KOGN:bqwqhhp) ([raw file](https://github.com/databendlabs/databend/blob/131d2d894e84fd8721b94c5384bea41d98e621eb/src/meta/protos/proto/client_session.proto#L26)):*
> ```proto
>   uint64 ver = 100;
>   uint64 min_reader_ver = 101;
>   string user_name = 2;
> ```

A removed field should be reserved to prevent further use with: https://protobuf.dev/programming-guides/proto3/#reserved

youngsofun:
>>>  Just treat them as two key segment
good idea, thank you

>>> A removed field should be reserved to prevent further use with: https://protobuf.dev/programming-guides/proto3/#reserved

this is not used in production for now,  can I skip the step and delete it directly?

drmingdrmer:
> > > > Just treat them as two key segment
> > > > good idea, thank you
> 
> > > > A removed field should be reserved to prevent further use with: https://protobuf.dev/programming-guides/proto3/#reserved
> 
> this is not used in production for now, can I skip the step and delete it directly?

I'd suggest always following the rule if it doesn't create additional work. 
Although it's unlikely to happen, there's still an odd that a binary with this field has been deployed somewhere we're unaware of. It's better to be cautious.

youngsofun:
@drmingdrmer  Your suggestions have all been adopted. please review again.

SkyFan2002 (approved):

drmingdrmer (approved):
> Reviewed 5 of 11 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/16592)* status: all files reviewed, 2 unresolved discussions (waiting on @youngsofun)

___
*[`src/query/service/src/servers/http/v1/session/client_session_manager.rs` line 111 at r2](https://reviewable.io/reviews/databendlabs/databend/16592#-O98DQFF-CX4QpNlXL_n:-O98DQFF-CX4QpNlXL_o:b4367to) ([raw file](https://github.com/databendlabs/databend/blob/a72b8b5aaf552b57a595e4c6864de497aaf08a31/src/query/service/src/servers/http/v1/session/client_session_manager.rs#L111)):*
> ```XML
> 
>     pub fn state_key(client_session_id: &str, user_name: &str) -> String {
>         format!("{client_session_id}:{user_name}")
> ```

The `client_session_id` and `user_name` should be escaped to prevent any `:` characters within them from interfering with the key format.

And if this `state_key` is not used for transport, the return value can be just any `Hash` type, such as `(client_session_id, user_name)`, which can be used as hash map key.



_Code quote:_
```XML
    pub fn state_key(client_session_id: &str, user_name: &str) -> String {
        format!("{client_session_id}:{user_name}")
```

youngsofun:
@drmingdrmer you are right, I will refactor it to return simply `(client_session_id, user_name)` in next pr coming soon.

PR #16632, opened 2024-10-17
fix: deadlock in table lock

zhyass:

#### Problem Overview
This PR addresses a deadlock issue that occurs when `max_running_queries` is set to 1 in Databend, resulting in errors like "table locked by other session." The root cause of the deadlock is the inconsistent timing of acquiring table locks during query execution, which leads to conflicts in acquiring the necessary locks.

#### Root Cause
The deadlock happens because queries acquire table locks at different stages of query processing:
- Some queries acquire the lock during the **binder phase**.
- Others acquire it during the **interpreter phase**.

This inconsistency creates a situation where:
- A query already in the execution queue (and actively running) may not be able to acquire a lock.
- Meanwhile, a query still waiting in the queue may hold the required table lock, blocking the active query and causing a deadlock.

#### Conditions That Trigger the Issue
- The issue is most likely to occur when the **`max_running_queries`** setting is set to 1.
- Under these conditions, only one query is allowed to run at a time, exacerbating lock contention between queued and running queries.

#### Solution
To resolve the deadlock, this PR introduces a new lock acquisition strategy **for queries that need require a table lock**(merge into, recluster, update, delete, compact, truncate, replace, ..): 
- **Queueing After SQL Parsing:** For queries that require a table lock, they are queued immediately after the SQL statement is parsed, rather than waiting after binder.
- **Lock Acquisition After Queuing:** The table lock is only acquired once the query exits the queue and begins execution. This ensures that the query can safely acquire the lock when needed, without blocking other queries.

This change ensures a more consistent and deadlock-free process of table lock acquisition across different types of queries.

#### Impact of the Fix
- The solution avoids the race condition between queued queries and running queries by ensuring that the lock acquisition only happens after queuing, preventing deadlocks and improving overall query execution stability in scenarios with limited concurrency.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16632)


zhang2014:
Can we add new check for sampling queries in the `is_heavy_action` function to solve this?

dantengsky (approved):

PR #16672, opened 2024-10-23
fix(query): spill hang

forsaken628:

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/transform_sort_spill.rs
+++ b/src/query/service/src/pipelines/processors/transforms/transform_sort_spill.rs
@@ -182,6 +182,7 @@ where
                         self.state = State::Merging;
                     }
                     self.input_data = Some(block);
+                    self.input.set_need_data();
                     Ok(Event::Async)
                 }
                 _ => unreachable!(),


src/query/service/src/pipelines/processors/transforms/transform_sort_spill.rs
```diff
@@ -182,6 +182,7 @@ where
                         self.state = State::Merging;
                     }
                     self.input_data = Some(block);
+                    self.input.set_need_data();
```
zhang2014, line 185: Maybe `self.input.finish()` is better if `meta.is_none()`

forsaken628: Here we need to mark the last block to implement some optimisations

wubx:
I tested this work fine.

sundy-li (approved):

PR #16701, opened 2024-10-28
fix(metrics): use f64::MAX as histogram bound to display +Inf properly

flaneur2020:

## Summary

on rendering the prometheus metrics, `prometheus-client` checks the upper bound with f64::MAX, and renders it as `+Inf`, which is a marker in calculating `histogram_quantile` on monitoring.

if `+Inf` is not displayed properly, it seems that `histogram_quantile` will return nothing.

the source look likes this:

```
            self.write_prefix_name_unit()?;
            self.write_suffix("bucket")?;

            if *upper_bound == f64::MAX {
                self.encode_labels(Some(&[("le", "+Inf")]))?;
            } else {
                self.encode_labels(Some(&[("le", *upper_bound)]))?;
            }

            self.writer.write_str(" ")?;
            self.writer
                .write_str(itoa::Buffer::new().format(cummulative))?;
```

to display `+Inf` properly, we need let the upper bound with `f64::MAX`.

## Tests

- [x] Unit test

## Type of change

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16701)


zhang2014 (approved):

zhang2014:
Cloud test passed

PR #16722, opened 2024-10-29
feat: filter null value before join

xudong963:

## Summary

Many columns in tpcds contain a lot of NULL values.
> The percent of NULL values in each non-primary key column varies from 4 to 100 percent based on the column. Most columns have 4 percent NULL values. The important rec_end_date columns have 50 percent NULL values.

When they're used as join keys for inner join, we can generate `not null` filters and push down them to scan, then reduce I/O cost and network cost.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16722)


TCeason:
https://github.com/databendlabs/databend/pull/16739 fix ci err

src/query/sql/src/planner/optimizer/rule/rewrite/rule_filter_nulls.rs
```diff
@@ -0,0 +1,191 @@
...
+use crate::optimizer::Statistics;
+use crate::plans::Filter;
+use crate::plans::FunctionCall;
+use crate::plans::Join;
+use crate::plans::JoinType;
+use crate::plans::RelOp;
+use crate::plans::RelOperator;
+use crate::ScalarExpr;
+
+const NULL_THRESHOLD_RATIO: f64 = 0.02;
```
Dousir9, line 34: I think this threshold should be increased, perhaps 0.5 is a good value. The reasons are: (1) DataBlock will be re-generated during the filter, which will be expensive if there are many columns; (2) During the hash join probe phase, keys that are null will not probe the hash table.

src/query/sql/src/planner/optimizer/rule/rewrite/rule_filter_nulls.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let left_stat = RelExpr::with_s_expr(&left_child).derive_cardinality()?;
+        let right_stat = RelExpr::with_s_expr(&right_child).derive_cardinality()?;
+        let mut left_null_predicates = vec![];
+        let mut right_null_predicates = vec![];
+        for join_key in join.equi_conditions.iter() {
+            let left_key = &join_key.left;
+            let right_key = &join_key.right;
+            if single_plan(&left_child) {
+                if let Some(filter) = Self::should_filter_nulls(
```
Dousir9, line 116: In some cases we can eliminate `is_not_null`, for example the filter is `is_not_null(a) and a > 3`, `a > 3` will filter out null values.

src/query/sql/src/planner/optimizer/rule/rewrite/rule_filter_nulls.rs
xudong963: Sure, but the 0.5 seems a little large. Prunning even 10% of the parts will give good results, especially in distributed scenarios, and reduce network overhead. But I don't have a clear value for how much.

Dousir9: Since most of the cases are standalone mode and the probe and build of join will naturally filter out null values ​​(other operators will also do this), the is_not_null filter will cause `DataBlock::take` overhead, so it would be better to enable this optimization only in distributed mode. We can use 0.2, which is also used in FilterExecutor.

src/query/sql/src/planner/optimizer/rule/rewrite/rule_filter_nulls.rs
Dousir9: Will be implemented in future PR.

src/query/sql/src/planner/optimizer/rule/rewrite/rule_filter_nulls.rs
xudong963: Let me do something performance tests

Dousir9 (approved):

Issue #16626, opened 2024-10-16
Tracking issues: refactor partition prune into part of pipeline


dqhl76:
**Summary**

This issues track the refactor:

- [ ] [refactor: split the partition prune stage into different processors](https://github.com/databendlabs/databend/pull/16625)

dqhl76:
This should be solved. Closed.

PR #16841, opened 2024-11-14
refactor: integrate fuse table block pruning into pipeline

dqhl76:

## Summary

part of #16626

Split the block prune part and put into processor

## Tests

- [ ] Unit Test
- [x] Logic Test
- [x] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16841)


src/query/catalog/src/table.rs
```diff
@@ -218,6 +218,17 @@ pub trait Table: Sync + Send {
         )))
     }
 
+    fn build_prune_pipeline(
+        &self,
+        table_ctx: Arc<dyn TableContext>,
+        plan: &DataSourcePlan,
+        source_pipeline: &mut Pipeline,
+    ) -> Result<Option<Pipeline>> {
+        let (_, _, _) = (table_ctx, plan, source_pipeline);
```
zhang2014, line 227: Maybe return error is better?

dqhl76: I think it’s better to keep `Option` here because as we continue adding support for different engines, conditions like `table.get_table_info().meta.engine == *"FUSE" && ...` will inevitably grow. The disadvantage of using `Option` is that it doesn’t align with other operations. I’m open to changing it here if you think it’s important.

zhang2014 (approved):

PR #16912, opened 2024-11-22
fix: ignore case when matching function name

notauserx:

## Summary

Update logic to search by the lowercase of func_name. Uses unicode crate to facilitate searching function name ignoring case

fixes: #[16737](https://github.com/databendlabs/databend/issues/16737)

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16912)


notauserx:
Does it make sense to add ignore case search in is_builtin_function(func_name) as well?

src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -721,9 +721,7 @@ impl<'a> TypeChecker<'a> {
             } => {
                 let func_name = normalize_identifier(name, self.name_resolution_ctx).to_string();
                 let func_name = func_name.as_str();
```
andylokandy, line 723: what about handling case here?

```suggestion
                let func_name = func_name.to_lowercase();
```

notauserx: Good suggestion. Since sugar and built-in functions are matched ignoring case, how about we move the login inside the functions, otherwise each call site we have to convert to lower case before calling the functions

sundy-li:
Let's add a test in `name_hit.test` to cover the issue.

src/query/sql/src/planner/semantic/type_check.rs
andylokandy: Ok. I see. Then I'll suggest replacing `&str` in `all_sugar_functions` and `builtin_functions` to `https://crates.io/crates/unicase`

notauserx: I've used the Ascii type from the unicase crate, a case-insensitive wrapper for ASCII strings. Is this assumption correct? If not, I can switch to the UniCase type, which provides a case-insensitive wrapper for general strings.

notauserx:
> Let's add a test in `name_hit.test` to cover the issue.

Added a test

BohuTANG:
Hi @notauserx You can continue after fixing conflicts, we're nearly at the end :)

sundy-li (approved):

BohuTANG:
@notauserx Thanks for your contribution.

PR #16925, opened 2024-11-24
feat: disable specifying copy options when create stage.

youngsofun:

## Overview

copy options clauses are no longer allowed in `create stage` stmt

## motivation

1. (main) it is not clear and error prone to set copy options when create stage, which intended to serve as default value of copy related to this stage.
2. it is a burden to maintain copy options in meta.
3. it is bad to put both load and unload options in one struct.

## Impact

1.  the only breaking change is that copy options clauses are no longer allowed in `create stage` stmt, I guess no one had ever use this.
2.  old stages with copy options can still be used as before, because the options does not take effect at all (covered by copy options from copy stmt, which is parsed with struct default values).

 fixes:  #(https://github.com/databendlabs/databend/issues/16910)

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16925)


dantengsky (approved):

BohuTANG:
The summary is not cleary that what's changed in this PR?

youngsofun:
> The summary is not cleary that what's changed in this PR?

updated

PR #16948, opened 2024-11-26
feat(query): speed up fetching mysql data from dictionaries via batch processing.

Dragonliu2018:

## Summary

**Previous Impl**: For the Column type, loop through the Column and get the values from MySQL **one by one** via sqlx.

**Current Impl**: We change the sql query to fetch data **in batches** via sqlx.

```sql
-- pre sql
SELECT {value} FROM {table} WHERE {key} = ? LIMIT 1; 
-- cur sql
SELECT {key},{value} FROM {table} WHERE {key} in ({key1,key2,...});
```

The performance comparison is shown in the following table.

| **strategy \ key count \ elapsed time(seconds)** | 1w     | 10w      | 100w     |
| ------------------------------------------------ | ------ | -------- | -------- |
| **Previous Impl (one by one)**                   | 26.839 | 1122.486 | \        |
| **Current Impl (in batches)**                    | 3.256  | 100.100  | 7654.867 |

Refer to `02_0077_function_dict_get.test` for reproduction.
***

part of https://github.com/databendlabs/databend/issues/16550
## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16948)

### DIFF
--- a/src/query/service/src/pipelines/processors/transforms/transform_dictionary.rs
+++ b/src/query/service/src/pipelines/processors/transforms/transform_dictionary.rs
@@ -14,6 +14,7 @@
 
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::string::String;
 use std::sync::Arc;
 
@@ -32,6 +33,7 @@ use databend_common_expression::types::StringType;
 use databend_common_expression::types::ValueType;
 use databend_common_expression::with_integer_mapped_type;
 use databend_common_expression::BlockEntry;
+use databend_common_expression::Column;
 use databend_common_expression::ColumnBuilder;
 use databend_common_expression::DataBlock;
 use databend_common_expression::Scalar;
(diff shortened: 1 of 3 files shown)


src/query/service/src/pipelines/processors/transforms/transform_dictionary.rs
```diff
@@ -239,72 +320,168 @@ impl DictionaryOperator {
...
+        default_value: &Scalar,
+        pool: &MySqlPool,
+        sql: &String,
+    ) -> Result<Value<AnyType>> {
+        let key_cnt = column.len();
+        let mut all_keys = Vec::with_capacity(key_cnt);
+        let mut unique_keys = Vec::with_capacity(key_cnt);
+        let mut key_set = HashSet::with_capacity(key_cnt);
+        for item in column.iter() {
+            let key = self.format_key(item.clone());
```
b41sh, line 380: If the key is a NULL value, we can ignore it here.

src/query/service/src/pipelines/processors/transforms/transform_dictionary.rs
```diff
@@ -239,72 +320,168 @@ impl DictionaryOperator {
...
+            all_keys.push(key.clone());
+            if !key_set.contains(&key) {
+                unique_keys.push(item);
+                key_set.insert(key);
+            }
+        }
+        let new_sql = format!("{} {}", sql, self.format_keys(unique_keys));
```
b41sh, line 387: we can use `key_set` instead of `unique_keys`.

b41sh (approved):

PR #16955, opened 2024-11-27
fix(query): make memory engine as non-local table

### LINKED ISSUE #16941: Bug: `create table xxx engine = memory as xxx` lost columns under cluster
**Summary**

Tables in `select` are from ci tpcds dataset,
```sql
create or replace table test engine =memory as SELECT sr_customer_sk AS ctr_customer_sk,           sr_store_sk AS ctr_store_sk,           sum(sr_return_amt) AS ctr_total_return    FROM store_returns,       date_dim    WHERE sr_returned_date_sk = d_date_sk      AND d_year = 2000    GROUP BY sr_customer_sk,             sr_store_sk;
```

After creating `test`, it's ok to query the count, but will report error if query the specific data.
```sql
mysql> select count() from test;
+---------+
| count() |
+---------+
|     364 |
+---------+
1 row in set (0.06 sec)
Read 364 rows, 0.00 B in 0.012 sec., 30.57 thousand rows/sec., 0.00 B/sec.
```
```
mysql> select * from test;
ERROR 1105 (HY000): PanicError. Code: 1104, Text = index out of bounds: the len is 0 but the index is 0.
````

After some investigations, I found that some data blocks written to the memory table `test` don't have columns(zero col but non-zero rows).

### PR DESCRIPTION
**Pull request by sundy-li**

## Summary

fix(query): make memory engine as non-local table

fixes #16941 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16955)

### DIFF
--- a/src/query/storages/memory/src/memory_table.rs
+++ b/src/query/storages/memory/src/memory_table.rs
@@ -54,6 +54,7 @@ use databend_storages_common_table_meta::table::OPT_KEY_TEMP_PREFIX;
 use parking_lot::Mutex;
 use parking_lot::RwLock;
 
+use crate::memory_part::MemoryLazyPartInfo;
 use crate::memory_part::MemoryPartInfo;
 
 #[derive(Clone)]
@@ -160,6 +161,12 @@ impl Table for MemoryTable {
         &self.table_info
     }
 
+    /// MemoryTable could be distributed table, yet we only insert data in one node per query
+    /// Because commit_insert did not support distributed transaction
+    fn is_local(&self) -> bool {
+        false
+    }
+
     fn support_column_projection(&self) -> bool {
         true
     }
@@ -176,8 +183,7 @@ impl Table for MemoryTable {
         _dry_run: bool,
     ) -> Result<(PartStatistics, Partitions)> {
         let blocks = self.blocks.read();
-
-        let statistics = match push_downs {
+        let mut statistics = match push_downs {
             Some(push_downs) => {
                 let projection_filter: Box<dyn Fn(usize) -> bool> = match push_downs.projection {
                     Some(prj) => {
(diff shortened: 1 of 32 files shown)

### CODE REVIEW

File: src/query/storages/memory/src/memory_table.rs [hunk 1]
```diff
@@ -214,11 +220,29 @@ impl Table for MemoryTable {
...
+            statistics.partitions_total = statistics.partitions_total.max(cluster.nodes.len());
+            statistics.partitions_scanned = statistics.partitions_scanned.max(cluster.nodes.len());
+
+            let parts = (0..cluster.nodes.len())
+                .map(|node| MemoryLazyPartInfo::create(node as u64))
+                .collect();
+
+            return Ok((
+                statistics,
+                Partitions::create(PartitionsShuffleKind::Mod, parts),
```
zhang2014, line 236: maybe PartitionsShuffleKind::Broadcast  is better?

SkyFan2002 (approved):

src/query/storages/memory/src/memory_table.rs
```diff
@@ -214,12 +194,19 @@ impl Table for MemoryTable {
...
-        let parts = Self::generate_memory_parts(
-            0,
-            ctx.get_settings().get_max_threads()? as usize,
-            blocks.len(),
-        );
-        Ok((statistics, parts))
+        let cluster = ctx.get_cluster();
+        if !cluster.is_empty() {
+            statistics.read_bytes = statistics.read_bytes.max(cluster.nodes.len());
+            statistics.read_rows = statistics.read_rows.max(cluster.nodes.len());
```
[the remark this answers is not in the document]
sundy-li: it's not accurate stats, so I keep a none zero bytes.

tests/sqllogictests/suites/mode/cluster/distributed_copy_into_table.test
```diff
@@ -183,5 +183,24 @@ select count(*) from test_order;
 ----
 4000000
 
+statement ok
```
BohuTANG, line 186: 'Explain' is better to know the memory engine is distributed or not, can we add one?

sundy-li: explain will always have `distributed insert plan` even in standalone mode, it makes no difference now.

PR #16967, opened 2024-11-28
fix(sqlparser): fix order by followed by scalar

### LINKED ISSUE #16914: bug: order by literal with limit return error
### Search before asking

- [X] I had searched in the [issues](https://github.com/databendlabs/databend/issues) and found no similar issues.

### Version

v1.2.662

### What's Wrong?

order by literal with limit return error

we can ignore order by item if it is a scalar value

### How to Reproduce?

```sql
MySQL [(none)]> create table tt(a int, b int);
Query OK, 0 rows affected (0.187 sec)

MySQL [(none)]> insert into tt values(1, 2),(3, 4);
Query OK, 2 rows affected (0.209 sec)

MySQL [(none)]> select eq(tt.a, b) FROM tt ORDER BY 'a' LIMIT 89, 7;
ERROR 1105 (HY000): BadArguments. Code: 1006, Text = Unable to get field named "0". Valid fields: ["4"].
```

chagelo:

## Summary

close #16914

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/16967)

### DIFF
--- a/src/query/ast/src/parser/query.rs
+++ b/src/query/ast/src/parser/query.rs
@@ -316,6 +316,10 @@ impl<'a, I: Iterator<Item = WithSpan<'a, SetOperationElement>>> PrattParser<I>
                 if query.offset.is_some() {
                     return Err("ORDER BY must appear before OFFSET");
                 }
+                let order_by = order_by
+                    .into_iter()
+                    .filter(|x| matches!(x.expr, Expr::ColumnRef { .. }))
+                    .collect();
                 query.order_by = order_by;
             }
             SetOperationElement::Limit { limit } => {
--- a/src/query/ast/src/parser/query.rs
+++ b/src/query/ast/src/parser/query.rs
@@ -316,10 +316,6 @@ impl<'a, I: Iterator<Item = WithSpan<'a, SetOperationElement>>> PrattParser<I>
                 if query.offset.is_some() {
                     return Err("ORDER BY must appear before OFFSET");
                 }
-                let order_by = order_by
-                    .into_iter()
-                    .filter(|x| matches!(x.expr, Expr::ColumnRef { .. }))
-                    .collect();
                 query.order_by = order_by;
             }
             SetOperationElement::Limit { limit } => {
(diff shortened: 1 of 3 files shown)


src/query/ast/src/parser/query.rs
```diff
@@ -316,6 +316,10 @@ impl<'a, I: Iterator<Item = WithSpan<'a, SetOperationElement>>> PrattParser<I>
                 if query.offset.is_some() {
                     return Err("ORDER BY must appear before OFFSET");
                 }
+                let order_by = order_by
```
sundy-li, line 319: we need filter out `Expr::Constant`

b41sh: Do not modify the parser, this is a semantic problem. It's better to deal with in the binder phase, check whether the `rewrite_scalar` is `ScalarExpr::ConstantExpr`.
https://github.com/databendlabs/databend/blob/main/src/query/sql/src/planner/binder/sort.rs#L150

chagelo:
I want to know about how we generally add logs, like that I want to use `debug!` to print `rewrite_scalar` [here](https://github.com/databendlabs/databend/blob/25d615f1cfa7e2dadd0717e1b98f04f452f78a2f/src/query/sql/src/planner/binder/sort.rs#L150)(just add a line `debug!("rewrite_scalar: {:?}", rewirte_scalar);`, but don't see any output in terminal or ./.databend), or any other ways.

b41sh:
> I want to know about how we generally add logs, like that I want to use `debug!` to print `rewrite_scalar` [here](https://github.com/databendlabs/databend/blob/25d615f1cfa7e2dadd0717e1b98f04f452f78a2f/src/query/sql/src/planner/binder/sort.rs#L150)(just add a line `debug!("rewrite_scalar: {:?}", rewirte_scalar);`, but don't see any output in terminal or ./.databend, I'm gonna get farmilar with usage of crate log), or is there any other ways.

The log is in this directory `.databend/logs_1`, for example: `.databend/logs_1/databend-query-test_cluster.2024-12-02-11.0`

b41sh (approved):
@chagelo LGTM, thanks for your contribution.

PR #17008, opened 2024-12-06
refactor: set path of cookie.

youngsofun:

## Summary

we used to assume client only care about name and value (ignore domain and path).
this fix is to  avoid accumulation of cookies of diff paths if client enabled cookie mistakenly but not ignore path.
however, we can not set domain freely, because it will not work.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17008)

### DIFF
--- a/src/query/service/src/servers/http/middleware/session.rs
+++ b/src/query/service/src/servers/http/middleware/session.rs
@@ -322,6 +322,12 @@ pub struct HTTPSessionEndpoint<E> {
     pub auth_manager: Arc<AuthMgr>,
 }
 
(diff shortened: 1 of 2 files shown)


sundy-li (approved):

youngsofun:
1. add logic test code for this pr
2. the cookie jar for python is simplified.

PR #17013, opened 2024-12-07
refactor: mget db names and table names remove duplicate ids to avoid too many parameters

b41sh:

## Summary

refactor call `mget_database_names_by_ids` and `mget_table_names_by_ids` remove duplicate ids to avoid too many parameters.
fix reorder of the returned names, the order of the returned names are same as the parameter id.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17013)

### DIFF
--- a/src/query/service/src/table_functions/show_grants/show_grants_table.rs
+++ b/src/query/service/src/table_functions/show_grants/show_grants_table.rs
@@ -14,6 +14,7 @@
 
 use std::any::Any;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use databend_common_catalog::plan::DataSourcePlan;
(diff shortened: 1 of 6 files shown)


src/query/service/src/table_functions/show_grants/show_grants_table.rs
```diff
@@ -14,6 +14,7 @@
 
 use std::any::Any;
 use std::collections::HashMap;
+use std::collections::HashSet;
```
TCeason, line 17: Better use btree map, keep order

b41sh: HashMap is ok, we don't need keep id order

TCeason (approved):

PR #17051, opened 2024-12-15
feat(cluster): support system-managed cluster

zhang2014:

## Summary
feat(cluster): support system-managed cluster

Add new sql statement
```sql
SHOW WAREHOUSES

SHOW ONLINE NODES

CREATE WAREHOUSE <warehouse> [(ASSIGN <node_size> NODES [FROM <node_group>] [, ...])] WITH [warehouse_size = <warehouse_size>]

DROP WAREHOUSE <warehouse>

RENAME WAREHOUSE <warehouse> TO <new_warehouse>

RESUME WAREHOUSE <warehouse>

SUSPEND WAREHOUSE <warehouse>

INSPECT WAREHOUSE <warehouse>

ALTER WAREHOUSE <warehouse> ADD CLUSTER <cluster> [(ASSIGN <node_size> NODES [FROM <node_group>] [, ...])] WITH [cluster_size = <cluster_size>]

ALTER WAREHOUSE <warehouse> DROP CLUSTER <cluster>

ALTER WAREHOUSE <warehouse> RENAME CLUSTER <cluster> TO <new_cluster>

ALTER WAREHOUSE <warehouse> ASSIGN NODES ( ASSIGN <node_size> NODES [FROM <node_group>] FOR <cluster> [, ...] )

ALTER WAREHOUSE <warehouse> UNASSIGN NODES ( UNASSIGN <node_size> NODES [FROM <node_group>] FOR <cluster> [, ...] )
```

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17051)


drmingdrmer (commented):
The naming seems a bit inconsistent with the keys used in the meta-service and the corresponding variables. Aligning them might improve clarity.

> Reviewed 1 of 9 files at r1, 10 of 61 files at r2.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17051#-:-OEtE9GA7eFdLFi6iJ2F:b-5bt1wt)* status: 11 of 62 files reviewed, 8 unresolved discussions (waiting on @zhang2014)

___
*[`src/query/management/src/warehouse/warehouse_api.rs` line 22 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEscWTL3nP_ek2PIi6R:-OEscWTL3nP_ek2PIi6S:bqbrmxg) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_api.rs#L22)):*
> ```XML
> #[derive(serde::Serialize, serde::Deserialize, Clone, Eq, PartialEq, Debug)]
> pub enum SelectedNode {
>     Random(Option<String>),
> ```

What does the string in it mean? If the code isn't self-explanatory, it deserves a documentation comment.
___
*[`src/query/management/src/warehouse/warehouse_api.rs` line 29 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEseVhV3NPEu26tNrwQ:-OEseVhV3NPEu26tNrwR:b3b0l9t) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_api.rs#L29)):*
> ```XML
> #[derive(serde::Serialize, serde::Deserialize, Eq, PartialEq, Debug)]
> pub enum WarehouseInfo {
>     SelfManaged(String),
> ```

Need doc to explain the meaning of this string.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 87 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEtCipa3cf20b4Yhx6X:-OEtCipa3cf20b4Yhx6Y:b-jdm8dp) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L87)):*
> ```XML
>                 escape_for_key(tenant)?,
>             ),
>         })
> ```

I'm kind of confused of the field names. A bit inconsistent with the name and the comment. GPT gives a refined version of this part. make sense?

It would be clearer to provide an explanation of the key format used under these prefixes.



_Suggestion:_
```XML
        Ok(WarehouseMgr {
            metastore,
            lift_time,
            // Prefix for all online nodes of the tenant
            online_nodes_key_prefix: format!(
                "{}/{}/online_nodes",
                WAREHOUSE_API_KEY_PREFIX,
                escape_for_key(tenant)?
            ),
            // Prefix for all online computing clusters of the tenant
            online_clusters_key_prefix: format!(
                "{}/{}/online_clusters",
                WAREHOUSE_API_KEY_PREFIX,
                escape_for_key(tenant)?
            ),
            // Prefix for all warehouses of the tenant (must ensure compatibility across all versions)
            warehouses_key_prefix: format!(
                "{}/v1/{}",
                WAREHOUSE_META_KEY_PREFIX,
                escape_for_key(tenant)?,
            ),
        })
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 138 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt8agR0cdJ0ff8GwV-:-OEt8agR0cdJ0ff8GwV0:bi7mgos) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L138)):*
> ```XML
> 
>         let mut txn = TxnRequest::default();
>         let node_key = format!("{}/{}", self.node_key_prefix, escape_for_key(&node.id)?);
> ```

This should be replaced with `self.node_key(&node)`.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 148 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt9CYuD-Q-M6f0owOG:-OEt9CYuD-Q-M6f0owOH:bdoqz6p) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L148)):*
> ```XML
>         ));
> 
>         let warehouse_node_key = self.cluster_key(&node)?;
> ```

The variable name `wharehouse_node_key` is kind of inconsistent with the right side method `cluster_key()`
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 154 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt9RvNEyliQq-kteFh:-OEt9RvO67w8pofL5sJ8:b-2xjpmi) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L154)):*
> ```XML
>             self.warehouse_key_prefix,
>             escape_for_key(&node.warehouse_id)?
>         );
> ```

introduce another key-building method for this?



_Code quote:_
```XML
        let warehouse_info_key = format!(
            "{}/{}",
            self.warehouse_key_prefix,
            escape_for_key(&node.warehouse_id)?
        );
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 157 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt9quG3WYoHmE1FE2N:-OEt9quG3WYoHmE1FE2O:bca7lpp) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L157)):*
> ```XML
> 
>         node.cluster_id = String::new();
>         node.warehouse_id = String::new();
> ```

This feels a bit odd: at the beginning of this method, it asserts that these two fields are non-empty, but later they are cleared. 

Would it make sense to pass `cluster_id` and `warehouse_id` as separate arguments to this method instead?



_Code quote:_
```XML
        node.cluster_id = String::new();
        node.warehouse_id = String::new();
```
___
*[`src/meta/types/src/cluster.rs` line 100 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEs_UzSBJJQWA8pX3Qy:-OEs_UzSBJJQWA8pX3Qz:bmvth0c) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/meta/types/src/cluster.rs#L100)):*
> ```XML
>     pub warehouse_id: String,
> 
>     pub runtime_resource_group: Option<String>,
> ```

This field does not need to specify `#[serde(skip_serializing_if = "Option::is_none")]`?

Encoding a `null` value may break older version query?

And for the three new fields, is `#[serde(skip)]` needed to allow it to decode an older version NodeInfo json that does not have these fields?

zhang2014:
*[`src/meta/types/src/cluster.rs` line 100 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEs_UzSBJJQWA8pX3Qy:-OFHFnkV1PLHpeffxQWX:bz0hv1f) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/meta/types/src/cluster.rs#L100)):*

This field does not need to specify `#[serde(skip_serializing_if = "Option::is_none")]`?

Encoding a `null` value may break older version query?

And for the three new fields, is `#[serde(skip)]` needed to allow it to decode an older version NodeInfo json that does not have these fields?
</blockquote></details>

We do not need to consider compatibility issues with older versions. The prefix has been modified, and now the new version will be written under the new prefix.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_api.rs` line 22 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEscWTL3nP_ek2PIi6R:-OFHG4cZCFTJtqJsLRl5:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_api.rs#L22)):*

What does the string in it mean? If the code isn't self-explanatory, it deserves a documentation comment.
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_api.rs` line 29 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEseVhV3NPEu26tNrwQ:-OFHG6ye34ZJMeMsRcIl:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_api.rs#L29)):*

Need doc to explain the meaning of this string.
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 87 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEtCipa3cf20b4Yhx6X:-OFHG9acAhkmpvcHakfZ:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L87)):*

I'm kind of confused of the field names. A bit inconsistent with the name and the comment. GPT gives a refined version of this part. make sense?

It would be clearer to provide an explanation of the key format used under these prefixes.
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 138 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt8agR0cdJ0ff8GwV-:-OFHGCDWD-ZU_ADUsSqL:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L138)):*

This should be replaced with `self.node_key(&node)`.
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 148 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt9CYuD-Q-M6f0owOG:-OFHGE6ACTWXDwUy0bUQ:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L148)):*

The variable name `wharehouse_node_key` is kind of inconsistent with the right side method `cluster_key()`
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 154 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt9RvNEyliQq-kteFh:-OFHGGI-AL96H7mPDhZL:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L154)):*

introduce another key-building method for this?
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 157 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt9quG3WYoHmE1FE2N:-OFHGI1wBye5NtPe4CPa:b7klf2c) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L157)):*

This feels a bit odd: at the beginning of this method, it asserts that these two fields are non-empty, but later they are cleared. 

Would it make sense to pass `cluster_id` and `warehouse_id` as separate arguments to this method instead?
</blockquote></details>

Done.  Extracted the `unload_warehouse_info` function and added comments.

// Unload the warehouse and cluster from the node.  
// 1. Used when a node is removed from the cluster.  
// 2. For cluster\_node\_key: node\_info, since its warehouse and cluster are already encoded in the key, we do not need to write the warehouse and cluster into its value again.

zhang2014:
system-managed warehouse?

zhang2014:
@drmingdrmer Thanks very helpful comments. I have resolved the issues. Please review again.

drmingdrmer (changes requested):
> Reviewed 3 of 59 files at r4, 2 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17051)* status: 9 of 77 files reviewed, 10 unresolved discussions (waiting on @zhang2014)

___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 157 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEt9quG3WYoHmE1FE2N:-OFHYPWoBu92nkegcU9A:b-aryxbw) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_mgr.rs#L157)):*
<details><summary><i>Previously, zhang2014 (Winter Zhang) wrote…</i></summary><blockquote>

Done.  Extracted the `unload_warehouse_info` function and added comments.

// Unload the warehouse and cluster from the node.  
// 1. Used when a node is removed from the cluster.  
// 2. For cluster\_node\_key: node\_info, since its warehouse and cluster are already encoded in the key, we do not need to write the warehouse and cluster into its value again.
</blockquote></details>

Have you considered removing `cluster_id` and `warehouse_id` from `NodeInfo`, since they are not actually stored but just used to generate some keys.
___
*[`src/query/management/src/warehouse/warehouse_api.rs` line 29 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEseVhV3NPEu26tNrwQ:-OFHMEuGARHHSZRsZWCn:be8uo0s) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_api.rs#L29)):*
<details><summary><i>Previously, zhang2014 (Winter Zhang) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

In such case, making it `SelfManaged{ cluster_id: String }` would be more self-explanatory.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 75 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHNuP4CihWgggLJXnn:-OFHNuP4CihWgggLJXno:by2u9kw) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L75)):*
> ```XML
>             // Prefix for all online computing clusters of the tenant
>             meta_key_prefix: format!(
>                 "{}/{}/online_clusters_v2",
> ```

The meaning of the `meta-key` is still unclear. what does `meta` mean here?
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 179 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHRdrr1PjWffu1xx7r:-OFHRdrs-H91kjmdOzvW:bmk7epg) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L179)):*
> ```XML
>             serde_json::to_vec(&warehouse_info)?,
>             Some(self.lift_time),
>         ));
> ```

The comment says that warehouse info is written conditionally. But the code always write the warehouse info unconditionally. 

The comment should be `// self-managed warehouse requires to persist warehouse info` or something else?



_Code quote:_
```XML
        // upsert warehouse info if self-managed.

        txn.if_then.push(TxnOp::put_with_ttl(
            warehouse_info_key.clone(),
            serde_json::to_vec(&warehouse_info)?,
            Some(self.lift_time),
        ));
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 186 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHV2WK9eiS2_1RMbXQ:-OFHV2WK9eiS2_1RMbXR:b-a0ziqv) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L186)):*
> ```XML
>         if seq != MatchSeq::Exact(0) {
>             return Ok(self.metastore.transaction(txn).await?);
>         }
> ```

It's a little bit  weird: why not retry if `seq != MatchSeq::Exact(0)`?

This function's retry behavior is inconsistent. When `seq` equals `MatchSeq::Exact(0)`, it retries until the update succeeds. However, for any other `seq` value, it makes no retry attempts. If retry behavior should be optional, it should be controlled by an explicit parameter rather than being tied to the `seq` value.



_Code quote:_
```XML
        if seq != MatchSeq::Exact(0) {
            return Ok(self.metastore.transaction(txn).await?);
        }
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 200 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHWytT1EBg0t0kl8mm:-OFHWytT1EBg0t0kl8mn:bf6917z) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L200)):*
> ```XML
>             warehouse_txn
>                 .else_then
>                 .push(TxnOp::get(warehouse_info_key.clone()));
> ```

It looks like this `else` operation of getting warehouse can be added before entering the loop?



_Code quote:_
```XML
            warehouse_txn
                .else_then
                .push(TxnOp::get(warehouse_info_key.clone()));
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 236 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHXAtY3ijWruRMv8s_:-OFHXAtY3ijWruRMv8sa:bskdw1) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L236)):*
> ```XML
>                 }
>                 response => Ok(response),
>             };
> ```

This piece of logic nested deeply and quite hard to understand. I asked GPT to rewrite it to reduce the indent.

Does it reflect the logic correctly?

```rust
let mut response = self.metastore.transaction(warehouse_txn).await?;
if response.success {
    return Ok(response);
}

let resp = match response.responses.pop().and_then(|x| x.response) {
    Some(Response::Get(data)) => data,
    _ => return Ok(response),
};

let value = match resp.value {
    None => return Ok(response),
    Some(value) if value.seq == 0 => return Ok(response),
    Some(value) => value,
};

let warehouse_info = serde_json::from_slice(&value.data)?;

match warehouse_info {
    WarehouseInfo::SystemManaged(_) => {
        return Err(ErrorCode::WarehouseAlreadyExists(
            "Already exists same name system-managed warehouse.",
        ));
    }
    WarehouseInfo::SelfManaged(_) => {
        // Check if node already exists
        if let Some(TxnOpResponse {
            response: Some(Response::Get(TxnGetResponse { value: Some(value), .. })),
        }) = response.responses.first()
        {
            if value.seq != 0 {
                return Ok(response);
            }
        }

        log::info!(
            "Self-managed warehouse has already been created by other nodes; attempt to join it. Retry count: {}",
            retry_count
        );
        retry_count += 1;
        exact_seq = value.seq;
        continue;
    }
}
```



_Code quote:_
```XML
            return match self.metastore.transaction(warehouse_txn).await? {
                mut response if !response.success => {
                    return match response.responses.pop().and_then(|x| x.response) {
                        Some(Response::Get(data)) => match data.value {
                            None => Ok(response),
                            Some(value) if value.seq == 0 => Ok(response),
                            Some(value) => match serde_json::from_slice(&value.data)? {
                                WarehouseInfo::SystemManaged(_) => {
                                    Err(ErrorCode::WarehouseAlreadyExists(
                                        "Already exists same name system-managed warehouse.",
                                    ))
                                }
                                WarehouseInfo::SelfManaged(_) => match response.responses.first() {
                                    // already exists node.
                                    Some(TxnOpResponse {
                                        response:
                                            Some(Response::Get(TxnGetResponse {
                                                value: Some(value),
                                                ..
                                            })),
                                    }) if value.seq != 0 => Ok(response),
                                    _ => {
                                        log::info!("Self-managed warehouse has already been created by other nodes; attempt to join it. Retry count: {}", retry_count);
                                        retry_count += 1;
                                        exact_seq = value.seq;
                                        continue;
                                    }
                                },
                            },
                        },
                        _ => Ok(response),
                    };
                }
                response => Ok(response),
            };
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 321 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHmkXn5FyRpvDS-7-y:-OFHmkXn5FyRpvDS-7-z:undefined) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L321)):*
> ```XML
>                 // std::mem::swap(&mut node_info.warehouse_id, &mut warehouse_id);
>                 Err(err)
>             }
> ```





_Suggestion:_
```XML
        match upsert_node.await? {
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 327 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHmg4SCptmkTykFaKq:-OFHmg4SCptmkTykFaKr:b2dlami) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L327)):*
> ```XML
>                 // std::mem::swap(&mut node_info.warehouse_id, &mut warehouse_id);
>                 Ok(seq)
>             }
> ```

No need to rollback right?



_Code quote:_
```XML
        match upsert_node.await {
            Err(err) => {
                // rollback
                // std::mem::swap(&mut node_info.cluster_id, &mut cluster_id);
                // std::mem::swap(&mut node_info.warehouse_id, &mut warehouse_id);
                Err(err)
            }
            Ok(response) if !response.success => {
                // rollback
                // std::mem::swap(&mut node_info.cluster_id, &mut cluster_id);
                // std::mem::swap(&mut node_info.warehouse_id, &mut warehouse_id);
                Ok(seq)
            }
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 444 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHoiP39rchd58FC28N:-OFHoiP39rchd58FC28O:b-7bbf1z) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L444)):*
> ```XML
>                     continue 'retry;
>                 }
>             }
> ```

Too many indents...hard to understand...

A rewritten version of this block with less indent looks like:

```rust
let response = self.metastore.transaction(after_txn).await?;
if !response.success {
    continue 'retry;
}

let mut consistent_nodes = Vec::with_capacity(response.responses.len());

for (idx, response) in response.responses.into_iter().enumerate() {
    let get_response = match response.response {
        Some(Response::Get(response)) => response,
        _ => continue 'retry,
    };

    let value = match get_response.value {
        Some(value) => value,
        _ => continue 'retry,
    };

    let node_info = serde_json::from_slice::<NodeInfo>(&value.data)?;
    
    assert_eq!(node_info.warehouse_id, id);
    assert!(!node_info.cluster_id.is_empty());

    consistent_nodes.push(ConsistentNodeInfo {
        node_seq: value.seq,
        cluster_seq: cluster_node_seq[idx],
        node_info,
    });
}

if consistent_nodes.len() != cluster_node_seq.len() {
    continue 'retry;
}

return Ok(ConsistentWarehouseInfo {
    info_seq: before_info.seq,
    warehouse_info: serde_json::from_slice(&before_info.data)?,
    consistent_nodes,
});
```



_Code quote:_
```XML
            match self.metastore.transaction(after_txn).await? {
                response if response.success => {
                    let mut consistent_nodes = Vec::with_capacity(response.responses.len());
                    for (idx, response) in response.responses.into_iter().enumerate() {
                        match response.response {
                            // TODO: maybe ignore none(not need retry)
                            Some(Response::Get(response)) => match response.value {
                                Some(value) => {
                                    let node_info =
                                        serde_json::from_slice::<NodeInfo>(&value.data)?;

                                    assert_eq!(node_info.warehouse_id, id);
                                    assert!(!node_info.cluster_id.is_empty());

                                    consistent_nodes.push(ConsistentNodeInfo {
                                        node_seq: value.seq,
                                        cluster_seq: cluster_node_seq[idx],
                                        node_info,
                                    });
                                }
                                _ => {
                                    continue 'retry;
                                }
                            },
                            _ => {
                                continue 'retry;
                            }
                        }
                    }

                    if consistent_nodes.len() == cluster_node_seq.len() {
                        return Ok(ConsistentWarehouseInfo {
                            info_seq: before_info.seq,
                            warehouse_info: serde_json::from_slice(&before_info.data)?,
                            consistent_nodes,
                        });
                    }
                }
                _ => {
                    continue 'retry;
                }
            }
```

flaneur2020:
do we have a config to disable these WAREHOUSE statements? we have a cloud deployment with the warehouse concept, but before we made the integration with the WAREHOUSE statements with the underlying kubernetes primitives, it might not bad to allow us disable the WAREHOUSE statements to avoid concept conflicts on corner cases.

zhang2014:
> do we have a config to disable these WAREHOUSE statements? we have a cloud deployment with the warehouse concept, but before we made the integration with the WAREHOUSE statements with the underlying kubernetes primitives, it might not bad to allow us disable the WAREHOUSE statements to avoid concept conflicts on possible corner cases.

By default(without the resources_management configuration), we return an unimplemented error. Is this enough?

flaneur2020:
> By default(without the resources_management configuration), we return an unimplemented error. Is this enough?

SGTM :+1:

zhang2014:
*[`src/query/management/src/warehouse/warehouse_api.rs` line 29 at r2](https://reviewable.io/reviews/databendlabs/databend/17051#-OEseVhV3NPEu26tNrwQ:-OFtKjQd7EyckqLuKJTJ:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/da6155e669f2e52921b231bab3419472a67db5a0/src/query/management/src/warehouse/warehouse_api.rs#L29)):*

In such case, making it `SelfManaged{ cluster_id: String }` would be more self-explanatory.
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 75 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHNuP4CihWgggLJXnn:-OFtKmep6Obs5nX_tryM:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L75)):*

The meaning of the `meta-key` is still unclear. what does `meta` mean here?
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 179 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHRdrr1PjWffu1xx7r:-OFtLIhj1g-O9bM2pwVX:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L179)):*

The comment says that warehouse info is written conditionally. But the code always write the warehouse info unconditionally. 

The comment should be `// self-managed warehouse requires to persist warehouse info` or something else?
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 186 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHV2WK9eiS2_1RMbXQ:-OFtLLQKCUUokGiX6hM1:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L186)):*

It's a little bit  weird: why not retry if `seq != MatchSeq::Exact(0)`?

This function's retry behavior is inconsistent. When `seq` equals `MatchSeq::Exact(0)`, it retries until the update succeeds. However, for any other `seq` value, it makes no retry attempts. If retry behavior should be optional, it should be controlled by an explicit parameter rather than being tied to the `seq` value.
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 200 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHWytT1EBg0t0kl8mm:-OFtLObSEwh9f05s117w:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L200)):*

It looks like this `else` operation of getting warehouse can be added before entering the loop?
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 236 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHXAtY3ijWruRMv8s_:-OFtLQmn0wbYExZSO1Tb:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L236)):*

This piece of logic nested deeply and quite hard to understand. I asked GPT to rewrite it to reduce the indent.

Does it reflect the logic correctly?

```rust
let mut response = self.metastore.transaction(warehouse_txn).await?;
if response.success {
    return Ok(response);
}

let resp = match response.responses.pop().and_then(|x| x.response) {
    Some(Response::Get(data)) => data,
    _ => return Ok(response),
};

let value = match resp.value {
    None => return Ok(response),
    Some(value) if value.seq == 0 => return Ok(response),
    Some(value) => value,
};

let warehouse_info = serde_json::from_slice(&value.data)?;

match warehouse_info {
    WarehouseInfo::SystemManaged(_) => {
        return Err(ErrorCode::WarehouseAlreadyExists(
            "Already exists same name system-managed warehouse.",
        ));
    }
    WarehouseInfo::SelfManaged(_) => {
        // Check if node already exists
        if let Some(TxnOpResponse {
            response: Some(Response::Get(TxnGetResponse { value: Some(value), .. })),
        }) = response.responses.first()
        {
            if value.seq != 0 {
                return Ok(response);
            }
        }

        log::info!(
            "Self-managed warehouse has already been created by other nodes; attempt to join it. Retry count: {}",
            retry_count
        );
        retry_count += 1;
        exact_seq = value.seq;
        continue;
    }
}
```
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 327 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHmg4SCptmkTykFaKq:-OFtLToYCMbctaAgoDjz:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L327)):*

No need to rollback right?
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 321 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHmkXn5FyRpvDS-7-y:-OFtLVnb3bW6Iwil23Cd:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L321)):*
> ```XML
>                 // std::mem::swap(&mut node_info.warehouse_id, &mut warehouse_id);
>                 Err(err)
>             }
> ```

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 444 at r5](https://reviewable.io/reviews/databendlabs/databend/17051#-OFHoiP39rchd58FC28N:-OFtL_qCAul-VqNU1DbS:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bc4b2ea819c232642634efb5f01c7fd0e156c4a/src/query/management/src/warehouse/warehouse_mgr.rs#L444)):*

Too many indents...hard to understand...

A rewritten version of this block with less indent looks like:

```rust
let response = self.metastore.transaction(after_txn).await?;
if !response.success {
    continue 'retry;
}

let mut consistent_nodes = Vec::with_capacity(response.responses.len());

for (idx, response) in response.responses.into_iter().enumerate() {
    let get_response = match response.response {
        Some(Response::Get(response)) => response,
        _ => continue 'retry,
    };

    let value = match get_response.value {
        Some(value) => value,
        _ => continue 'retry,
    };

    let node_info = serde_json::from_slice::<NodeInfo>(&value.data)?;
    
    assert_eq!(node_info.warehouse_id, id);
    assert!(!node_info.cluster_id.is_empty());

    consistent_nodes.push(ConsistentNodeInfo {
        node_seq: value.seq,
        cluster_seq: cluster_node_seq[idx],
        node_info,
    });
}

if consistent_nodes.len() != cluster_node_seq.len() {
    continue 'retry;
}

return Ok(ConsistentWarehouseInfo {
    info_seq: before_info.seq,
    warehouse_info: serde_json::from_slice(&before_info.data)?,
    consistent_nodes,
});
```
</blockquote></details>

Done.

drmingdrmer (approved):
LGTM. Some nits as following:

> Reviewed 13 of 61 files at r2, 37 of 59 files at r4, 2 of 4 files at r8, 2 of 3 files at r9, 19 of 21 files at r10, 1 of 4 files at r11, 2 of 4 files at r12, 2 of 4 files at r13, 2 of 2 files at r14, 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17051#-:-OFuJLCL50Hf1Vtod2v-:b173j9u)* status: all files reviewed, 10 unresolved discussions (waiting on @flaneur2020 and @zhang2014)

___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 186 at r10](https://reviewable.io/reviews/databendlabs/databend/17051#-OFoL7tf5W2kKGUP8bEy:-OFoL7tf5W2kKGUP8bEz:b-vnoocf) ([raw file](https://github.com/databendlabs/databend/blob/8bd09f8d97fffcf96bc2c05b50ea4363fa22fd9e/src/query/management/src/warehouse/warehouse_mgr.rs#L186)):*
> ```XML
>         node.cluster_id = String::new();
>         node.warehouse_id = String::new();
>     }
> ```

`&self` is not used in this method. It looks like `unload_warehouse_info()` should be a method of `NodeInfo`.



_Code quote:_
```XML
    // Unload the warehouse and cluster from the node.
    // 1. Used when a node is removed from the cluster.
    // 2. For cluster_node_key: node_info, since its warehouse and cluster are already encoded in the key, we do not need to write the warehouse and cluster into its value again.
    fn unload_warehouse_info(&self, node: &mut NodeInfo) {
        node.cluster_id = String::new();
        node.warehouse_id = String::new();
    }
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 371 at r10](https://reviewable.io/reviews/databendlabs/databend/17051#-OFoNxBV4EyXCjmjw9MB:-OFoNxBV4EyXCjmjw9MC:bp4r88n) ([raw file](https://github.com/databendlabs/databend/blob/8bd09f8d97fffcf96bc2c05b50ea4363fa22fd9e/src/query/management/src/warehouse/warehouse_mgr.rs#L371)):*
> ```XML
> 
>         // If the warehouse has already been assigned.
>         if !node.cluster_id.is_empty() && !node.warehouse_id.is_empty() {
> ```

Seems there should be a method `NodeInfo::is_assigned_to_warehouse()` that check these two fields.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 421 at r10](https://reviewable.io/reviews/databendlabs/databend/17051#-OFoRL9yEz6EwuLAtbhz:-OFoRL9yEz6EwuLAtbi-:b-y56x8m) ([raw file](https://github.com/databendlabs/databend/blob/8bd09f8d97fffcf96bc2c05b50ea4363fa22fd9e/src/query/management/src/warehouse/warehouse_mgr.rs#L421)):*
> ```XML
> 
>         let reply = self
>             .heartbeat_system_managed(leave_node.clone(), seq)
> ```

Does this txn need to remove the node from the `cluster_node_key` prefix `*/<warehouse>/<cluster_id>/`? It looks like `heartbeat_system_managed()` does not remove such key.

Reusing `heartbeat_system_managed()` here makes it unclear why `heartbeat` and `leave` share the same logic. It would be clearer to have both `heartbeat` and `leave` call a separate method, such as `try_update_system_managed_node()`.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 463 at r10](https://reviewable.io/reviews/databendlabs/databend/17051#-OFoSqpJBX5rwwaW-fGE:-OFoSqpJBX5rwwaW-fGF:b-mjjtfj) ([raw file](https://github.com/databendlabs/databend/blob/8bd09f8d97fffcf96bc2c05b50ea4363fa22fd9e/src/query/management/src/warehouse/warehouse_mgr.rs#L463)):*
> ```XML
>     }
> 
>     async fn resolve_conflicts(&self, reply: TxnReply, node: &mut NodeInfo) -> Result<u64> {
> ```

This method should not receive a `TxnReply` as argument. `TxnReply` is designed for meta-store API and this method might be affected if `TxnReply`changes in the future.

This method just need a seq number and a `NodeInfo` that is retrieved from meta-store.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 679 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFthVXT5UTRq-FeBDOv:-OFthVXT5UTRq-FeBDOw:b-l0sqy) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L679)):*
> ```XML
>                     id
>                 )));
>             };
> ```

It would be better to return `Result<Option<WarehouseSnapshot>>`. An `ErrorCode` could represent meta-store network/internal errors or similar issues, which are distinct from a "warehouse not found" case. Handling these separately allows the caller to address errors more appropriately.



_Code quote:_
```XML
            let Some(before_info) = self.metastore.get_kv(&warehouse_info_key).await? else {
                return Err(ErrorCode::UnknownWarehouse(format!(
                    "Unknown warehouse or self managed warehouse {:?}",
                    id
                )));
            };
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 697 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtl4Nc7gFcQHtnp3kL:-OFtl4Nc7gFcQHtnp3kM:undefined) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L697)):*
> ```XML
> 
>                 let node_key = format!("{}/{}", self.node_key_prefix, node);
>                 after_txn.if_then.push(TxnOp::get(node_key));
> ```





_Suggestion:_
```XML
/// Fetch the seq of node under `node_prefix/**`
after_txn.if_then.push(TxnOp::get(node_key));
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 698 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtmw4P2m6ASa8Rvgm0:-OFtmw4P2m6ASa8Rvgm1:undefined) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L698)):*
> ```XML
>                 let node_key = format!("{}/{}", self.node_key_prefix, node);
>                 after_txn.if_then.push(TxnOp::get(node_key));
>                 cluster_node_seq.push(value.seq);
> ```





_Suggestion:_
```XML
// While we always assert the seq of `node_key_prefix/**` nodes during updates to ensure a consistent snapshot view,
// the seq of `cluster_node_key_prefix/**` nodes is also asserted as a defensive measure.
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 2033 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtftSq86XtRShy4350:-OFtftSq86XtRShy4351:b1z2edz) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L2033)):*
> ```XML
>             escape_for_key(warehouse)?,
>             escape_for_key(cluster)?
>         );
> ```

Key / prefix building should be extracted into a method for reuse.

And BTW `escape_for_key()` should not return an error at all.



_Code quote:_
```XML
        let cluster_prefix = format!(
            "{}/{}/{}/",
            self.cluster_node_key_prefix,
            escape_for_key(warehouse)?,
            escape_for_key(cluster)?
        );
```
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 2067 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtfJovBT9TB4hdBJCo:-OFtfJow8KblDIJuNTVw:undefined) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L2067)):*
> ```XML
>     }
> 
>     async fn discover(&self, node_id: &str) -> Result<Vec<NodeInfo>> {
> ```





_Suggestion:_
```XML
/// Discovers and returns all the nodes that are in the same cluster with the given node, including the provided one.
///
/// If the given node is not in a cluster yet, just return itself.
async fn discover(&self, node_id: &str) -> Result<Vec<NodeInfo>> {
```
___
*[`src/query/management/src/warehouse/warehouse_api.rs` line 67 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtMRxS4L3DMP4alNnl:-OFtMRxS4L3DMP4alNnm:b4ct8tf) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_api.rs#L67)):*
> ```XML
> pub trait WarehouseApi: Sync + Send {
>     /// Start a new node.
>     async fn start_node(&self, node: NodeInfo) -> Result<(u64, NodeInfo)>;
> ```

Use `SeqV<NodeInfo>` to represent a combination of `seq` and `T`. It would be more clear about the purpose of the `u64`.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 371 at r10](https://reviewable.io/reviews/databendlabs/databend/17051#-OFoNxBV4EyXCjmjw9MB:-OFuL2RV6vvxlFqWjXHs:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bd09f8d97fffcf96bc2c05b50ea4363fa22fd9e/src/query/management/src/warehouse/warehouse_mgr.rs#L371)):*

Seems there should be a method `NodeInfo::is_assigned_to_warehouse()` that check these two fields.
</blockquote></details>

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 2067 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtfJovBT9TB4hdBJCo:-OFuNkVUEypfB6iumZro:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L2067)):*
> ```XML
>     }
> 
>     async fn discover(&self, node_id: &str) -> Result<Vec<NodeInfo>> {
> ```

Done.

zhang2014:
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 2033 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtftSq86XtRShy4350:-OFuNjpf3FWUpWXG24Tz:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L2033)):*

Key / prefix building should be extracted into a method for reuse.

And BTW `escape_for_key()` should not return an error at all.
</blockquote></details>

Done.

zhang2014 (commented):
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17051)* status: 78 of 83 files reviewed, 10 unresolved discussions (waiting on @drmingdrmer and @flaneur2020)

___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 186 at r10](https://reviewable.io/reviews/databendlabs/databend/17051#-OFoL7tf5W2kKGUP8bEy:-OFuRxAk3kT6M0AUEhix:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bd09f8d97fffcf96bc2c05b50ea4363fa22fd9e/src/query/management/src/warehouse/warehouse_mgr.rs#L186)):*

`&self` is not used in this method. It looks like `unload_warehouse_info()` should be a method of `NodeInfo`.
</blockquote></details>

Done.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 421 at r10](https://reviewable.io/reviews/databendlabs/databend/17051#-OFoRL9yEz6EwuLAtbhz:-OFuLbh_AKurjtdz7Y9H:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/8bd09f8d97fffcf96bc2c05b50ea4363fa22fd9e/src/query/management/src/warehouse/warehouse_mgr.rs#L421)):*

Does this txn need to remove the node from the `cluster_node_key` prefix `*/<warehouse>/<cluster_id>/`? It looks like `heartbeat_system_managed()` does not remove such key.

Reusing `heartbeat_system_managed()` here makes it unclear why `heartbeat` and `leave` share the same logic. It would be clearer to have both `heartbeat` and `leave` call a separate method, such as `try_update_system_managed_node()`.
</blockquote></details>

Done.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 679 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFthVXT5UTRq-FeBDOv:-OFuSiCHBff6y2BzoV6f:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L679)):*

It would be better to return `Result<Option<WarehouseSnapshot>>`. An `ErrorCode` could represent meta-store network/internal errors or similar issues, which are distinct from a "warehouse not found" case. Handling these separately allows the caller to address errors more appropriately.
</blockquote></details>

Done.
___
*[`src/query/management/src/warehouse/warehouse_api.rs` line 67 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtMRxS4L3DMP4alNnl:-OFuP3B1ETGK7GTOF8XU:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_api.rs#L67)):*

Use `SeqV<NodeInfo>` to represent a combination of `seq` and `T`. It would be more clear about the purpose of the `u64`.
</blockquote></details>

Done.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 697 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtl4Nc7gFcQHtnp3kL:-OFuSRn7EqtSO2NVBn-u:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L697)):*
> ```XML
> 
>                 let node_key = format!("{}/{}", self.node_key_prefix, node);
>                 after_txn.if_then.push(TxnOp::get(node_key));
> ```

Done.
___
*[`src/query/management/src/warehouse/warehouse_mgr.rs` line 698 at r12](https://reviewable.io/reviews/databendlabs/databend/17051#-OFtmw4P2m6ASa8Rvgm0:-OFuSkm366VRAXtlmfPp:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/5da3a5e74757317f2e3b3302b77e755ccd037b47/src/query/management/src/warehouse/warehouse_mgr.rs#L698)):*
> ```XML
>                 let node_key = format!("{}/{}", self.node_key_prefix, node);
>                 after_txn.if_then.push(TxnOp::get(node_key));
>                 cluster_node_seq.push(value.seq);
> ```

Done.

PR #17089, opened 2024-12-19
feat(query): refactor vacuum temp files

sundy-li:

## Summary

feat(query): refactor vacuum temp files

-  `VacuumTempOptions::QueryHook` only vacuum temp files after the query finish
   1. It will vacuum files by hook limit
   2. If there are files remained, store ${query_id}_${index}.list` mark that there are remaining files to be cleaned in the future. 

-  `VacuumTempOptions::VacuumCommand` is called via command `VACUUM TEMPORARY FILES`
    It will list the spill directory
   1. If it scans the meta file, it will clean the data by meta
   3. If it scans the directory, it will get the meta file firstly. If found then delete by meta else delete by list
   4. If it scans other files, delete by iterator.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17089)


src/query/ee/src/storages/fuse/operations/vacuum_temporary_files.rs
```diff
@@ -34,229 +36,228 @@ const DEFAULT_RETAIN_DURATION: Duration = Duration::from_secs(60 * 60 * 24 * 3);
...
-                        };
-
-                    vacuum_finished_query(
-                        &abort_checker,
-                        start_time,
-                        &mut removed_temp_files,
-                        &mut total_cleaned_size,
-                        &mut batch_size,
-                        &de,
+                let meta = operator.stat(de.path()).await;
```
Xuanwo, line 90: We can use `de.metadata().mode()` to check if it's a dir.

src/query/service/src/sessions/query_ctx.rs
```diff
@@ -387,6 +426,42 @@ impl QueryContext {
...
+            })
+            .cloned()
+            .collect::<Vec<_>>();
+
+        w.clear();
+
+        if !remote_spill_files.is_empty() {
+            let location_prefix = self.query_tenant_spill_prefix();
+            let node_idx = self.get_cluster().ordered_index();
+            let meta_path = format!("{}/{}_{}.meta", location_prefix, self.get_id(), node_idx);
```
Xuanwo, line 447: the `meta`'s name is a bit confuse, how about using `spill_idx` or something?

src/query/ee/src/storages/fuse/operations/vacuum_temporary_files.rs
sundy-li: we need to check `last_modified`, so `stat` op can't be avoided.

Xuanwo: > we need to check `last_modified`, so `stat` op can't be avoided.

`de.metadata()` will also have `last_modified` for most object storage services.

src/query/service/src/spillers/serialize.rs
```diff
@@ -158,7 +160,7 @@ fn bare_blocks_to_parquet<W: Write + Send>(
     let table_schema = infer_table_schema(&data_schema)?;
 
     let props = WriterProperties::builder()
-        .set_compression(Compression::LZ4_RAW)
+        .set_compression(Compression::SNAPPY)
```
forsaken628, line 163: How much of a performance difference is there between snappy and lz4?

forsaken628 (approved):

src/query/service/src/spillers/serialize.rs
sundy-li: It's change by mistake, thanks for the tips.

everpcpc (approved):

PR #17092, opened 2024-12-20
refactor(query): refactor udf server batch rows

b41sh:

## Summary

Setting `request_batch_rows` too small will cause a large number of concurrent connections to communicate with the udf server, which will affect the performance. For example, if the user sets `request_batch_rows` to 8 and a block has 65,536 rows, the block will generate 8192 flight connections.

This PR was modified as follows:
1. Add a new setting `external_server_request_max_threads` to control the total number of flight connections.
2. If the number of UDF threads exceeds `external_server_request_max_threads`, these threads will wait until the end of previous threads.
3. Change functions loop so that data in the block is split and concat only once, avoiding repeated split and concat operations when there are multiple functions.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17092)


src/query/service/src/pipelines/processors/transforms/transform_udf_server.rs
```diff
@@ -237,9 +257,9 @@ impl AsyncTransform for TransformUdfServer {
 
             let task_len = tasks.len() as u64;
             record_running_requests_external_start(func.name.clone(), task_len);
-            let blocks = futures::future::join_all(tasks).await;
+            let task_results = futures::future::join_all(tasks).await;
```
sundy-li, line 260: maybe use `futures::stream::iter(tasks).for_each_concurrent`

src/query/service/src/pipelines/processors/transforms/transform_udf_server.rs
```diff
@@ -190,19 +201,27 @@ impl AsyncTransform for TransformUdfServer {
 
     #[async_backtrace::framed]
     async fn transform(&mut self, mut data_block: DataBlock) -> Result<DataBlock> {
+        let rows = data_block.num_rows();
+        // Use block row numbers and request max threads to calculate the number of rows for each threads.
+        // In each thread, the block is split by the batch rows and call the udf server serially,
+        // so that the flight connection can be reused.
+        let mut batch_rows = rows / self.request_max_threads;
+        if batch_rows < self.request_batch_rows {
```
sundy-li, line 209: batch_rows should still be self.request_batch_rows

src/query/service/src/pipelines/processors/transforms/transform_udf_server.rs
sundy-li: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b737a80a798e0f603be01fbb7caeb09f

sundy-li: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4435eb4674a69b2b68f5c59cde7bc5ca

sundy-li: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=04620f7a98d4ebb9496e9086cced031f

I prefer this one.

sundy-li: But we may need a global permit during udf processors * thread, so the second one can meet our needs.

src/query/service/src/pipelines/processors/transforms/transform_udf_server.rs
```diff
@@ -73,11 +81,16 @@ impl TransformUdfServer {
...
     async fn transform_inner(
         ctx: Arc<QueryContext>,
+        semaphore: Arc<Semaphore>,
         connect_timeout: u64,
         request_timeout: u64,
         func: UdfFunctionDesc,
         mut data_block: DataBlock,
     ) -> Result<DataBlock> {
+        // Must obtain the permit to execute, prevent too many connections being executed concurrently
+        let _permit = semaphore.acquire_owned().await.map_err(|e| {
```
sundy-li, line 91: this should be inside the `batch_blocks ` iteration.

tests/suites/1_stateful/02_query/02_0010_udf.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+. "$CURDIR"/../../../shell_env.sh
+
+echo "SET external_server_request_max_threads=1; SELECT ping(number) FROM numbers(100000) IGNORE_RESULT;" | $BENDSQL_CLIENT_CONNECT
```
sundy-li, line 6: let's also decrease the setting of `batch_rows `

echo "SET external_server_request_batch_rows = 1024; SET external_server_request_max_threads=1; SELECT ping(number) FROM numbers(100000) IGNORE_RESULT;" | $BENDSQL_CLIENT_CONNECT

sundy-li (approved):

Issue #17060, opened 2024-12-17
Feature: Implement Date and Time Functions for Interval Type


TCeason:
**Summary**

Now Databend already supports Interval Type https://github.com/databendlabs/databend/pull/16990.

We need to implement several date and time functions in Databend that revolve around interval handling. These functions will enhance the capability of our SQL engine to manipulate and extract components from time intervals, which is crucial for users working with temporal data.

**Proposed Functions:**

- [ ] **Addition of an INTERVAL:**
  - Example: `INTERVAL '1 HOUR' + INTERVAL '5 HOUR'`
  - Expected Result: `INTERVAL '6 HOUR'`

- [ ] **Addition to a TIMESTAMP:**
  - Example: `TIMESTAMP '1992-03-22 01:02:03' + INTERVAL '5 DAY'`
  - Expected Result: `TIMESTAMP '1992-03-27 01:02:03'`

- [ ] **Subtraction of an INTERVAL:**
  - Example: `INTERVAL '5 HOUR' - INTERVAL '1 HOUR'`
  - Expected Result: `INTERVAL '4 HOUR'`

- [ ] **Subtraction from a TIMESTAMP:**
  - Example: `TIMESTAMP '1992-03-27 01:02:03' - INTERVAL '5 DAY'`
  - Expected Result: `TIMESTAMP '1992-03-22 01:02:03'`

- [ ] **epoch(interval):**
  - Get the total number of seconds, as a double precision floating point number, represented in the interval.

- [ ] **to_centuries(integer):**
  - Construct a century interval based on the provided integer value.

- [ ] **to_days(integer):**
  - Construct a day interval based on the provided integer value.

- [ ] **to_decades(integer):**
  - Construct a decade interval based on the provided integer value.

- [ ] **to_hours(integer):**
  - Construct an hour interval based on the provided integer value.

- [ ] **to_microseconds(integer):**
  - Construct a microsecond interval based on the provided integer value.

- [ ] **to_millennia(integer):**
  - Construct a millennium interval based on the provided integer value.

- [ ] **to_milliseconds(integer):**
  - Construct a millisecond interval based on the provided integer value.

- [ ] **to_minutes(integer):**
  - Construct a minute interval based on the provided integer value.

- [ ] **to_months(integer):**
  - Construct a month interval based on the provided integer value.

- [ ] **to_seconds(integer):**
  - Construct a second interval based on the provided integer value.

- [ ] **to_weeks(integer):**
[...]

PR #17093, opened 2024-12-20
feat(query): support interval function

TCeason:

## Summary

- fix: #17060

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17093)

### DIFF
--- a/src/common/column/src/types/native.rs
+++ b/src/common/column/src/types/native.rs
@@ -13,7 +13,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::cmp::Ordering;
 use std::convert::TryFrom;
+use std::hash::Hash;
+use std::hash::Hasher;
 use std::ops::Neg;
 use std::panic::RefUnwindSafe;
 
@@ -252,11 +255,7 @@ impl NativeType for days_ms {
     Copy,
     Clone,
     Default,
-    PartialEq,
-    PartialOrd,
-    Ord,
     Eq,
-    Hash,
     Zeroable,
     Pod,
     Serialize,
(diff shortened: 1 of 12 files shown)


src/query/functions/src/scalars/comparison.rs
```diff
@@ -79,6 +81,39 @@ const ALL_FALSE_DOMAIN: BooleanDomain = BooleanDomain {
     has_false: true,
 };
 
+fn register_interval_cmp(registry: &mut FunctionRegistry) {
+    registry.register_comparison_2_arg::<IntervalType, IntervalType, _, _>(
```
sundy-li, line 85: maybe `register_simple_domain_type_cmp`

src/query/functions/src/scalars/interval.rs
```diff
@@ -79,3 +86,227 @@ fn register_interval_to_string(registry: &mut FunctionRegistry) {
...
+
+    registry
+        .register_passthrough_nullable_2_arg::<TimestampType, IntervalType, TimestampType, _, _>(
+            "plus",
+            |_, _, _| FunctionDomain::MayThrow,
+            vectorize_with_builder_2_arg::<TimestampType, IntervalType, TimestampType>(
+                |a, b, output, ctx| {
+                    // plus microseconds and days
+                    let ts = a
+                        .wrapping_add(b.microseconds())
```
sundy-li, line 113: why not use `total_micros `

TCeason: Add total micros is not right. 

If interval compression 

30day is equal with 1 month.

But if timestamp add 30 days and add 1month maybe diff.

src/common/column/src/types/native.rs
```diff
@@ -268,6 +267,33 @@ impl NativeType for days_ms {
 #[repr(C)]
 pub struct months_days_micros(pub i128);
 
+const MICROS_PER_DAY: i64 = 24 * 3600 * 1_000_000;
+const MICROS_PER_MONTH: i64 = 30 * MICROS_PER_DAY;
```
b41sh, line 271: The number of days in each month is different, can we use a fixed number?

TCeason: > The number of days in each month is different, can we use a fixed number?

Well, there's no need to think about it. The same is true in other implementations of the interval type of db. Like duckdb and snowflake

src/query/functions/src/scalars/comparison.rs
TCeason: Yes.

b41sh (approved):

sundy-li (approved):

PR #17106, opened 2024-12-24
refactor(query): refactor udf server connection reuse endpoint

b41sh:

## Summary

Reuse the `Endpoint` of the flight connection to avoid repeated reading of the `ClientTlsConfig` file and speed up the query. For querying 100,000 rows data, the execution time reduced from 18 seconds to 8 seconds

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17106)


Xuanwo:
Hi, can we reuse a TlsConnector instead? 

The map of Arc<Endpoint> seems not cool for me. We still need to load certs for every domains and queries.

If we can use a global TlsConnector, we just need to load certs once instead.

src/query/service/src/pipelines/processors/transforms/transform_udf_server.rs
```diff
@@ -66,24 +69,37 @@ impl TransformUdfServer {
         let retry_times = settings.get_external_server_request_retry_times()? as usize;
         let semaphore = Arc::new(Semaphore::new(request_max_threads));
 
-        let s = Self {
+        let mut endpoint_map: BTreeMap<String, Arc<Endpoint>> = BTreeMap::new();
```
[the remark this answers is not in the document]
b41sh: I think it's OK. Not much difference.

Xuanwo (approved):

src/query/service/src/pipelines/processors/transforms/transform_udf_server.rs
```diff
@@ -66,33 +69,42 @@ impl TransformUdfServer {
         let retry_times = settings.get_external_server_request_retry_times()? as usize;
         let semaphore = Arc::new(Semaphore::new(request_max_threads));
 
-        let s = Self {
+        let mut endpoints: BTreeMap<String, Arc<Endpoint>> = BTreeMap::new();
+        for func in funcs.iter() {
+            let server_addr = func.udf_type.as_server().unwrap();
+            if let Some(endpoint) = endpoints.get(server_addr) {
```
Xuanwo, line 75: We can continue if endpoints.get(addr).is_some().

b41sh: fixed, thanks.

Issue #17143, opened 2024-12-30
Implement spill fallback logic during runtime


Xuanwo:
https://github.com/databendlabs/databend/blob/7548f99a9b7b722c43218c2cbc055d7612cb313f/src/query/config/src/config.rs#L3023-L3030

            

Directly altering the configuration is error-prone; we should implement this fallback logic during runtime.

Issue #17142, opened 2024-12-30
feat: add storage.spill bucket


BohuTANG:
**Summary**

To direct the spill data to a different bucket (e.g., a non-versioned bucket):
```
# Storage config.
[storage]

...

[storage.spill]
bucket = "spill-bucket"
```

* If the `[storage.spill]` section is not configured (i.e., set to None), the system will default to using the path: `${storage.bucket}/${storage.root}/_spill/${queryID}/`

* If the `[storage.spill]` section is configured, the system will use the path: `${storage.spill.bucket}/${storage.root}/_spill/${queryID}/`

PR #17159, opened 2025-01-02
feat(query): support direct the spill data to a different bucket 

forsaken628:

## Summary

- fixes: #17142
- fixes: #17143

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17159)


src/query/service/src/pipelines/processors/transforms/transform_stream_sort_spill.rs
```diff
@@ -1089,7 +1089,7 @@ mod tests {
         let fixture = TestFixture::setup().await?;
         let ctx = fixture.new_query_ctx().await?;
 
-        let op = DataOperator::instance().operator();
+        let op = DataOperator::instance().spill_operator();
```
sundy-li, line 1092: also in `vacuum_temporary_files.rs`

Xuanwo:
We are modifying as much code as we are adding with `spill.storage`, so why not use `spill.storage` instead? Adding a `spill_bucket` doesn’t seem like a good idea to me. It’s also difficult to document for users. For instance, there is no `bucket` concept in `azblob`.

forsaken628:
https://github.com/databendlabs/databend/blob/7ec5a8d41d4ded33a047787f9638f1b4bf46b5e7/src/query/config/src/config.rs#L239-L253

This is a better layout, but there's no time to refactor it.

sundy-li:
We need a test to cover that spill could work out with other storage params.

Maybe in `.github/actions/test_ee_standalone_linux/action.yml`

sundy-li (approved):

PR #17263, opened 2025-01-13
fix: vacuum index not work if index is dropped by create or replace

SkyFan2002:

## Summary

The `create or replace index` operation may cause an index to be dropped, and such dropped indexes should also be recorded in `__fd_marked_deleted_index`

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17263)

### DIFF
--- a/src/meta/api/src/name_id_value_api.rs
+++ b/src/meta/api/src/name_id_value_api.rs
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::any::Any;
 use std::future::Future;
 
 use databend_common_meta_app::data_id::DataId;
 use databend_common_meta_app::id_generator::IdGenerator;
 use databend_common_meta_app::primitive::Id;
+use databend_common_meta_app::schema::IndexMeta;
 use databend_common_meta_app::tenant_key::ident::TIdent;
 use databend_common_meta_app::tenant_key::resource::TenantResource;
 use databend_common_meta_app::KeyWithTenant;
@@ -38,6 +40,7 @@ use log::debug;
 use crate::kv_pb_api::KVPbApi;
 use crate::kv_pb_api::UpsertPB;
 use crate::meta_txn_error::MetaTxnError;
+use crate::schema_api_impl::mark_index_as_deleted;
 use crate::txn_backoff::txn_backoff;
 use crate::txn_op_del;
 use crate::util::fetch_id;
@@ -113,6 +116,15 @@ where
                         let kvs = associated_records(seq_id.data);
                         for (k, _v) in kvs {
                             txn.if_then.push(TxnOp::delete(k));
+                        }
+
+                        let index_meta = <dyn Any>::downcast_ref::<IndexMeta>(&seq_meta.data);
+                        if let Some(index_meta) = index_meta {
+                            txn.if_then.push(mark_index_as_deleted(
+                                name_ident.tenant(),
+                                index_meta.table_id,
+                                *seq_id.data,
+                            )?);
                         }
                     } else {
                         return Ok(Err(seq_id));
--- a/src/meta/api/src/name_id_value_api.rs
+++ b/src/meta/api/src/name_id_value_api.rs
@@ -71,6 +71,9 @@ where
     /// Such operations do not have any condition constraints.
     /// For example, a `name -> id` mapping can have a reverse `id -> name` mapping.
     ///
+    /// `mark_delete_records` is used to generate additional key-values for implementing `mark_delete` operation.
+    /// For example, when an index is dropped by `override_exist`,  `__fd_marked_deleted_index/<table_id>/<index_id> -> marked_deleted_index_meta` will be added.
+    ///
     /// If there is already a `name_ident` exists, return the existing id in a `Ok(Err(exist))`.
     /// Otherwise, create `name -> id -> value` and returns the created id in a `Ok(Ok(created))`.
     async fn create_id_value<A, M>(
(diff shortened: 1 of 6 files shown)


drmingdrmer (commented):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17263)* status: all files reviewed, 2 unresolved discussions (waiting on @b41sh and @SkyFan2002)

___
*[`src/meta/api/src/name_id_value_api.rs` line 128 at r1](https://reviewable.io/reviews/databendlabs/databend/17263#-OGUVaha7z5kGVVJFD_8:-OGUVaha7z5kGVVJFD_9:b-m4ha7u) ([raw file](https://github.com/databendlabs/databend/blob/9508095871dd673d9621c05ca2f2fee6d737fe2e/src/meta/api/src/name_id_value_api.rs#L128)):*
> ```XML
>                                 *seq_id.data,
>                             )?);
>                         }
> ```

use the `associated_records` to add additional key-value to write.



_Code quote:_
```XML
                        let index_meta = <dyn Any>::downcast_ref::<IndexMeta>(&seq_meta.data);
                        if let Some(index_meta) = index_meta {
                            txn.if_then.push(mark_index_as_deleted(
                                name_ident.tenant(),
                                index_meta.table_id,
                                *seq_id.data,
                            )?);
                        }
```
___
*[`src/meta/kvapi/src/kvapi/value.rs` line 21 at r1](https://reviewable.io/reviews/databendlabs/databend/17263#-OGUVzmqBGfkquPpmobv:-OGUVzmqBGfkquPpmobw:b-5emih7) ([raw file](https://github.com/databendlabs/databend/blob/9508095871dd673d9621c05ca2f2fee6d737fe2e/src/meta/kvapi/src/kvapi/value.rs#L21)):*
> ```XML
> 
> /// A value that can be stored in kvapi::KVApi.
> pub trait Value: Debug + Any {
> ```

remove Any

src/meta/api/src/name_id_value_api.rs
```diff
@@ -114,6 +117,15 @@ where
                         for (k, _v) in kvs {
                             txn.if_then.push(TxnOp::delete(k));
                         }
+
+                        let index_meta = <dyn Any>::downcast_ref::<IndexMeta>(&seq_meta.data);
```
b41sh, line 121: How about implementing the `create_index` function without using the generic `create_id_value` function ? Special treatment here may not be a good idea.

drmingdrmer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17263)* status: all files reviewed, 3 unresolved discussions (waiting on @SkyFan2002)

___
*[`src/meta/api/src/name_id_value_api.rs` line 128 at r1](https://reviewable.io/reviews/databendlabs/databend/17263#-OGUVaha7z5kGVVJFD_8:-OGWBQXuDaVzTh3REtIu:b-qq04na) ([raw file](https://github.com/databendlabs/databend/blob/9508095871dd673d9621c05ca2f2fee6d737fe2e/src/meta/api/src/name_id_value_api.rs#L128)):*

use the `associated_records` to add additional key-value to write.
</blockquote></details>

Hmm... this part seems a bit strange. `IndexNameIdent` only contains two pieces of information: `tenant` and `index-name`:  
https://github.com/datafuselabs/databend/blob/88fb9e3e1e20d57cf26571fa72feeb9861b5fb2a/src/meta/app/src/schema/index_name_ident.rs#L19

But `IndexMeta` includes `table-id`:  
https://github.com/datafuselabs/databend/blob/88fb9e3e1e20d57cf26571fa72feeb9861b5fb2a/src/meta/app/src/schema/index.rs#L56

So does this imply that when a new index overwrites an old one, it can overwrite it as an index for another table?

If we want to adapt to this scenario, we can pass the `meta-value` to `associated_records()` to create the key-values that need to be deleted together. The `name-id-value-api` is a low-level API and should not rely on specific value types.

```rust
async fn create_id_value<A>(
    &self,
    name_ident: &K,
    value: &IdRsc::ValueType,
    override_exist: bool,
    associated_records: A,
) -> Result<Result<DataId<IdRsc>, SeqV<DataId<IdRsc>>>, MetaTxnError>
where
    A: Fn(DataId<IdRsc>) -> Vec<(String, Vec<u8>)> + Send,
    // Updated to include the value type
    A: Fn(DataId<IdRsc>, &IdRsc::ValueType) -> Vec<(String, Vec<u8>)> + Send,
```

SkyFan2002:
@drmingdrmer Yes, the aggregating index was implemented as an object independent from the table, when a new index overwrites an old one, it can overwrite it as an index for another table. Thank you for the suggestion to pass the meta-value to associated_records()

src/meta/api/src/name_id_value_api.rs
SkyFan2002: Thanks. Special treatment is indeed not a good idea.

drmingdrmer (approved):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17263)* status: all files reviewed, 1 unresolved discussion (waiting on @b41sh)

b41sh (approved):

PR #17266, opened 2025-01-13
feat(metrics): display scan rows/bytes, write rows/bytes in global counter metrics

flaneur2020:

## Summary

currently we've tracked the scan rows/bytes, write rows/bytes progress value in query stats. and collected them into metrics in query_finished.

but this have an issue like, given a COPY statement which executed for 10 hours. this `databend_query_scan_rows` will kept to be zero during the executation, can not reflect the executing progress in metrics.

this pr add a hook on the `Progress` struct to increment the global metrics as well as the local stats.

## Tests

- [x] No Test - simple metrics enhancement

## Type of change

- [x] Other (please describe): observability

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17266)

### DIFF
--- a/src/common/base/src/base/progress.rs
+++ b/src/common/base/src/base/progress.rs
@@ -19,8 +19,6 @@ use std::sync::Arc;
 use serde::Deserialize;
 use serde::Serialize;
 
-use crate::runtime::metrics::Counter;
-
 #[derive(Debug, Clone, Deserialize, Serialize, Default)]
 pub struct ProgressValues {
     pub rows: usize,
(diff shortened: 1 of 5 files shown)


zhang2014:
The metrics recorder has some locks, let's run the performance tests.

src/common/base/src/base/progress.rs
```diff
@@ -24,24 +24,43 @@ pub struct ProgressValues {
...
+            bytes: self.bytes,
+            hook: Some(hook),
         }
     }
 
     pub fn incr(&self, progress_values: &ProgressValues) {
         self.rows.fetch_add(progress_values.rows, Ordering::Relaxed);
         self.bytes
             .fetch_add(progress_values.bytes, Ordering::Relaxed);
+        if let Some(hook) = &self.hook {
```
zhang2014, line 61: this might result in double counting if cluster mode.

flaneur2020: can you help explain more about this?

flaneur2020:
close this due to performance decrease, let me try another with prometheus' collector

PR #17291, opened 2025-01-15
feat: support vacuum inverted index

SkyFan2002:

## Summary

This PR implements a new table function `fuse_vacuum_drop_inverted_index()` to clean up the data of dropped and outside the retention period inverted indexes.

### Implemention
A new key-value pair is added to the meta-service:

```
__fd_marked_deleted_table_index/table_id/index_name/index_version-> marked_deleted_index_meta
```

When an inverted index is dropped or replaced, the `fd_marked_deleted_table_index` key-value pair is added.

When a vacuum is triggered, the meta-service will check the `__fd_marked_deleted_table_index` key. And filter out the indexes that is in retention period with `MarkedDeletedIndexMeta.dropped_on`.

The vacuum will delete the index data that is not in retention period, by identifying the index files with `index_name` and `index_version`. After that, the meta-service will remove the index meta from the `__fd_marked_deleted_table_index/index_name/index_version` key.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17291)


b41sh (approved):

drmingdrmer (changes requested):
@b41sh mentioned that Index and Table may not have a one-to-one mapping, as an aggregate index could be built from multiple tables. At the time I suggested adding `table_id` to the index key, I wasn’t aware of this. Should we reconsider the design of the marked-deleted key in light of this information?

> Reviewed 13 of 21 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17291#-:-OGibeNV6TWLMUSVbJPP:b-d34tow)* status: 13 of 21 files reviewed, 2 unresolved discussions (waiting on @SkyFan2002)

___
*[`src/meta/app/src/schema/table.rs` line 889 at r1](https://reviewable.io/reviews/databendlabs/databend/17291#-OGiewg20isvTfYlTpiA:-OGiewg20isvTfYlTpiB:b-lia0hr) ([raw file](https://github.com/databendlabs/databend/blob/653d4061607b61d9eca1963f530480fc1d6e24d1/src/meta/app/src/schema/table.rs#L889)):*
> ```XML
> #[derive(Clone, Debug, PartialEq, Eq)]
> pub struct GetMarkedDeletedTableIndexesReply {
>     pub table_indexes: HashMap<u64, Vec<(String, String, MarkedDeletedIndexMeta)>>,
> ```

What about introducing two type alias `IndexName = String` and `IndexVersion = String` to improve readability?

Or make it more strict, by defining a explicit type for index version: `struct IndexVersion(String);`
___
*[`src/meta/api/src/schema_api_impl.rs` line 2655 at r1](https://reviewable.io/reviews/databendlabs/databend/17291#-OGigdalDO8JxYIHNfzg:-OGigdalDO8JxYIHNfzh:b-1985py) ([raw file](https://github.com/databendlabs/databend/blob/653d4061607b61d9eca1963f530480fc1d6e24d1/src/meta/api/src/schema_api_impl.rs#L2655)):*
> ```XML
>     #[logcall::logcall]
>     #[fastrace::trace]
>     async fn get_marked_deleted_table_indexes(
> ```

The implementation of this method is actually a `list` operation thus this method should be `list_marked_deleted_table_indexes()`

SkyFan2002:
@drmingdrmer adding `table_id` to the index key makes `vacuum index from specific table` more convient and efficient. 

Currently, we don't have indexes associated with multiple tables. If we add such indexes in the future, they would be fundamentally different from our current indexes. I propose that if we add such indexes in the future, we should use a separate key to store these dropped indexes.

drmingdrmer (approved):
> Reviewed 5 of 21 files at r1, 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17291)* status: all files reviewed, 4 unresolved discussions (waiting on @SkyFan2002)

___
*[`src/meta/app/src/schema/table.rs` line 893 at r2](https://reviewable.io/reviews/databendlabs/databend/17291#-OGks8Sm3K-1PmmDHGu9:-OGks8Sm3K-1PmmDHGuA:b-f9uhlp) ([raw file](https://github.com/databendlabs/databend/blob/5dabc316d21171a65ad6cf567c192446006850f0/src/meta/app/src/schema/table.rs#L893)):*
> ```XML
> 
> pub type IndexName = String;
> pub type IndexVersion = String;
> ```

Since these aliases already exist, use them throughout the codebase to make the code more self-explanatory and improve readability.
___
*[`src/meta/api/src/schema_api_impl.rs` line 2679 at r2](https://reviewable.io/reviews/databendlabs/databend/17291#-OGkn_u19MYtG8eE3Txq:-OGkn_u19MYtG8eE3Txr:undefined) ([raw file](https://github.com/databendlabs/databend/blob/5dabc316d21171a65ad6cf567c192446006850f0/src/meta/api/src/schema_api_impl.rs#L2679)):*
> ```XML
>                 DirName::new_with_level(ident, 3)
>             }
>         };
> ```





_Suggestion:_
```XML
        let dir = {
            let table_id = table_id.unwrap_or_default();
            let ident = MarkedDeletedTableIndexIdIdent::new_generic(
                tenant,
                MarkedDeletedTableIndexId::new(
                    table_id,
                    "dummy".to_string(),
                    "dummy".to_string(),
                ),
            );
            DirName::new_with_level(ident, 2)
        };
```
___
*[`src/meta/api/src/schema_api_impl.rs` line 3249 at r2](https://reviewable.io/reviews/databendlabs/databend/17291#-OGkrV9B9oHCzaleO1Bq:-OGkrV9B9oHCzaleO1Br:b-39al3p) ([raw file](https://github.com/databendlabs/databend/blob/5dabc316d21171a65ad6cf567c192446006850f0/src/meta/api/src/schema_api_impl.rs#L3249)):*
> ```XML
>         tenant: &Tenant,
>         table_id: u64,
>         indexes: &[(String, String)],
> ```

Is there need to remove a collection of indexes in  transaction? AFAIK, they are independent and can be removed one by one.



_Code quote:_
```XML
    async fn remove_marked_deleted_table_indexes(
        &self,
        tenant: &Tenant,
        table_id: u64,
        indexes: &[(String, String)],
```

Issue #17292, opened 2025-01-15
Refactor udf script


sundy-li:
**Summary**

Deadlock issue of global python runtime：

 

1. Investigate creating python/js/wasm runtime per processor
2. enable `sandboxed(true)` in python runtime

```
/// Click on each function to see the example.
///
/// # Parallelism
///
/// As we know, Python has a Global Interpreter Lock (GIL) that prevents multiple threads from executing Python code simultaneously.
/// To work around this limitation, each runtime creates a sub-interpreter with its own GIL. This feature requires Python 3.12 or later.
///
/// [`add_function`]: Runtime::add_function
/// [`add_aggregate`]: Runtime::add_aggregate
/// [`call`]: Runtime::call
/// [`call_table_function`]: Runtime::call_table_function
/// [`create_state`]: Runtime::create_state
/// [`accumulate`]: Runtime::accumulate
/// [`accumulate_or_retract`]: Runtime::accumulate_or_retract
/// [`merge`]: Runtime::merge
/// [`finish`]: Runtime::finish
pub struct Runtime {
    interpreter: SubInterpreter,
    functions: HashMap<String, Function>,
    aggregates: HashMap<String, Aggregate>,
    converter: pyarrow::Converter,
}
```

PR #17304, opened 2025-01-16
refactor(query): introduce udf runtime pool

forsaken628:

## Summary

- fixes: #17292

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17304)


sundy-li:
Need shell tests that execute 4 sqls with python/js udf on parallel.

20+_others/20_0023_udf_script.sh

src/query/service/src/pipelines/processors/transforms/runtime_pool.rs
```diff
@@ -0,0 +1,54 @@
...
+        E: From<B::Error>,
+    {
+        let mut runtimes = self.runtimes.lock().unwrap();
+        let runtime = match runtimes.pop() {
+            Some(runtime) => runtime,
+            None => self.builder.build()?,
+        };
+        drop(runtimes);
+
+        let result = op(&runtime)?;
```
sundy-li, line 47:
```suggestion
         let result = op(&runtime);

         let mut runtimes = self.runtimes.lock().unwrap();
         runtimes.push(runtime);
         result
```

src/query/service/src/pipelines/processors/transforms/transform_udf_script.rs
```diff
@@ -212,79 +254,39 @@ impl Transform for TransformUdfScript {
...
-                    let new_runtime = ScriptRuntime::try_create(lang, code_opt, runtime_num)
-                        .map(Arc::new)
-                        .map_err(|err| {
-                            ErrorCode::UDFDataError(format!(
-                                "Failed to create UDF runtime for language {lang:?} with error: {err}",
-                            ))
-                        })?;
-                    entry.insert(new_runtime).clone()
-                }
+            if let Entry::Vacant(entry) = script_runtimes.entry(func.name.clone()) {
```
b41sh, line 279: key should include function language and function name, different language may have same function name

src/query/service/src/pipelines/processors/transforms/runtime_pool.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct Pool<R, B: RuntimeBuilder<R>> {
+    pub builder: B,
+    runtimes: Mutex<Vec<R>>,
+}
+
+impl<R, B: RuntimeBuilder<R>> Pool<R, B> {
+    pub fn new(builder: B) -> Self {
+        Self {
+            builder,
+            runtimes: Mutex::new(vec![]),
```
b41sh, line 31: Why not generate some runtimes during initialisation, the default number of threads is the same as the number of cpus, will the initial runtime number 0 cause a scramble on the first run and affect performance?

b41sh (approved):

src/query/service/src/pipelines/processors/transforms/transform_udf_script.rs
forsaken628:
```
create or replace FUNCTION echo1 () RETURNS INT64 LANGUAGE javascript HANDLER = 'echo' AS $$
export function echo() {
    return 1;
}
$$;

create or replace FUNCTION echo2 () RETURNS INT64 LANGUAGE javascript HANDLER = 'echo' AS $$
export function echo() {
    return 2;
}
$$;

select echo1(),echo2();
-- 1,1
```
the handler is not  unique.

src/query/service/src/pipelines/processors/transforms/runtime_pool.rs
forsaken628: For udf, the maximum number of instances is the number of threads; for udaf it's a bit more complicated because udaf will have multiple phases.
The current runtime's lifecycle is within one sql. What is the benefit of early initialization? The benefit of delayed initialization is obvious: it is possible that a few instances are enough. This implementation is a copy of arrow_udf_wasm, which is why wasm doesn't have a pool, so it might make sense to put the runtime pool inside arrow_udf, for example, to save parse script ast every time.

src/query/service/src/pipelines/processors/transforms/transform_udf_script.rs
b41sh: Actually, the name here is the handler name and not the function name, which is confusing. It's better to add some comments or change the field name to `handler_name`.

src/query/service/src/pipelines/processors/transforms/runtime_pool.rs
b41sh: Ok, may be we can contribute this feature to `arrow_udf`.

b41sh (approved):

sundy-li (approved):

BohuTANG:
Need backport to 680?

PR #17366, opened 2025-01-24
chore(query): bump jiff crate

sundy-li:

## Summary

chore(query): bump jiff crate

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17366)


src/query/expression/tests/it/types.rs
```diff
@@ -35,6 +37,36 @@ fn test_timestamp_to_string_formats() {
     );
 }
 
+#[test]
+fn test_parse_jiff() {
+    let (mut tm, offset) = BrokenDownTime::parse_prefix(
```
TCeason, line 42: Maybe can consider https://docs.rs/jiff/latest/jiff/fmt/strtime/fn.format.html it will be better if format is valid.

sundy-li: but this's just for testing..

src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -698,24 +597,15 @@ fn register_to_string(registry: &mut FunctionRegistry) {
         |_, _, _| FunctionDomain::MayThrow,
         vectorize_with_builder_2_arg::<TimestampType, StringType, NullableType<StringType>>(
             |micros, format, output, ctx| {
-                if format.is_empty() {
-                    output.push_null();
-                } else {
-                    let items = StrftimeItems::new(format);
-                    if items.clone().any(|item| matches!(item, Item::Error)) {
-                        ctx.set_error(output.len(), "Invalid format string".to_string());
+                let ts = micros.to_timestamp(ctx.func_ctx.jiff_tz.clone());
```
TCeason, line 600: The name of jiff_tz is a different with Chrono tz. But now we will drop Chrono in function crate, so maybe we can rename it as tz

tests/sqllogictests/suites/query/functions/02_0012_function_datetimes_tz.test
```diff
@@ -824,8 +824,10 @@ select to_datetime('1', '%s')
 ----
 1970-01-01 00:00:01.000000
 
-statement error 1006
+query T
 select to_timestamp('200,2000', '%s,%Y');
+----
+2000-01-01 00:03:20.000000
```
TCeason, line 830: Why the time part is 00:03:20?

And in this pr we modify the function "date_format" so the item of formatting maybe changed. So we need to know the new item and what's it means.

E.g. 
%Y -> means year

src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -692,30 +600,30 @@ fn register_number_to_date(registry: &mut FunctionRegistry) {
...
+                match write!(
+                    output.builder.row_buffer,
+                    "{}",
+                    ts.strftime(format.as_ref())
+                ) {
+                    Ok(_) => {
+                        output.builder.commit_row();
+                        output.validity.push(true);
+                    }
+                    Err(e) => {
```
TCeason, line 620:
```sql
in pr:

:) select date_format('2022-02-02 01:00:00','%X%\%C');
error: APIError: QueryFailed: [1104]a formatting trait implementation returned an error when the underlying stream did not


in main :

:) select date_format('2022-02-02 01:00:00','%X%\%C');
error: APIError: QueryFailed: [1006]Invalid format string while evaluating function `to_string('2022-02-02 01:00:00.000000', '%X%\%C')` in expr `to_string('2022-02-02 01:00:00.000000', '%X%\%C')`
```

Issue #17384, opened 2025-01-29
Feature: Allow Extract EPOCH from timestamp


rad-pat:
It is possible to create a timestamp from an integer as number of seconds since UNIX epoch.

Can we please have a function to extract the same from a timestamp value?

https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT

TCeason:
Please see doc: https://docs.databend.com/sql/sql-functions/datetime-functions/extract

rad-pat:
I see the doc, but it does not allow EPOCH:
```
EXTRACT( YEAR | QUARTER | MONTH | WEEK | DAY | HOUR | MINUTE | SECOND | DOW | DOY FROM <date_or_time_expr> )
```

TCeason:
> I can do this, but it's a little ugly and not very intuitive
> ```
> select to_timestamp('2025-01-01 10:00:01') - to_timestamp(0)
> 1735725601000000
> ```

Fine you need this 

SELECT EXTRACT(EPOCH FROM ts);

We can add it.

rad-pat:
> We can add it.

Thanks 🙏

TCeason:
> I can do this, but it's a little ugly and not very intuitive
> 
> ```
> select to_timestamp('2025-01-01 10:00:01') - to_timestamp(0)
> 1735725601000000
> ```

you can use to_unix_timestamp tmply

TCeason:
https://docs.databend.com/sql/sql-functions/datetime-functions/to-unix-timestamp

rad-pat:
> Unfortunately it truncates to integer:
> 
> ```
> SELECT (to_timestamp('2023-11-12 09:38:18.965575') - to_timestamp(0)) / 1000000.0, 
> 	to_unix_timestamp('2023-11-12 09:38:18.965575')
> 
> (to_timestamp('2023-11-12 09:38:18.965575') - to_timestamp(0)) / 1000000.0|to_unix_timestamp('2023-11-12 09:38:18.965575')|
> --------------------------------------------------------------------------+-----------------------------------------------+
>                                                          1699781898.965575|                                     1699781898|
> ```

I think we can still use it though 👍

PR #17385, opened 2025-01-29
feat(query): support extract EPOCH from timestamp

TCeason:

## Summary

- fixes: #17384

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17385)

### DIFF
--- a/src/query/functions/src/scalars/timestamp/src/datetime.rs
+++ b/src/query/functions/src/scalars/timestamp/src/datetime.rs
@@ -1525,7 +1525,7 @@ fn register_to_number_functions(registry: &mut FunctionRegistry) {
     );
 
     registry.register_passthrough_nullable_1_arg::<TimestampType, Float64Type, _, _>(
-        "epoch_microsecond",
+        "epoch",
         |_, domain| {
             FunctionDomain::Domain(SimpleDomain::<F64> {
                 min: (domain.min as f64 / 1_000_000f64).into(),
--- a/src/query/functions/src/scalars/timestamp/src/datetime.rs
+++ b/src/query/functions/src/scalars/timestamp/src/datetime.rs
@@ -1524,6 +1524,19 @@ fn register_to_number_functions(registry: &mut FunctionRegistry) {
         }),
     );
 
+    registry.register_passthrough_nullable_1_arg::<TimestampType, Float64Type, _, _>(
+        "epoch_microsecond",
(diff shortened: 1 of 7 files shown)


src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -1516,6 +1516,19 @@ fn register_to_number_functions(registry: &mut FunctionRegistry) {
         }),
     );
 
+    registry.register_passthrough_nullable_1_arg::<TimestampType, Float64Type, _, _>(
+        "epoch_microsecond",
+        |_, _| FunctionDomain::Full,
+        vectorize_1_arg::<TimestampType, Float64Type>(|val, ctx| {
```
sundy-li, line 1522: why not return the val directly?

sundy-li: the epoch number will be consistent in all timezones.

src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -1524,6 +1524,12 @@ fn register_to_number_functions(registry: &mut FunctionRegistry) {
         }),
     );
 
+    registry.register_passthrough_nullable_1_arg::<TimestampType, Float64Type, _, _>(
+        "epoch_microsecond",
```
sundy-li, line 1528: val is already in `microsecond`, the domain should be same as input.

TCeason: Done

sundy-li:
Better to rewrite `epoch(timestamp)` ---> `timestamp`  in binder type_checker

TCeason:
> type_checker

You mean try_rewrite_sugar_function?

sundy-li:
> You mean try_rewrite_sugar_function?

Yes

TCeason:
> try_rewrite_sugar_function

We also have epoch(i64)->intervaltype in interval.rs. So impl a new function is suitable.

sundy-li (approved):

TCeason:
cc @soyeric128 doc needed

Issue #17398, opened 2025-02-03
bug: parse datetime started failing after update to 1.2.691


rad-pat:
### Search before asking

- [x] I had searched in the [issues](https://github.com/databendlabs/databend/issues) and found no similar issues.

### Version

1.2.691-nightly

### What's Wrong?

We upgrade from 1.2.680-p3 to 1.2.691 to include latest fixes and a date parsing in a CI test began to fail for us.

We tried settings such as those below, but still get the error message.
```
set GLOBAL enable_strict_datetime_parser = 0;
set GLOBAL parse_datetime_ignore_remainder = 1;
```

### How to Reproduce?

```
SELECT to_timestamp('2019_09', '%Y_%m');

Query failed: QueryErrors{code=1006, message=BadArguments. Code: 1006, Text = 2019_09 to datetime error date required to parse datetime: a month/day, day-of-year or week date must be present to create a date, but none were found. while evaluating function `to_timestamp('2019_09', '%Y_%m')` in expr `to_timestamp('2019_09', '%Y_%m')`} cause: nu
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

TCeason:
https://github.com/databendlabs/databend/pull/17366

Maybe date_format need to use Chrono.

PR #17399, opened 2025-02-03
fix(query): fix time parser with enable_strict_datetime_parser = 0

sundy-li:

## Summary

fix(query): fix time parser with enable_strict_datetime_parser = 0

 fixes:  #17398
 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17399)


src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -374,6 +374,15 @@ fn string_to_format_timestamp(
         let _ = tm.set_second(Some(0));
     }
 
+    if !ctx.func_ctx.enable_strict_datetime_parser {
```
TCeason, line 377: Maybe hour minute second also need respect this check? Now we pad it directly

TCeason (approved):

Issue #17400, opened 2025-02-03
bug: group checker incorrect


sundy-li:
**Summary**

```

CREATE OR REPLACE  TABLE u (
  id VARCHAR NULL
);



CREATE OR REPLACE TABLE c (
  id VARCHAR NULL,
  encounter_id VARCHAR NULL,
  person_id VARCHAR NULL,
  client_code VARCHAR NULL,
  what_fuck BOOLEAN NOT NULL,  -- change to `what_fuck BOOLEAN NULL` or remove this column will be fine
  payload VARIANT NULL
);


select
    ids.id,
    array_filter(array_agg(case when px.payload is not null and px.payload:px_code_set <> 'CPT' then px.payload:px_key else null end),x-> x is not null) as px_payload
  from u ids
    left join c  px on px.encounter_id = ids.id
  group by ids.id;
 

error: APIError: QueryFailed: [1065]error:
  --> SQL:3:139
  |
1 | select
2 |     ids.id,
3 |     array_filter(array_agg(case when px.payload is not null and px.payload:px_code_set <> 'CPT' then px.payload:px_key else null end),x-> x is not null) as px_payload
  |                                                                                                                                           ^ column "x" must appear in the GROUP BY clause or be used in an aggregate function

```

PR #17402, opened 2025-02-03
fix(query): fix lambda function bind column failed

b41sh:

## Summary

We use the number of bound columns as the column index of the lambda column, but since the column index may be discontinuous, e.g. if there is a not null column in the join table, a new nullable wrapper is generated to replace the original column, and a new column index is generated. In this case, the parameter of lambda function will got a wrong column index and cause resolve lambda function fail, this PR fixes this bug.

- fixes: #17400

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17402)

### DIFF
--- a/src/query/sql/src/planner/expression_parser.rs
+++ b/src/query/sql/src/planner/expression_parser.rs
@@ -366,13 +366,14 @@ pub fn parse_lambda_expr(
     // The column index may not be consecutive, and the length of columns
     // cannot be used to calculate the column index of the lambda argument.
     // We need to start from the current largest column index.
-    let mut column_index = 0;
-    for column in lambda_context.all_column_bindings().iter().rev() {
-        if column.index >= column_index {
-            column_index = column.index + 1;
-        }
-    }
+    let mut column_index = lambda_context
+        .all_column_bindings()
+        .iter()
+        .map(|c| c.index)
+        .max()
+        .unwrap_or_default();
     for (lambda_column, lambda_column_type) in lambda_columns.iter() {
+        column_index += 1;
         lambda_context.add_column_binding(
             ColumnBindingBuilder::new(
                 lambda_column.clone(),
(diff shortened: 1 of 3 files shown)


tests/sqllogictests/suites/query/functions/02_0061_function_array.test
```diff
@@ -419,6 +419,32 @@ SELECT arrays_zip(col1, col2) FROM t3;
...
+CREATE OR REPLACE TABLE c (
+  id VARCHAR NULL,
+  what_fuck BOOLEAN NOT NULL,
+  payload VARIANT NULL
+);
+
+statement ok
+INSERT INTO c VALUES(1, true, '[1,2]'),(1, false, '[3,4]'),(2, true, '123');
+
+query IT
```
sundy-li, line 440: ordering

src/query/sql/src/planner/expression_parser.rs
```diff
@@ -356,30 +356,39 @@ pub fn parse_computed_expr_to_string(
...
+
+    // The column index may not be consecutive, and the length of columns
+    // cannot be used to calculate the column index of the lambda argument.
+    // We need to start from the current largest column index.
+    let mut column_index = 0;
+    for column in lambda_context.all_column_bindings().iter().rev() {
+        if column.index >= column_index {
+            column_index = column.index + 1;
+        }
+    }
```
sundy-li, line 374:
```suggestion
let column_index =  lambda_context.all_column_bindings().iter().map(|c| c.index).max().unwrap_or_default() + 1;
```

Issue #17440, opened 2025-02-12
bug: Full Table Scan Issue with UNNEST(SPLIT(...)) and IN/CTE Queries in Databend


wubx:
### Search before asking

- [x] I had searched in the [issues](https://github.com/databendlabs/databend/issues) and found no similar issues.

### Version

main

### What's Wrong?

The following SQL queries cause full table scan on a large table 
```
create or replace table t250212 as select number as id from numbers(1000000);
explain select id from t250212 where id in (select * from unnest(split('1000',',')) as t);

explain with t(id) as (
select * from unnest(split('1000',','))
)
select t1.id from t250212 t1 join t on t1.id= t.id;
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #17448, opened 2025-02-13
fix(query): fold constant subquery to build filter plan instead of join plan

b41sh:

## Summary

Fold a subquery that returns a constant value to a constant value scalar, convert the join plan to a filter plan, which can be filtered using indexes to avoid reading the full amount of data and speed up the query.

for example:
main
```sql
root@0.0.0.0:48000/default> explain select * from numbers(10) as t where t.number in (select * from unnest([1,2]));
HashJoin
├── output columns: [t.number (#0)]
├── join type: LEFT SEMI
├── build keys: [CAST(subquery_2 (#2) AS UInt64 NULL)]
├── probe keys: [CAST(t.number (#0) AS UInt64 NULL)]
├── filters: []
├── estimated rows: 10.00
├── EvalScalar(Build)
│   ├── output columns: [unnest([1, 2]) (#2)]
│   ├── expressions: [get(1)(unnest([1, 2]) (#1))]
│   ├── estimated rows: 3.00
│   └── ProjectSet
│       ├── output columns: [unnest([1, 2]) (#1)]
│       ├── estimated rows: 3.00
│       ├── set returning functions: unnest([1, 2])
│       └── DummyTableScan
└── TableScan(Probe)
    ├── table: default.system.numbers
    ├── output columns: [number (#0)]
    ├── read rows: 10
    ├── read size: < 1 KiB
    ├── partitions total: 1
    ├── partitions scanned: 1
    └── estimated rows: 10.00

25 rows explain in 0.005 sec. Processed 0 rows, 0 B (0 row/s, 0 B/s)
```
this PR
```sql
root@0.0.0.0:48000/default> explain select * from numbers(10) as t where t.number in (select * from unnest([1,2]));
Filter
├── output columns: [t.number (#0)]
├── filters: [(t.number (#0) = 1 OR t.number (#0) = 2)]
├── estimated rows: 0.01
└── TableScan
    ├── table: default.system.numbers
    ├── output columns: [number (#0)]
    ├── read rows: 10
    ├── read size: < 1 KiB
    ├── partitions total: 1
    ├── partitions scanned: 1
    ├── push downs: [filters: [(numbers.number (#0) = 1 OR numbers.number (#0) = 2)], limit: NONE]
    └── estimated rows: 10.00

13 rows explain in 0.007 sec. Processed 0 rows, 0 B (0 row/s, 0 B/s)
```

- part of: #17440

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17448)


src/query/sql/src/planner/optimizer/decorrelate/decorrelate.rs
```diff
@@ -517,4 +523,221 @@ impl SubqueryRewriter {
...
+                    return Ok(None);
+                };
+                match &const_scalar.value {
+                    Scalar::EmptyArray => {
+                        return Ok(Some(ScalarExpr::ConstantExpr(ConstantExpr {
+                            span: subquery.span,
+                            value: Scalar::Null,
+                        })));
+                    }
+                    Scalar::Array(array_column) => {
```
sundy-li, line 689: what if array_column has large length.

Need to respect `max_inlist_to_or` setting

Dousir9 (approved):

Issue #13408, opened 2023-10-24
Feature: The index cannot be matched because the input type is coerced during query


raymondchen-byte:
**Summary**

Define a table field type as int
Query using int::text = '123'
As a result, the query fails to match the index and more files are scanned

BohuTANG:
Could you give the SQL and the errors more?  Thanks

raymondchen-byte:
> 你能提供更多的SQL和错误吗？谢谢
There's nothing wrong with it. It's just an optimization

raymondchen-byte:
explain
select * from wq_app_use.t_order_header where create_time ='2021-01-01 00:00:05';
![image](https://github.com/datafuselabs/databend/assets/12689259/4a026581-4d10-421d-9cb9-166589764814)

explain
select * from wq_app_use.t_order_header where create_time::text = '2021-01-01 00:00:05'::text;
![image](https://github.com/datafuselabs/databend/assets/12689259/9c63c3ab-7fb8-4091-b6ed-890f1fc32b7d)
The same query if the field type packaging, the filter push-down scan of the partition is very different, I think this can be optimized, after all, can not require every warehouse engineer to write sql strictly according to the type

BohuTANG:
> explain select * from wq_app_use.t_order_header where create_time ='2021-01-01 00:00:05'; ![image](https://user-images.githubusercontent.com/12689259/281656476-4a026581-4d10-421d-9cb9-166589764814.png)
> 
> explain select * from wq_app_use.t_order_header where create_time::text = '2021-01-01 00:00:05'::text; ![image](https://user-images.githubusercontent.com/12689259/281656548-9c63c3ab-7fb8-4091-b6ed-890f1fc32b7d.png) The same query if the field type packaging, the filter push-down scan of the partition is very different, I think this can be optimized, after all, can not require every warehouse engineer to write sql strictly according to the type

@xudong963 This maybe a pushdown issue?

sundy-li:
cast will invalidate the range pruner, cause it may destroye the monotonicity or throw exceptions.

forsaken628:
I don't think this needs any more work.

PR #17469, opened 2025-02-17
chore(query): add monotonicity function property

sundy-li:

## Summary

chore(query): add monotonicity function property

part of #13408

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17469)


src/query/expression/src/function.rs
```diff
@@ -578,6 +578,16 @@ impl EvalContext<'_> {
         }
     }
 
+    pub fn has_error(&self, index: usize) -> bool {
+        if self.suppress_error {
+            return false;
+        }
+        match &self.errors {
+            Some((b, _)) => !b.get(index),
```
b41sh, line 586: Need to check if the index is out of bounds, otherwise it may cause a panic.

b41sh (approved):

PR #17503, opened 2025-02-21
feat: Add bendsave which can backup and restore databend data

Xuanwo:

## Summary

This PR will add bendsave: the DR tool for databend that can backup and restore databend data.

The RFC could be seen at: https://docs.databend.com/guides/community/rfcs/disaster-recovery

This PR tested in this way:

- Build a new databend query node with tpch data.
- Perform tpch sqllogictests.
- Run backup.
- Destory all data entirely, remove the metasrv's data.
- Run restore.
- Perform tpch sqllogictests again.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17503)


drmingdrmer:
> * Build a new databend query node with tpch data.
> * Perform tpch sqllogictests.
> * Run backup.
> * Distory all data entirely, remove the metasrv's data.
> * Run restore.
> * Perform tpch sqllogictests again.

Before conducting a thorough review, I have a question:
Is the second pass of SQL logic tests identical to the first pass? 
Does it focus on reading the restored data or executing a new test on a recovered metadata/data set?

Xuanwo:
> Is the second pass of SQL logic tests identical to the first pass?

Yes, exactly the same.

> Does it focus on reading the restored data or executing a new test on a recovered metadata/data set?

I'm trying to run a new test on a recovered metadata/data set to ensure that all the data has been restored correctly.

I chose the TPC-H test for this since we are only executing queries and not inserting new data or performing DML operations, which do not need to be tested again.

.cargo/config.toml
```diff
@@ -4,4 +4,4 @@ JEMALLOC_SYS_WITH_MALLOC_CONF = "percpu_arena:percpu,oversize_threshold:0,backgr
 
 ## If you are using aarch64-apple-darwin, you may need to set the following:
 ## JEMALLOC_SYS_WITH_LG_PAGE = "14"
-## JEMALLOC_SYS_WITH_MALLOC_CONF = "oversize_threshold:0,dirty_decay_ms:5000,muzzy_decay_ms:5000"
+## JEMALLOC_SYS_WITH_MALLOC_CONF = "oversize_threshold:0,dirty_decay_ms:5000,muzzy_decay_ms:5000"
```
sundy-li, line 7: Is this change ok?

src/bendsave/src/backup.rs
```diff
@@ -0,0 +1,78 @@
...
+            .into_bytes_stream(..)
+            .await?
+            .map_ok(Buffer::from)
+            .map_err(|err| anyhow!("read databend query data failed: {err:?}"));
+
+        let mut file = efs.create_file(entry.path()).await?;
+        file.sink(stream).await?;
+        file.commit().await?;
+        info!("databend query file {} has been backed up", entry.path());
+    }
```
dantengsky, line 75: Seems the backup process is "append-only"? How to clean up the objects that have been deleted?

src/bendsave/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+use databend_bendsave::restore;
+use logforth::append;
+use logforth::filter::EnvFilter;
+use logforth::Dispatch;
+use logforth::Logger;
+
+#[derive(Parser)]
+#[command(name = "bendsave")]
+#[command(about = "Databend backup and restore tool", long_about = None)]
+struct Cli {
```
sundy-li, line 28: Can we add backup/restore subcommand in databend-query binary ?

Thus, we don't need to add an extra binary in release tars.

.cargo/config.toml
Xuanwo: Will remove.

src/bendsave/src/backup.rs
Xuanwo: We will need to impement `vacuum` for this, not implemented so far.

src/bendsave/src/main.rs
Xuanwo: Possibly. But databend-bendsave will be released under an elastic License and will require a license key to run, so I assume it will need to be used as a separate CLI.

drmingdrmer (approved):
> Reviewed 24 of 24 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17503)* status: all files reviewed, 3 unresolved discussions (waiting on @dantengsky, @everpcpc, and @sundy-li)

drmingdrmer:
An excellent PR! Clean and concise!

dantengsky:
Looks good to me.

As we continue improving this feature, here are a few considerations:

Given that the table data could grow quite large, we should probably give some extra thought to the efficiency of backup and restore operations. If I understand correctly, the current approach involves periodically backing up data from A to B, then restoring it to C, with B potentially storing multiple versions.

Even if we avoid redundant data objects across those versions at B, this could still lead to significant additional storage costs. The efficiency of restoring from B to C might also negatively impact the RTO.

Additionally, since the backed-up data already includes time travel data, this might raise questions about the necessity of multi-version backups. The vacuum operation on backups could introduce complexity in both implementation and maintenance.

Xuanwo:
All CI passed. I'm going to merge this PR now.

PR #17519, opened 2025-02-25
feat: support `WITHIN GROUP` clause for `AggregateFunction`

KKould:

## Summary

```sql
[ WITHIN GROUP ( <orderby_clause> ) ]
```
```sql
[ WITHIN GROUP ( <orderby_clause> ) ]
    OVER ( [ PARTITION BY <expr2> ] )
```

https://github.com/databendlabs/databend/issues/16472

```sql
select listagg(concat('Hello ', s), '|') within group (order by y + 1 desc, x desc) from t3;

Hello xyz|Hello abc|Hello def

SELECT listagg(cast(salary as varchar), '|') WITHIN GROUP (ORDER BY empno DESC) OVER (PARTITION BY depname ORDER BY empno) FROM empsalary;

3500|3900
3900
4200
4500|6000|4200
4800|4800|5000
4800|5000
5000
5200|4500|6000|4200
5200|5200|4500|6000|4200
6000|4200
```

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17519)


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -1734,16 +1752,45 @@ impl<'a> TypeChecker<'a> {
...
+                        .position(|arg_expr| arg_expr == &scalar_expr)
+                        .ok_or(
+                            ErrorCode::SemanticError(
+                                "the expression in order by must be one of the function parameters",
+                            )
+                            .set_span(expr.span()),
+                        )?;
+
+                    Ok(AggregateFunctionSortDesc {
+                        column: pos,
```
b41sh, line 1775: Don't use column position to identify the sort column here, use `ScalarExpr` instead, and rewrite as a column in `AggregateRewriter` like arguments. The columns of order by can also be other non-argument expressions.

src/query/functions/src/aggregates/aggregate_string_agg.rs
```diff
@@ -44,6 +49,51 @@ pub struct StringAggState {
...
+        self.inner.add_batch(column, validity)
+    }
+
+    fn merge(&mut self, rhs: &Self) -> Result<()> {
+        self.inner.merge(&rhs.inner)
+    }
+
+    fn merge_result(&mut self, builder: &mut ColumnBuilder) -> Result<()> {
+        if let Some(opts) = self.sort_desc {
+            self.inner.values.sort_by(|a, b| {
```
b41sh, line 78: better to build a `DataBlock` and use `sort` function to do the order by.
https://github.com/databendlabs/databend/blob/main/src/query/expression/src/kernels/sort.rs#L77

src/query/functions/src/aggregates/aggregate_array_agg.rs
```diff
@@ -417,6 +418,7 @@ pub fn try_create_aggregate_array_agg_function(
     display_name: &str,
     _params: Vec<Scalar>,
     argument_types: Vec<DataType>,
+    _sort_descs: Vec<AggregateFunctionSortDesc>,
```
b41sh, line 421: We can also implement within group for `array_agg` function, which is similar to `string_agg` function.

src/query/functions/src/aggregates/aggregate_function_factory.rs
```diff
@@ -117,6 +128,34 @@ impl AggregateFunctionDescription {
...
+    Copy,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    Debug,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct AggregateFunctionSortDesc {
```
b41sh, line 143: May be we can use `SortItem` instead this struct.

src/query/sql/src/planner/semantic/type_check.rs
KKould: I tried to replace it with `Expr`, because the introduction of `ScalarExpr` crate will cause circular dependency. In the implementation I am trying to implement, the tuple type is passed into the sort state, and the columns after the first column in the tuple are all order by columns and `DataBlock::sort` is used. This should be able to implement expressions of non-parameter columns.
```rust
impl SortAggregateStringAggFunction {
    fn accumulate(
        &self,
        place: AggrState,
        columns: InputColumns,
        validity: Option<&Bitmap>,
        _input_rows: usize,
    ) -> Result<()> {
        let mut block =
            DataBlock::new_from_columns(columns.as_slice().map(|s| s.to_vec()).unwrap_or_default());
        if let Some(validity) = validity {
            block = block.filter_with_bitmap(validity)?;
        }
        let context = FunctionContext::default();
        let evaluator = Evaluator::new(&block, &context, &BUILTIN_FUNCTIONS);
        let mut tuples = Vec::with_capacity(self.sort_descs.len() + 1);
        tuples.push(columns[0].clone());
        for desc in self.sort_descs.iter() {
            tuples.push(
                evaluator
                    .run(&desc.expr)?
                    .convert_to_full_column(desc.expr.data_type(), block.num_rows()),
            )
        }
        let state = place.get::<SortStringAggState>();
        state.add_batch(&Column::Tuple(tuples), validity)?;

        Ok(())
    }
}

impl SortStringAggState {
    fn merge_result(&mut self, result_builder: &mut ColumnBuilder) -> Result<()> {
        let values = &self.inner.values;
        let mut builders = self
            .sort_descs
            .iter()
            .map(|desc| ColumnBuilder::with_capacity(&desc.ty, values.len()))
            .collect_vec();
        for tuple in values.iter() {
            if let Some(values) = tuple.as_tuple() {
                for (i, value) in values.iter().enumerate() {
                    builders[i].push(value.as_ref());
                }
            }
        }
        let columns = builders
            .into_iter()
            .map(|builder| BlockEntry::new(builder.data_type(), Value::Column(builder.build())))
            .collect_vec();
        let mut block = DataBlock::new(columns, values.len());
        if block.columns().len() > 1 {
            let sort_descs = (1..block.columns().len())
                .into_iter()
                .map(|offset| {
                    let desc = &self.sort_descs[offset];
                    SortColumnDescription {
                        offset: 0,
                        asc: desc.asc,
                        nulls_first: desc.null_first,
                    }
                })
                .collect_vec();

            block = DataBlock::sort(&block, &sort_descs, None)?;
        }
        let builder = StringType::try_downcast_builder(result_builder).unwrap();
        if !block.is_empty() {
            let string = (0..block.num_rows())
                .into_iter()
                .flat_map(|i| block.value_at(0, i))
                .join(&self.delimiter);
            builder.put_and_commit(string);
        } else {
            builder.put_and_commit("");
        }
        Ok(())
    }
}
```

But the problem I am facing now is that Expr cannot support `serde::Serialize, serde::Deserialize`, this is the problem I am facing now, do you have any suggestions for this?

src/query/functions/src/aggregates/aggregate_function_factory.rs
KKould: I tried using `SortItem`, but that requires introducing `databend-common-sql` which results in circular dependencies (like `ScalarExpr`)

b41sh: ok

src/query/functions/src/aggregates/aggregate_string_agg.rs
```diff
@@ -44,6 +55,115 @@ pub struct StringAggState {
...
+        builders.push(ColumnBuilder::with_capacity(
+            &Box::new(DataType::String),
+            values.len(),
+        ));
+        builders.push(ColumnBuilder::with_capacity(
+            &Box::new(DataType::String),
+            values.len(),
+        ));
+        for desc in self.sort_descs.iter() {
+            builders.push(ColumnBuilder::with_capacity(&desc.data_type, values.len()));
```
b41sh, line 116: Aggregate function will remove `NULL` value automatically, if there is a `NULL` value in the sorted column, the row will be removed, resulting in incorrect result. In this case, we need to add the function name to `NEED_NULL_AGGREGATE_FUNCTIONS` and manually remove the `NULL` value when merging the results.
https://github.com/databendlabs/databend/blob/main/src/query/functions/src/aggregates/aggregate_function_factory.rs#L30

src/query/functions/src/aggregates/adaptors/aggregate_sort_adaptor.rs
```diff
@@ -0,0 +1,275 @@
...
+        validity: Option<&Bitmap>,
+        input_rows: usize,
+    ) -> Result<()> {
+        let state = Self::get_state(place);
+        Self::init_columns(columns, state, Some(input_rows));
+
+        match validity {
+            Some(validity) => {
+                for (i, column) in columns.iter().enumerate() {
+                    column.iter().zip(validity.iter()).for_each(|(v, b)| {
```
b41sh, line 91: If the order by column and the argument column are the same, we don't need to store data for both columns, we need to remove duplicate args in `InputColumns::Block`
for example `select array_agg(a) within group (order by a asc nulls first) from t;`

b41sh (approved):

sundy-li:
`list_agg` or `listagg` ?

b41sh:
> `list_agg` or `listagg` ?

[snowflake](https://docs.snowflake.com/en/sql-reference/functions/listagg) and [redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_LISTAGG.html) is `listagg`, we'd better be compatible with them.

BohuTANG:
@soyeric128 Please help to add this function: `listagg` to documention, just find this is useful and I forgot it.

PR #17533, opened 2025-02-27
feat(query): bloom index support prune expr with simple casts

forsaken628:

## Summary

For any injective function y = f(x), the `eq(column, inverse_f(const))` is a necessary condition for `eq(f(column), const)`.
For example, the result of `col = '+1'::int` contains `col::string = '+1'`.

part of #13408

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17533)


src/query/expression/src/evaluator.rs
```diff
@@ -2261,6 +2286,135 @@ impl<'a, Index: ColumnIndex> ConstantFolder<'a, Index> {
...
+
+        Some((func_expr, domain))
+    }
+
+    fn is_injective_cast(src: &DataType, _src_domain: &Domain, dest: &DataType) -> bool {
+        if src == dest {
+            return true;
+        }
+
+        // Ignore FunctionDomain::MayThrow
```
sundy-li, line 2386: I did not see any codes to match the comment.

src/query/expression/src/evaluator.rs
```diff
@@ -2261,6 +2286,135 @@ impl<'a, Index: ColumnIndex> ConstantFolder<'a, Index> {
...
+                true
+            }
+
+            (DataType::Number(src), DataType::Number(dest))
+                if src.is_integer() && dest.is_integer() =>
+            {
+                true
+            }
+            (DataType::Number(src), DataType::Decimal(_)) if src.is_integer() => true,
+            (DataType::Number(_), DataType::String) => true,
```
sundy-li, line 2398: need to check that string literal matches `[1-9][0-9]*`

src/query/functions/tests/it/scalars/testdata/fold_for_prune.txt
```diff
@@ -0,0 +1,124 @@
...
+optimized expr : eq<Int16, Int16>(x, 1_i16)
+output type    : Boolean
+output domain  : Some(Boolean(BooleanDomain { has_false: true, has_true: true }))
+
+
+ast            : x::string = '+1'
+raw expr       : eq(CAST(x::Int16 AS String), '+1')
+checked expr   : eq<String, String>(to_string<Int16>(x), "+1")
+input domain  : {0: Number(Int16(SimpleDomain { min: 1, max: 100 }))}
+optimized expr : eq<Int16, Int16>(x, 1_i16)
```
sundy-li, line 50: should be false.

src/query/expression/src/evaluator.rs
forsaken628: This comment means that Ignore FunctionDomain::MayThrow is a precondition for is_injective_cast. It will be refined later.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -395,7 +395,7 @@ impl BloomIndex {
         )?;
 
         let (new_expr, _) =
-            ConstantFolder::fold_with_domain(&expr, &domains, &self.func_ctx, &BUILTIN_FUNCTIONS);
+            ConstantFolder::fold_for_prune(&expr, &domains, &self.func_ctx, &BUILTIN_FUNCTIONS);
```
b41sh, line 398: It looks like `fold_for_prune` is meant to rewrite the equal expression with a cast column function, like `col::dest_type = val` as `col = val::src_type`, which we can rewrite directly inside the `visit_expr_column_eq_constant` function and don't need redefine `fold_for_prune` function. If `val::src_type` fails, the result must be false.

sundy-li:
does this work for range index ?

b41sh (approved):

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -406,6 +355,64 @@ impl BloomIndex {
...
+                        return (id, domain);
+                    }
+                };
+                let domain = data_schema
+                    .column_id_of(&id)
+                    .ok()
+                    .and_then(|col_id| {
+                        column_stats.get(&col_id).map(|stat| {
+                            match Domain::from_min_max(stat.min.clone(), stat.max.clone(), &ty) {
+                                Domain::Nullable(NullableDomain { value, .. }) => {
```
sundy-li, line 387: ty may be not nullable.

sundy-li (approved):

PR #17549, opened 2025-03-04
fix(query): fix incorrect domain folder

sundy-li:

## Summary

fix(query): fix incorrect domain folder for string cast to timestamp/date

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17549)


src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -502,7 +537,37 @@ fn register_number_to_timestamp(registry: &mut FunctionRegistry) {
 fn register_string_to_date(registry: &mut FunctionRegistry) {
     registry.register_passthrough_nullable_1_arg::<StringType, DateType, _, _>(
         "to_date",
-        |_, _| FunctionDomain::MayThrow,
+        |ctx, d| {
```
forsaken628, line 540: Here just check, the maximum and the minimum, what happens if the value is_err in the middle?

sundy-li: I think it will throw errors during runtime ( to_date will process all values of this column).

So it maybe ok to include it in domain calculation?

TCeason: Yes. If is_err return maythrow domain. Real err check in eval logic

src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -502,7 +537,37 @@ fn register_number_to_timestamp(registry: &mut FunctionRegistry) {
...
+            let mut res = Vec::with_capacity(2);
+            let mut is_extended = false;
+            for (i, v) in [&d.min, &max].iter().enumerate() {
+                let mut d = string_to_date(v, &ctx.tz);
+                if d.is_err() && i == 1 && v.len() > 10 {
+                    d = string_to_date(&v[0..10], &ctx.tz);
+                    is_extended = true;
+                }
+
+                if d.is_err() {
```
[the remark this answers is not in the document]
sundy-li: use extend_value   -1 or  +1 to expand the ranges

src/query/functions/src/scalars/timestamp/src/datetime.rs
```diff
@@ -221,7 +223,42 @@ fn register_string_to_timestamp(registry: &mut FunctionRegistry) {
...
+            for (i, v) in [&d.min, &max].iter().enumerate() {
+                let mut extend_num = 0;
+                if i == 1 && d.max.is_none() {
+                    // the max domain is unbounded
+                    res.push(TIMESTAMP_MAX);
+                    break;
+                }
+                let mut d = string_to_timestamp(v, &ctx.tz);
+                // the string max domain maybe truncated into `"2024-09-02 00:0�"`
+                const MAX_LEN: usize = "1000-01-01".len();
```
TCeason, line 238: Why do not use a constant value 10?

sundy-li: it's be compiled into constant number.

TCeason:
In this pr:

Date min value is 0001-01-01 , Datetime min value is 0001-01-01 00:00:00.000000. So maybe the doc need to modify cc @soyeric128 

```rust
/// 0001-01-01
pub const DATE_MIN: i32 = -719162;

/// Minimum valid timestamp `0001-01-01 00:00:00.000000`, represented by the microsecs offset from 1970-01-01.
pub const TIMESTAMP_MIN: i64 = -62135596800000000;

```

TCeason (approved):

PR #17555, opened 2025-03-05
feat(query): support nondeterministic update

forsaken628:

## Summary

Added the `error_on_nondeterministic_update` setting. Specifies whether to return an error when the UPDATE command is used to update a target row that joins multiple source rows and the system cannot determine the action to perform on the target row.

**Effects:**
- When enabled: An error is returned that includes values from one of the target rows that caused the error.
- When disabled: No error is returned and the update completes, but the results of the update are nondeterministic.

For compatibility with previous behavior, the default enabled

**Implementation modalities:**
support nondeterministic update by inserting an aggregate on top of a join

**Example:**
```sql
CREATE OR REPLACE TABLE target (
    id INT,
    price DECIMAL(10, 2),
    stock INT
);

INSERT INTO target VALUES
(1, 299.99, 50),
(2, 399.99, 30),
(3, 199.99, 20),
(4, 89.99, 100);

CREATE OR REPLACE TABLE source (
    id INT,
    price DECIMAL(10, 2),
    stock INT
);

INSERT INTO source VALUES
(1, 279.99, 55),
(2, 399.99, 25),
(3, 189.99, 22),
(4, 99.99, 50),
(4, 39.99, 60);

-- This update fails when error_on_nondeterministic_update is enabled.
UPDATE target
SET target.price = source.price
FROM source
WHERE target.id = source.id;

SELECT * FROM target ORDER BY id;
----
┌──────────────────────────────────────────────────────────────┐
│        id       │           price          │      stock      │
│ Nullable(Int32) │ Nullable(Decimal(10, 2)) │ Nullable(Int32) │
├─────────────────┼──────────────────────────┼─────────────────┤
│               1 │ 279.99                   │              50 │
│               2 │ 399.99                   │              30 │
│               3 │ 189.99                   │              20 │
│               4 │ 99.99                    │             100 │
└──────────────────────────────────────────────────────────────┘
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17555)


BohuTANG:
@forsaken628 

👍 
But we need more PR summary,  like this:

```
Added [settings name] that [brief purpose].

**Effects:**
- When enabled: [what happens]
- When disabled: [what happens]

**Example:**
```

src/query/settings/src/settings_default.rs
```diff
@@ -1214,6 +1214,13 @@ impl DefaultSettings {
                     scope: SettingScope::Both,
                     range: Some(SettingRange::Numeric(0..=1)),
                 }),
+                ("error_on_nondeterministic_update", DefaultSettingValue {
+                    value: UserSettingValue::UInt64(1),
```
BohuTANG, line 1218: Why default value is true

tests/sqllogictests/suites/base/03_common/03_0048_nondeterministic_update.test
```diff
@@ -0,0 +1,62 @@
+statement ok
+set error_on_nondeterministic_update = 0;
```
BohuTANG, line 2: For the test we need two parts:
For the same dataset and same operations, check the differences.

1. error_on_nondeterministic_update is false
2. error_on_nondeterministic_update is true

sundy-li (approved):

PR #17570, opened 2025-03-07
feat: `string_agg` is compatible with multiple types as arguments instead of just String

KKould:

## Summary
`string_agg` will convert argment if it is not of `DataType::String`

```sql
create table t3(x int null, b boolean null);

insert into t3 values
    (3, true),
    (2, false),
    (1, null),
    (0, true);


select string_agg(x + 1, '|') from t3;
----
4|3|2|1

select string_agg(b, '|') from t3;
----
true|false|true

```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17570)

### DIFF
--- a/src/query/functions/src/aggregates/aggregate_string_agg.rs
+++ b/src/query/functions/src/aggregates/aggregate_string_agg.rs
@@ -18,7 +18,6 @@ use std::sync::Arc;
 
 use borsh::BorshDeserialize;
 use borsh::BorshSerialize;
-use databend_common_exception::ErrorCode;
 use databend_common_exception::Result;
 use databend_common_expression::types::Bitmap;
 use databend_common_expression::types::DataType;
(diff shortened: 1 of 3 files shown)


src/query/functions/src/aggregates/aggregate_string_agg.rs
```diff
@@ -190,11 +217,21 @@ pub fn try_create_aggregate_string_agg_function(
...
+            | DataType::Number(_)
+            | DataType::Decimal(_)
+            | DataType::Timestamp
+            | DataType::Date
+            | DataType::Variant
+            | DataType::Interval
+    ) {
         return Err(ErrorCode::BadDataValueType(format!(
-            "The argument of aggregate function {} must be string",
+            "The argument of aggregate function {} must be one of boolean, string, number, decimal, timestamp, date, varint, interval",
```
b41sh, line 234: `varint` is a typo, we can just return the following message.
```rust
return Err(ErrorCode::BadDataValueType(format!(
    "{} does not support type '{:?}'",
    display_name, arguments[0]
))),
```

b41sh (approved):

PR #17580, opened 2025-03-10
chore(storage): add block file size

zhyass:

## Summary

Introduces a new option, `file_size`, which allows to specify the desired size of Parquet files when blocks are written. The default value is set to `10M`. This option is primarily used during the `recluster` operation to control the size of the resulting blocks.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17580)

### DIFF
--- a/src/query/ee/src/hilbert_clustering/handler.rs
+++ b/src/query/ee/src/hilbert_clustering/handler.rs
@@ -25,8 +25,6 @@ use databend_common_storages_fuse::pruning::create_segment_location_vector;
 use databend_common_storages_fuse::statistics::reducers::merge_statistics_mut;
 use databend_common_storages_fuse::FuseTable;
 use databend_common_storages_fuse::SegmentLocation;
-use databend_common_storages_fuse::DEFAULT_BLOCK_PER_SEGMENT;
-use databend_common_storages_fuse::FUSE_OPT_KEY_BLOCK_PER_SEGMENT;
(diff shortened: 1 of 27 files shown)


BohuTANG:
We need unit/logic tests for this PR. Our team is currently adding more tests to previously untested code sections.

zhyass:
> We need unit/logic tests for this PR(No tests related to this PR?). Our team is currently adding more tests to previously untested code sections.

Added

src/query/ee/src/hilbert_clustering/handler.rs
```diff
@@ -98,7 +100,7 @@ impl HilbertClusteringHandler for RealHilbertClusteringHandler {
         }
 
         let rows_per_block =
-            block_thresholds.calc_rows_per_block(checker.total_size, checker.total_rows) as u64;
+            block_thresholds.calc_rows_per_block(checker.total_size, checker.total_rows, 0) as u64;
```
dantengsky, line 103: Why passes 0 as `calc_rows_per_block` here?

zhyass: The code here will be refactored. calc_rows_per_block takes the maximum of bytes and compressed, so setting it to 0 won't have any effect.

PR #17596, opened 2025-03-12
fix(query): /v1/roles return all roles for account admin

everpcpc:

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17596)

### DIFF
--- a/src/common/io/src/constants.rs
+++ b/src/common/io/src/constants.rs
@@ -26,12 +26,16 @@ pub const INF_BYTES_LONG: &str = "Infinity";
 
 // The size of the I/O read/write block buffer by default.
 pub const DEFAULT_BLOCK_BUFFER_SIZE: usize = 100 * 1024 * 1024;
+// The size of the block compressed by default.
+pub const DEFAULT_BLOCK_COMPRESSED_SIZE: usize = 10 * 1024 * 1024;
 // The size of the I/O read/write block index buffer by default.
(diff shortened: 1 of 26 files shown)


src/query/service/src/servers/http/v1/roles.rs
```diff
@@ -46,14 +47,31 @@ async fn handle(ctx: &HttpQueryContext) -> Result<ListRolesResponse> {
...
     let mut roles = vec![];
-    for role in user.grants.roles() {
-        let is_current = role == current_role;
-        let is_default = role == default_role;
-        roles.push(RoleInfo {
-            name: role.clone(),
-            is_current,
-            is_default,
-        });
+    if user.is_account_admin() {
```
TCeason, line 50: How to test? And why we need this? Please provide summary

everpcpc: We are replacing `show roles` in manage mode with `/v1/roles`, and just keeping them consistent.

TCeason: Got cc @flaneur2020 please review this pr

PR #17603, opened 2025-03-13
chore(query): optimize show tables from iceberg catalog

TCeason:

## Summary

optimize show tables from iceberg catalog

add new API list_tables_names return Vec<String>

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17603)

### DIFF
--- a/src/query/storages/system/src/tables_table.rs
+++ b/src/query/storages/system/src/tables_table.rs
@@ -42,6 +42,7 @@ use databend_common_functions::BUILTIN_FUNCTIONS;
 use databend_common_management::RoleApi;
 use databend_common_meta_app::principal::OwnershipObject;
 use databend_common_meta_app::schema::database_name_ident::DatabaseNameIdent;
+use databend_common_meta_app::schema::CatalogType;
 use databend_common_meta_app::schema::TableIdent;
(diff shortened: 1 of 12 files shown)


src/query/storages/system/src/tables_table.rs
```diff
@@ -124,14 +126,139 @@ where TablesTable<WITH_HISTORY, WITHOUT_VIEW>: HistoryAware
...
+                                catalog_name = catalog.to_string();
+                            }
+                        }
+                    }
+                    Ok(())
+                });
+            }
+        };
+
+        let mut only_iceberg_catalog = false;
```
sundy-li, line 173: this should not be a special case.

TCeason: Default catalog should check privileges.

TCeason: And we already disable refresh when query system.tables.

Issue #17620, opened 2025-03-18
bug: cast nested type cause panic


b41sh:
### Search before asking

- [x] I had searched in the [issues](https://github.com/databendlabs/databend/issues) and found no similar issues.

### Version

main

### What's Wrong?

```
panicked at src/common/column/src/bitmap/bitmap_ops.rs:119:5:
assertion `left == right` failed
  left: 5
 right: 11
```

### How to Reproduce?

```sql
root@0.0.0.0:48000/default> create table tt5(a int, b array(int));
root@0.0.0.0:48000/default> insert into tt5 values(1, [1,2,3,4]),(2, null),(3, [5]),(4, [6,7]),(5, null),(6,[]),(7,[8,9,10]),(8,[11]);
root@0.0.0.0:48000/default> select cast(b, array(string)) from tt5 limit 3;
error: APIError: QueryFailed: [1104]assertion `left == right` failed
  left: 5
 right: 11
```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #17622, opened 2025-03-18
fix(query): fix cast nested types cause panic

b41sh:

## Summary

`ArrayColumn` only intercepts offsets when executing the slice function without modifying the underlying values, resulting in a panic when the length of the values is not matched with the offsets when using values directly. this PR changes the fields values and offsets of `ArrayColumn` from public to private, the values inside `ArrayColumn` can be used by calling `underlying_column` and `underlying_offsets` to avoid getting incorrect values.

By the way, the `set variable` syntax has been changed to allow query with parenthesis to be executed correctly.

- fixes: #17620

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17622)

### DIFF
--- a/src/query/formats/src/field_encoder/json.rs
+++ b/src/query/formats/src/field_encoder/json.rs
@@ -138,10 +138,10 @@ impl FieldEncoderJSON {
         let start = unsafe { *column.offsets().get_unchecked(row_index) as usize };
         let end = unsafe { *column.offsets().get_unchecked(row_index + 1) as usize };
         out_buf.push(b'[');
-        let inner = &T::upcast_column(column.values());
+        let inner = &T::upcast_column(column.values().clone());
         for i in start..end {
             if i != start {
-                out_buf.extend_from_slice(b",");
+                out_buf.push(b',');
             }
             self.write_field(inner, i, out_buf);
         }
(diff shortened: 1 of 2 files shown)


sundy-li (approved):

src/query/expression/src/values.rs
```diff
@@ -1156,18 +1156,18 @@ impl Column {
                 })
             }
             Column::Array(col) => {
-                if col.len() == 0 || col.values.len() == 0 {
+                if col.len() == 0 {
```
TCeason, line 1159: No need to check underlying_column().values.len()?

src/query/expression/src/values.rs
```diff
@@ -1156,18 +1156,18 @@ impl Column {
...
                     Domain::Array(None)
                 } else {
-                    let inner_domain = col.values.domain();
+                    let inner_domain = col.underlying_column().domain();
                     Domain::Array(Some(Box::new(inner_domain)))
                 }
             }
             Column::Map(col) => {
-                if col.len() == 0 || col.values.len() == 0 {
+                if col.len() == 0 {
```
TCeason, line 1167: ditto

src/query/expression/src/values.rs
b41sh: `ArrayColumn` only modifies the `offsets` in the slice operation, the `underlying_column` is unchanged, so checking the length of the `underlying_column` may give a wrong result.

PR #17625, opened 2025-03-18
chore: optimize building bloom index for null

zhyass:

## Summary

1. optimization to the Bloom index construction process. Specifically, it skips building a Bloom index for columns that contain only a single unique value or are entirely composed of NULL values. Because we donot need to generate bloom index, it will never be used since range index is checked first.
2. Calculate the block's memory without convert to full column.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17625)

### DIFF
--- a/tests/suites/5_ee/01_vacuum/01_0002_ee_vacuum_drop_table.sh
+++ b/tests/suites/5_ee/01_vacuum/01_0002_ee_vacuum_drop_table.sh
@@ -119,7 +119,7 @@ echo "select * from test_vacuum_drop_4.b"  | $BENDSQL_CLIENT_CONNECT
 
 ## test vacuum table output
 echo "create table test_vacuum_drop_4.c(c int)" | $BENDSQL_CLIENT_CONNECT
-echo "INSERT INTO test_vacuum_drop_4.c VALUES (1)" | $BENDSQL_CLIENT_OUTPUT_NULL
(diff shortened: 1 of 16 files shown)


tests/suites/5_ee/01_vacuum/01_0002_ee_vacuum_drop_table.sh
```diff
@@ -119,7 +119,7 @@ echo "select * from test_vacuum_drop_4.b"  | $BENDSQL_CLIENT_CONNECT
 
 ## test vacuum table output
 echo "create table test_vacuum_drop_4.c(c int)" | $BENDSQL_CLIENT_CONNECT
-echo "INSERT INTO test_vacuum_drop_4.c VALUES (1)" | $BENDSQL_CLIENT_OUTPUT_NULL
+echo "INSERT INTO test_vacuum_drop_4.c VALUES (1),(2)" | $BENDSQL_CLIENT_OUTPUT_NULL
```
dantengsky, line 122: Failed to understand the purpose of this adjustment, if we don’t insert the extra line here, will this test case fail?

sundy-li (approved):

Issue #17607, opened 2025-03-14
Feature: Add some rewrite rules that `WeTune` discovers


b41sh:
**Summary**

[WeTune](https://ipads.se.sjtu.edu.cn/_media/publications/wtune_extend.pdf) is a rule generator that automatically discovers new rewrite rules. We can add some rewrite rules in Databend.

for example:
```sql
SELECT id FROM notes
WHERE
  type = 'D'
  AND id IN (
    SELECT id FROM notes WHERE commit_id = 7
  )
```
can rewrite as
```sql
SELECT id FROM notes
  notes
WHERE
  type = 'D' AND commit_id = 7
```

```sql
SELECT
  o_auth_applications.id
FROM
  o_auth_applications
  INNER JOIN authorizations ON o_auth_applications.id = authorizations.o_auth_application_id
WHERE
  authorizations.user_id = 1465
```
can rewrite as
```sql
SELECT
  authorizations.o_auth_application_id AS o_auth_application_id
FROM
  authorizations AS authorizations
WHERE
  authorizations.user_id = 1465
```

PR #17628, opened 2025-03-19
feat: impl `Eliminate InSubquery`

KKould:

## Summary

This rule is used to eliminate simple subqueries and move the filter in the subquery to Left.
```sql
select t3.* from t1 as t3 where t3.c = 'D' and t3.a in (select t4.a from t1 as t4 where t4.b = 7);
```
👇👇👇
```sql
select t3.* from t1 as t3 where t3.c = 'D' and t3.b = 7;
```

part of #17607

## Tests

- [ ] Unit Test (TODO)
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17628)


src/query/sql/src/planner/optimizer/rule/rewrite/rule_eliminate_subquery.rs
```diff
@@ -0,0 +1,290 @@
...
+        }
+    }
+}
+
+impl Rule for RuleEliminateSubquery {
+    fn id(&self) -> RuleID {
+        self.id
+    }
+
+    fn apply(&self, s_expr: &SExpr, state: &mut TransformResult) -> Result<()> {
```
BohuTANG, line 73: Better to add some unit tests for this rule apply, the logic test is not enough :/

src/query/sql/src/planner/optimizer/decorrelate/subquery_rewriter.rs
```diff
@@ -453,6 +462,199 @@ impl SubqueryRewriter {
...
+                    .source_table_index(left_source_table_index)
+                    .column_position(left_column.column_position)
+                    .virtual_expr(left_column.virtual_expr)
+                    .build(),
+                );
+            }
+            new_column_bindings
+        };
+        // restore ColumnBinding of same table in Subquery to table in Left
+        let mut body_visitor = SubqueryBodyVisitor::new(new_column_bindings, left.clone());
```
b41sh, line 645: we should rewrite the subquery child plan to replace equivalent columns, not the input `s_expr`.

src/query/sql/src/planner/optimizer/decorrelate/subquery_rewriter.rs
```diff
@@ -452,6 +460,169 @@ impl SubqueryRewriter {
...
+        }
+
+        let right_expr_binding = &subquery.output_column;
+        let (SubqueryType::Any, Some(box ScalarExpr::BoundColumnRef(left_column))) =
+            (&subquery.typ, &subquery.child_expr)
+        else {
+            return Ok(None);
+        };
+        // `xx in null` != `where xx = null`
+        if left_column.column.data_type.is_nullable() || !is_conjunctive_predicate {
```
b41sh, line 553: if data type of left column is nullable, we can still apply this optimization by adding a filter expr `xx is not null`

src/query/sql/src/planner/optimizer/decorrelate/subquery_rewriter.rs
```diff
@@ -452,6 +460,169 @@ impl SubqueryRewriter {
...
+        // restore possible aliases or duplicate loaded tables by `source_table_id` to determine whether they are the same columns of the same table
+        let (Some(left_source_binding), Some(right_source_binding)) = (
+            left_column.column.as_source(),
+            right_expr_binding.as_source(),
+        ) else {
+            return Ok(None);
+        };
+        // when left is a join, filtering becomes complicated. If in subquery is eliminated,
+        // the filter will be pushed down to Scan before join, resulting in inaccurate data results.
+        if left.has_join(Some(&JoinType::Inner)) || left_source_binding != right_source_binding {
```
b41sh, line 571: Even if the left column and the right column are not the same column, we can optimize them by adding the filter `left_col = right_col`.

src/query/sql/src/planner/optimizer/decorrelate/subquery_rewriter.rs
```diff
@@ -452,6 +460,169 @@ impl SubqueryRewriter {
...
+        }
+        let new_column_bindings = {
+            let guard = self.metadata.read();
+
+            let left_columns = guard.columns_by_table_index(left_table_index);
+            let right_columns = guard.columns_by_table_index(right_table_index);
+            let left_table = guard.table(left_table_index);
+            let right_table = guard.table(right_table_index);
+            // filter table function
+            if left_table.database() == "system" || right_table.database() == "system" {
```
b41sh, line 582: we don't need check database is `system`.

src/query/service/tests/it/sql/planner/optimizer/eliminate_in_subquery_test.rs
```diff
@@ -0,0 +1,421 @@
...
+            inverted_indexes: None,
+            attached_columns: None,
+        },
+    ]
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_query_rewrite() -> Result<()> {
+    test_query_rewrite_impl("parquet").await?;
+    test_query_rewrite_impl("native").await
```
b41sh, line 369: don't need test 'native' format, as this feature is only used to optimize the query plan and has nothing to do with the storage layer.

src/query/service/tests/it/sql/planner/optimizer/eliminate_in_subquery_test.rs
```diff
@@ -0,0 +1,421 @@
...
+}
+
+fn get_test_suites() -> Vec<TestSuite> {
+    vec![
+        // Elimination Successful
+        TestSuite {
+            name: "base",
+            query: "SELECT t3.* FROM t1 t3 WHERE t3.b IN (SELECT t4.b FROM t1 t4 WHERE t4.a = 5);",
+            explain: "EvalScalar
+├── scalars: [t3.a (#0) AS (#0), t3.b (#1) AS (#1), t3.c (#2) AS (#2)]
```
b41sh, line 59: those tests seems like explain test, we can write them in this file
`tests/sqllogictests/suites/mode/standalone/explain/subquery.test`
here we can use `Matcher` to check result plan after optimize

Issue #17635, opened 2025-03-21
chore: enable limit total number of log files


BohuTANG:
## Summary 

**Problem**
When using hourly log rotation, the system would count each date separately. For example:
- `databend-query-default.2025-03-18-10.log`
- `databend-query-default.2025-03-18-11.log`
- `databend-query-default.2025-03-19-09.log`
- `databend-query-default.2025-03-19-10.log`

These were treated as different groups, allowing up to 48 files for EACH date. This caused excessive accumulation - in one case, we found 458 log files when the limit was supposed to be 48.

PR #17637, opened 2025-03-21
chore(logger): upgrade logforth to support global max_files limit for log_dir

BohuTANG:

## Summary

**Upgrade logforth with Global File Limit Capability**

Upgrades to logforth from https://github.com/datafuse-extras/logforth (global-max-files-v0.14 branch), which adds the ability to limit the total number of log files in `log_dir` via [PR #1](https://github.com/datafuse-extras/logforth/pull/1).

- fixes: #17635 

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - tests already in datafuse-extras/logforth

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [x] Other (please describe): enhance the log files limitition

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17637)


Xuanwo:
This is a temporary solution, right? I see logforth 0.23 has already supported the max file limit: https://github.com/fast/logforth/blob/d9620447ba9be53a730d8960f103445de79076a9/src/append/rolling_file/rolling.rs#L137-L142

BohuTANG:
> This is a temporary solution, right? I see logforth 0.23 has already supported the max file limit: https://github.com/fast/logforth/blob/d9620447ba9be53a730d8960f103445de79076a9/src/append/rolling_file/rolling.rs#L137-L142

The max_files is for rotation(if hourly the max_files mean not over max_files in a hour) not global, see : https://github.com/fast/logforth/pull/107

wubx:
The log section with: 
```
[log.file]
limit = 12
max-size=1024
format = "text"
include_node_id=true
dir = "/var/log/databend"
level = "WARN,databend=info,opendal=info,openraft=info"
```
Test use
```
for i in `seq 1 1000` ; do echo "select $i"|bendsql ; done
```
"It feels like the number of generated log files is far beyond what was expected."

image

BohuTANG:
cc @wubx 

From my local test, this works as expected.

```
[log.file]
limit = 3
max-size=256
format = "text"
include_node_id=true
dir = "./logs"
level = "INFO"
```

I only tested as:
Case 1: 
The logs have many log files(64) and start the query, the log files reduce to 3.

Case 2:
Do many times like `select sum(numbers) from numbers(10000)`, the logs directory always keep 3 log files.

tisonkun:
From the latest code I don't see the date time being used as part of filename (prefix), then we may not need this workaround now?

See https://github.com/fast/logforth/pull/207.

PR #17641, opened 2025-03-24
fix(query): to_date parse string with format should use UTC timezone

TCeason:

## Summary

to_date parse string with format should use UTC timezone

**Changes:**

- Implemented a conditional check based on the `parse_timestamp` flag to determine whether to parse with timezone awareness.
- When `parse_timestamp` is true the behaviour is as before using the context's timezone.
- When `parse_timestamp` is false, it will parse the timestamp with UTC timezone.

Try to parse string into date in different time zones. Make sure the results are consistent.

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


TCeason:
cc @wubx to_date bug fix in this pr.

sundy-li (approved):

PR #17664, opened 2025-03-27
feat: support specifying compression when unloading to parquet.

youngsofun:

## Summary

before this pr，zstd is used for unloading to parquet files.

this pr add a file format option ` compression = zstd | snappy`  to file format parquet.

unlike the `compression` option of other formats(ndjson/csv):
1. it is used only when unloading, instead of both load and unload
2. is it used for compression of internal block of file, instead of the whole file
3. the default is zstd instead of none

for example

```
copy into @data/unload/ from (select * from numbers(3)) file_format=(type = parquet compression=snappy)
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17664)

### DIFF
--- a/src/meta/app/src/principal/file_format.rs
+++ b/src/meta/app/src/principal/file_format.rs
@@ -109,7 +109,7 @@ impl FileFormatParams {
             FileFormatParams::NdJson(v) => v.compression,
             FileFormatParams::Json(v) => v.compression,
             FileFormatParams::Xml(v) => v.compression,
-            FileFormatParams::Parquet(_) => StageFileCompression::None,
+            FileFormatParams::Parquet(v) => v.compression,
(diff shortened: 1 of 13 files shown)


drmingdrmer:
Is that correct for the same meta-data, the previous version treat it a `None` compression value(for no-compression I think?) and this version consider it as zstd. Then the this version will try to decompress the data in a parquet, which is actually not compressed?

sundy-li (approved):

BohuTANG:
Will merge and release a new build. cc @drmingdrmer @youngsofun

PR #17667, opened 2025-03-28
refactor: use `BlocksSerializer` to replace `StringBlock` to simplify the serialization

KKould:

## Summary

- fixes: https://github.com/databendlabs/databend/issues/17662

when using `StringBlocks` to query large amounts of data, data was frequently materialized and there were meaningless data conversions. For example, the String type was first converted to Bytes and then to String.

This PR uses `BlocksSerializer` to collect `DataBlocks` that need to be serialized. When serde serialization is performed, it will imitate the behavior of Vec for serialization, thus avoiding the need for `DataBlock` to be converted to string line by line in the original `StringBlock` and then collected as `Vec<Vec<Option<String>>>`

When extracting data from a `DataBlock`, `BlocksSerializer` will try to directly extract data from String, Variant, and other types that are Strings, avoiding the meaningless conversion to Bytes in the middle.

Originally, when processing `remaining_size` and `remaining_rows`, the specific size of the data is calculated row by row. However, when the amount of data is large, this will take up some overhead. Therefore, this PR uses the `Proportional Splitting`. Assuming that the data in DataBlock is uniform, `take_rows = (remain_size * block_rows) / block_size` is used to determine how many rows of data can be obtained under the `remaining_size` limit, and then further limit it through `remaining_row`.

test case: 
```sql
CREATE TABLE test AS
SELECT 
    number AS id,
    JSON_OBJECT(
        'id', CAST(number % 100000000 AS STRING),
        'event_type', CONCAT('submit_abcd', CAST(number % 100 AS STRING)),
        'client_code', CONCAT('aaaaaaaaaaaaaaaaaaa', CAST(number % 100 AS STRING)),
        'activity_id', CONCAT('aaaaaaaaaaaaaaaaabbbbbbbbbbbbbbbbbbbbbbbbbccccccccccccccddddddddddddddddeeeeeeeeeeeeeeeee', CAST(number % 100 AS STRING))
    ) AS payload
FROM numbers(50000000);


SELECT * FROM test;
```

Release Mode: 
- Before: 50000000 rows read in 66.891 sec. Processed 50 million rows, 10.19 GiB (747.48 thousand rows/s, 155.92 MiB/s)
- After: 50000000 rows read in 53.671 sec. Processed 50 million rows, 10.19 GiB (931.6 thousand rows/s, 194.33 MiB/s)

e132dab7e5a558b90f124541436ed401

[after.pb.gz](https://github.com/user-attachments/files/19501389/after.pb.gz)
[before.pb.gz](https://github.com/user-attachments/files/19501390/before.pb.gz)

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - There are already tests covering this: `test_max_size_per_page` & `test_max_size_per_page_total_rows`

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17667)


src/query/service/src/servers/http/v1/query/blocks_serializer.rs
```diff
@@ -0,0 +1,214 @@
...
+            }
+        }
+        serialize_seq.end()
+    }
+}
+
+struct RowSerializer<'a> {
+    format: &'a FormatSettings,
+    data_block: &'a [Column],
+    encodeer: &'a FieldEncoderValues,
```
youngsofun, line 116: “encoder” is typo？

src/query/service/src/servers/http/v1/query/page_manager.rs
```diff
@@ -124,65 +129,87 @@ impl PageManager {
...
+        serializer: &mut BlocksSerializer,
         block: DataBlock,
         remain_rows: &mut usize,
         remain_size: &mut usize,
     ) -> Result<()> {
-        let format_settings = {
+        if block.is_empty() {
+            return Ok(());
+        }
+        if !serializer.has_format() {
```
youngsofun, line 140: no need to check，since always init with `BlocksSerializer::new(self.format_settings.read().clone())`?

src/query/service/src/servers/http/v1/query/blocks_serializer.rs
```diff
@@ -0,0 +1,133 @@
...
+        Column::Null { .. } => true,
+        Column::Nullable(box inner) => !inner.validity.get_bit(row_index),
+        _ => false,
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct BlocksSerializer {
+    // Vec<Column> for a Block
+    columns: Vec<(Vec<Column>, usize)>,
```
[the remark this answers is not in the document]
youngsofun: what about define a local ` type Block = (Vec<Column>, usize)`?

youngsofun (approved):

src/query/service/src/servers/http/v1/query/page_manager.rs
KKould: Some places, such as [`QueryResponse::from_internal`](https://github.com/databendlabs/databend/pull/17667/files#diff-ff304d95292e66b2d2555d20bcc927d0531a74f5e346616b13514d0dfc372177R168), may need an empty `BlocksSerializer` and `FormatSettings` cannot be obtained here, so it is difficult to remove Option here.

sundy-li (approved):

b41sh (approved):

Issue #17563, opened 2025-03-07
Feature: query iceberg table by version


sundy-li:
**Summary**

Description for this feature.

```
select count(t.c) from t12_0004 at (TIMESTAMP => 'xxxx') as t;
select count(t.c) from t12_0004 at (snapshot => 'xxxx') as t
```

PR #17693, opened 2025-04-02
feat(query): support query iceberg table by version

TCeason:

## Summary

```sql
select * from lineitem at (timestamp=>'2025-03-12 23:18:22.665000'::Timestamp) limit 1;

select * from lineitem at (snapshot=>'4524280171191816067') limit 1;
```

Note:

If timestamp is less or bigger than min snapshot ts will return error.

If snapshot id is not exists will return error.

Fix: #17563 
## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17693)

### DIFF
--- a/src/query/storages/iceberg/src/table.rs
+++ b/src/query/storages/iceberg/src/table.rs
@@ -462,6 +462,11 @@ impl Table for IcebergTable {
             }
         };
 
+        if snapshot_id.is_none() {
+            return Err(ErrorCode::TableHistoricalDataNotFound(
+                "No historical data found at given point",
+            ));
+        }
         let (table, statistics) = Self::parse_engine_options(&self.info.meta.engine_options)?;
         Ok(Arc::new(IcebergTable {
             info: self.info.clone(),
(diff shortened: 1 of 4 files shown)


src/query/storages/iceberg/src/table.rs
```diff
@@ -405,4 +418,56 @@ impl Table for IcebergTable {
...
+                return Err(ErrorCode::Unimplemented(format!(
+                    "Time travel operation is not supported for the table '{}', which uses the '{}' engine.",
+                    self.name(),
+                    self.get_table_info().engine(),
+                )));
+            }
+        };
+
+        let (table, statistics) = Self::parse_engine_options(&self.info.meta.engine_options)?;
+        Ok(Arc::new(IcebergTable {
```
sundy-li, line 466: better to throw an error if no snapshot found, which keep consistent with fuse table

TCeason: Done

sundy-li (approved):

PR #17702, opened 2025-04-03
refactor: add `query_profile` and `query_details` persistent log tables

dqhl76:

## Summary

**This PR included:**

1. **Added persistent log tables `query_profile` and `query_details`**
2. **Refactored code to simplify the process of adding new tables to the persistent_system**
3. **Implemented schema migration support (by renaming and recreation)**

We will have three tables (if enable in config):
```
select * from persistent_system.query_log;
select * from persistent_system.query_profile;
select * from persistent_system.query_details;
```

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17702)


src/query/service/src/persistent_log/global_persistent_log.rs
```diff
@@ -188,38 +221,67 @@ impl GlobalPersistentLog {
...
+            let table_name = table.table_name();
+            let old_table = context
+                .get_table(CATALOG_DEFAULT, "persistent_system", table_name)
+                .await;
+            if old_table.is_ok() {
+                let old_schema = old_table?.schema();
+                if !table.schema_equal(old_schema) {
+                    let rename_target =
+                        format!("`{}_old_{}`", table_name, chrono::Utc::now().timestamp());
+                    let rename = format!(
```
zhang2014, line 243: need lock

dqhl76: Done

src/query/service/src/persistent_log/global_persistent_log.rs
```diff
@@ -188,57 +292,102 @@ impl GlobalPersistentLog {
...
+
+        for table in &self.tables {
+            let session = create_session(&self.tenant_id, &self.cluster_id).await?;
+            let context = session.create_query_context().await?;
+            let table_name = table.table_name();
+            let old_table = context
+                .get_table(CATALOG_DEFAULT, "persistent_system", table_name)
+                .await;
+            if old_table.is_ok() {
+                let old_schema = old_table?.schema();
```
zhang2014, line 310: When we rename a table name, what happens to older versions of `databend-query` that are still running?

dqhl76: The concern makes sense. The older versions will continue to fail during execution because of schema changes.

However, I think the newer version of databend-query can execute successfully, so logs can still be transferred to the table, although the interval between transfers has become longer.

zhang2014 (approved):

PR #17710, opened 2025-04-07
chore(query): Optimize cast expr error display

forsaken628:

## Summary

cast expr is executed recursively, the previous implementation did not pass the outer display name, making the error message hard to read

e.g.
```
select 1 + 'a';"

before
invalid digit found in string while evaluating function `to_uint64('a')` in expr `to_uint64()`, during run expr: `1 + CAST('a' AS UInt64)`

after
invalid digit found in string while evaluating function `to_uint64('a')` in expr `CAST('a' AS UInt64)`, during run expr: `1 + CAST('a' AS UInt64)`

select cast((1::int null,'a') as tuple(string,int not null)).1 +3;

before
invalid digit found in string while evaluating function `to_int32('a')` in expr `to_int32()`, during run expr: `to_uint64(get(1)(CAST((1, 'a') AS Tuple(String, Int32)))) + 3`

after
invalid digit found in string while evaluating function `to_int32('a')` in expr `CAST((1, 'a') AS Tuple(String, Int32))`, during run expr: `CAST(get(1)(CAST((1, 'a') AS Tuple(String, Int32))) AS UInt64) + 3`
```

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17710)


src/query/expression/src/filter/selector.rs
```diff
@@ -549,16 +549,19 @@ impl<'a> Selector<'a> {
                 );
                 let mut eval_options = EvaluateOptions::new(selection);
                 let value = self.evaluator.get_select_child(expr, &mut eval_options)?.0;
+                let expr_display = &expr.sql_display();
```
sundy-li, line 552: can we pass a lambda function to initialize the string?

sundy-li (approved):

PR #17766, opened 2025-04-12
feat(meta-client): add `cache` crate for databend-meta service

drmingdrmer:

##### feat(meta-client): add `cache` crate for databend-meta service

Crate `databend-common-meta-cache` provide a process-wide in memory
cache for databend-meta service.
It stores a continous range of key values locally and watches the
updates on a remote databend-meta and continously apply new changes to
local store.

The cache may be obsolete if the connection to the datbend-meta service
is unstable. Although in the background it keep trying to re-connecting.

This cache requires at least databend-meta 1.2.677; if the server is
older the cache returns a `Unsupported` error when accessing it.

Usage:

```rust
let client = MetaGrpcClient::try_create(/*..*/);
let cache = Cache::new(
        client,
        "your/cache/key/space/in/meta/service",
        "your-app-name-for-logging",
).await;

let value = cache.try_get("key").await?;
```

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Other

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17766)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3594,6 +3594,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "databend-common-meta-cache"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "databend-common-base",
+ "databend-common-meta-client",
+ "databend-common-meta-types",
+ "futures",
+ "log",
+ "pretty_assertions",
+ "thiserror 1.0.65",
+ "tokio",
+ "tonic",
+]
+
+[[package]]
 name = "databend-common-meta-client"
 version = "0.1.0"
 dependencies = [
@@ -4988,6 +5004,7 @@ dependencies = [
  "databend-common-grpc",
  "databend-common-http",
  "databend-common-meta-api",
+ "databend-common-meta-cache",
  "databend-common-meta-client",
  "databend-common-meta-kvapi",
  "databend-common-meta-raft-store",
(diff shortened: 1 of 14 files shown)


BohuTANG:
Can we integrate this into the MetaClient? Like:
```
let client = MetaGrpcClient::try_create(/*..*/).with_cache("your/cache/key/space/in/meta/service","your-app-name-for-logging",);
let value = client.xx().await?; -- using the cache in the client internal function
```

drmingdrmer:
> Can we integrate this into the MetaClient? Like:
> 
> ```
> let client = MetaGrpcClient::try_create(/*..*/).with_cache("your/cache/key/space/in/meta/service","your-app-name-for-logging",);
> let value = client.xx(...).await?; -- using the cache in the client internal function
> ```

I don't want to encourage such usage. While providing a transparent cache layer may seem convenient, it actually hides network layer errors and could significantly complicate troubleshooting.

TCeason:
I see this pr and I think it's a cache impl inside meta. So we can merge this pr. 

When we actually cache some kv with it, we can talk about the actual problem again, right?

drmingdrmer:
> I see this pr and I think it's a cache impl inside meta. So we can merge this pr.
> 
> When we actually cache some kv with it, we can talk about the actual problem again, right?

Yes. Then I'll just merge it.

TCeason (approved):

PR #17771, opened 2025-04-14
feat: hybrid column data cache

dantengsky:

## Summary

- Replace on-disk table data cache with a hybrid cache (in-memory and on-disk).
- Add new cache config option data-cache-in-memory-bytes (default: 0, disabled).
  - Specifies the maximum number of bytes for in-memory data cache.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17771)


SkyFan2002 (approved):

src/query/storages/common/cache/src/cache_items.rs
```diff
@@ -0,0 +1,67 @@
...
+pub use parquet::file::metadata::ParquetMetaData;
+
+use crate::HybridCache;
+
+pub struct ColumnData(Bytes);
+
+impl ColumnData {
+    pub fn from_merge_io_read_result(bytes: Bytes) -> Self {
+        // Bytes are from merge read result, may refer to a large chunk of memory;
+        // Caching this large buffer wastes memory, so we need to copy it out.
```
sundy-li, line 39: I don't get the comments.
If `bytes` are continuous, maybe we don't need to copy it out?

sundy-li: I get it, so bytes are sliced.

src/query/storages/fuse/src/io/read/block/block_reader_merge_io_async.rs
```diff
@@ -112,7 +113,11 @@ impl BlockReader {
                     .owner_memory
                     .get_chunk(*chunk_idx, &merge_io_result.block_path)?;
                 let data = chunk_data.slice(range.clone());
-                column_data_cache.insert(column_cache_key.as_ref().to_owned(), data);
+
+                column_data_cache.insert(
+                    column_cache_key.as_ref().to_owned(),
+                    ColumnData::from_merge_io_read_result(data),
```
sundy-li, line 119: `chunk_data` is read from merge io, it's wasteful to store all chunk_data into memory, but it's ok to store all chunk_data into disk by ranges.  Maybe later we can optimize it.

sundy-li (approved):

Issue #17810, opened 2025-04-18
Feature: create iceberg table with partition and properties


TCeason:
**Summary**

Description for this feature.

PR #17812, opened 2025-04-18
feat(query): support create iceberg table with partition and properties

TCeason:

## Summary

```sql
CREATE TABLE [IF NOT EXISTS] [database.]table_name
(column_definition1[, column_definition2, ...])
[PARTITION BY (par_col1[, par_col2...])]
[PROPERTIES ("key" = "value", ...)]
```

Note:

1. partition key supports data types other than FLOAT, DOUBLE, DECIMAL, and DATETIME, and does not support NULL values.
2. If create table that in default/hive catalog assign partition or PROPERTIES it will be failed
Add new struct TablePartition. Only support identity transform.
3. forbidan alter column name when table engine is iceberg
4. Table PROPERTIES and PARTITION will send to iceberg-rust by CreateTableReq

- fixes: #17810

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)


TCeason:
Rerun ci, wait https://github.com/databendlabs/databend/pull/17813 merge.

tests/sqllogictests/suites/tpch_iceberg/utils.test
```diff
@@ -105,8 +105,11 @@ id INT YES NULL (empty)
 statement ok
 create or replace table abc.t1(c1 int, c2 string);
 
+statement error 1001
+create or replace table abc.t1(c1 datetime, c2 string) partition by (c1, c2);
```
BohuTANG, line 109: If we apply the partition by on the fuse engine table, what's happened? By design, it will return not support error. For the user, they don't know the table is fuse or iceberg, but the capability is not same. We should handle them on the underlying.

TCeason: Oh yes I will add this check

TCeason: Now if create a table in default/hive catalog with a partition or PROPERTIES it will return a definite err.

src/query/service/src/interpreters/interpreter_table_create.rs
```diff
@@ -347,6 +349,18 @@ impl CreateTableInterpreter {
             self.build_request(stat)
         }?;
 
+        let catalog_type = catalog.info().catalog_type();
+
+        if matches!(catalog_type, CatalogType::Default | CatalogType::Hive)
```
sundy-li, line 354:
```
if !catalog.support_partition() {

}
```

src/query/service/src/interpreters/interpreter_table_create.rs
```diff
@@ -347,6 +349,18 @@ impl CreateTableInterpreter {
...
+            && (req.table_meta.iceberg_table_properties.is_some()
+                || req.table_meta.iceberg_partition.is_some())
```
sundy-li, line 356: just use 

```
table_properties and iceberg_partition
```
These keeps the abstraction in table.

TCeason: How about 

iceberg_properties and iceberg_partition?

sundy-li: I don't think it's good to have `iceberg_` prefix.

What if we will integrate with other catalog with partition suppport.

TCeason: Well, rigorous thinking, I'll change it later

TCeason: Done

drmingdrmer (changes requested):
> Reviewed 2 of 47 files at r1, 20 of 49 files at r3.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: 22 of 57 files reviewed, 4 unresolved discussions (waiting on @BohuTANG, @sundy-li, @TCeason, and @Xuanwo)

___
*[`src/meta/app/src/schema/table.rs` line 257 at r3](https://reviewable.io/reviews/databendlabs/databend/17812#-OOFlCboBOQE5zuy7J05:-OOFlCboBOQE5zuy7J06:b-x3fvws) ([raw file](https://github.com/databendlabs/databend/blob/a99a6041f45e15c8e36e6b49529eba8aff188370/src/meta/app/src/schema/table.rs#L257)):*
> ```XML
> #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
> pub enum TablePartition {
>     Identity { columns: Vec<String> },
> ```

Is it safe to use column name as partition keys, while user might rename a column.

TCeason:
> 
> 
> > Reviewed 2 of 47 files at r1, 20 of 49 files at r3.
> > *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: 22 of 57 files reviewed, 4 unresolved discussions (waiting on @BohuTANG, @sundy-li, @TCeason, and @Xuanwo)
> 
> ___
> *[`src/meta/app/src/schema/table.rs` line 257 at r3](https://reviewable.io/reviews/databendlabs/databend/17812#-OOFlCboBOQE5zuy7J05:-OOFlCboBOQE5zuy7J06:b-x3fvws) ([raw file](https://github.com/databendlabs/databend/blob/a99a6041f45e15c8e36e6b49529eba8aff188370/src/meta/app/src/schema/table.rs#L257)):*
> > ```XML
> > #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Eq, PartialEq)]
> > pub enum TablePartition {
> >     Identity { columns: Vec<String> },
> > ```
> 
> Is it safe to use column name as partition keys, while user might rename a column.
> 
> 
> 
> 

A good issue.

I think we should forbidden iceberg table alter and return a definite error
. Like view and stream engine 

https://github.com/databendlabs/databend/blob/main/src/query/service/src/interpreters/interpreter_table_rename_column.rs#L76

Xuanwo:
Hi, I’m a bit concerned that we’re planning to store this data in metasrv. Are we intending to use metasrv as an iceberg catalog, or will we support creating Iceberg tables as internal tables, similar to what the Fuse engine does?

TCeason:
> Hi, I’m a bit concerned that we’re planning to store this data in metasrv. Are we intending to use metasrv as an iceberg catalog, or will we support creating iceberg tables as internal tables, similar to what the Fuse engine does?

Actually we can add the partition info in req not tablemeta.

If we do not store this information, we must load table when we use it later

Xuanwo:
> Actually we can add the partition info in req not tablemeta.
> 
> If we do not store this information, we must load table when we use it later

It depends on our design and our objectives. Typically, we can't assume that we are the only ones writing to or reading from an iceberg table. Users can perform various actions on it, such as renaming column names with Spark. While we can cache this metadata, we can't persist it.

TCeason:
> > Actually we can add the partition info in req not tablemeta.
> 
> > 
> 
> > If we do not store this information, we must load table when we use it later
> 
> 
> 
> It depends on our design and our objectives. Typically, we can't assume that we are the only ones writing to or reading from an iceberg table. Users can perform various actions on it, such as renaming column names with Spark. While we can cache this metadata, we can't persist it.

If so it's better to add these info in req. maybe we need to cost some time and discuss it tomorrow

drmingdrmer (commented):
> Reviewed 1 of 47 files at r1, 1 of 38 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: 2 of 66 files reviewed, 4 unresolved discussions (waiting on @BohuTANG, @sundy-li, @TCeason, and @Xuanwo)

___
*[`src/meta/app/src/schema/table.rs` line 257 at r3](https://reviewable.io/reviews/databendlabs/databend/17812#-OOFlCboBOQE5zuy7J05:-OOMI9Ao0Lwag3pH3rG-:b-wkjw3x) ([raw file](https://github.com/databendlabs/databend/blob/a99a6041f45e15c8e36e6b49529eba8aff188370/src/meta/app/src/schema/table.rs#L257)):*
<details><summary><i>Previously, TCeason wrote…</i></summary><blockquote>

A good issue.

I think we should forbidden iceberg table rename col name and return a definite error
. Like view and stream engine 

https://github.com/databendlabs/databend/blob/main/src/query/service/src/interpreters/interpreter_table_rename_column.rs#L76 

</blockquote></details>

Then how to deal with the column rename? :thinking:

TCeason:
> 4\. PROPERTIES and PARTITION

For write : We set the partition in Req not store these info.
For read: We load the data in iceberg catalog

drmingdrmer (changes requested):
> Reviewed 3 of 47 files at r1, 1 of 25 files at r2, 23 of 52 files at r5, 37 of 38 files at r6.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: all files reviewed, 8 unresolved discussions (waiting on @BohuTANG, @sundy-li, @TCeason, and @Xuanwo)

___
*[`src/query/storages/iceberg/src/database.rs` line 201 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMUMZO15Q5FUWjNjhC:-OOMUMZO15Q5FUWjNjhD:b-lf5iz4) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/query/storages/iceberg/src/database.rs#L201)):*
> ```XML
>         };
> 
>         let table_create_option = if let Some(ref partition) = req.table_partition {
> ```

The deep nesting since this line makes the logic hard to understand. It would be better extract these part into another method and flatten the nesting logic. 

BTW, the `builder` should be created only once.
___
*[`src/query/service/src/interpreters/interpreter_table_create.rs` line 497 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMV25k-rZCwRhlF2sh:-OOMV25k-rZCwRhlF2si:b47jgip) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/query/service/src/interpreters/interpreter_table_create.rs#L497)):*
> ```XML
>                     columns: self.plan.table_partition.clone(),
>                 })
>             },
> ```

Use uniformed type for property and partition.

All types should be defined with `Option`, or none types with `Option`.



_Code quote:_
```XML
            table_properties: if self.plan.table_properties.is_empty() {
                None
            } else {
                Some(self.plan.table_properties.clone())
            },
            table_partition: if self.plan.table_partition.is_empty() {
                None
            } else {
                Some(TablePartition::Identity {
                    columns: self.plan.table_partition.clone(),
                })
            },
```
___
*[`src/query/sql/src/planner/binder/ddl/table.rs` line 489 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMVbJ91bcpQEuvtco0:-OOMVbJ91bcpQEuvtco1:b-4b4wl5) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/query/sql/src/planner/binder/ddl/table.rs#L489)):*
> ```XML
>                 table_partition.push(iceberg_table_partition.to_string());
>             }
>         }
> ```

`iceberg_table_partition.into_iter().collect()` would be good enough to convert a BTreeMap to Vec?



_Code quote:_
```XML
        let mut table_partition: Vec<String> = Vec::new();
        if let Some(iceberg_table_partition) = iceberg_table_partition {
            for iceberg_table_partition in iceberg_table_partition {
                table_partition.push(iceberg_table_partition.to_string());
            }
        }
```
___
*[`src/meta/app/src/schema/table.rs` line 441 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMLOzl1jzU5mfKLeLm:-OOMLOzl1jzU5mfKLeLn:b-g5tvw1) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/meta/app/src/schema/table.rs#L441)):*
> ```XML
>         TablePartition::Identity { columns: vec![] }
>     }
> }
> ```

Why not derive Default for it?



_Code quote:_
```XML
impl Default for TablePartition {
    fn default() -> Self {
        TablePartition::Identity { columns: vec![] }
    }
}
```
___
*[`src/query/sql/src/planner/plans/ddl/table.rs` line 56 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMMclZACq_LUZlGs_G:-OOMMclZACq_LUZlGs_H:bxpvaaf) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/query/sql/src/planner/plans/ddl/table.rs#L56)):*
> ```XML
>     pub options: TableOptions,
>     pub table_properties: TableOptions,
>     pub table_partition: Vec<String>,
> ```

If partitioning is optional, these two field should be optional.

If the default value, such as `vec![]` indicating no-partitioning, everywhere else should use `vec![]` to represent a no-partitioning decision.

To me, `Option<Vec<String>>` would be more accurate about the purpose.



_Code quote:_
```XML
    pub table_properties: TableOptions,
    pub table_partition: Vec<String>,
```

TCeason (commented):
> Dismissed @drmingdrmer from a discussion.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: 50 of 66 files reviewed, 7 unresolved discussions (waiting on @BohuTANG, @drmingdrmer, @sundy-li, and @Xuanwo)

___
*[`src/meta/app/src/schema/table.rs` line 441 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMLOzl1jzU5mfKLeLm:-OON19_7BdPlPyMldK44:bvo3vqa) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/meta/app/src/schema/table.rs#L441)):*

Why not derive Default for it?
</blockquote></details>

Maybe support other type in the future. So I impl a Default.
___
*[`src/query/storages/iceberg/src/database.rs` line 201 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMUMZO15Q5FUWjNjhC:-OON11Ub7R_ietxLPf17:bb6un1j) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/query/storages/iceberg/src/database.rs#L201)):*

The deep nesting since this line makes the logic hard to understand. It would be better extract these part into another method and flatten the nesting logic. 

BTW, the `builder` should be created only once.
</blockquote></details>

Done

TCeason (commented):
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: 50 of 66 files reviewed, 7 unresolved discussions (waiting on @BohuTANG, @drmingdrmer, @sundy-li, and @Xuanwo)

___
*[`src/query/sql/src/planner/binder/ddl/table.rs` line 489 at r6](https://reviewable.io/reviews/databendlabs/databend/17812#-OOMVbJ91bcpQEuvtco0:-OON1LT01-ft7u_0nqkG:b-896fix) ([raw file](https://github.com/databendlabs/databend/blob/a4dcf718cd44cffd36d185630944f8a2150762c4/src/query/sql/src/planner/binder/ddl/table.rs#L489)):*

`iceberg_table_partition.into_iter().collect()` would be good enough to convert a BTreeMap to Vec?
</blockquote></details>

Done.

sundy-li (approved):

drmingdrmer (changes requested):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: all files reviewed, 7 unresolved discussions (waiting on @BohuTANG, @sundy-li, @TCeason, and @Xuanwo)

___
*[`src/query/storages/iceberg/src/database.rs` line 308 at r9](https://reviewable.io/reviews/databendlabs/databend/17812#-OOVVutJ0XUBISpIEHHO:-OOVVutJ0XUBISpIEHHP:bazvpdd) ([raw file](https://github.com/databendlabs/databend/blob/0ed64474dac38721b0db732af8fd9afabfba5cbb/src/query/storages/iceberg/src/database.rs#L308)):*
> ```XML
>             "Partition key can not empty.".to_string(),
>         ));
>     }
> ```

Is that illegal if a partition does not contain any column?

It seems to be the same as **no partition** policy and should be valid IMHO.



_Code quote:_
```XML
    if columns.is_empty() {
        return Err(ErrorCode::Internal(
            "Partition key can not empty.".to_string(),
        ));
    }
```
___
*[`src/query/storages/iceberg/src/database.rs` line 326 at r9](https://reviewable.io/reviews/databendlabs/databend/17812#-OOV_ZcE8AvmHq5SQXTO:-OOV_ZcE8AvmHq5SQXTP:b-efj4vi) ([raw file](https://github.com/databendlabs/databend/blob/0ed64474dac38721b0db732af8fd9afabfba5cbb/src/query/storages/iceberg/src/database.rs#L326)):*
> ```XML
>                 source_id: field.id,
>                 name: field.name.to_string(),
>                 field_id: Some(i as i32),
> ```

Is it safe to use a transient index value as `field_id`?
___
*[`src/query/storages/iceberg/src/database.rs` line 360 at r9](https://reviewable.io/reviews/databendlabs/databend/17812#-OOVV_5RCK1GLAVK_Obu:-OOVV_5RCK1GLAVK_Obv:beo9soj) ([raw file](https://github.com/databendlabs/databend/blob/0ed64474dac38721b0db732af8fd9afabfba5cbb/src/query/storages/iceberg/src/database.rs#L360)):*
> ```XML
>                 if columns.is_empty() {
>                     return Ok(builder.build());
>                 }
> ```

Is this `is_empty()` necessary? I thought the result of this function would be the same with or without this check.



_Code quote:_
```XML
                if columns.is_empty() {
                    return Ok(builder.build());
                }
```

TCeason (commented):
> *[Reviewable](https://reviewable.io/reviews/databendlabs/databend/17812)* status: all files reviewed, 7 unresolved discussions (waiting on @BohuTANG, @drmingdrmer, @sundy-li, and @Xuanwo)

___
*[`src/query/storages/iceberg/src/database.rs` line 326 at r9](https://reviewable.io/reviews/databendlabs/databend/17812#-OOV_ZcE8AvmHq5SQXTO:-OOVc_aECg76bxFIvxhr:biiphs5) ([raw file](https://github.com/databendlabs/databend/blob/0ed64474dac38721b0db732af8fd9afabfba5cbb/src/query/storages/iceberg/src/database.rs#L326)):*

Is it safe to use a transient index value as `field_id`?
</blockquote></details>

Yes, i is usize . 
___
*[`src/query/storages/iceberg/src/database.rs` line 360 at r9](https://reviewable.io/reviews/databendlabs/databend/17812#-OOVV_5RCK1GLAVK_Obu:-OOVd9eX6JNQSQLsNGAC:b-k93i1u) ([raw file](https://github.com/databendlabs/databend/blob/0ed64474dac38721b0db732af8fd9afabfba5cbb/src/query/storages/iceberg/src/database.rs#L360)):*

Is this `is_empty()` necessary? I thought the result of this function would be the same with or without this check.
</blockquote></details>

This should be return error . I will directly call build_identity_partition_spec

Xuanwo (approved):
Thank you for working on this!

PR #17839, opened 2025-04-23
fix: wrong calc of #rows of page in http handler.

youngsofun:

## Summary

1. fix：when `block_memory_size > *remain_size` and the size limit is the dominant factor, we expect remain_size to be zero. However, in reality, it may be not zero due to integer division. （introduced in v1.2.717 https://github.com/databendlabs/databend/pull/17667）
2. optimize：call `block.memory_size(); `  only once

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17839)


KKould (approved):
LGTM

sundy-li (approved):

PR #17843, opened 2025-04-24
fix(query): align roles api with table functions

everpcpc:

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17843)


src/query/service/src/servers/http/v1/roles.rs
```diff
@@ -35,47 +37,34 @@ pub struct RoleInfo {
     pub is_default: bool,
 }
 
+/// same as `show_roles` in
+/// src/query/service/src/table_functions/show_roles/show_roles_table.rs
 #[async_backtrace::framed]
 async fn handle(ctx: &HttpQueryContext) -> Result<ListRolesResponse> {
-    let user = ctx.session.get_current_user()?;
+    let mut all_roles = ctx.session.get_all_available_roles().await?;
```
TCeason, line 44:
```rust
    async fn get_all_effective_roles(&self) -> Result<Vec<RoleInfo>> {
        let secondary_roles = self.session_ctx.get_secondary_roles();

        // if secondary_roles is not set, return all the available roles
        if secondary_roles.is_none() {
            let available_roles = self.get_all_available_roles().await?;
            return Ok(available_roles);
        }
        ...
}
secondary_roles is 

/// To SET SECONDARY ROLES ALL, the session will have all the roles take effect. On the other hand,
/// SET SECONDARY ROLES NONE will disable all the roles except the current role.
/// By default, the SECONDARY ROLES is ALL, which is None here. There are a few cases that the SECONDARY
/// ROLES is preferred to be empty, which is Some([]) here:
/// 1. The user comes from an external authenticator, which maps to a single role.
/// 2. The role is intentionally restricted by the sql client, to run SQLs with a restricted privileges.


So maybe get_all_effective_roles is better in this context? cc @flaneur2020

Comment (everpcpc) [reply src/query/service/src/servers/http/v1/roles.rs:44]: Then we should also use `get_all_effective_roles` for table function?

File: src/query/service/src/servers/http/v1/verify.rs [hunk 2]
```diff
@@ -35,9 +35,10 @@ pub async fn verify_handler(ctx: &HttpQueryContext) -> PoemResult<impl IntoRespo
         .session
         .get_current_user()
         .map_err(HttpErrorCode::server_error)?;
+    let roles = ctx.session.get_all_effective_roles().await?;
     Ok(Json(VerifyResponse {
         tenant: tenant.tenant_name().to_string(),
         user: user.name.to_string(),
-        roles: user.grants.roles(),
+        roles: roles.into_iter().map(|r| r.name).collect(),
```
TCeason, line 42: Maybe this also need to modify ? I think the roles maybe also need auth_role? cc @flaneur2020 

https://github.com/datafuselabs/databend/blob/1263e53ccbcf55bb5e77be611e1c016c55aa0405/src/query/service/src/servers/http/v1/users.rs#L74

```rust
#[async_backtrace::framed]
async fn handle(ctx: &HttpQueryContext) -> Result<ListUsersResponse> {
    let tenant = ctx.session.get_current_tenant();
    let user_api = UserApiProvider::instance();
    let mut users = vec![];
    for user in user_api.get_users(&tenant).await? {
        users.push(UserItem {
            name: user.name.clone(),
            hostname: user.hostname.clone(),
            auth_type: user.auth_info.get_type().to_str().to_string(),
            default_role: user.option.default_role().cloned().unwrap_or_default(),
            roles: user.grants.roles(),
            disabled: user.option.disabled().cloned().unwrap_or_default(),
            network_policy: user.option.network_policy().cloned(),
            password_policy: user.option.password_policy().cloned(),
            must_change_password: user.option.must_change_password().cloned(),
        });
    }
    Ok(ListUsersResponse { users })
}
```

TCeason (approved):

src/query/service/src/servers/http/v1/roles.rs
TCeason: After discussion : available is ok, check logic will call verify

PR #17852, opened 2025-04-25
feat: impl `NgramIndex` for `FuseTable`, improve like query performance

KKould:

## Summary

part of: https://github.com/databendlabs/databend/issues/17724

Implement Ngram Index to improve the retrieval speed of Like query

Its working principle is to insert String type data into multiple substrings in the form of ngram and insert them into BloomFilter. When querying Like, it determines whether there is a substring after ngram that does not exist in BloomFilter to filter out the Block that must not have data in Like in advance.

Therefore, when using Ngram Index, the insertion time will be longer due to ngram (depending on the length of each line of string and the total number of data lines).

# Storage
Ngram Index is essentially a data segmentation method based on Bloom Index using Ngram. Therefore, Ngram Index shares Meta with Bloom Index and uses the same storage file.

# Benchmark
Using amazon_reviews as the benchmark, the total data size is 39.2 GB, and review_body is 17 GB
```
CREATE OR REPLACE TABLE `amazon_reviews_ngram` (
                                  `review_date` int(11) NULL,
                                  `marketplace` varchar(20) NULL,
                                  `customer_id` bigint(20) NULL,
                                  `review_id` varchar(40) NULL,
                                  `product_id` varchar(10) NULL,
                                  `product_parent` bigint(20) NULL,
                                  `product_title` varchar(500) NULL,
                                  `product_category` varchar(50) NULL,
                                  `star_rating` smallint(6) NULL,
                                  `helpful_votes` int(11) NULL,
                                  `total_votes` int(11) NULL,
                                  `vine` boolean NULL,
                                  `verified_purchase` boolean NULL,
                                  `review_headline` varchar(500) NULL,
                                  `review_body` string NULL,
                                  NGRAM INDEX idx1 (review_body) gram_size = 10 bitmap_size = 2097152
) Engine = Fuse bloom_index_columns='review_body';

copy into amazon_reviews_ngram from @data/ngram_test/amazon_reviews_2010.snappy.parquet file_format = (type = PARQUET);
copy into amazon_reviews_ngram from @data/ngram_test/amazon_reviews_2011.snappy.parquet file_format = (type = PARQUET);
copy into amazon_reviews_ngram from @data/ngram_test/amazon_reviews_2012.snappy.parquet file_format = (type = PARQUET);
copy into amazon_reviews_ngram from @data/ngram_test/amazon_reviews_2013.snappy.parquet file_format = (type = PARQUET);
copy into amazon_reviews_ngram from @data/ngram_test/amazon_reviews_2014.snappy.parquet file_format = (type = PARQUET);
copy into amazon_reviews_ngram from @data/ngram_test/amazon_reviews_2015.snappy.parquet file_format = (type = PARQUET);
```

Using this SQL to test Ngram, the total file size of BloomFilter is 1.5 GB

## Query: 
```
SELECT
    product_id,
    any(product_title),
    AVG(star_rating) AS rating,
    COUNT() AS count
FROM
    amazon_reviews_ngram
WHERE
    review_body LIKE '%The first track with Chris Botti is beautiful%'
GROUP BY
    product_id
ORDER BY
    count DESC,
    rating DESC,
    product_id
    LIMIT 5;
```
Ngram: 
- 1 row read in 1.126 sec. Processed 786.43 thousand row, 444.15 MiB (698.43 thousand rows/s, 394.45 MiB/s)

Without Ngram: 
- 1 row read in 13.045 sec. Processed 135.59 million row, 52.91 GiB (10.39 million rows/s, 4.06 GiB/s)

## Insert:
Ngram: 
- 2010: 38.227 sec
- 2011: 46.212 sec
- 2012: 67.140 sec
- 2013: 112.430 sec
- 2014: 132.978 sec
- 2015: 102.655 sec

Without Ngram:
- 2010: 6.090 sec
- 2011: 6.468 sec
- 2012: 9.751 sec
- 2013: 15.562 sec
- 2014: 23.374 sec
- 2015: 14.587 sec

Tips: The factors that affect the insertion time are as follows:
- The length of each row of data
- Number of data rows
- BloomFilter Bitmap Size
- N (gram_size) of Ngram

Therefore, this benchmark is the parameter I chose for query purposes. In actual applications, users need to weigh the insertion speed and filtering effect.

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17852)


tests/sqllogictests/suites/ee/04_ee_inverted_index/04_0000_inverted_index_base.test
```diff
@@ -476,7 +476,7 @@ idx2 INVERTED books(title, author, description)index_record='"basic"' tokenizer=
 query III
 select row_count, bloom_filter_size, inverted_index_size from fuse_block('test_index', 't1')
 ----
-10 438 2390
+10 439 2390
```
KKould, line 479: add a bit to distinguish whether the filter is Xor8 or Bloom

tests/sqllogictests/suites/base/09_fuse_engine/09_0006_func_fuse_history.test
```diff
@@ -21,7 +21,7 @@ insert into t values (1)
 query III
 select block_count, row_count, index_size from fuse_snapshot('db_09_0006', 't') order by row_count desc limit 1
 ----
-1 1 0
```
KKould, line 24: Why is the index size 0? The bloom filter is created by default, so after inserting a row of data, the index size will always be greater than 1 in theory.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -347,6 +357,50 @@ impl BloomIndex {
...
+        fn_call: F,
+    ) -> impl Iterator<Item = Vec<T>> + 'a
+    where
+        F: Fn(&str) -> T + 'a,
+    {
+        (0..arg.len()).filter_map(move |i| {
+            arg.index(i).and_then(|scalar| {
+                scalar.as_string().and_then(|text| {
+                    let text = text.to_lowercase();
+                    if text.is_empty() || gram_size == 0 {
```
b41sh, line 372: `gram_size` can't be zero, it's invalid argument.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -399,12 +453,18 @@ impl BloomIndex {
         }
     }
 
+    pub fn build_filter_ngram_name(field: &TableField) -> String {
+        format!("Ngram({})", field.column_id())
```
b41sh, line 457: Ngram field name should also contain `gram_size`, because user may modify this value in table index, which will resulting in the already created ngram index not being filtered correctly.

KKould: But build_filter_ngram_name is difficult to obtain the gram_size parameter for NgramArg

Because when the corresponding NgramArg is obtained, its column name is needed for search, but the combination of column name requires the gram_szie of NgramArg
- https://github.com/databendlabs/databend/pull/17852/files#diff-3d9a5741355c92e4764a46869c567fda641956263edd4fba0bbe742142db348cR168:~:text=acc.push(BloomIndex%3A%3Abuild_filter_ngram_name(field))%3B
- https://github.com/databendlabs/databend/pull/17852/files#diff-ccbfd55cd2c4c1adde6621b0da13070ad2f24c9c8a5806628c990a7b54c68a52R1062:~:text=BloomIndex%3A%3Abuild_filter_ngram_name(table_field)

tests/sqllogictests/suites/ee/04_ee_inverted_index/04_0000_inverted_index_base.test
sundy-li: Does it break the old bloom filter index compatibility?

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -347,6 +357,50 @@ impl BloomIndex {
...
+                        words.push(fn_call(&text[start..end]));
+                    }
+                    Some(words)
+                })
+            })
+        })
+    }
+
+    pub fn ngram_hash(s: &str) -> u64 {
+        cityhash64_with_seed(s.as_bytes(), 1575457558)
```
sundy-li, line 401: Better put `1575457558 ` as constant static value

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -641,129 +784,172 @@ where T: EqVisitor
...
+                    ..
+                }), Expr::Constant(Constant { scalar, .. })]
+                | [Expr::Constant(Constant { scalar, .. }), Expr::ColumnRef(ColumnRef {
+                    id,
+                    data_type: column_type,
+                    ..
+                })] => {
+                    if let Some(pattern) = scalar.as_string() {
+                        if let LikePattern::SurroundByPercent(v) =
+                            generate_like_pattern(pattern.as_bytes(), 1)
```
sundy-li, line 807: StartOfPercent and EndOfPercent should also encounter this branch

tests/sqllogictests/suites/ee/04_ee_inverted_index/04_0000_inverted_index_base.test
KKould: Yes, so we are currently trying other methods that do not change the original serialization

KKould:
Please note that the filter has been adjusted: the original BloomFilter has been removed, and the size is controlled by taking the remainder using Xor8Filter. This may have a significant impact on the benchmark, and it still needs to be tested.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -347,6 +365,58 @@ impl BloomIndex {
...
+        fn_call: F,
+    ) -> impl Iterator<Item = Vec<T>> + 'a
+    where
+        F: Fn(&str, usize) -> T + 'a,
+    {
+        (0..arg.len()).filter_map(move |i| {
+            arg.index(i).and_then(|scalar| {
+                scalar.as_string().and_then(|text| {
+                    let text = text.to_lowercase();
+                    if text.is_empty() {
```
b41sh, line 381: we can check `gram_size` and char count first
```rust
if text.is_empty() || gram_size > text.chars().count() {
    return None;
}
```

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -453,31 +586,85 @@ impl BloomIndex {
 
 pub struct BloomIndexBuilder {
     func_ctx: FunctionContext,
-    columns: Vec<ColumnXor8Builder>,
+    columns: Vec<ColumnFilterBuilder>,
```
b41sh, line 589: better to define two different fields for bloom filter and ngram bloom filter.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -575,9 +768,36 @@ impl BloomIndexBuilder {
...
+                index_column.builder.add_digests(it);
             } else {
-                bloom_index_column.builder.add_digests(column.deref());
+                index_column.builder.add_digests(column.deref());
+            }
+        }
+        for (index, index_column) in ngram_iter {
+            let field_type = &block.get_by_offset(index_column.index).data_type;
+            if !Xor8Filter::supported_type(field_type) {
+                keys_to_remove.push(index);
```
b41sh, line 779: don't need check type here, ngram filter must be a string column.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -641,129 +868,185 @@ where T: EqVisitor
...
-                // debug_assert_eq!(scalar_type, column_type);
-                // If the visitor returns a new expression, then replace with the current expression.
-                if scalar_type == column_type {
-                    self.0
-                        .enter_target(*span, id, scalar, column_type, return_type)?
-                } else {
-                    ControlFlow::Continue(None)
+        if id.name() == "like" {
+            match args.as_slice() {
+                // patterns like `Column = <constant>`, `<constant> = Column`
```
b41sh, line 878: this comment is not correct.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -641,129 +868,185 @@ where T: EqVisitor
...
+                [Expr::ColumnRef(ColumnRef {
+                    id,
+                    data_type: column_type,
+                    ..
+                }), Expr::Constant(Constant { scalar, .. })]
+                | [Expr::Constant(Constant { scalar, .. }), Expr::ColumnRef(ColumnRef {
```
b41sh, line 884: only support left argument is a column case.

src/query/storages/common/index/tests/it/bloom_pruner.rs
```diff
@@ -477,17 +532,44 @@ fn eval_index_expr(
...
+        for words in BloomIndex::calculate_ngram_nullable_column(
+            Value::Scalar(scalar),
+            ngram_arg.gram_size(),
+            ngram_arg.bitmap_size(),
+            |w, _| w.to_string(),
+        ) {
+            for word in words {
+                let hash = BloomIndex::ngram_hash(&word, ngram_arg.bitmap_size());
+
+                scalar_map.entry(Scalar::String(word)).or_insert(hash);
```
b41sh, line 566: don't put ngram scalar string in `scalar_map`, the word may be conflict with scalar string in bloom scalar, which will result in wrong result.
add a new map to store ngram scalar string map and use the string as the key, not the splitted word, so that we don't need to caculate hash values in each blocks.

src/query/storages/common/index/src/bloom_index.rs
```diff
@@ -1023,24 +1298,46 @@ impl EqVisitor for ShortListVisitor {
             return Ok(ControlFlow::Continue(None));
         };
 
-        let Some(field) = self.found_field(id) else {
-            return Ok(ControlFlow::Break(None));
-        };
-        if !Xor8Filter::supported_type(src_type) || !is_injective_cast(src_type, dest_type) {
-            return Ok(ControlFlow::Break(None));
-        }
+        if is_like {
```
b41sh, line 1301: `like` don't support cast, because only string type have ngram index.

src/query/storages/common/index/src/filters/xor8/bloom_filter.rs
```diff
@@ -0,0 +1,288 @@
...
+        for hash in mem::take(&mut self.inner) {
+            filter.add(hash);
+        }
+        Ok(filter)
+    }
+}
+
+#[derive(Clone)]
+pub struct BloomFilter {
+    size: usize,
```
b41sh, line 107: use `u64` instead of `usize`, because the size of `usize` is related to the platform, it may cause error during deserialization

src/query/storages/common/index/src/filters/xor8/mod.rs
```diff
@@ -14,12 +14,135 @@
...
+                let mut bytes = filter.to_bytes()?;
+                bytes.splice(0..0, "bloom".as_bytes().iter().cloned());
+                bytes
+            }
+        })
+    }
+
+    fn from_bytes(buf: &[u8]) -> Result<(Self, usize), Self::CodecError> {
+        Ok(if buf.starts_with("bloom".as_bytes()) {
+            BloomFilter::from_bytes(&buf[5..])
```
sundy-li, line 108: what if the old xor filter starts with `bloom` ?

I think we can read the first byte to indicate which kind of the filter.


Look at the `Cbor::decode` of `decode_hdr`

`major` ranges from [0, 7] is `Xor8Filter`

Let's check the first byte, if it's `n`, we can take it as Ngram, otherwise fallback to use `Xor`

dantengsky:
LGTM. One concern to consider for future optimization work:

- Hash derivation strategy in BloomFilter::find/add methods

  Perhaps the Kirsch-Mitzenmacher approach:
  "Given two hash functions h_1(x) and h_2(x), an i-th additional hash function g_i(x) can be simulated as g_i(x) = h_1(x) + i * h_2(x)"
  could provide a lower false positive rate in ngram scenario?

dantengsky (approved):

PR #17864, opened 2025-04-28
feat: unload preserve order if the source is ordered.

youngsofun:

## Summary

for unloading like `copy into <location> from (select ...  order by ... ) `  
order are preserve in result files 

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17864)


src/query/sql/src/planner/binder/copy_into_location.rs
```diff
@@ -109,6 +109,15 @@ impl Binder {
                     .await
             }
         }?;
+        println!("qyery pln: {:?}", &query);
```
sundy-li, line 112: need add `git diff | grep print` in git commit hook


```
cd .git/hooks
 touch pre-commit
 chmod +x pre-commit 
```

```
  #!/bin/bash

   # Check for 'print' in the diff
   if git diff --cached | grep -q 'print'; then
       echo "Error: Commit contains 'print' statements."
       exit 1  # Exit with a non-zero status to prevent the commit
   fi
```

sundy-li:
> order are preserve in result files by (file_name, row_number)

The description is wrong, orders are kept by custom expression.

src/query/sql/src/planner/binder/copy_into_location.rs
```diff
@@ -109,6 +109,15 @@ impl Binder {
...
+        println!("qyery pln: {}", &query);
+        let mut is_ordered = false;
+        if let Plan::Query { s_expr, .. } = &query {
+            let p = s_expr.derive_relational_prop()?;
+            if !p.orderings.is_empty() || p.partition_orderings.is_some() {
```
sundy-li, line 117: `if !p.orderings.is_empty() `

partition_orderings is not used for keeping order

sundy-li (approved):
ALMOST LGTM

src/query/catalog/src/plan/datasource/datasource_info/stage.rs
```diff
@@ -42,9 +46,11 @@ pub struct StageTableInfo {
     // - may need to be purged as well (depends on the copy options)
     pub duplicated_files_detected: Vec<String>,
     pub is_select: bool,
-    pub copy_into_location_options: CopyIntoLocationOptions,
     pub copy_into_table_options: CopyIntoTableOptions,
-    pub stage_root: String,
+
+    // copy into location only
+    pub copy_into_location_ordered: bool,
+    pub copy_into_location_options: CopyIntoLocationOptions,
```
BohuTANG, line 53:
```suggestion
    pub copy_into_location_ordered: bool,
    pub copy_into_location_options: CopyIntoLocationOptions,
```
This structure marked as Serialize/Deserialize, will this new field break existing compatibility?

youngsofun: no problem since it is only for transfer, not stored

src/query/sql/src/planner/binder/copy_into_location.rs
youngsofun: done

youngsofun:
> > order are preserve in result files by (file_name, row_number)
> 
> The description is wrong, orders are kept by custom expression.

I intend to say it is not only sorted in file, but between files.

sundy-li (approved):

PR #17920, opened 2025-05-12
chore(query): add some function alias

sundy-li:

## Summary

Below is a brief changelog summary for the changes:

1. support `create or replace` for task 
2. some functions behavior changes:
 - iff alias for if function
 - `greatest` and `least` will return error if input args contains null
 -  add `greatest_ignore_nulls` and `least_ignore_nulls` which ignore null (works as the previous behavior of greatest and least)
 - add `var_pop`, `variance_pop` alias for function `covar_pop`
 -  add `var_samp`, `variance_samp`` alias for function `covar_samp`
 -  support `insert overwrite into` syntax, works same as `insert overwrite`
 
## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [x] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17920)

### DIFF
--- a/src/query/sql/src/planner/semantic/type_check.rs
+++ b/src/query/sql/src/planner/semantic/type_check.rs
@@ -3423,6 +3423,7 @@ impl<'a> TypeChecker<'a> {
             Ascii::new("connection_id"),
             Ascii::new("timezone"),
             Ascii::new("nullif"),
+            Ascii::new("iff"),
             Ascii::new("ifnull"),
             Ascii::new("nvl"),
             Ascii::new("nvl2"),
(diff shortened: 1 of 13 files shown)


src/query/sql/src/planner/semantic/type_check.rs
```diff
@@ -3509,6 +3512,7 @@ impl<'a> TypeChecker<'a> {
                     arg_x,
                 ]))
             }
+            ("iff", args) => Some(self.resolve_function(span, "if", vec![], args)),
```
b41sh, line 3515: better to registe alias function name in this file `src/query/functions/src/scalars/control.rs`
```rust
registry.register_aliases("if", &["iff"]);
```

sundy-li: evaluator judge `if name == "if"` in multi times, so we don't introduce function alias in expression level.

Issue #17928, opened 2025-05-14
bug: `InferFilterOptimizer` generate invalid filter expr with wrong data type


b41sh:
### Search before asking

- [x] I had searched in the [issues](https://github.com/databendlabs/databend/issues) and found no similar issues.

### Version

main

### What's Wrong?

`InferFilterOptimizer` generate invalid filter expr with wrong data type

### How to Reproduce?

```sql
root@0.0.0.0:48000/default/default> create table t1 (a int, b int);
root@0.0.0.0:48000/default/default> create table t2 (a int, b json);

root@0.0.0.0:48000/default/default> select * from t2 where a = b and strip_null_value(b) is not null;
error: APIError: QueryFailed: [1065]no function matches signature `strip_null_value(Int32 NULL)`, you might need to add explicit type casts.

candidate functions:
  strip_null_value(Variant NULL) :: Variant NULL  : unable to unify `Int32` with `Variant`

root@0.0.0.0:48000/default/default> select * from t1 join t2 on t1.b = t2.b where strip_null_value(t2.b) is not null;
error: APIError: QueryFailed: [1065]no function matches signature `strip_null_value(Int32 NULL)`, you might need to add explicit type casts.

candidate functions:
  strip_null_value(Variant NULL) :: Variant NULL  : unable to unify `Int32` with `Variant`
```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #17929, opened 2025-05-14
fix(query): Prevent invalid filter expression generation in `InferFilterOptimizer`

b41sh:

## Summary

This PR addresses a bug in the `InferFilterOptimizer` that could lead to the generation of invalid filter expressions, causing query execution failures.

**Background:**

The `InferFilterOptimizer` aims to improve query performance by inferring additional filter conditions and pushing them down to the scan operator. This allows for earlier filtering of data, reducing the amount of data processed during the join. For example, the condition `t1.a = t2.a and t1.a = 5` can be optimized to `t1.a = t2.a and t1.a = 5 and t2.a = 5`.

**The Issue:**

The previous implementation lacked type checking during the filter condition inference process. This could result in the creation of expressions that are incompatible with the underlying data types.

Consider the following scenario: `t1.a = t2.a and strip_null_value(t1.a) is not null`. The optimizer might incorrectly infer `t1.a = t2.a and strip_null_value(t1.a) is not null and strip_null_value(t2.a) is not null`. If `t1.a` is a `VARIANT` column and `t2.a` is an `INT` column, the `strip_null_value(t2.a)` expression will fail because the `strip_null_value` function only supports `VARIANT` arguments.

**The Fix:**

This PR introduces type checking within the `InferFilterOptimizer`. The optimizer now verifies that the inferred expressions are compatible with the data types of the involved columns before generating new filter conditions. This prevents the creation of invalid expressions and resolves the bug.

- fixes: #17928

## Tests

- [x] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17929)


src/query/sql/src/planner/optimizer/optimizers/operator/filter/infer_filter.rs
```diff
@@ -188,6 +199,28 @@ impl<'a> InferFilterOptimizer<'a> {
             Some(index) => self.expr_equal_to[*index].push(left.clone()),
             None => self.add_expr(right, vec![], vec![left.clone()]),
         };
+
+        true
+    }
+
+    // equal expr must have the same type, otherwise the function may fail on execution.
+    fn check_equal_expr_type(left_ty: &DataType, right_ty: &DataType) -> bool {
```
BohuTANG, line 207: Do we need(I am not sure other places has) some infer types compare?
Like Number Type cmp  Decimal Type?

b41sh: Yes, you're right that this approach might miss some potential optimizations. But the functions supporting `Number` types might not be directly applicable to `Decimal` types. Fully supporting these kinds of cross-type inferences would require a more complex logic to ensure type compatibility. I haven't yet devised a clean and robust solution for handling these scenarios. We can consider exploring more advanced type inference strategies as a future enhancement.

BohuTANG: This PR is OK, let's tracking this in new issue.

BohuTANG (approved):

Issue #17936, opened 2025-05-14
Feature: support async function for merge not match clause


zhang2014:
**Issue Summary**

**Problem:**  
When executing a `MERGE` statement with an async function in the "NOT MATCHED" clause, the query fails with an index out of bounds error.  

**Steps to Reproduce:**  
 1. Create sequence:
```sql
CREATE SEQUENCE test_seq;
```
2. Create target and source tables:  
   ```sql
   CREATE TABLE target_table (id int default nextval(test_seq), name varchar(50), value int);
   CREATE TABLE source_table (id INT, name VARCHAR(50), value INT);
   ```  
3. Insert data into the source table:  
   ```sql
   INSERT INTO source_table (name, value) VALUES ('Item1', 100);
   ```  
4. Execute the `MERGE` statement:  
   ```sql
   MERGE INTO target_table AS target 
   USING source_table AS source 
   ON target.id = source.id 
   WHEN MATCHED THEN 
      UPDATE SET target.name = source.name, target.value = source.value 
   WHEN NOT MATCHED THEN 
      INSERT (name, value) VALUES (source.name, source.value);
   ```  

**Observed Error:**  
```
APIError: QueryFailed: [1104] index out of bounds: the len is 2 but the index is 18446744073709551610
```  

**Expected Behavior:**  
The `MERGE` statement should execute successfully, inserting or updating records as specified.  

**Root Cause:**  
The databend does not properly support async functions (e.g., sequence nextval()) in the NOT MATCHED clause of MERGE.

PR #17670, opened 2025-03-28
feat: support nextval as field default value.

youngsofun:
I hereby agree to the terms of the CLA available at: https://docs.databend.com/dev/policies/cla/

## Summary

support  `nextval(a_seq_name)` as field defaut value
    
```
CREATE SEQUENCE user_id_seq;

CREATE TABLE users (
  id INT DEFAULT NEXTVAL(user_id_seq),
  name VARCHAR
);
```

limition:
     
1. not support  expr with nextval as a part, e.g. `nextval(seq1) + 1`
2. when copy into table the above `user` from stage files
    1. if `copy into user(name)...`, column `id` will be generated
    2. if `copy into user(id, name)...`, column `id` will come from data files,  currently require every row contains column `id`

fixes https://github.com/databendlabs/databend/issues/17616

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [x] New Feature (non-breaking change which adds functionality)
- [ ] Breaking Change (fix or feature that could cause existing functionality not to work as expected)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17670)

PR #17941, opened 2025-05-15
feat: support async function for merge not match clause

SkyFan2002:

## Summary
Reused the `DefaultExprBinder` introduced in #17670 and implemented `TransformBranchedAsyncFunction` to handle `SourceSchemaIndex` in merge into.
- fixes: #17936

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17941)


src/query/service/src/pipelines/processors/transforms/transform_branched_async_function.rs
```diff
@@ -0,0 +1,94 @@
...
+                        &mut block,
+                        sequence_name,
+                        &async_func_desc.data_type,
+                    )
+                    .await?;
+                }
+                AsyncFunctionArgument::DictGetFunction(_) => unreachable!(),
+            }
+        }
+        let block = cast_schema(
```
youngsofun, line 84: cast can be skipped  too if from==to

zhang2014 (approved):

tests/sqllogictests/suites/base/09_fuse_engine/09_0048_issue_17936.sql
```diff
@@ -0,0 +1,44 @@
+statement ok
+CREATE OR REPLACE SEQUENCE test_seq;
+
+statement ok
```
youngsofun, line 4: we can add another column of uint64 not null，which cover the code path where no cast needed，my lasr pr miss this too. 

by the way, nextval is uint64, int may be not big enough?  all cases in docs also use int: https://docs.databend.com/sql/sql-functions/sequence-functions/nextval
https://docs.databend.com/sql/sql-commands/ddl/table/ddl-create-table#column-default-values

youngsofun (approved):

PR #17942, opened 2025-05-15
feat(parse): Simplify the matching pattern when parse function, avoid exponential backtracking

KKould:

## Summary
Databend may take a very long time to parse multiple layers of function calls
```sql
select ROUND(6378.138 * 2 * ASIN(SQRT(POW(SIN((IF(1 = 2, NULL, 10) * PI() / 180 - 40.051816 * PI() / 180) / 2), 2))) * 1000, 0) distance;

Before: 
1 row read in 25.504 sec. Processed 1 row, 1 B (0.04 row/s, 0 B/s)

After:
1 row read in 1.115 sec. Processed 1 row, 1 B (0.90 row/s, 0 B/s)

```
This is because there are many similar matching patterns when parsing function calls, and simple function calls need to fail to match complex expressions before entering. This will cause multiple nested simple function calls to fail to match correctly (the matching order is the last pattern).
94b9082cd84cc97d4fde9e3fe2a24b9

This PR simplifies the matching patterns in function calls as much as possible and rearranges the order to avoid this situation.

## Tests

- [ ] Unit Test
- [ ] Logic Test
- [x] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17942)

### DIFF
--- a/src/query/ast/benches/bench.rs
+++ b/src/query/ast/benches/bench.rs
@@ -16,12 +16,14 @@ fn main() {
     divan::main()
 }
 
-// bench                  fastest       │ slowest       │ median        │ mean          │ samples │ iters
-// ╰─ dummy                             │               │               │               │         │
-//    ├─ deep_query       122 µs        │ 324.3 µs      │ 127.2 µs      │ 130.1 µs      │ 100     │ 100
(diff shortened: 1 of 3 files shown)


sundy-li (approved):
Impressive performance

PR #17947, opened 2025-05-15
feat: Use query id as trace id

Xuanwo:

## Summary

Use the same ID for both the query ID and the trace ID.

The OTEL specification requires the trace ID to be 128 bits (32 hex chars). To enable us to use our query ID as the trace ID, I have changed the format of our query ID from `67e55044-10b1-426f-9247-bb680e5fe0c8` to `67e5504410b1426f9247bb680e5fe0c8`.

I believe this should be a non-breaking change, as it is unlikely that any system depends on the specific representation of the query ID. However, we should still review this carefully to ensure we haven't introduced any unexpected issues.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17947)


sundy-li (approved):

hantmac:
Driver will gennerate `query-id` as the http header:
- Go driver: https://github.com/databendlabs/databend-go/blob/main/connection.go#L195
- JDBC: https://github.com/databendlabs/databend-jdbc/blob/main/databend-jdbc/src/main/java/com/databend/jdbc/DatabendConnection.java#L709

hantmac:
For Databend query, after implementing a compatibility feature that removes all hyphens ("-") from the received query ID, and use this new query ID and trace ID for downstream transmission. Then we can merge the following pr in clients:

- JDBC: https://github.com/databendlabs/databend-jdbc/pull/317
- Go: https://github.com/databendlabs/databend-go/pull/152
- Rust: https://github.com/databendlabs/bendsql/pull/623

Xuanwo:
> For Databend query, after implementing a compatibility feature that removes all hyphens ("-") from the received query ID, and use this new query ID and trace ID for downstream transmission. Then we can merge the following pr in clients:

Thank you for this! I will work on this PR directly.

everpcpc (approved):

youngsofun (approved):

PR #17953, opened 2025-05-17
feat: support select from AVRO files.

youngsofun:

## Summary

# Complete Avro File Loading Support

In https://github.com/databendlabs/databend/pull/17548 we introduced the `COPY INTO <table>` feature for Avro files. This PR completes the Avro loading functionality by adding support for:

1. Direct querying of Avro files
2. COPY with transformation
3. Accessing metadata columns `metadata$filename` and `metadata$file_row_number`

## Summary of Avro Loading Features

After these changes, Avro files behave very similarly to NDJSON files when reading:

1. **COPY without transformation**:
   - Assumes top-level schema is a record and matches fields by name
   - More efficient for bulk loading
   - Doesn't require all files to have identical schema:
     - Field order doesn't matter
     - Supports file format option `missing_field_as = ERROR | FIELD_DEFAULT` (including nested records)
     - Supports loading different Avro types into one Databend field via auto-casting

2. **Query and COPY with transformation**:
   - Treats each row as a variant referenced by `$1`
   - Each Avro type is mapped to a variant type
   - More flexible and stable as type mappings are less likely to change

## Examples

Given an Avro file with schema `user` and table `user(id int, name string)`:

```json
{
  "type": "record",
  "name": "user",
  "fields": [
    {
      "name": "id",
      "type": "long"
    },
    {
      "name": "name",
      "type": "string"
    }
  ]
}
```
SQL exmaples:

```
-- Basic COPY
COPY INTO user FROM @stage file_format=(type=avro);

-- Query with metadata
SELECT 
    metadata$filename as file,
    metadata$file_row_number as row, 
    cast($1:id as int) as id, 
    $1:name as name
FROM @stage(file_format=>'avro');

-- COPY with transformation
COPY INTO user FROM (
    SELECT 
        $1:id::int as id,
        $1:name as name
    FROM @stage(file_format=>'avro')
) file_format=(type=avro);
```

## Type Mapping to Variant

Variants are stored as JSONB in Databend. While most mappings are straightforward, some special cases to note:

- Time Types:
   - JSONB doesn't have a Time type, so TimeMillis and TimeMicros are mapped to int64
   - Users should keep the original type in mind when working with these values
- Decimal Types:
   - Decimals will be loaded as either decimal128 or decimal256
   - May report an error if the precision is too large
- Enum Types:
   - Mapped to string values

## Tests

- [ ] Unit Test
- [x] Logic Test
- [ ] Benchmark Test
- [ ] No Test - _Explain why_

## Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17953)

### DIFF
--- a/src/query/storages/stage/src/read/avro/avro_to_jsonb.rs
+++ b/src/query/storages/stage/src/read/avro/avro_to_jsonb.rs
@@ -15,54 +15,126 @@
 use std::collections::BTreeMap;
 
 use apache_avro::types::Value;
+use apache_avro::Schema;
+use databend_common_expression::types::i256;
+use databend_common_expression::types::Decimal;
+use databend_common_expression::types::MAX_DECIMAL128_PRECISION;
+use databend_common_expression::types::MAX_DECIMAL256_PRECISION;
+use num_bigint::BigInt;
 
-pub(super) fn to_jsonb(value: &Value) -> Result<jsonb::Value, String> {
-    let jvalue = match value {
-        Value::Null => jsonb::Value::Null,
-        Value::Boolean(v) => jsonb::Value::Bool(*v),
-        Value::Int(v) => jsonb::Value::from(*v),
-        Value::Long(v) => jsonb::Value::from(*v),
-        Value::Float(v) => jsonb::Value::from(*v),
-        Value::Double(v) => jsonb::Value::from(*v),
-        Value::String(v) => jsonb::Value::from(v.as_str()),
-        Value::Enum(_, v) => jsonb::Value::from(v.as_str()),
-        Value::Uuid(v) => jsonb::Value::from(v.to_string()),
-        Value::Union(_, v) => to_jsonb(v)?,
-        Value::Array(v) => {
+pub(super) fn to_jsonb<'a>(value: &'a Value, schema: &Schema) -> Result<jsonb::Value<'a>, String> {
+    let jvalue = match (value, schema) {
+        // primaries
+        (Value::Null, Schema::Null) => jsonb::Value::Null,
+        (Value::Boolean(v), Schema::Boolean) => jsonb::Value::Bool(*v),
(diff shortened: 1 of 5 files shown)


src/query/storages/stage/src/read/avro/avro_to_jsonb.rs
```diff
@@ -15,54 +15,126 @@
...
-        Value::Int(v) => jsonb::Value::from(*v),
-        Value::Long(v) => jsonb::Value::from(*v),
-        Value::Float(v) => jsonb::Value::from(*v),
-        Value::Double(v) => jsonb::Value::from(*v),
-        Value::String(v) => jsonb::Value::from(v.as_str()),
-        Value::Enum(_, v) => jsonb::Value::from(v.as_str()),
-        Value::Uuid(v) => jsonb::Value::from(v.to_string()),
-        Value::Union(_, v) => to_jsonb(v)?,
-        Value::Array(v) => {
+pub(super) fn to_jsonb<'a>(value: &'a Value, schema: &Schema) -> Result<jsonb::Value<'a>, String> {
```
b41sh, line 25: better to add some unit test for this function to ensure that the value can be correctly converted to jsonb format value and can be converted back to the original value.

youngsofun: I will add some test later, but not all type can be converted back.

youngsofun: add tests mainly for decimal,
for the other types, unit test code are almost the same as the code being tests.
the will be tested in logic test after cast functions are add https://github.com/databendlabs/databend/issues/17940

b41sh (approved):

PR #17969, opened 2025-05-20
refactor(metactl): when restore without --initial-cluster, add default single node cluster config

drmingdrmer:

##### refactor(metactl): when restore without --initial-cluster, add default single node cluster config

When using `databend-metactl import` without `--initial-cluster`, it
should fill in a default cluster config which has only one nodes. So
that a default `import` command can just build a working cluster of one
node.

This feature is to improve usability of `bendsave`.

Other changes: rewrite meta new-cluster test by python.

## Tests

- [x] Unit Test
- [ ] Logic Test
- [ ] Benchmark Test

## Type of change

- [x] Other

## Related Issues

- - -
This change is [Reviewable](https://reviewable.io/reviews/databendlabs/databend/17969)


everpcpc (approved):

Xuanwo (approved):
Thank you for fixing this!

